PR #259, opened 2023-08-17
deps

cbini:
## Merge Pre-Check
- [ ] <kbd>Format</kbd> has been run on all requested files
- [ ] `dbt build` successfully run for the entire project[^1]

[^1]: If only requesting 1 file, it's acceptable to run build only for that and its downstream dependencies.

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -219,15 +219,15 @@ files = [
 
 [[package]]
 name = "click"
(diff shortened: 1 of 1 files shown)


kevinverhoff (approved):
I have no idea what's going on here. Seems like it's good? @cbini when you're back, can you explain what this is?

PR #297, opened 2023-09-18
Mg tax fix again

cristinabaldor:
# Pull Request

## Merge Pre-Check

- [x] <kbd>Format</kbd> has been run on all requested files
- [x] `dbt build` successfully run for the entire project[^1]

[^1]:
    If only requesting 1 file, it's acceptable to run build only for that and its downstream
    dependencies.


kevinverhoff (approved):
Moving to regex makes sense. I'm going to approve for now, but is there a risk that we will end up getting additional tag_name_shorts down the line because we're removing that where ?

PR #2296, opened 2024-12-10
Gradebook audit new dm test

GabyRangelB:
# Pull Request

## Summary & Motivation

## Self-review

- [x] Update **due date**, **assignee**, and **priority** on our

  _If this is a same-day request, please flag that in Slack_

- [x] <kbd>Format</kbd> has been run on all modified files

- [x] Ensure you are using the `union_dataset_join_clause()` macro for queries that employ any
      models from these datasets:
  - deanslist
  - edplan
  - iready
  - pearson
  - powerschool
  - renlearn
  - titan

## Troubleshooting

---
  - https://app.asana.com/0/0/1208948220684703


src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
```diff
@@ -0,0 +1,197 @@
...
+            teacher_running_total_assign_by_cat,
+            teacher_avg_score_for_assign_per_class_section_and_assign_id,
+
+            audit_category,
+            cte_grouping,
+            audit_flag_name,
+            audit_flag_value,
+
+        from {{ ref("int_tableau__gradebook_audit_final_roster") }}
+        group by all
```
cbini, line 63: remove: unnecessary

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
```diff
@@ -0,0 +1,197 @@
...
+            is_missing,
+
+            audit_category,
+            cte_grouping,
+            audit_flag_name,
+            audit_flag_value as flag_value,
+
+        from {{ ref("int_tableau__gradebook_audit_final_roster") }}
+        where audit_flag_value = 1
+        group by all
```
cbini, line 133: remove: unnecessary

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
```diff
@@ -0,0 +1,197 @@
...
+    v.score_entered,
+    v.assign_final_score_percent,
+    v.is_exempt,
+    v.is_late,
+    v.is_missing,
+
+    v.flag_value,
+
+from teacher_aggs as t
+left join
```
cbini, line 184: I don't think this does anything. You're joining `int_tableau__gradebook_audit_final_roster` where `audit_flag_value = 1`. You should be able to work with `int_tableau__gradebook_audit_final_roster` as-is

src/dbt/kipptaf/models/extracts/tableau/intermediate/int_tableau__gradebook_audit_flags.sql
cbini, line 1: replace `'' as ...` with `null as ...`

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
cbini, line 1: If we're deleting `rpt_tableau__gradebook_audit`, we should rename this to that

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
GabyRangelB: it is necessary. with the group by all, the total row count is 1,860,687. without it, 25,946,277. the purpose of this CTE is to generate an aggregated view of all possible flags from the teacher perspective. int_tableau__gradebook_audit_final_roster (correctly) lists ALL possible errors from a teacher AND a student perspective. the request was to list all possible flags for the teacher, and only actual flags for the students.

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
GabyRangelB: it does exactly what the request is. teacher_aggs gives me a master list of all teachers with possible errors by category/flag, even if they dont actually have that. the valid flags gives me only the errors. therefore, i am able to calculate % errors and % gradebooks complete at a school with the correct N from teacher_aggs, AND provide the errors only from valid_flags.

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
GabyRangelB: not yet. i need to swap the sources on the tableau workbook and redo all of the calculations to fit the new model. this will take me 1-2 days. we cannot have the dashboard go down, especially now that EOQ report card auditing is happening.

src/dbt/kipptaf/models/extracts/tableau/intermediate/int_tableau__gradebook_audit_flags.sql
GabyRangelB: i specifically have '' because otherwise i get a data type mismatch error for the union. DBT wants me to use NULL for ints, and '' for non-numeric

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
GabyRangelB: let me review this one. the row count of group_by_all is 726,898. the non-group_by_all is  727,719

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
cbini: OK we can revisit

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
cbini: there are no aggregate functions being used, so it's unclear what's actually happening here and why

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
cbini: there are no aggregate functions being used, so it's unclear what's actually happening here and why

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
cbini: Yeah, this comes from me confusion above

src/dbt/kipptaf/models/extracts/tableau/intermediate/int_tableau__gradebook_audit_flags.sql
cbini: Looking into this myself

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
GabyRangelB: it is removing the student rows, which cause duplicates i dont need. i could use distinct, but it is usually not allowed. group by all seemed to be faster and tableau didnt complain about it like it did with distinct

src/dbt/kipptaf/models/extracts/tableau/intermediate/int_tableau__gradebook_audit_flags.sql
GabyRangelB: idk where exactly, but if you review the commits, you will find the countless commits i failed because dbt didnt like the column type mismatches. actually... they might have been the commits from 1/22 around midnight

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
cbini: `distinct` goes against best practices when it's used to mask structural issues upstream. If using `group by` without an explicit aggregation effectively does the same thing, it's equivalently bad.

I encourage you to figure out precisely what the remaining issues with the row scaffold you're using and address them directly, but if this produces the output you need, we can merge this view as-is since this is just a reporting view and has no downstream dependencies. However, troubleshooting any future issues will be much more difficult as a result.

cbini (approved):

cbini:
@GabyRangelB this is approved, but I'm not sure if you're still working on anything. You should be able to merge this if you're ready.

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v2.sql
GabyRangelB: got it. i think the missing piece here is that scaffold you created doesnt support all the needs of the report. not throwing shade at all, but when you simplified the model, an assumption was made that all assignment flags could come from student assignments, and all teacher flags could come from teacher assignments, and that the joins could all be assignment_category_code. 

however, if you review the gradebook flag g-sheet that is used as reference, you will see that the column called code is NOT just for W, F, and S (thats why i called it code, and not gradebook_category_code). code sometimes holds Q1 to Q4, AND there is an additional field called grade level because some flags apply only to certain grade levels. the majority of these flags do not use the category field because: a) not needed and b) causes dups.

since all these flags were essentially incorrectly omitted by the new scaffold you created, and i needed to put them back somehow, this was the best solution i could do that still used your scaffold AND brought the millions of rows that were left out. if you wanna give the intent of the data model a second look and create new scaffolds for the flags that were left out, let's discuss. for now, i agree that we need to move this fwd AND we probably want to wait until we get new marching orders from the meetings they are having this spring for gradebook policy changes before we invest more time in this.

thanks!

GabyRangelB:
> @GabyRangelB this is approved, but I'm not sure if you're still working on anything. You should be able to merge this if you're ready.

thanks! merging now

cbini (approved):

PR #2613, opened 2025-04-10
fix: row number calc in google form responses

cristinabaldor:
# Pull Request

## Summary & Motivation

## Self-review

**If this is a same-day request, please flag that in our Slack channel!**

- [ ] Update **due date**, **assignee**, and **priority** on our
- [ ] <kbd>Format</kbd> has been run on all modified files

### SQL

- [ ] Ensure you are using the `union_dataset_join_clause()` macro for queries
      that employ any models using these datasets: `deanslist` `edplan` `iready`
      `overgrad` `pearson` `powerschool` `renlearn` `titan`
      necessity
- [ ] If you are adding a new external source, run:

      dbt run-operation stage_external_sources --vars
      "ext_full_refresh: true"

## Troubleshooting

---
  - https://app.asana.com/0/0/1209952262918429

### DIFF
--- a/src/dbt/kipptaf/dbt_project.yml
+++ b/src/dbt/kipptaf/dbt_project.yml
@@ -237,6 +237,8 @@ models:
     surveys:
       +schema: surveys
       +materialized: table
+      disabled:
+        +enabled: false
     tableau:
       +schema: tableau
       staging:
--- a/src/dbt/kipptaf/models/surveys/disabled/int_surveys__support_survey_details.sql
+++ b/src/dbt/kipptaf/models/surveys/disabled/int_surveys__support_survey_details.sql
@@ -1,0 +1,127 @@
+with
+    response_identifiers as (
+        select
+            fr.form_id as survey_id,
+            fr.info_title as survey_title,
+            fr.response_id as survey_response_id,
+            fr.respondent_email,
+            fr.last_submitted_timestamp as date_submitted,
+            fr.create_timestamp as date_started,
+
+            up.employee_number as respondent_df_employee_number,
+
+            rt.academic_year as campaign_academic_year,
+            rt.name as campaign_name,
+            rt.code as campaign_reporting_term,
+        from {{ ref("base_google_forms__form_responses") }} as fr
+        inner join
+            {{ ref("stg_ldap__user_person") }} as up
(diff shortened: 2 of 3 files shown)


GabyRangelB (dismissed):

kevinverhoff (dismissed):
Approving, but also wondering if we should switch it from BASE to INT?

cbini:
> Approving, but also wondering if we should switch it from BASE to INT?

that would be nice :)

src/dbt/kipptaf/models/google/forms/base/base_google_forms__form_responses.sql
```diff
@@ -37,8 +37,7 @@ with
             }},
 
             row_number() over (
-                partition by
-                    f.form_id, f.item_id, r.respondent_email, rata.is_null_value
+                partition by f.form_id, r.respondent_email, rata.is_null_value
```
cbini, line 40: @cristinabaldor is the point of this to get the most recent overall form response record? that already exists here: https://github.com/TEAMSchools/teamster/blob/4c1ceb1f6506709ecacec3370c622d855d2c3d37/src/dbt/kipptaf/models/google/forms/staging/stg_google_forms__responses.sql#L10-L13

cbini (changes requested):

src/dbt/kipptaf/models/google/forms/base/base_google_forms__form_responses.sql
cristinabaldor: yes it is! let me see if it makes a difference, i feel like something wasn't working which is why we added a separate rn calc

cbini: that would make more sense, yeah. potential problem is that it's also long by response item, so I'm not entirely sure if that will get you the desired outcome

cristinabaldor: gonna resubmit now using dense_rank, which looks like it's working as we want it to -- i didn't understand Kevin's comment, you want me to rename base_forms to Int_forms?

cbini (approved):

PR #2686, opened 2025-05-21
New fafsa fields

GabyRangelB:
# Pull Request

## Summary & Motivation

## Self-review

**If this is a same-day request, please flag that in our Slack channel!**

- [ ] Update **due date**, **assignee**, and **priority** on our
- [x] <kbd>Format</kbd> has been run on all modified files

### SQL

- [ ] Ensure you are using the `union_dataset_join_clause()` macro for queries
      that employ any models using these datasets: `deanslist` `edplan` `iready`
      `overgrad` `pearson` `powerschool` `renlearn` `titan`
      necessity
- [ ] If you are adding a new external source, run:

      dbt run-operation stage_external_sources --vars
      "ext_full_refresh: true"

## Troubleshooting

---
  - https://app.asana.com/0/0/1210340474825757


src/dbt/kipptaf/models/extracts/powerschool/rpt_powerschool__autocomm_students.sql
```diff
@@ -17,6 +17,11 @@ with
                     'ELA' as `s_nj_stu_x__graduation_pathway_ela`
                 )
             )
+    ),
+
+    ps_fafsa_status as (
+        select _dbt_source_relation, studentsdcid, fafsa as ps_fafsa_status,
+        from {{ ref("stg_powerschool__s_stu_x") }}
```
cbini, line 24: no need for a CTE here, you can join this table right to the select statement

src/dbt/kipptaf/models/extracts/powerschool/rpt_powerschool__autocomm_students.sql
cbini, line 1: have you discussed this with Jabari? we'll need to update the autocomm config when this goes live

cbini (changes requested):

cbini:
@GabyRangelB we need to add the new field to the regional variants of `rpt_powerschool__autocomm_students`. Should it get applied to every region?

GabyRangelB:
> @GabyRangelB we need to add the new field to the regional variants of `rpt_powerschool__autocomm_students`. Should it get applied to every region? This can be a follow up PR too if you want to this closed out.

darn. forgot those. nj only for now. do you prefer to do that since i cant build those or want me to take a stab at them?

cbini:
you can go ahead, it's only a one line change on each of them so I don't think we need to test it that strictly

GabyRangelB:
> you can go ahead, it's only a one line change on each of them so I don't think we need to test it that strictly

done!

PR #3227, opened 2026-01-21
Marts: Teammate Metrics

cbini:
# Pull Request

## Summary & Motivation

Things included/what I'm trying to do here:

1. New dedupe method @cbini and I discussed to reduce number of irrelevant rows. In dim_teammates I'm doing a select distinct of all the fields we need for the model and deduping based on field changes that we actually care about.
2. New methodology for attrition. I'm getting every year someone was active and then comparing that roster to the roster of active people the following year. If you don't exist in the roster for the following year, is_attrition is true. Also unioning the different definitions of attrition into one view so when we need different ones we can just filter on attrition_type.
3. New methodology for terminations. Previously this was a subset on the attrition view just to show us termination details on that group, but I had to (I think) create a new fact table to accommodate more attrition types. Same sort of deal here, I'm getting a roster of people for the year, if you have a termination record that year is_termination is true. For the attrition breakdowns we would join on employee_number and academic_year (see below). 

```
'semantic_layer' as metric_type,
avg(if(a.is_attrition and (t.termination_reason like '%Non-Renewal%' or t.termination_reason like '%NonRenewal%' or t.termination_reason like '%Termination%' or t.termination_reason like '%Temp%' or t.termination_reason in ('Administrative','Dismissal','Elimination of Position','Internship Ended','Performance') or (t.is_termination and t.termination_reason is null)),1,0)) as metric_value 
from `teamster-332318`.`zz_cristinabaldor_marts`.`fct_teammate_attrition` as a 
inner join `teamster-332318`.`zz_cristinabaldor_marts`.`fct_teammate_terminations` as t
on a.employee_number = t.employee_number
and a.academic_year = t.academic_year
where a.attrition_type = 'foundation'
```

### General

- [ ] If this is a same-day request, please flag that in the #data-team Slack
- [ ] Run <kbd>Format</kbd> on all modified files

### dbt

- [ ] Include a corresponding `[model name].yml` properties file for all models.
      These can be generated by running
      `dbt run-operation generate_model_yaml --args '{"model_names": ["MODEL NAME"]}'`
      and saving the console output to a YAML file.

      models:
        - name: [model name]
          config:
            contract:
              enforced: false  # required while creating a contracted model, delete before merging
          columns:  # required for contracted models
            - name: ...
              data_type: ...
              data_tests:  # column tests, optional
                - ...
          data_tests:  # model tests, optional
            - ...
            - dbt_utils.unique_combination_of_columns:  # recommended for intermediate models
                arguments:
                  combination_of_columns:
                    - ...
                config:
                  store_failures: true

- [ ] Include (or update) an
      [exposure](https://docs.getdbt.com/reference/exposure-properties) for all
      models that will be consumed by a dashboard, analysis, or application:

      exposures:
        - name: [exposure name, snake_case]
          label: [exposure name, Title Case]
          type: dashboard | notebook | analysis | ml | application
          owner:
            name: Data Team
          depends_on:
            - ref("[model name]")
            - ...
          url: ...  # optional
          config:
            meta:
              dagster:
                kinds:
                  - tableau | googlesheets | ...
                  - ...
                asset:
                  metadata:
                    id: [lsid]  # required for Tableau Server workbooks
                    cron_schedule:  # required for Dagster automation
                      - * * * * *
                      - ...

[Dagster "kinds" Reference](https://docs.dagster.io/guides/build/assets/metadata-and-tags/kind-tags#supported-icons)

### SQL

- [ ] Use the `union_dataset_join_clause()` macro for queries that employ models
      that use regional datasets
- [ ] Do not use `group by` without any aggregations when you mean to use
      `distinct`
      necessity
- [ ] Do not use `order by` for `select` statements. That should be done in the
      reporting layer.
- [ ] If you are adding a new external source, before building, run:

      dbt run-operation stage_external_sources --vars "{'ext_full_refresh': 'true'}" --args select: [model name(s)]

## Troubleshooting

### DIFF
--- a/src/dbt/kipptaf/models/marts/properties/dim_staff.yml
+++ b/src/dbt/kipptaf/models/marts/properties/dim_staff.yml
@@ -96,8 +96,7 @@ models:
 
       - name: is_teacher
         data_type: boolean
-        description:
-          "{{ doc('staff_is_teacher') }}"
+        description: "{{ doc('staff_is_teacher') }}"
 
       - name: is_manager
         data_type: boolean


src/dbt/kipptaf/models/marts/dim_teammates.sql
```diff
@@ -11,40 +41,53 @@ with
                 )
             }} as academic_year,
         from {{ ref("int_people__staff_roster_history") }}
+        where primary_indicator and assignment_status != 'Pre-Start'
+    ),
+
+    dedupe as (
```
cristinabaldor: should we do this on the int view eventually?

src/dbt/kipptaf/models/marts/fct_teammate_terminations.sql
```diff
@@ -0,0 +1,72 @@
...
+            assignment_status as termination_status,
+            assignment_status_reason as termination_reason,
+            min(assignment_status_effective_date) as min_effective_date,
+            min(assignment_status_effective_date) as termination_effective_date,
+        from teammate_history
+        where
+            assignment_status = 'Terminated'
+            and assignment_status_reason
+            not in ('Import Created Action', 'Upgrade Created Action')
+        group by all
```
cbini: specify columns in group by

src/dbt/kipptaf/models/marts/dim_teammates.sql
cbini: add yaml with column schema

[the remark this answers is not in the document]
cristinabaldor: So staff would be like personal fields (name, ethnicity, gender, alumni status) and we would just take the most recent value, and then join to work assignment history when we need it?

src/dbt/kipptaf/models/marts/dim_teammates.sql
cristinabaldor: I'll add an asana task to discuss after pilot is done.

cristinabaldor:
hey @cbini thanks for style changes. here's everything with yaml. I had to kill the dbt semantic layer exposure so that I could build the dim_dates one because it's configured differently. It's not in use and won't be unless Cube goes wrong so I'm ok with taking it out completely.

kevinverhoff (approved):

PR #3360, opened 2026-03-04
fix: bound ancestor recursion to view deps to reduce CPU/memory

cbini:
## Summary & Motivation

Fix exponential CPU/memory blowup in `_build_any_ancestor_updated` and enable view/table-specific automation conditions.

### View-bounded ancestor recursion
- `_build_any_ancestor_updated` was exploring **all** upstream deps at each nesting level, causing exponential blowup (d^n paths) in wide DAGs
- Restrict recursion to only follow view edges via `.allow(view_selection)`, stopping at table boundaries (v^n paths where v << d)

### Dagster bug patch
- `DepsAutomationCondition._get_dep_keys` uses in-place set operations (`&=`, `-=`) on the mutable set returned by `parent_entity_keys`, permanently corrupting the asset graph's dependency data
- Monkey-patch copies the set before mutating — affects all automation conditions globally as a pure correctness fix

### Translator-derived automation conditions
- `CustomDagsterDbtTranslator.get_automation_condition()` now branches on materialization type:
  - **view** models → `dbt_view_automation_condition` (only re-materializes on `newly_missing` or `code_version_changed`)
  - **table/incremental/snapshot** models → `dbt_table_automation_condition` (re-materializes on upstream data changes, with view-bounded ancestor lookthrough)
- `get_tags()` now adds `dagster/materialized` tag from dbt config, used by `_VIEW_SELECTION` to identify views

### Materialization updates
- `int_google_forms__form_responses` → table
- `int_overgrad__students` → table
- `int_performance_management__observations` → table
- Reduces max consecutive view chain depth from 5 to 3

### Tests
- 6 unit tests with synthetic assets for core automation condition logic
- 6 integration tests validating kipptaf dbt asset specs (tags, conditions, materialization types)
- 6 integration tests using real kipptaf chain topologies with `evaluate_automation_conditions()`
- All 18 tests passing

## Test plan
- [x] `pytest tests/test_automation_conditions.py` — 18 tests pass
- [x] Verify automation conditions evaluate correctly in Dagster UI

### DIFF
--- a/src/teamster/core/automation_conditions.py
+++ b/src/teamster/core/automation_conditions.py
@@ -40,10 +40,9 @@ def _patched_get_dep_keys(
 DepsAutomationCondition._get_dep_keys = _patched_get_dep_keys
 
 
-def _external_source_selection() -> AssetSelection:
-    return AssetSelection.all(include_sources=True) - AssetSelection.all(
-        include_sources=False
-    )
+_EXTERNAL_SOURCE_SELECTION = AssetSelection.all(
+    include_sources=True
+) - AssetSelection.all(include_sources=False)
 
 
 def _build_any_ancestor_updated(
@@ -91,7 +90,7 @@ def dbt_view_automation_condition(
             AutomationCondition.newly_requested() | AutomationCondition.newly_updated()
         )
         & ~AutomationCondition.any_deps_missing().ignore(
-            ignore_selection | _external_source_selection()
+            ignore_selection | _EXTERNAL_SOURCE_SELECTION
         )
         & ~AutomationCondition.any_deps_in_progress().ignore(ignore_selection)
         & ~AutomationCondition.in_progress()
@@ -120,7 +119,7 @@ def dbt_table_automation_condition(
             AutomationCondition.newly_requested() | AutomationCondition.newly_updated()
         )
         & ~AutomationCondition.any_deps_missing().ignore(
-            ignore_selection | _external_source_selection()
+            ignore_selection | _EXTERNAL_SOURCE_SELECTION
         )
         & ~AutomationCondition.any_deps_in_progress().ignore(ignore_selection)
         & ~AutomationCondition.in_progress()


cbini:
## Dagster Bug Report: `DepsAutomationCondition._get_dep_keys` corrupts asset graph

### What's the issue?

`DepsAutomationCondition._get_dep_keys` applies in-place set operations (`&=`, `-=`) directly on the mutable set returned by `asset_graph.get(key).parent_entity_keys` when `allow_selection` or `ignore_selection` is set:

```python
# dagster/_core/definitions/declarative_automation/operators/dep_operators.py
dep_keys = asset_graph.get(key).parent_entity_keys  # returns the actual mutable set
if self.allow_selection is not None:
    dep_keys &= self.allow_selection.resolve(...)  # mutates the original set in-place
if self.ignore_selection is not None:
    dep_keys -= self.ignore_selection.resolve(...)  # mutates the original set in-place
```

Since `parent_entity_keys` returns the asset graph's own internal set (not a copy), these operations permanently remove dependencies from the graph. Any subsequent evaluation — including by other automation conditions on other assets — sees a corrupted DAG with missing edges.

### What did you expect to happen?

`.allow()` and `.ignore()` should filter the dep keys for the current evaluation only, without modifying the asset graph's internal dependency data. Other conditions evaluating against the same graph should still see the full set of dependencies.

### How to reproduce?

1. Define a DAG with assets A → B → C
2. Apply `.allow(selection_that_excludes_A)` on a `any_deps_match()` condition for C
3. After evaluation, inspect `asset_graph.get(B).parent_entity_keys` — A is permanently removed
4. Any other automation condition that checks B's parents now sees an incomplete graph

### Dagster version

Discovered on `dagster==1.9.12.post6`

### Deployment type

Local

### Deployment details

N/A

### Additional information

**Fix:** Copy the set before mutating:

```python
dep_keys = set(asset_graph.get(key).parent_entity_keys)  # copy first
```

The `&=` and `-=` then operate on the local copy, leaving the asset graph intact. We are currently using this as a monkey-patch in production.

PR #3479, opened 2026-03-18
fix(google-directory): retry only transient HTTP errors, fix callback bug and kwargs double-passing

cbini:
## Summary

- **Selective retry**: Replaces broad `retry_on=(errors.HttpError,)` with `_retryable_execute()` factory + `_TransientHttpError` subclass so `backoff` only retries transient 5xx/429 responses — 4xx client errors now propagate immediately
- **Callback bug fix**: `self._exceptions = []` was reset on every individual callback invocation, silently dropping all but the last exception per batch; now reset once per batch before the loop
- **kwargs double-passing fix**: `list_users`, `list_groups`, `list_roles`, `list_role_assignments` used `kwargs.get()` + `**kwargs`, causing `TypeError: got multiple values for keyword argument` when callers passed those keys explicitly; fixed with `pop()`
- **Sleep fix**: `time.sleep(1)` was called unconditionally including after the last batch; now skipped on the final iteration
- **Docstrings**: Full Google-style docstrings on all public methods, `setup_for_execution`, and module level in `schema.py`; mkdocstrings-compatible (summary-first, `Attributes:`, `Raises:`, API reference links)
- **Test**: `test_list_retries_on_503_mid_pagination` verifies retry behaviour mid-pagination
- **CLAUDE.md**: Python standards, mkdocstrings rules, kwargs/chunk gotchas, updated retry pattern docs

## Test plan

- [ ] `uv run pytest tests/resources/test_resource_google_directory.py` passes
- [ ] `uv run pytest tests/test_dagster_definitions.py` passes
- [ ] Confirm `_list` retries on 503 but raises immediately on 403

### DIFF
--- a/src/teamster/libraries/google/CLAUDE.md
+++ b/src/teamster/libraries/google/CLAUDE.md
@@ -30,7 +30,17 @@ as `None`. Type both as `X | None`.
 **Unit testing**: To mock the API without live credentials, set
 `resource._resource = MagicMock()` and target
 `mock_resource.<api>.return_value.list.return_value.execute`. Patch
-`dagster._utils.backoff.time.sleep` to suppress retry delays.
+`dagster._utils.backoff.time.sleep` to suppress retry delays in `_list`. To test
+batch methods, set `mock_api.new_batch_http_request.side_effect` to a factory
+that captures the `callback` kwarg and calls it inside `execute()` — see
+`_make_batch_side_effect()` in
+`tests/resources/test_resource_google_directory.py`. Patch
+`teamster.libraries.google.directory.resources.time.sleep` (not the backoff
+path) to suppress inter-batch delays.
+
+**Empty-page responses**: Some API endpoints return `{}` instead of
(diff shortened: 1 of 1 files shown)


cbini:
Fixed the three actionable items:

**`response.get(response_data_key, [])` in `_list`** — done. `resources.py:183`.

**`list_roles` docstring** — softened to \"defaults to 100\" rather than \"capped at 100\". Didn't add a clamp since no existing caller passes `max_results` to `list_roles`, and the API's own 400 is a clear enough signal if one ever does.

**`PrivateAttr(default_factory=list)`** — done.

**Callback regression test** — `test_callback_preserves_earlier_exceptions_within_same_batch` already covers the invariant (pre-populate, call again, verify both survive). Added `test_batch_insert_users_collects_all_exceptions_from_multi_failure_batch` as an end-to-end path through the batch method with two failures in the same batch, which is the scenario the original bug actually silently dropped.

**Single-request bare `.execute()` methods** — deferred to a follow-up.

PR #3603, opened 2026-04-09
fix(kipptaf): retry Tableau version discovery, gate deploys on dep code changes

cbini:
# Pull Request

## Summary & Motivation

1. **Retry Tableau API version discovery** — `tableauserverclient` silently falls back to API version 2.4 when the server info endpoint is unreachable, causing `EndpointUnavailableError: Workbooks is not available in API version 2.4` on the next workbook operation. This adds a tenacity retry (3 attempts, exponential jitter) around `use_server_version()` with an explicit guard against the default version, so transient network failures resolve within `setup_for_execution` instead of failing the run.

2. **Gate dbt materialization on upstream code changes** — adds `_dep_code_version_pending` to `_build_dbt_condition()` which blocks materialization when a direct dependency has a `code_version_changed()` that hasn't been applied yet. Prevents schema errors when a deploy adds columns through a TABLE → VIEW chain and both fire `code_version_changed()` simultaneously — the VIEW now waits for the TABLE to re-materialize first.

## AI Assistance

Both fixes were identified and implemented by Claude Code during a day-2 operations review. The Tableau root cause was traced through TSC source code (`_determine_highest_version` swallows all exceptions). The automation condition gap was identified from the dbt run failure timeline — the VIEW materialized before its upstream TABLE after a deploy.

### General

      code location

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.


cbini:
Addressed the important finding — replaced `default_server_version` private import with initial-version capture in `c30ecd145`. `Server.version` is captured before `use_server_version()` and compared after; no private symbols involved.

Skipping the `before_sleep_log` minor — consistent with existing project patterns and the retry window is short (3 attempts, ~1-10s jitter).

PR #3715, opened 2026-04-22
docs(cube): Cube infrastructure implementation

### LINKED ISSUE #3591: feat: implement Cube Cloud integration
Configure `src/cube/` as a fully version-controlled Cube project with a security model, `cube.js` config, Cube Cloud deployment, and local dev setup — the infrastructure layer that cube model YAML files will sit on top of.

## Scope

- `src/cube/` repository structure (`cube.js`, `package.json`, `.env.example`, `SETUP.md`)
- Security model: Google group naming convention, three-layer access enforcement
- `cube.js` config: dbt Cloud metadata, `contextToGroups`, `queryRewrite`, access policies
- New dbt model: `dim_staff_hierarchy` in `marts/dimensions/`
- Cube Cloud setup: Git sync, prod deployment, per-engineer dev deployments
- Local dev: VS Code task, ADC auth, local manifest workflow, `CUBE_GROUP_MAP`

## Out of scope

- Cube model YAML files (67 models across 12 domains) — follow-up spec
- Pre-aggregations — follow-up spec
- Downstream integrations (Tableau, Superset, Streamlit) — follow-up spec
- Google Group creation and member assignment in Workspace Admin — separate issue for IT admin team
- Vercel and Supabase integrations — future work

## Design spec

[docs/superpowers/specs/2026-04-15-cube-infrastructure-design.md](https://github.com/TEAMSchools/teamster/blob/cristinabaldor/feat/cube-implementation/docs/superpowers/specs/2026-04-15-cube-infrastructure-design.md)

## Implementation plan

TBD — pending spec review

## Prerequisites / follow-up flags

- [ ] **Spot-check #3678 (YAML contract drift)** for `fct_student_attendance_daily` and `dim_students` before writing Cube YAML — enforced contracts will fail at Cube schema load if `data_type` in YAML doesn't match BigQuery's actual type. Low risk for attendance models (no timestamp columns), but confirm before building.
- [ ] **`dim_staff_hierarchy` must be materialized before staff cubes go live** — it is the lookup table for the `queryRewrite` org-hierarchy filter. Not needed for initial student/attendance BI connection testing, but is a hard prerequisite for any staff cube query reaching BigQuery.

---

## Reviewer Context

Background on the key design decisions. Not needed for implementation — included here for discussion.

### Glossary

| Term | Plain English |
| --- | --- |
| **`contextToGroups`** | Cube hook that runs at request time to resolve which `cube-*` groups the authenticated user belongs to. Calls the Google Admin Directory API using the user's email from `securityContext`. Drives all three security layers. |
| **`queryRewrite`** | Cube hook that modifies every query before it hits BigQuery — injects location and org-hierarchy WHERE clauses based on the user's resolved groups. Uses `securityContext.email` directly in the org-hierarchy subquery via a JOIN through `dim_staff`. |
| **`accessPolicy`** | YAML annotation on a cube dimension that hides it from users who don't have the required group. Applied to PII fields and salary data. |
| **Transitive closure** | A table that pre-computes every (ancestor, descendant) pair in a hierarchy at any depth. `dim_staff_hierarchy` is a transitive closure of the org chart — one row for every manager-report relationship, direct or indirect. |
| **ADC** | Application Default Credentials — GCP's mechanism for local auth. Already configured in the devcontainer; no service account key needed locally. |
| **Domain-wide delegation** | A Google Workspace Admin setting that allows a service account to call the Admin Directory API on behalf of any user. Required for `contextToGroups` to look up group memberships in Cloud deployments. |

#### Security enforced once in Cube, nowhere else

All row-level, column-level, and cube-visibility rules live in `cube.js` and cube YAML. No downstream tool (Tableau, Superset, Streamlit) has its own access config. This means a permission change takes effect everywhere the moment it is deployed — no per-tool maintenance.

#### Google Groups as the identity carrier

Rather than maintaining a separate permissions table, we use Google Workspace groups. The `cube-` prefix keeps Cube-relevant groups filterable without a mapping table. Groups are additive — a user's effective permissions are the union of all their `cube-*` groups, with no role hierarchy to maintain.

#### Cube never handles Google OAuth — downstream tools mint the JWT

Users log in to Superset or Streamlit via Google OAuth, not to Cube directly. User identity enters Cube's `securityContext` as a signed JWT that the downstream tool is responsible for minting:

- **Streamlit**: app backend signs a Cube JWT using `CUBEJS_API_SECRET` after Google OAuth, with `{ "email": "user@apps.teamschools.org" }` in the payload.
- **Superset**: connects to Cube's SQL API as `cube-superset-service`, then issues `SET ROLE` per query. `canSwitchSqlUser` gates the switch to the end-user's identity.

#### Three-layer security model

| Layer | Mechanism | Controls |
| --- | --- | --- |
| 1. Identity | `contextToGroups` | Resolves `cube-*` groups from Google Admin Directory API using `securityContext.email` |
| 2. Row-level | `queryRewrite` | Which school/region rows they can see; org-hierarchy filter on staff via `dim_staff` JOIN |
| 3. Column-level | `accessPolicy` in YAML | Which sensitive fields are visible |

Each layer is independent. A user can have network-wide row access but no salary fields, or school-level row access with PII fields — any combination.

### Architectural Decision Rationale

| Decision | Choice | Rationale |
| --- | --- | --- |
| Group naming: scope | `cube-{network\|region-X\|school-X}-{detail\|summary}` | Encodes geographic scope and access level in one group name; filterable by prefix |
| Group naming: detail vs summary | `-detail` = row-level, `-summary` = aggregates only | Plain-language pair; avoids "trends" which sounds like a data concept rather than an access level |
| Group naming: access | `cube-access-{data-area}-{data-type}` | Names the data being gated, not the people accessing it — self-documenting |
| `cube-access-staff-all` | Bypass org-hierarchy filter only | Does not override location scope — a Newark Head of Schools with this group sees all Newark staff, not all network staff |
| `cube-access-student-data` | Gates all student-domain cubes | Users without it (e.g. finance) see no student cubes in schema introspection — not just empty results |
| `dim_staff_hierarchy` | Transitive closure of org chart | Enables a single subquery to resolve any manager's full reporting chain at any depth, without recursive SQL at request time. Uses `securityContext.email` via JOIN through `dim_staff` — no separate employee number lookup needed. |
| `canSwitchSqlUser` | Scoped to `cube-superset-service` only | Required for Superset per-user SQL API impersonation. Gated on `current_user === 'cube-superset-service'` and `new_user.endsWith('@apps.teamschools.org')` — a blanket allow would let any SQL API user impersonate anyone. |
| dbt manifest: local vs Cloud | Local `dbt compile` for dev; dbt Cloud API for Cube Cloud | Cube Cloud fetches the manifest automatically; local dev uses a one-time compile step that only needs re-running when model definitions change |
| `CUBE_GROUP_MAP` | Local dev fallback | Bypasses the Directory API call — lets engineers simulate any permission combination without Workspace Admin access |

cristinabaldor:
## Summary & Motivation

When merged, this pull request will add the `src/cube/` project scaffold needed to run Cube Core locally and deploy to Cube Cloud, closing the infrastructure milestone of [#3591](https://github.com/TEAMSchools/teamster/issues/3591).

**What's included:**

- `src/cube/cube.js` — four hooks: `driverFactory` (restricts scanning to `kipptaf_marts`), `contextToGroups` (Google Directory API + `CUBE_GROUP_MAP` local dev bypass), `queryRewrite` (location scope filter + org-hierarchy segment injection), `canSwitchSqlUser`
- `src/cube/package.json` + `package-lock.json` — `@cubejs-backend/bigquery-driver` + `@cubejs-backend/server`
- `src/cube/.env.example` — documented vars for local dev
- `src/cube/.gitignore` — excludes `.env`, `node_modules/`, `.cubestore/`
- `src/cube/model/cubes/` + `src/cube/model/views/` — empty stubs for YAML models (follow-up spec)
- `src/cube/SETUP.md` — Cube Cloud one-time setup + local dev guide
- `.vscode/tasks.json` — **Cube: Dev Server** task
- Updated spec and new implementation plan

**Deferred to follow-on PRs:**

- `bridge_staff_hierarchy` dbt model — blocked on [#3729](https://github.com/TEAMSchools/teamster/issues/3729) (`dim_staff_work_assignments.staff_key` OID vs. worker ID join mismatch)
- Cube YAML model files (67 models across 12 domains) — follow-up spec
- Cube Cloud one-time setup — manual, performed after merge

**AI Assistance:** Claude Code authored the `src/cube/` scaffold, `cube.js` hooks, VS Code task, and spec/plan updates under human direction. All logic reviewed and approved before commit.

### General

_No dbt model changes in this PR._

_No schedule/sensor/integration changes._

## CI checks

- [x] **Trunk** — passes.


docs/superpowers/specs/2026-04-15-cube-infrastructure-design.md
```diff
@@ -0,0 +1,334 @@
...
+cube-region-{region}-summary # all schools in region, aggregates only
+
+cube-school-{slug}-detail    # one school, row-level
+cube-school-{slug}-summary   # one school, aggregates only
+```
+
+`{region}` matches `location_region` values from `dim_locations` (lowercased,
+hyphenated). `{slug}` matches `location_abbreviation` from `dim_locations`.
+
+A user with `-detail` access can see everything a `-summary` user can see, plus
```
kevinverhoff, line 74: I like this approach

docs/superpowers/specs/2026-04-15-cube-infrastructure-design.md
```diff
@@ -0,0 +1,334 @@
...
+  WHERE s.google_email = '{securityContext.email}'
+)
+```
+
+### Layer 3 — Column-level access policies (YAML)
+
+Sensitive dimensions are marked with `accessPolicy` in their cube YAML:
+
+- Salary, pay rate fields on staff cubes → requires
+  `cube-access-staff-compensation`
```
kevinverhoff, line 169: 🔥

kevinverhoff (dismissed):
this is dope. no notes.

docs/superpowers/specs/2026-04-15-cube-infrastructure-design.md
```diff
@@ -0,0 +1,334 @@
...
+depth                       INT64    NOT NULL
+```
+
+One row per (ancestor, descendant) pair at any depth, including self-referential
+rows (`depth = 0`). Built recursively from `dim_staff.manager_employee_number`.
+Refreshes on the standard marts cadence so org chart changes propagate
+automatically.
+
+`dim_staff_hierarchy` is not a Cube model — it is a small lookup table used only
+as a filter subquery in `queryRewrite`.
```
cbini: can you confirm this has everything that queryRewrite needs?

cbini: this might also might be a "bridge" model

cristinabaldor: Yes — `queryRewrite` only needs `ancestor_employee_number` and `descendant_employee_number`. The email-to-employee-number resolution happens via a JOIN to the existing `dim_staff` table (which has `google_email`), so the hierarchy table itself doesn't need to carry the email. The schema covers everything.

cristinabaldor: Good call — a transitive org chart closure is a many-to-many relationship between staff members (one ancestor → many descendants, one descendant → many ancestors at different depths), which is exactly the bridge pattern. Renaming to `bridge_staff_hierarchy` and moving to `marts/bridges/` in the spec now.

cristinabaldor:
@claude please review the changes to the spec, the implementation plan and new files

cristinabaldor:
@claude please review changes

cristinabaldor:
@claude the replica cache issue probably won't affect us at our scale, we can ignore

cristinabaldor:
@claude check changes

cristinabaldor:
@claude check that plan steps 5b and 5c now match the shipped cube.js exactly.

src/cube/SETUP.md
```diff
@@ -0,0 +1,45 @@
...
+3. Set the Cube project path to `src/cube/`
+4. Set the production branch to `main` — merges trigger automatic redeploy
+5. Set the following environment variables in Cube Cloud:
+   - `CUBEJS_DB_TYPE=bigquery`
+   - `CUBEJS_DB_BQ_PROJECT_ID=teamster-332318`
+   - `CUBEJS_DB_BQ_CREDENTIALS` — service account JSON (base64-encoded)
+   - `CUBEJS_API_SECRET` — generate a strong random secret
+   - `GOOGLE_DIRECTORY_SA_KEY` — Admin Directory API service account
+     (base64-encoded)
+   - `GOOGLE_DIRECTORY_SA_SUBJECT` — email of the Workspace super-admin that
```
cristinabaldor: @cbini is this the right address for the super-admin?

cbini:
## Code Review

### Strengths
- Clean separation: hooks isolated in `cube.js`, model dirs stubbed empty, deps minimal.
- `cube.js:18-21` — `driverFactory` hard-codes `database: "kipptaf_marts"`, correctly limiting BigQuery dataset exposure.
- `cube.js:26-85` — `contextToGroups` defaults to deny-on-error, uses `member.readonly` minimum scope, caches with 5-min TTL.
- `cube.js:135-147` — `queryRewrite` defaults to `values: []` when no scope group present (fail-closed).
- `cube-` group prefix filter applied consistently in both bypass and Directory-fetch paths.

### Critical
None.

### Important
1. **`CUBE_GROUP_MAP` bypass not gated to dev** (`cube.js:31`) — only checks "is var set?". If leaked into Cube Cloud prod, it short-circuits the Directory API entirely. Gate on `NODE_ENV !== "production"` and document in `SETUP.md` that it must never be set in Cube Cloud.
2. **`schemaVersion` reads attacker-controllable input** (`cube.js:23-24`) — if `securityContext.schemaVersion` comes from a user-supplied JWT, it's a cache-poisoning vector across users. Derive server-side or remove until needed.
3. **Region/school slug regex permits dot/path chars** (`cube.js:106-111`) — `.+?` matches `..`, `/`, quotes. Use `[a-z0-9_]+` and explicit allow-list validation.
4. **Group `email` may be undefined** (`cube.js:70`) — `g.email.split("@")[0]` throws on missing field, dropping into the catch and denying all for that user. Use `(g.email ?? "").split("@")[0]`.
5. **`canSwitchSqlUser` hardcodes `"cube-superset-service"`** (`cube.js:169-171`) but `CUBEJS_SQL_SUPER_USER` is configurable in `SETUP.md`. Read from env (`process.env.CUBEJS_SQL_SUPER_USER`) to prevent drift.
6. **`STUDENT_CUBES`/`STAFF_CUBES` placeholders silent** — if a real student cube YAML lands later without updating these arrays, students would be exposed. Add a TODO with link to the follow-up issue.

### Minor
- `.vscode/tasks.json` Cube task: `[ -d node_modules ] || npm install && npm run dev` — `&&` binds tighter than `||`, so when `node_modules` exists the dev server never starts. Use `([ -d node_modules ] || npm install) && npm run dev`.
- `package-lock.json` is 10,191 lines — verify only `registry.npmjs.org` URLs (no internal/private registries).
- Confirm `.gitignore` lists `.env`.
- `queryRewrite` mutates via reassignment then spreads at return — works but fragile; consider building one final object.

### Assessment
**Needs minor changes before merge.** Important items 1, 2, and 5 are real security/operational hazards (bypass leakage, cache poisoning, config drift) — fix before any production Cube Cloud deployment. Item 1 is the most pressing. The `tasks.json` shell precedence bug will block local dev on second run. Otherwise the scaffold is acceptable given YAML models and staff hierarchy are explicitly deferred.

cristinabaldor:
**Item 1 — `CUBE_GROUP_MAP` not gated to `NODE_ENV`**

Fixed. Added `process.env.NODE_ENV !== "production"` guard so the bypass is a no-op in Cube Cloud even if the variable were accidentally set. Added a warning to `SETUP.md` that `CUBE_GROUP_MAP` must never be configured in Cube Cloud.

**Item 2 — `schemaVersion` reads JWT**

Removed `schemaVersion` entirely rather than hardcoding `"v1"`. The hook is only needed when the schema itself varies per user (e.g., different tenants see different cube definitions). KIPP's security model is entirely in `queryRewrite` — the YAML schema is identical for every user, only query results differ. Every user gets the same compiled schema regardless of their groups. Omitting the hook lets Cube handle cache invalidation via file hashes, which is the correct default for a single-schema deployment and removes the JWT read entirely.

**Item 3 — Regex permits `.`, `/` in slug**

Tightened to `[a-z0-9][a-z0-9-]*` rather than the suggested `[a-z0-9_]+`. The spec defines region/school keys as lowercased and hyphenated (e.g., `north-jersey`), so hyphens must be allowed in the middle portion — `[a-z0-9_]+` would silently reject valid multi-word region groups.

**Item 4 — `g.email` may be undefined**

Added `(g.email ?? "").split("@")[0]`. The outer try/catch already handled this as a default deny, but the guard keeps the error from swallowing valid groups when only one entry in the response has a missing field.

**Item 5 — `canSwitchSqlUser` hardcodes username**

Fixed. Now reads `process.env.CUBEJS_SQL_SUPER_USER` to stay in sync with the documented env var.

**Item 6 — `STUDENT_CUBES` placeholder has no issue reference**

Added a TODO referencing the YAML implementation as the follow-up to PR #3715. `STAFF_CUBES` already had `#3729`; `STUDENT_CUBES` now has equivalent treatment so a future YAML PR can't land without updating both arrays.

**`tasks.json` shell precedence**

No change. Bash gives `&&` and `||` equal precedence, evaluated left-to-right — unlike C or Python. `A || B && C` parses as `(A || B) && C`. When `node_modules` exists: `[ -d node_modules ]` is true, `||` short-circuits, `&& npm run dev` executes. When absent: `[ -d node_modules ]` is false, `npm install` runs, `&& npm run dev` follows. Both paths work correctly with the current command.

**`package-lock.json` registry URLs**

Verified — all `resolved` URLs point to `registry.npmjs.org`. No private or internal registries present.

**`.gitignore`**

`.env` is line 1.

**`queryRewrite` mutation pattern**

No change. The reassignment-via-spread pattern is intentional — `query` is conditionally rebuilt at two independent branch points (student cube filtering and staff segment injection). A single-pass construction would require nested ternaries and be harder to read. The final `return { ...query, filters }` correctly merges everything; there's no in-place mutation at any point.

cristinabaldor:
@claude confirm these changes and whether we are safe to merge

cristinabaldor:
@claude please review

PR #3730, opened 2026-04-23
fix(tableau): cast max_score to int64 in college assessment benchmark calcs

GabyRangelB:
## Summary

- Casts `max_score` from `NUMERIC` to `INT64` in `rpt_tableau__college_assessment_dashboard_benchmark_calcs`
- Updates the dbt contract YAML to reflect the new `int64` type
- Fixes Tableau error 9773A965 ("Cannot expand domain of data type string with a value of data type real") caused by a type mismatch between the `NUMERIC` `max_score` field (which Tableau maps to Real) and the Integer benchmark goal parameters

## Test plan

- [ ] Confirm dbt model compiles and runs cleanly against staging
- [ ] Verify Tableau workbook no longer throws error 9773A965 on the `Met Benchmark Goal Status` field after refreshing the data source

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

---
  - https://app.asana.com/0/0/1214244015850864

### DIFF
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__college_assessment_dashboard_benchmark_calcs.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__college_assessment_dashboard_benchmark_calcs.yml
@@ -78,7 +78,7 @@ models:
           area across all assessment administrations. Null for students with no
           qualifying score. Used as the score value for benchmark percent-met
           calculations in Tableau.
-        data_type: numeric
+        data_type: int64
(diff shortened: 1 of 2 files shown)


GabyRangelB:
@claude Skipping the inline comment per our project convention: comments are only added when the WHY is non-obvious. The assumption that SAT Suite scores are always integers is documented in this PR description and in the dbt contract type change itself (`numeric → int64`). A future maintainer who needs context can read the PR history — an inline comment would just add noise that rots as the codebase evolves.

kevinverhoff (approved):
i ❤️ int64

PR #3740, opened 2026-04-24
feat(tableau): add benchmark_goal and met_benchmark_goal to college assessment benchmark calcs

GabyRangelB:
## Summary

- Adds `benchmark_goal` (INT64) to the scaffold CTE via a `CASE` on `benchmark_group`, hardcoding the score threshold for each scope/subject/benchmark combination
- Propagates `benchmark_goal` through the `base` CTE to the final `SELECT`
- Adds `met_benchmark_goal` (STRING) to the final `SELECT` — `'Met'` / `'Not Met'` / `'No Data'` — computed as a simple comparison of `max_score >= benchmark_goal`
- Moves all benchmark threshold logic out of Tableau calculated fields and into the dbt model, eliminating the Tableau type-mismatch error (Error 9773A965) that occurred when filtering on string fields derived from numeric comparisons

## Test plan

- [ ] Confirm dbt model compiles and runs cleanly against staging
- [ ] Verify `benchmark_goal` values are correct for each scope/subject/benchmark combination
- [ ] Verify `met_benchmark_goal` returns `Met`, `Not Met`, and `No Data` as expected
- [ ] Refresh Tableau extract and confirm `met_benchmark_goal` can be added to the filter shelf without error

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__college_assessment_dashboard_benchmark_calcs.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__college_assessment_dashboard_benchmark_calcs.yml
@@ -82,7 +82,17 @@ models:
       - name: benchmark_name
         description: >
           Readiness benchmark label. One of College-Ready, HS-Ready, EA/ED-Ready
-          (for Combined rows) or EBRW, Math (for subject-specific rows). Tableau
-          consumers filter to a single benchmark_name and compare max_score
-          against a dynamic threshold parameter to compute percent met.
+          (for Combined rows) or EBRW, Math (for subject-specific rows).
+        data_type: string
+      - name: benchmark_goal
+        description: >
+          Hardcoded score threshold for the benchmark. Derived from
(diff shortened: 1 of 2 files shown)


GabyRangelB:
@claude Done — added `not_null` on `benchmark_goal` and `accepted_values` on `met_benchmark_goal`, with both columns moved to the top of the `columns:` list per convention.

Not addressing the remaining items:
- **Sync comment above CASE**: per our project convention, comments are only added when the WHY is non-obvious to a future reader. The `not_null` test on `benchmark_goal` now catches any sync drift at CI time, which is a stronger guard than a code comment.
- **ST06 ordering in final SELECT**: `max_score` is a plain column ref at this SELECT level — the linter doesn't flag it, and the formatter didn't reorder it. Not changing it.

The CI failure is unrelated to this PR — transient Google Sheets service overload on `stg_google_sheets__surveys__scd_answer_crosswalk`.

anthonygwalters (approved):

PR #3760, opened 2026-04-29
feat(kipptaf): add grade level to award ceremony GPA extract

GabyRangelB:
## Summary
- Adds `grade_level` column to `rpt_gsheets__award_ceremony_gpa`
- Expands student scope from seniors only (grade 12) to all HS students (grades 9–12)

## Test plan
- [ ] `dbt build --select rpt_gsheets__award_ceremony_gpa` passes locally (view created, uniqueness test passes)
- [ ] Verify Google Sheet consumer handles additional grade levels correctly

---
  - https://app.asana.com/0/0/1214399308568914

### DIFF
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__award_ceremony_gpa.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__award_ceremony_gpa.yml
@@ -1,31 +1,48 @@
 models:
   - name: rpt_gsheets__award_ceremony_gpa
+    description: >
+      Per-student subject GPA extract for award ceremony use, scoped to active
+      HS students (grades 9–12) in the current academic year. Reports both
+      stored (Y1) and live (Q4) GPA points pivoted by credit type.
     data_tests:
       - dbt_utils.unique_combination_of_columns:
           arguments:
             combination_of_columns:
               - school
+              - grade_level
               - student_number
     columns:
       - name: school
         data_type: string
+        description: School abbreviation from PowerSchool.
       - name: grade_level
         data_type: int64
+        description: Student's current grade level (9–12).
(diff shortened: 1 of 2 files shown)


GabyRangelB:
@claude — thanks for the review! A few notes to address your findings:

## What changed

- Added `grade_level` to the output so the Google Sheet consumer can filter/group by grade
- Expanded student scope from grade 12 only to all HS students (grades 9–12)
- Updated the uniqueness test to include `grade_level` (per your critical flag — good catch)
- Added model and column descriptions

## On the `gpa_type` drop and mixed sources

The `gpa_type` column in `grade_source` is intentionally dropped in `calculations` — this is by design, not an oversight.

**Context:** this extract powers an award ceremony that runs *before* the end of the school year. At ceremony time, final Y1 grades don't exist yet for Q4. To produce a meaningful GPA per credit type, the model combines two sources for each course:

1. **Stored Y1** (`storedgrades`, `storecode = 'Y1'`) — PowerSchool's running cumulative grade for the course, reflecting completed quarters (Q1–Q3) weighted together
2. **Live Q4** (`final_grades`, `storecode = 'Q4'`) — the current in-progress Q4 grade

Both records intentionally exist for the same course at the same time. By summing `gpa_points * earnedcrhrs` across both and dividing by the total `potentialcrhrs`, `calculations` produces a blended projected Y1 GPA — incorporating finalized Q1–Q3 performance alongside the running Q4 grade. This is the intended "fake" Y1 for ceremony purposes.

`max()` in the PIVOT is required syntax; since `calculations` already produces one row per student × credit type after the `GROUP BY`, it acts as a pass-through.

The `gpa_type` column was included in `grade_source` for transparency/debugging but is correctly excluded from the aggregation — merging both sources into a single GPA figure is the whole point.

cristinabaldor (approved):

Issue #3766, opened 2026-04-29
fix(dbt): fct_survey_responses FK gaps — survey_question_key (39K), survey_submission_key (415K)


cbini:
## Context

Surfaced by dbt CI run [70403205283381](https://rn998.us1.dbt.com/deploy/116510/projects/211862/runs/70403205283381/) on PR #3765. Two relationships-test warnings on `fct_survey_responses`.

## Failing relationships tests

| Child model | Column | Parent | Orphan rows |
|---|---|---|---:|
| `fct_survey_responses` | `survey_question_key` | `dim_survey_questions.survey_question_key` | 39,426 |
| `fct_survey_responses` | `survey_submission_key` | `fct_survey_submissions.survey_submission_key` | 415,421 |

## Probable root cause

Same audit class as #3723 (`bridge_student_contacts.student_key`) and #3724 (`fct_behavioral_consequences.behavioral_incident_key` fact-to-fact). The fact carries response rows that reference question / submission keys absent from their respective parent models.

For `survey_submission_key` (fact-to-fact, 415K), candidate causes:

- Submission filter on `fct_survey_submissions` excluding rows that still appear in `fct_survey_responses` (e.g. status / completeness filter).
- Upstream Alchemer dedup mismatch — related to #3629 but distinct: dedup fixes row-level duplication, not surrogate-key coverage.

For `survey_question_key` (39K), candidate causes:

- `dim_survey_questions` may be definition-grain after #3646 lands; current population may exclude question variants referenced by responses.
- Question keys generated in the fact from a different composition than the dim.

## Acceptance

- [ ] Identify which submissions / questions are missing from the parent and why
- [ ] Decide whether to expand parent coverage or filter the fact
- [ ] Both relationships tests return 0 rows

## Related

- #3723 — sibling FK gap (bridge_student_contacts.student_key)
- #3724 — sibling fact-to-fact FK gap (fct_behavioral_consequences)
- #3646 — definition-grain catalog intermediates for dim_assessments / dim_surveys (may resolve survey_question_key half)
- #3629 — Alchemer dedup (related but not the FK gap)
- #3713 — parent FK-constraints initiative

Issue #3737, opened 2026-04-24
fix(dbt): dim_student_assessment_expectations Paterson 2026-02-02 term gap (~105 rows)


cbini:
After PR #3734, ~105 Paterson assessment rows dated 2026-02-02 still surface with `term_key = NULL`. The Paterson RT sheet coverage starts 2025-07-01, so the date itself is in range — the miss is likely a `school_id` mismatch between the Paterson scaffold and the sheet's Paterson RT rows.

Scope:

- Compare `int_assessments__scaffold.powerschool_school_id` for Paterson 2026-02-02 rows against the `school_id` column on Paterson RT sheet rows covering 2026-02-02.
- Either fix the scaffold's school_id population or add the missing Paterson RT row(s) in the sheet.

Follow-up to #3717.

cbini:
Spike confirms the diagnosis. On 2026-02-02 the Paterson rows in `int_assessments__scaffold` carry `powerschool_school_id = 1234` (~105 rows, matching the issue's count), while the `Paterson RT3` row in `stg_google_sheets__reporting__terms` covering 2026-02-02 carries `school_id = 2`. The Paterson RT sheet has only two distinct `school_id` values total (`2` and `1234`, 4 rows each), so one of them is a placeholder/typo for what should be a single canonical Paterson school_id.

**Ops ask:** which is canonical for Paterson — `2` or `1234`?

The scaffold's `1234` derives from PowerSchool data (regional source); the sheet's `2` looks like a placeholder. If `1234` is canonical, the sheet's `school_id = 2` rows need to be updated to `1234`. If `2` is canonical, the underlying PS data needs reconciliation (much larger fix).

This is a prerequisite for the batch F PR to merge — `bridge_assessment_expectations_student_scoped`'s `term_key → dim_terms` relationships test will report ~105 Paterson 2026-02-02 orphans until the alignment is done.

Once Ops edits the sheet cell, downstream rebuild via `dbt build --select stg_google_sheets__reporting__terms+1 --exclude resource_type:test` and the test should return 0 orphans.

cc @cbini

Refs PR for batch F (forthcoming): branch `cbini/feat/claude-batch-f-assessment-survey-catalog`.

cbini:
[...]

Issue #3648, opened 2026-04-15
fix(dbt): add assessment_key FK crosswalks for dim_assessment_comparisons and dim_assessment_targets


cbini:
## Context

Two spec-required FKs from assessment-related dimensions to `dim_assessments`
are missing because upstream Google Sheet inputs don't have enough columns to
construct the composite `assessment_key` (`assessment_type, title,
subject_area, scope, module_code, grade_level`). Both need a crosswalk from
the sheet-level keys to assessment key components.

### dim_assessment_comparisons (was #3638) — spec line 388

Comparison data from `stg_google_sheets__state_test_comparison` has
`(test_name, test_code)` but no grade_level — comparisons are at the test
level (e.g., "NJSLA ELA") with no grade_level, so they can't match a specific
assessment row.

### dim_assessment_targets (was #3639) — spec line 389

Target data from `stg_google_sheets__assessments__academic_goals` has
`(state_assessment_code, illuminate_subject_area, grade_level)` but no mapping
from `state_assessment_code` to `(title, scope, module_code)`.

## Fix options (per dim)

1. Add assessment key component columns to the source Google Sheet
2. Create a crosswalk seed mapping source codes to assessment key components
3. Join through an intermediate model that resolves the mapping
4. (Comparisons only) Reconsider the spec — comparisons may be at a coarser
   grain than `dim_assessments` and the FK may not be appropriate

## Related

- Discovered in PR #3636 — PK/FK compliance audit against star schema spec
- Consolidates: #3638, #3639

cbini:
## Scope expansion: promote to Group D umbrella

Surfaced by #3713 / PR #3714 (CI run 70403204236701).

### Additional failing relationships tests (8 total in this group, ~9M rows)

| Child model | Column | Orphan rows |
|---|---|---:|
| `fct_assessment_scores_enrollment_scoped` | `assessment_key` | 4,384,576 |
| `fct_assessment_scores_enrollment_scoped` | `student_key` | 19,927 |
| `dim_student_assessment_expectations` | `assessment_key` | 4,437,125 |
| `dim_student_assessment_expectations` | `student_key` | 1,186 |

Plus the term_key failures on `dim_student_assessment_expectations` tracked under the dim_terms umbrella (#3717).

### Subsumes

- #3640 — add `student_section_enrollment_key` FK to `dim_student_assessment_expectations` (same table, same domain)

### Linked as root-cause candidates

[...]

PR #3770, opened 2026-04-30
feat(dbt): batch F — assessment/survey catalog refactor

### LINKED ISSUE #3628: fix(dbt): deduplicate upstream int_assessments__response_rollup and int_assessments__college_assessment
## Context

Both `int_assessments__response_rollup` and `int_assessments__college_assessment` contain true duplicate rows (identical on all columns). The mart models (`fct_assessment_scores_enrollment_scoped`, `fct_assessment_scores_student_scoped`) handle this with `dbt_utils.deduplicate()`, but the duplicates should be fixed at the source.

## Scope

Investigate and fix the deduplication at the intermediate layer so downstream models don't need workarounds.

## Related

- Parent: #3543 (star schema data mart)

### LINKED ISSUE #3629: fix(dbt): deduplicate upstream int_surveys__survey_responses Alchemer data
## Context

`int_surveys__survey_responses` contains duplicate rows from the Alchemer survey data source. The mart model `fct_survey_responses` handles this with deduplication, but the duplicates should be fixed at the source.

## Scope

Investigate the Alchemer ingestion pipeline and fix deduplication in the intermediate layer.

## Related

- Parent: #3543 (star schema data mart)

### LINKED ISSUE #3640: fix(dbt): add student_section_enrollment_key FK to dim_student_assessment_expectations
## Problem

`dim_student_assessment_expectations` is missing
`student_section_enrollment_key` FK to `dim_student_section_enrollments` per
the star schema spec (line 390).

## Root cause

The upstream `int_assessments__scaffold` generates expectations at the student
x assessment x school grain, not at the section level. The
`student_section_enrollment_key` requires `(cc_dcid, _dbt_source_relation)` —
the specific course section enrollment — but the scaffold doesn't know which
section drives the expectation, only which school.

## Fix options

1. Enhance `int_assessments__scaffold` to join through
   `base_powerschool__course_enrollments` and resolve the section enrollment
   that matches the assessment's subject area
2. Accept the coarser grain and remove this FK from the spec

## Discovered in

PR #3636 — PK/FK compliance audit against star schema spec.

### LINKED ISSUE #3646: fix(dbt): create definition-grain catalog intermediates for dim_assessments and dim_surveys
## Context

Multiple mart dimensions extract definitions from upstream models at response
grain, forcing `SELECT DISTINCT` to collapse to definition grain. Creating
definition-grain catalog intermediates eliminates this workaround.

### dim_assessments (was #3634)

Pulls from 5 per-student-grain upstreams, each needing DISTINCT:

- `int_assessments__response_rollup` (Illuminate internal)
- `int_pearson__all_assessments` (NJ state)
- `int_fldoe__all_assessments` (FL state)
- `int_assessments__college_assessment` (SAT/ACT/PSAT)
- `int_assessments__ap_assessments` (AP)

### dim_surveys + dim_survey_questions + bridge_survey_questions (was #3635)

All extract from response-grain upstreams with DISTINCT.

## Scope

Create assessment-definition-grain and survey/question-definition-grain
intermediate models so the dimensions can select directly without DISTINCT.

## Related

- Parent: #3543 (star schema data mart)
- Consolidates: #3634, #3635

### LINKED ISSUE #3736: fix(dbt): int_assessments__scaffold.region IS NULL for 1.6M rows
`int_assessments__scaffold` emits ~1.6M rows with `region IS NULL`, of which ~273K also have `administered_at IS NULL`. These rows cannot match any term in `stg_google_sheets__reporting__terms` via `(region, date)` and therefore surface as `term_key = NULL` in `dim_student_assessment_expectations` after PR #3734.

Scope:

- Investigate upstream region population in `int_assessments__scaffold` and its inputs.
- Either populate `region` on every row via the school/location lookup, or drop the 273K no-date rows if they represent unusable assessment records.

Not blocking PR #3734 — this is a separate data-quality issue in the assessments domain, not a `term_key` logic bug.

Follow-up to #3717.

cbini:
## Summary

7-issue batch implementing PR batch F of the Cube prerequisite work.

**Closes:**

- #3628 (dedup `int_assessments__response_rollup` + `int_assessments__college_assessment`)
- #3629 (dedup `int_surveys__survey_responses`)
- #3640 (replace `dim_student_assessment_expectations` with `bridge_assessment_expectations_enrollment_scoped` + `bridge_assessment_expectations_student_scoped`)
- #3646 (eliminate response-grain DISTINCT workarounds; introduce `dim_assessment_administrations`)
- #3736 (populate `region` in `int_assessments__scaffold`)

**Partial close:**

- #3648 — comparisons half resolved as spec edit; targets half carved out (Ops decision needed on `Test_Program` column).
- #3766 — Partial close. The 39K `survey_question_key` orphans are resolved by extending `dim_survey_questions` (and `bridge_survey_questions`) to cover Alchemer and Manager survey questions in addition to the prior Google Forms + SCD sources. The 415K `survey_submission_key` orphans persist — root cause is in `fct_survey_submissions` filter logic (excludes rows that still appear in `fct_survey_responses`); that model isn't in this PR's modified surface. Needs separate follow-up.

**Out-of-band prerequisite:**

- #3737 — Paterson 2026-02-02 RT sheet cell. Ops to align `school_id` (currently 1234 in scaffold vs 2 in sheet). Tracked in #3737.

## Architecture

- **Phase 1** — Source dedup at originating int layer (Tasks 1-3) replaces `dbt_utils.deduplicate()` workarounds in `fct_assessment_scores_*` and `fct_survey_responses`. Each int gained a `unique_combination_of_columns` test.
- **Phase 2** — `int_assessments__scaffold` populates `region` via `int_people__location_crosswalk` (Task 4) and threads `cc_dcid` + `_dbt_source_relation` from `base_powerschool__course_enrollments` (Task 5).
- **Phase 3** — Marts cutover (Tasks 6-17): new `dim_assessment_administrations` parallels `dim_survey_administrations`; `dim_assessments` stays definition-grain. The factless many-to-many bridge replacing `dim_student_assessment_expectations` is split by `is_internal_assessment` to remove diamond paths to `dim_terms` and `dim_students`. Assessment fact tables drop dedup workarounds and R9-violating columns; switch FK from `assessment_key` to `assessment_administration_key`.

## Substantive findings during execution

- **Task 1**: 6-key partition (added `powerschool_school_id`) preserves cross-school student fan-out. The fact's previous 5-key dedup workaround was incorrectly collapsing those rows.
- **Task 7**: `dim_assessment_administrations` PK extended to 12 columns (added `season` + `administration_window`) — without these, state NJ/FL had 56 uniqueness violations.
- **Task 11/12**: bridges re-join `int_assessments__assessments` to lift canonical `grade_level` and `academic_year` (scaffold uses `coalesce(illuminate_grade_level_id, agl.grade_level_id)` and renames `academic_year_clean` to `academic_year`, both off-by-one from the dim's source). NULL `cc_dcid` rows filtered out (116K rows; can't link to enrollment). Final row counts: `bridge_assessment_expectations_enrollment_scoped` 2,769,792 rows; `bridge_assessment_expectations_student_scoped` 1,297 rows.
- **Task 12 scope**: bridge covers K-8 replacement only. The "all other assessments" subset (1.6M rows in scaffold) sources from `int_assessments__assessments` Illuminate catalog only — `dim_assessment_administrations` filters those rows out (`where regions_assessed is not null`). The original `dim_student_assessment_expectations` had the same coverage gap; this is structural cleanup not regression.
- **Task 14 — fact-side grade_level fix**: pre-existing inconsistency where the fact's internal-branch hash passed `cast(null as int64)` for grade_level while `dim_assessments` illuminate CTE used the actual `grade_level`. Resolved by re-joining `int_assessments__assessments` in the fact's internal CTE (mirrors the Task 11/12 bridge pattern).
- **Task 14 — data-quality flags, not structural orphans**: 49 WARN-level relationships orphans on `assessment_administration_key`, in two distinct buckets:
  - **43 on `fct_assessment_scores_enrollment_scoped`** — cross-region administration drift. The student's resolved region (e.g., Paterson) is not in the assessment's declared `regions_assessed` (e.g., "Camden, Newark"). The dim is built by unnesting `regions_assessed`, so these out-of-region scoring events have no matching dim row. Two upstream interpretations, both legitimate Ops/data-team escalations: (a) `regions_assessed` is incomplete and should be widened, OR (b) the student took an assessment they shouldn't have (mis-administered).
  - **6 on `bridge_assessment_expectations_student_scoped`** — distinct cause: `int_people__location_crosswalk` returns NULL region for some `ssa.site_id` values on K-8 replacement rows. Upstream crosswalk gap.
  - 73K dev-reported figure was stale-dev artifact; CI on PR branch and a fresh prod-cloned dev build both show the same 43 + 6. Fix is upstream catalog/data, not in this PR. Relationships tests stay WARN-level for ongoing monitoring.
- **Task 15**: 0 FK orphans; clean refactor.
- **Task 16**: #3766 `survey_question_key` half closed in this PR by extending `dim_survey_questions` and `bridge_survey_questions` to source Alchemer + Manager survey questions; verified 0 orphans on `fct_survey_responses.survey_question_key`. The `survey_submission_key` half (415K orphans) persists from a separate root cause in `fct_survey_submissions` filter logic and needs separate follow-up.

## Diamond walk (manual)

| Mart | FKs | Same-ancestor paths? |
|---|---|---|
| `dim_assessment_administrations` | `assessment_key`, `administered_date_key` | none |
| `bridge_assessment_expectations_enrollment_scoped` | `assessment_administration_key`, `student_section_enrollment_key` | none |
| `bridge_assessment_expectations_student_scoped` | `assessment_administration_key`, `student_key`, `term_key` | none |
| `fct_assessment_scores_enrollment_scoped` | `assessment_administration_key`, `student_key`, `test_date_key` | `dim_dates` reached via admin and directly — **role-disambiguated** per CLAUDE.md `created_date_key`/`solved_date_key` convention; not a diamond |
| `fct_assessment_scores_student_scoped` | same shape | same role-disambiguation note |

## Test plan

- [x] All relationships tests on the new bridges at the documented orphan counts (or 0)
- [x] Ops aligns Paterson 2026-02-02 RT sheet cell (#3737) before merge
- [x] dbt Cloud CI build green on this branch
- [x] Trunk check --ci clean


cbini:
## Summary of actions since the [initial Claude review](https://github.com/TEAMSchools/teamster/pull/3770#issuecomment-4349662128)

### Review findings addressed

Commit `c391b7c12` — `fix(dbt): address PR #3770 review findings`:

- **Critical** — Added `dim_assessment_administrations` to `cube.yml` exposure (was missed alongside the two new bridges).
- **Important** — Renamed stale `rn` → `round_rn` in `int_surveys__survey_responses.yml` to match the SQL output.
- **Important** — Removed stale `row_count_by_test_date` entry from `int_assessments__college_assessment.yml` (column not produced).
- **Important** — Documented `discipline` column in `int_assessments__scaffold.yml` with per-branch null-pattern description.
- **Important** — Added `dbt_utils.unique_combination_of_columns` test to `int_assessments__course_enrollments.yml` on the existing dedup partition (closes the intermediate-uniqueness convention gap).
- **Minor** — Added projection comment to the new `school_to_region` CTE in `int_assessments__scaffold.sql`.

### Re-evaluation against fresh prod-cloned dev

After cloning prod into dev, re-ran the audit and found Task 14's "73K orphans" claim was a stale-dev artifact. CI on the PR branch and fresh dev build both show **43 fact + 6 bridge** WARN-level orphans. PR description updated to correct the figure and reframe these as data-quality flags rather than structural orphans:

- 43 fact orphans → cross-region administration drift (student's resolved region not in assessment's `regions_assessed`)
- 6 bridge orphans → distinct upstream gap (`int_people__location_crosswalk` returns NULL region for some K-8 replacement `ssa.site_id` values)

Both surfaces require upstream catalog/Ops review, not code changes here. Relationships tests stay WARN-level for ongoing monitoring.

### Documentation cleanup

Commit `0e2afbe3e` — `chore(dbt): address minor review followups for batch F PR`:

- Documented the `superscore desc` tiebreaker in `int_assessments__college_assessment.sql`.
- Added `config.meta.source_column` pointer to `_dbt_source_relation` in `int_assessments__course_enrollments.yml` (parity with `cc_dcid`).
- Documented the `FallBlock → Fall` season normalization on `dim_assessment_administrations.season`.
- Renamed Alchemer branch's `rn_round` alias → `round_rn` in `int_surveys__survey_responses.sql` for consistency with the YAML and the Google Forms branch.

### Reverts

Commit `b0e890a5f` — `revert(dbt): keep dim_assessment_administrations sourced from regions_assessed unnest`:

`0e2afbe3e` accidentally included an SQL change that sourced the dim's illuminate branch from `int_assessments__scaffold` instead of unnesting `int_assessments__assessments.regions_assessed`. That direction was explicitly rejected during review — the dim grain must reflect declared administrations (`regions_assessed`), not observed student responses. Cross-region scoring events surfacing as 49 WARN-level orphans are intended data-quality flags, not bugs to mask by widening the dim. Other 6 items from `0e2afbe3e` preserved.

Commit `4d34b7267` — `revert(dbt): drop tautological not_null tests on bridge FK keys`:

The `not_null` tests added on bridge FK keys (`student_section_enrollment_key`, `assessment_administration_key`, `student_key`) check the output of `dbt_utils.generate_surrogate_key`, which never returns NULL even with all-NULL inputs. The tests passed by construction and caught nothing the relationships tests wouldn't catch first. Real upstream invariants (`cc_dcid IS NOT NULL` etc.) are enforced in the bridges' `WHERE` clauses, not by mart-level tests.

### PR description refinements

- Disambiguated the 43 fact orphans vs 6 bridge orphans into separate root causes.
- Called out Task 14's fact-side `grade_level` hash fix (re-join `int_assessments__assessments` to lift canonical `grade_level`, mirrors the Task 11/12 bridge pattern; resolves a pre-existing inconsistency where the fact passed `cast(null as int64)`).
- Cited the specific upstream variants the bridges re-join around (`coalesce(illuminate_grade_level_id, agl.grade_level_id)` and the `academic_year_clean` rename) so the re-join rationale is concrete.

### Item explicitly skipped

- Model-level `dbt_utils.unique_combination_of_columns` on the 12 hash inputs of `dim_assessment_administrations` (suggested as defense-in-depth). Mathematically redundant with the existing PK uniqueness on `assessment_administration_key` — for a deterministic hash, `unique(hash(N inputs))` ≡ `unique_combination_of_columns(N inputs)`. The PR-branch already has the PK test.

### Branch state

HEAD: `8b7e89786`. Pending: dbt Cloud CI re-run on the description/test cleanup, and Ops resolving #3737.

## Subsequent work

Commit `8b7e89786` — `feat(dbt): extend dim_survey_questions + bridge_survey_questions to cover Alchemer and Manager surveys`. Closes the `survey_question_key` half of #3766 (39,426 → 0 orphans). The 415K `survey_submission_key` half remains (root cause in `fct_survey_submissions` filter logic, outside this PR's modified surface).

PR description's #3766 entry moved from "Investigated, not closed" to "Partial close" reflecting that.

cbini:
@claude update your review

cbini:
Fixed in 1c298e919 —  order_by now `question_type nulls last, question_text asc`. Verified the underlying claim: `int_surveys__survey_responses` unions Google Forms + Alchemer, so GF questions did appear in both CTEs (one typed, one NULL).

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__college_assessment.yml
```diff
@@ -1,5 +1,13 @@
 models:
   - name: int_assessments__college_assessment
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          arguments:
+            combination_of_columns:
+              - student_number
+              - score_type
+              - test_date
+              - rn_highest
```
cbini: what are some examples of students taking the same test on the same date?

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__response_rollup.yml
```diff
@@ -1,5 +1,15 @@
...
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          arguments:
+            combination_of_columns:
+              - illuminate_student_id
+              - assessment_id
+              - response_type
+              - response_type_id
+              - response_type_code
+              - powerschool_school_id
```
cbini: why is powerschool_school_id necessary for uniqueness?

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__college_assessment.sql
```diff
@@ -239,64 +239,81 @@ with
...
-            then avg_running_max_superscore
-            when scope = 'SAT'
-            then sum_running_max_superscore
-        end,
-        0
-    ) as runnning_superscore,
-
-from score_calcs
+        }}
+    )
```
cbini: select * not necessary. use bare macro or put into cte

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
```diff
@@ -219,56 +219,71 @@ with
...
+from
+    (
+        {{
+            dbt_utils.deduplicate(
+                relation="enriched",
+                partition_by="illuminate_student_id, assessment_id, response_type, response_type_id, response_type_code, powerschool_school_id",
+                order_by="date_taken desc, points desc",
+            )
+        }}
+    )
```
cbini: select * not necessary. use bare macro or put into cte

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__scaffold.sql
```diff
@@ -23,6 +23,18 @@ with
...
+    -- canonical-school grain (one row per powerschool_school_id).
+    school_to_region as (
+        select distinct
+            location_powerschool_school_id as powerschool_school_id,
+            initcap(
+                regexp_extract(location_dagster_code_location, r'kipp(\w+)')
+            ) as region,
+        from {{ ref("int_people__location_crosswalk") }}
+        where location_powerschool_school_id != 0
+    ),
```
cbini: can we just join to stg_people__locations?

src/dbt/kipptaf/models/marts/bridges/bridge_assessment_expectations_student_scoped.sql
cbini, line 1: do these comments belong in the yaml description?

src/dbt/kipptaf/models/marts/bridges/bridge_survey_questions.sql
cbini, line 1: can these CTEs pull from upstream models to avoid distinct usage?

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
```diff
@@ -0,0 +1,211 @@
+with
+    -- DISTINCT projects from response grain to administration grain (one row
+    -- per scheduled occurrence). Internal assessments use administered_at as
+    -- the occurrence date; regions_assessed is a comma-separated list of
+    -- KIPP regions and is unnested to one row per region.
+    illuminate_administrations as (
+        select distinct
```
cbini: does distinct matter here? assessments are unique by assessment_id

src/dbt/kipptaf/models/marts/dimensions/dim_assessments.sql
cbini, line 1: can we look upstream to avoid distinct?

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
cbini, line 1: can we look upstream to avoid distinct?

src/dbt/kipptaf/models/marts/dimensions/dim_survey_questions.sql
cbini, line 1: can we look upstream to avoid distinct?

src/dbt/kipptaf/models/marts/dimensions/dim_surveys.sql
cbini, line 1: can we look upstream to avoid distinct?

src/dbt/kipptaf/models/surveys/intermediate/properties/int_surveys__survey_responses.yml
```diff
@@ -1,5 +1,13 @@
 models:
   - name: int_surveys__survey_responses
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          arguments:
+            combination_of_columns:
+              - survey_id
+              - survey_response_id
+              - respondent_identifier
```
cbini: is respondent_identifier required for uniqueness?

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
```diff
@@ -1,111 +1,146 @@
...
+            dbt_utils.deduplicate(
+                relation="with_identifier",
+                partition_by=(
+                    "survey_id, survey_response_id, "
+                    "respondent_identifier, question_shortname"
+                ),
+                order_by="answer asc",
+            )
+        }}
     )
```
cbini: not necessary. use bare macro or move to cte.

cbini: we should investigate the dupes. did we introduce them in this PR? this will have a downstream impact outside of the marts

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__college_assessment.yml
cbini: There are real dupes in prod — 173 of them. e.g. Student A has two rows for `(act_composite, 2023-02-01, rn_highest=1)` that differ on `superscore` (11 vs 14) but are otherwise identical. So the dedup is _not_ collapsing all dupes today, and the uniqueness test against `(student_number, score_type, test_date, rn_highest)` is failing (warn).

Two separate problems here:
1. `rn_highest` doesn't actually disambiguate the natural grain — most dup pairs share the same rn_highest value. The right fix is probably to drop `rn_highest` from the partition AND surface it as a flag on the row (so we keep the highest scoring record), then tighten the uniqueness test to `(student_number, score_type, test_date)`.
2. The upstream KIPPADB unpivot is producing duplicates that survive `rn_highest`-based partitioning.

This is out of scope for batch F (it's an existing data-quality issue, not introduced by this PR). I'll open a follow-up issue.

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__response_rollup.yml
cbini: `powerschool_school_id` is in the `GROUP BY` of both `internal_assessment_rollup` and `non_internal_rollup` (the two CTEs that feed `enriched`), so it's a natural grain dimension after aggregation. Dropping it from the uniqueness test would require also dropping it from the GROUP BYs upstream — at which point a student attending multiple schools in the same year would collapse, losing the school context the rollup currently preserves. The test as written matches the actual aggregation grain.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__college_assessment.sql
cbini: Fixed in 5746f7f75 — dropped the `select * from (...)` wrapper, dbt_utils.deduplicate is now the bare final macro.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
cbini: Fixed in 5746f7f75.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__scaffold.sql
cbini: Direct join doesn't quite work — `stg_google_sheets__people__locations.location_region` is the legal-entity name (e.g. "TEAM Academy Charter School"), not the friendly value used by `int_assessments__assessments.regions_assessed` ("Newark", "Camden", "Miami", "Paterson"). The crosswalk parse `initcap(regexp_extract(location_dagster_code_location, r'kipp(\w+)'))` is what produces the matching value.

The scaffold currently uses `int_people__location_crosswalk` (alias-grain) and DISTINCTs to canonical-school grain. We could instead parse `dagster_code_location` directly from `stg_google_sheets__people__locations` (already canonical grain — no DISTINCT needed). Want me to make that change here?

src/dbt/kipptaf/models/marts/bridges/bridge_assessment_expectations_student_scoped.sql
cbini: Moved to YAML description in 5746f7f75.

src/dbt/kipptaf/models/marts/bridges/bridge_survey_questions.sql
cbini: Each CTE projects from response grain to (survey_id, question_shortname) pair grain. The cleaner pattern would be a per-source `int_*__questions` model (or column projection in the existing intermediate) that already lives at pair grain — then the bridge becomes a plain UNION ALL with no DISTINCT.

This is the same pattern the user is flagging across `dim_assessments`, `dim_assessment_administrations`, `dim_surveys`, and `dim_survey_questions`. It's a consistent architectural cleanup but it touches every survey/assessment intermediate and would inflate this PR considerably. Should I open a follow-up issue ("Push entity-grain projections from marts into intermediate") and resolve there, or do you want it folded into batch F?

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
cbini: `int_assessments__assessments` is unique by `assessment_id`, but this CTE does `cross join unnest(split(regions_assessed, ','))` — one assessment with `regions_assessed = 'Newark, Camden, Miami'` becomes 3 rows. Live data also has both `'Camden,Newark'` and `'Camden , Newark'` (whitespace variants), and after `trim()` those collapse to the same `(assessment_id, region)` pair. DISTINCT is what catches that whitespace-variant collapse.

The cleaner fix is to push the unnest+trim into intermediate (covered by the `dim_*` distinct theme — see [reply on bridge_survey_questions](https://github.com/TEAMSchools/teamster/pull/3770#discussion_r3169785341)). Same scoping question applies.

src/dbt/kipptaf/models/marts/dimensions/dim_assessments.sql
cbini: Same theme as [bridge_survey_questions](https://github.com/TEAMSchools/teamster/pull/3770#discussion_r3169785341) and [dim_assessment_administrations](https://github.com/TEAMSchools/teamster/pull/3770#discussion_r3169786090). The fix is to push the response-grain → entity-grain projection into intermediate. Awaiting scope direction (this PR vs follow-up issue).

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
cbini: Same theme as [bridge_survey_questions](https://github.com/TEAMSchools/teamster/pull/3770#discussion_r3169785341) and [dim_assessment_administrations](https://github.com/TEAMSchools/teamster/pull/3770#discussion_r3169786090). The fix is to push the response-grain → entity-grain projection into intermediate. Awaiting scope direction (this PR vs follow-up issue).

src/dbt/kipptaf/models/marts/dimensions/dim_survey_questions.sql
cbini: Same theme as [bridge_survey_questions](https://github.com/TEAMSchools/teamster/pull/3770#discussion_r3169785341) and [dim_assessment_administrations](https://github.com/TEAMSchools/teamster/pull/3770#discussion_r3169786090). The fix is to push the response-grain → entity-grain projection into intermediate. Awaiting scope direction (this PR vs follow-up issue).

src/dbt/kipptaf/models/marts/dimensions/dim_surveys.sql
cbini: Same theme as [bridge_survey_questions](https://github.com/TEAMSchools/teamster/pull/3770#discussion_r3169785341) and [dim_assessment_administrations](https://github.com/TEAMSchools/teamster/pull/3770#discussion_r3169786090). The fix is to push the response-grain → entity-grain projection into intermediate. Awaiting scope direction (this PR vs follow-up issue).

src/dbt/kipptaf/models/surveys/intermediate/properties/int_surveys__survey_responses.yml
cbini: `respondent_identifier = coalesce(employee_number, email)` and is required because `survey_response_id` isn't unique across the GF + Alchemer UNION (Alchemer's `response_id` is namespaced to its survey, GF's `response_id` is its own ID space). Without `respondent_identifier`, two respondents who happened to land on the same `survey_response_id` collide.

That said, this ties directly to the dedup discussion below — if we push the dedup upstream and stop UNIONing two ID spaces under one column, the test could simplify.

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
cbini: Fixed in 5746f7f75.

cbini: Confirmed — this PR introduced the dupes. Before commit `cad690412` `int_surveys__survey_responses` was Google Forms only; the Alchemer + Manager UNION was added in this PR (issue #3629) and the dedup was a defensive add at the introducing layer.

Downstream impact: `fct_survey_responses`, `fct_survey_submissions`, and any consumer of `int_surveys__survey_responses` now see deduped output instead of raw response grain.

The right next step is to identify _where_ the dupes originate — likely either (a) Alchemer base data has the same response repeated (their internal `response_id` collisions), (b) cross-source `respondent_identifier` collisions when the same email exists in both GF and Alchemer for the same survey window, or (c) the GF-half itself has dupes that were previously hidden by the lack of a uniqueness test. Then push the fix upstream (Alchemer staging or GF intermediate).

Not a small investigation — should I open a follow-up issue, or treat it as in-scope for batch F and resolve here?

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__scaffold.sql
cbini: you can use city on stg_google_sheets__people__locations as well

src/dbt/kipptaf/models/marts/bridges/bridge_survey_questions.sql
cbini: let's add follow-up issues

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__response_rollup.yml
cbini: are there confirmed violations of this test without school id?

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__college_assessment.yml
cbini: let's investigate on this PR but only fix if it doesn't significantly expand the scope

src/dbt/kipptaf/models/marts/dimensions/dim_assessments.sql
cbini: follow-up

src/dbt/kipptaf/models/marts/dimensions/dim_surveys.sql
cbini: follow-up

src/dbt/kipptaf/models/marts/dimensions/dim_survey_questions.sql
cbini: follow-up

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
cbini: follow-up

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
cbini: fix upstream on this PR

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
cbini: this has to be fixed in this PR. it would be acceptable to make the minimal changes necessary and move disruptive changes downstream into the marts models if that's a simpler fix

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__response_rollup.yml
cbini: Yes — 1014 confirmed violations. Without `powerschool_school_id`: 23,008,308 distinct partition keys vs 23,009,322 rows. With school_id: 23,009,322 distinct (zero violations). So school_id is materially required.

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__college_assessment.yml
cbini: Investigated. PR-branch table is clean — 38,477 rows, all distinct on `(student, score_type, test_date, rn_highest)` (prod has 173 stale dupes that this PR collapses).

Real examples of "same test, same date" with different `rn_highest` (student numbers redacted):
- Student A, `sat_total_score`, 2015-10-19: scored 1440 (rn_highest=1) and 1320 (rn_highest=2) — two real SAT attempts on the same date.
- Student B, `sat_total_score`, 2019-03-16: 790 (rn_highest=1) and 790 (rn_highest=2) — same value, two distinct attempts.

Dropping `rn_highest` from the partition would collapse the rn_highest=2 records via `order_by scale_score desc` — for Student A we'd lose the 1320 attempt entirely. So `rn_highest` is doing real work; keeping it. No action needed on this PR.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__scaffold.sql
cbini: Done in 7d45f7ba2. Switched `school_to_region` to source `(powerschool_school_id, city)` from `stg_google_sheets__people__locations` directly. `city` is `Newark`/`Camden`/`Miami`/`Paterson`, matching `regions_assessed`. The crosswalk reference is gone, no more DISTINCT.

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
cbini: Done in 7d45f7ba2. New `int_assessments__administrations` model collapses to admin grain upstream (turned out multiple `assessment_id` rows can share the natural admin key — region-specific copies of the same module checkpoint — so this is the right place to dedupe). Also added a `regions_assessed_array` array column to `int_assessments__assessments` so the unnest+trim+dedup happens once at the source. `dim_assessment_administrations.illuminate_administrations` now selects from `int_assessments__administrations` without DISTINCT.

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
cbini: Resolved in 7d45f7ba2. Investigation showed dupes were not caused by the Alchemer UNION — Google Forms grid and check-all-that-apply questions naturally produce multiple rows per (response, question_id), one per row shortname or selected option. Pushed `question_id` through from Alchemer (it was being NULLed) and reverted the intermediate dedup. Uniqueness test now matches the actual data grain: `(survey_id, survey_response_id, survey_question_id, question_shortname, answer)`. 8 edge-case warnings remain (~0.0002% of ~4M rows), no errors. Downstream consumers see raw response grain again — same shape they had before this PR for GF, plus new Alchemer/Manager rows.

src/dbt/kipptaf/models/marts/bridges/bridge_survey_questions.sql
cbini: Tracked in #3780.

src/dbt/kipptaf/models/marts/dimensions/dim_assessments.sql
cbini: Tracked in #3780.

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
cbini: Tracked in #3780.

src/dbt/kipptaf/models/marts/dimensions/dim_survey_questions.sql
cbini: Tracked in #3780.

src/dbt/kipptaf/models/marts/dimensions/dim_surveys.sql
cbini: Tracked in #3780.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__assessments.sql
```diff
@@ -19,6 +19,12 @@ select
     iae.regions_report_card,
     iae.regions_progress_report,
 
+    array(
+        select distinct trim(r),
+        from unnest(split(iae.regions_assessed, ',')) as r
+        where trim(r) != ''
+    ) as regions_assessed_array,
```
cbini: move to staging

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__college_assessment.sql
```diff
@@ -239,64 +239,76 @@ with
...
-from score_calcs
+    -- 80% of dup groups tie on scale_score; superscore breaks the tie
+    -- deterministically.
+    {{
+        dbt_utils.deduplicate(
+            relation="enriched",
+            partition_by="student_number, score_type, test_date, rn_highest",
+            order_by="scale_score desc, superscore desc",
+        )
+    }}
```
cbini: we need to roll this back to preserve the state of production tools. can we can fix it downstream on a marts model and create a followup to do a root cause fix?

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
```diff
@@ -219,56 +219,66 @@ with
...
-    on ru.date_taken between rtt.start_date and rtt.end_date
-    and ru.powerschool_school_id = rtt.school_id
-    and rtt.type = 'RT'
+    {{
+        dbt_utils.deduplicate(
+            relation="enriched",
+            partition_by="illuminate_student_id, assessment_id, response_type, response_type_id, response_type_code, powerschool_school_id",
+            order_by="date_taken desc, points desc",
+        )
+    }}
```
cbini: we need to roll this back to preserve the state of production tools. can we can fix it downstream on a marts model and create a followup to do a root cause fix?

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__administrations.sql
cbini, line 1: I didn't intend for the fix to create a new model. Move these into dim_assessment_administrations as CTEs

cbini: Done in ebb5585cb. Inlined as `illuminate_unnested` CTE + `dbt_utils.deduplicate` directly in `dim_assessment_administrations.illuminate_administrations`. Deleted the standalone `int_assessments__administrations` model.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
cbini: Done in ebb5585cb. Reverted the dedup in `int_assessments__response_rollup` (back to pre-PR state — no test, no dedup). Re-added the `dbt_utils.deduplicate` workaround in `fct_assessment_scores_enrollment_scoped` so consumers outside the fact still see the same row count they did before this PR. Follow-up for root-cause fix: #3782.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__college_assessment.sql
cbini: Done in ebb5585cb. Reverted the dedup in `int_assessments__college_assessment` (back to pre-PR state). Re-added the `dbt_utils.deduplicate` workaround in `fct_assessment_scores_student_scoped`. Follow-up for root-cause fix: #3781.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__assessments.sql
cbini: Done in ebb5585cb. Moved the `regions_assessed_array` derivation into `stg_google_appsheet__illuminate_assessments_extension`. `int_assessments__assessments` now just passes the column through from staging.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__college_assessment.sql
cbini, line 1: this can be entirely reverted

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
cbini, line 1: this can be entirely reverted

src/dbt/kipptaf/models/google/appsheet/staging/stg_google_appsheet__illuminate_assessments_extension.sql
```diff
@@ -4,6 +4,12 @@ select
     grade_level + 1 as illuminate_grade_level_id,
 
     concat(module_type, module_sequence) as module_code,
+
+    array(
+        select distinct trim(r),
+        from unnest(split(regions_assessed, ',')) as r
+        where trim(r) != ''
+    ) as regions_assessed_array,
```
cbini: could we just do a simpler regex replace?

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
```diff
@@ -0,0 +1,235 @@
...
+                        relation="illuminate_unnested",
+                        partition_by=(
+                            "title, subject_area, scope, module_code, "
+                            "grade_level, administered_date, academic_year, "
+                            "region"
+                        ),
+                        order_by="title",
+                    )
+                }}
+            )
```
cbini: unnest into a CTE

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
cbini: do we have a followup issue for GF dupes and edge-cases?

src/dbt/kipptaf/models/surveys/intermediate/properties/int_surveys__survey_responses.yml
cbini: where did we land on this? is this one of the follow-ups?

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__response_rollup.yml
cbini: where did we land on this? is this one of the follow-ups?

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__college_assessment.yml
cbini: where did we land on this? is this one of the follow-ups?

cbini: Folded into #3781 (it's the same investigation — why does `int_kippadb__standardized_test_unpivot` produce these dupes, and what's the right uniqueness key once the upstream is clean). The current PR no longer adds a test on this model, so the question is dormant until #3781 is picked up.

src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__response_rollup.yml
cbini: Folded into #3782 — same investigation (whether `powerschool_school_id` is part of the natural grain depends on what's actually causing the upstream dupes). PR no longer adds a test on this model, so dormant until #3782.

src/dbt/kipptaf/models/surveys/intermediate/properties/int_surveys__survey_responses.yml
cbini: Resolved in this PR. The new uniqueness grain is `(survey_id, survey_response_id, survey_question_id, question_shortname, answer)` — `respondent_identifier` is no longer in the test. The column still exists on the model for downstream consumers to use as a stable respondent join key, but uniqueness is now keyed on source-system IDs.

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
cbini: Opened #3784 to track the residual 8 edge-case dupes.

src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql
cbini: Done in 641aea8dc. Pulled the deduplicate out into its own `illuminate_deduped` CTE; `illuminate_administrations` is now a clean SELECT with no inline subquery.

src/dbt/kipptaf/models/google/appsheet/staging/stg_google_appsheet__illuminate_assessments_extension.sql
cbini: Done in 641aea8dc. Now `split(regexp_replace(regions_assessed, r'\s+', ''), ',') as regions_assessed_array` — strips all whitespace then splits on comma. The dim's `dbt_utils.deduplicate` handles any within-row duplicates.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
cbini: Done in 641aea8dc — fully reverted to origin/main.

src/dbt/kipptaf/models/assessments/intermediate/int_assessments__college_assessment.sql
cbini: Done in 641aea8dc — fully reverted to origin/main.

cbini:
## Summary of actions since the [previous summary](https://github.com/TEAMSchools/teamster/pull/3770#issuecomment-4354207976)

Four review rounds: 36 inline comments addressed, 5 follow-up issues opened, the PR scope tightened back toward "preserve production-tool behavior."

### Round 1 — `dim_survey_questions` dedup determinism (1c298e919)

- Changed `dbt_utils.deduplicate` `order_by` so Google Forms rows with a real `question_type` win over `alchemer_questions` rows where `question_type` is NULL — the two CTEs sourced the same GF questions and a non-deterministic dedup could drop the typed row.

### Round 2 — initial review fixes + minor cleanups (5746f7f75)

- Unblocked CI: BigQuery rejects `asc nulls last` in `array_agg`; switched to `question_type desc` (default NULLS LAST for desc).
- Dropped `select * from (...)` wrappers around `dbt_utils.deduplicate` in three intermediates.
- Moved `bridge_assessment_expectations_student_scoped` SQL header comment into the YAML description.

### Round 3 — first pass at "fix upstream" feedback (7d45f7ba2)

- `int_assessments__scaffold.school_to_region` now sources `(powerschool_school_id, city)` directly from `stg_google_sheets__people__locations`. No more crosswalk DISTINCT.
- New `int_assessments__administrations` model with `dbt_utils.deduplicate` (multiple `assessment_id` rows can share the natural admin key — region-specific copies of the same module checkpoint). `dim_assessment_administrations` consumed it without DISTINCT.
- Added `regions_assessed_array` (split/trimmed/deduped array) on `int_assessments__assessments` so the unnest+trim happens once.
- Reverted `int_surveys__survey_responses` dedup. Pushed Alchemer's `question_id` through (was being NULLed). Uniqueness test moved to the natural data grain `(survey_id, survey_response_id, survey_question_id, question_shortname, answer)` — Google Forms grid + check-all questions legitimately produce multiple rows per `(response, question_id)`.

### Round 4 — preserve production-tool behavior (ebb5585cb, 641aea8dc, 8045b02f4)

Per direction, dedup work was moved out of intermediates and back into marts so existing consumers see unchanged row counts:

- Fully reverted `int_assessments__college_assessment` and `int_assessments__response_rollup` to `origin/main` (no dedup, no test). Re-added `dbt_utils.deduplicate` workarounds in `fct_assessment_scores_student_scoped` and `fct_assessment_scores_enrollment_scoped`.
- Moved `regions_assessed_array` derivation from `int_assessments__assessments` into `stg_google_appsheet__illuminate_assessments_extension` (right layer for source-data normalization). Simplified to `split(regexp_replace(regions_assessed, r'\s+', ''), ',')`.
- Inlined the deleted `int_assessments__administrations` model as `illuminate_unnested` + `illuminate_deduped` CTEs inside `dim_assessment_administrations`; `illuminate_administrations` is now a clean SELECT.

### PII redaction

- Edited two earlier review replies to redact specific student numbers (now "Student A" / "Student B").

### Resolved without code change

- **`rn_highest` in college_assessment uniqueness** — confirmed `rn_highest` is doing real work (e.g. one student with two SAT attempts on the same date scoring 1440 and 1320 — both must survive the dedup).
- **`powerschool_school_id` in response_rollup uniqueness** — confirmed required: 1014 actual violations on the natural grain without it.
- **`respondent_identifier` in survey_responses uniqueness** — no longer in the test (replaced by source-system IDs).

### Open follow-ups

| Issue | Scope |
|---|---|
| #3780 | Push entity-grain projections from marts into intermediate (5 dim/bridge DISTINCT cleanups deferred from batch F) |
| #3781 | Root-cause `int_assessments__college_assessment` dupes (workaround currently in `fct_assessment_scores_student_scoped`) |
| #3782 | Root-cause `int_assessments__response_rollup` dupes (workaround currently in `fct_assessment_scores_enrollment_scoped`) |
| #3784 | Investigate 8 residual `int_surveys__survey_responses` GF edge-case dupes |
| #3785 | Root-cause illuminate admin-grain dupes in `int_assessments__assessments` (workaround currently in `dim_assessment_administrations.illuminate_deduped`) |

CI green at HEAD.

cbini:
@claude update your review

cbini (commented):
## Code Review — batch F (assessment/survey catalog refactor)

**Verdict: Ready with fixes.** No Critical blockers, but 4 Important items to address before merge.

### Strengths

- **Hash-input ordering verified consistent** across all 5 fact branches (illuminate, state_nj, state_fl, college, AP) and both bridges — the 12-column input order matches `dim_assessment_administrations.assessment_administration_key`. This is the highest-risk surface area of the PR and it's correct.
- Test coverage (`unique_combination_of_columns`) matches grain claims on every new model. The "dedup at int layer + drop `dbt_utils.deduplicate()` from facts" pattern matches the conventions in `src/dbt/CLAUDE.md`.
- `regions_assessed_array` lifted to staging eliminates `split(...)` repetition across scaffold, dim, and downstream models.
- All upstream references via `ref()`/`source()`. No hardcoded table names introduced.
- No dangling refs to the deleted `dim_student_assessment_expectations`; `cube.yml` exposure updated to add both new bridges + `dim_assessment_administrations`.
- Documented deviations from the plan (Task 7 PK extension to 12 cols, Task 14 fact-side `grade_level` fix, Task 12 NULL `cc_dcid` filter) all surfaced in the PR body — no silent departures.

### Issues

#### Important

1. **`test_type` excluded from `assessment_administration_key` may cause collisions in college branch** — `src/dbt/kipptaf/models/marts/dimensions/dim_assessment_administrations.sql:127-138`. The college CTE pulls from `int_assessments__college_assessment`, which contains both `Official` and `Practice` SAT/ACT rows, but the 12-column surrogate-key hash omits `test_type`. If any student has both an Official and a Practice record in the same `(score_type, test_date, administration_round)` window, two distinct CTE rows generate the **same** `assessment_administration_key` → uniqueness violation in the dim and a fact-side FK collision. **Verify before merge:**
   ```sql
   select score_type, test_date, administration_round,
          count(*) as rows, count(distinct test_type) as test_types
   from `kipptaf_assessments.int_assessments__college_assessment`
   group by 1,2,3
   having count(*) > count(distinct test_type)
   ```
   If any rows return, add `test_type` to the surrogate-key hash in the dim and in `fct_assessment_scores_student_scoped.sql` college CTE.

2. **2.7M-row enrollment bridge defaults to `view`** — `src/dbt/kipptaf/models/marts/bridges/bridge_assessment_expectations_enrollment_scoped.sql`. Per `dbt_project.yml`, `marts/` is view by default. With Cube downstream joining `dim_students` → `dim_student_section_enrollments` → bridge → `dim_assessment_administrations`, BigQuery's 16-level nested-view limit becomes reachable. Add `{{ config(materialized='table') }}` (the small student-scoped bridge at 1.3K rows can stay a view).

3. **`regions_assessed_array` YAML claim is false** — `src/dbt/kipptaf/models/google/appsheet/staging/stg_google_appsheet__illuminate_assessments_extension.sql:7` does `split(regexp_replace(regions_assessed, r'\s+', ''), ',')` — **not** deduped, **not** empty-string-filtered — but the column description in `stg_google_appsheet__illuminate_assessments_extension.yml` and `int_assessments__assessments.yml` claims it is. Either fix the SQL (`array(select distinct x from unnest(...) as x where x != '')` — preferred, eliminates risk of `region = ''` rows that won't FK-match real regions from `int_people__location_crosswalk`) or fix the description in both YAML files.

4. **Redundant `cc_dcid is not null` filter unexplained** — `src/dbt/kipptaf/models/marts/bridges/bridge_assessment_expectations_enrollment_scoped.sql:30-31`. The filter has both `not is_replacement` and `cc_dcid is not null`. Per `int_assessments__course_enrollments.sql:54-55`, `cc_dcid` is null only on synthesized ELA-replacement branch rows — so `not is_replacement` should already cover them. The PR body's 116K-row figure implies non-replacement rows also have null `cc_dcid` (likely from LEFT JOIN behavior in the K-12 scaffold branch when course-enrollment match fails). Add a comment explaining the filter, and verify whether these are legitimately unmatchable students or a join bug upstream in `int_assessments__scaffold`.

#### Minor

5. **DISTINCT-as-projection comments lost their `-- TODO:` marker** in `dim_assessments.sql`. Per `src/dbt/CLAUDE.md`, `DISTINCT` requires a `-- TODO:` comment for grep-ability. The new comments explain the rationale clearly but no longer carry the marker — keep it for tooling, or amend the convention.

6. **Mix of `select distinct` and `dbt_utils.deduplicate`** in `dim_assessment_administrations` CTEs (illuminate uses `deduplicate`, state/college/AP use `select distinct`). Pick one for consistency.

7. **`bridge_survey_questions.sql` `order_by="is_required desc nulls last"`** correctly prefers concrete booleans over null casts when alchemer/manager-vs-SCD pairs collide — but the choice is subtle and undocumented. Add a one-liner.

8. **`int_assessments__scaffold.sql:28-32` school_to_region uses `stg_google_sheets__people__locations` (canonical grain)** rather than `int_people__location_crosswalk` (alias grain). This is **correct** per `kipptaf/CLAUDE.md`, but worth a one-line comment so a future maintainer doesn't "fix" it to use the crosswalk.

### Recommendations

1. Run the `test_type` collision query (#1) before merge.
2. Materialize the enrollment-scoped bridge as `table` (#2).
3. Fix the `regions_assessed_array` SQL/YAML discrepancy (#3).
4. Wait for Ops on #3737 — already listed as blocker in PR body.
5. Open a follow-up issue for #3766's `survey_submission_key` half (good scoping discipline to carve it out here).

### Plan Alignment

Implementation tracks the spec and plan closely. None of the Important issues indicate fundamental design problems — they are implementation-detail gaps in an otherwise carefully-executed refactor.

---

cbini:
## Response to round-5 review

Two reviews evaluated; one Important pushback, two Important fixed, all Minors fixed or scoped.

### Important #1 — `test_type` collision in college branch — pushback (verified)

`int_assessments__college_assessment` is Official-only:

```sql
select distinct test_type from `kipptaf_assessments.int_assessments__college_assessment`
-- → 'Official' (single row)
```

The Practice variant lives in a separate model `int_assessments__college_assessment_practice` which the dim doesn't consume. The collision query also returned `n_test_types = 1` for every `(score_type, test_date, administration_round)` window. No collision possible at the current grain — adding `test_type` to the hash isn't required for correctness today.

That said, **Practice rows are missing from the semantic layer**. Tracking this as a coverage gap in **#3787** — when Practice rows land in the dim/fct, `test_type` should join the hash inputs.

### Important #2 — bridge materialization — keep as `view`

Cube depends on this bridge but production-tool query patterns don't currently chain enough mart views to hit BigQuery's 16-view nesting limit. The 2.7M-row size doesn't warrant overriding the project default. Leaving as `view`.

### Important #3 — `regions_assessed_array` YAML description inaccurate — fixed in `af54afc32`

The prior description claimed trim/dedup/empty-filter, but the simplified regex (`split(regexp_replace(..., r'\s+', ''), ',')`) does only a whitespace strip + comma split. Updated descriptions in both `stg_google_appsheet__illuminate_assessments_extension.yml` and `int_assessments__assessments.yml` to match actual behavior, and to point at where residual edges (within-row dupes / empty strings) get resolved (`illuminate_deduped` CTE in the dim).

### Important #4 — `cc_dcid is not null` filter — documented in `af54afc32`

Verified: 116K `is_internal_assessment AND NOT is_replacement` rows in the scaffold have NULL `cc_dcid`. Source: ES Writing rows in `int_assessments__course_enrollments` synthesize `cast(null as int64) as cc_dcid` because no PowerSchool course-enrollment record exists for them. Without filtering, all NULL rows would hash to the same `generate_surrogate_key` placeholder on `student_section_enrollment_key` and collide on the bridge's PK. Added an inline comment explaining this.

### Minor — addressed

- **#5** (DISTINCT-as-projection `-- TODO:` markers in `dim_assessments`) — folded into existing **#3780**; that issue covers pushing entity-grain projections from marts into intermediate.
- **#6** (mix of `select distinct` / `dbt_utils.deduplicate` in `dim_assessment_administrations`) — same; consistency falls out of #3780.
- **#7** (`bridge_survey_questions` `order_by` documentation) — added inline comment in `af54afc32`.
- **#8** (`school_to_region` design choice comment) — added inline comment in `af54afc32`.

### Observations — pre-existing gaps

`stg_google_appsheet__illuminate_assessments_extension.yml` and `int_assessments__assessments.yml` lack uniqueness tests / model descriptions. Out of scope for batch F.

### Updated follow-up issue map

| Issue | Scope |
|---|---|
| #3780 | Push entity-grain projections from marts into intermediate (5 dim/bridge DISTINCT cleanups) |
| #3781 | Root-cause `int_assessments__college_assessment` dupes |
| #3782 | Root-cause `int_assessments__response_rollup` dupes |
| #3784 | Investigate residual GF edge-case dupes in `int_surveys__survey_responses` |
| #3785 | Root-cause illuminate admin-grain dupes in `int_assessments__assessments` |
| **#3787** | Add SAT/ACT Practice rows to `dim_assessment_administrations` + `fct_assessment_scores_student_scoped` |

cbini:
Related: the residual ~105 Paterson 2026-02-02 rows with NULL `term_key` after this PR were caused by date gaps in the Paterson RT sheet for `school_id = 1234` (Paterson Prep Elementary), not a code issue. Closed in #3737 via sheet edits — no follow-up code change needed.

PR #3843, opened 2026-05-07
chore(cube): sync dbt descriptions into cube YAMLs and enrich attendance views

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will:

- Add a one-shot utility script (`scripts/sync_cube_descriptions.py`) that reads dbt mart column descriptions and patches matching Cube dimension YAMLs in place. The script is **not wired into CI** — future runs are manual (`uv run --with "ruamel.yaml>=0.18" scripts/sync_cube_descriptions.py`).
- Apply the first-ever sync run: 69 dimensions across 8 cube files now carry descriptions sourced from dbt marts (`fct_student_attendance_daily`, `dim_students`, `dim_school_calendars`, `dim_student_enrollments`, `dim_dates`, `dim_locations`, `dim_regions`, `dim_terms`).
- Remove two `dim_students` dimensions (`has_iep`, `meal_eligibility_status`) whose underlying warehouse columns are absent from the current mart — keeping them would have surfaced broken metadata in Cube Cloud.
- Add view-level `description` and `meta.folders` entries to both attendance views (`attendance_summary.yml`, `attendance_detail.yml`), making them discoverable in Cube Cloud's folder browser.
- Correct three prose typos (`eachone`, `usually be1`, `attendancevalue`) in `fct_student_attendance_daily.yml`'s upstream column descriptions, and propagate the fixes to the cube YAML.

Cube Cloud production redeploys automatically on merge to `main`; metadata changes take effect after that redeploy with no manual deploy step required.

> **AI Assistance:** Script, tests, and fixture files were fully AI-generated (Claude Code). Cube YAML edits were applied by running the generated script against the local model. Human direction: chose sync approach, approved design, reviewed output.

### General

- [ ] Run \`uv run dagster definitions validate\` for any modified code location
- [ ] Run \`uv run pytest tests/test_dagster_definitions.py\` for any modified
      code location

- [ ] Include a \`[model_name].yml\` properties file for all new models — see
- [ ] Include (or update) an
      (\`stg_\`, \`rpt_\`, mart) will break downstream exposures. Coordinate with
      summary above.
- [ ] If adding or modifying an external source, run \`stage_external_sources\`
      with **\`--target staging\`** so the dbt Cloud CI job can find the table.
      See

      catalog: \`uv run scripts/gen-automations-doc.py\`
      \`CLAUDE.md\` by running \`/claude-md-management:revise-claude-md\`

## CI checks

- [ ] **Trunk** — passes. If it fails, run \`trunk check\` and \`trunk fmt\`
      expand **Invoke \`dbt build ...\`** → **Debug Logs**. Or tag **Claude** on

## Troubleshooting

### DIFF
--- a/scripts/sync_cube_descriptions.py
+++ b/scripts/sync_cube_descriptions.py
@@ -22,6 +22,19 @@ import re
 _COLUMN_RE = re.compile(r"^[a-z_][a-z0-9_]*$")
 
 
+_EXPECTED_SCHEMA = "kipptaf_marts"
+
+
+def _resolve_table_from_sql_table(sql_table: str | None) -> str | None:
+    """Extract ``<table>`` from ``kipptaf_marts.<table>``; ``None`` otherwise."""
+    if not isinstance(sql_table, str):
+        return None
+    parts = sql_table.split(".", 1)
+    if len(parts) != 2 or parts[0] != _EXPECTED_SCHEMA:
+        return None
+    return parts[1].strip() or None
+
+
 def _resolve_dbt_column(sql: str | None) -> str | None:
     """Return the dbt column name from a Cube dimension's ``sql:`` value.
 
--- a/scripts/sync_cube_descriptions.py
+++ b/scripts/sync_cube_descriptions.py
@@ -17,7 +17,9 @@ Usage:
 
 from __future__ import annotations
 
+import argparse
 import re
+import sys
 from pathlib import Path
 
 import yaml
@@ -31,6 +33,8 @@ _DBT_MART_DIRS: tuple[Path, ...] = (
 )
 
 _EXPECTED_SCHEMA = "kipptaf_marts"
+
+_CUBE_MODEL_DIR = Path("src/cube/model/cubes")
 
 
 def _load_dbt_descriptions(
@@ -166,3 +170,37 @@ def _patch_cube_file(
         with path.open("w") as f:
             ry.dump(doc, f)
     return counts
+
+
+def main(argv: list[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--check",
+        action="store_true",
+        help="Exit non-zero if any file would change.",
+    )
+    args = parser.parse_args(argv)
+    files = sorted(_CUBE_MODEL_DIR.rglob("*.yml"))
+    total: dict[str, int] = {
+        "updated": 0,
+        "skipped_already": 0,
+        "skipped_expr": 0,
+        "skipped_no_match": 0,
+        "skipped_no_table": 0,
+    }
+    any_changes = False
+    for f in files:
+        counts = _patch_cube_file(f)
+        if counts["updated"]:
+            any_changes = True
+            print(f"{f}: {counts['updated']} updated")
+        for k, v in counts.items():
+            total[k] = total.get(k, 0) + v
+    print(f"\nTotal: {total}")
+    if args.check and any_changes:
+        return 1
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/scripts/sync_cube_descriptions.py
+++ b/scripts/sync_cube_descriptions.py
@@ -1,0 +1,18 @@
+# /// script
+# requires-python = ">=3.13"
+# dependencies = ["pyyaml>=6.0", "ruamel.yaml>=0.18"]
+# ///
+
+"""Sync dbt mart YAML descriptions into Cube cube YAML files.
+
+Reads each cube file under ``src/cube/model/cubes/``, locates the matching
+dbt mart YAML via ``sql_table:``, and inserts dbt column descriptions onto
+cube dimensions that lack them. Patches dimensions only — measures are
+calculations and require hand-authored descriptions. Existing Cube
+descriptions are preserved.
+
+Usage:
+    uv run scripts/sync_cube_descriptions.py [--check]
+"""
+
+from __future__ import annotations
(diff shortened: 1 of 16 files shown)


cbini:
All review items addressed in `ba1bd33d6` and `54e6a5fc5`.

**Important — fixed:**

1. **`from __future__ import annotations`** removed from `scripts/sync_cube_descriptions.py` and `tests/test_sync_cube_descriptions.py`. Confirmed against `dignified-python/versions/python-3.13.md` ("DO NOT use `from __future__ import annotations`"). Note: `scripts/propagate_mart_descriptions.py` (the precedent that was used as the pattern) also has it — pre-existing drift, not addressed in this PR.
2. **`encoding="utf-8"`** added to `path.read_text()` (line 53), `path.open()` (line 137), and `path.open("w")` (line 170). Aligns with `propagate_mart_descriptions.py` which specifies `encoding="utf-8"` everywhere.

**Minor — fixed:**

3. **PR description** corrected: `disability_type_code` / `special_education_classification` → `has_iep` / `meal_eligibility_status`. The wrong names came from the agent that drafted the PR body; the actual diff was always correct.
4. **Prose typos** fixed at the source in `src/dbt/kipptaf/models/marts/facts/properties/fct_student_attendance_daily.yml` (`eachone` → `each one`, `usually be1` → `usually be 1`, `attendancevalue` → `attendance_value`) and propagated manually to `src/cube/model/cubes/attendance/attendance.yml` since the script preserves existing descriptions.
5. **Accumulator simplified**: `total[k] = total.get(k, 0) + v` → `total[k] += v`.
6. **`sys.modules` registration** added to the test loader. One citation correction: this convention is documented in `scripts/CLAUDE.md` ("Testing standalone PEP 723 scripts"), not `tests/CLAUDE.md`. The reviewer's substance is correct; the script has no `@dataclass` so the registration is currently belt-and-suspenders, but it aligns with the documented project rule.

**Verification post-fix:**
- 18/18 pytest pass.
- `--check` exits 0 (`updated: 0`).
- Trunk clean across all touched files.

cbini:
@claude update your review

cristinabaldor (commented):
## View metadata feedback

### `meta.folders` — rename `Outcome` to `Attendance`

The `Outcome` folder in both views groups things that aren't really outcomes. In the detail view especially, it mixes measures (`avg_daily_attendance`, `pct_tardy`, etc.) with raw attendance event dimensions (`attendance_code`, `attendance_category`, `attendance_value`, `is_oss`, `is_iss`, etc.). `is_oss` is a suspension event type, not an outcome of a pattern.

Also worth noting: Cube Cloud already separates measures from dimensions natively in the field picker, so measures don't need a named folder to be discoverable.

**Suggestion:** rename `Outcome` → `Attendance` in both views. The raw event dimensions (codes, values, flags) are what the folder is actually grouping — that name fits. Measures will surface on their own.

---

### `meta.usage` — remove, fold relevant guidance into `description`

`meta` in Cube is a free-form passthrough object intended for application-specific metadata that a custom downstream tool reads programmatically from the `/v1/meta` API (e.g. a custom frontend using `meta.display_format` to control rendering). `meta.folders` works because Cube Cloud explicitly recognizes that key; arbitrary keys like `meta.usage` are just passed through and nothing reads them.

Concretely: the Cube Cloud UI won't surface `meta.usage` in the field browser, and the Cube MCP server — which enables LLM queries against this semantic layer — reads dimension/measure `description` fields, not arbitrary `meta` keys.

The guidance is good but in the wrong place:
- `ada` — redundant with `avg_daily_attendance` already being a named measure with a description
- `point_in_time` and `truancy_regional` — useful context that would actually reach end users and LLMs if folded into the view-level `description`, which this PR already fills in well

**Suggestion:** drop `meta.usage` from both views. Check whether `point_in_time` and `truancy_regional` guidance is already covered in the view `description`; if not, add it there.

---

## Script minor notes (FYI, not blocking)

- **`skipped_no_table` conflates two different failure modes** — "cube references a non-`kipptaf_marts` schema" (intentional) and "dbt YAML not found for a `kipptaf_marts` table" (unexpected, could mask a broken path) both increment the same counter. Consider splitting into `skipped_wrong_schema` and `skipped_no_dbt_yaml`.

- **`_patch_cube_file` silently processes only `cubes[0]`** — fine given every cube file in this repo has exactly one cube, but it's an implicit assumption. An `assert len(cubes) == 1` would make it explicit and catch any future file that breaks the pattern.

- **No test for `--check` / `main()`** — low priority for a one-shot script, but the `--check` exit-code behavior is untested.

- **`ruamel.yaml` in the test environment** — tests that call `_patch_cube_file` will invoke `_get_ryaml()`. Confirm `uv run pytest` has `ruamel.yaml` available (either via `--with "ruamel.yaml>=0.18"` or added to dev deps), otherwise those tests will fail outside of a `uv run` context.

cbini:
Follow-up at #3844 (off `main` since this PR is merged).

**Implemented:**

- **`Outcome` → `Attendance`** in both views, measures removed from the folder. Agreed on the framing: the folder groups raw event dimensions; Cube Cloud picks up measures separately.
- **`meta.usage` dropped** from both views. Confirmed it's free-form passthrough that doesn't reach the chat agent. Summary view's description already had all three guidance bullets; added point-in-time and regional-truancy sentences to detail view's description so nothing is lost.
- **`skipped_no_table` split** into `skipped_wrong_schema` (cube → non-`kipptaf_marts` schema) and `skipped_no_dbt_yaml` (kipptaf_marts table missing its dbt YAML). New tests cover both branches.
- **Multi-cube guard**: `_patch_cube_file` raises `ValueError` if a file declares more than one cube under `cubes:`. Used `raise ValueError` over `assert` per project trunk/bandit posture (asserts get stripped under `-O`). Regression test added.

**Deferred (per "FYI, not blocking"):**

- `main()` / `--check` test — would require turning `_CUBE_MODEL_DIR` into a parameter. Low value for a one-shot script.
- `ruamel.yaml` available without `--with` — the canonical invocation in this repo's PEP 723 testing convention is `uv run --with "ruamel.yaml>=0.18" pytest tests/test_sync_cube_descriptions.py`. No precedent for promoting script-only deps to project dev deps.

20/20 pytest pass; `--check` exits 0 with `updated: 0`; trunk clean on the four touched files.

PR #3606, opened 2026-04-09
perf(dbt): optimize tableau gradebook and topline weekly materializations

cbini:
# Pull Request

## Summary & Motivation

> "When merged, this pull request will..."

Reduce BigQuery costs by restructuring the gradebook audit pipeline and topline
weekly models. The original monolithic `int_tableau__gradebook_audit_flags`
(1,091-line 6-branch UNION ALL) was split, aggregation patterns were refactored,
and materialization was guided by query plan analysis.

### Structural refactoring

- **Split monolithic flags model** into student/teacher sub-models:
  - `int_tableau__gradebook_audit_flags__student` (view, 4-branch UNION ALL)
  - `int_tableau__gradebook_audit_flags__teacher` (view, 2-branch UNION ALL)
- **Split multi-grain scaffolds** into single-grain models with proper
  uniqueness tests:
  - `int_tableau__gradebook_audit_section_week_scaffold` (view)
  - `int_tableau__gradebook_audit_section_week_category_scaffold` (view)
  - `int_tableau__gradebook_audit_section_week_student_scaffold` (view)
  - `int_tableau__gradebook_audit_section_week_category_student_scaffold` (view)
- **New teacher aggregation model**:
  - `int_tableau__gradebook_audit_teacher_aggs` (**table** — only materialized
    intermediate) — replaces large GROUP BY patterns with `logical_or()`
    aggregation from pre-UNPIVOT sources using 3-7 column GROUP BY keys
- **Fixed upstream duplication**: removed redundant expectations join in student
  scaffold that was causing 2x row inflation

### Materialization strategy (query-plan-driven)

All intermediate models default to views. Query plan analysis on BigQuery
identified `teacher_aggs` as the single highest-leverage materialization point:

| Metric | All views | teacher_aggs as table | Reduction |
|---|---|---|---|
| Stages (gradebook_audit) | 934 | 419 | 55% |
| Slot-ms (gradebook_audit) | 14.2M | 4.7M | 67% |
| Stages (es_comments) | 473 | 237 | 50% |
| Stages (ms_hs_comments) | 356 | 121 | 66% |

Materializing `__student` (30M rows) was tested but rejected — it increases
bytes processed 5x (2 GB to 12 GB) with diminishing returns.

### Reporting models

- **Reverted to views**: all 4 `rpt_tableau__gradebook_*` reports and 6
  `int_topline__*_weekly` models (consumed ~270x/month by Tableau, were being
  materialized ~1,400x/month by Dagster)
[...]

PR #3862, opened 2026-05-11
fix(dbt-kipptaf): move is_current_week to rpt view and convert to view

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will:
- Move `is_current_week` computation from `int_tableau__gradebook_audit_teacher_scaffold` (a materialized TABLE) into `rpt_tableau__gradebook_audit` so the value is always evaluated dynamically at query time
- Convert `rpt_tableau__gradebook_audit` from `materialized: table` back to the `extracts/` default (view)

**Root cause of the incident (May 11, 2026):** The week of May 4–10 was missing from the gradebook audit dashboard. `is_current_week` was computed in the scaffold TABLE at materialization time. During an unusually quiet 6-day stretch (May 5–11) with no upstream data changes, Dagster's automation condition (`deps_updated SINCE last_updated`) had no trigger — so the scaffold was never rebuilt at week turnover. The stale `is_current_week = true` value for May 4–10 filtered those rows out of the report.

**Why convert back to a view:** The rpt was originally converted to a TABLE on 2026-03-20 because Tableau refreshes timed out (~2 GB scan, >1 hour). Since then, Q1 and Q2 data were removed from the dataset, reducing load to ~1M rows. A view is now viable and eliminates the stale-data risk entirely.

**Row count validation:** New view logic produces 1,077,352 rows vs prod table's 1,077,115 — a 237-row difference attributable to data changes between the table build and query time. No rows are lost.

## AI Assistance

Fully AI-assisted investigation and implementation. Human-directed: scope decision (targeted fix only — not Charlie's full PR #3606 refactor which removes rows for teachers with no errors).

### General

- [ ] Include a `[model_name].yml` properties file for all new models — see [dbt Conventions](https://teamschools.github.io/teamster/reference/dbt-conventions/#model-properties-file)
- [ ] Include (or update) an [exposure](https://teamschools.github.io/teamster/reference/dbt-conventions/#exposures) for all models consumed by a dashboard, analysis, or application
- [ ] **Breaking change?** Renaming or removing columns in a contracted model (`stg_`, `rpt_`, mart) will break downstream exposures. Coordinate with affected teams before merging and document your rollback plan in the summary above.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

---
  - https://app.asana.com/0/0/1214717502238939


GabyRangelB:

anthonygwalters (approved):

PR #3925, opened 2026-05-14
perf(amplify-mclass-sftp-sensor): prune unchanged subtrees via dir_mtimes cursor

### LINKED ISSUE #3924: perf(amplify-mclass-sftp-sensor): prune unchanged subtrees via dir_mtimes cursor
## Background

The Amplify mClass SFTP sensor (`build_amplify_mclass_sftp_sensor` in [src/teamster/libraries/amplify/mclass/sftp/sensors.py](../blob/main/src/teamster/libraries/amplify/mclass/sftp/sensors.py)) calls `SSHResource.listdir_attr_r()` unconditionally every tick — a recursive walk of the entire Amplify drop tree. When Amplify's SFTP server is slow, the sensor function exceeds Dagster's 600s execution cap and the tick fails with `DagsterUserCodeUnreachableError`.

Observed twice on 2026-05-13/14 day-2 window (kippnewark + kipppaterson, single-tick blips, no run impact, recovered on next tick).

## Fix

`SSHResource.listdir_attr_r` ([src/teamster/libraries/ssh/resources.py:35](../blob/main/src/teamster/libraries/ssh/resources.py#L35)) already accepts a `dir_mtimes: dict[str, float]` parameter that prunes recursion into any subdir whose current `st_mtime <= cached_mtime`. The Amplify mClass sensor doesn't currently wire this through.

Change: store `dir_mtimes` in the sensor cursor alongside the existing per-asset mtime map, pass it into `listdir_attr_r`, persist the mutated dict back to the cursor. Stable subtrees collapse to a single readdir at the root after the first tick.

## Acceptance

- Cursor structure additive (existing per-asset cursor keys preserved for backward compatibility on first tick after deploy)
- `uv run dagster definitions validate -m teamster.code_locations.kippnewark.definitions` passes
- No behavior change on first tick (empty `dir_mtimes` → full walk, populates dict)
- Subsequent ticks skip recursion into unchanged subdirs

## Caveat

Some SFTP servers don't bump directory `st_mtime` on file create/delete inside the directory. Amplify's behavior should be confirmed before merging — if it doesn't bump mtime, the pruning is a no-op (safe, but ineffective). Fallback alternative: tick-rotated subdir traversal (separate issue if needed).

cbini:
# Pull Request

## Summary & Motivation

> When merged, this pull request will cut the per-tick work the Amplify mClass SFTP sensor does when Amplify's directory tree is stable, by passing a `dir_mtimes` cache through the sensor cursor so unchanged subdirectories are skipped during recursion.

The sensor was calling `SSHResource.listdir_attr_r()` unconditionally every tick — a full recursive walk of the Amplify drop tree. Twice in the 2026-05-13/14 day-2 window (kippnewark + kipppaterson) the listing exceeded Dagster's 600s sensor execution cap and the tick failed with `DagsterUserCodeUnreachableError`. Both recovered on the next tick (single-tick blips, no run impact), but the failure mode is structural.

`SSHResource.listdir_attr_r` already accepts a `dir_mtimes: dict[str, float]` that prunes recursion into subdirs whose `st_mtime <= cached_mtime` (see `src/teamster/libraries/ssh/resources.py:55-71`). This PR just wires that through:

- Read `__dir_mtimes__` from the existing cursor JSON (reserved key, namespaced underscore so it can't collide with `asset.key.to_python_identifier()` output)
- Pass it into `listdir_attr_r` — the function mutates it in place
- Persist back to the cursor

First tick after deploy starts with an empty dict, does a full walk, populates the cache. Subsequent ticks short-circuit on unchanged subdirs.

## AI Assistance

AI-assisted: Claude diagnosed the two day-2 sensor timeouts, identified the existing unused `dir_mtimes` parameter, and wired it through. Human-directed: triage and approval of approach.

Refs #3924

### Dagster

- [x] Skipped `uv run dagster definitions validate` — codespace env-var gaps misfire it per `src/teamster/CLAUDE.md`. Verified via `uv run python -c "from teamster.libraries.amplify.mclass.sftp import sensors"` instead.
- [ ] `uv run pytest tests/test_dagster_definitions.py` — N/A, no test changes; sensor factory has no direct test.

### Caveat

Some SFTP servers don't bump directory `st_mtime` when files inside change. If Amplify's server doesn't, the pruning is a no-op (safe, but ineffective). Worth watching tick durations on `kippnewark__amplify__mclass__sftp_asset_job_sensor` and `kipppaterson__amplify__mclass__sftp_asset_job_sensor` post-deploy. Fallback plan: tick-rotated subdir traversal (separate change).

## CI checks

- [ ] Trunk
- [ ] dbt Cloud
- [ ] Dagster Cloud

### DIFF
--- a/src/teamster/libraries/amplify/mclass/sftp/sensors.py
+++ b/src/teamster/libraries/amplify/mclass/sftp/sensors.py
@@ -20,6 +20,8 @@ from dagster_shared import check
 
 from teamster.libraries.ssh.resources import SSHResource
 
+DIR_MTIMES_KEY = "__dir_mtimes"
+
 
 def build_amplify_mclass_sftp_sensor(
     code_location: str,
@@ -56,7 +58,7 @@ def build_amplify_mclass_sftp_sensor(
         run_request_kwargs = []
         run_requests = []
         cursor: dict = json.loads(context.cursor or "{}")
-        dir_mtimes: dict[str, float] = cursor.get("__dir_mtimes__", {})
+        dir_mtimes = cursor.pop(DIR_MTIMES_KEY, {})
 
         with (
             ssh_amplify.get_connection() as connection,
@@ -125,7 +127,7 @@ def build_amplify_mclass_sftp_sensor(
                 )
             )
 
-        cursor["__dir_mtimes__"] = dir_mtimes
+        cursor[DIR_MTIMES_KEY] = dir_mtimes
 
         return SensorResult(run_requests=run_requests, cursor=json.dumps(obj=cursor))
 
--- a/src/teamster/libraries/amplify/mclass/sftp/sensors.py
+++ b/src/teamster/libraries/amplify/mclass/sftp/sensors.py
@@ -56,13 +56,16 @@ def build_amplify_mclass_sftp_sensor(
         run_request_kwargs = []
         run_requests = []
         cursor: dict = json.loads(context.cursor or "{}")
+        dir_mtimes: dict[str, float] = cursor.get("__dir_mtimes__", {})
 
         with (
             ssh_amplify.get_connection() as connection,
             connection.open_sftp() as sftp_client,
         ):
             files = ssh_amplify.listdir_attr_r(
-                sftp_client=sftp_client, remote_dir=remote_dir_regex
+                sftp_client=sftp_client,
+                remote_dir=remote_dir_regex,
+                dir_mtimes=dir_mtimes,
             )
 
         for asset in asset_selection:
@@ -121,6 +124,8 @@ def build_amplify_mclass_sftp_sensor(
                     asset_selection=[g["asset_key"] for g in group],
                 )
             )
+
+        cursor["__dir_mtimes__"] = dir_mtimes
 
         return SensorResult(run_requests=run_requests, cursor=json.dumps(obj=cursor))
 


cbini:
Addressed in `729b06c3f` — a different fix than what was suggested, because grepping for `dir_mtimes` turned up a prior precedent I'd missed: `src/teamster/code_locations/kipptaf/adp/workforce_now/sftp/sensors.py:17,31,77`. That sensor uses a module-level `DIR_MTIMES_KEY = "__dir_mtimes"` (single trailing underscore), `cursor.pop` to keep `cursor.values()` numeric, and no type annotation. I aligned this sensor to match — sidesteps the annotation issue entirely and keeps cursor-key naming consistent across the codebase.

On the specific annotation nit: the library signature in `resources.py:41` is `dict[str, float] | None`, and `int` is assignable to `float` per PEP 484, so the original wasn't wrong for mypy. But dropping the annotation in favor of inference (matching ADP) is cleaner than picking either `int` or `float | int` here while leaving the library signature inconsistent.

Acknowledged the other items:
- Stale cursor entries: agree, not worth fixing now.
- Pre-existing wall-clock cursor at line 111: will file separately.
- Post-deploy tick-duration monitoring: in the PR body.

cbini:
@claude update your review

PR #3951, opened 2026-05-18
fix(libraries/ssh): re-enable ssh-rsa host keys for edplan SFTP

cbini:
# Pull Request

## Summary & Motivation

Restore edplan SFTP ingestion for kippcamden and kippnewark. The host
`secureftp.easyiep.com` (GlobalSCAPE EFT 8.1) advertises only `ssh-rsa` as its
host-key algorithm, which paramiko 5.0 dropped from its default `_preferred_keys`.
Negotiation failed deterministically with `paramiko.ssh_exception.IncompatiblePeer:
no acceptable host key`, blocking three consecutive nights of
`<loc>__edplan__njsmart_powerschool` materializations and producing 100/100
failed ticks per location for the two edplan SFTP sensors.

This PR:

1. Adds an opt-in `enable_legacy_rsa: bool = False` field on `SSHResource`. When
   set, `get_connection()` appends `"ssh-rsa"` to `paramiko.Transport._preferred_keys`
   for the duration of the connect (restored in `finally`).
2. Enables the flag on `SSH_EDPLAN` only — every other SFTP resource keeps
   paramiko's modern defaults.
3. Narrows the tenacity retry predicate on `get_connection()` from
   `(SSHException, TimeoutError, socket.gaierror)` to
   `(NoValidConnectionsError, TimeoutError, socket.gaierror, ConnectionResetError)`.
   The old predicate caught deterministic config failures (`IncompatiblePeer`,
   `BadHostKeyException`, `BadAuthenticationType`) and burned ~30s × 5 retries
   on each one — confirmed in the run logs for the failed njsmart_powerschool
   runs (5 connect attempts in 33s before raising).
4. Hoists the duplicated branch-deployment environment-variable check (4 inline
   copies) to a module-level `IS_BRANCH_DEPLOYMENT` constant.
5. Documents the post-handshake `time.sleep(1.0)` in `open_ssh_tunnel()` as a
   race-condition guard for downstream forwarded-port consumers.

A separate backfill of `kippcamden__edplan__njsmart_powerschool` and
`kippnewark__edplan__njsmart_powerschool` for 2026-05-16, 05-17, 05-18 should
follow once this merges.

## AI Assistance

AI-assisted: full diagnosis (sensor / run / pod log correlation, paramiko
internals investigation, fix design and implementation) plus this PR body.
Human-directed: confirmed the failure mode with a local `ssh -vv` against the
host, vetoed the use of `getattr`/`setattr` over direct attribute access,
approved the `IS_BRANCH_DEPLOYMENT` refactor, and supplied the rationale for
the existing `time.sleep(1.0)`.

### General

### Dagster

      code location

## CI checks

- [ ] **Trunk** — passes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,7 +60,10 @@ file; domain specifics live in the nearest subdirectory CLAUDE.md.
   explicitly. Use `git -C <worktree>` on every git call (bare `git` from the
   main repo silently commits to `main`) and
   `uv run dbt ... --project-dir <worktree>/src/dbt/<project>` on every dbt call.
-  Otherwise prefer absolute paths.
+  For Python execution from the main repo, prefix `VIRTUAL_ENV=` and use
+  `uv --directory <worktree> run python ...` — bare `uv run --active` reads the
+  main repo's `.venv` and misses worktree-only changes. Otherwise prefer
+  absolute paths.
 
 - **Branch switch**: `gh issue develop <number> --name <branch> --checkout`.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -128,8 +128,12 @@ file; domain specifics live in the nearest subdirectory CLAUDE.md.
   `.trunk/tools/trunk check --force <files>` to verify before claiming the
   change is lint-clean.
 
-- **Linter**: Use `# trunk-ignore(<linter>/<rule>)` with a reason comment — not
-  linter-native disable syntax. Binary:
+- **Linter**: Suppress with `# trunk-ignore(linter/rule): reason` (e.g.
+  `# trunk-ignore(bandit/B603): static argv, no shell`) on the line immediately
+  before the flagged line — not linter-native disable syntax. Wrapping the
+  reason onto extra comment lines silently breaks the suppression (trunk only
+  honors the directive on the adjacent line), and CI also flags it with
+  `trunk/ignore-does-nothing`. Binary:
   `/workspaces/teamster/.trunk/tools/trunk`.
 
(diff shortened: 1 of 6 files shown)


cbini:
Addressed in c943656af.

**1. `trunk-ignore(bandit/B603)` removal** — verified: `trunk check --force --filter=bandit` fires B603 on the `subprocess.Popen` call. The pre-push diff-only check let the removal slip through (it flags new B603 lines, not removed suppressions on unchanged code). Comment restored.

**2. `Transport._preferred_keys` race** — your scenario is the worst case (ssh-rsa leaking permanently after a save/restore interleave), not the retry case I originally documented. Fixed with a module-level `threading.Lock` around the save/mutate/connect/restore critical section. Comment rewritten to match.

**3. `ConnectionResetError`** — precautionary, not seen in the edplan logs. Added as defense against generic mid-stream TCP RST since `NoValidConnectionsError` and `TimeoutError` don't cover that case. Leaving as-is; happy to drop if you'd prefer YAGNI.

**4. `IS_BRANCH_DEPLOYMENT` import-time evaluation** — `grep -r "DAGSTER_CLOUD_IS_BRANCH_DEPLOYMENT" tests/` returns zero matches. No test currently sets this flag, so the import-time evaluation is safe. If a test needs to flip it later, patching `teamster.core.resources.IS_BRANCH_DEPLOYMENT` directly is the path.

**5. `libraries/ssh/CLAUDE.md` table** — already added in d10f1f030 (the docs commit landed after your review of c4541aa1b). `enable_legacy_rsa` row + a Notes bullet documenting the paramiko 5.0 default change and the `ssh -vv` diagnostic.

PR #3962, opened 2026-05-19
fix(kipptaf): filter dibels bm goals calculations to At/Above goal type only

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will fix `rpt_gsheets__dibels_bm_goals_calculations` to correctly select grade goals by filtering to `grade_goal_type = 'At/Above'` in the final SELECT, rather than relying on `max(grade_goal)` aggregation across multiple goal-type rows.

**Root cause of the old bug**: The previous model grouped rows and used `max(grade_goal)` to collapse At/Above and Well Below goal rows. For upper grades (5–8) where the Well Below goal is numerically larger than the At/Above goal, `max()` silently picked the Well Below value. This caused incorrect expected counts and gap calculations for Camden grades 5–8, Newark grades 7–8, and Paterson grades 6–7.

**Fix**: Propagate `grade_goal_type` through the CTEs and filter `where grade_goal_type = 'At/Above'` in the final SELECT, dropping the `max()`/`avg()`/`sum()` aggregation layer entirely. The `* except (grade_goal_type)` removes the helper column from the output contract.

**Also fixes**: `n_admin_season_school_gl_all` and `n_admin_season_region_gl_all` were declared `float64` in the YAML (artifact of the old `avg()` aggregation). Now declared `int64`, matching the `count()` window function output.

## AI Assistance

Claude Code (claude-sonnet-4-6) identified the root cause (`max()` picking Well Below goal for upper grades), proposed and implemented the refactor, and updated the YAML properties file. Human-directed: branch creation, final SQL validation, gap-column impact analysis against prod data.

### General

- [x] Include a `[model_name].yml` properties file for all new models — updated `rpt_gsheets__dibels_bm_goals_calculations.yml`
- [ ] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application
- [ ] **Breaking change?** Output schema is unchanged (same columns, same names). `grade_goal` and `grade_range_goal` values will differ for upper-grade rows where the old model was using the Well Below goal value.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

### DIFF
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__dibels_bm_goals_calculations.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__dibels_bm_goals_calculations.yml
@@ -20,13 +20,13 @@ models:
       - name: grade_range_goal
         data_type: float64
       - name: n_admin_season_school_gl_all
-        data_type: float64
+        data_type: int64
       - name: n_admin_season_school_gl_at_above
         data_type: int64
       - name: n_admin_season_school_gl_bl_wb
         data_type: int64
       - name: n_admin_season_region_gl_all
-        data_type: float64
+        data_type: int64
       - name: n_admin_season_region_gl_at_above
         data_type: int64
       - name: n_admin_season_region_gl_bl_wb
--- a/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__dibels_bm_goals_calculations.sql
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__dibels_bm_goals_calculations.sql
@@ -156,13 +156,40 @@ with
     )
 
 select
-    * except (grade_goal_type),
+    c.academic_year,
+    c.region,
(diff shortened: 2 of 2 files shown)


GabyRangelB:
@claude **Issue 1 (bl_wb always zero) has been addressed** in the latest commit (`fa17b7fd6`).

The fix uses a self-join on `needed_count_calcs` to pull `n_admin_season_school_gl_bl_wb` and `n_admin_season_region_gl_bl_wb` from the `grade_goal_type = 'Well Below'` rows, which carry the correct Below/Well Below student counts. The At/Above rows (which are zero for bl_wb due to the partition scoping) no longer source those columns.

**Issue 2 (non-deterministic row_number)** was investigated against current data — zero NULL `grade_goal` values exist across 7,146 At/Above rows, so the foundation goals table fully covers every grade/region/period combination. Not fixing.

**Items 3 and above** (explicit column enumeration, uniqueness test, descriptions) will not be addressed in this PR in the interest of time.

anthonygwalters (approved):

Issue #3972, opened 2026-05-20
fix(amplify-mclass-sftp-sensor): drop unsound dir_mtimes prune


cbini:
Amplify mClass SFTP sensors (kippnewark, kipppaterson) have not spawned a run since 2026-05-14 ~08:40 UTC. Root cause: the `dir_mtimes` subtree prune added in 84cc25a56 + 729b06c3f assumes POSIX dir-mtime propagation that the Amplify SFTP server does not honor. Remove the `dir_mtimes` wiring from `build_amplify_mclass_sftp_sensor`; keep the parameter on `listdir_attr_r` (ADP WFN sensor uses it and is healthy).

PR #3975, opened 2026-05-20
test(sftp): audit dir-mtime propagation across SFTP servers

### LINKED ISSUE #3974: test(sftp): audit dir-mtime propagation across all SFTP servers
Add a passive integration test that walks each SFTP server we connect to and flags any directory whose `st_mtime` is older than the max `st_mtime` of its children — the smoking-gun signature of a server that doesn't propagate parent-dir mtime on entry changes (the failure mode behind #3972). Test lives under `tests/resources/`, uses the existing conftest 1Password bootstrap, and skips per-server when required environment variables are missing.

cbini:
## Summary & Motivation

> When merged, this pull request will add an opt-in integration test that detects SFTP servers where `listdir_attr_r(dir_mtimes=)` would silently drop data — the failure mode behind #3972.

The test walks each SFTP server's directory tree (bounded to 200 directories / depth 4) and fails when any directory's `st_mtime` is older than the max `st_mtime` of its children. That mismatch is the smoking-gun signature of a server that doesn't propagate parent-dir mtime on entry changes; servers exhibiting it cannot safely opt into the `dir_mtimes=` subtree-prune optimization.

Lives at `tests/resources/test_resource_ssh_dir_mtime_propagation.py` with a sibling `conftest.py` that registers one parametrized fixture per SFTP `SSHResource`. Authentication piggybacks on the root `tests/conftest.py` 1Password bootstrap — no new credential plumbing. Per-server skip when the relevant environment variables are unset, so the test exits cleanly in CI without 1Password access.

## Initial audit findings

Run via `uv run pytest tests/resources/test_resource_ssh_dir_mtime_propagation.py` in a codespace with 1Password bootstrap:

| Result | Servers |
|---|---|
| PASSED (3) | `littlesis`, `renlearn`, `titan` |
| SKIPPED (1) | `lattice` (environment variables not bootstrapped — separate config issue) |
| FAILED (11) | `amplify`, `adp_workforce_now`, `clever`, `couchdrop`, `coupa`, `deanslist`, `edplan`, `egencia`, `idauto`, `illuminate`, `iready` |

Only **two** sensors have ever wired `dir_mtimes=` into `listdir_attr_r`:

- `libraries/amplify/mclass/sftp/sensors.py` — already removed in #3973
- `code_locations/kipptaf/adp/workforce_now/sftp/sensors.py` — **still using it in production**

The ADP WFN sensor produces SUCCESS ticks today because its file-level `min_mtime` filter and the current vendor upload pattern happen to align with subtrees that do propagate. But the audit shows ADP WFN's server has the same structural fragility as Amplify did before 5/14 — a layout shift on the vendor side could silently break ingestion the same way. **Removing `dir_mtimes=` from the ADP WFN sensor will be filed as a separate PR.**

The other 9 failing servers don't currently opt into the optimization, so their violations are latent. The test's value for those servers is forward-looking: it prevents future opt-ins on servers that would silently drop data.

## Why a passive walk rather than an active probe?

An active probe (upload-check-delete) would give a definitive answer but writes to production SFTP servers — clutter risk, vendor confusion, and requires a known-safe write path on each. The passive walk is fully read-only. A smoking-gun failure is definitive proof of non-propagation; absence of a smoking gun is suggestive but not proof. That trade-off is the right one for an opt-in safety check.

## CI behavior

The test only runs when explicitly invoked (`uv run pytest tests/resources/...`). Default CI doesn't trigger it. It's intended as a developer tool — run before opting a sensor into `dir_mtimes=` to verify the server is safe.

Closes #3974

## AI Assistance

AI-assisted: test design (passive vs. active probe trade-off, bounded walk parameters), conftest and test code, audit execution and result interpretation, scope decisions (separating ADP WFN removal into its own PR). Human-directed: choice of probe method (passive), server scope (all SFTP resources), placement (`tests/resources/`, opt-in), and PR sequence.

### Dagster

- [x] Test collects cleanly (`uv run pytest --collect-only`): 15 tests parametrized
- [x] Live audit run completed (3 passed / 1 skipped / 11 failed) — findings above

### Docs

- [x] No new schedule or sensor — no automations catalog update needed

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — N/A (no dbt changes)
- [ ] **Dagster Cloud** — N/A (test-only, no asset changes)

### DIFF
--- a/tests/resources/conftest.py
+++ b/tests/resources/conftest.py
@@ -8,6 +8,7 @@ unset (e.g. CI without 1Password access).
 
 from collections.abc import Callable
 from dataclasses import dataclass
+from typing import Any
 
 import pytest
 from dagster import EnvVar
@@ -39,7 +40,7 @@ def _password_ssh(
         resolved_port = port
         if port_env is not None:
             resolved_port = int(check.not_none(value=EnvVar(port_env).get_value()))
-        kwargs: dict = {
+        kwargs: dict[str, Any] = {
             "remote_host": EnvVar(host_env),
             "remote_port": resolved_port,
             "username": EnvVar(user_env),
--- a/tests/resources/conftest.py
+++ b/tests/resources/conftest.py
@@ -1,0 +1,259 @@
+"""Per-server SSHResource fixtures for SFTP integration tests.
+
+Authentication is handled by the root ``tests/conftest.py`` session-scoped
+fixture that bootstraps 1Password secrets into ``os.environ`` on first run.
+Individual fixtures here skip when their required environment variables are
+unset (e.g. CI without 1Password access).
+"""
+
+from collections.abc import Callable
+from dataclasses import dataclass
+
+import pytest
+from dagster import EnvVar
+from dagster_shared import check
+
+from teamster.libraries.ssh.resources import SSHResource
+
+
+@dataclass(frozen=True)
+class SFTPServer:
+    name: str
+    env_vars: tuple[str, ...]
+    builder: Callable[[], SSHResource]
+    remote_dir: str
+    exclude_dirs: tuple[str, ...] = ()
+
+
+def _password_ssh(
+    *,
+    host_env: str,
+    user_env: str,
+    pass_env: str,
+    port: int = 22,
+    port_env: str | None = None,
+    timeout: int | None = None,
+    enable_legacy_rsa: bool = False,
+) -> Callable[[], SSHResource]:
+    def build() -> SSHResource:
+        resolved_port = port
+        if port_env is not None:
+            resolved_port = int(check.not_none(value=EnvVar(port_env).get_value()))
+        kwargs: dict = {
+            "remote_host": EnvVar(host_env),
+            "remote_port": resolved_port,
+            "username": EnvVar(user_env),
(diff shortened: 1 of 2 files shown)


cbini:
Fixed in 1369cf345:

- Annotated `_walk(sftp: SFTPClient)` with the concrete paramiko type.
- Switched `kwargs: dict` to `dict[str, Any]` in `_password_ssh`.
- Added a docstring on `_walk` explaining why the root directory itself isn't audited (no parent to stat it from via `listdir_attr`).

On the `sftp_server: SFTPServer` parameter annotation — pushing back. I tried that import path earlier and pytest collection failed with `ModuleNotFoundError: No module named 'tests'`. Pytest auto-loads `conftest.py` for fixture discovery but does not add `tests.resources` to `sys.path` as an importable module (would need `tests/__init__.py` files, which the repo deliberately avoids). The annotation could go behind `TYPE_CHECKING` with a string forward-reference, but the value at a test-function parameter is minimal — pytest doesn't introspect annotations on fixture-injected params. Leaving it untyped.

The `exclude_dirs` path-matching observation is noted as a known fragility; not changing for now.

Issue #3820, opened 2026-05-04
refactor(dbt): mart surrogate keys hash region prefix only, not full _dbt_source_relation


cbini:
## Context

`dim_courses.sql` and `dim_student_section_enrollments.sql` both use a `replace()` workaround on `_dbt_source_relation` inside `generate_surrogate_key()` to align producer / consumer hashes when the producer (e.g. `stg_powerschool__courses`) and consumer (e.g. `dim_course_sections` via `base_powerschool__sections`) read from differently-named upstream models in the same region.

```sql
{{
    dbt_utils.generate_surrogate_key(
        [
            "c.course_number",
            (
                "replace(c._dbt_source_relation,"
                " 'stg_powerschool__courses',"
                " 'base_powerschool__sections')"
            ),
        ]
    )
}} as course_key,
```

This pattern is fragile: a future district whose source model is named differently silently breaks the FK chain (the `replace()` no-ops, the hash diverges, FK resolves to nothing).

## Long-term fix

Hash on the region prefix extracted from `_dbt_source_relation` (e.g. `regexp_extract(_dbt_source_relation, r'(kipp\w+)_')`) rather than the full table-qualified relation. The region prefix is consistent across producer and consumer regardless of which upstream model they read.

Apply consistently to every mart surrogate key that includes `_dbt_source_relation` as a hash input.

## Affected files (initial sweep)

- `src/dbt/kipptaf/models/marts/dimensions/dim_courses.sql`
- `src/dbt/kipptaf/models/marts/dimensions/dim_course_sections.sql`
- `src/dbt/kipptaf/models/marts/dimensions/dim_student_section_enrollments.sql`
- (others — full sweep needed)

## Acceptance

- [ ] Audit every mart `generate_surrogate_key()` call that includes `_dbt_source_relation`.
- [ ] Replace the full-relation hash with a region-prefix hash, applied identically across producer and consumer.
- [ ] Remove the `replace()` workarounds from `dim_courses` and `dim_student_section_enrollments`.
- [ ] All cross-mart FK relationships tests stay at 0 orphans.

## Related

- PR #3809 — introduced the `replace()` pattern in `dim_courses.sql` (rerouted source from `base_powerschool__course_enrollments` to `stg_powerschool__courses`).
- `dim_student_section_enrollments.sql:30-46` — original site of the pattern.

cbini:
### Status update — partial migration landed on PR #3985

[...]

PR #4004, opened 2026-05-22
refactor(dagster): drop dep-code-version gate

### LINKED ISSUE #4003: fix(dagster): scope dep-code-version gate to same code location
## Symptom

305 `kipptaf__dbt_assets` runs failed 2026-05-21 23:52 ET through 2026-05-22 11:13 ET, all with `Database Error in model dim_courses ... Name _dbt_source_project not found inside c`. `kipptaf/powerschool/stg_powerschool__courses` (the parent that c32f3a8d6 added the column to) was not being auto-materialized to pick up the new column. dim_courses (a plain view, no gate) kept retrying on every tick.

## Root cause

The `_dep_code_version_pending` gate in `_build_dbt_condition()` (commit 615a21610, 2026-04-09) blocks the kipptaf staging asset from being requested. The Dagster condition tree shows:

- `kippmiami/powerschool/stg_powerschool__courses ((code_version_changed) SINCE (newly_updated))` = `[1/1]`
- → `any_deps_match(...)` = TRUE → outer `NOT(...)` = FALSE → kipptaf-level staging never requested

`kippmiami/powerschool/stg_powerschool__courses` last materialized 2025-08-22; its current code_version differs from stored. The SINCE memory correctly identifies this as a dep with unapplied code change.

The gate's SINCE semantic is correct (verified against `CodeVersionChangedCondition.evaluate()` — it's a transition detector that needs SINCE memory to express "code changed since last mat"). The deadlock isn't a logic bug — it's that **cross-code-location deps reach kipptaf via dbt `source()` rather than `ref()`, and the kippmiami sensor isn't reaching its staging asset to clear the gate.**

## Fix

Scope the gate to deps with the same code-location key prefix. Within a code location, dbt's internal DAG ordering already handles correctness within a single `dbt build` invocation (the gate is belt-and-suspenders). Across code locations, the gate currently causes deadlock when an upstream sensor doesn't reach its asset.

Implementation:

- `_build_dbt_condition` takes new `guard_dep_selection: AssetSelection | None = None`. When set, applies `.allow(selection)` to the gate's `any_deps_match(...)`.
- `dbt_table_automation_condition` and `dbt_union_relations_automation_condition` take new `code_location: str | None = None`, passed through to `_same_code_location_selection()` (`AssetSelection.key_prefixes(code_location)`).
- `CustomDagsterDbtTranslator.get_automation_condition` passes `self.code_location` to both.

Same-CL race protection is preserved. Cross-CL race becomes possible but is mitigated by (a) dbt validates view SQL at read time anyway, (b) cross-CL deploys are already uncoordinated, (c) manual `launch_run` is the documented escape valve for any cross-CL coordination issue.

## Tests

- `test_table_gate_scoped_to_same_code_location` — verifies cross-CL deps with unapplied code changes don't block downstream.
- `test_table_gate_still_blocks_same_code_location_deps` — verifies same-CL gate still fires.

## Investigation notes

- First fix attempt (drop `.since(newly_updated())` from the gate) verified unsafe via Dagster source inspection — would lose race protection between transition ticks. Reverted.
- District staging assets (`kippmiami/powerschool/stg_powerschool__courses` etc.) have zero condition evaluation records, suggesting the district auto-mat sensor isn't covering them. Worth investigating separately, but addressing that root cause is out of scope here.
- Immediate unblock for the active 305-run incident: manual `launch_run` on `kipptaf/powerschool/stg_powerschool__courses`. This fix prevents recurrence.

cbini:
## Summary & Motivation

When merged, this PR removes the `_dep_code_version_pending` gate from `_build_dbt_condition` entirely (tables, union_relations views, and the now-unused parameter). The translator no longer plumbs `code_location` through to condition factories.

### Background

Discovered while diagnosing a 305-run kipptaf__dbt_assets failure cascade on 2026-05-21/22. PR #3820 added a `_dbt_source_project` column to `kipptaf/powerschool/stg_powerschool__courses` and to `dim_courses`'s hash inputs. Every downstream `dim_courses` view re-creation failed with `Name _dbt_source_project not found inside c` because the kipptaf-level staging table was never rematerialized to pick up the new column. The Dagster automation condition evaluator showed the gate firing on `kippmiami/powerschool/stg_powerschool__courses ((code_version_changed) SINCE (newly_updated)) = [1/1]`.

### What we learned

Confirmed via `mcp__dagster__get_asset_staleness`: all four regional `stg_powerschool__courses` assets (kippmiami, kippnewark, kippcamden, kipppaterson) are **FRESH** — current code_version matches stored mat code_version. There is no actual divergence anywhere in kipptaf's parents.

Yet `code_version_changed().since(newly_updated())` evaluates `[1/1]` for kippmiami's row in kipptaf's gate evaluator. The reason: `code_version_changed()` is a **cursor-based transition detector** (compares to the evaluator's prior tick, not the dep's stored materialization tag). The SINCE memory wraps that into "transition observed at some point since dep's last mat" — but the cursor can be perturbed by sensor restarts, condition tree changes, or manifest re-parses, capturing phantom "true" state that never resets on a FRESH dep (no `newly_updated` event will fire to clear it).

### Why drop the gate instead of scoping it

An earlier version of this PR (now superseded) scoped the gate to same-code-location deps via `AssetSelection.key_prefixes(code_location)`. That worked for today's incident but post-hoc rationalized the mechanism — phantom SINCE can also strike same-CL parents, just less often.

The race the gate nominally prevents:

- **Within a code location**: a deploy adds a column to a parent and a downstream that references it. Both have `code_version_changed=TRUE` in the same auto-mat tick, both are requested into the same `<code_location>__dbt_assets` run, and dbt's intra-build DAG ordering materializes parent first. No race. (Also: `~any_deps_missing()` and `~any_deps_in_progress()` already cover the parent-not-ready cases.)
- **Cross code location**: parent and downstream are in different dbt builds — a race is possible. The failure mode is the downstream task fails with a BigQuery schema error and is retried (manual or auto-retry). Not data corruption, not silent wrongness. The escape valve (manual `launch_run`) already exists.

Net assessment: the gate's belt-and-suspenders within-CL protection is moot, and its cross-CL protection trades a recoverable retry-once race for permanent false-positive deadlocks. The cure is worse than the disease.

### Root-cause fix not in scope

A true root-cause fix exists: a custom `BuiltinAutomationCondition` that compares current code_version against the dep's stored mat code_version directly (no cursor, no SINCE memory). The owner of this branch chose not to build that custom operand — drop the gate and rely on dbt's intra-build DAG ordering instead.

### Diff

- `_build_dbt_condition` loses `guard_dep_code_version` and `guard_dep_selection` parameters; the entire `if guard_dep_code_version:` block is removed.
- `dbt_view_automation_condition`, `dbt_union_relations_automation_condition`, and `dbt_table_automation_condition` simplified to call `_build_dbt_condition(...)` with their trigger trees only.
- Translator passes no extra args.
- New `test_table_not_blocked_by_parent_pending_code_version` mirrors the existing view test for tables.
- The two interim "scoped gate" tests are removed.
- CLAUDE.md in `src/teamster/core/` rewritten to document the gate's absence and what to investigate when a downstream looks stuck.

Closes #4003

## AI Assistance

Claude Code (Opus 4.7, 1M ctx) drove the systematic-debugging investigation across multiple iterations: enumerating gate semantics, reading Dagster's `CodeVersionChangedCondition.evaluate()` source to disprove a "drop SINCE only" fix that would have lost race protection between transition ticks, then designing the same-CL-scoped gate, then disproving that framing using `get_asset_staleness` data showing parents are FRESH, and finally landing on the gate-removal change. The conditional logic and tests were AI-authored; each fix attempt and disproof was discussed with the human reviewer before committing.

### General

### Dagster

- [x] `tests/test_automation_conditions.py` — `test_table_not_blocked_by_parent_pending_code_version` covers the new behavior; existing `test_view_not_blocked_by_parent_pending_code_version` continues to pass.
- [x] All in-scope pytest tests pass locally (20 passed, 16 deselected for pre-existing failures or environment-variable-dependent integration tests unrelated to this change).
- [ ] Run `uv run dagster definitions validate` — requires secrets unavailable to the Claude session; reviewer to confirm.
- [ ] Run `uv run pytest tests/test_dagster_definitions.py` — same as above.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — N/A (no dbt changes).

### DIFF
--- a/src/teamster/core/CLAUDE.md
+++ b/src/teamster/core/CLAUDE.md
@@ -109,16 +109,18 @@ deploy rollover, the materialization may be stamped with the new deployment's
 code version. `code_version_changed()` returns false permanently — manual
 materialization is the only fix. See dagster-io/dagster#33708.
 
-**Deploy ordering gate**: `_dep_code_version_pending` in
-`_build_dbt_condition()` blocks materialization when a direct dependency has
-`code_version_changed().since(newly_updated())`. Applied to tables and
-union_relations views via `guard_dep_code_version=True`; plain views opt out
-(`guard_dep_code_version=False`). **The gate is scoped per code location** via
-`guard_dep_selection=AssetSelection.key_prefixes(code_location)` —
-cross-code-location deps (kipptaf reading kippmiami via dbt `source()`) bypass
-the gate. When debugging "downstream table not auto-materializing despite
-code_version_changed," check whether a same-CL parent has stuck SINCE memory;
-cross-CL parents are excluded by design.
+**No dep-code-version gate**: `_build_dbt_condition()` does NOT block
+materialization when a direct dep has
+`code_version_changed().since(newly_updated())`. A previous gate did, but the
+operator is cursor-based — its SINCE memory could capture phantom "true" state
+from any past tick (sensor restart, condition change, manifest re-parse) and
+never reset on a FRESH dep (no `newly_updated` event to clear it), producing
+permanent deadlocks. The in-CL race the gate nominally prevented is already
+covered by dbt's intra-build DAG ordering plus `any_deps_missing` /
+`any_deps_in_progress`; cross-CL races fail at BigQuery query time (recoverable,
+not silent corruption). If a downstream table looks "stuck," the cause is
+elsewhere — start with `any_deps_missing` / `any_deps_in_progress` evaluator
+nodes.
 
 **Dep fan-out rule**: An unpartitioned dep of a partitioned asset fans out to
 ALL partitions on every materialization. To preserve per-partition triggering,
--- a/src/teamster/core/CLAUDE.md
+++ b/src/teamster/core/CLAUDE.md
@@ -109,15 +109,16 @@ deploy rollover, the materialization may be stamped with the new deployment's
 code version. `code_version_changed()` returns false permanently — manual
 materialization is the only fix. See dagster-io/dagster#33708.
 
-**Deploy ordering gate**: `_dep_code_version_pending` blocks materialization
-when a direct dependency has `code_version_changed().since(newly_updated())` —
-prevents schema errors when a deploy adds columns through a TABLE → VIEW chain.
-Applied in `_build_dbt_condition()` to tables and union_relations views via the
-default `guard_dep_code_version=True`. Plain views opt out
-(`dbt_view_automation_condition` passes `False`): a view never bakes parent
-columns into stored state, so a pending parent code change can't poison it, and
-the guard would only strand the view's own `code_version_changed` refresh behind
-unrelated upstream churn.
+**Deploy ordering gate**: `_dep_code_version_pending` in
+`_build_dbt_condition()` blocks materialization when a direct dependency has
+`code_version_changed().since(newly_updated())`. Applied to tables and
+union_relations views via `guard_dep_code_version=True`; plain views opt out
+(`guard_dep_code_version=False`). **The gate is scoped per code location** via
+`guard_dep_selection=AssetSelection.key_prefixes(code_location)` —
+cross-code-location deps (kipptaf reading kippmiami via dbt `source()`) bypass
+the gate. When debugging "downstream table not auto-materializing despite
+code_version_changed," check whether a same-CL parent has stuck SINCE memory;
+cross-CL parents are excluded by design.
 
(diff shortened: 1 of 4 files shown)


cbini:
@claude I pivoted on the solution, update your review

PR #4010, opened 2026-05-22
feat(cube/dbt): snapshot anchor auto-injection and named period-end measures

### LINKED ISSUE #4009: feat(cube/mcp): add AY normalization and CA anchor validation to server.py
## Problem

Two recurring query mistakes cause silent overcounting in the Cube MCP:

1. **Academic year phrasing** — users and LLMs send `"AY2025"`, `"ay2025"`, `"2025-26"`, or `"2025–2026"` as filter values on `dim_terms_academic_year`. Cube expects `"2025"` (the start year of the 2025–26 school year). These pass through silently and return 0 rows.

2. **Chronic absence without an anchor filter** — CA measures (`pct_chronically_absent`, `count_chronically_absent`, `pct_tier_1_2`, `pct_tier_3`) are cumulative per student per day. A query with no date pin or `is_latest_record = true` filter counts each student once per matching day — e.g. a student CA on 30 days is counted 30 times. The result looks plausible but is meaningless.

## Solution

Add two validation helpers and wire them into both `load` and `sql`:

- **`_rewrite_ay_filters`** — silently normalises AY filter values before the query reaches Cube (`"AY2025"` / `"2025-26"` → `"2025"`). Fires on every call; no error on success.
- **`_validate_query`** — checks CA queries for a valid anchor filter (single `date_day` equality pin OR `is_latest_record = true`). On failure, returns a structured error dict (not a raised exception) with a plain-English explanation, two fix options, and a `suggested_fix` field with the original query patched to add an Option A stub. The LLM reads the error, fills in the real date, and retries.

## Files changed

- `src/cube/mcp/server.py`

cristinabaldor:
Closes #4009

## Summary & Motivation

When merged, this pull request will establish a reusable pattern for Cube cubes built on fact tables with cumulative daily-status flags, and apply it to attendance as the first domain.

**Background:** Some fact tables have one row per entity per day where key columns are cumulative running statuses — re-stamped on every row with the entity's status as of that day. Without a point-in-time anchor, `count_distinct` on any such flag across a date range overcounts (a student flagged on 30 days is counted 30 times). This PR introduces a general mechanism to prevent that class of error.

**`cube.js` — snapshot anchor auto-injection**

Adds a `queryRewrite` guard for cubes registered in `SNAPSHOT_CUBES`. For any unanchored query against a snapshot measure, injects the appropriate anchor automatically:
- No granularity or year-level → `is_latest_record = true`
- `granularity: "month"` → `is_month_end_record = true`
- `granularity: "week"` → `is_week_end_record = true`

Named `_year_end` / `_month_end` / `_week_end` measures bypass injection — anchors are baked into their SQL filters. Throws a descriptive error for unsupported granularities or for named period-end measures used without matching granularity. To add a new domain: append the cube name to `SNAPSHOT_CUBES` and its measure stems to `SNAPSHOT_MEASURE_STEMS`.

**`attendance.yml` + views** — first application of the pattern. Adds `is_month_end_record` and `is_week_end_record` dimensions and named measures in three families (`_year_end`, `_month_end`, `_week_end`) for CA counts/pcts, ADA tier pcts, and truancy, exposed on both `attendance_detail` and `attendance_summary`.

**`fct_student_attendance_daily.sql`** — adds `is_month_end_record` and `is_week_end_record` columns: last full membership day per student per calendar month/week.

**`server.py`** — removes now-redundant CA anchor guidance from the MCP system prompt; the `queryRewrite` guard handles it automatically.

## AI Assistance

Architecture and implementation were Claude-assisted (snapshot anchor pattern, named measure family design, dbt window function logic). Human-directed: scope decisions, measure naming conventions, which stems belong in `SNAPSHOT_MEASURE_STEMS`.

### General

### dbt

- [x] Properties files updated for all new columns (`fct_student_attendance_daily.yml`)
- [x] No new exposures needed — `fct_student_attendance_daily` is an in-place mart update

## CI checks

- [x] **Trunk** — passes
- [ ] **dbt Cloud** — passes
- [ ] **Dagster Cloud** — not triggered (no Dagster changes)


cristinabaldor:

cristinabaldor:
**Additional fix: `"2025-2025"` same-year range now returns a structured error**

`_expand_ay_range` previously returned `None` for `"2025-2025"` (start ≥ end guard), letting it fall through to Cube as a raw string. Cube would reject it with a generic float cast error. Changed `start >= end` → `start > end` so the value expands to `["2025"]` and gets caught by the existing multi-year range error path with a correct `suggested_fix`. One test added.

---

**Local end-to-end testing against staging**

Wired the local `server.py` (this branch) to the Cube Cloud staging endpoint via stdio MCP and ran two smoke tests directly in a Claude Code session:

- **CA anchor validation**: sent `pct_chronically_absent` with no anchor filter → server returned `__query_validation_error__: true` with the expected error message and a `suggested_fix` stub.
- **AY normalisation**: sent a filter with `"AY2025"` → compiled SQL showed parameter `"2025"`, confirming the rewrite happened before the query reached Cube.

cbini (changes requested):
## Recommended changes

1. **Move the CA anchor guard to Cube `queryRewrite`** (inject `attendance.is_latest_record`, skip when a date is pinned or grouped) — not the MCP server. Covers all consumers; fixes the trend false-positive.
2. **Add a baked `count_chronically_absent_year_end` measure** as a self-documenting complement for the YoY case.
3. **Drop the AY normalization** — `type: number` already rejects non-numeric values deterministically.
4. **Fix the AY off-by-one via a model change** — centralize academic-year onto `dim_dates` (add `academic_year_label`, present AY via `dim_dates`, retire the `dim_terms` exposure).

_(Logistics: this can all stay in this PR; validate the model changes on Cube branch staging before merge.)_

---

I dug into this against the live (pre-merge) Cube MCP, the compiled SQL, and the warehouse. Summary: **keep the CA guard's intent, move it into Cube; replace the AY normalization with a model change.**

### CA overcounting — confirmed, and material

Probed production, academic year 2024 (`count_chronically_absent`):

| Query | Result |
|---|---|
| Anchored (`is_latest_record = true`) | **2,863** |
| No anchor | **5,529** |

Unanchored inflates ~1.93x, no error. Right thing to prevent.

**The mechanism in the issue is wrong, though, and it changes the fix.** Compiled SQL shows a `count(distinct student_enrollment_key)` over `fct_student_attendance_daily` (one row per student per day):

```sql
count(distinct CASE WHEN attendance.is_chronically_absent = true AND ...
  THEN attendance.student_enrollment_key END)
FROM kipptaf_marts.fct_student_attendance_daily AS attendance
```

A student is **counted once**, not "30 times." `is_chronically_absent` is a *daily cumulative* flag, and cumulative ADA recovers (dip below 90% in October, climb back by June). So unanchored = "ever CA on any day" (5,529); anchored = "CA at the snapshot" (2,863) — two valid distinct counts answering different questions. The defect is **grain ambiguity**, not row multiplication.

### This belongs in `queryRewrite`, not the MCP server

The invariant is needed by every consumer — Superset, SQL API, Playground — not just the MCP path. `queryRewrite` runs server-side on every request and already does this shape of work for the `dim_locations` scope filter and the staff `reporting_chain` segment. Cube docs confirm it can push filters or throw.

I verified the injected member by compiling both views: one cube-level member, **`attendance.is_latest_record`**, resolves through both `attendance_summary` and `attendance_detail`, producing the identical `WHERE attendance.is_latest_record = CAST(? AS BOOL)` predicate that yields 2,863. Rule: *if a CA measure is present, and there's no `is_latest_record` filter, no single `date_day` equality pin, and no `date_day` in dimensions/timeDimensions, inject `attendance.is_latest_record = true`.* ~10 lines, mirrors existing patterns, covers all consumers.

**Model-level complement (optional):** a clearly-named `count_chronically_absent_year_end` measure with `filters: [{ sql: "{CUBE}.is_latest_record = true" }]` baked in makes the dominant YoY case correct-by-construction and self-documenting, shrinking what `queryRewrite` has to guard down to just the explicit point-in-time measure. It can't stand alone, though — point-in-time needs a user-supplied date, and "default to latest unless a date is pinned or we're trending by `date_day`" is query-shape logic a context-free measure can't express. `queryRewrite` stays the load-bearing fix.

### Concrete false-positive in the current PR

`_validate_query` inspects only `filters`. A legitimate CA *trend* — `measures: [pct_chronically_absent]`, `dimensions: [attendance_summary.dim_dates_date_day]`, no filters — has no pin and no `is_latest_record`, so it hits the no-anchor error branch and gets **rejected**, even though grouping CA by date is a correct "accumulation over the year" query. `queryRewrite` sees `dimensions`/`timeDimensions`, so it can skip injection there.

### AY normalization — minor convenience, mis-scoped, and pointed away from the real risk

**1. The non-numeric case is already handled deterministically — by the type system.** `academic_year` is `type: number`, so `"AY2024"` / `"2024-25"` are rejected by Cube's coercion with an immediate 400 before any rows are scanned, for every consumer:

```text
values: ["AY2024"]  ->  400 {"error": "Bad double value: AY2024"}
```

The normalization replaces that deterministic *reject* with a partial *heuristic rewrite* — it fixes the formats someone regexed and still 400s on the rest, so behavior gets inconsistent rather than safer. A guard that guesses is the opposite of a type cast that rejects. Net benefit is "skip one recoverable round-trip," which doesn't justify the em-dash / 3-digit / `"2025-2025"` breadth.

**2. Wrong layer.** If first-try tolerance for `"2025-26"` is worth having, Superset and SQL-API users want it too — a one-line normalize in `queryRewrite` covers everyone.

**3. It can't touch the only AY mistake that's actually dangerous** — the off-by-one, below.

### The off-by-one is the real risk, and the fix is to centralize the academic-year surface

The silent failure is the start-year convention: filtering `academic_year = 2025` to mean the year that *ended* in 2025 (2024-25), when the convention is 2025 = the 2025-26 year. `2025` is a valid number, so it returns plausible data for the wrong year, no error. Normalization can't catch it — you can't infer intent from a valid integer.

This is a modeling problem, not an MCP one. `academic_year` is defined on **three** cubes (`dim_terms`, `dim_dates`, `dim_student_enrollments`), and the attendance views source it from `dim_terms` even though `dim_dates` — already joined — is the richer surface, carrying `academic_year` **and** `is_current_academic_year`. The convention already ships in the member description that reaches `/meta`, and off-by-one is still possible — so reminding the LLM has a proven ceiling. The durable fix is to stop making it apply the convention at all.

Recommendation — **centralize onto `dates`, no new dim:**

1. Add `academic_year_label` ("2024-25") to `dates.yml`, beside the `is_current_academic_year` it already has.
2. Present academic year through `dim_dates` in the attendance views; retire the parallel `dim_terms` AY exposure from the analyst surface (keep the raw column for internal joins). This also closes a latent diamond — AY is reachable via both `dim_dates` and `dim_terms` today.
3. The LLM then filters on unambiguous surfaces — `is_current_academic_year` for "this year," the string label for "2024-25" — and rarely needs the raw integer or the convention.

Verified safe to switch: across 100% of attendance-fact rows (every academic year 2007–2025), `dim_dates.academic_year` and `dim_terms.academic_year` are identical, so repointing the source changes no values. (One residual: confirm unmatched-key NULL behavior is unchanged when switching the join path.)

### Minor: error-return pattern

If any validation stays server-side, the official MCP SDK gives two first-class channels — raise (surfaces as `CallToolResult.isError`) or a typed structured-output model. The `__query_validation_error__` sentinel-in-a-success-dict is a bespoke third path; a small Pydantic result model would be more idiomatic. (Mostly moot if CA moves to Cube.)

Happy to pair on the `queryRewrite` diff and the `dates` change.

cristinabaldor:
All four items addressed, plus some additional scope that fell out of the analysis.

## Item 1 — CA anchor guard → `queryRewrite`

Moved to `cube.js` `queryRewrite` so all consumers (Superset, SQL API, Playground) are covered, not just the MCP path. Implementation mirrors the existing `dim_locations` scope filter and `reporting_chain` segment patterns.

Generalized as a `SNAPSHOT_CUBES` array — adding a future snapshot cube is a one-liner. Shared constants (`SNAPSHOT_ANCHOR_DIMENSIONS`, `SNAPSHOT_SELF_ANCHORED_SUFFIXES`) capture the convention without repeating per-cube entry.

**False positive fixed:** the original `_validate_query` only inspected `filters`, so a CA trend query grouped by `date_day` in `dimensions` or `timeDimensions` was incorrectly rejected. `queryRewrite` sees the full query shape and skips injection for day-level trends.

The guard also routes week/month granularities to `is_month_end_record` / `is_week_end_record` (see item 2), and throws a clear error for unsupported granularities (quarter/year) rather than returning silently wrong data.

## Item 2 — Named `_year_end` measure (and more)

Added `_year_end`, `_month_end`, and `_week_end` variants for all four CA measures and both truancy measures. `is_truant` is a daily cumulative flag with the same grain problem as `is_chronically_absent`, so it gets the same treatment.

This required two new dbt columns on `fct_student_attendance_daily`: `is_month_end_record` and `is_week_end_record` — last full membership day per month/week per enrollment. Non-membership days are excluded. Verified via BigQuery: ~9,600 enrollment-months where a naive window would have flagged the wrong day; the membership-aware window handles these correctly.

Each named measure has a private denominator anchored to the same period so the pct measures are correct-by-construction, not just the counts.

One additional guard in `queryRewrite`: `_month_end` measures require `timeDimensions` with `granularity: "month"`; `_week_end` requires `"week"`. Without this, an unanchored `count_chronically_absent_month_end` over a date range returns "CA at any month-end during the range" — a meaningful-looking but wrong number.

`_month_end` and `_week_end` measures are blocked on this PR's dbt changes reaching production and Dagster rebuilding the affected tables. `_year_end` measures are live on the branch now.

## Item 3 — Drop AY normalization

Removed `_rewrite_ay_filters`, `_normalise_ay_value`, and the associated regexes. `type: number` already rejects non-numeric values deterministically; the normalization was replacing a clean 400 with an inconsistent heuristic.

Kept the multi-year range check ("2023-2025" suggests `["2023", "2024", "2025"]` with `suggested_fix`). The raw Cube error for that case is "Bad double value: 2023-2025" with no actionable guidance; the check is narrow enough to earn its place.

## Item 4 — AY off-by-one via model change

Added `academic_year_label` (e.g. "2025-26") to `dim_dates` in both dbt and Cube. Retired `dim_terms_academic_year` from both attendance views; AY is now exposed through `dim_dates` (`dim_dates_academic_year`, `dim_dates_academic_year_label`, `dim_dates_is_current_academic_year`). Closes the diamond — AY was reachable via both `dim_dates` and `dim_terms` before.

Verified: across all attendance-fact rows, `dim_dates.academic_year` and `dim_terms.academic_year` are identical. `academic_year_label` is also blocked on dbt reaching production.

## Minor — error-return pattern

`__query_validation_error__` is now used only for the multi-year range check. Everything else is in `queryRewrite` where Cube's own error channels handle it. Left as-is since changing it would only affect that one remaining server-side check.

## Verified on staging

Against the branch staging endpoint with production BigQuery data:

| Test | Result |
|---|---|
| `pct_chronically_absent_year_end` YoY (no anchor filter) | AY2022: 47.5%, AY2023: 38.1%, AY2024: 27.5%, AY2025: 21.7% |
| Unanchored `count_chronically_absent` AY2024 (guard injection) | 2,863 |
| `academic_year_label`, `_month_end`, `_week_end` | Pending dbt to prod + Dagster rebuild |

The 2,863 for AY2024 matches your analysis exactly.

cristinabaldor:
@claude update your review

src/dbt/kipptaf/models/marts/dimensions/dim_dates.sql
```diff
@@ -42,6 +42,22 @@ select
...
+            )
+        }},
+        mod(
+            {{
+                date_to_fiscal_year(
+                    date_field="date_value", start_month=7, year_source="start"
+                )
+            }} + 1, 100
+        )
+    ) as academic_year_label,
```
cbini: use 4 digit year for both start and end years

src/cube/mcp/server.py
```diff
@@ -314,6 +297,73 @@ def _meta_cache_path(email: str) -> Path:
...
+    m = _AY_RANGE_4_RE.match(value.strip())
+    if not m:
+        return None
+    start, end = int(m.group(1)), int(m.group(2))
+    if start > end or end - start > 10:
+        return None
+    return [str(y) for y in range(start, end + 1)]
+
+
+def _validate_query(query: dict[str, Any]) -> dict[str, Any]:
```
cbini: do we still need this, or did the model restructuring handle it? are there residual issues that we did not originally address?

cbini (changes requested):
Re-reviewed at head `f717f91`. The academic-year semantics work has been spun off to #4084 so this PR can stay focused on the chronic-absenteeism fix — so this review covers **CA correctness only**, plus a scoping note on what moves to #4084. Two reviewers (Cube/JS and dbt) went over the diff independently; the dbt half is solid, but the `queryRewrite` guard has two correctness regressions to fix before merge.

## Must fix (CA correctness)

### 1. The snapshot guard fires on *every* attendance measure, not just CA/truancy 🔴
[cube.js](src/cube/cube.js) — `const measures = (query.measures ?? []).filter((m) => m.startsWith("attendance"))`

Because view members are `attendance_summary.*` / `attendance_detail.*`, `startsWith("attendance")` matches **all** measures, including non-snapshot ones (`avg_daily_attendance`, `count_students`, `pct_tardy`, `count_absent_days`). Two consequences I traced through the logic:

- `avg_daily_attendance` grouped by **quarter or year** granularity now **throws** the "does not support … granularity" error — but ADA is additive and perfectly valid at those grains. Regression.
- A full-year `avg_daily_attendance` with null granularity gets `is_latest_record = true` injected, collapsing ADA to last-day-only — silently wrong.

Fix: scope the guard to the CA/truancy snapshot-flag measure families (an explicit measure-name set or suffix match), not every measure on the cube.

### 2. Point-in-time via `timeDimensions` gets the wrong anchor injected 🔴
[cube.js](src/cube/cube.js) — the `alreadyAnchored` block

`alreadyAnchored` inspects `filters` and `dimensions` but not `timeDimensions`. A point-in-time query written the idiomatic Cube way:

```json
{ "timeDimensions": [{ "dimension": "...dim_dates_date_day", "dateRange": ["2025-11-01","2025-11-01"] }] }
```

has `granularity = null`, isn't caught by any `alreadyAnchored` clause, isn't `granularity === "day"`, so `is_latest_record = true` is injected — replacing the user's pinned-date result with a year-end snapshot, silently. The `instructions=` block actively steers the LLM toward `timeDimensions` for dates, so this path will be hit.

Important nuance: the **correct** fix is narrow — treat a `timeDimensions` entry on `dim_dates_date_day` as anchored only when it is a single day:

```js
(query.timeDimensions ?? []).some(
  (td) =>
    td.dimension?.endsWith("dim_dates_date_day") &&
    ((Array.isArray(td.dateRange) && td.dateRange[0] === td.dateRange[1]) ||
      td.granularity === "day"),
)
```

Treating *any* `date_day` timeDimension as anchored (the broader version) would re-open the overcount: a wide `dateRange` with null granularity would skip injection and count "ever-CA-in-range." Use the single-day condition above.

### 3. Stale `is_latest_record` description — contingent on the AY scoping
[attendance.yml](src/cube/model/cubes/attendance/attendance.yml) — the `is_latest_record` description still says "use with `dim_terms.academic_year`." If the AY changes move to #4084 and `dim_terms_academic_year` stays exposed, this self-resolves. If any AY consolidation remains here, update it to the `dim_dates` member. Flagging so it isn't missed either way.

## Verified good (no change needed)

- **dbt snapshot flags** — `is_month_end_record` / `is_week_end_record` were empirically validated against the PR CI build (783k partitions: exactly one flag per membership period, none on no-membership periods, PK uniqueness intact). The `if(membership_value = 1, date_key, null) desc nulls last` + trailing `and membership_value = 1` pattern is correct.
- **Named period-end measures** — numerator and denominator anchored to the same period; eligibility filters consistent within each family. `pct_*_year_end` is correct-by-construction.
- Minor (non-blocking): `membership_value` is `{0, 1, NULL}` in prod, so the truancy `membership_value > 0` vs CA `membership_value = 1` difference is a no-op today. Worth a one-line comment noting the flags rely on integral membership, in case fractional dual-enrollment ever lands.

## Scope — move to #4084

To keep this PR CA-only, these AY changes should move to #4084:

- `academic_year_label` (in `dim_dates.sql`, `dim_dates.yml`, and Cube `dates.yml`)
- Retiring `dim_terms_academic_year` from the views / exposing the `dim_dates` AY members
- The `_validate_query` multi-year-range check in `server.py` and the AY rewrite of the `instructions=` block

Staying here (CA): the `queryRewrite` snapshot guard, the named `_year_end`/`_month_end`/`_week_end` measures, and the `is_month_end_record`/`is_week_end_record` dbt columns.

## Assessment

**Request changes.** Fix #1 and #2 (both correctness regressions in the guard); resolve #3 per whichever way the AY scoping lands; move the AY changes to #4084. The dbt work and the measure math are good to go.

cbini:
@claude update your review

cbini (approved):

PR #3755, opened 2026-04-28
docs(cube): Cube model YAML design spec

cristinabaldor:
## Summary & Motivation

When merged, this pull request will add the Cube model YAML design spec for the semantic layer sitting on top of the dbt mart models in BigQuery, and bring existing models and cube.js configuration into alignment with the spec.

The spec covers:
- How all 72 dbt mart models map to ~42 cube files across 14 domains, with primary models and inlined models documented in a domain breakdown table
- Three cube structural patterns: conformed dims, fact-based domain cubes, and SCD2 period intersection cubes
- Two views per domain (detail and summary) with access policy patterns enforcing the three-layer security model
- Six design decisions documenting non-obvious architectural choices (TIMESTAMP joins, period intersection in Cube SQL, strict-chain traversal, role-playing FKs, GREATEST/LEAST LEFT JOIN constraint, detail/summary split)

All reference YAML examples have been verified against actual dbt model property files for correct column names, PKs, and FKs.

## AI Assistance

Spec was collaboratively authored with Claude Code: structure, patterns, and domain breakdown were human-directed; Claude drafted YAML examples, verified column names against mart model YAMLs, and wrote the design decision entries.

### General

- [ ] Update **due date** and **assignee** on the [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
- [ ] Review the **Claude Code Review** comment posted on this PR. Address valid feedback; dismiss false positives with a brief reply explaining why.

## CI checks

- [ ] **Trunk** — passes. If it fails, run `trunk check` and `trunk fmt` locally, fix any issues, and push again.

Closes #3658

Issue #3658, opened 2026-04-17
feat: Cube model YAML design


cristinabaldor:
Define the YAML conventions, patterns, and reference implementations for all 67 Cube model files (cubes + views) that sit on top of the infrastructure established in #3591.

## Scope

- File and folder organization for `src/cube/model/cubes/` and `src/cube/model/views/`
- Two cube types: conformed dimensions and domain cubes
- SCD2 period intersection pattern for point-in-time queries
- Two views per domain: `{domain}_detail` and `{domain}_summary`
- Access policy patterns for PII, compensation, student data, and detail/summary enforcement
- Domain breakdown: which dbt mart models map to which cube files, which are inlined
- One reference implementation per domain (12 domains + conformed layer)

## Out of scope

- Full field enumeration for all 67 models — implementation plan
- Pre-aggregations — follow-up spec
- Downstream integrations (Tableau, Superset, Streamlit) — follow-up spec
- `cube.js` configuration — covered in #3591

## Design spec

`docs/superpowers/specs/2026-04-17-cube-model-yaml-design.md` (this branch)

## Implementation plan

TBD — pending spec review

---

### Glossary

| Term | Plain English |
| --- | --- |
| **Conformed dimension** | A dimension with a consistent definition across multiple domains, shared as a join target. `dim_locations` and `dim_dates` are the clearest examples — every domain filters and groups by them, and they mean the same thing in every context. |
| **SCD Type 2 (versioned)** | A table where each change adds a new row with `effective_date_start` and `effective_date_end`, preserving full history. `dim_work_assignment_jobs` has a row for every job title change, not just the current one. |
| **Period intersection** | When two or more SCD2 tables are joined together for a point-in-time query, their effective date ranges may not align. Period intersection computes overlapping periods using `GREATEST(start_a, start_b)` and `LEAST(end_a, end_b)`, producing one composite row per valid combination with its own effective date range. A single `BETWEEN` join to `dim_dates` then slices all attributes to the same point in time correctly. |
| **Fact-based domain cube** | A domain cube built on a fact table that already has a date on every row (e.g., `fct_student_attendance_daily`). No period intersection needed — the fact date drives the join to `dim_dates`. |
| **SCD2 domain cube** | A domain cube built on snapshot data with no single event date (e.g., headcount). Uses period intersection across Type 2 child tables to produce composite effective dates. |
[...]

PR #4067, opened 2026-05-28
test(dbt): guard primary-exclusivity invariant for Cube manager resolver

### LINKED ISSUE #3838: feat(cube): point-in-time role-playing manager via Cube period-intersection cube
## Problem

Gradebook and observations Cube views need a teacher's (or observed staff's) **manager** alongside their own name. The original ask was to denormalize `manager_name` / `manager_staff_key` onto `dim_staff`. That's the wrong solution: it violates the marts strict-chain rule (denormalizes another staff row's `full_name`) and is **current-only** — it can't answer "who was their manager *at the time* of this observation/gradebook record?", which is the real need.

The premise "Cube cannot join the same cube twice under different aliases" is also imprecise. Cube identifies a join by the target cube's `name`; you can join the same physical table twice by giving each role its own cube name via `extends`. The spec already uses this for `staff_observer` / `staff_survey_subject`. So a role-playing `staff_manager` alias is the right mechanism — no `dim_staff` change.

## Why it isn't a single existing join

"Manager at date D" is a three-table, point-in-time intersection, because the facts carry a *person* FK (`teacher_staff_key`) while the manager lives at *work-assignment* grain:

```text
primary assignment @ D        (dim_work_assignment_primary,                SCD2)
  -> reporting-relationship @ D (dim_work_assignment_reporting_relationships, SCD2)
    -> manager_staff_key        (role FK -> dim_staff)
```

`staff_key` is not unique in `dim_staff_work_assignments` (a person can hold multiple assignments), so the primary assignment at date D must be resolved before the manager.

## Resolution (decision: resolve in Cube SQL, not dbt)

Per the cube-model spec's design decision *"Period intersection lives in Cube SQL, not dbt,"* build the intersection as a shared period-intersection cube (Pattern 3), mirroring `staff_work_history`:

- **`staff_reporting_relationships`** cube (`public: false`) — `sql:` block joins `dim_staff_work_assignments` (bridges `work_assignment_key` to `staff_key`) inner-join `dim_work_assignment_primary` (INNER, `is_primary_position` — the "hard requirement for primary") inner-join `dim_work_assignment_reporting_relationships`, with `GREATEST`/`LEAST` over the spans. Exposes `(staff_key, manager_staff_key, effective_start_date, effective_end_date)`. The `staff_` prefix keeps it under `queryRewrite`'s `startsWith("staff")` gate.
- Each staff-keyed fact date-range-joins it `many_to_one` on `{role}_staff_key = staff_key AND {record_date_key} BETWEEN effective_start_date AND effective_end_date` (inclusive, per the student-status-dim precedent).
- Manager attributes resolve through `staff_manager` (`extends: staff`, `public: false`) joined on `manager_staff_key`. Generalizes to `observer` / `survey_subject`.

`dim_staff` is untouched. No new dbt mart model.

## Validated against prod (`kipptaf_marts`)

- Resolver grain is **unique**: 10,387 rows = 10,387 distinct `(staff_key, effective_start_date)` across 3,231 staff. **No concurrent-primary fan-out.**
- Real-fact join is **`many_to_one`**: of 47,585 `fct_staff_observations`, **0 fan-out** (`max_matches = 1`); 47,537 resolve to exactly one manager, 48 to none (observations dated before the teacher's primary/reporting span -> NULL manager, correct).
- Manager coverage strong: 0 NULL `manager_staff_key` across the intersected set (the #3869 gap doesn't bite the primary-assignment population).
- Nesting: source dims are all views; resolver + `dim_staff`-joined-twice executes clean. A Cube inline cube adds no persisted view level.

## Scope of this issue

1. **dbt:** a guard data test asserting the **primary-exclusivity invariant** (≤1 primary assignment per staff at any date) on `dim_work_assignment_primary` × `dim_staff_work_assignments`. The resolver's unique grain depends on it; it holds today but has no test.
2. **Spec:** `docs/superpowers/specs/2026-05-28-staff-reporting-relationships-dim-design.md` — the authoritative pattern doc that unblocks the Cube work.
3. **Cube cube (`staff_reporting_relationships` + `staff_manager`):** depends on the `staff` cube, which lives on the cube-model branch (`cristinabaldor/feat/claude-cube-model-yaml-design`), not `main`. Authored there, not on this branch. Supersedes that spec's current-manager gradebook note.

## Out of scope (follow-up)

The "who can see all of a person's data **today**" reporting chain — the `queryRewrite` `dim_staff.reporting_chain` segment ([`cube.js`](https://github.com/TEAMSchools/teamster/blob/main/src/cube/cube.js#L178-L187)) — is a separate access/security mechanism (current transitive closure), already scoped to the cube-infra follow-up. Tracked separately.

## Dependencies

- [#3869](https://github.com/TEAMSchools/teamster/issues/3869) — Dayforce-era manager rows; only affects historical coverage completeness, not the structure (and prod shows strong current coverage).

cbini:
## Summary & Motivation

> When merged, this pull request adds the dbt guard test that protects the invariant behind a point-in-time, role-playing **manager** in the Cube gradebook/observations/surveys views — resolved without denormalizing manager attributes onto `dim_staff`.

The manager-resolver design itself is being incorporated into the cube-model spec (#3658 / PR #3755) — the `staff_reporting_relationships` (period-intersection cube) + `staff_manager` (`extends: staff`) approach supersedes that spec's current-manager gradebook note. This PR carries only the **dbt slice** of that work.

**In this PR:**

- `tests/dim_work_assignment_primary__no_overlapping_primary_spans.sql` — a singular test asserting no staff has two simultaneously-primary work assignments (`lead()` overlap check over `dim_work_assignment_primary` joined to `dim_staff_work_assignments`, filtered to `is_primary_position`). This primary-exclusivity invariant underpins the unique grain of the Cube resolver; it holds today but had no test. **Passes against prod (0 violations).**
- `docs/superpowers/specs/2026-05-28-staff-reporting-relationships-dim-design.md` — design record + prod validation evidence for the resolver. (Disposition pending: keep as a standalone deep-dive, or fold into the #3755 spec.)

**Validated against prod (`kipptaf_marts`):** resolver grain unique (10,387 rows = 10,387 PKs, 3,231 staff); real-fact join to `fct_staff_observations` is `many_to_one` (0 fan-out across 47,585 observations); 0 NULL managers in the intersected set.

**Out of scope:** the `staff_reporting_relationships` + `staff_manager` cubes themselves — they land in the #3755 spec implementation, against the `staff` base cube.

## AI Assistance

Collaboratively authored with Claude Code: design direction, the strict-chain pushback, and the Cube-vs-dbt placement decision were human-directed; Claude validated the resolver against prod, wrote the spec and the guard test, and verified the test passes.

### dbt

- [x] Properties entry (description + `meta.dagster.ref` → `dim_work_assignment_primary`) included for the new singular test. No new model, so no `[model_name].yml` model properties file applies.
- [x] No exposure change — no dashboard/application consumes a new or renamed model.
- [x] Not a breaking change — no contracted-model column renames/removals.

## CI checks

- [ ] **Trunk** — lint-clean locally (`trunk check --force` on both files).
- [ ] **dbt Cloud** — the new test runs in CI.
- [ ] **Dagster Cloud** — not triggered (no Dagster paths changed).

Refs #3658

### DIFF
--- a/docs/superpowers/specs/2026-05-28-staff-reporting-relationships-dim-design.md
+++ b/docs/superpowers/specs/2026-05-28-staff-reporting-relationships-dim-design.md
@@ -1,186 +1,216 @@
-# Staff Reporting-Relationships Dim Design
+# Point-in-Time Role-Playing Manager for Cube
 
 **Date:** 2026-05-28 **Status:** Active **Issue:**
 [#3838](https://github.com/TEAMSchools/teamster/issues/3838)
 
 ## Overview
 
-Add a person-grained, point-in-time dimension —
-`dim_staff_reporting_relationships` — that exposes "who was this staff member's
-manager on date D." It lets the gradebook, observations, and surveys Cube cubes
-surface a teacher's / observed staff's / survey subject's manager via a single
-date-range join plus a role-playing `staff_manager` cube (`extends: staff`),
-without denormalizing manager attributes onto `dim_staff` and without each
-consuming cube re-deriving a fragile multi-table period intersection.
+Let the gradebook, observations, and surveys Cube cubes surface a teacher's /
+observed staff's / survey subject's **manager as of the record date** via a
+single date-range join plus a role-playing `staff_manager` cube
+(`extends: staff`) — without denormalizing manager attributes onto `dim_staff`,
+and without each consuming cube re-deriving a fragile multi-table period
+intersection.
+
+The manager-at-a-date resolution is built as a shared **Cube period-intersection
+cube** (`staff_reporting_relationships`), per the cube-model spec's design
+decision _"Period intersection lives in Cube SQL, not dbt."_ The only dbt change
+is a **guard test** protecting the invariant the resolver's grain depends on.
 
 This supersedes the original #3838 plan (denormalize `manager_name` /
 `manager_staff_key` onto `dim_staff`), which violated the marts strict-chain
 rule and was current-only.
 
-## Goals
+## Decision: resolve in Cube SQL, not a dbt dim
 
-- A reusable dbt mart dim, keyed by the subordinate's `staff_key`, that answers
-  "manager at a point in time" with one date-range join.
-- Strict-chain compliant: `dim_staff` is untouched; manager identity is a role
-  FK (`manager_staff_key`) reachable by traversal, not a denormalized name.
-- Resolve the primary-assignment + reporting-relationship period intersection
-  **once, upstream**, so gradebook / observations / surveys cubes each do a
-  single clean join.
+Two strict-chain-compliant placements were considered for the
+primary-assignment + reporting-relationship intersection:
 
-## Non-goals
+- **Cube SQL** (chosen) — a shared `staff_reporting_relationships` cube whose
+  `sql:` block computes the intersection, mirroring `staff_work_history`.
+- **dbt dim** — a new `dim_staff_reporting_relationships` mart at
+  `(staff_key, effective_start_date)` grain.
 
-- **Cube YAML implementation** (the `staff_reporting_relationships` cube, the
-  `staff_manager extends staff` alias, the per-fact joins) — that lands in the
-  Cube model work on `cristinabaldor/feat/claude-cube-model-yaml-design`. This
-  spec covers the dbt dim only, and documents the intended Cube consumption for
-  context.
-- **The "who can see their data today" access chain** — the `queryRewrite`
-  `dim_staff.reporting_chain` segment is a separate current-transitive-closure
-  security mechanism, already scoped to the cube-infra follow-up. Out of scope.
-- **Widening manager history** — pre-2021 NJ (Dayforce-era) manager coverage is
-  owned by [#3869](https://github.com/TEAMSchools/teamster/issues/3869). This
-  dim reads downstream of that and will gain coverage when #3869 ships.
+Cube SQL wins because:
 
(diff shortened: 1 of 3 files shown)


cbini:
Addressed in a6942a8c5.

- **Wrong test name in validation command** — fixed; was `test_staff_primary_assignment_no_overlap`, now `dim_work_assignment_primary__no_overlapping_primary_spans`.
- **Spec SQL diverged from implementation** — fixed; the spec's guard-test section now shows the `lead()` SQL that actually shipped, with a one-line note that it replaces the self-join draft (equivalent for sorted spans, matches the sibling `dim_student_*_status__no_overlapping_spans` tests).
- **Cardinality assumption** — added a comment in both the test and the spec. Sharpening the framing: it's not just "holds today" — `work_assignment_key` is the **primary key** of `dim_staff_work_assignments`, so the join attaches exactly one `staff_key` per primary span and structurally cannot fan out the `lead()` ordering.
- **BETWEEN context in the Cube YAML example** — added an inline comment on the join (inclusive end-dates, prod-validated, zero fan-out).
- **Single-line final SELECT** — not changed. trunk-fmt produced it: the column list fits in 88 chars so the formatter collapses it to one line; the sibling tests stay multi-line only because their longer column lists exceed 88. Forcing multi-line would be reverted by the `trunk fmt` pre-commit hook, so this is formatter-deterministic, not a style choice.

Issue #4070, opened 2026-05-29
fix(google_sheets): strip leading space from Fisk University Account_ID in NSC crosswalk


anthonygwalters:
## Problem

The NSC crosswalk Google Sheet has a stale row with a leading space in `Account_ID`:

| Field | Value |
|-------|-------|
| `College_Name_NSC` | `FISK UNIVERSITY` |
| `College_State_NSC` | `TN` |
| `College_Code_NSC` | `003490-00` |
| `Account_ID` | `" 0018000000Y9oYrAAJ"` (19 chars; canonical Salesforce ID is 18 chars without the leading space) |

Sheet: <https://docs.google.com/spreadsheets/d/1smCSUKz2SXwnohmz59LNptoN5jHzYD2TfZNHZwK7Dts> (tab `src_kippadb__nsc_crosswalk_v2`).

## Impact

The leading space causes string-equality joins from `stg_google_sheets__kippadb__nsc_crosswalk.account_id` against the canonical Salesforce account ID in `stg_kippadb__enrollment.school` to miss. In the NSC enrollment-overlap check, this produces false-positive "new enrollment" rows for Fisk students who already have a Salesforce enrollment at that school.

## Temporary mitigation (already shipped)

`stg_google_sheets__kippadb__nsc_crosswalk` now applies `trim(account_id)` defensively, so the join works even with stale sheet data. The defensive trim can stay or be removed once the sheet is clean.

## Ask of Ops

Delete the leading space character in the `Account_ID` cell for the `FISK UNIVERSITY` / `003490-00` row in the linked sheet. After the edit, the Dagster external-table refresh will land the clean value automatically — no dbt rerun required.

## Verification

After Ops edit, this should return 0 rows:

```sql
select account_id
from `teamster-332318.kipptaf_google_sheets.stg_google_sheets__kippadb__nsc_crosswalk`
where account_id != trim(account_id)
```

PR #4072, opened 2026-05-29
fix(kipptaf): rebuild NSC enrollment grain on continuous stints

anthonygwalters:
## Summary

- Rebuild the three NSC rpt views around a new `int_nsc__enrollment_stints` intermediate that chains terms into stints (breaks at gaps > 6 months) and attaches graduation to the stint containing NSC's `graduation_date`.
- Fix a case-sensitivity bug: NSC uppercases its `requester_return_field` echo while Salesforce stores 18-char IDs in mixed case, so the prior `contact_id = student` join missed virtually every match. Now recovered in staging via `upper()`-keyed lookup against `stg_kippadb__contact`.
- Trim `account_id` in the NSC crosswalk staging defensively; the underlying Google Sheet has one stale row with a leading space (tracked separately in #4070).

## Row-count impact (dev vs prod)

| View | Prod | New | Read |
|------|-----:|----:|------|
| `nsc_enrollment_new` | 2,344 | 440 | 81% fewer false positives |
| `nsc_enrollment_updates` | 97 | 717 | Case-fix exposed real updates that were never surfaced |
| `nsc_terms_new` | 202 | 1,599 | Same root cause |

## Hand-audit

Six diverse cases hand-audited against Salesforce after the rebuild — all six produce correct rows.

## Test plan

- [x] dbt build of stg_nsc__student_tracker+ passes
- [x] Uniqueness tests pass on all four NSC models (int_nsc__enrollment_stints, rpt_gsheets__nsc_enrollment_new, rpt_gsheets__nsc_enrollment_updates, rpt_gsheets__nsc_terms_new)
- [x] Trunk clean on all touched files
- [x] Hand-audited 6 sample rows from nsc_enrollment_new against underlying NSC + Salesforce data

Refs #4070


anthonygwalters:
Addressed in e55c5c561.

**Item 1** (NULL stint_end) — confirmed: 0 NSC rows currently have non-NULL `enrollment_begin` paired with NULL `enrollment_end`, but the schema allows it for in-progress terms. Added `coalesce(stint_end, date(9999, 12, 31))` defensively in the overlap predicates of `rpt_gsheets__nsc_enrollment_new` and `rpt_gsheets__nsc_enrollment_updates`.

**Item 2** (BETWEEN) — fixed in `rpt_gsheets__nsc_terms_new` (term-to-parent join is exactly the chained-enrollment shared-boundary case the convention targets). Leaving the BETWEEN in `int_nsc__enrollment_stints` graduation check: it's a single date checked against a single stint's range (not range-meets-range), and stints are built with > 6-month gaps so adjacent stints can't share a boundary date — fan-out is structurally impossible. The convention is about chained-range joins, not range-contains-date.

**Item 3** — added the grain-projection annotation on `covered_stints`.

**Item 4** — added description for `contact_id`.

**Minor (uniqueness invariant)** — added an inline comment on the `nsc_enrollment_new` uniqueness test explaining why `(student, school, start_date)` is a valid PK proxy for `(student, school, stint_num)`.

Re-built and re-tested all four NSC models against the dev schema; all uniqueness tests pass. Row-count delta from the changes was ≤3 rows on each view.

anthonygwalters:
> Addressed in [e55c5c5](https://github.com/TEAMSchools/teamster/commit/e55c5c5616da42eff8eace2062ae91e20dc7ba16).
> 
> **Item 1** (NULL stint_end) — confirmed: 0 NSC rows currently have non-NULL `enrollment_begin` paired with NULL `enrollment_end`, but the schema allows it for in-progress terms. Added `coalesce(stint_end, date(9999, 12, 31))` defensively in the overlap predicates of `rpt_gsheets__nsc_enrollment_new` and `rpt_gsheets__nsc_enrollment_updates`.
> 
> **Item 2** (BETWEEN) — fixed in `rpt_gsheets__nsc_terms_new` (term-to-parent join is exactly the chained-enrollment shared-boundary case the convention targets). Leaving the BETWEEN in `int_nsc__enrollment_stints` graduation check: it's a single date checked against a single stint's range (not range-meets-range), and stints are built with > 6-month gaps so adjacent stints can't share a boundary date — fan-out is structurally impossible. The convention is about chained-range joins, not range-contains-date.
> 
> **Item 3** — added the grain-projection annotation on `covered_stints`.
> 
> **Item 4** — added description for `contact_id`.
> 
> **Minor (uniqueness invariant)** — added an inline comment on the `nsc_enrollment_new` uniqueness test explaining why `(student, school, start_date)` is a valid PK proxy for `(student, school, stint_num)`.
> 
> Re-built and re-tested all four NSC models against the dev schema; all uniqueness tests pass. Row-count delta from the changes was ≤3 rows on each view.

@claude re-review in light of this comment

src/dbt/kipptaf/models/nsc/staging/stg_nsc__student_tracker.sql
```diff
@@ -1,39 +1,87 @@
...
 
-    left(requester_return_field, length(requester_return_field) - 1) as contact_id,
-from {{ source("nsc", "src_nsc__student_tracker") }}
+    /* NSC uppercases the requester_return_field echo; recover canonical
+       mixed-case Salesforce ID via case-insensitive contact lookup so
+       downstream joins to kippadb succeed. */
+    coalesce(c.id, n.contact_id_raw) as contact_id,
+from nsc_raw as n
+left join
+    {{ ref("stg_kippadb__contact") }} as c on upper(n.contact_id_raw) = upper(c.id)
```
cbini, line 87: i'd prefer this be downstream but it's fine for now

cbini (approved):

PR #4093, opened 2026-06-02
feat(dbt): add stg_kippadb__kipp_aid staging model

anthonygwalters:
## Summary & Motivation

Add a staging model for the Salesforce `KIPP_Aid__c` object, bringing it in line with the other `kippadb` staging models.

- `stg_kippadb__kipp_aid.sql` — flat select against `{{ source("kippadb", "kipp_aid") }}` (already declared in `sources-bigquery.yml`), `__c` suffixes stripped and columns snake_cased, reserved/keyword-like columns backticked (`name`, `date`, `type`), `_airbyte_*` plumbing columns dropped, and the standard soft-delete filter `where not isdeleted`.
- `stg_kippadb__kipp_aid.yml` — minimal properties file matching the existing kippadb staging models (column + `data_type`), plus a `unique` test on `id` (`severity: error`).

Grain verified: `id` is unique at 3,533 rows; all rows currently have `IsDeleted = false`. The model builds clean (`dbt build`) and trunk is green.

**AI Assistance:** Claude Code authored the model and YAML, profiled the source table, and built/tested it; column naming, the soft-delete convention, and the uniqueness test were human-directed via the existing kippadb pattern.

### dbt

- [ ] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — n/a (no downstream consumer added)
- [ ] **Breaking change?** — no, additive only
- [ ] If adding or modifying an external source, run `stage_external_sources` with `--target staging` — n/a, this is a BQ-native source (`kipptaf_kippadb.KIPP_Aid__c`), not an external/GCS source

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

### DIFF
--- a/src/dbt/kipptaf/models/kippadb/staging/properties/stg_kippadb__kipp_aid.yml
+++ b/src/dbt/kipptaf/models/kippadb/staging/properties/stg_kippadb__kipp_aid.yml
@@ -1,0 +1,37 @@
+models:
+  - name: stg_kippadb__kipp_aid
+    columns:
+      - name: id
+        data_type: string
+        data_tests:
+          - unique:
+              config:
+                severity: error
+      - name: name
+        data_type: string
+      - name: created_by_id
+        data_type: string
+      - name: created_date
+        data_type: timestamp
+      - name: last_activity_date
+        data_type: date
+      - name: last_modified_by_id
+        data_type: string
+      - name: last_modified_date
+        data_type: timestamp
+      - name: system_modstamp
+        data_type: timestamp
+      - name: amount
+        data_type: numeric
+      - name: date
+        data_type: date
+      - name: kipp_aid_count
+        data_type: numeric
+      - name: notes
+        data_type: string
+      - name: status
+        data_type: string
+      - name: student
+        data_type: string
+      - name: type
+        data_type: string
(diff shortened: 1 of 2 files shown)


anthonygwalters:
Thanks for the review. Addressing each point — no code changes, with reasoning:

**`quote: true` for `name` / `date` / `type` — declining (false positive).**

- None of `name`, `date`, or `type` are BigQuery reserved words. `DATE` and `TYPE` are type/function names, not reserved keywords (see the [reserved keyword list](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#reserved_keywords)). The CLAUDE.md `quote: true` rule is scoped to reserved words, so it doesn't trigger here — these are backticked in SQL only to match the existing kippadb convention.
- The contracted sibling `stg_kippadb__gpa` already declares `type`, `source`, and `name` columns with no `quote: true` and is in production, so the "contract check can fail / silently misresolve" concern is empirically ruled out.
- `dbt build` on this model passed contract enforcement locally (kippadb staging is `contract: enforced: true`) with the YAML as-is.

Keeping it consistent with the 15 existing kippadb staging models.

**`not_null` on `id` — declining.** Salesforce `Id` is the record primary key and is never null; the `unique` test already guards the grain, and no sibling kippadb staging model adds `not_null`. Leaving it off to match the pattern.

**Descriptions — declining for now.** CLAUDE.md does call for them on new models, but all 15 existing kippadb staging models omit them and matching that pattern was the explicit ask here. Better to backfill descriptions across the whole kippadb staging folder in one pass than to do this one model in isolation.

Appreciate the confirmation on the well-done items (staging-layer soft-delete, `__c` stripping, `student` naming, top-sorted tested column, no `_airbyte_*` columns).

Issue #4091, opened 2026-06-02
fix(claude-hooks): harden check-sensitive.sh & check-output.sh — confirmed bypasses + fail-open defaults


cbini:
## Summary

An adversarial multi-agent code review of the two security hooks — `.claude/hooks/check-sensitive.sh` (PreToolUse) and `.claude/hooks/check-output.sh` (PostToolUse) — surfaced **6 Critical** and **~19 Important/Minor** issues. Every bypass was empirically reproduced by piping fixtures through the real hooks (`deny=0` = allowed through, `deny=1` = blocked); three plausible-but-wrong leads were refuted by the verification pass.

Full per-finding harness evidence is retained locally (`.claude/scratch/hook-review-report.md` + workflow transcripts) and intentionally **not** pasted here — the fixtures are secret-shaped strings that should stay off external surfaces per the PII convention.

These hooks are documented as the **sole enforcement layer** (the Codespace sandbox is ineffective), so each confirmed bypass defeats a boundary with no backstop.

**Assessment: not mergeable as-is.** Highest-leverage fix = fail-closed defaults + `tool_name` normalization (small, safe diff that neutralizes a cluster of findings).

## Critical (must fix)

1. **Cert / private-key files under any non-`file_path` key are not blocked** — `check-sensitive.sh:73,79` + lines 48-50. Rule 1 requires the cert/key extension to be a whole path *segment*; the bare-extension catch (Rule 1b) scans only four named keys with no recursive descent. PEM/CER/KEY files referenced under other MCP keys (`arg`, `uri`, `url`, `source`, `localPath`) or a nested object are allowed (`deny=0`). Fix: move the bare-extension check into the all-string scan, or build the path corpus with recursive `.. | strings`.

2. **Symlink resolution covers only `.tool_input.file_path`** — `check-sensitive.sh:36`. `readlink -f` runs on one key only; Grep (`path`), Glob (`pattern`), and MCP path keys get no resolution, so a symlink to a sensitive dir is read through unresolved (`deny=0`), and Grep returns the target file contents. Fix: resolve symlinks for every path-bearing field.

[...]

PR #4096, opened 2026-06-03
fix(claude-hooks): harden security hooks — batches 1-3 (#4091)

cbini:
## Summary & Motivation

When merged, this PR hardens the two Claude Code security hooks
(`.claude/hooks/check-sensitive.sh` PreToolUse and `.claude/hooks/check-output.sh`
PostToolUse) against confirmed bypasses surfaced in #4091. These hooks are the
sole enforcement layer in Codespaces, so each bypass defeated a boundary with no
backstop.

This is the first three of five planned batches, ordered by ascending blast
radius. Full plan: `docs/superpowers/plans/2026-06-03-claude-hooks-4091-hardening.md`.
Each fix was driven by a failing regression test first, then validated against
the full suite.

### Batches in this PR

- **Fail-closed defaults + tool_name normalization** — both hooks now DENY on
  empty stdin, unparseable JSON, a non-object frame, or a missing tool_name
  instead of silently allowing; tool_name is matched case- and
  whitespace-insensitively. Closes 6, 8, 9, 26.
- **Anchored variable allowlist + declare dump forms** — the shell-variable
  strip no longer lets a safe prefix leak a longer secret variable name; the
  process-dump rule now catches `declare -p`, bare `declare`, and `declare -px`.
  Closes 4, 5.
- **BigQuery read-only gate** — Section 3 now gates every `mcp__bigquery__*`
  tool (not just two), with a per-tool read-only allowlist for `execute_sql`, an
  embedded-write denylist (now covering TRUNCATE / LOAD DATA, INSERT without
  INTO, and UPDATE..SET split across lines) for `forecast` and
  `analyze_contribution`, and a corpus denylist for unknown tools. Closes 3,
  10, 11, 12, 13, and 28 (a bare table-id is no longer wrongly denied).

Design note: `ask_data_insights` is intentionally NOT SQL-gated. It has no
raw-SQL argument — it takes a natural-language question plus structured table
references, and the server performs read-only NL-to-SQL. A denylist on its prose
would block ordinary analyst questions (words like "drop", "deleted"). Deliberate
deviation from the literal wording of 3.

### Deferred to follow-up batches (issue stays open)

- 1 / 2 — scan all string leaves for cert/key files and resolve symlinks for
  every path field (path-corpus rework; highest regression risk).
- 21 / 22 — secret-value egress scan for write-capable MCP tools and the
  outbound URL scanner; folds in 6e / 20.

Refs #4091.

## AI Assistance

Authored by Claude Code via the systematic-debugging workflow — each finding was
reproduced against the live hooks before any fix, failing test first. Scope,
batch sequencing, and the `ask_data_insights` / unknown-tool design decisions
were human-directed. The protected hook files were applied and committed by the
human.

### General

- [ ] Review the Claude Code Review comment posted on this PR; address valid
      feedback or dismiss false positives with a brief reply.

## CI checks

- [ ] Trunk — passes.

## Test evidence

All hook regression suites green against the patched hooks:
Sensitive-paths 52, Bypass 73, Variable-protection 78, BigQuery 43,
Output-scanner 47, Self-protection 20, Fail-closed 21 — 7/7 suites passed.


cbini:
Thanks for the review — triaged all four findings.

**Finding 1 (wildcard safe-var bypass) — FIXED in `e5e624e2f`.** Confirmed real and reproduced: `COMP[A-Z_]*` and `XDG_[A-Z_]+` greedily swallowed the whole underscore-joined name, so `$COMP_TOKEN` / `$XDG_SESSION_SECRET` were stripped and allowed through. Replaced all three wildcard patterns (`COMP_*`, `XDG_*`, `LC_*`) with explicit enumerations; combined with the anchored strip this is exact-match only, so secret-suffixed names now deny while the real completion / XDG / locale vars stay allowed. Added regression tests (Variable-protection suite 78 → 90, full suite 7/7 green).

**Finding 4 (`declare -fp`) — already enforced; test added.** The `declare[[:space:]]+-[a-zA-Z]*p[a-zA-Z]*` pattern already matched `-fp`; added an explicit deny test to document the intent.

**Findings 2 & 3 (deny-helpers don't assert exit 0; shared `/tmp` stderr path) — valid, tracked for a follow-up.** Both are pre-existing test-infra concerns with low real risk (`deny()` always exits 0 structurally; `run_all.sh` runs serially today). Folding them into a dedicated test-hardening change rather than expanding this PR.

Also resolved the Trunk SC2312 failures in `837d80a0f` by switching the `tool_name` normalization and the newline-flatten to bash parameter expansion (no `tr` subshell).

On the closing repo-state note: that was a CI-checkout artifact, not a PR defect — the pushed commits are intact. The full suite is green against the committed hooks and Trunk is green on the pushed SHA, both of which exercise the `_bq_has_write` helper, the `case` gate, and the fail-closed block.

cbini:
@claude update your review

cbini:
One correction on the updated review: `expect_deny_exit0` is not dead code — it has 9 call sites: 6 in `tests/hooks/test_output_scanner.sh` (103/106/109/122/125/130) and 3 in `tests/hooks/test_sensitive_paths.sh` (166/168/170). `test_fail_closed.sh` uses `expect_deny_raw` (that's what it was written with), but the helper itself is exercised by the two suites above.

Finding 2's underlying point still stands as the follow-up item: the bulk deny assertions (`expect_deny` / `expect_deny_json` / `expect_deny_raw`) validate stdout content but not the exit-0 half of the protocol — only those 9 `expect_deny_exit0` spot-checks do that. Findings 1 and 4 are resolved; 2 and 3 are tracked for a dedicated test-hardening change.

PR #4097, opened 2026-06-03
fix(claude-hooks): scan all leaves for cert/key + resolve symlinks for every path field (#4091)

cbini:
## Summary & Motivation

When merged, this completes the path-corpus hardening from #4091 (Batch 4 of the
plan), building on the merged #4096:

- **#1 — cert / private-key files under non-path keys.** Rule 1b previously
  scanned only file_path / command / path / pattern, so a cert or key file
  referenced under an MCP key (uri, nested localPath, source) slipped both Rule 1
  (whole-segment match) and Rule 1b. It now scans `no_content` (all string leaves
  except Write/Edit body), so those are caught while Python attribute access
  (e.g. `asset.key`) in written code / doc content stays exempt.
- **#2 — symlink resolution for every path field.** `readlink -f` ran only on
  `.tool_input.file_path`, so a symlink supplied via Grep `path`, Glob `pattern`,
  or an MCP path key was read through unresolved. Resolution now covers every
  path-bearing string leaf that exists on disk (body content excluded).

Failing tests written first; full hook suite 7/7 green (Sensitive-paths 55 → 60),
shellcheck clean (the jq-in-process-substitution was captured to a variable to
avoid a masked-return lint).

Refs #4091.

## AI Assistance

Authored by Claude Code via the systematic-debugging workflow — each bypass was
reproduced against the live hooks before the fix, failing test first. The
protected hook file was applied and committed by the human.

### General

- [ ] Review the Claude Code Review comment posted on this PR; address valid
      feedback or dismiss false positives with a brief reply.

## CI checks

- [ ] Trunk — passes.

## Test evidence

Failing-first regression tests in `tests/hooks/test_sensitive_paths.sh`: 5 prior
bypasses now deny (cert / key file under uri / nested localPath / source; symlink
via Grep path and MCP uri); content-exemption and benign-symlink controls still
allow. Full suite 7/7 against the patched hooks.

### DIFF
--- a/.claude/hooks/check-sensitive.sh
+++ b/.claude/hooks/check-sensitive.sh
@@ -47,13 +47,20 @@ fi
 
 normalized=$(echo "${path}" | _normalize)
 
-# Resolve symlinks on file_path to prevent symlink-based bypass — cache result for reuse below
-file_path=$(jq -r '.tool_input.file_path // ""' <<<"${input}")
-if [[ -n ${file_path} && -e ${file_path} ]]; then
-  resolved=$(readlink -f "${file_path}" 2>/dev/null || echo "${file_path}")
-else
-  resolved=""
-fi
+# Resolve symlinks for EVERY path-bearing string leaf that exists on disk (not
+# just file_path) so a symlink to a sensitive target is caught regardless of
+# which key holds it — path (Grep), pattern (Glob), MCP uri/localPath/source,
+# nested objects, etc. Body content (content/new_string/old_string) is excluded:
+# that is doc/code text, not a path the tool will open.
+resolved=""
+_cands=$(jq -r '
+  [.tool_input | to_entries[] | select(.key | test("^(content|new_string|old_string)$") | not) | .value | .. | strings] | .[]
+' <<<"${input}")
+while IFS= read -r _cand; do
+  [[ -n ${_cand} && -e ${_cand} ]] || continue
+  _r=$(readlink -f "${_cand}" 2>/dev/null) || continue
+  [[ -n ${_r} ]] && resolved+=" ${_r}"
+done <<<"${_cands}"
 [[ -n ${resolved} ]] && normalized="${normalized} ${resolved}"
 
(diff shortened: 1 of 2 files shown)


cbini:
Verified the false positive and fixed it in `71bce6a56`.

Reproduced: with Rule 1b scanning `no_content`, `SELECT record.key FROM t` and `SELECT a.pem FROM t` in an `execute_sql` `sql` field both denied — the dotted column ref matched the cert/key extension once the `sql` field was in scope.

Fix: Rule 1b scans `path_only` again, but `path_only` is now rebuilt to collect named path keys **recursively** — `file_path` / `command` / `path` / `pattern` plus MCP `uri` / `url` / `localPath` / `source` at any nesting depth. So cert/key files under those keys are still caught (including a nested `localPath`), while free-text fields like `sql` are never path-scanned. The two SQL inputs now allow; the `uri` → cert-file control still denies. Both added as allow tests. Full suite 7/7, shellcheck clean.

Went with recursive named-keys rather than the top-level named-fields suggestion because one of the #1 cases is a nested `localPath`, which a top-level-only lookup would miss.

cbini:
@claude update your review

cbini:
Verdict confirmed — ready to merge, no code change needed. Two notes:

**Housekeeping (working-tree `M check-sensitive.sh`):** that's a CI-checkout artifact, not the PR. The branch working tree is clean (`git status` empty at `71bce6a56`) and the committed file has the recursive `path_only`:

```text
[.tool_input | .. | objects | (.file_path?, .command?, .path?, .pattern?, .uri?, .url?, .localPath?, .source?) | strings] | join(" ")
```

So there is nothing to discard — `git checkout HEAD -- ...` would be a no-op against the clean tree. The reviewed/committed state is the correct one.

**`url` edge case:** keeping `url` in the named-key list is intentional. Issue #4091's finding #1 lists uri/url/source/localPath as the keys to cover, and a URL whose path segment ends in `.key`/`.pem`/`.cer` is reasonably treated as a cert/key reference. Accepting as-is.

PR #4098, opened 2026-06-03
test(claude-hooks): assert exit 0 in all deny helpers + mktemp stderr (#4091)

cbini:
## Summary & Motivation

When merged, this addresses the two deferred test-infrastructure findings from
the #4096 review (re-confirmed as the follow-up during #4097's review):

- **Finding 2 — deny helpers did not validate exit code.** `expect_deny`,
  `expect_deny_json`, `expect_deny_raw`, `expect_deny2`, and `check_output`'s
  deny path checked only that deny JSON appeared on stdout — not that the hook
  exited 0. Claude Code treats a non-zero exit as a non-blocking error (the deny
  is ignored and the tool runs), so a hook that printed deny JSON but exited 1
  would pass these tests while being ineffective in production. All
  deny-expecting helpers now route through `expect_deny_exit0`, which validates
  the full protocol: exit 0 + deny JSON on stdout + deny JSON NOT on stderr.
- **Finding 3 — shared `/tmp` stderr path.** `expect_deny_exit0` wrote stderr to
  a fixed `/tmp` file; two concurrent runs would race on it. Now uses a
  per-invocation `mktemp`.

No hook behavior changes — `helpers.sh` only.

Refs #4091.

## AI Assistance

Authored by Claude Code. Findings raised by the automated PR review on #4096 /
#4097; implemented and validated here.

### General

## CI checks

- [ ] Trunk — passes.

## Test evidence

Full suite stays 7/7 green against unchanged hooks (confirming the hooks already
satisfy exit 0 on every deny). A teeth check with deliberately non-compliant
fake hooks confirms the stricter assertion now works: a compliant deny+exit0
PASSes, while deny+exit1 and deny-on-stderr-only both FAIL.

### DIFF
--- a/tests/hooks/helpers.sh
+++ b/tests/hooks/helpers.sh
@@ -29,11 +29,13 @@ _is_deny() {
 # Usage: expect_deny_exit0 <description> <hook_script> <json_input>
 expect_deny_exit0() {
   local desc="$1" hook="$2" input="$3"
-  local stdout stderr exit_code
-  stdout=$(echo "${input}" | bash "${hook}" 2>/tmp/_hook_stderr)
+  local stdout stderr exit_code tmpfile
+  # Per-invocation temp file (mktemp) — a fixed /tmp path races under parallel runs.
+  tmpfile=$(mktemp)
+  stdout=$(printf '%s' "${input}" | bash "${hook}" 2>"${tmpfile}")
   exit_code=$?
-  stderr=$(cat /tmp/_hook_stderr)
-  rm -f /tmp/_hook_stderr
+  stderr=$(cat "${tmpfile}")
+  rm -f "${tmpfile}"
   local ok=true
   if [[ ${exit_code} -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


cbini:
On Finding 3 (`mktemp` cleanup): the `rm -f "${tmpfile}"` runs unconditionally right after the stderr capture, before any verdict branch, and the function has no early `return` — so normal execution and every logical path already clean up. The only uncovered case is a SIGKILL between `mktemp` and `rm`, which a `trap … RETURN` does not fire on. So a `RETURN` trap wouldn't close that gap; leaving cleanup as the unconditional `rm`.

Findings 1 (echo vs printf in the allow-path helpers) and 2 (debug-line format) noted: 1 is pre-existing and cosmetic, out of scope for this change; 2 is the intended trade — the per-reason messages (`[exit 0]` / `[deny on stdout]` / `[no deny on stderr]`) are more actionable than the old input dump. No code changes; merging as-is.

PR #4099, opened 2026-06-03
fix(claude-hooks): outbound secret-value egress scan + check-output schema fallback (#4091)

cbini:
## Summary & Motivation

When merged, this is the final planned hardening batch (Batch 5) for #4091:

- **#22 — outbound secret-value egress scan (new Section 4 in `check-sensitive.sh`).**
  Sections 1-3 match sensitive PATHS / read-only enforcement; this catches a
  secret VALUE being SENT to a write-capable MCP tool (GitHub / Drive / Asana /
  Slack writes — gated by a write-verb name pattern) or carried in a WebFetch
  URL. Reuses `check-output.sh`'s secret-pattern set (1Password-style references,
  private-key headers, cloud tokens, connection strings, service-account JSON,
  GitHub tokens). Read-only MCP tools (bigquery / dagster / dbt
  `get_`/`list_`/`search_`) are not scanned.
- **#6e/#20 — `check-output.sh` schema fallback.** It read only `.tool_response`,
  so a payload arriving under a different key skipped scanning entirely. Now
  falls back to the whole payload when `.tool_response` is absent.
- **#21 — doc reconciliation.** `.claude/CLAUDE.md` documented a WebFetch/MCP
  URL-keyword scanner that never existed in code. Replaced that paragraph with an
  accurate description of the Section 4 egress scan. No keyword URL scanner is
  built — only secret-value shapes match (not trivially evaded, low
  false-positive rate).

Failing-test-first; full hook suite 8/8 green (new Egress suite 8, Output Scanner
+2 for #20). shellcheck clean (only the pre-existing trunk-ignored decode loop).

Operational note: the #22 scan covers GitHub MCP writes, so a PR/issue/comment
body containing a real secret-shaped value will be denied — redact such values.

Refs #4091.

## AI Assistance

Authored by Claude Code via the systematic-debugging workflow — each gap
reproduced against the live hooks before the fix, failing test first. The
protected hook files were applied and committed by the human.

### General

## CI checks

- [ ] Trunk — passes.

## Test evidence

New `tests/hooks/test_egress.sh`: write-MCP / WebFetch carrying a secret value →
deny; read-only MCP tool with a reference, benign write, benign URL → allow. Plus
two payload-key-drift cases in `test_output_scanner.sh`. Full suite 8/8 against
the patched hooks.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -28,11 +28,17 @@ regression test suite (`expect_deny_exit0`) enforces both invariants.
 
 ## What is blocked
 
-**WebFetch / MCP input scanning** — PreToolUse scans URLs and query strings for
-sensitive keywords (e.g., `/auth` in URL paths, `JWT` / `secret` / `credential`
-/ `signing key` in context7 query strings). Symptom: "Cannot access sensitive
-path" with no further detail. Rephrase the URL or query (generic terms like
-"user context", "header format") to get past it.
+**Outbound secret-value egress scan** (PreToolUse, Section 4) — write-capable
+MCP tools (tool name contains
+`create`/`update`/`write`/`add`/`comment`/`upload`/
+`send`/`post`/`put`/`delete`/`append`/`insert`/`merge`/`push`/`reply`) and
+WebFetch URLs are scanned for secret VALUES (`op://` refs, private-key headers,
+cloud tokens, connection strings — the same pattern set as `check-output.sh`). A
+match is blocked to stop exfiltration. Practical effect: a GitHub issue/PR write
+or an Asana/Drive write whose body contains a real-looking secret is denied —
+redact it (e.g. `op://…` → `op-uri`). Read-only MCP tools (bigquery / dagster /
+dbt `get_`/`list_`/`search_`) are not scanned. There is no keyword-based URL
+scanner — only secret-value shapes match.
 
 **Secret paths** (all tools blocked) — dotenv files, private key/cert files, SSH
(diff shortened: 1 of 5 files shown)


cbini:
On #1 (the `(.tool_response // .)` fallback): keeping it as-is. The suggested `(.tool_response // empty | .)` would re-open the #20 passthrough — when `.tool_response` is null/absent, `// empty` yields nothing, so the scan sees an empty list and the payload is never inspected. Verified with jq:

```text
(.tool_response // .)         on {tool_name, output:{...}}  -> ["x","SECRET"]  (scanned)
(.tool_response // empty | .) on the same                   -> []              (NOT scanned)
```

The harmless cost you flagged — `.. | strings` reaching `tool_name` on the fallback path — is intentional fail-closed-on-drift; `tool_name` values don't match any secret pattern.

#2 (no 120-char entropy catch-all in Section 4): intentional. On outbound write bodies (issue/PR prose, base64-embedded images, long IDs) a length-only rule false-positives heavily; the named-pattern set is the deliberate trade.

#4 (`request_copilot_review`): it takes owner/repo/pullNumber only — no free-text body — so no secret-egress surface; not gating it, since the broad `review`/`request` verbs would over-match read tools.

#3 (egress deny tests for connection strings / service-account JSON / AKIA / `merge`+`delete` verbs) and #5 (a `# mirrors check-output.sh` cross-reference comment on the shared regex): both valid, tracked for the follow-up that also tackles the still-open detection-quality items (#14–19/#23). Merging this as-is.

PR #4101, opened 2026-06-03
fix(claude-hooks): credential-file paths + 1Password verbs in denylists (#4091)

cbini:
## Summary & Motivation

When merged, this closes the two remaining Important denylist gaps from #4091:

- **#23 — credential files in Rule 1 (Section 1, all tools).** Added the gcloud
  application-default credentials file, the `git-credentials` file, the `netrc`
  file, and the 1Password CLI config dir (`.config/op`). These join the existing
  sensitive-path set (SSH dir, key/cert files, etc.), so reads/greps/MCP-path
  references to them are blocked.
- **#14 — 1Password CLI verbs in Rule 4 (Bash only).** Extended the escalation
  guard beyond `vault/item/read/run/document/inject` to also cover the
  credential-minting / account paths: `signin`, `account`, `user`,
  `service-account`, `connect`, and plural `items`. (`whoami` is intentionally
  left allowed — it is read-only identity info, and the suite asserts it.)
- **#5 — reciprocal cross-reference comment** in Section 4 so the egress regex
  and `check-output.sh`'s regex are kept in sync ("update BOTH if adding a
  token type").

Failing tests written first; the suite caught one over-reach (I had initially
added `whoami`, which an existing allow-test flagged — removed). Full hook suite
**8/8 green**, shellcheck clean.

Refs #4091.

## AI Assistance

Authored by Claude Code. Each gap was reproduced (red test) before the fix; the
protected `check-sensitive.sh` was applied and committed by the human.

### General

## CI checks

- [ ] Trunk — passes.

## Test evidence

`test_sensitive_paths.sh` +8 (#23: deny the four credential files via Read /
Bash / Grep; allow lookalike non-credential filenames). `test_env_protection.sh`
+11 (#14: deny the new verbs incl. existing-verb controls; allow `op --version`
and non-op commands). Full suite 8/8.

## Remaining on #4091 after this

#15 (encoding-bypass variants — deferred as the most FP-prone), #3 (egress
deny tests), and minors #27/#31/#32 — one small final batch closes the issue.

### DIFF
--- a/.claude/hooks/check-sensitive.sh
+++ b/.claude/hooks/check-sensitive.sh
@@ -94,8 +94,10 @@ fi
 # Section 1: All tools — path-based protection
 # ═══════════════════════════════════════════════════════════════════
 
-# 1. Sensitive file/directory patterns (case-insensitive)
-if echo "${no_content}" | grep -qiE '\.env[.a-z]*|(^|[ /])(\.ssh|\.kube|\.pem|\.key|\.cer|secrets\.json|credentials\.json|secret-volume)([ /]|$)|(^|[ /])(\.?/)?env(/|[ ]|$)|\.devcontainer/tpl/'; then
+# 1. Sensitive file/directory patterns (case-insensitive). Credential files added
+#    (#23): gcloud ADC application_default_credentials.json, git-credentials,
+#    netrc, and the 1Password CLI config dir (.config/op).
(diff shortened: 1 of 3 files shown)


cbini:
Both points checked — no changes needed.

**Working tree divergence:** CI-checkout artifact, not the PR. `git status --short` on the branch is empty (clean), and as you noted the committed blob has the new patterns — so `git checkout` would be a no-op. (The same phantom ` M check-sensitive.sh` has appeared in the CI review environment on the prior check-sensitive PRs too.)

**`op whoami` allow test:** already covered — `test_bypass_protection.sh:32` asserts `expect_allow "op whoami"`. That existing test is exactly what flagged an initial over-reach (I'd briefly added `whoami` to the deny verbs, then removed it), so the allow intent is documented. The review only saw the two changed test files, hence the apparent gap.

Regex review confirmed correct on all additions. Ready to merge.

PR #4105, opened 2026-06-03
fix(claude-hooks): #4091 tail — egress pattern sync, encoding variants, minors

### LINKED ISSUE #4091: fix(claude-hooks): harden check-sensitive.sh & check-output.sh — confirmed bypasses + fail-open defaults

cbini:
## Summary & Motivation

When merged, this completes the #4091 hook-hardening tail — the last cluster of
findings on the two security hooks (the sole enforcement layer in Codespaces).

- **#5 / I6** — sync the Slack, Stripe, Slack-webhook, and contextual AWS-secret
  patterns into `check-sensitive.sh` Section 4 (the outbound egress value-scan),
  so the PreToolUse scan mirrors the PostToolUse `check-output.sh` set. Closes
  the asymmetry flagged in PR #4100's review.
- **#15** — encoding-bypass variants: a decode (base64/xxd) consumed by `eval` /
  `source` / the dot-builtin through process substitution, here-strings, or
  command substitution; plus a decode-to-shell split across a newline (Rule 5
  now flattens newlines to a separator before matching).
- **#27** — Rule 7 now matches single-char uppercase variables (the quantifier
  was multi-char-only). Lowercase and loop variables stay ignored to avoid false
  positives; the last-arg special is allow-listed.
- **#31** — the PostToolUse scan gate now includes Glob output.
- **#32** — dotenv placeholder templates (example/sample/template/dist) are no
  longer over-blocked for path reads (Read/Grep/Glob/MCP); real config files
  stay blocked, and the Bash word-rule asymmetry is documented in the tests.

## AI Assistance

Authored by Claude Code. Each finding was reproduced against the live hooks
before the fix; new tests were written failing-first and validated against a
scratch copy via the suite's `HOOK`/`OUTPUT_HOOK` override before the protected
hooks were applied and committed by the human.

### General

- [ ] Review the Claude Code Review comment posted on this PR. Address valid

## CI checks

- [ ] Trunk — passes.
- [ ] dbt Cloud — not triggered (no dbt changes).
- [ ] Dagster Cloud — not triggered (no Dagster changes).

## Test evidence

Full hook suite 9/9 green against the patched hooks. New deny tests for every
#15 form and the synced #5/I6 patterns fail against the pre-patch hooks and pass
after (red to green confirmed); #27/#31/#32 covered with allow/deny controls;
trunk-clean (shellcheck, gitleaks, ignore-does-nothing).

Refs #4091.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -190,3 +190,17 @@ blocked (Rule 2), and trigger tokens placed in the command self-block. To test a
 rule, `Write` a harness into `.claude/scratch/` (Write `content` is exempt from
 scanning) that pipes fixtures into the hook by absolute path, then run
 `bash .claude/scratch/<name>.sh` (the command string carries no triggers).
+
+## Editing the hooks — recurring gotchas
+
+- **The CI `claude-review` bot recurringly reports a phantom unstaged
+  "working-tree revert"** of an edited hook (e.g. ` M check-sensitive.sh`, with
+  the new patterns "missing"). It's a CI-checkout artifact, not the PR — the
+  committed blob is correct. Confirm `git status` is clean and dismiss; do NOT
+  `git checkout` to "fix" a clean tree.
+- **trunk's shellcheck enables `SC2312` (masked-return) on pipelines**:
+  `echo`/`printf` are exempt, but `tr` / `base64` / `gunzip` / `jq` inside a
+  `$(...)` are flagged. Prefer bash parameter expansion (`${v,,}`, `${v//x/y}`,
+  `${v//$'\n'/ }`) over a `tr` subshell, or put a
+  `# trunk-ignore(shellcheck/SC2312)` on the line immediately before the
+  substitution. (Raw `shellcheck` won't show it; trunk's config does.)
--- a/.claude/hooks/check-output.sh
+++ b/.claude/hooks/check-output.sh
@@ -25,7 +25,7 @@ if [[ -z ${tool_name} ]]; then
 fi
 
 # Scan output from tools that can return sensitive content (names normalized)
-[[ ! ${tool_name} =~ ^(bash|read|grep|notebookedit|webfetch|websearch|mcp__.*)$ ]] && exit 0
+[[ ! ${tool_name} =~ ^(bash|read|grep|glob|notebookedit|webfetch|websearch|mcp__.*)$ ]] && exit 0
 
 # Extract all string values from tool_response (Claude Code's PostToolUse payload
 # key). Fall back to the whole payload when .tool_response is absent so a
--- a/.claude/hooks/check-sensitive.sh
+++ b/.claude/hooks/check-sensitive.sh
@@ -96,8 +96,13 @@ fi
 
 # 1. Sensitive file/directory patterns (case-insensitive). Credential files added
 #    (#23): gcloud ADC application_default_credentials.json, git-credentials,
-#    netrc, and the 1Password CLI config dir (.config/op).
-if echo "${no_content}" | grep -qiE '\.env[.a-z]*|(^|[ /])(\.ssh|\.kube|\.pem|\.key|\.cer|secrets\.json|credentials\.json|application_default_credentials\.json|\.git-credentials|\.netrc|secret-volume)([ /]|$)|(^|[ /])(\.?/)?env(/|[ ]|$)|\.config/op([ /]|$)|\.devcontainer/tpl/'; then
+#    netrc, and the 1Password CLI config dir (.config/op). Dotenv templates
+#    (.env.example/.sample/.template/.dist) are placeholder files, not secrets —
+#    stripped from the dotenv branch only (#32); every other pattern still scans
+#    the full corpus.
+_env_scan=$(echo "${no_content}" | sed -E 's/\.env\.(example|sample|template|dist)([^a-zA-Z]|$)/\2/g')
+if echo "${_env_scan}" | grep -qiE '\.env[.a-z]*' ||
+  echo "${no_content}" | grep -qiE '(^|[ /])(\.ssh|\.kube|\.pem|\.key|\.cer|secrets\.json|credentials\.json|application_default_credentials\.json|\.git-credentials|\.netrc|secret-volume)([ /]|$)|(^|[ /])(\.?/)?env(/|[ ]|$)|\.config/op([ /]|$)|\.devcontainer/tpl/'; then
   deny
 fi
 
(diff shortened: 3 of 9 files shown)


cbini:
## Review triage

Verified against the committed branch (head was `a7de117`, now `7383ace`). The **Critical finding and Findings 1–5 are false positives** — they are the CI-checkout phantom this PR already documents in `.claude/CLAUDE.md`: the review reasoned from a working-tree snapshot where the two hook scripts appeared unstaged, and concluded the patches were not on the branch. Both hook patches **are** committed and present in the PR diff.

Verification against the committed files:

- **Critical (hooks not committed)** — both `check-sensitive.sh` (+26/−9) and `check-output.sh` (+1/−1) are in this PR's changed-files list with their patches.
- **F1 (egress patterns)** — the Section 4 outbound value-scan regex includes the four synced shapes (Slack bot/app/user tokens, Stripe live/test keys, Slack incoming-webhook URLs, contextual AWS-secret assignment).
- **F2 (Rule 7 quantifier)** — committed regex is `\$\{?[A-Z_][A-Z0-9_]*\}?` (the `*`, so single-char uppercase like `$X` is matched); `_` was added to the safe-var list so the last-arg special is not snagged.
- **F3 (Glob gate)** — the output-scanner gate is `^(bash|read|grep|glob|notebookedit|webfetch|websearch|mcp__.*)$` — `glob` is present.
- **F4 (Rule 5 / 5a)** — Rule 5 flattens newlines to `;` before matching (catches a decode and consuming interpreter split across a newline); Rule 5a covers `eval` / `source` / the dot builtin consuming process substitution, here-strings, and command substitution.
- **F5 (dotenv carve-out)** — Rule 1 strips the `example` / `sample` / `template` / `dist` placeholder suffixes before the dotenv match; real config files still match.

All five `test_*.sh` cases the review predicted would fail pass against the committed hooks (suite is 9/9 green).

The **Minor is valid**: the garbled `Pattern 1b` comment heading in `test_sensitive_paths.sh` was a leftover find/replace artifact. Fixed in `7383ace`.

cbini:
@claude update your review

PR #4125, opened 2026-06-04
feat(cube): clarify academic-year semantics across filter surfaces

### LINKED ISSUE #4084: feat(cube): clarify academic-year semantics across filter surfaces (start-year vs SY/end-year)
Spun off from the #4010 review discussion. #4010 stays focused on the chronic-absenteeism overcounting fix; this issue captures the academic-year (AY) semantics exploration, which needs more design before implementing.

## Problem

Users refer to the same school year three different ways, and two of them anchor on **different years**:

| Phrasing | Anchors on | Example for the 2025–26 year |
|---|---|---|
| Integer `academic_year` | **start** year | `2025` |
| Full span label | start year | `2025-2026` |
| "School year" / SY notation | **end** year | `SY26` |

The collision: someone holding "SY26" maps it to `26` / `2026`, filters the numeric `academic_year` on `2026`, and silently gets the **2026–27** year instead. **This wrong-year mistake is real and observed in practice** — it is the motivation for this issue, not a hypothetical.

## What we've learned (from the #4010 investigation)

- **`academic_year` is the start year** (`2025` = the 2025–26 year), documented in the dimension description. SY notation uses the end year. The two anchors are the root cause.
- **The numeric member's failure modes are asymmetric:** a non-numeric value (`AY2025`) is rejected **loudly** by Cube's type coercion (FLOAT64 cast error). But a valid-but-wrong integer (the off-by-one) returns plausible data with **no error** — the dangerous, silent case.
- **A string label member fails the opposite way:** a malformed/aliased value (`SY26`, `26`, `2025-26`) is a valid string, matches nothing, and returns **0 rows silently** — no loud error. So switching to a label-only surface trades a loud failure for a silent one unless paired with a resolver.
- **`academic_year` was historically defined on three cubes** (`dim_terms`, `dim_dates`, `dim_student_enrollments`). #4010 consolidates the attendance views onto `dim_dates` and retires the `dim_terms` exposure (closes a latent diamond). Verified: `dim_dates.academic_year` and `dim_terms.academic_year` are identical across all attendance-fact rows, so consolidation is value-preserving.
- **Cube has no native value-allowlist / pattern validation.** Filters pass straight to SQL; there is no enum/regex guard that rejects out-of-set values. Relevant findings:
  - `suggestFilterValues: true` is a discovery **hint** for BI tools, not a constraint.
  - The supported way to enumerate values is to **query the dimension** for distinct values at runtime (data-driven, not declared).
  - `type: switch` ("predefined enumerated values") is the only declared-enum-like feature, but the docs mark it **Tesseract-engine only** — availability and exact behavior here are unconfirmed.
- **`dim_dates.is_current_academic_year` already exists** and removes year-naming entirely for relative phrasing ("this year" / "current"), which is the highest-reliability, lowest-cost lever for the most common questions.

## Proposed solutions (for refinement)

Recommended combination:

1. **Canonical filter surface = the 4-digit span label** (`2025-2026`). Keep numeric `academic_year` for sort / group / math only. (Note: 4-digit-both also makes the label sort correctly for year-over-year and removes a century-turn formatting edge.)
2. **Put the full crosswalk in every AY member description** (this reaches `/meta`, which the LLM reads at filter-choice time), and explicitly name the gotcha: "academic_year 2025 = SY26 = the 2025–26 school year. SY notation uses the END year; the integer uses the START year."
3. **Use `is_current_academic_year`** for relative phrasing so common questions never name a year.
4. **Add a thin MCP resolver** (the repurposed server-side check):
   - Deterministically translate unambiguous shorthands (`SY26`, `2025-26`, `2025–2026`) to the canonical label.
   - For the genuinely ambiguous bare integer, return a `suggested_fix` naming both readings — do not silently pick.
   - Echo the resolved interpretation back ("interpreted as the 2025–26 school year / SY26") so a wrong year is visible, not silent.
5. **Optionally expose `school_year_label` (`SY26`) as its own member** so SY notation lands directly for human Superset users instead of being guess-mapped.
6. **Set `suggestFilterValues: true`** on the label dimension and lean on dynamic value enumeration so consumers can pick from the real value list rather than constructing a string from convention.

## Open questions requiring more research

- **Does `SY26` get its own dimension, or live only in the resolver + descriptions?** Depends on how much of the audience is human Superset users (who would filter `SY26` directly) vs the MCP/LLM path (where a resolver suffices).
- **Is the Tesseract engine available (or adoptable) here?** If so, a `type: switch` dimension might give a declared, enumerated AY surface natively — verify Cube version/engine config and whether `switch` actually validates filter values or just declares the set.
- **What phrasings do users actually use most** (relative vs explicit span vs bare year vs SY)? The distribution determines how much to invest in each surface.
- **Can a numeric dimension be kept for sort/group but removed as a filter surface?** In Cube, setting `public: false` removes it entirely (filtering and grouping are not separable per-member). If numeric must stay public for grouping, the off-by-one filter surface remains and the resolver is still required.
- **Bare-integer disambiguation policy:** default to the start-year reading and echo it, or always ask? Needs a decision.
- **Does Cube Cloud's chat agent honor `suggestFilterValues`** / read it at all? (Repo notes the chat agent ignores some `meta.*` keys.)
- **Backward compatibility:** confirm no Superset dashboards or saved Cube queries reference the retired `dim_terms_academic_year` member before relying on the consolidation.

## Scope boundary

This issue is **academic-year semantics only**. The chronic-absenteeism overcounting fix (the `queryRewrite` snapshot guard and the named period-end measures) stays in #4010.

cristinabaldor:
## Summary & Motivation

Closes #4084. Fixes the silent off-by-one academic year bug where users holding "SY26" filter on `2026` and silently get 2026–27 data instead of 2025–26.

- Adds `academic_year_label` string dimension to `dim_dates` (e.g. `"2025-2026"`) as the canonical, unambiguous filter surface
- Swaps attendance views off `dim_terms_academic_year` onto `dim_dates_academic_year` / `dim_dates_academic_year_label` (consolidation deferred from #4010)
- Updates `academic_year` descriptions in `dim_terms` and `is_latest_record` in `attendance.yml` with the start-year/SY crosswalk note
- Adds `resolve_academic_year` MCP tool that translates any year phrasing (`SY26`, `AY2025`, `2025-26`, `26`, etc.) to the canonical integer + label before a Cube query is built, and updates `instructions=` to require calling it before any year-based query
- 15 new tests covering all resolver input patterns

## AI Assistance

Fully AI-assisted (Claude) with human direction on scope, design decisions (AY prefix handling, bare-integer policy, folder placement), and review.

## Self-review

_No Dagster changes._

_No dbt changes._

_No schedule/sensor/integration changes._

## CI checks


cristinabaldor:
## What changed

**`src/cube/model/cubes/conformed/dates.yml`**
- Added `academic_year_label` dimension: CONCAT-derived string (e.g. `"2025-2026"`) as the canonical, unambiguous filter surface. Integer `academic_year` retained for sort/group/math.

**`src/cube/model/cubes/conformed/terms.yml`**
- Updated `academic_year` description to include the start-year/SY crosswalk: `academic_year 2025 = the 2025-26 school year = SY26. SY notation uses the END year; this integer uses the START year.`

**`src/cube/model/cubes/attendance/attendance.yml`**
- Updated `is_latest_record` description: replaced `dim_terms.academic_year` reference with `dim_dates.academic_year` or `dim_dates.academic_year_label`.

**`src/cube/model/views/attendance/attendance_summary.yml`** and **`attendance_detail.yml`**
- Removed `academic_year` from `join_path: attendance.dim_terms` includes.
- Added `academic_year` and `academic_year_label` to `join_path: attendance.dim_dates` includes (sourced from the date dimension, not the term dimension).
- Moved `dim_dates_academic_year` and `dim_dates_academic_year_label` into the `Date` folder in `meta.folders`; removed `dim_terms_academic_year` from the `Term` folder.

**`src/cube/mcp/server.py`**
- Added `_resolve_academic_year` pure Python helper: translates any year phrasing (`SY26`, `SY2026`, `AY2025`, `AY25`, `2025-2026`, `2025-26`, `25-26`, bare `2026`, bare `26`) to canonical `academic_year` integer + `academic_year_label` string + `school_year` + `interpreted_as`. Bare integers default to start-year with a `note` field.
- Added `resolve_academic_year` async MCP tool wrapping the helper.
- Updated `instructions=` block: replaced stale `dim_terms_academic_year` reference; added REQUIRED instruction to call `resolve_academic_year` before any year-based Cube query and echo `interpreted_as` inline before results.

**`tests/cube/test_mcp_server.py`**
- 15 new tests covering all resolver input patterns including case-insensitive prefixes, en-dash variants, and `interpreted_as` format assertions.

---

## What did NOT change

- `src/cube/model/views/students/student_enrollments_summary.yml` — untouched
- `src/cube/model/views/students/student_enrollments_detail.yml` — untouched
- `src/cube/model/cubes/students/student_enrollments.yml` — untouched
- `academic_year_label` is NOT exposed on enrollment views (enrollment facts don't join through `dim_dates`; deferred to a future issue)
- No changes to any dbt models, Dagster assets, or other Cube cubes/views

cristinabaldor:
@claude update your review

cristinabaldor:
@claude update your review

cristinabaldor:
@cbini ready for you now with updated file names

cbini:
## Eval: does `resolve_academic_year` earn its keep?

I ran a model-in-the-loop eval to test, with data, whether the `resolve_academic_year` tool actually reduces wrong-year answers versus plain `instructions=` plus the new `academic_year_label` dimension. Harness is committed to this branch at `src/cube/mcp/eval/` (commit e72686c60); reproduction steps at the bottom.

### TL;DR — recommendations

1. **Keep `academic_year_label`.** It is the fix — it drove wrong-year answers to ~0 in *every* configuration, including a baseline with no resolver and no special instruction. No action needed; this is already in the PR.
2. **Move the crosswalk into `instructions=` as worked examples** (the "arm B" text below). Strengthened instructions matched or beat the resolver tool in every cell, with no extra tool round-trip.
3. **Drop `resolve_academic_year`, OR change its bare-integer behavior to surface both readings** instead of silently defaulting to start-year. On the trap case it returned the *wrong* reading and only the model's own reasoning prevented a wrong answer (details below). This is the empirical version of proposal # 4 in #4084.
4. **Optional:** if you want statistical confidence before removing the tool, re-run with more reps on the trap family (it is n=20 per cell today).

### What I ran

- 3 arms, holding the `meta` catalog + dimension descriptions constant; only the `instructions=` paragraph and the presence of the resolver tool vary:
  - **A baseline** — current instructions minus the "REQUIRED: call resolve_academic_year" paragraph; no resolver tool.
  - **B instructions** — same, plus an inline crosswalk with worked examples (model resolves the year itself).
  - **C tool** — the branch as-is (REQUIRED paragraph + resolver tool).
- 2 models (Sonnet 4.6, Haiku 4.5) x 24 prompts x 5 reps = **720 conversations**, run hermetic (only in-process mock cube tools; `load` is stubbed and the AY filter the model builds is captured). 0 errors, 0 tool leaks.
- Prompts span: clear notation (`SY26`, `2025-2026`, `2025-26`, `AY2025`), the observed trap (an SY-thinker who says "SY26 ... pull the 2026 numbers"), and an ambiguous bare integer.

### Results (correct / clarified, by prompt family)

| Model and arm | Clear-notation correct (n=80) | Trap correct (n=20) | Trap clarified (n=20) | Wrong-year, any family | Resolver called |
| haiku A baseline | 96% | 60% | 40% | 0 | n/a |
| haiku B instructions | 100% | 95% | 5% | 0 | n/a |
| haiku C tool | 100% | 75% | 25% | 0 | 100% |
| sonnet A baseline | 96% | 100% | 0% | 3/80 (clear) | n/a |
| sonnet B instructions | 100% | 100% | 0% | 0 | n/a |
| sonnet C tool | 100% | 100% | 0% | 0 | 100% |

**Silent wrong-year answers: 0 of 720.** ("Clarified" = the model recognized the conflicting prompt and asked which year was meant — a safe outcome, never wrong.)

### The key finding

On the trap, **the resolver produced the dangerous reading and the model had to override it.** All 5 Haiku arm-C "misses" were the rate prompt, where Haiku passed the literal `"2026"` (the number the prompt emphasized) to `resolve_academic_year`. The tool returned `"2026 -> SY27 (2026-2027)"` — the exact off-by-one this work exists to prevent. What avoided a wrong answer was the model distrusting the tool: *"the resolver interprets 2026 as SY27, but you said SY26 and today is June 2026, so let me clarify."* The deterministic resolver, fed a bare integer, institutionalizes the start-year guess; the model's context reasoning is what saved it.

Meanwhile the `academic_year_label` dimension is doing the real work: models filter on `academic_year_label = "2025-2026"`, which is unambiguous by construction, so wrong-year was ~0 even at baseline.

### Recommended `instructions=` crosswalk (arm B)

This text (no tool) matched or beat the resolver everywhere:

```text
ACADEMIC YEAR — resolve it yourself before building any query that names a year:
- academic_year is the START year; 'SY' notation uses the END year.
- 'SY26' -> academic_year 2025, label '2025-2026' (SY end year minus 1).
- '2025-26', '2025-2026', 'AY2025' -> academic_year 2025, label '2025-2026'.
- bare '2026' -> treat as the START year (academic_year 2026, label '2026-2027');
  if the user's wording implies SY / end-year, note the other reading.
State your interpretation inline (e.g. 'Interpreting as the 2025-2026 school year')
before showing results, then proceed.
```

### Caveats (don't over-read)

- **n=20 per trap cell.** Haiku B-beats-C on the trap (19/20 vs 15/20) has overlapping confidence intervals — directional, not conclusive. The robust claims are: wrong-year ~0 everywhere, C never beats B in any cell, and C's trap misses have a clear reproducible mechanism.
- "Clarified" outcomes are counted separately from "correct"; for the trap, clarifying is safe, so raw "correct" undercounts good behavior.
- The disambiguation metric on the ambiguous family uses lenient text matching — it shows the model states a year, not that it handles ambiguity perfectly.
- Run in a hermetic harness, not the live claude.ai connector (different host prompt, other tools). Resolver call-rate in particular could differ in the real connector.

### Reproduce

Subscription auth, no API key needed; run in a terminal where the claude CLI is logged in:

```bash
cd src/cube/mcp/eval
uv run --with claude-agent-sdk --with pyyaml python run_eval_cc.py --reps 5
```

See `src/cube/mcp/eval/README.md` for arms, metrics, and the pre-registered decision rules.

(Eval was Claude-assisted; flagging it as analysis to inform the resolver decision, not a unilateral change.)

cbini (changes requested):
see review

cbini (approved):

PR #4131, opened 2026-06-05
fix(adp): retry transient WFN 5xx instead of failing on non-JSON body

### LINKED ISSUE #4130: fix(adp): WFN workers retry swallows transient 504 as non-retryable JSONDecodeError; no alert fires
## Summary

Overnight, 20 partitions of `kipptaf/adp/workforce_now/workers` failed (asset health `DEGRADED`, 20 `FAILED` of 1995). They all came from a single schedule tick (`19088723220`, 00:30 ET) and Dagster's run-level auto-retry re-failed each one. **No Dagster+ alert fired.**

Two distinct problems:

1. A retry bug converts a transient ADP `504` into a **non-retryable** `JSONDecodeError`, so the op-level `tenacity` retry never engages.
2. None of the existing alert paths structurally catch this failure mode.

## Problem 1 — root cause of the failures

Error chain from run `e7ebef4f-0bc1-4fd5-adf0-daa51b5d0700` (partition `06/11/2026`), read bottom-up:

```text
1. requests.exceptions.HTTPError: 504 Server Error for url:
   https://api.adp.com/hr/v2/workers?asOfDate=06/11/2026&$select=...&$top=100&$skip=3800   <- real cause
2. simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)            <- 504 body is not JSON
3. requests.exceptions.JSONDecodeError  -> step "kipptaf__adp__workforce_now__workers" fails
```

The bug is in `src/teamster/libraries/adp/workforce_now/api/resources.py` (`_request`):

```python
except HTTPError as e:
    response_json = response.json()          # 504 gateway page is HTML/empty -> raises JSONDecodeError HERE
    if response.status_code == 429 or response.status_code >= 500:
        self._log.warning(msg=response_json)
        raise                                # intended "retry me" path is NEVER reached
    self._log.error(msg=response_json)
    raise AdpWorkforceNowError(response_json) from e
```

The `@retry` decorator only retries `(RequestsConnectionError, Timeout, HTTPError)`. `requests.exceptions.JSONDecodeError` is none of those (MRO: `InvalidJSONError -> RequestException -> IOError`), so tenacity re-raises it immediately with **zero retries**. The status-code check that was meant to re-raise the `HTTPError` for retry never runs because `response.json()` throws first. The retry path for 5xx/429 is effectively dead whenever the error body is not JSON — which is the normal case for gateway 502/503/504 pages.

**Why it is intermittent / ~20 of 45 partitions:** the schedule re-materializes a rolling 45-partition window every night (`partition_keys[-45:]`, `end_offset=14` in `api/schedules.py`). Each partition pulls the full ~3,800-worker roster in 100-row pages (38+ sequential calls). Whichever partitions hit a 504 on a deep page (`$skip=3800`) that night fail permanently; the rest succeed. All 20 failures fall inside that window (05/05 to 06/16); nothing older fails. Dagster run-level auto-retry fired once (`retry_number: 1`) and re-failed because it re-runs the whole asset into the same bug.

## Problem 2 — why no alert fired

The 20 failures are at the **run** level. The schedule **tick succeeded** (it launched all 45 run requests). Walking each alert type:

- **Schedule/sensor tick failure**: no — tick succeeded.
- **Asset check failure**: no — `assetChecksStatus = HEALTHY`. The `avro_schema_valid` check is yielded after the API call, so a fetch failure means the check never runs.
- **Freshness policy**: no — `freshnessStatus = NOT_APPLICABLE` on this raw asset. The only freshness policy is on downstream dbt people models (1:15am), and the current-day partition succeeded, so those models stayed fresh. The failures are on future/backfill-window partitions that do not feed today's reporting.
- **Run failure**: this is the only type that could catch it. The runs carry only `dagster/*` system tags (no custom tag), and they execute as `__anonymous_asset_job_kipptaf__adp__workforce_now__workers_schedule`.

Leading hypothesis: either there is no deployment-wide run-failure policy (the team leans on the downstream freshness policy, which this mode evades), or the run-failure policy is tag-scoped and these untagged anonymous-asset-job runs do not match. Alert policies live in the Dagster+ UI, not this repo, so the exact policy scope needs UI confirmation.

### resources.py (root cause)

- In `_request`, check the status code **before** touching `response.json()`. The 5xx/429 branch should log `response.text` and re-raise the `HTTPError` (which tenacity retries with backoff). The 4xx branch should wrap `.json()` in try/except and fall back to `response.text`. Do **not** simply add `JSONDecodeError` to the retry predicate — that would mask genuine parse failures on 200 responses.
- TDD: add a unit test that mocks a 504 whose `.json()` raises, asserting `_request` raises `HTTPError` (retryable) and that tenacity makes the expected number of attempts — not `JSONDecodeError`.

### Alerting (prevention)

- Make this asset alertable independent of UI tag scoping: e.g. a freshness check on the raw `workers` asset, and/or an alert tag on the schedule's `RunRequest`s, and/or a deployment-wide run-failure policy.
- Verify in the Dagster+ UI that the run-failure policy covers anonymous asset jobs.

### Secondary (optional, separate)

- Re-materializing 45 full-roster pulls nightly is ~1,700 API calls for this one asset and inflates the 504 surface. Worth evaluating whether the window must be that wide. Design question, not the bug.

## Evidence

- Asset: `kipptaf/adp/workforce_now/workers` — 20 FAILED / 1995, health DEGRADED.
- Tick: `19088723220` (SUCCESS) launched the runs; failures are run-level.
- Failed partitions: 05/05, 05/07, 05/10-13, 05/17-18, 05/21, 05/24, 05/28, 06/03, 06/06-08, 06/11-12, 06/14-16.
- Sample run: `e7ebef4f-0bc1-4fd5-adf0-daa51b5d0700` (504 on `$skip=3800`).

## Related

- #3762 — independently flags that the avro check is `WARN`-only and does not trigger a default Dagster+ failure alert (same asset, related alerting gap).

cbini:
## Summary & Motivation

When merged, this PR makes the ADP Workforce Now `_request` helper retry transient gateway 5xx errors (e.g. 504) instead of failing immediately.

The error handler called `response.json()` **before** checking the status code. A gateway 504 returns a non-JSON body, so `.json()` raised `JSONDecodeError` — which is not in the tenacity retry predicate `(RequestsConnectionError, Timeout, HTTPError)` — converting a retryable 504 into an immediate, unrecoverable step failure. ~20 `kipptaf/adp/workforce_now/workers` partitions failed overnight this way (deep-pagination 504 at `$skip=3800`), with no recovery.

Fix: check the status code first. 429/5xx re-raise the `HTTPError` so tenacity retries with backoff (logging `response.text`); deterministic 4xx guard the JSON parse and fall back to `response.text`. Adds 3 regression tests for non-JSON 5xx (retries, then `RetryError[HTTPError]`) and non-JSON 4xx (`AdpWorkforceNowError`, no retry) — the existing `_FakeResponse` always returned valid JSON, which is why the existing 5xx test missed this.

Closes #4130

## AI Assistance

AI-authored (Claude Code): systematic debugging from the failed run to root cause, then TDD (failing tests → minimal fix). Human-directed scope and review.

A related alerting gap was investigated and **deliberately deferred** as a rare event: there is no `JOB_FAILURE` policy, the deployment-wide `ASSET_HEALTH_DEGRADED` policy did not fire on a confirmed health transition, and freshness policies cannot catch partial-partition failures (most partitions, including the latest, succeed). Analysis recorded on #4130.

### Dagster

- [x] `uv run pytest tests/resources/test_resource_adp_workforce_now.py -k request` — 5 passed
- [ ] `uv run dagster definitions validate` — not run locally (requires 1Password-provisioned configuration unavailable to this session); the change is to a library resource method, no definition wiring changed

## CI checks

- [ ] **Trunk** — `trunk check --force` clean locally on both changed files
- [ ] **dbt Cloud** — n/a (no dbt changes)
- [ ] **Dagster Cloud** — branch deploy

### DIFF
--- a/src/teamster/libraries/adp/workforce_now/api/resources.py
+++ b/src/teamster/libraries/adp/workforce_now/api/resources.py
@@ -7,7 +7,7 @@ from pydantic import PrivateAttr
 from requests import Response
 from requests.auth import HTTPBasicAuth
 from requests.exceptions import ConnectionError as RequestsConnectionError
-from requests.exceptions import HTTPError, Timeout
+from requests.exceptions import HTTPError, JSONDecodeError, Timeout
 from requests_oauthlib import OAuth2Session
 from tenacity import (
     retry,
@@ -84,18 +84,26 @@ class AdpWorkforceNowResource(ConfigurableResource):
 
             return response
         except HTTPError as e:
-            response_json = response.json()
-
             # 429 (rate limited) and 5xx (server-side) errors are transient;
-            # re-raise the HTTPError so tenacity retries with backoff.
+            # re-raise the HTTPError so tenacity retries with backoff. Check the
+            # status code *before* parsing the body: gateway 5xx responses (e.g.
+            # a 504 HTML page) are not JSON, and JSONDecodeError is not in the
+            # retry predicate, so parsing here would convert a retryable error
+            # into a non-retryable one.
             if response.status_code == 429 or response.status_code >= 500:
-                self._log.warning(msg=response_json)
+                self._log.warning(msg=response.text)
                 raise
 
             # other 4xx are deterministic client errors — surface a specific
-            # exception (never a bare Exception) and do not retry.
-            self._log.error(msg=response_json)
-            raise AdpWorkforceNowError(response_json) from e
+            # exception (never a bare Exception) and do not retry. Guard the JSON
+            # parse: a 4xx body may also be non-JSON.
+            try:
+                detail = response.json()
+            except JSONDecodeError:
+                detail = response.text
+
+            self._log.error(msg=detail)
+            raise AdpWorkforceNowError(detail) from e
 
     def post(
         self, endpoint: str, subresource: str, verb: str, payload: dict
--- a/tests/resources/test_resource_adp_workforce_now.py
+++ b/tests/resources/test_resource_adp_workforce_now.py
@@ -8,7 +8,10 @@ from dagster import build_resources
 from requests.exceptions import HTTPError, JSONDecodeError
 from tenacity import RetryError, wait_none
 
-from teamster.libraries.adp.workforce_now.api.resources import AdpWorkforceNowResource
+from teamster.libraries.adp.workforce_now.api.resources import (
+    AdpWorkforceNowError,
+    AdpWorkforceNowResource,
+)
 
 
 def get_adp_wfn_resource() -> AdpWorkforceNowResource:
@@ -183,10 +186,6 @@ def test_request_retries_on_server_error(monkeypatch: pytest.MonkeyPatch):
 
 def test_request_does_not_retry_on_client_error(monkeypatch: pytest.MonkeyPatch):
     """A non-429 4xx is deterministic: raise a specific error without retrying."""
-    from teamster.libraries.adp.workforce_now.api.resources import (
-        AdpWorkforceNowError,
-    )
-
     monkeypatch.setattr(AdpWorkforceNowResource._request.retry, "wait", wait_none())  # pyright: ignore[reportFunctionMemberAccess]
 
     calls = {"n": 0}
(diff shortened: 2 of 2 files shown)


cbini:
Addressed the one actionable note in 55e12c3e5: `AdpWorkforceNowError` is now a module-level import, and both inline imports are removed (the new `test_request_non_json_client_error_raises_adp_error` and the pre-existing `test_request_does_not_retry_on_client_error`). Tests remain 5/5, trunk clean.

The other observations (verbose inline comment, EAFP `try/except JSONDecodeError` in the 4xx branch, bare `raise` for 5xx/429) are intentional and left unchanged.

PR #4077, opened 2026-06-01
feat(kippcamden): add u_expectations ODBC ingestion and enable dbt staging

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will enable the `U_EXPECTATIONS` PowerSchool plugin table for Camden, ingesting it via ODBC into BigQuery and including Camden in the `stg_powerschool__u_expectations` kipptaf union view alongside Newark.

Changes:
- Adds `kippcamden/powerschool/u_expectations` Dagster ODBC ingestion asset (`assets-full.yaml`, partitioned by `whenmodified`)
- Enables `src_powerschool__u_expectations` source and `stg_powerschool__u_expectations` staging model in `kippcamden/dbt_project.yml`
- Adds Camden source entry to `kipptaf/models/powerschool/sources-kippcamden.yml`
- Adds Camden to the `dbt_utils.union_relations()` call in the kipptaf union view

AI-assisted: Claude Code drafted all changes and ran parse/asset-registration checks. Human-directed: confirmed the `U_EXPECTATIONS` table was created in Camden's PowerSchool last week and that Camden should be added to the union.

### General

- [ ] Update **due date** and **assignee** on the [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
- [ ] Review the **Claude Code Review** comment posted on this PR. Address valid feedback; dismiss false positives with a brief reply explaining why.

### Dagster _(skip if no Dagster changes)_

- [ ] Run `uv run dagster definitions validate` for any modified code location
- [ ] Run `uv run pytest tests/test_dagster_definitions.py` for any modified code location
- [ ] New integrations follow the [Library + Config pattern](https://teamschools.github.io/teamster/reference/adding-an-integration/) with the correct asset key format and IO manager

### dbt _(skip if no dbt changes)_

- [ ] Include a `[model_name].yml` properties file for all new models — see [dbt Conventions](https://teamschools.github.io/teamster/reference/dbt-conventions/#model-properties-file)
- [ ] Include (or update) an [exposure](https://teamschools.github.io/teamster/reference/dbt-conventions/#exposures) for all models consumed by a dashboard, analysis, or application
- [ ] **Breaking change?** Renaming or removing columns in a contracted model (`stg_`, `rpt_`, mart) will break downstream exposures. Coordinate with affected teams before merging and document your rollback plan in the summary above.
[...]

Issue #3900, opened 2026-05-12
fix(dbt): collapse overlapping date ranges in base_powerschool__student_enrollments upstream to remove qualify tiebreakers in marts


cbini:
Refs the stale `#3633` references across multiple mart models. `#3633` was about `int_people__location_crosswalk` deduplication (completed 2026-04-22); the qualify-tiebreaker references in marts were mis-cited at it.

## Root cause

`base_powerschool__student_enrollments` (and `base_powerschool__course_enrollments`) have **genuinely overlapping date ranges** for some students at the same school/section, not just boundary off-by-ones. This is documented in `src/dbt/CLAUDE.md` → "Enrollment join fan-out (known upstream issue)":

> `base_powerschool__student_enrollments` and `base_powerschool__course_enrollments` have genuinely overlapping date ranges for some students at the same school/section (not just boundary overlaps). Half-open interval joins reduce but don't eliminate fan-out. When a date-range enrollment/CC join still produces duplicates, add a `qualify` tiebreaker picking the latest `entrydate` / `cc_dateenrolled`, with a `-- TODO: #3633` comment.

Empirical sizing (prod, 2026-05-12): 3 overlapping pairs in `int_extracts__student_enrollments` across 2 students. Small but real, and the protection has to be in place to keep PK uniqueness on consumer marts.

## Affected workarounds (`qualify row_number() = 1` tiebreakers)

- `src/dbt/kipptaf/models/marts/facts/fct_student_attendance_daily.sql:100`
- `src/dbt/kipptaf/models/marts/facts/fct_student_attendance_streaks.sql:45`
- `src/dbt/kipptaf/models/marts/facts/fct_grades_assignments.sql:127`

All currently reference closed `#3633`. The CLAUDE.md guidance text itself also cites `#3633` and should be updated to point here.

## Fix direction

Resolve the overlapping date ranges in the source data. Two viable paths:

1. **Upstream fix in PowerSchool data**: identify the rows producing overlaps (data-quality issue in source) and remediate via PowerSchool data fix or a deterministic dedup at the staging layer.
2. **Intermediate-layer collapse**: build `int_powerschool__student_enrollments_canonical` (or similar) at one row per (student, school, day) by deterministically picking the active enrollment when overlaps exist. Downstream marts source from the canonical model, no qualify needed.

## Scope

- All three mart `qualify` tiebreakers above.
- The CLAUDE.md "Enrollment join fan-out" reference (update to this issue once filed).
[...]

PR #4132, opened 2026-06-05
feat(dbt): rebuild gradebook audit pipeline for AY 2026-2027 (teacher/student split)

### LINKED ISSUE #3908: docs(dbt): document gradebook audit data model and plan AY 2026–2027 changes
# Gradebook Audit — AY 2026-2027 (shipped state)

> The AY 2026-2027 rebuild is delivered. The living reference for the current model is [`docs/reference/gradebook-audit-data-model.md`](https://github.com/TEAMSchools/teamster/blob/main/docs/reference/gradebook-audit-data-model.md); the design spec and implementation plan live under `docs/superpowers/`. This issue's original May–June 2026 design writeup is preserved as a historical record at the bottom.

## What shipped

- **Quarter-grain pipeline** — moved from week-grain to one row per section × quarter × category. The week time spine (`int_powerschool__calendar_week`) is gone, full Q1–Q4 coverage is restored, and EOQ flags fire year-round.
- **Teacher/student split (July 2026)** — the single week-grain model was replaced by two terminal reports that read one shared intermediate:
  - `rpt_tableau__gradebook_audit` — teacher/section-facing, **zero student PII**. Row types `category_summary` (exactly 4 per section × quarter) and `assignment_detail`, plus two health booleans (`is_healthy_gradebook_all_flags` / `is_healthy_gradebook_excl_comments`).
  - `rpt_gsheets__gradebook_audit_student_flags` — student-level flags (grade above 100; grade below 70 with a missing EOQ comment) to a Google Sheet for ops follow-up.
  - `int_extracts__gradebook_audit_student_flags` — the shared intermediate both reports read (removed an earlier `rpt_`-reading-`rpt_` layering violation).
  - The interim `int_tableau__gradebook_audit_flags_calculations` model was deleted.
- **PS-native expectations** — `int_powerschool__u_expectations_qtd_unpivot` (reads the `U_EXPECTATIONS` PowerSchool plugin) replaced `stg_google_sheets__gradebook_expectations_assignments` (disabled).
- **Exceptions mechanism removed** — every join to `stg_google_sheets__gradebook_exceptions` was dropped; that sheet and the flag allowlist (`stg_google_sheets__gradebook_flags`) are disabled. Flags are now hardcoded boolean columns, not sheet-driven config.
- **Coverage** — Miami removed (moved to Focus gradebook); Paterson MS added, with expectations spoofed from Newark MS (a `paterson_spoof` CTE) until Paterson gets its own plugin.
- **Flag changes** — FYI flags and the Summative-200 point-value flags removed; `qt_student_is_ada_80_plus_gpa_less_2` migrated off the audit to `int_extracts__student_enrollments` and `rpt_tableau__gradebook_gpa`.

## Coverage (AY 2026-2027)

| Region   | School level | Coverage                                        |
| -------- | ------------ | ----------------------------------------------- |
| Camden   | MS, HS       | Full audit                                      |
| Newark   | MS, HS       | Full audit                                      |
| Paterson | MS           | Full audit (expectations spoofed from Newark MS) |
| Camden   | ES           | EOQ comments only                               |
| Newark   | ES           | EOQ comments only                               |
| Paterson | ES           | EOQ comments only                               |
| Miami    | n/a          | Removed — moved to Focus                        |

## Remaining work

- [ ] **Install the gradebook-audit PS plugin on Paterson's PowerSchool instance** so Paterson MS uses its own `U_EXPECTATIONS` instead of the Newark spoof, then remove the `paterson_spoof` CTE.
- [ ] **Make student-course selection deterministic when enrollment dates overlap** — `int_extracts__course_enrollments_by_term`'s `row_number()` tiebreak is frequently a true tie when PowerSchool holds overlapping course-enrollment records (#3900 double-writes), so the section/teacher in scope for a student is arbitrary. Add a deterministic tiebreak, or fix the upstream double-writes (#3900).
- [ ] **Re-derive or remove the `section_quarter_count >= 2` filter** in `int_extracts__course_schedule_by_term` — it silently excludes single-quarter sections with no recoverable rationale.
- [ ] **Tableau workbook (workbook-side, not dbt):** school-level summary view (per school, share of classrooms with an active flag) and click-through from that summary to a filtered teacher list. The dbt side (manager columns) is shipped.

---

<details>
<summary>Historical record — original May–June 2026 design and scope (superseded by the shipped model above)</summary>

## Background

The gradebook audit Tableau dashboard helps school leaders and instructional coaches audit teacher gradebooks for compliance with KIPP TAF grading policies. Teachers are expected to maintain gradebooks according to policy — correct assignment categories, appropriate point values, timely grading, valid scores — and the dashboard surfaces deviations as flags for review each week.

The dashboard is powered by `rpt_tableau__gradebook_audit` and has been in production for approximately two years. It covers the current academic year only and is organized by calendar week within each quarter. A full lineage diagram and model-by-model reference lives in [`docs/reference/gradebook-audit-data-model.md`](https://github.com/TEAMSchools/teamster/blob/main/docs/reference/gradebook-audit-data-model.md) (added in this issue).

### Dashboard coverage (AY 2025-2026)

| Region | School level | Coverage |
|---|---|---|
| Camden | MS, HS | Full audit (all applicable flags) |
| Camden | ES | EOQ comments only — ES schools do not enter assignments in PS gradebook |
| Newark | MS, HS | Full audit (all applicable flags) |
| Newark | ES | EOQ comments only — ES schools do not enter assignments in PS gradebook |
| Miami | ES, MS | Full audit — **removing AY 2026-2027** (moving to Focus) |
| Paterson | n/a | Not on dashboard — **adding AY 2026-2027** |

### Configuration layer

Three Google Sheets staging models drive all flag behavior:

**`stg_google_sheets__gradebook_flags`** — the flag allowlist. Defines which `audit_flag_name` values are active per `region`, `school_level`, and `code` (quarter code or gradebook category code). Flags are only surfaced in the dashboard if a matching row exists here. Deactivating a flag means disabling or removing its row from this sheet.

**`stg_google_sheets__gradebook_expectations_assignments`** — defines the expected assignment count per category (W, H, F, S) per `region` / `school_level` / `quarter` / `week_number`. Source sheet uses a wide format (one column per category); staging unpivots to one row per category with a computed `assignment_category_term` (e.g. `W3` for Work Habits in Q3).

**`stg_google_sheets__gradebook_exceptions`** — a multi-key suppression table. Each row suppresses a flag or set of rows from a specific model CTE, identified by `view_name` and `cte`. Suppressions can be permanent (no `is_quarter_end_date_range` scoping) or temporary (only during non-EOQ or EOQ periods). Scope fields: `region`, `school_level`, `school_id`, `course_number`, `credit_type`, `gradebook_category`, `audit_flag_name`, `grade_level`. Used in 15+ LEFT JOINs across five intermediate models — the most operationally complex part of the data model.

### Scaffold layer

**`int_tableau__gradebook_audit_teacher_scaffold`** — the time spine. Joins active sections to term weeks (quarters x calendar weeks) for the current academic year only. Produces two scaffold variants tagged by `scaffold_name`:

- `teacher_scaffold` — one row per section x week (no category)
- `teacher_category_scaffold` — one row per section x week x category, enriched with `expectation` (required assignment count) from `stg_google_sheets__gradebook_expectations_assignments`

`is_quarter_end_date_range` is computed here: a date window around the last in-session day of each quarter during which EOQ-only flags are active. The window varies by region and school level:

| Context | Window |
|---|---|
| Miami | 9 days before to 28 days after `quarter_end_date_insession` |
| HS Q3 | 9 days after to 20 days after `quarter_end_date_insession` (extended window) |
| All others | 5 days before to 14 days after `quarter_end_date_insession` |

Currently excludes Q1 and Q2 (see [Recent change](#recent-change-q1q2-removal) below).

**`int_tableau__gradebook_audit_student_scaffold`** — adds enrolled students to the teacher scaffold. Joins `int_extracts__student_enrollments` to `base_powerschool__course_enrollments` and `int_tableau__gradebook_audit_teacher_scaffold`. Pulls quarter course grades from `base_powerschool__final_grades` (current year) and `stg_powerschool__storedgrades` (prior year Q). Produces two scaffold variants:

- `student_scaffold` — computes EOQ flags directly as boolean columns: conduct code missing/incorrect, comment missing, grade > 100, ADA >= 80% + GPA < 2.0
- `student_category_scaffold` — one row per student x category; computes `w_grade_inflation` (student's W% deviates >= 30 points from the class-wide W average), `qt_effort_grade_missing`, `qt_formative_grade_missing`, `qt_summative_grade_missing`

### Assignment and category rollup layer

**`int_tableau__gradebook_audit_assignments_student`** — one row per student x assignment x week. Joins the student scaffold to `int_powerschool__gradebook_assignments_scores` (by student + section + category + due date). Computes per-student per-assignment flags: null score, score above max, wrong score for missing assignments (ES/MS: should be 5; HS: should be 0), S score below 50%, score not on the school-level conversion chart.

**`int_tableau__gradebook_audit_assignments_teacher`** — one row per section x assignment x week. Joins the teacher scaffold to `int_powerschool__gradebook_assignments` (by `duedate` within the week window) and `int_powerschool__gradebook_assignments_scores` (rollup). Computes class-level max-score flags (W/H/F must be 10 pts; Miami S must be <= 100 pts), running assignment counts by category, and section-level score-count aggregates (`n_students`, `n_late`, `n_missing`, `n_null`, etc.).

**`int_tableau__gradebook_audit_categories_teacher`** — one row per section x category x week. Computes: expected assignment count not met (`teacher_running_total_assign_by_cat < expectation`), percent graded < 70% (`n_expected_scored / n_expected`), S total point value out of range (MiamiES: target 100; all others: target 200).

### Flag assembly: `int_tableau__gradebook_audit_flags`

Six-branch `UNION ALL` that unpivots each source model's boolean flag columns into `audit_flag_name / audit_flag_value` rows:

| CTE | Source | Flags |
|---|---|---|
| `student_unpivot` | `int_tableau__gradebook_audit_assignments_student` | 16 assignment-student flags |
| `teacher_unpivot_cca` | `int_tableau__gradebook_audit_assignments_teacher` | 4 class-category-assignment flags |
| `teacher_unpivot_cc` | `int_tableau__gradebook_audit_categories_teacher` | 12 class-category flags |
| `eoq_items` | `int_tableau__gradebook_audit_student_scaffold` | 7 EOQ student flags (non-conduct) |
| `eoq_items_conduct_code` | `int_tableau__gradebook_audit_student_scaffold` | 5 conduct code flags (grade-level scoped) |
| `student_course_category` | `int_tableau__gradebook_audit_student_scaffold` | 4 student-category flags |

Each CTE INNER JOINs to `stg_google_sheets__gradebook_flags` on the relevant keys — only rows with a matching flag config row are kept. Each CTE also LEFT JOINs to `stg_google_sheets__gradebook_exceptions` (one or more joins per CTE) and filters to `include_row is null` to apply suppressions.

### Final extract: `rpt_tableau__gradebook_audit`

Two CTEs (`teacher_aggs` and `valid_flags`) from `int_tableau__gradebook_audit_flags`, combined in a 5-branch `UNION ALL`. All branches filter `audit_start_date <= current_date` and `not is_current_week` — the current week is excluded because grading is still in progress.

## Recent change: Q1/Q2 removal

In May 2026, Q1 and Q2 were removed from `int_tableau__gradebook_audit_teacher_scaffold` (`t.term not in ('Q1', 'Q2')`) to address Tableau Server refresh failures caused by row volume. The audit now covers **Q3 and Q4 only**. This was a direct response to query size limits on the Tableau side, not a policy change.

## Pain points

- **Volume** — covering only the current year and only Q3+Q4, the model still pushes against Tableau Server limits. Every new flag or branch added since launch has compounded the row count with no offsetting reduction. Full-year coverage (Q1-Q4) is not currently feasible without a structural change.
- **Exceptions complexity** — `stg_google_sheets__gradebook_exceptions` is called in 15+ LEFT JOINs across five intermediate models with different `view_name` / `cte` keys at each site. Adding or debugging an exception requires knowing the exact call-site identifier, which is not surfaced in the sheet itself.
- **Multi-grain UNION** — the final model merges rows at different grains (student x assignment, student x category, section x category) into one output schema with many `null` placeholders per branch. This makes the Tableau data source large and complicates field reuse across views.
- **Flag management via sheet** — activating or deactivating a flag requires coordinated edits to `stg_google_sheets__gradebook_flags` and (often) `stg_google_sheets__gradebook_exceptions`, with no validation that the `audit_flag_name` values in the sheet match what the SQL actually produces.

### Documentation
- [x] Write `docs/reference/gradebook-audit-data-model.md`
- [x] Add nav entry in `mkdocs.yml`
- [x] Create `gradebook-audit` Claude Code skill at `.claude/skills/gradebook-audit/` — a domain expert skill for anyone working with this data model. Covers explaining the model, annual flags rollover (with pre-built query), adding/removing flags, adding regions, and debugging. Shared across all team members who use Claude Code in this repo.

### AY 2026-2027: Add Paterson

Paterson MS joins the dashboard for the first time (Paterson has no HS schools).

- [x] Add Paterson rows to `stg_google_sheets__gradebook_flags` (confirm which flags apply with T&L)
- [ ] Add Paterson rows to `stg_google_sheets__gradebook_expectations_assignments` (expected counts per category / quarter / week) — or confirm covered by PS-native source if that migration lands first
- [ ] Verify teacher scaffold picks up Paterson sections correctly (no SQL change expected, but confirm)

### AY 2026-2027: Remove Miami ES and MS

Miami ES and MS are migrating to Focus gradebook.

- [x] Remove Miami rows from `stg_google_sheets__gradebook_flags`
- [x] Remove Miami rows from `stg_google_sheets__gradebook_expectations_assignments` — Miami is moving to Focus gradebook and will not receive the U_EXPECTATIONS PS plugin. No rows were added for AY 2026 and none will be added going forward.
- [x] Audit and remove Miami-only SQL branches — resolved in the AY 2026-2027 pipeline rebuild. The new `int_tableau__gradebook_audit_flags_calculations` has no Miami-specific flag branches; Miami is excluded entirely via `_dbt_source_project != 'kippmiami'` in all scaffold branches.

### AY 2026-2027: Other
- [x] Restore full academic year coverage (Q1-Q4) — resolved as a natural consequence of the quarter-grain scaffold refactor (see Design decisions below). The `term not in ('Q1', 'Q2')` filter is removed when the `term_weeks` CTE is eliminated. No separate structural solution required.

### AY 2026-2027: Flag removals

- [x] **Remove FYI Flags from the audit model** — `w_grade_inflation`, `assign_s_hs_score_not_conversion_chart_options`, and `assign_s_ms_score_not_conversion_chart_options` removed. Resolved in the AY 2026-2027 pipeline rebuild: the new `int_tableau__gradebook_audit_flags_calculations` does not include these flags.
- [x] **Migrate `qt_student_is_ada_80_plus_gpa_less_2`** — removed from the gradebook audit and split into two new booleans: `is_ada_above_or_at_80_cum_gpa_less_2` on `int_extracts__student_enrollments` (cumulative GPA, student grain) and `is_ada_above_or_at_80_gpa_y1_less_2` on `rpt_tableau__gradebook_gpa` (year-to-date per-course GPA grain). Also removed the `region != 'Paterson'` filter from `rpt_tableau__gradebook_gpa` to add Paterson to that view.
- [x] **Remove Summative 200 point-value flags** — `qt_teacher_s_total_greater_200` and `qt_teacher_s_total_less_200` removed. **Reason:** the makeup work policy means teachers legitimately exceed or fall short of the 200-point target without it indicating a compliance problem — false positives, not actionable. Resolved in the AY 2026-2027 pipeline rebuild: the new pipeline does not include these flags.

### AY 2026-2027: Logic changes

- [x] **QTD cumulative assignment count** — change `w/h/f/s_expected_assign_count_not_met` from a weekly per-category check to a quarter-to-date cumulative check. **Format question resolved:** QTD expectations come from the PS-native intermediate model (`int_powerschool__u_expectations_unpivot`), not the Google Sheet. At quarter grain, cumulative counts are natural — every row already represents the full quarter-to-date.
- [x] ~~**7-day grace period for percent-graded flags**~~ — **dropped as policy (July 2026); never implemented.** T&L officially dropped the grace period. `is_expected` in `int_powerschool__gradebook_assignments_scores` carries no due-date condition — an assignment counts as expected the moment it is due (see the code comment there and the design spec's grace-period section). `percent_graded_min_not_met` shipped at per-assignment grain (threshold 0.90) in `int_powerschool__gradebook_assignment_scores_rollup` as one input to `assignment_has_flags`, without the grace condition.
- [x] **EOQ flags fire year-round** — `is_quarter_end_date_range` is removed entirely from the scaffold and flag conditions. At quarter grain, EOQ conditions (missing comments, missing conduct codes) are either true or false for the quarter regardless of calendar date. T&L approved. This supersedes the original "move suppression to Tableau" approach.

### AY 2026-2027: Output redesign

- [x] **Remove exceptions mechanism** — T&L has decided to eliminate the exceptions mechanism. Removed all LEFT JOINs to `stg_google_sheets__gradebook_exceptions` across five intermediate models and all `WHERE e.include_row IS NULL` filters. The staging model is **disabled** (`config: enabled: false`), not deleted, pending operational decisions after July 1, 2026. This is the largest single simplification in the AY 2026-2027 scope.
- [x] **Simplify the "in the clear" row pattern** — implemented as a three-branch UNION in `rpt_tableau__gradebook_audit` (fully overwritten; no parallel `_v4` model shipped): Branch 1 = one 'No Flags' anchor row per section x quarter for healthy teachers; Branch 2 = `expected_assign_count_not_met` rows per assignment in a short category (or one null-assignmentid row for zero-entry categories); Branch 3 = the two unpivoted `student_course` flags. `is_healthy_gradebook` is computed in the `health_calc` CTE (`not max(audit_flag_value)` grouped by teacher x school x quarter — true means no flag fired). `int_tableau__gradebook_audit_scaffold_unpivot` was deleted.

### AY 2026-2027: Tableau workbook requirements

User feedback collected from school leaders:

- [ ] **School-level summary view** — a view showing, per school, the number or percentage of classrooms with at least one active flag. Requires the anchor-row pattern above so that total classroom count (denominator) is available even for fully compliant teachers.
- [ ] **Click-through to teacher list** — from the school-level summary, school leaders want to click through to a filtered list of teachers with active flags so they can have direct coaching conversations. Tableau action filter from school summary → teacher detail. The dbt side of this requirement is satisfied by the manager columns (`manager_employee_number`, `manager_name`, `manager_tableau_username`) being added to the teacher scaffold in this PR. The Tableau action filter itself is a workbook change handled separately.

### PS-native expectations source (promoted from exploration to in-scope)

A pilot was completed to move `stg_google_sheets__gradebook_expectations_assignments` out of Google Sheets and into PowerSchool via a custom PS plugin. The staging model `stg_powerschool__u_expectations` already exists in the `powerschool` dbt project (sourced from the `U_EXPECTATIONS` PS extension table via ODBC).

Current schema of `stg_powerschool__u_expectations`:

| Column | Type | Notes |
|---|---|---|
| `id` | int64 | Plugin-assigned PK |
| `school_level` | string | ES / MS / HS |
| `quarter` | string | 1-4 |
| `week_number` | int64 | Week within quarter |
| `cnt_w` / `cnt_h` / `cnt_f` / `cnt_s` | int64 | Expected count per category (wide format) |
| `notes` | string | Free-text |

**Schema gaps vs. the current Google Sheet source** that must be resolved before this can replace it:

- No `academic_year` column — source is always current year; the kipptaf model would need to inject `current_academic_year` at join time
- No `region` column — currently Newark-only; Camden and Paterson would need the PS plugin deployed to their instances
- Wide format (`cnt_w/h/f/s`) — needs unpivoting to match the `region / school_level / quarter / week_number / assignment_category_code / expectation` grain used by the teacher scaffold join

**Status update:** Camden has the U_EXPECTATIONS plugin deployed (PR #4077 pending full integration by Bini). Newark data already exists in `stg_powerschool__u_expectations`. The migration is now in scope for this PR at Newark-only coverage, with Camden and Paterson added as their deployments land. `stg_google_sheets__gradebook_expectations_assignments` is deprecated and **disabled** (`config: enabled: false`), not deleted.

- [x] Camden plugin deployed (PR #4077 for full integration)
- [x] Write `int_powerschool__u_expectations_unpivot` — intermediate model unioning regions, injecting `academic_year`; name includes `_unpivot` because the model performs UNPIVOT internally (Charlie's naming convention)
- [x] Update `int_tableau__gradebook_audit_teacher_scaffold` to join the new model (Newark only at launch) — the AY 2026-2027 rebuild replaced `int_tableau__gradebook_audit_teacher_scaffold` with `int_tableau__gradebook_audit_flags_calculations`, which already joins `int_powerschool__u_expectations_qtd_unpivot` for all active regions.
- [x] Add Camden to `int_powerschool__u_expectations_unpivot` once PR #4077 lands
- [ ] Add Paterson to `int_powerschool__u_expectations_unpivot` once PS instance access is available
- [x] Disable Google Sheet source (not delete, pending July 1 ops decisions)

## Design decisions (added during design phase, May-June 2026)

The following decisions were made during the design phase and expand on or supersede the original scope. Documented here for future maintainers.

### Quarter-grain scaffold refactor

**What:** The entire audit pipeline moves from week-grain to quarter-grain. The teacher scaffold currently generates one row per section x week x category. The new design generates one row per section x quarter x category.

**Why:** Week-grain was the root cause of the volume problem documented in Pain points above. Moving to quarter grain eliminates week-level row multiplication entirely, restores Q1/Q2 coverage at no volume cost, and allows the scaffold to collapse from four CTEs (`sections`, `term_weeks`, `school_level_mod`, `final`) to a single `sections` CTE plus a direct two-branch UNION ALL SELECT. `int_powerschool__calendar_week` is no longer a dependency of the audit scaffold.

**Consequences:** All downstream models that join on week date windows must update to quarter date windows. Assignment counts become quarter-to-date naturally. EOQ flags fire year-round (see below). New columns added to scaffold output: manager employee number, manager name, manager Tableau username (sourced from `int_people__staff_roster`). `school_abbreviation` and `school_level` added to `base_powerschool__sections` as a prerequisite.

### EOQ flags fire year-round

**What:** `is_quarter_end_date_range` removed. EOQ flags fire unconditionally.

**Why:** At quarter grain, each row represents a full quarter. The concept of "weeks before EOQ" no longer applies. T&L approved.

### Summative 200 flags — false positives

**Why removed:** The makeup work policy means teachers legitimately exceed or fall short of the 200-point Summative target without it indicating a compliance problem. Documented here so the flags are not re-added in a future year without revisiting this reasoning.

### Exceptions mechanism — disable, not delete

**Why:** Pending operational decisions after July 1, 2026. Models are disabled (`config: enabled: false`) rather than deleted in case those decisions affect the approach.

## Open questions

**For T&L:**

- **ADA + GPA flag threshold** — the current flag fires when a student's GPA is strictly below 2.0. A student with exactly 2.0 does not currently get flagged. Should students with exactly 2.0 also be flagged? This applies to both the cumulative version going to `int_extracts__student_enrollments` and the per-course version going to `rpt_tableau__gradebook_gpa`.
- **HS ADA weighting for AY 2026-2027 and beyond** — `ada_above_or_at_80` and the new `is_ada_above_or_at_80_cum_gpa_less_2` both use `academic_year >= 2025` to switch HS from unweighted to weighted ADA. This condition applies permanently to all future years. Was weighted ADA intended as a one-year pilot for AY 2025-2026 only, or is it permanent going forward? If permanent, no code change is needed.

</details>

GabyRangelB:
## Summary & Motivation

The full AY 2026-2027 rebuild of the gradebook audit pipeline — moving it from week-grain to quarter-grain, splitting the single Tableau model into a teacher-facing report and a student-flags Google Sheet over a shared intermediate, replacing the Google-Sheet flag/expectation config with a PowerSchool-native source, and removing the exceptions mechanism.

The living reference for the resulting model is [`docs/reference/gradebook-audit-data-model.md`](https://github.com/TEAMSchools/teamster/blob/main/docs/reference/gradebook-audit-data-model.md); the design spec and implementation plan are under `docs/superpowers/`. This PR's iteration-by-iteration history — which describes several intermediate architectures that were later superseded — is preserved in the collapsed **Development history** below.

### What this PR delivers

- **Quarter-grain pipeline** — one row per section × quarter × category (was section × week × category). The week time spine (`int_powerschool__calendar_week`) is gone, Q1–Q4 coverage is restored, and EOQ flags fire year-round.
- **Teacher/student split** — the audit is now two terminal reports over one shared intermediate:
  - `rpt_tableau__gradebook_audit` — teacher/section-facing, **zero student PII**. `row_type` = `category_summary` (exactly 4 per section × quarter, built as a grain-projection `SELECT DISTINCT`) or `assignment_detail`, plus two health booleans (`is_healthy_gradebook_all_flags` / `is_healthy_gradebook_excl_comments`).
  - `rpt_gsheets__gradebook_audit_student_flags` — student-level flags (grade above 100; grade below 70 with a missing EOQ comment) to a Google Sheet for ops follow-up.
  - `int_extracts__gradebook_audit_student_flags` — the shared intermediate both reports read (no `rpt_`-reading-`rpt_`).
  - The eight v3 `int_tableau__gradebook_audit_*` models and the interim `int_tableau__gradebook_audit_flags_calculations` are disabled/deleted.
- **New scaffolds** — `int_extracts__course_schedule_by_term`, `int_extracts__course_enrollments_by_term`, `int_powerschool__gradebook_assignment_scores_rollup`.
- **PS-native expectations** — `int_powerschool__u_expectations_qtd_unpivot` (reads the `U_EXPECTATIONS` PS plugin) replaces `stg_google_sheets__gradebook_expectations_assignments` (disabled). Paterson can't host the plugin, so Paterson MS expectations are spoofed from Newark MS via a `paterson_spoof` CTE in `stg_powerschool__u_expectations`.
- **Exceptions mechanism removed** — all joins dropped; the exceptions and flag-allowlist sheets (and their sources) are disabled. Flags are hardcoded boolean columns, not sheet-driven config.
- **Coverage** — Miami removed (moved to Focus); Paterson MS added. Flag removals: FYI flags, Summative-200; `qt_student_is_ada_80_plus_gpa_less_2` migrated off the audit to `int_extracts__student_enrollments` and `rpt_tableau__gradebook_gpa`.
- **Docs & tooling** — reference doc, the `gradebook-audit` skill, design spec, implementation plan, and an `mkdocs.yml` nav entry.

### Breaking change

`rpt_tableau__gradebook_audit` (contract-enforced) is **fully replaced** with a new column set — a `row_type`-based `category_summary` / `assignment_detail` shape, two health booleans, and manager columns; all `_week_` columns (`audit_qt_week_number`, `is_quarter_end_date_range`, `is_current_week`, etc.) are gone. **The Tableau workbook must be rebuilt against the new schema post-merge.**

### Remaining before merge

- **CODEOWNERS review** from analytics-engineers on `src/dbt/` (required approval; CI is green).

Refs #3908

---

<details>
<summary>Development history — superseded intermediate architectures and per-round review notes</summary>

This PR implements the full AY 2026-2027 revamp of the gradebook audit pipeline: annual config rollover, PS-native expectations model, quarter-grain scaffold redesign, and flag cleanup.

**Scope correction (post-review):** the eight `int_tableau__gradebook_audit_*` v3 models (`teacher_scaffold`, `student_scaffold`, `assignments_teacher`, `assignments_student`, `categories_teacher`, `flags`) are **disabled** (`config: enabled: false`), not rewritten in place — a new pipeline (`int_tableau__gradebook_audit_flags_calculations` → `rpt_tableau__gradebook_audit`) ships alongside them. MS has now signed off on the same audit model as HS, so `rpt_tableau__gradebook_audit` is fully overwritten by the new pipeline rather than left as a parallel `_v4` model — there is no separate v3 report model left to retire.

**Docs and tooling:**
- Adds `docs/reference/gradebook-audit-data-model.md` — full lineage, flag inventory, and model-by-model reference for the gradebook audit pipeline
- Adds `.claude/skills/gradebook-audit/` — domain skill covering annual rollover, flag management, region additions, and debugging procedures
- Adds `docs/superpowers/plans/2026-05-14-gradebook-audit-ay2627-revamp.md` — complete implementation plan for the AY 2026-2027 revamp
- Adds nav entry in `mkdocs.yml` for the reference doc

**Exceptions mechanism removed (Task 2):**
- `stg_google_sheets__gradebook_exceptions`: disabled (`config: enabled: false`)
- All 15+ exception LEFT JOINs removed from intermediate models as part of Task 6 rewrites

**PS-native expectations (Task 3):**
- Creates `int_powerschool__u_expectations_qtd_unpivot` — Newark and Camden covered (PR #4077 merged). Paterson: the **KIPP NJ Gradebook Audit** PS plugin cannot be installed on the Paterson PowerSchool instance — this is not a pending deploy, it's a hard blocker. Plan is to spoof Paterson expectations from Newark's data; not yet implemented (tracked separately from this PR)

**Prerequisite model updates (Tasks 4–5):**
- `stg_google_sheets__gradebook_flags`: remove `alt_code` CASE; drop `grade_level` column (Miami-only) — the model (and, per the code-review fixes below, its source) is disabled, so this only matters if a future year re-enables it
- `rpt_tableau__gradebook_gpa`: add `is_ada_above_or_at_80_gpa_y1_less_2`; remove Paterson exclusion

**Quarter-grain scaffold refactor (Task 6):**
The audit pipeline moves from week-grain to quarter-grain. The eight `int_tableau__gradebook_audit_*` v3 models above are disabled; their function is now covered by:
- `int_extracts__course_schedule_by_term` (new): one row per teacher x section x quarter
- `int_extracts__course_enrollments_by_term` (new): one row per student x academic year x course x quarter; enrollment stints deduped to latest exitdate
- `int_powerschool__gradebook_assignment_scores_rollup` (new): per-assignment score-flag rollup, replacing the inline CTE previously in `assignments_teacher`/`categories_teacher`
- `int_tableau__gradebook_audit_flags_calculations` (new, replaces `int_tableau__gradebook_audit_flags`): three-branch `UNION ALL` (`sections_teacher`, `assignment_teacher`, `student_course`); summer-toggle markers at all year-filter lines and the `grades_type` filter
- `rpt_tableau__gradebook_audit` (fully rewritten): three-branch `UNION ALL` over `count_not_met_flag` → `health_calc` (computes `is_healthy_gradebook` as a `GROUP BY` aggregate — `not max(audit_flag_value)`, true when **no** flag fired) → `flags_unpivot` (unpivots the two `student_course` flags for Branch 3's output rows)
- `int_tableau__gradebook_audit_scaffold_unpivot` — never shipped; an interim model from this PR's own iteration, removed before merge
- `rpt_tableau__gradebook_es_comments`: complete rewrite as a standalone model, sourced from `int_extracts__course_enrollments_by_term`; ES removed from the main flags pipeline entirely
- `rpt_tableau__assignment_checks`: deprecated (superseded by the quarter-grain pipeline)

**Code-review fixes (this round):**
- `int_powerschool__gradebook_assignments_scores`: fixed a `BETWEEN` date-range join to the half-open form; added a missing uniqueness test; corrected the `is_expected` description (no due-date grace period is implemented — the 7-day grace scoped in the design spec was not built)
- `int_extracts__course_enrollments_by_term`: fixed a grain defect — the dedup partition and uniqueness key omitted `academic_year`, so the same `course_number` across two academic years could collide
- `rpt_tableau__gradebook_es_comments`: added a missing uniqueness test and full column descriptions
- `rpt_tableau__gradebook_gpa`: added a uniqueness test (warn-severity; residual violations trace to the known `base_powerschool__course_enrollments` double-write duplicates, #3900)
- Fixed several inaccurate YAML descriptions (`int_extracts__course_schedule_by_term.school_level_alt`, `int_extracts__course_enrollments_by_term.rn`) and added missing descriptions (`int_powerschool__gradebook_assignment_scores_rollup`)
- Renamed `int_powerschool__u_expectations_qtd.yml` to match its model file (`..._unpivot`)
- Removed a logically-redundant join predicate in `rpt_tableau__gradebook_audit`'s Branch 2
- Corrected the reference doc, design spec, and `SKILL.md`, which inverted the `is_healthy_gradebook` formula and branch count — all three described a 2-branch `UNION ALL` with `is_healthy_gradebook = max(audit_flag_value)` (true = at least one flag fired); the actual pipeline is 3-branch with `not max(...)` (true = no flag fired). Also fixed `SKILL.md`'s relative doc links (were one directory short) and its incorrect claim that ES students appear in student-course EOQ rows (ES is excluded from all three branches)
- Disabled the `src_google_sheets__gradebook_flags` and `src_google_sheets__gradebook_exceptions` sources (`sources-external.yml`), closing out an unaddressed review comment asking for both — with MS now on the same model as HS, nothing reads either sheet; the earlier Task 1 sheet-edit prep (AY rows, Paterson rows, Miami removal) is moot, so that paragraph was dropped from this description. Dagster no longer pulls either sheet into BigQuery

**Second review pass — `expected_assign_count_not_met` was checking the wrong thing:**
- The flag computed `total_assign_count_qtd_by_cat_section_actual != ..._no_flags` (fires only when at least one flagged assignment exists) and never compared against `expectation` at all, despite the flag's name and the carried `expectation` column implying a count check. Concretely: a category with zero assignments entered, or with fewer clean assignments than expected, produced `actual == no_flags` and passed silently — verified 4,165 zero-entry section-categories were invisible under the old logic
- Corrected to `total_assign_count_qtd_by_cat_section_no_flags < expectation` — a flagged assignment no longer counts toward the valid total, so falling short of the weekly expectation (including entering nothing) now fires. Verified: 4,132 previously-invisible zero-entry categories now surface; healthy anchor rows dropped 746 → 442 as more teachers are correctly marked short; a flagged-but-still-sufficient category no longer fires
- Added `assignment_category_code` to `rpt_tableau__gradebook_audit`'s uniqueness key — zero-entry categories emit a null-`assignmentid` row, so multiple short categories in one section would otherwise collide
- `int_tableau__gradebook_audit_flags_calculations`: moved a preserved-table filter (`s.quarter_start_date <= current_date`) out of a `LEFT JOIN`'s `ON` clause into `WHERE` (was silently not filtering `s`; harmless today since no future quarters exist yet, but incorrect as written)
- Updated the reference doc and Paterson notes to match: the flag computation description, and Paterson's blocker restated as permanent (plugin cannot be installed) rather than "pending deploy," with the Newark-spoofing plan noted as undecided
- This delivers Task 9 (see below) — moved out of post-merge/TBD

**Remaining before merge (tracked in plan):**
- ~~6b.3: disable `stg_google_sheets__gradebook_expectations_assignments`~~ — done
- ~~6h.1–6h.2: flag removals spot-check + commit~~ — done
- ~~6h.6: `rpt_tableau__gradebook_ms_hs_comments` Miami cleanup~~ — done (no Miami refs; teacher scaffold exclusion covers it)
- ~~6h.5: deprecate `rpt_tableau__assignment_checks`~~ — done
- ~~Task 7: reference doc AY 2026-2027 section update~~ — done
- ~~Overwrite `rpt_tableau__gradebook_audit` namespace vs. parallel `_v4` model~~ — done, MS signed off
- ~~CI unblock: stale `dbt_cloud_pr_70403104388001_4132_*` schemas~~ — dbt Cloud CI passes on the latest commit; no longer blocked
- 6h.8: exposure split (`gradebook_audit` / `gradebook_gpa`) — pending LSID confirmation

**Task 9 — Assignment validity filter for QTD count: delivered (was post-merge/TBD):**

Stakeholders raised this at the June 2026 HS all-hands: assignments should only count toward QTD if they meet gradebook policy (score entered on ≥90% of expected students, <50% exempt, no null/above-max scores, S-category scores ≥50%, plus the H/W/F policy checks). This is now met — not via the new column/flag originally sketched in the plan, but because the validity signal (`assignment_has_flags`, and the `..._no_flags` count that excludes flagged assignments) already existed in `int_powerschool__gradebook_assignment_scores_rollup`; the "Second review pass" fix above is what actually wires it into `expected_assign_count_not_met`. One flag fires at the assignment/category level per the original ask. Full mapping from the original design sketch to what shipped is in plan Task 9. Still open from the original ask: whether to keep the individual per-student score flags alongside this assignment-level one, or consolidate (plan Step 9.4) — a T&L/Tableau-layout decision, not a code gap.

</details>

## AI Assistance

AI-assisted: plan authorship, SQL drafting, iterative review, and refactoring were Claude-assisted. All design decisions (quarter-grain migration, teacher/student split, FYI flag removal, per-assignment categories grain, exceptions removal, Paterson inclusion, manager columns, Miami exclusion, MS/HS model convergence) were human-directed.

### General

- [x] Include (or update) an
- [ ] **Breaking change?** Yes — `rpt_tableau__gradebook_audit` (contracted) is fully replaced with a new `row_type`-based column set (see the Breaking change note above). The Tableau workbook must be rebuilt post-merge.

### Docs

- [x] Nav entry added to `mkdocs.yml` for `reference/gradebook-audit-data-model.md`

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes.
- [ ] **Dagster Cloud** — not triggered (no Dagster-relevant paths changed; this PR is dbt/docs-only).

---
  - https://app.asana.com/0/0/1215457788396037


GabyRangelB:
@claude - can you update your review please? note: we are trying to collapse the model. we have added two new int_tableau views to replace implementation plan's model 6a and 6b for now. can you focus on analyzing whether these two new views are thus far correctly replacing 6a and 6b? also note that flag exceptions are no longer a feature of this data model. thanks!

GabyRangelB:
@claude this PR has been updated since your last review. Can you take a fresh look?

GabyRangelB:
@claude - made some changes. please update your review

GabyRangelB:
@claude — responses to all open items from the June 22 review:

**1. `assign_s_score_less_50p` non-HS asymmetry** — intentional by design. Confirmed via `stg_google_sheets__gradebook_flags`: `assign_s_score_less_50p` is configured as an MS-only flag for AY 2026 (Camden MS, Newark MS, Paterson MS). The HS policy is handled by the separate `assign_s_hs_score_less_50p` flag, which was split out in AY 2025 precisely because HS requires a different rule (no missing-assignment guard). The `is_missing` asymmetry in the SQL reflects two distinct policies controlled by the flags config sheet.

**2. Paterson absence comment in `category_grades` and `gradebook_assignments`** — addressed. Block comments have been added to both models explaining that `kipppaterson_powerschool` is intentionally absent pending the GradeBook PS plugin deployment, tracked in #3908.

**3. `teacher_scaffold_assignment` branch unused by `flags_calculations`** — intentional placeholder. This PR is still in progress; the branch will be consumed by the assignment-grain teacher flags view in a future step. Keeping it in place is by design.

**4. Window partition on `assignment_category_term` vs `quarter`** — intentional. `assignment_category_term` is defined as `concat(assignment_category_code, right(quarter, 1))` (e.g. W1, H2, F3, S4), so it already encodes the quarter. Partitioning on it is equivalent to partitioning on both category and quarter together, which is the correct grain for this calculation.

**5. `flags_calculations` not wired into downstream pipeline** — expected. Wiring into `int_tableau__gradebook_audit_flags` and `rpt_tableau__gradebook_audit` is a pending step tracked in the implementation plan.

GabyRangelB:
@claude can you do a review, please?

GabyRangelB:
@claude — addressing your review feedback. Three fixes applied in the latest commit:

**1. `is_healthy_gradebook` semantics inverted (critical)**

`max(audit_flag_value)` evaluates to `true` when a flag fires, making the column name backwards. Fixed by negating:

```sql
not max(audit_flag_value) over (
    partition by _dbt_source_project, academic_year, schoolid, teacher_number, `quarter`
) as is_healthy_gradebook
```

Now `true` = no flags fired (gradebook is clean). Both UNION branches updated accordingly.

**2. Missing `sectionid` in JOIN conditions (critical)**

Both UNION branch JOINs between `int_extracts__course_schedule_by_term` and `flags_unpivot` were missing `sectionid`, causing cross-section fan-out for teachers with multiple sections. Added `and s.sectionid = f.sectionid` to both branches.

**3. Uniqueness test added to YML**

`dbt_utils.unique_combination_of_columns` added on `(_dbt_source_project, sectionid, quarter, audit_flag_name, cte_grouping)`. The key may need `assignmentid` for the assignment branch — still under review.

---

**On the missing `cron_schedule` in the `gradebook_audit` exposure:** this is intentional. Dagster does not own the Tableau workbook refresh for this dashboard. The exposure is there to document the dependency, not to drive a schedule. A cron will be added once we confirm ownership of the refresh cadence.

GabyRangelB:
@claude - still working on the fan-out. can you check the rest?

GabyRangelB:
@claude - please review one more time

GabyRangelB:
@claude - made some corrections. thoughts?

GabyRangelB:
@claude hopefully all done?

GabyRangelB:
**Note on `.claude/skills/gradebook-audit/SKILL.md`** — the claude-review bot flagged this file, but it is being **added** by this branch, not deleted.

`git diff origin/main --name-status` shows:

```
A       .claude/skills/gradebook-audit/SKILL.md
```

Status `A` = Added. The file does not exist on `main` and this PR creates it. Please do not delete it during review — the bot appears to have misread the diff direction.

src/dbt/kipptaf/models/extracts/tableau/intermediate/properties/int_tableau__gradebook_audit_student_scaffold.yml
```diff
@@ -1,7 +1,7 @@
 models:
   - name: int_tableau__gradebook_audit_student_scaffold
-    # config:
-    #   materialized: table
```
cbini, line 4: retain materialization config

src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__gradebook_audit_v4.yml
cbini, line 1: why not overwrite existing rpt_tableau__gradebook_audit?

src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__gradebook_audit.yml
```diff
@@ -1,5 +1,7 @@
 models:
   - name: rpt_tableau__gradebook_audit
+    config:
+      enabled: false
```
cbini: drop if no compelling reason not to reuse existing namespace

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v4.sql
cbini, line 1: https://github.com/TEAMSchools/teamster/pull/4132/changes#r3492971984

same question

src/dbt/kipptaf/models/google/sheets/staging/stg_google_sheets__gradebook_flags.sql
cbini, line 1: why the change if we're disabling this?

cbini:
we need to disable the google sheet sources as well

src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__gradebook_audit_v4.yml
GabyRangelB: HS is on board with the changes. MS has not signed off yet. the new person taking over MS gradebook policies started while I was on PTO and i wont meet with her until this week or next. once MS tells me they agree to the same model as HS, i will override.

src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit_v4.sql
GabyRangelB: same answer https://github.com/TEAMSchools/teamster/pull/4132#discussion_r3571432651

src/dbt/kipptaf/models/google/sheets/staging/stg_google_sheets__gradebook_flags.sql
GabyRangelB: the case statement was to accommodate flags from miami, and flags that are gone from hs. grade level only existed for miami. even if ms decides to keep the same audit model as last yr, the alt code and grade level arent needed. i changed it to prepare it in case i am not able to fully disable this model for ms.

src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__gradebook_audit.yml
GabyRangelB: need to wait for ms signing off on new model

GabyRangelB:
> we need to disable the google sheet sources as well

not to sound like a broken record, but waiting on ms sign off

cbini (commented):
## Code review — gradebook audit AY 2026-2027

Reviewed at head `298f0b80` vs base `5699a97` via three focused passes (upstream data models, tableau report-layer SQL, docs/skill/plan + PR hygiene). The report-layer pass validated grain/types against the live PR-CI build (aggregate-only queries, no PII). AI-assisted review.

### What this PR actually ships (framing)

The title/body read as an in-place rewrite of the whole `rpt_tableau__gradebook_audit` pipeline. It isn't. The eight "Task 6 rewritten" `.sql` files are **untouched** — they're **disabled** (`config: enabled: false` in their `.yml`) — and a **new parallel `v4` pipeline** ships alongside them, with the `gradebook_audit` exposure re-pointed to it:

```text
int_extracts__course_enrollments_by_term ┐
int_extracts__course_schedule_by_term    ┤
int_powerschool__gradebook_assignment_scores_rollup ┤→ int_tableau__gradebook_audit_flags_calculations → rpt_tableau__gradebook_audit_v4
int_powerschool__u_expectations_qtd_unpivot ┤        (3 branches)                                        (3 branches, contracted)
base_powerschool__final_grades / storedgrades ┘
int_extracts__course_enrollments_by_term → rpt_tableau__gradebook_es_comments (rewritten)
```

The migration is DAG-coherent: no enabled model references a disabled one, and the deleted `int_powerschool__u_expectations_qtd` has no dangling refs. The pipeline builds, contracts pass, branches are mutually exclusive (0 uniqueness violations on the 301k/83k-row PR-CI build). **The code is in good shape; the docs and the PR description are not.**

### Strengths

- v3 teardown is disciplined — every renamed/removed column (`_dbt_source_relation`, `school_level` to `school_level_alt`) is absorbed by disabling all v3 descendants; no enabled model strands a ref.
- `is_healthy_gradebook` logic is correct in code: `not max(audit_flag_value)` (true = no flags). Verified true on exactly the 686 anchor rows, false on all 82k flag rows; branches are mutually exclusive with 0 overlap.
- `cast(null as string)` for `quarter_comment_value` is applied in every branch of both `flags_calculations` and `v4` — the STRING contract holds.
- Exposure/threshold/toggle wiring is accurate: `gradebook_audit` to `rpt_tableau__gradebook_audit_v4`; `min_graded_percent` `0.90` at rollup line 94; summer-toggle markers exactly where the skill claims.

### Issues

#### Important (should fix before merge)

**1. The primary deliverables (docs + skill) invert the model's central signal.** This is a `docs(dbt)` PR, and the docs describe the opposite of the code:

- `docs/reference/gradebook-audit-data-model.md` lines 326-334 and the design spec state `is_healthy_gradebook = max(audit_flag_value) over (...)`, "true when at least one flag fired." Code (`rpt_tableau__gradebook_audit_v4.sql:25`) is `not max(audit_flag_value)` via a GROUP BY aggregate (not a window), true when NO flag fired. Formula (missing `not`), mechanism (aggregate vs window), CTE name (`health_calc`, not `flags_unpivot`), and the Branch-1/Branch-2 gloss are all wrong or reversed.
- Both `int_tableau__gradebook_audit_flags_calculations` and `v4` are documented as 2-branch UNIONs; both are actually 3-branch (`sections_teacher` / `assignment_teacher` / `student_course`). The doc omits `sections_teacher` — the branch that produces v4's "No Flags" anchor.

**2. `SKILL.md` procedures will actively misdirect maintainers.**

- "Add/Remove a flag" and the debug step say to edit the `flags_unpivot` CTE with a 3-flag list. In code `flags_unpivot` unpivots only 2 flags; the 3-flag list lives in `health_calc`; `expected_assign_count_not_met` is emitted as a literal in v4 branch 2. Following the procedure would silently break the health signal.
- Says "ES students appear in student branches for EOQ flags." Code excludes ES from all three branches (`s.school_level_alt != 'ES'`, line 379). ES is handled by the separate `es_comments` model.
- The three "Always read first" relative links are off by one directory (`../../` should be `../../../`), so all three resolve to nonexistent `.claude/docs/...` paths.

**3. PR body + "Breaking change" checkbox materially misdescribe the change.** The body describes in-place `.sql` rewrites and column removals/renames on `rpt_tableau__gradebook_audit`; that model's SQL is unchanged (it's disabled). The body also lists models not in the diff at all (`int_powerschool__terms`, `base_powerschool__sections`, `stg_powerschool__storedgrades`, `int_powerschool__category_grades`, `stg_powerschool__calendar_day`, `int_extracts__student_enrollments`) as if they landed here, and omits a model that did ship (`int_extracts__course_schedule_by_term`). Suggest restating as: "8 contracted models disabled; parallel v4 pipeline added; `gradebook_audit` exposure re-pointed to `rpt_tableau__gradebook_audit_v4`."

**4. `int_powerschool__gradebook_assignments_scores`: no uniqueness test + BETWEEN date-range join.** Two compounding gaps on a model whose grain is now load-bearing for the new rollup's flag counts:

- `int_powerschool__gradebook_assignments_scores.sql:76` uses `and a.duedate between e.cc_dateenrolled and e.cc_dateleft` — violates the `src/dbt/CLAUDE.md` "Date-range joins" rule (use half-open). The sibling mart `fct_grades_assignments.sql:99-100` already uses the correct half-open form against the same `cc` table (which has known duplicate rows, #3900). Pre-existing, but sits inside the join block this PR rewrote.
- The `.yml` has no uniqueness test (per-layer requirement for intermediate models). With the BETWEEN, any boundary fan-out silently inflates every `count(students_dcid)` / `countif(is_expected)` in the rollup, undetected. Add `dbt_utils.unique_combination_of_columns` on `(_dbt_source_project, assignmentsectionid, students_dcid)` and switch to the half-open form.

**5. `is_expected` description (and PR body) claim a 7-day grace the code doesn't implement.** `properties/int_powerschool__gradebook_assignments_scores.yml` lines 113-118 (changed here) say `is_expected` is true when "at least 7 days have passed since the assignment due date." The CASE (`int_powerschool__gradebook_assignments_scores.sql:29-35`) only checks `isexempt` + `iscountedinfinalgrade` — no due-date logic. No literal 7-day grace exists anywhere; downstream `flags_calculations` uses a `duedate <= week_end_sunday` weekly cutoff instead. Reconcile the code, the description, and the PR-body claim.

**6. `rpt_tableau__gradebook_es_comments`: no uniqueness test.** Fully rewritten `rpt_` model with a new grain, but the `.yml` has no uniqueness test (per-layer requirement). Grain is clean (0 dups verified) — add `dbt_utils.unique_combination_of_columns` on `(_dbt_source_project, studentid, sectionid, quarter)` to lock it.

**7. `int_extracts__course_enrollments_by_term` grain omits `academic_year`.** No year filter; the dedup partition and uniqueness key are `(_dbt_source_project, studentid, course_number, quarter)`. `course_number` recurs across years, so a student taking the same course in two years collapses to one row (latest `exitdate` wins). Current-year rows survive (today's dashboard is unaffected), but it's a latent grain defect and needlessly materializes full history as a `table`. Add `where academic_year = {{ var("current_academic_year") }}`, or add `academic_year` to the partition + key.

#### Questions for the author (confirm intent — likely not bugs)

- **v4 drops 22.5% of section-quarters** (1,510 of 6,710): a clean section under an otherwise-flagged teacher appears in no branch, because health is partitioned at teacher x school x quarter. Documented as intentional in `rpt_tableau__gradebook_audit_v4.yml` lines 326-330, but please confirm the dashboard measures teacher-level health — a per-section compliance view would silently omit those sections.
- **`RHET` dropped** from the `es_comments` credit-type filter (`rpt_tableau__gradebook_es_comments.sql:58`: `in ('HR', 'MATH', 'ENG')`); the AY2025-26 `qt_es_comment_missing` definition included `RHET`. Intentional?
- **Branch 2 emits a section/category flag at assignment grain** (~79k rows). Fine for `is_healthy_gradebook` and drill-down, but any Tableau measure that counts flag rows would over-weight it vs the ~3.6k student-course rows.

#### Minor

- Missing model/column descriptions: `int_powerschool__gradebook_assignment_scores_rollup.yml` (no model description + several undescribed columns); `rpt_tableau__gradebook_es_comments.yml` (zero descriptions).
- `int_powerschool__u_expectations_qtd.yml` filename not renamed to `..._unpivot` (binds on `name:`, so it works).
- `int_extracts__course_schedule_by_term.yml` lines 306-310 say `school_level_alt` is null otherwise, but the SQL returns `d.school_level`. `int_extracts__course_enrollments_by_term.yml` lines 1217-1222 describe `rn` via "qualify" though the model correctly avoids QUALIFY.
- Reference doc internal inconsistencies: header still names the old model; "four active flags" vs 2/3 in code; a "Deleted `int_tableau__gradebook_audit_scaffold_unpivot`" section for a model in neither base nor head; "Eliminated/Deleted" for models that are only disabled.
- Disabled v3 models retain dead refs to removed columns — a re-enable would fail to build; worth a `-- TODO` at those refs or a note in the plan.
- `rpt_tableau__gradebook_gpa` has no uniqueness test (pre-existing); severity inconsistency across the new unique tests; redundant `and not h.is_healthy_gradebook` at `rpt_tableau__gradebook_audit_v4.sql:267`.

### Verdict

**Not ready — with fixes.** The v4 code is sound, builds, and its grain/typing/health logic verified correct. But for a docs-primary PR, the reference doc, design spec, and skill invert the central `is_healthy_gradebook` semantic and misdescribe the branch structure (Issues 1-2), and the PR body misrepresents the scope (Issue 3). Those, plus the two missing uniqueness tests and the BETWEEN/grain gaps (Issues 4-7), should land before merge. The three author questions each want a one-line confirmation.

cbini:
@GabyRangelB ran an independent review. Feel free to push back on any of these.

GabyRangelB:
Done — disabled `src_google_sheets__gradebook_flags` and `src_google_sheets__gradebook_exceptions` at the source level in `sources-external.yml`. Neither staging model has a live consumer, so Dagster no longer pulls either sheet into BigQuery.

GabyRangelB:
@claude made some changes. can you review please?

GabyRangelB:
@claude - please review the updates

GabyRangelB:
@claude - are you back? can you review the changes?

cbini:
@GabyRangelB I forced the review update

cbini (dismissed):
@GabyRangelB approved for merge once you sign off on the final claude review. Also have claude debug why the dbt build is failing--probably some staging artifact

src/dbt/kipptaf/models/powerschool/sources-kipppaterson.yml
```diff
@@ -545,6 +545,15 @@ sources:
...
                 - stg_powerschool__u_def_ext_students
+      - name: stg_powerschool__u_expectations
+        config:
+          meta:
+            dagster:
+              group: powerschool
+              asset_key:
+                - kipppaterson
+                - powerschool
+                - stg_powerschool__u_expectations
```
cbini: move this under a new source (kipppaterson_fake?) where the schema points to kippnewark_powerschool

src/dbt/kipppaterson/models/powerschool/sources-kipppaterson.yml
cbini, line 1: copy this into src/dbt/kipptaf/models/powerschool/sources-kipppaterson.yml and just rename the source name to something like kipppaterson_fake

src/dbt/kipppaterson/models/powerschool/sis/staging/stg_powerschool__u_expectations.sql
cbini, line 1: this can be deleted after source do-si-do

src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__u_expectations.sql
```diff
@@ -9,6 +9,9 @@ with
                     source(
                         "kippnewark_powerschool", "stg_powerschool__u_expectations"
                     ),
+                    source(
+                        "kipppaterson_powerschool", "stg_powerschool__u_expectations"
```
cbini: use the spoofed source name here

GabyRangelB:
@cbini re: the Paterson `U_EXPECTATIONS` spoof threads (sources-kipppaterson.yml, the kipppaterson-project sources file, its staging model, and kipptaf's `stg_powerschool__u_expectations`) — pushed the restructure you suggested. Wanted to explain the one place I diverged from the literal suggestion and why.

**What changed:** the spoof no longer lives as an override model inside the `kipppaterson` project. `kipptaf/models/powerschool/staging/stg_powerschool__u_expectations.sql` now handles it directly: `dbt_utils.union_relations()` covers only the two real relations (Camden, Newark), and a hand-written `paterson_spoof` CTE reads Newark's real source directly, filtered to `school_level = 'MS'`, with a hardcoded `'kipppaterson' as _dbt_source_project`. The two are combined with a plain `UNION ALL` at the top level. The `kipppaterson`-project override model and its misplaced sources file are deleted.

**Why not just add Paterson as a third relation inside `union_relations()`** (closer to the literal ask): tested this first and it doesn't work cleanly, for two reasons:

1. `union_relations()` derives `_dbt_source_relation` from the relation's actual physical path (`project.schema.table`). Since the Paterson entry would point at the exact same physical table as the real Newark entry, both would resolve to the identical `_dbt_source_relation` string — `extract_code_location()` would then derive `_dbt_source_project = 'kippnewark'` for both, so the Paterson rows would be indistinguishable from real Newark rows downstream, not labeled `kipppaterson` at all.
2. The macro's `where` param is global to every relation in the list — there's no way to scope the `school_level = 'MS'` filter to just the Paterson entry without also dropping Camden/Newark's HS rows.

I also want to flag the specific value you floated (`_dbt_source_relation` resolving to `kipppaterson_powerschool_fake`) — tested it directly rather than eyeballing it:

```sql
select regexp_extract('kipppaterson_powerschool_fake', r'(kipp\w+)_') -- -> 'kipppaterson_powerschool'
select regexp_extract('kipppaterson_fake', r'(kipp\w+)_')             -- -> 'kipppaterson'
```

`extract_code_location()`'s regex is greedy, so a value with two underscores resolves through the last one, not the first — `kipppaterson_powerschool_fake` doesn't cleanly extract back to `kipppaterson`. Your own follow-up (drop the fake source, just hardcode `kipppaterson` as the source project) sidesteps this entirely, which is what's implemented now — no new source alias at all, just a direct read of the already-declared `kippnewark_powerschool` source with an explicit literal.

Verified end-to-end via a dev build: `stg_powerschool__u_expectations` through `rpt_tableau__gradebook_audit` all build clean, all tests pass, and Paterson MS rows show up correctly (688 real `assignment_teacher` rows in `flags_calculations`) — same real Newark-sourced data as before, just flowing through an honestly-named path instead of a project-boundary-crossing override. `kipppaterson` reparses clean after the deletions too.

GabyRangelB:
@claude - can you check the changes? do you have any suggestions for the current view of rpt_tableau__gradebook_audit? im trying to simplify it

GabyRangelB:
@claude - final review, please. be thorough

GabyRangelB:
**Re: the Critical finding on Paterson MS assignment counts — confirmed valid, root-caused, and fixed (`317b356e0`).**

Verified it: `int_powerschool__gradebook_assignments` unioned only newark/camden/miami, so Paterson MS (in-scope in the scaffold, not excluded like Miami/ES) had zero assignment rows → `not_enough_assignments` true for every section → both health booleans false for 100% of Paterson MS.

It turned out to be a **two-model cascade**, not just that one union. After adding Paterson to the assignments union, Paterson was still ~100% short because `stg_powerschool__assignmentscore` **also** omitted Paterson — assignments with no scores read as 99.9% `percent_graded_min_not_met`. Paterson's underlying gradebook ingestion is already in prod (`kipppaterson_powerschool` tables populated, AY2024–2026), so adding it to both unions is globally correct.

**Fix:** added `kipppaterson_powerschool` to both `int_powerschool__gradebook_assignments` and `stg_powerschool__assignmentscore`, and seeded the `zz_stg_kipppaterson_powerschool` copies so dbt Cloud CI resolves the new relations.

**Validated in dev — Paterson now matches the other regions:**

| metric | before | after | Newark / Camden |
|---|---|---|---|
| assignment flag rate | 99.9% | 23% | 40% / 43% |
| `percent_graded_min_not_met` | 99.9% | 8.2% | 6% / 12% |
| `not_enough_assignments` | 100% | 75.3% | 46% / 68% |

dbt Cloud CI is green on the fix commit (no new warnings). The separate `U_EXPECTATIONS`-plugin item (Paterson expectations still spoofed from Newark, since the plugin can't be installed on Paterson) is unchanged and tracked in the reference doc.

cc @cbini — this adds Paterson to those two shared powerschool unions and seeds their staging copies; flagging since it touches the powerschool layer.

GabyRangelB:
@claude are we finally done?

cbini (approved):

PR #4141, opened 2026-06-08
feat(adp): capture additional WFN worker attributes (#4106)

### LINKED ISSUE #4106: feat(adp): capture all ADP WFN worker attributes in schema + staging/int models
## Problem

The ADP WFN workers asset's `$select` and the `Worker` pydantic model don't cover every field the API returns, so attributes the API emits are silently dropped (the Avro schema-valid check only *warns* on extras, it doesn't capture them).

Confirmed missing on `workAssignments` in live data (full-population check, 4,562 workers / 4,740 assignments):

- `jobFunctionCode` (present on 64 assignments; **absent from `workers/meta`** — proving meta is not a complete field list)
- `customCountryInputs`
- `rehireEligibleIndicator`

There are likely more across `person` and the top-level `worker` object once we look at the full default representation.

## Goal

Capture **every attribute and subattribute** the ADP WFN workers API returns in:

1. the `Worker` pydantic model (`src/teamster/libraries/adp/workforce_now/api/schema.py`) and the generated `WORKER_SCHEMA` Avro,
2. the staging model `stg_adp_workforce_now__workers` (+ properties YAML / contract),
3. the intermediate models (`int_adp_workforce_now__workers*`, especially `__work_assignments`) and their properties YAML.

## Decided approach

- **Enumerate from live data, not meta.** `workers/meta` is incomplete (misses `jobFunctionCode`). Re-fetch the **full worker population with `$select` dropped** (ADP's full default representation) and union every field path across all records to catch sparse fields. (~43MB local PII pull, kept local only.)
- **Drop the `$select`** in the workers asset so the API returns its full representation going forward (vs. a curated list).
- **Capture everything**, including person-level identifiers (governmentIDs incl. SSN/ITIN, addresses, DOB, etc.). Treated as PII per existing handling.

## Impact / risks to plan for

- **Surrogate-key hash change.** Staging keys on `to_json_string(workAssignments)` / the worker struct; adding captured fields changes those hashes -> SCD2 re-key in staging and ripple into int models + downstream marts. Deliberate hash-change event per hash discipline.
- **External-table + contract churn.** New Avro fields change the external source columns; staging is contract-enforced, so properties YAML must add every new column. Requires `stage_external_sources --target staging` before dbt Cloud CI passes.
- **Descriptions.** Every new staging/int column needs a `description:` per dbt conventions.

## Out of scope

- The jobs validation table (tracked in #4071).
- Any change to the `workers_sync` (update) asset.

## References

- Field-coverage analysis and the `meta`-incompleteness finding were established while investigating #4071.

cbini:
## Summary & Motivation

When merged, this captures additional ADP WFN worker attributes that were being
silently dropped (#4106). The `Worker` pydantic model now declares the complete
retrievable representation; a curated `$select` and the contract-enforced
staging model surface the new fields, while unsynced sensitive PII stays off the
wire and out of the warehouse.

**AI Assistance:** Human-directed design and decisions (curation policy, cut
list, rollout); Claude Code ran the live-data field enumeration, made the
schema/SQL/contract edits, and validated each layer against live ADP + a seeded
staging external table.

### What changed

- **Schema** (`libraries/adp/workforce_now/api/schema.py`): model the full
  retrievable Worker representation — adds `jobFunctionCode`,
  `rehireEligibleIndicator`, `customCountryInputs`, `religionCode`, `pagers`,
  `payGradePayRange`, `laborUnion`, `workShiftCode`, `deathDate`,
  `adjustedServiceDate`, `retirementDate`, custom amount/percent/telephone
  fields, address `deliveryPoint`, and `identityDocuments`. Validated against
  live data: the Avro schema-valid check reports no extras.
- **`$select`** (kipptaf workers asset): adds the `person` sub-paths that return
  data, plus `religionCode`.
- **Staging** (`stg_adp_workforce_now__workers`): flattens `religionCode` and
  passes through `person__communication__pagers`; regenerates the
  `work_assignments` and `other_personal_addresses` contract structs. Built
  clean against the wide external table.
- **Intermediate** (`int_…__work_assignments`): flattens `jobFunctionCode` and
  `rehireEligibleIndicator`.

### Divergence from #4106

The issue proposed dropping `$select` and capturing everything including
`governmentIDs` (SSN/ITIN). This PR instead:

- keeps a **curated `$select`** — minimum PII on the wire (the pydantic model is
  the real Avro gate, so dropping `$select` saves no modeling work);
- **models** the full representation but **excludes `governmentIDs` and
  `identityDocuments`** from `$select` and staging (ADP-masked PII/SPI, no
  warehouse consumer) — they stay declared in the model so a future need is a
  one-line `$select` change;
- captures + warehouses `religionCode` per maintainer decision (a FERPA indirect
  identifier — tagged `contains_pii`).

### Rollout notes

- CI reads the `zz_stg` external table, seeded from a branch-deployment-style
  materialization in `teamster-test`. BigQuery Avro autodetect uses the
  last-alphabetical file's schema, so the max-partition file carries the wide
  schema.
- **Post-merge:** a recent-window re-pull of
  `kipptaf/adp/workforce_now/workers` is required so prod-bucket Avro carries the
  wide schema and prod staging picks up the new columns.
- Expect a **one-time SCD2 re-key** on `stg`/`int` (deliberate whole-struct hash
  recomposition), rippling into downstream people/work-assignment marts.

### dbt

- [x] Properties files updated for all modified models.
- [x] Ran `stage_external_sources --target staging` for the external source.
- [ ] Breaking change: no column renames/removals on contracted models; only
      additive columns + a deliberate surrogate-key recomposition.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Closes #4106.


cbini:
Triaged against the live full-population enumeration (4,562 workers, `$select` dropped) that drove this PR. Fixed 3/5/6; pushing back on 1/2/4 with that data.

### Fixed (commit e02f88a)

- **#3** — `test_avro_schema_includes_new_fields` now serializes the generated schema and asserts all 18 new fields — including deeply-nested `religionCode`, `laborUnion`, `payGradePayRange`, `adjustedServiceDate`, etc. — survive `py_avro_schema.generate()`.
- **#5** — added `deliveryPoint STRING` to the `assigned_work_locations` address struct in the intermediate YML to match the staging contract.
- **#6** — inline comment distinguishing `PayGradeRate` (amountValue only) from `Rate`.

### Pushing back — 1, 2, 4 are intentional, not gaps

The live enumeration confirms these fields **are not returned for KTAF** (absent from the dropped-`$select` full response):

- `person/deathDate` — not present
- `workerDates/adjustedServiceDate`, `workerDates/retirementDate` — `workerDates` returns only `originalHireDate` / `rehireDate` / `terminationDate`
- `person/communication/faxes` — `communication` returns `emails` / `landlines` / `mobiles` / `pagers`, no `faxes`. That's exactly why `pagers` is surfaced and `faxes` isn't (rule: flatten what returns data). The premise that faxes arrives because `communication` is in `$select` doesn't hold.

By design (approved in the spec), the pydantic model is the **complete retrievable representation**, while `$select` and staging surface only what ADP returns today. The rest are modeled for forward-compatibility so adding one later is a one-line `$select` change rather than a schema/Avro change. Surfacing them now would add permanently-null columns — and for `deathDate`, request a sensitive field with no data and no consumer. I added a module docstring to `schema.py` making this split explicit so the model↔`$select` asymmetry isn't read as accidental. The `test_new_fields_parse` assertions verify model *parsing* of the superset, not warehouse capture.

### #7 — no action

Staging doesn't output a surrogate key (it's internal to the SCD2 dedupe), so downstream can't key on it. The one-time re-key is the deliberate hash recomposition flagged in the PR body; the `--full-refresh` CI run rebuilt the downstream work-assignment/people marts and their uniqueness tests passed (the re-key adds new SCD2 segments, not duplicate rows).

PR #4162, opened 2026-06-11
feat(kipptaf): add KIPP Aid award columns to kfwd dashboard

CGibson17:
# Pull Request

## Summary & Motivation

When merged, this pull request will surface each student's KIPP Aid awards on
`rpt_tableau__kfwd_dashboard`, joined from `rpt_tableau__kfwd_aid_report` on
`sf_contact_id` and `academic_year` (matching the dashboard's existing
per-student/per-year grain).

Awards are pivoted to **wide format** so the dashboard's one-row-per-student-per-year
grain is preserved (no fan-out). New columns:

- `total_aid_amount` — sum of all the student's awards that academic year
- `aid_award_count` — count of all awards that year
- `award_1_name` / `award_1_amount` … `award_10_name` / `award_10_amount` — each
  award in its own columns, ordered earliest-first

The slot count is a single tunable Jinja var (`{%- set max_awards = 10 -%}` at
the top of the model). Awards beyond the cap still contribute to
`total_aid_amount` and `aid_award_count`.

**AI Assistance:** Pivot structure, Jinja loop, and YAML contract entries were
Claude-authored; the join keys, aggregate-then-pivot approach, and the slot cap
were human-directed across iteration.

### Open items for reviewer

- **`max_awards = 10` is provisional.** It was not possible to query the
  warehouse during development to confirm the true max awards-per-student-per-year.
  Please run `select max(c) from (select count(*) c from
  rpt_tableau__kfwd_aid_report group by sf_contact_id, academic_year)` and bump
  the var if any student exceeds 10 (otherwise their extra awards drop from the
  `award_N_*` slots, though totals remain correct).
- Local `dbt build` / `trunk` could not run in the dev environment (package
  registry egress blocked); the Jinja was rendered to confirm it compiles to 20
  award columns matching the 22 YAML contract entries. **dbt Cloud CI is the
  validation gate.**

### dbt

- [x] Properties file updated — `rpt_tableau__kfwd_dashboard.yml` carries
      contract types + descriptions for all new columns
- [ ] Exposure — `rpt_tableau__kfwd_dashboard` already has its Tableau exposure;
      no new consumer added
- [x] **Breaking change?** Additive only — new columns, no renames/removals on
      existing dashboard columns

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

### DIFF
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__kfwd_dashboard.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__kfwd_dashboard.yml
@@ -610,3 +610,37 @@ models:
           Overgrad. False if the student has no Overgrad top-choice data.
       - name: es_attended
         data_type: string
+      - name: id
+        data_type: string
+        description:
+          KIPP Aid award record ID (Salesforce) joined from
+          rpt_tableau__kfwd_aid_report on contact and academic year.
+      - name: aid_name
+        data_type: string
+        description: KIPP Aid record name / auto-number (e.g. KA-016473).
+      - name: amount
+        data_type: numeric
+        description: KIPP Aid award amount in dollars.
+      - name: award_date
+        data_type: date
+        description: Date the KIPP Aid award was recorded.
+      - name: created_date
+        data_type: timestamp
+        description: Timestamp the KIPP Aid record was created in Salesforce.
+      - name: fund_type
+        data_type: string
+        description:
+          Maher funding type parsed from the MAHER_FUND tag (Emergency,
+          Enrichment, or Unclassified); null when untagged.
+      - name: kipp_aid_count
+        data_type: numeric
+        description: KIPP Aid count value from the source record.
+      - name: notes
+        data_type: string
+        description:
+          Raw KIPP Aid award note, including any appended MAHER_FUND marker tag.
+      - name: notes_clean
+        data_type: string
+        description:
+          KIPP Aid award note with the MAHER_FUND tag and surrounding whitespace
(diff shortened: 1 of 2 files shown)


CGibson17:
Thanks — both findings addressed in c7fcb1a36.

**Medium — `rpt_`→`rpt_` coupling + redundant join.** Confirmed `stg_kippadb__kipp_aid` already exposes every column the pivot needs (`student`, `academic_year` via `date_to_fiscal_year`, `name`, `amount`, `` `date` ``, `created_date`, `id`) and applies `where not isdeleted`. `aid_awards` now reads directly from staging, so the dashboard no longer references `rpt_tableau__kfwd_aid_report` and skips re-running its roster/enrollment join. Row set is identical (same staging source, same soft-delete filter).

**Low — non-deterministic slot assignment on null dates.** Added `id` as a final `order_by` tiebreaker (`order by \`date\` asc, created_date asc, id asc`), so slot assignment is stable even when both date fields are null — no warehouse count needed to de-risk it.

**Open item (`max_awards = 10`).** Still needs confirmation against the data before merge — I couldn't reach the warehouse from the dev environment. The sizing query (now against the staging model):

```sql
select max(c) from (
    select count(*) c
    from {{ ref("stg_kippadb__kipp_aid") }}
    group by student, academic_year
)
```

If the result is greater than 10, bump `max_awards` at the top of the model and the YAML contract will need the matching `award_N_*` entries.

Also fixed the Trunk Check failure (d25346432 — `trunk fmt` on the Jinja pivot).

GabyRangelB (approved):

PR #4201, opened 2026-06-17
feat(finalsite): Focus SFTP output models + crosswalks (Component 4)

cbini:
# Pull Request

## Summary & Motivation

> When merged, this adds the Finalsite to Focus "Focus-shaped output" layer (Component 4 of the enrollment integration, tracked in #4211), restructured per review into the network's standard cross-project + method-folder conventions.

**Finalsite package** (`src/dbt/finalsite`)
- Staging split into method folders `api/` (contacts, contact_relationships) and `sftp/` (status_report), matching the amplify convention.
- New `api/intermediate/` layer (SIS-agnostic): `int_finalsite__enrollment_lifecycle` (promoted from kippmiami) plus two array-pivot views — `int_finalsite__contact_id_attributes` (→ `focus_student_id`, var-driven) and `int_finalsite__contact_custom_attributes` (→ `is_latino_hispanic`).
- Package var defaults (`current_academic_year: 0`, `finalsite_focus_student_id_field: ""`) so all consumers compile.
- `finalsite: api: +enabled: false` in Newark/Camden/Paterson (no Finalsite API ingestion there); SFTP/`status_report` stays enabled network-wide (kipptaf consumes it).

**kipptaf** (`src/dbt/kipptaf/models/extracts/focus`)
- The 5 `rpt_focus__*` views with all transformation logic, sourcing `kippmiami_finalsite` (staging + intermediate) and `ref()`-ing the in-project Google-Sheets crosswalks. Joins the pivot views (no inline array subqueries). Unit tests live here (7, all passing).

**kippmiami** (`src/dbt/kippmiami/models/extracts/focus`)
- Thin enumerated `rpt_focus__*` wrappers sourcing `kipptaf_extracts` (rpt_powerschool__autocomm pattern).

**Crosswalks** — 4 Google-Sheets crosswalks in kipptaf (grade, school, enrollment-code keyed on `lifecycle_action`, drop-code), populated from the live Sheet.

Component 5 (SFTP transport assets) is out of scope.

## Validation

Built end-to-end against staged Miami data (`zz_stg_kippmiami_finalsite`): `dbt build` of the kipptaf views = PASS 37, ERROR 0, 7 warns (all expected data-completeness gaps). 7/7 unit tests pass. All affected projects parse clean; trunk clean.

### Known follow-ups (warn-level, not blocking)

[...]

Issue #4073, opened 2026-05-29
feat: Finalsite → Focus enrollment integration (KIPP Miami)


cbini:
## Goal

Push accepted/enrolled student records out of Finalsite Enrollment and into Focus (the SIS) for KIPP Miami, so new enrollees are set up in Focus for the school year without manual re-entry.

## Scope

- **Miami only.** Focus is Miami's SIS; the other three regions use PowerSchool. Build in the `kippmiami` code location, alongside the existing Focus DLT and Finalsite assets.
- Direction: Finalsite → Focus (outbound).

## Approach (agreed in brainstorming)

Warehouse-first pipeline with a **pluggable transport seam** — everything up to "Focus-shaped output" is shared; the transport (SFTP vs API) stays swappable and is decided after vendor confirmation.

```text
Finalsite (API now; new SFTP export later)
  -> GCS (Avro) -> BigQuery raw [kippmiami]
  -> dbt staging (normalize source)
  -> dbt eligibility filter (accepted/enrolled, target year)
  -> dbt identity resolution (+ persisted crosswalk)
  -> dbt Focus-shaped output models (API shape and/or 5 SFTP template shapes)
  -> transport: build_bigquery_query_sftp_asset  OR  new OAuth2 FocusResource (POST /student, /multi)
```

Components: source-agnostic ingestion, eligibility filter, identity resolution + persisted crosswalk, Focus-shaped output models, transport seam, optional write-back of the canonical ID to Finalsite, and a reconciliation model comparing the Finalsite eligible roster to existing Focus students.

## Key findings from investigation

- Focus exposes both an officially-supported **SFTP template import** (5 coordinated CSVs: Demographics, Student_Enrollment, Address, Contacts, Linked_Students; keyed on a caller-supplied `STDT_ID`) and a **REST API** with a purpose-built `POST /student` admissions endpoint (OAuth2 client_credentials; one call carries student + demographics + 2 guardians + address + accepting grade/school/program/year; plus a `/multi` batch endpoint).
- The student-enrollment record may only be creatable via the SFTP `STUDENT_ENROLLMENT` file: `enrollments` / `school_enrollments` are GET-only in the API. `POST /student` accepting fields may cover it — unconfirmed.

## Open questions / vendor-confirmation gates

[...]

PR #4217, opened 2026-06-18
feat(dbt): map Finalsite language values to Focus codes in rpt_focus__demographics

### LINKED ISSUE #4206: feat(dbt): map Finalsite lang and promotion-status values to Focus codes
Follow-up to #4201 (Component 4 of #4073).

## Context

Two Focus output fields pass raw Finalsite values through without translating to Focus's expected value codes:

- `LANG` / `PRIMARY_HOME_LANG` / `NATIVE_PARENT_LANG` ← `lang_parent_ss` (e.g. English / Spanish / Creole)
- `GRDE_PROM_ST` ← `track_attributes.promotion_status_ss` (e.g. `Retained`)

## Work

- Confirm with the registrar the Focus value codes these fields expect.
- If a mapping is required, add Google-Sheets crosswalk(s) following the enrollment-code / drop-code crosswalk pattern and join them in `rpt_focus__demographics` / `rpt_focus__student_enrollment`.

## Blocked on

Registrar confirmation of Focus value codes.

Refs #4073

cbini:
## Summary & Motivation

When merged, this pull request will translate Finalsite `lang_parent_ss` values into Focus language value codes in `rpt_focus__demographics`, so the `LANG`, `PRIMARY_HOME_LANG`, and `NATIVE_PARENT_LANG` columns emit Focus codes (e.g. `EN`, `SP`, `HC`) instead of raw Finalsite labels (`English`, `Spanish`, `Haitian Creole`).

Follow-up to #4201 (Component 4 of #4073). Implements the language half of #4206.

Changes, following the existing enrollment-code / drop-code crosswalk pattern:

- New Google-Sheets source `src_google_sheets__focus__language_code_crosswalk` (same spreadsheet as the other Focus crosswalks, tab `src_focus__language_code_crosswalk`) and staging model `stg_google_sheets__focus__language_code_crosswalk` — 94 rows, one per Finalsite picklist value.
- Left-join the crosswalk in `rpt_focus__demographics`; the three language columns now carry `focus_language_code`.
- Warn-level singular test `rpt_focus__demographics__language_crosswalk_resolves` flags any non-null `lang_parent_ss` with no crosswalk entry, so a new Finalsite picklist value surfaces (Ops signal to add it) instead of silently passing through as null.
- Updated the unit test and column descriptions.

Open registrar item (does not block this PR): the Focus `LANG` field historically carried a legacy 3-option set (`EN` / `English` / `Spanish`); this PR assumes it accepts the same FLDOE code as the home/native-language fields. An inline comment in the model flags this. If `LANG` must keep its legacy values, a second crosswalk column is a follow-up.

### Validation

- Staging crosswalk built against staging: 94 rows, `not_null` + `unique` pass; all 94 codes verified against valid Focus option codes (no typos).
- `rpt_focus__demographics` builds end-to-end; real-data output maps English to `EN`, Spanish to `SP`, Creole to `ZF`, Haitian Creole to `HC`, Farsi to `FA`, with no non-null source values falling through to null.
- Unit test and the new crosswalk-resolves test pass.

### AI Assistance

AI-assisted: warehouse exploration, mapping draft, and all model/test scaffolding were authored by Claude Code. Human-directed: the crosswalk sheet values and the worktree/PR decisions.

### dbt

- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — done for `src_google_sheets__focus__language_code_crosswalk`
- [ ] **Breaking change?** `rpt_focus__demographics` is contracted, but column values change (raw label to code); no column rename/removal. The extract is not yet live in Focus, so no downstream consumer breaks.

## CI checks

- [ ] **Trunk** — clean locally on all changed files
- [ ] **dbt Cloud** — pending
- [ ] **Dagster Cloud** — not expected (no Dagster code changes)

Refs #4206

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -552,6 +552,16 @@ Hyphenated identifiers in INFORMATION_SCHEMA paths need backticks — `region-us
 as a bare token fails with "Syntax error: Expected end of input but got '-'".
 Write `` `teamster-332318`.`region-us`.INFORMATION_SCHEMA.TABLES ``.
 
+Single quotes inside a BigQuery string literal escape with a **backslash**
+(`'O\'odham'`), not by doubling (`''`) — the doubled form fails with
+"concatenated string literals must be separated by whitespace".
+
+The BigQuery MCP service account **cannot read GOOGLE_SHEETS external tables**
+("Access Denied: ... while getting Drive credentials", 403) — it lacks Drive
+scope. To inspect a sheet-backed source's rows, build the staging model via dbt
+(`dbt build --select <stg_model> --target staging`; ADC has Drive scope), then
+query the materialized `zz_stg_*` table — a native BQ table, not Drive-backed.
+
 `bq` CLI fallback for shell contexts (Monitor poll loops): binary at
 `/usr/local/share/google-cloud-sdk/bin/bq`, `--project_id=teamster-332318`. Same
 SELECT-only constraints apply.
--- a/src/dbt/focus/CLAUDE.md
+++ b/src/dbt/focus/CLAUDE.md
@@ -9,6 +9,15 @@ dbt projects (currently `kippmiami`).
 Focus Postgres → dlt `sql_database` → BigQuery (`dagster_<project>_dlt_focus`) →
 dbt staging models → dbt intermediate models
 
+## Focus field value codes
+
+A Focus custom field's allowed value codes live in
+`dagster_<district>_dlt_focus.custom_fields` (find the field by `title` /
+`column_name`) joined to `custom_field_select_options` on
+`custom_field_select_options.source_id = custom_fields.id` — `code` is the value
+Focus expects, `label` is the human name. Use this to build Finalsite→Focus
+value crosswalks.
+
 ## Model Structure
 
 ```text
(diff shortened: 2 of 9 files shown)


cbini:
All three findings addressed in 06751a9 — all verified against the codebase first.

**Ref anchor (minor):** Re-anchored the singular test to `int_finalsite__contact_custom_attributes`. Confirmed this is a kipptaf-level model (`kipptaf/models/finalsite/intermediate/`), so the `meta.dagster.ref` needs no `package:` qualifier — same as the `int_collegeboard__ap_unpivot__crosswalk_resolves` precedent. The check now fires on source-data materialization rather than sheet edits.

**`not_null` on `focus_language_code` (minor):** Added at `severity: error`. This is a distinct gap from the singular test (which catches an unmapped `finalsite_language`); `not_null` catches an existing row with a blank code. Passes (sheet is fully populated).

**Unit-test miss case (informational):** Added a third fixture — a student whose `lang_parent_ss` has no crosswalk entry — asserting the three language columns emit null, not the raw label. Passes.

Local validation: unit test, the new `not_null`, and the re-anchored singular test all pass; trunk clean on changed files.

PR #4224, opened 2026-06-18
fix(cube): use human-readable region name for cube-region-* group scoping

cristinabaldor:
## Summary & Motivation

When merged, this pull request will fix `cube-region-*` Google group scoping in `queryRewrite` to filter on the human-readable region name instead of a surrogate hash key.

Previously, the region filter compared the group slug against `locations.region_key` (an MD5 hash derived from `business_unit_code`), making groups like `cube-region-newark-detail` impossible to use. It now filters on `regions.region_name`, so `cube-region-newark-detail` correctly scopes to all Newark schools.

Because BigQuery string equality is case-sensitive and the stored region names are not uniformly cased (`TAF` is all-caps; the city regions are title-case), the slug is mapped to its exact stored name through an explicit `REGION_NAMES` lookup rather than a mechanical title-case transform — a transform could never produce `TAF`. An unmapped slug yields empty filter values, preserving the default-deny behavior.

**AI-assisted:** authored with human direction and review.

### General

## CI checks

- [x] **Trunk** — passes. If it fails, run `trunk check` and `trunk fmt` locally, fix any issues, and push again.
- [x] **dbt Cloud** — passes. If it fails: click **Details** on the check, expand **Invoke `dbt build ...`**, then **Debug Logs**. Or tag **Claude** on the PR with the error details.
- [x] **Dagster Cloud** — passes or not triggered (only runs when relevant paths change and the PR is not a draft). If it fails: check the **Actions** tab for the workflow run logs.

### DIFF
--- a/src/cube/CLAUDE.md
+++ b/src/cube/CLAUDE.md
@@ -109,9 +109,10 @@ Default-deny, group-driven. Read [`cube.js`](cube.js) before modifying.
 - **`queryRewrite`** enforces three filters:
   - Strips student-domain dims/measures (members where `isStudentMember` is
     true) for users without `cube-access-student-data`.
-  - Adds a `locations` filter based on the highest-priority scope group: network
-    (no filter) → region (`region_key`) → school (`abbreviation`). No scope
-    group → empty `IN ()` filter (default deny).
+  - Adds a location filter based on the highest-priority scope group: network
+    (no filter) → region (`regions.region_name`) → school
+    (`locations.abbreviation`). No scope group → empty `IN ()` filter (default
+    deny).
   - For queries touching a staff-domain cube (`isStaffMember`), injects the
     `staff.reporting_chain` segment unless the user has `cube-access-staff-all`.
 - **`isStudentMember` / `isStaffMember` prefix helpers.** Domain gating is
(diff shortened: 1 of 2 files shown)


cristinabaldor:
Addressing the Claude review. The approach changed after the review, so noting how each item lands against the current code:

**Important #1 — case-sensitivity of `regions.region_name`:** Resolved. Verified `kipptaf_marts.dim_regions` stores `Camden / Miami / Newark / Paterson / TAF`, and the region branch now maps the slug to its exact stored name via an explicit `REGION_NAMES` lookup rather than a mechanical title-case transform. This is what fixes `TAF` (all-caps) — a transform could only ever produce `Taf`. An unmapped slug yields empty filter values, preserving default-deny, so a bad/new slug fails safe instead of silently mis-scoping.

**Minor #2 — no unit test for the slug-to-name transform / leading-digit edge:** Superseded. The transform no longer exists; it's a static literal map, so there is no `charAt`/digit edge case. There's also no JS test runner in `src/cube` (no test script in `package.json`). The meaningful invariant — map values matching the DB casing — is a data check, which was run, not a unit test.

**Minor #3 — split/join readability:** Moot. That code was removed; the replacement carries an explanatory comment.

PR #4233, opened 2026-06-20
fix(kipptaf): derive pursuing_degree_type__c from NSC class_level + Account.type

anthonygwalters:
## Summary & Motivation

When merged, this PR will populate `pursuing_degree_type__c` on `rpt_gsheets__nsc_enrollment_new` for every row that has any usable signal — today, that column resolves to NULL on every row because the previous derivation keyed off NSC's `two_year_four_year` field, which is 100% NULL at the source (NSC isn't emitting it for our SFTP feed).

Replaces the dead derivation with a two-step resolution:

1. **NSC `class_level`** (most-recent term in the stint) maps to its credential equivalent — A/B/C/N/M/D/L/P/G per the StudentTracker Detail Report spec, plus institution-emitted non-spec codes A (Associate's-seeking) and B (Bachelor's-seeking) that NSC passes through unchanged.
2. **Fallback to Salesforce `Account.type`** (Public/Private 2 yr/4 yr) for year-in-school codes (F/S/J/R) and NULL `class_level`.

Drops the now-unused `current_two_year_four_year` carrier from `int_nsc__enrollment_stints` and adds `current_class_level` in its place. Code mapping and resolution precedence documented inline in the rpt view and on `current_class_level` in the intermediate's YAML.

Verified output in dev: 215 Associate's, 192 Bachelor's, 17 Master's, 6 Certificate, 6 Course Credit, 3 Graduate Degree (previously: all 439 rows NULL).

## AI Assistance

Claude-assisted: investigated the NULL root cause via BigQuery, confirmed the NSC `class_level` code mapping against the official StudentTracker Detail Report spec, drafted the two-step CTE design, and wrote the implementation. Human-directed on the derivation shape (rejected an inline-CASE-only design in favor of the two-signal CTE + coalesce structure for readability) and on dropping a dead MBA carve-out that the data flow couldn't fire.

### dbt

- [x] Properties YAML updated for both modified models
- [x] No breaking change: `current_two_year_four_year` drop has zero remaining consumers (verified via grep); replaced by `current_class_level` in `int_nsc__enrollment_stints`
- [x] `dbt build --select int_nsc__enrollment_stints+` passes in dev (6 PASS / 0 WARN / 0 ERROR)

## CI checks

- [x] Trunk — clean
- [ ] dbt Cloud
- [ ] Dagster Cloud (not applicable; dbt-only)

### DIFF
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__nsc_enrollment_new.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__nsc_enrollment_new.yml
@@ -43,8 +43,16 @@ models:
       - name: pursuing_degree_type__c
         data_type: string
         description: >
-          Degree type mapped from NSC two-year/four-year classification:
-          "Bachelor's (4-year)" or "Associate's (2 year)".
+          Salesforce picklist value for the credential the alum is pursuing at
+          this institution. Resolved with precedence — the credential implied by
+          the NSC class_level code from the stint's most-recent term (see
+          int_nsc__enrollment_stints.current_class_level for the full code
+          mapping), then a fallback to the Salesforce Account.type ("Public 2
+          yr" / "Private 4 yr" / etc.) for year-in-school codes (F/S/J/R) and
+          NULL class_level. NULL when no signal is available. Specific degree
+          titles (MBA, JD, etc.) are not resolved here — NSC only emits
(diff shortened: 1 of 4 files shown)


anthonygwalters:
@claude values in warehouse re: finding 1:

pursuing_degree_type
Certificate
Bachelor's (4-year)
High School Diploma
Associate's (2 year)
Graduate Degree
Course Credit (Non-Degree Seeking)
Master's
GED
MBA

Issue #4213, opened 2026-06-17
feat: Focus SIS (KIPP Miami) — Phase B: staging + intermediate models + kipptaf integration


cbini:
Phase B of the Focus SIS integration for KIPP Miami. Sub-issue of #3584.

## Scope

- `stg_focus__*` staging models (one per populated Phase 1 source table; contract-enforced, PK uniqueness tests, curated columns + descriptions)
- `int_focus__*` intermediate models — currently only `int_focus__student_enrollment` is buildable; attendance / schedule / gradebook / report-card intermediates depend on empty tables tracked in #4220
- Custom-field unpivot model(s) — wide tables (`students`, etc.) carry curated core columns; `custom_*` fields unpivot to long via the `custom_fields` crosswalk
- kippmiami: surface Focus models via the local `focus` package (package already wired)
- kipptaf: region source + `stg_kippmiami__focus__*` wrappers — split out to #4257

## Status

Staging + intermediate layer complete: batches A–I and `int_focus__student_enrollment` are all merged. The kipptaf-level integration has been split out to #4257 (needs its own design pass). The encoded-custom-value decode is a separate Wave 2 follow-up. 54 of 76 configured tables populated; the 22 empty tables are tracked in #4220.

## Phase B plans

Each item = one plan doc in `docs/superpowers/plans/` + one PR. Curation anchor, custom-field handling, PK/soft-delete maps, and batch contents are in the Phase B plan index (`docs/superpowers/plans/2026-06-18-focus-phase-b-index.md`).

- [x] C — Attendance lookups (4)
- [x] D — Courses & scheduling (8)
- [x] E — Grades & standards (14)
- [x] F — Discipline lookups (2)
- [x] H — Custom fields (6)
- [x] I — Test history (5)
- [x] A — Students / people (3 tables)
- [x] B — Enrollment & org (7) + `int_focus__student_enrollment`
- [x] G — Users & permissions (5)
- [x] kipptaf integration — split out to #4257 (needs a separate design pass)

## Design spec

[docs/superpowers/specs/2026-04-03-focus-sis-integration-design.md](https://github.com/TEAMSchools/teamster/blob/main/docs/superpowers/specs/2026-04-03-focus-sis-integration-design.md)

Refs #3584

cbini:
## Spec review (2026-06-18) — Phase B is unblocked

[...]

PR #4244, opened 2026-06-22
feat(dbt): add Focus courses & scheduling staging models (Batch D)

cbini:
## Summary & Motivation

When merged, this pull request will add contract-enforced `stg_focus__*` staging models for the eight populated Focus courses & scheduling tables: `course_subjects`, `course_weights`, `course_code_directory`, `co_teacher_days`, `resources`, `courses`, `master_courses`, `course_periods`.

Each model does light cleaning only — explicit column projection (drops `_dlt_*` bookkeeping and the `created_by_*`/`updated_by_*` audit-quad), a `unique` + `not_null` test on the table PK at `severity: error`, and a `description:` on the model and every column. The four wide tables (`course_subjects`, `courses`, `master_courses`, `course_periods`) additionally project an inline block of the currently-populated custom columns, each aliased to its slug. Custom `select`/`multiple` values are left as stored Focus option codes — decoding is a deferred follow-up.

Part of Focus Phase B. Refs #4213.

### AI Assistance

AI-authored under human direction. Each model was built and its PK tests verified locally via `uv run dbt build --select <model> --project-dir src/dbt/kippmiami`.

### dbt

- [x] Includes a `[model_name].yml` properties file for all new models
- [ ] Exposures — none consumed by a dashboard/app yet (staging layer only)
- [x] Not a breaking change — net-new models, no contracted columns renamed/removed
- [x] Source unchanged — Focus dlt tables are already declared in `src/dbt/focus/models/staging/sources-bigquery.yml` (BQ-native, not an external sheet), so `stage_external_sources` is N/A

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes


cbini:
Reviewed the bot feedback; CI is green on a9e9212 (dbt Cloud success, Trunk Check / CodeQL / claude-review all passing) and stays green after the doc push.

### Important — `course_subjects` / `courses` grain vs. single-column PK

Fixed in 0ad8274. Confirmed the single-column PK test is correct and left it unchanged: the focus CLAUDE.md documents `course_id` as the table's natural PK, and Focus rollover assigns a new id each year while recording the prior id in `rollover_id`, so `subject_id` / `course_id` are globally unique across years rather than recycled. The misleading part was the "scoped by school year and school" phrasing, so I rewrote both model descriptions to state that the id is globally unique across years (via rollover) and is therefore the PK even though each row also carries its school year and school. No test change — switching to a composite `unique_combination_of_columns` would over-test a key that is already unique on its own and the build passes today.

### Minor 1 — `resources.type` missing `quote: true`

Declining. `TYPE` is not a strict DML reserved word in BigQuery, the SQL projects it unquoted without error, and the existing `stg_focus__marking_periods.sql` (line 9) already uses `type` unquoted with no `quote: true`. No model in the focus project sets `quote: true` for this column. The dbt SQL-style convention only requires backtick-quoting for reserved words, so adding it here would introduce a pattern the codebase does not follow. Matches the existing convention as-is.

### Minor 2 — identical custom-column descriptions across the four wide models

Declining (no change). As the review notes, this is accurate: the focus CLAUDE.md confirms the positional `custom_N` / `custom_field_N` slots on these tables are not present in the extracted `custom_fields` catalog (their labels live in Focus config, not the extract), so the shared boilerplate is correct rather than overlooked.

Observations were all marked correct and need no action.

PR #4246, opened 2026-06-22
feat(dbt): add Focus discipline-lookup staging models (Batch F)

cbini:
## Summary & Motivation

When merged, this pull request will add contract-enforced `stg_focus__*` staging models for the two populated Focus discipline-lookup tables: `referral_actions` and `referral_codes`.

Light cleaning only — explicit column projection (drops `_dlt_*` and the audit-quad; neither table has `uuid` or a soft-delete column), a `unique` + `not_null` test on the `id` PK at `severity: error`, and a `description:` on the model and every column. `referral_codes.priority` is typed `numeric` (not `int`/`float64`).

Part of Focus Phase B. Refs #4213.

### AI Assistance

AI-authored under human direction. Both models were built and their PK tests verified locally via `uv run dbt build --select <model> --project-dir src/dbt/kippmiami`.

### dbt

- [x] Includes a `[model_name].yml` properties file for all new models
- [ ] Exposures — none consumed by a dashboard/app yet (staging layer only)
- [x] Not a breaking change — net-new models, no contracted columns renamed/removed
- [x] Source unchanged — Focus dlt tables already declared in `sources-bigquery.yml`, so `stage_external_sources` is N/A

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes


cbini:
Reviewed the bot pass — appreciate it. CI is green on the head commit: dbt Cloud status `success`, Trunk Check `success`, CodeQL `success`, claude-review `success`.

Both minor items, evaluated point-by-point. Declining both, with reasoning:

**Minor 1 — block-continuation descriptions (`withdraw_code`, `mobile_app`): declining.**
These two are not a hand-introduced inconsistency — they're trunk-fmt's enforced wrapping. I edited both to the inline form and ran `trunk check`; trunk-fmt auto-reverted them back to the continuation form because each inline description exceeds the 80-char line limit (81 and 84 chars). The other columns are inline only because they're short enough to fit on one line. The block style is mechanically enforced by the repo formatter, so normalizing them would just bounce on the next commit. No change.

**Minor 2 — `sesir` description precision: declining.**
Verified against the source. In `dagster_kippmiami_dlt_focus.referral_codes`, `sesir` is `INT64`, and the stored values are integer only (`1` and `NULL` — no alpha category codes like `BU`/`DR`). So `data_type: int` is correct and "SESIR (state incident reporting) code id" accurately describes an integer column. No change.

**Minor 3 — `type` unquoted: no action (matches your read).**
Confirmed pre-existing: `referral_actions.sql` projects `type,` unquoted, identical to `stg_focus__attendance_codes.sql` on main, and no `quote: true` or backtick-quoted `type` exists anywhere in the Focus staging layer. Not introduced here.

No commits pushed.

PR #4256, opened 2026-06-23
feat(dbt): add Focus users and permissions staging (Batch G)

cbini:
## Summary and Motivation

Adds contract-enforced `stg_focus__*` staging for the five populated Focus users/permissions tables (Batch G of the Focus SIS Phase B rollout): `stg_focus__users`, `stg_focus__user_enrollment`, `stg_focus__user_profiles`, `stg_focus__permission`, `stg_focus__login_history`. Each reads only its `focus` source; built and tested via the `kippmiami` project. Refs #4213.

Light cleaning per staging conventions: explicit column projection (drops `_dlt_*` and the audit-quad), soft-delete filter where present (`users` only — `where deleted is null`), a `unique` + `not_null` PK test at `severity: error`, and descriptions on every model and column.

### Deviations from the plan (verified against live Miami data before writing)

- **`stg_focus__users` PK is `staff_id`, not `profile_id`.** `profile_id` is null in 179 of 181 rows (1 distinct value) and would fail `not_null`; `staff_id` is unique and non-null across all rows. The plan, the Phase B index, and `focus/CLAUDE.md` all documented `profile_id` — corrected here.
- **Dropped the `password` column** from `stg_focus__users` entirely (raw credential — excluded from the projection, matching the Batch A students/address precedent). Other auth columns (`desktop_token`, `password_token`, `force_password_change`) were never in the projection.
- **Added two populated native columns** the plan omitted from `stg_focus__users`: `erp_display_name` (tagged PII) and `imported`.
- **PII tagging** (`config.meta.contains_pii: true`) applied on the broad Batch A scope — direct + indirect identifiers across `users`, `user_enrollment`, and `login_history`. `user_profiles` and `permission` are role / permission-definition lookups (no personal data) and carry no tags.

The 10 populated `custom_*` columns on `users` are inline-aliased to their `custom_fields` catalog slugs; select/multiple values remain encoded (decoding deferred). Every other custom field (including the semantic-named ones like `birth_date`, `race_*`, `charter_*`) is unpopulated in Miami and out of scope.

Verification: all 5 models build and 10/10 tests pass locally; `trunk check` clean on all 10 files.

## AI Assistance

Authored by Claude Code (executing-plans skill). Human-directed: the plan, the worktree/extra-column scope decisions, and the PII-tagging scope were directed and approved by the requester. Live-data verification (PK validity, custom-field population, contract types, STRUCT-collision check) was AI-run.

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — one per model, contract columns + types + descriptions
- [ ] Exposure — n/a (no dashboard/application consumes these yet)
- [ ] Breaking change — no (new additive models; no contracted columns renamed/removed)
- [ ] External source — n/a (BQ-native `focus` sources already declared in `sources-bigquery.yml`)

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes


cbini:
Review dispositions (Batch G):

**1. `focus/CLAUDE.md` PK note — fixed** in `189674c7`. Corrected `users→profile_id` to `users→staff_id` (with a note that `profile_id` is null for nearly all rows). The same stale note exists in the Phase B index plan doc; left there as a historical record (can update if preferred).

**2. `permission.sql` single-line SELECT — won't change.** That single-line form is trunk-fmt's output, not a style choice: the model fits within 88 chars so the formatter collapses it, while the longer models stay multi-line because they don't. The pre-commit `fmt` hook re-collapses any manual expansion, and the repo convention is to not run `trunk fmt` manually. `trunk check` passes on it as-is.

**3. `login_history.date` PII tag — keeping, by directive.** Tagging the login timestamp was an explicit requirement for this batch (login_history treated as staff-PII-heavy, login timestamps named). This repo's staging models only carry a per-column `contains_pii` mechanism — there is no model-level PII convention in use to express row-level sensitivity. Flagged to the requester as a judgment call; will narrow if they prefer.

**4. SSN passthrough — acknowledged, no change here.** `social_security_number` stays projected and PII-tagged, consistent with the Batch A students precedent. Noted for the intermediate-layer design that SSN warrants stricter column-level access / masking before any mart or export consumes it.

dbt Cloud CI: success, no warnings.

cbini:
Update to disposition #4 (SSN): per maintainer request, `social_security_number` (`custom_556`) is now dropped from `stg_focus__users` entirely — excluded from the projection like `password`, rather than projected-and-tagged. Commit `ea1fd219c`. The model description notes the exclusion; the other 9 populated custom fields are unchanged.

Also pushed `c649c3fd` — CLAUDE.md session learnings (Focus custom-field profiling gotcha, GitHub MCP body-encoding, claude-review trigger scope).

Issue #4268, opened 2026-06-25
feat(cube): strip inaccessible members in queryRewrite instead of blocking queries


cristinabaldor:
## Problem

Cube's `access_policy` `member_level` security blocks an entire query when a user requests a field they don't have access to. For example, a teacher (no `staff-pii`) querying `staff_detail` with `birth_date` included gets a query denial rather than a result set with that column absent.

This creates a poor experience for BI tool consumers — particularly Tableau, where the publisher's field picker reflects their own (broader) access, and viewers with narrower access hit errors at runtime. It also means any mixed-audience dashboard (some viewers have `staff-pii`, some don't) requires separate views or workbooks per audience tier, which is the maintenance burden we're trying to avoid.

## Desired behavior

When a user's query includes members they don't have access to, `queryRewrite` should silently strip those members before execution rather than letting the query reach Cube's access policy gate and fail. The query runs and returns results for the members the user _can_ see; inaccessible members are simply absent from the response.

This mirrors how row-level filters already work in `queryRewrite` — they narrow the result transparently rather than blocking the request.

## Proposed approach

Extend `queryRewrite` in `cube.js` to inspect `query.dimensions` and `query.measures` against the user's resolved groups, and strip any members the user's `access_policy` would block before the query executes. The `access.js` helper layer already detects staff PII members (`STAFF_PII_MEMBERS`, `STAFF_SENSITIVE_SCOPE_BY_MEMBER`) and could be extended with a `stripInaccessibleMembers(query, groups)` helper that handles both student and staff tiers.

`access_policy` entries in the view YAMLs would remain as-is — they serve as the documented source of truth for what each tier can see, and as a backstop for any query path that bypasses `queryRewrite`.

## Context

- Refold-c security redesign: `dim_staff_cube_access` access rows, `access.js` helpers, `staff-directory`/`staff-pii` tiers
- Affects any BI tool consuming `staff_detail` where viewers have mixed `staff-pii` access
- Superset (SQL API path) handles this more gracefully than Tableau since field lists are filtered per-user at connection time — but stripping in `queryRewrite` would make all BI paths consistent

cristinabaldor:
## Why the asymmetry between student stripping and staff gating

The code review on #4269 flagged that the two domains handle "no access" differently in `queryRewrite`, and it's worth documenting the nuance here.

[...]

PR #4269, opened 2026-06-25
feat(cube): refold-c security redesign — HR-derived access, open directory, per-field scopes

### LINKED ISSUE #4102: feat(cube): cube security model redesign
Redesign the Cube semantic layer security model and naming convention. All access control is declared inside Cube (`cube.js` + view YAML). Group membership is derived from a new dbt model (`dim_staff_cube_access`) queried at request time — no Google Group enrollment, no Google Admin Directory API.

Supersedes #4114 (closes as duplicate). All subsequent domain plans depend on this.

## Motivation

The current model has two problems:

1. **Manual group enrollment.** Location scope and access level are encoded in Google Group names (e.g. `cube-region-newark-detail`). Membership is maintained by IT in Google Workspace — it drifts, requires tickets, and is invisible to the data team.
2. **Static cube arrays.** `STUDENT_CUBES`/`STAFF_CUBES` in `cube.js` are maintained by hand. Every new cube requires a code change to register it in the right access tier.

### `cube.js` changes

- Drop `STUDENT_CUBES` / `STAFF_CUBES` static arrays
- Add `isStudentMember(member)` / `isStaffMember(member)` prefix helpers — a cube is student-domain if its name starts with `student_`; staff-domain if it starts with `staff_`
- Add `withSyntheticGroups(groups, members)` — emits `detail-access` / `summary-access` synthetic groups for view access policies based on the user's resolved scope tier
- Replace `contextToGroups` Google Admin Directory API call with a BigQuery query against `kipptaf_marts.dim_staff_cube_access` — returns the user's scope tier and access flags as a group list; cached until midnight Eastern per email
- `CUBE_GROUP_MAP` (local dev only, gated on `NODE_ENV !== "production"`) is the sole bypass — unchanged
- `queryRewrite` updated to use `isStudentMember` / `isStaffMember` helpers instead of array lookups

### New dbt model: `dim_staff_cube_access`

**Location:** `src/dbt/kipptaf/models/marts/people/dim_staff_cube_access.sql`

One row per staff email. Columns drive `contextToGroups` directly:

| Column | Type | Description |
| --- | --- | --- |
| `google_email` | STRING | PK; matches `securityContext.email` |
| `scope_level` | STRING | `network`, `region`, or `school` |
| `scope_key` | STRING | NULL for network; `region_key` for region; school `abbreviation` for school |
| `access_tier` | STRING | `detail` or `summary` |
| `has_student_data` | BOOL | May query student-domain cubes |
| `has_student_pii` | BOOL | May see student PII fields |
| `has_staff_pii` | BOOL | May see staff PII fields |
| `has_staff_compensation` | BOOL | May see salary/pay rate fields |
| `has_staff_all` | BOOL | Bypasses org-hierarchy filter on staff cubes |

Derived from HR data (work assignments, job classifications, location hierarchy). No manual enrollment required.

### Cube renames (breaking — existing cubes must be updated)

All existing cubes with `dim_` or `fct_` prefixes are renamed to unprefixed or domain-prefixed names. **Any Cube Cloud staging environments or downstream tool connections using old cube names will break.**

| Old name | New name |
| --- | --- |
| `attendance` | `student_attendance` |
| `dim_student_ell_status` | `student_ell_status` |
| `dim_student_iep_status` | `student_iep_status` |
| `dim_student_meal_eligibility_status` | `student_meal_eligibility_status` |
| `dim_staff` | `staff` |
| `fct_staff_attrition` | `staff_attrition` |
| `fct_staff_observations` | `staff_observations` |
| `dim_dates` | `dates` |
| `dim_locations` | `locations` |
| `dim_regions` | `regions` |
| `dim_terms` | `terms` |
| `dim_school_calendars` | `school_calendars` |

Attendance views rename accordingly: `attendance_detail` → `student_attendance_detail`, `attendance_summary` → `student_attendance_summary`. All join references within YAML files must be updated to match.

### Schema validation test

`tests/cube/test_cube_schema.py`: cube names must not use `dim_` or `fct_` prefix — enforced as a CI check going forward.

### View access policies

Updated to `detail-access` / `summary-access` synthetic group pattern emitted by `withSyntheticGroups`. PII tiers use `cube-access-student-pii` / `cube-access-staff-pii` / `cube-access-staff-compensation` group names — resolved from `dim_staff_cube_access` flags, not Google Groups.

## Security model summary

`contextToGroups` queries BigQuery once per request (cached until midnight ET). It returns a synthetic group list derived from `dim_staff_cube_access`:

- Scope: `cube-network-detail`, `cube-region-{key}-summary`, `cube-school-{slug}-detail`, etc. — same naming convention as before, now computed not enrolled
- Access: `cube-access-student-data`, `cube-access-student-pii`, `cube-access-staff-pii`, `cube-access-staff-compensation`, `cube-access-staff-all` — from boolean flags on the dbt model

`queryRewrite` and view `access_policy` blocks are unchanged in structure — they still match on `cube-*` group names. The only change is where those group names come from.

## Finding: member-level access policies intersect, not union

Investigated in PR #3755. Despite Cube docs stating OR/union semantics, `member_level` blocks **intersect** across groups in practice. A user in both `cube-access-staff-pii` and `cube-access-staff-compensation` gets neither — the intersection of the two `includes` lists is empty.

**Implication:** `contextToGroups` must resolve a single effective policy block per user — not multiple additive blocks. The `dim_staff_cube_access` approach handles this correctly: each user gets one resolved role with all applicable flags set, mapped to one `includes` list in the view access policy.

## Design spec

To be written on the feature branch: `docs/superpowers/specs/2026-04-17-cube-model-yaml-design.md`

cristinabaldor:
## Summary and Motivation

Replaces Cube's Google Admin Directory group-based security with an HR-derived BigQuery model, and enforces row-level security through **Cube-native per-view `access_policy`** driven by a server-enriched `securityContext`. Identity becomes a first-class data artifact — `dim_staff_cube_access` (a dbt mart from ADP work assignments and three Google Sheets crosswalks) plus `dim_staff_reporting_chain` (org-tree closure) — with no Google admin permissions required.

### How access works

- **Resolution in the auth hooks.** `cube.js` `resolveAccess(email)` reads the access row and reporting chain, precomputes the location-and-department remit into allow-list arrays, and builds a flat `securityContext`. It is called from `checkAuth` (REST/MCP — verifies the HS256 JWT against `CUBEJS_API_SECRET`, trusts only the `email` claim) and `checkSqlAuth` (SQL API). Fail-closed to an empty context on any error; cached to midnight ET.
- **`access.js`** (pure, unit-tested, 26 tests): `buildGroups` emits scope-specific groups — `student-region` / `student-school` / `student-network`, `staff-directory` (every resolved viewer), and one `staff-pii-*` per `staff_pii_scope`; `buildSecurityContext` shapes the flat context; `computeAllowedAbbreviations` / `computeAllowedDepartmentGroups` precompute the remit.
- **RLS lives entirely in per-view `access_policy`.** Each gated view declares `row_level` filters that interpolate `securityContext` values. A viewer holds exactly one scope-group per axis, so exactly one policy is ever active — no AND/OR combination to reason about — and a `none` scope emits no group, so the default-deny floor applies. `queryRewrite`'s RLS half is removed; only the snapshot-anchor guard remains.
- **Staff views split.** `staff_directory` (open roster and work-contact fields, no PII) and `staff_pii` (the six sensitive fields, gated per `staff_pii_scope` by a location-intersect-department remit precomputed into `securityContext.allowed_abbreviations` / `allowed_department_groups`, reused as-is when compensation/observations/benefits views are built).
- **Student views** carry three location-scope policies each (region / school / network).

Row-level policy `member:` references use flat view-member names (a cube-qualified path fails to compile), and scope selection is group-based because `conditions.if` cannot evaluate `==` in this Cube version.

### Validation

Live REST `/load` viewer matrix against a dev-schema build (8/8): a network viewer saw all rows (10,721 students / 1,443 staff); a region viewer saw exactly its region's slice; a school viewer exactly its school; `none`-scope and unresolved-JWT viewers were denied on both student and staff views.

### AI Assistance

Implemented with Claude Code via subagent-driven execution of the pivot plan (`docs/superpowers/plans/2026-07-07-cube-access-policy-pivot.md`). Human-directed: security-model design, scope semantics, the design forks (group-based vs conditions, precompute vs queryRewrite residual), and live validation. AI-assisted: `access.js` / `cube.js` / view YAML, docs, and unit tests.

### dbt

- [x] Properties file for all new models
- [x] Exposure for consumed models — N/A for this pivot (Cube-layer and docs only; no new consumed models)
- [x] `stage_external_sources --target staging` for the three cube-access sources — done in the base rework; the `access_policy` pivot changes no dbt models or `sources-external.yml`, so the staged externals still stand

### Docs

- N/A — no schedule, sensor, or integration changes. Cube docs (`src/cube/CLAUDE.md`, `docs/guides/cube.md` including a new local-RLS-testing section, and the reviewer guide) updated for the new model.

## CI checks

- [ ] **dbt Cloud** — re-running on the pivot push (no dbt model changes; expected to stay green)
- [x] **Trunk** — passes
- [x] **Dagster Cloud** — not triggered

## Follow-ups

Non-blocking follow-ups tracked in #4342 (a dbt invariant hardening the empty-allow-list path, minor `cube.js` polish, dead-code trim, and a Cube Cloud `securityContext`-vs-`userAttributes` confirmation).


cristinabaldor:
## Review response

**Addressed (4 commits)**

- **Finding 2** — added `not_null` + `unique` at `severity: warn` on `google_email` in `dim_staff_cube_access.yml`
- **Finding 1** — documented the role crosswalk fan-out invariant in `dim_staff_cube_access.sql` header (wildcard + specific rows for the same `job_function_code` must not coexist; the existing `staff_key` unique test catches it at CI but the constraint is now explicit for sheet authors)
- **Finding 3** — documented the isolated-node gap in `dim_staff_reporting_chain.sql` (staff with no reporting relationships get no self-pair; `reporting_chain` scope denies them silently; debug via `dim_work_assignment_reporting_relationships`)
- **Finding 8** — added `job_function_code__code_value` to the `dim_work_assignment_jobs` SCD2 `attribute_hash`; no downstream consumers join on `work_assignment_job_key` yet so no hash churn

**Not taken / deferred**

- **Finding 4** (`timeDimensions`/`filters` not stripped for no-student-access users) — known limitation, already tracked in #4268
- **Finding 5** (inline multi-table JOIN in `staff.yml`) — intentional; added a comment explaining why these access-control attributes live in `dim_staff_cube_access` rather than `dim_staff` and why the join cannot be pre-materialized without coupling two distinct concerns (same reasoning as the documented SCD2 range-join pattern)
- **Finding 6** (two untested paths in `access.test.js`) — minor; deferred to a follow-up test pass
- **Finding 7** (`test_cube_schema.py` group name convention guard) — minor; deferred

cristinabaldor:
@claude update your review?

cristinabaldor:
@claude update review

kevinverhoff (dismissed):
This is really cool to see. Nice work!

cristinabaldor:
@claude update your review

cbini (changes requested):
## Code review — Cube security redesign

Reviewed the full diff (base `22e7461` to head `8febbc4`) across four surfaces — the `cube.js` auth core, the pure `access.js` logic, the per-view `access_policy` RLS, and the dbt access models — plus current CI state. **No Critical defects and no auth bypass:** the two load-bearing invariants hold — JWT auth trusts only the `email` claim (no client-injected scope), and every traced error path fails closed to an empty default-deny context. Requesting changes for the items below: one blocking convention violation (the PR now contradicts its own hardened rule), two decisions that need an owner, plus should-fix hardening. None of it is a runtime bug or auth bypass.

~### CI status — the red dbt Cloud check is NOT this PR's code~

~dbt Cloud run `70403216870734` fails on two models: `stg_collegeboard__ap` (enforced-contract: `source_file_name` missing in contract) and `int_extracts__student_enrollments` (`contact_1_email_current` not found in `base_powerschool__student_enrollments`). **Neither model is touched by this PR** — both are stale-base drift. This branch is 86 commits behind `main` (last merge 2026-07-08) and is missing `main`'s July fix for the collegeboard `source_file_name` contract. **Fix: merge `origin/main` into the branch and re-run** — that clears both. Trunk and CodeQL are green.~

### Blocking (must fix before merge)

- **Inline `JOIN` in `staff.yml` — now forbidden by this PR's own convention.** The `staff` cube uses `sql: SELECT s.*, ca.job_function_level, ... FROM dim_staff LEFT JOIN dim_staff_cube_access` to pull three access attributes onto the cube. This PR *also* hardens `src/cube/CLAUDE.md` to a rule that a cube's `sql:` reads exactly one dbt model — no `JOIN`, no `SELECT t.*` — so as written the PR ships code that violates a convention it itself introduces (the range-join recipe the code comment cites applies to a *join's* `sql:` predicate, not a cube-body `sql:`). `SELECT s.*` is separately a latent break: if `dim_staff` ever gains a `job_function_level` / `job_function_code` / `department_group` column, the `s.* + ca.*` collision fails the staff cube's compile and takes down both `staff_directory` and `staff_pii` on deploy. Fix: expose `dim_staff_cube_access` as its own `public: false` cube (surfacing only those three dimensions, scope columns unexposed) and bring them in via a Cube join on `staff_key`; both staff views consume the three, so the join serves both.

### Two items that need a decision (not a code fix)

1. **Student direct-identifier PII tier was removed — the single largest access change, and it's invisible in the view diffs alone.** Previously student names/DOB/`state`/`lea` identifiers sat behind a separate `cube-access-student-pii` group. The new `student_*_view.yml` files expose all fields to anyone with a student location scope (and `student_assessment_scores_view` additionally surfaces `district_student_identifier` and `salesforce_contact_id`). Location scope is now the *only* control on student identifiers. It's intentional and documented in CLAUDE.md and the reviewer guide — but given FERPA obligations it warrants explicit data-governance sign-off recorded on this PR, not silent acceptance.

2. **Empty-allow-list fail-closed rests on an unverified Cube runtime behavior.** `access.js` correctly emits `[]` on an empty remit, but end-to-end safety then depends on Cube compiling `equals []` into `IN ()` / zero rows rather than dropping the filter (the exact behavior deferred to #4342). The reachable trigger is a viewer with `staff_pii_scope` other than `none` but `staff_location_scope = none` (a plausible spreadsheet mis-entry), guarded today only by a `severity: warn` dbt test. The 8/8 live matrix validated the `none`-scope case (no group emitted), not this active-group-empty-remit case. **Before merge:** run one SQL-API check with such a viewer and confirm the compiled SQL is `WHERE (1=0)`; if Cube drops the predicate instead, that viewer sees all staff PII network-wide.

### Important (should fix before merge)

- **`checkSqlAuth` fails *open* on an unset `CUBEJS_SQL_PASSWORD`.** `cube.js` returns the `CUBEJS_SQL_PASSWORD` environment variable's value unconditionally; if that variable is ever blank, Cube validates presented passwords against `undefined` and may accept blank-password connections. The REST path fails closed (an unset secret makes `jwt.verify` throw a 403). Fix: return `null` / throw when the SQL password is absent.

- **dbt role-crosswalk join can fan out to a duplicate `staff_key`, guarded only by a data test.** `dim_staff_cube_access.sql`'s `LEFT JOIN cube_access_role ON job_function_code = ... AND rl.entity IN ('any', e.entity)` matches both a wildcard (`any`) and a specific (`KTAF`/`Region`) row if the sheet ever holds both for one code, producing two conflicting scope rows for one staff member. The sheet-level `unique_combination_of_columns` does not prevent this. `dim_staff_cube_access` is materialized as a *view* (confirmed in prod), so `cube.js`'s `SELECT * ... ORDER BY staff_key LIMIT 1` (both rows share the same `staff_key`) silently picks one nondeterministically rather than erroring — the `unique` test is the only detection, and only when it next runs. Fix: make the specific-over-`any` pick structural (rank/`coalesce` before the join).

- **`entity` fallback grants on unknown/NULL `business_unit_name`.** `if(business_unit_name = 'KIPP TEAM and Family Schools Inc.', 'KTAF', 'Region')` routes every non-KTAF value — including a NULL from an unresolved org unit — to `Region`, a granting classification. Fail-toward-grant on the identity axis. Fix: derive `entity` from an explicit allow-list and emit a deny sentinel for unrecognized/NULL business units, and/or force location scope to `none` when `region_key`/`location_abbreviation` is NULL.

- **SQL-API impersonation via the shared password.** In production `email = user` (the connecting username, verbatim), authenticated only by the single shared `CUBEJS_SQL_PASSWORD`; `canSwitchSqlUser` gates only a post-connection SET-USER, not the initial connection. This is Cube's documented model and fails closed on identity resolution, but the whole boundary reduces to that one secret. Fix: confirm the SQL port is reachable only from the Superset integration (network policy) and that the password is strong and rotated.

- **Security-boundary test gaps.** `cube.test.js` covers expiry/`maxAge`/`iat` well but is missing the attacks this PR defends against: an `alg:none` token rejected, a wrong-secret token rejected, and a forged `groups`/`securityContext` claim ignored (this locks the only-email-trusted invariant). `checkSqlAuth` has zero tests, and all current tests route through the `CUBE_GROUP_MAP` dev bypass — the production BigQuery `resolveAccess` path and its fail-closed catch are never exercised. `access.js` also has untested branches (`buildGroups({})`, unknown-scope-denies, empty-universe department path).

- **`buildGroups` grants `staff-directory` to any truthy object (even `{}`).** `access.js` guards only `if (!row)`. The wired caller passes `null` for not-found so this isn't live, but for the module whose whole job is to be independently correct, gate the push on a real identity field (e.g. `if (row.staff_key)`).

### Minor / defensive

- `computeAllowedAbbreviations` compares `regionKey` with `===`; a location with a null `region_key` would match a null-region viewer (same root as the `entity`-NULL item). Guard `regionKey` null to `[]`.
- Scope columns have `accepted_values` but no `not_null` (accepted_values passes NULLs silently); `entity` and `department_group` have no test.
- Doc/code mismatch: comments reference `securityContext.email`, but `resolveAccess` never sets it (latent — fix before a future policy interpolates an always-undefined field).
- No `clockTolerance` on `jwt.verify` (5-min tokens can spuriously 403 on small clock skew — fails closed). No-token REST requests get 200 + empty context rather than 403 (data protected, but inconsistent with the invalid-token 403).
- Stale "summary" references in `access.js` comments (the open summary view was removed). Unit-test count is 20, not the 21 stated in the description.
- `staff_directory` now exposes `job_function_level` / `job_function_code` / `department_group` network-wide (role metadata, not PII) and is open to every resolved staff member — a deliberate shift from the old location-scoped staff RLS. Flagging for awareness.

### Done well

- JWT verification pins HS256, enforces `exp` plus an `iat`-derived `maxAge`, and discards every payload field except `email`; the `mcp/server.py` minter was correctly updated to add the required `iat`.
- Fail-closed is uniform across null/absent email, no row, malformed row, and BigQuery errors; errors are not cached, so recovery is immediate.
- PII containment is clean: the six sensitive staff fields appear only in the gated `staff_pii` view, never in the open `staff_directory` (verified field-by-field).
- `access.js` is default-deny throughout and always returns real `[]` arrays; the dbt spine is genuinely 1:1 on `staff_key` (test-enforced), scopes coalesce to `none` on no match, and the recursive reporting-chain closure is depth-bounded. The `dbt_utils.deduplicate` usages are guide-compliant here (TODO-tagged), not anti-patterns.

### Assessment

Request changes. **Blocking:** fix the `staff.yml` inline JOIN so the PR stops violating the `src/cube/CLAUDE.md` rule it adds (expose `dim_staff_cube_access` as a cube + join on `staff_key`). Also before merge:
~1. merge `main` to clear CI,~
2. governance sign-off on the student PII de-tiering, 
3. one-time runtime confirmation that an empty allow-list compiles to `IN ()`, and 
4. close the two `dim_staff_cube_access` seams and the `checkSqlAuth` unset-password guard. 

Security-boundary tests are strongly recommended pre-merge; the remaining Minor items can follow. The PR is also `mergeable_state: blocked` pending the required CODEOWNERS (analytics-engineers) review on `src/dbt/`.

*Review produced with multi-agent assistance (four scoped reviewers plus independent verification of the auth path, the group-name contract, PII containment, and CI root cause).*

cristinabaldor:
## Review response

Verified every finding against the code and implemented the fixes (just pushed). Summary by section; all of it is covered by tests (39 JS unit tests pass), `dbt parse` is clean, and trunk is green locally.

### Blocking — `staff.yml` inline JOIN

Fixed, per your `fd37aec94` one-model rule. `dim_staff_cube_access` is now its own `public: false` cube (`staff_cube_access`) exposing only `job_function_level` / `job_function_code` / `department_group`; `staff` is a thin `sql_table: kipptaf_marts.dim_staff` with the three attrs brought in via a Cube join.

Implementation note: I put the join on **`staff_work_history`**, not `staff`. On `staff` it broke `staff_directory` — `staff_manager` `extends staff`, so it inherited the join and created an ambiguous second path to `staff_cube_access` (the attrs came back null). On `staff_work_history` (nothing extends it) both views resolve — verified on the dev server. It is `many_to_one` on `staff_key`; current-role semantics are identical to the old inline LEFT JOIN.

One thing to flag on `fd37aec94`: as written ('a cube's `sql:` reads exactly ONE dbt model — never a `JOIN`') it also bans `staff_work_history` and `staff_reporting_relationships`, which carry multi-table **SCD2 period-intersection** joins in their cube-body `sql:` — `greatest()`/`least()` collapsing several effective-dated child dims into one contiguous-window grain. Cube genuinely can't express that as a cube-to-cube join (it is many-to-many and row-synthesizing). I've **proposed an edit** to that CLAUDE.md bullet adding an explicit SCD2-period-intersection exception naming those two cubes, keeping your strict default for the `staff.yml`-style plain-join case. Flagging for your call — if you'd rather migrate those two to dbt marts instead, say so and I'll drop the carve-out.

### Decisions

- **Student PII de-tiering** — confirmed intended by the data team: single-gate (location scope) is the design; network/region-scoped viewers see student identifiers by design. Recording that sign-off here.
- **Empty-allow-list behavior** — verified on the dev server, and the documented behavior was wrong: Cube (Tesseract) does not compile `equals []` into `IN ()` / zero rows — it throws 'Values required for filter' and fails the query. Still fail-closed (I confirmed an `all_in_scope` + `staff_location_scope=none` viewer errors rather than returning rows — no leak), but via error, not clean deny. So I (a) hardened `access.buildGroups` to not emit a `staff-pii` group whose remit/chain resolved empty — routing that viewer to the clean no-group default-deny path and removing the dependency on Cube's empty-array behavior entirely; (b) tightened the dbt guard `warn` to `error`; (c) corrected the CLAUDE.md claim. 0 viewers are in the offending state today. Follow-up for the one case the probe did not cover (`reporting_chain_or_below_rank`'s nested empty-array branch; 0 viewers hold that scope): #4403.
- **SQL-API impersonation** — Cube's model, and an infra item I own rather than code: the SQL API is public (`cubecloudapp.dev`), password-gated, and the password is set-but-undistributed with no Superset wired yet. I'll network-restrict it at Superset time and keep the password strong/rotated.

### Should-fix — done

- `checkSqlAuth` returns `null` (rejects every connection) on an unset/blank `CUBEJS_SQL_PASSWORD`.
- Role-crosswalk fan-out: the specific-over-`any` pick is now structural (`role_picked` ranks a specific entity match ahead of the wildcard, one row per staff). Defensive — no wildcard+specific overlap in the crosswalk today.
- `entity` fallback: replaced `else 'Region'` with an explicit allow-list; unrecognized/NULL business units resolve to an `unknown` deny sentinel that matches only entity-agnostic `any` rows, never Region-specific grants. All current business units are allow-listed.
- Security-boundary tests: `alg:none` rejected, wrong-secret rejected, forged `groups`/`securityContext` ignored (only email trusted), `checkSqlAuth` unset/valid, and the production `resolveAccess` BigQuery fail-closed catch (mocked).
- `buildGroups` gates on `row.staff_key` — no grant to a keyless object.

### Minor — done

`computeAllowedAbbreviations` null-region guard; `not_null` on the scope columns plus an `entity` accepted-values test; corrected the `securityContext.email` comments; `clockTolerance` on `jwt.verify` plus a note that the no-token 200-vs-403 behavior is deliberate; `access.js` test count corrected in the PR description (26).

cbini:
## SCD2 carve-out — decision: migrate to dbt marts in this PR

Verified both halves of the claim before deciding. You're right that Cube's join graph can't express the period intersection — joins traverse existing rows and never synthesize them, many-to-many requires an associative cube mirroring a junction table that would itself have to *be* the intersection, and a join's `sql:` can't reference a third cube, so pairwise range-join chaining can't encode the mutual-overlap predicates either. The narrow technical claim stands.

But "cube-body `sql:` acceptable when every input is already a published dim" is a placement preference, not a necessity — and it carries two real costs the carve-out doesn't price in:

- **No lineage.** The cube SQL references `kipptaf_marts.dim_*` as raw strings — a rename or refactor of any of those six dims won't flag these cubes in dbt's graph; they break at query time instead of at `dbt build`.
- **No tests on the synthesized grain.** The intersection's invariants (windows don't mutually overlap, `effective_start_date` less-than-or-equal-to `effective_end_date`, one manager per period) are enforced nowhere today. The cube docs here already say join invariants "must be test-enforced upstream in dbt" — impossible while the grain is manufactured downstream of dbt.

Since the migration is mechanical (the SQL moves as-is; no recursion, no contract obstacle), fold it into this PR rather than grandfathering:

1. **Two mart models** (`kipptaf_marts`): move each cube-body `sql:` verbatim into a mart — suggest `dim_staff_work_history` and `dim_staff_reporting_periods` (avoids colliding with `dim_work_assignment_reporting_relationships`). Convert raw table refs to `ref()`. Filter the non-overlapping combinations (`effective_start_date` greater than `effective_end_date`) with a `where` clause in the mart instead of relying on the dates `BETWEEN` join to silently exclude them — that behavior is currently load-bearing and invisible.
2. **Properties + tests**: contract enforced (plain SQL, no recursion), `not_null` on the keys, `unique_combination_of_columns` on the grain (`work_assignment_key` + `effective_start_date`; `staff_key` + `effective_start_date`), `expression_is_true` for `effective_start_date <= effective_end_date`, and `relationships` back to the parent dims. A mutual-non-overlap test per key would be ideal; fine as a follow-up if it needs a custom test.
3. **Cubes become thin**: `staff_work_history` / `staff_reporting_relationships` keep their names and join graphs, `sql:` replaced by `sql_table: kipptaf_marts.<model>`. The dates `BETWEEN` range join stays in the join `sql:` — that's the sanctioned use.
4. **CLAUDE.md**: drop the exception block entirely; restore "(SCD2 period-intersection / status spines)" to the belongs-in-dbt list. The one-model rule then holds with zero exceptions, which is the whole point of having it.
5. Register both marts in `exposures/cube.yml` alongside the existing entries.

Everything else in your response is verified and looks good — the `staff_cube_access` join placement (confirmed the `staff_manager` extends-inheritance diamond; your placement is correct), the empty-array discovery (good catch on the docs being wrong — the `buildGroups` suppression is the right fix), and the should-fix/minor items all check out in the code, 39/39 tests passing locally. One nit for the same push: `staff_cube_access.yml`'s header comment still says "Joined 1:1 to `staff`" — update it to name `staff_work_history` and the extends rationale.

The change request stays open pending the migration; ping when pushed and I'll re-verify.

cbini:
## SCD2 cubes — decision: migrate to dbt marts in this PR (supersedes my hidden earlier comment)

Self-contained restatement after deeper verification; the earlier version of this comment is hidden.

### Verified: your impossibility claim is correct, but narrower than "no alternative"

Cube's join graph genuinely cannot produce the period-intersection grain — joins traverse existing rows and never synthesize them; many-to-many requires an associative cube mirroring a junction table (which would have to *be* the intersection — circular); join `sql:` is pairwise, so six-way mutual-overlap predicates have nowhere to live; and the point-in-time workaround (each child range-joined to `dates`) is a six-way diamond that leaves five timelines unconstrained. Views add nothing — they define no members of their own, and multi-fact blending can't constrain a measure (or an `access_policy` remit) across subquery trees.

So the real choice is cube-body `sql:` vs dbt mart — a placement decision, and the mart wins on grounds the carve-out didn't price in:

- **Lineage.** The cube SQL references `kipptaf_marts.dim_*` as raw strings; a refactor of any of the six inputs breaks these cubes at query time instead of at `dbt build`.
- **Tests.** The synthesized grain's invariants (windows don't overlap, start before end, one manager per period) are currently enforced nowhere.
- **Precedent.** The marts layer already materializes `greatest()`/`least()` period-intersections on the student side (`dim_student_ell_status`, `dim_student_iep_status`, `dim_student_meal_eligibility_status`). The staff analog is the missing piece of the same design, not a new pattern.
- **Deduplication, not duplication.** I checked for overlap with existing marts: none — no mart produces either target grain. Better: the reporting-periods logic currently exists in TWO cube files (`staff_reporting_relationships` and copy-pasted as the `mgr` subquery inside `staff_work_history`), and `dim_staff_reporting_chain` re-derives its current slice a third time. The migration collapses three copies into one layered derivation.

(`int_people__staff_roster_history` covers similar ground but is cross-layer legacy — flat ADP lineage, documented build-time-identifier caveats, not readable by cubes — so it is the surface this family eventually replaces, not a competitor.)

### Requested changes

1. **`dim_staff_reporting_periods`** (new mart): the `staff_reporting_relationships` cube-body SQL verbatim — `dim_work_assignment_primary` ∩ `dim_staff_work_assignments` ∩ `dim_work_assignment_reporting_relationships`, `greatest`/`least` windows — converted to `ref()`s. Grain: `staff_key` x effective range → `manager_staff_key`.
2. **`dim_staff_work_history`** (new mart): the `staff_work_history` cube-body SQL, with the inline `mgr` subquery replaced by a `ref()` to `dim_staff_reporting_periods`. Filter the non-overlapping combinations (start after end) with an explicit `where` — today the dates `BETWEEN` join silently drops them, which is load-bearing and invisible.
3. **Rebase `dim_staff_reporting_chain` onto `dim_staff_reporting_periods`** (current slice → transitive closure), so the edge derivation exists exactly once:

   ```text
   primary + assignments + reporting_rel → dim_staff_reporting_periods → dim_staff_reporting_chain
   status/jobs/types/orgs/locations ─────→ dim_staff_work_history ─────→ staff_work_history (thin cube)
   ```

4. **Properties + tests** for both marts: contract enforced (plain SQL, no recursion), `not_null` on keys, `unique_combination_of_columns` on the grain (`work_assignment_key` + `effective_start_date`; `staff_key` + `effective_start_date`), `expression_is_true` for start-not-after-end, `relationships` to parent dims. A per-key mutual-non-overlap test is ideal; fine as a follow-up if it needs a custom test.
5. **Cubes go thin**: `staff_work_history` / `staff_reporting_relationships` keep their names and join graphs; `sql:` becomes `sql_table:`. The dates `BETWEEN` range join stays in the join `sql:` (sanctioned use).
6. **CLAUDE.md**: drop the SCD2 exception block entirely; restore "(SCD2 period-intersection / status spines)" to the belongs-in-dbt list. The one-model rule then holds with zero exceptions.
7. Register both marts in `exposures/cube.yml`; nit from before: `staff_cube_access.yml`'s header comment still says "Joined 1:1 to `staff`" — update to name `staff_work_history` and the extends rationale.

### Everything else from your response: verified, no further asks

The `staff_cube_access` join placement is correct (confirmed the `staff_manager` extends-inheritance diamond — my suggested placement would have broken `staff_directory`); the empty-array discovery and `buildGroups` suppression check out (good catch on the docs being wrong); all should-fix and minor items verified in code; 39/39 tests pass locally; CI green on both surfaces after your main merge.

Change request stays open pending the migration; ping when pushed and I'll re-verify.

cristinabaldor:
## Migration pushed — SCD2 period-intersection moved to dbt marts

All seven requested changes are in (executed the plan at `docs/superpowers/plans/2026-07-17-cube-scd2-marts-migration.md`). Commit map:

1. `dim_staff_reporting_periods` mart — `ba503d7d`
2. `dim_staff_work_history` mart — `a6c69b68`
3. Rebase `dim_staff_reporting_chain` onto reporting_periods — `f25da9c8`
4. Slim both cubes to `sql_table:` (names/joins/dims/measures unchanged; dates BETWEEN join stays) — `4c15725c`
5. Register both marts in `cube.yml` + fix `staff_cube_access` header — `9f867f34`
6. Drop the SCD2 exception in `src/cube/CLAUDE.md` (one-model rule now zero-exceptions) — `ce69f53f`
7. Bonus: sqlfluff ST09 join-order — `7d22b840`; refreshed the `docs/guides/cube.md` access-groups section (it still described the old queryRewrite/single-tier model) + added a region-isolation dev-server test — `a2afc7c7`

### Two deviations from the letter of the ask (flagging explicitly)

- **Grain uniqueness uses `unique` on the surrogate PK, not `unique_combination_of_columns`.** `marts/CLAUDE.md` says drop the combination test when its columns equal the surrogate-key hash inputs — which they do (both PKs hash exactly the grain). Same enforcement, repo-conventional form. Happy to switch to the explicit combination test if you prefer.
- **Per-key mutual-non-overlap test deferred** — per your fine-as-follow-up note.

### One behavior change worth your eyes: the `wp` fold-in

The uniqueness test surfaced a latent fan-out that was already present in the original cube SQL: `dim_work_assignment_primary` (`wp`) was joined to read `is_primary_position` but its own effective window was NOT folded into the greatest/least intersection. So a period spanning a primary-status flip emitted two rows identical except `is_primary_position`, colliding on `(work_assignment_key, effective_start_date)` — 290 keys.

I folded `wp` into the intersection like the other five children, which collapses 290 collisions to 1, removes ~299 spurious duplicate rows, and makes `is_primary_position` period-accurate. Validated the new mart is a **strict subset** of the old cube output (0 rows added or changed; only the duplicates dropped), with identical distinct-staff counts (all-time 4,685 = 4,685; today 4,676 = 4,676). The single primary-headcount delta traced to a correct fix: it surfaced an assignment with no `dim_work_assignment_locations` row that the old fan-out was masking.

### The 1 residual PK warn is a real upstream data anomaly

The lone remaining collision is a genuine **overlapping dual-primary assignment**: one staff held two `is_primary_position=true` assignments overlapping ~5 weeks into termination, each with a different manager (the manager join keys on `staff_key`). Not a clock/UTC artifact — every effective column is `DATE`. Network-wide it is the only case, and the overlap is already in the past. Left as a warn + TODO (project-default severity) and filed upstream as #4431 (Tier 2 / work-assignment cleanup) with a proposed non-overlap check on `dim_work_assignment_primary`.

### Reporting-chain rebase is behavior-preserving

`dim_staff_reporting_chain` now reads the current slice of `dim_staff_reporting_periods`, restricted to currently-employed staff via `dim_staff_work_assignments.is_current` — reproduces the prior edge set exactly (1,593 direct edges). I kept the mart itself all-history and unopinionated (the current-employee filter lives in the chain, not the mart), so historical / point-in-time reporting analysis stays available for future fact cubes.

### Validation

- dbt: 3 models build; tests pass (1 documented warn).
- Cube JS 39/39, `test_cube_schema.py` 5/5, trunk clean.
- Live SQL-API RLS matrix on student views: region-scoped viewers saw only their region, a network viewer saw all four, a none-scope viewer was denied — confirms `resolveAccess` (which reads the rebased chain) still builds the right `securityContext`.

Note for re-verify: the two new marts are not in prod `kipptaf_marts` yet, so branch-staging Cube reads them only after merge + materialize — a live RLS re-run on the branch staging instance needs that first. dbt Cloud CI is running on the push.

Ready for re-review — ping me with anything.

cristinabaldor:

cristinabaldor:
Addressed the one review note in `ea8d4f44`.

**`dim_staff_work_history` residual warn → linked to #4431.** I also corrected the model's header comment: it previously described the residual as "concurrent reporting relationships," but the deeper trace showed it is actually an **overlapping dual-primary-assignment** anomaly (the #4431 record). The comment now names that root cause and points at #4431.

**`dim_staff_reporting_chain`'s dedupe is a distinct case — confirmed, not #4431.** Checked the current data: 0 reportees have more than one current manager (that dedupe's target case), and the #4431 record does not appear as a dual-manager in the chain (its terminated primary drops out of the current slice). So that dedupe guards concurrent CURRENT reporting relationships — dormant today (0 instances) and inherited via the rebase, not this anomaly. Left it untouched rather than mislabel it as #4431.

**CI was fully green on `ba4ba065`:** dbt Cloud status success (both new marts built in the `state:modified+ --full-refresh` run), plus CodeQL, Trunk, and all Analyze checks. The only dbt warning is the documented `dim_staff_work_history` PK (the single #4431 row). `ea8d4f44` is a comment-only change, so CI is re-running on it now and should stay green.

cbini (approved):

PR #4271, opened 2026-06-26
fix: add drifted avro schema fields for amplify mclass and ldap

cbini:
## Summary & Motivation

When merged, this PR clears the `avro_schema_valid` drift warnings on two sources by declaring the vendor fields that started appearing in the feeds, so the columns are captured to the warehouse instead of dropped on Avro write:

- **amplify mclass `pm_student_summary`** (Newark + Paterson, shared library model `PMStudentSummary`): adds `aimline_value_by_date`, `aimline_status`, `growth_goal_set`. The first two already exist on the sibling `PMStudentSummaryAimline` model, confirming they are genuine Amplify columns.
- **ldap `user_person`** (kipptaf, `UserPerson`): adds `extensionAttribute1` next to the existing `extensionAttribute13/14/15`.

All four are real, named vendor fields (not positional fillers), so updating the schema is the correct fix — `check_avro_schema_valid` has no allowlist mechanism, and one would be inappropriate for real fields. The single shared-library edit covers both Newark and Paterson (they generate from `PMStudentSummary`).

Surfaced by the daily Day-2 operations check.

## AI Assistance

AI-assisted: Claude Code investigated the drift warnings, identified the fields and their source, made the edits, and verified the generated Avro schemas now include all four fields via a runtime smoke test (PM 49 to 52 fields; LDAP 173 to 174, `extensionAttribute*` ordered 1/13/14/15). Human-directed: scope, the update-vs-allowlist decision, and approval to ship.

### General

### Dagster

- Verified locally: generated Avro schemas (`PM_STUDENT_SUMMARY_SCHEMA`, `USER_PERSON_SCHEMA`) build and include the new fields; `trunk check` clean on both files.
- [ ] Run `uv run dagster definitions validate` for the modified code locations (kippnewark, kipppaterson, kipptaf)
- [ ] Run `uv run pytest tests/test_dagster_definitions.py`

### dbt

- N/A — no dbt changes.

### Docs

- N/A — no schedule, sensor, or new-integration changes.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes or not triggered (no dbt paths changed).

### DIFF
--- a/.claude/skills/dagster-day2/scripts/day2_summarize.py
+++ b/.claude/skills/dagster-day2/scripts/day2_summarize.py
@@ -160,7 +160,7 @@ def _print_details(data: dict) -> None:
     failures = data.get("step_01_failed_runs", {}).get("failures", [])
     print(f"\n[step_01] failures ({len(failures)}):")
     for f in failures:
-        run_id = (f.get("runId") or "?")[:8]
+        run_id = f.get("runId") or "?"
         print(
             f"  {run_id} job={f.get('jobName')} loc={f.get('locationName')} "
             f"cat={f.get('category')} start={f.get('startTime')} end={f.get('endTime')}"
@@ -172,8 +172,8 @@ def _print_details(data: dict) -> None:
     retries = data.get("step_02_retries", {}).get("retries", [])
     print(f"\n[step_02] retries ({len(retries)}):")
     for r in retries:
-        parent = (r.get("parentRunId") or "?")[:8]
-        retry = (r.get("retryRunId") or "?")[:8]
+        parent = r.get("parentRunId") or "?"
+        retry = r.get("retryRunId") or "?"
         print(f"  parent={parent} retry={retry} status={r.get('retryStatus')}")
 
     sensor = data.get("step_03_sensor_ticks", {})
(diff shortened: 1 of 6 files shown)


cbini:
Dispositions on the review:

**[Important] aimline field order** — Aligned in d0a39ae72. `PMStudentSummary` now orders `aimline_status` → `aimline_value_by_date` to mirror `PMStudentSummaryAimline`. Cosmetic only (`check_avro_schema_valid` is set-based, not positional), but worth the cross-model consistency.

**[Minor] `growth_goal_set` vs `goal`** — No change. These live on distinct Amplify record types (the `pm_student_summary` feed vs the aimline report), and I can't confirm `growth_goal_set` and `goal` are the same semantic field without the vendor spec. I won't add an inline comment asserting a relationship I haven't verified; if the data team confirms they're equivalent, happy to annotate then.

**[Minor] `extensionAttribute1` in `Group`** — No change. No drift has been observed on group feeds, so adding it now is speculative (YAGNI). It will surface as its own `avro_schema_valid` warning if it ever appears, and we'll add it at that point.

Issue #4237, opened 2026-06-22
feat(cube): small-cell suppression for aggregate demographic breakdowns


cristinabaldor:
## Problem

Aggregate demographic breakdowns (e.g. `staff_summary` grouped by `race` x `gender_identity`, and future survey / performance / compensation views) can expose an individual when a cell count is very low. If only one person matches an intersection (e.g. one Asian transgender employee), the "aggregate" is effectively personal data.

Current informal benchmark is to suppress groups smaller than 4. Goal: adopt a defensible, consistent minimum-cell-size standard and enforce it in code.

## Why this is separate from the column-level PII gate (#4236)

The `staff_detail` PII gate (#4236) is static, column-level access control — Cube's `access_policy` handles it natively. Small-cell suppression is **value-dependent, post-aggregation** and `access_policy` cannot express it. Two hard constraints:

1. **No native Cube mechanism.** The closest option is injecting a `count >= k` measure filter via `queryRewrite`, which would fire on every query globally (breaking legitimate exact headcounts) unless carefully scoped to demographic queries.
2. **Complementary suppression.** Suppressing only the small cell while still showing the column total and sibling cells leaks the suppressed value by subtraction. Correct enforcement requires secondary suppression across the whole table.

## Governance decision needed (People Ops / legal)

Pick a minimum-cell-size threshold. Reference points (to be confirmed, not authoritative):

- ESSA accountability minimum n-size: states chose 10–30.
- NCES / privacy-reporting suppression: commonly suppress n < 10, plus secondary suppression.
- General k-anonymity floor: k >= 5; higher for sensitive attributes.

Current n=4 is below the most common education thresholds; n < 10 with secondary suppression would align with NCES-style practice.

## Scope

Network-wide: applies to any future Cube view exposing demographic, survey, performance-management, or compensation breakdowns — not just `staff_summary`.

## Next step

Brainstorm a design once the threshold is set. Likely involves a `queryRewrite` strategy plus a documented suppression policy. Needs its own spec.

PR #4282, opened 2026-06-29
feat(dbt,cube): single weekly-grain staff attrition fact + foundation window fix

### LINKED ISSUE #4275: feat(cube): staff attrition views with over-the-year trend
Add Cube views for staff attrition: aggregate rates by methodology, termination reason/type, and staff characteristics as of exit, with a weekly/monthly trend axis.

Spec: `docs/superpowers/specs/2026-06-26-staff-attrition-cube-views-design.md` on branch `cristinabaldor/feat/claude-staff-attrition-cube-views`.

## Design summary

**`fct_staff_attrition` (refactor in place, cohort grain unchanged)** — one row per `(employee_number, academic_year, attrition_type)`. No daily expansion. Gains: `staff_key` (join key + count_distinct measure key), `window_start_date` / `window_close_date` (trend spine), `as_of_exit_date` (last active assignment start, capped at term date — the `staff_work_history` SCD2 pin), `termination_type` (derived leading token). Artifact reasons (`Import Created Action`, `Upgrade Created Action`, `Internship Ended`) excluded from the termination pick. DRY 3-window CTE refactor (9 near-identical CTEs collapsed to one `windows` unnest + cohort/returner/terminations/as_of/attrition) — review with user before implementing.

**Cube `staff_attrition`** — two joins: point-in-time `staff_work_history` (pinned on `as_of_exit_date`; traverses all work/location/manager/demographic context, exactly as `staff_detail` does) and one-to-many `attrition_periods` (the trend spine). Per-methodology measures only — `count_cohort_*`, `count_attritors_*`, `attrition_rate_*`, `retention_rate_*` for each of `foundation` / `nj_compliance` / `recruitment`. No generic blendable `attrition_rate`.

**Cube `attrition_periods extends dates`** — cube-layer alias; no new warehouse model. Resolves the diamond (`staff_work_history` already joins `dates`). Inherits all `dim_dates` dimensions: `calendar_week_end_date` (ISO weeks), `school_week_end_date` (Monday-based year-round grid, covers summer), monthly granularity.

**View `staff_attrition_summary`** — aggregate rates + trend. `attrition_methodology` not a free dimension (encoded in measure names, so cross-methodology blending is impossible). Single `cube-access-staff-data` policy.

**Detail views deferred** — per-methodology `staff_attrition_{type}_detail` views (row-locked via static `row_level` access-policy filter) are out of scope for Phase 1.

**No `cube.js` changes** — `count_distinct(staff_key)` is correct per period bucket without snapshot-guard machinery; school-week grouping uses the plain `dim_dates` grid, not the PowerSchool per-school in-session grid.

## Scope (Phase 1)

- dbt: DRY CTE refactor + `termination_type` + artifact exclusion + `staff_key` / `window_*` / `as_of_exit_date` columns + tests
- Cube: `staff_attrition` cube; `attrition_periods extends dates`; `staff_attrition_summary` view; `cube.yml` exposure update

## Follow-on phases

- **Phase 2**: migrate `int_topline__staff_retention` off Engine B onto this fact
- **Phase 3** (#3744): union pre-2021 Dayforce history into work-assignment dims

cristinabaldor:
## Summary & Motivation

Closes #4275.

When merged, this exposes staff attrition in the Cube semantic layer with
per-methodology rates (foundation / nj_compliance / recruitment), work-context
and demographic breakdowns as-of the time staff left, and a cumulative
weekly/monthly trend — via the new `staff_attrition_summary` view.

**Design pivot during prototyping.** The original plan kept `fct_staff_attrition`
at cohort grain and expanded the weekly trend at query time through an
`attrition_periods` (`extends dates`) spine cube. Cube validation showed this
fought the join planner — final-outcome queries fanned across the whole date
spine, and the ratio measure tripped Cube's fan-trap guard. Per `src/cube`
conventions a period-intersection grain belongs in dbt, so the design
consolidated to a single materialized table:

- **`fct_staff_attrition_weekly`** — one row per
  `(employee x academic_year x methodology x Monday-Sunday week)` (~1.3M rows).
  `is_attritor_as_of` is a monotonic flag: `count_distinct(staff_key)` over it
  grouped by `week_end_date` gives the cumulative trend, and collapses to the
  exact final-outcome count when ungrouped — so the rate measures are fan-safe
  on one table and no snapshot anchor is needed.
- **Foundation termination-window bug fixed.** The capture window closed at the
  cohort window-close (4/30) but attrition is measured to the 9/1 return check,
  so summer (esp. June) departures were flagged as attrition with their
  termination date dropped. Extending capture to the return-check date recovers
  ~1014 foundation termination dates and drops genuinely-null terminations from
  59% to under 1%.
- **`staff_work_history` cube** gains two additive backing members (`staff_key`,
  `effective_end_date`) so the point-in-time exit-context join resolves.

Reconciliation verified against the dev schema (AY2023: foundation 1573 cohort /
316 attritors, identical to the prior `is_attrition` final outcome) and all
seven prototype queries (V1-V7) pass in the Cube dev server.

**Breaking change:** `fct_staff_attrition` is renamed to
`fct_staff_attrition_weekly` and changes grain. It had no downstream dbt-model
consumers (only the `cube.yml` exposure pre-wire, updated here). Rollback:
revert this PR; nothing in prod reads the old fact yet.

## AI Assistance

AI-assisted (Claude Code): the DRY refactor, the weekly-expansion SQL, the cube
and view YAML, and the spec/plan amendments. Human-directed: the design pivot to
a single weekly fact, the call to fix the foundation window bug, the
Monday-Sunday week and residual-null decisions, and all prototype validation.

### dbt

- [x] Properties file included for the new model
- [x] Exposure updated (`cube.yml` now references `fct_staff_attrition_weekly`)
- [x] Breaking change noted above with rollback plan

## CI checks

- [x] Trunk — passes locally (`trunk check --force` clean on all touched files)
- [x] dbt Cloud — pending
- [ ] Dagster Cloud — not triggered (no Dagster code paths changed)


cristinabaldor:
Addressed the Claude review findings in 5d6e343e2:

**Important**
1. **Join operand type mismatch** — cast the left operand to DATE so all three `BETWEEN` operands are DATE (`CAST({CUBE}.as_of_exit_date AS DATE) BETWEEN ... AND ...`). Resolves the TIMESTAMP-vs-DATE asymmetry; matches the date-pinned semantics of `as_of_exit_date`.
2. **Redundant uniqueness test** — dropped the model-level `dbt_utils.unique_combination_of_columns`. The PK `unique` test on `staff_attrition_weekly_key` (hash of employee_number, academic_year, attrition_type, week_end_date) covers the same violations, per the marts conventions doc. Re-ran `dbt build` in dev: all 7 tests pass, including PK uniqueness.

**Minor**
- Made `staff_work_history.effective_end_date` non-public — it is only needed as a join-backing member, so this avoids the net-new public exposure.
- Trimmed the "Read by the staff_attrition Cube cube" consumer reference from the model description (consumer tracking lives in `cube.yml` `depends_on`).

Thanks for the review — the type-cast catch in particular was a real fragility.

cbini (changes requested):
## Requesting changes

Strong change overall — the weekly-grain design is the right call, the foundation termination-window bug fix is correct, and I independently reconciled the core numbers against production. There are **two blockers to fix before merge**, both about how the new view *exposes and gates* data (not the math), plus a few should-fix items. Findings in plain terms, grouped by severity.

### What's working well (verified, not eyeballed)

- **Weekly grain is the right design.** One row per person-per-week, counted with `count_distinct` over a monotonic flag, means a single table serves week / month / quarter / final-outcome — no second fact needed, and it rolls up cleanly to any coarser time grain.
- **Foundation bug fix is correct and surgical.** Before this PR, staff who left over the summer were correctly counted as attrition but their departure *date* was discarded (the capture window closed 4/30, before the 9/1 return check). I reproduced the baseline in prod: AY2023 foundation = 1,573 cohort / 316 attritors, with 59.2% of attritors missing a termination date — matching the PR's claim, and foundation-specific. Extending capture to the return-check date fixes it.
- **Leave-of-absence is handled correctly.** I checked the warehouse: `status_code` is only ever A (active), T (terminated), or L (leave — reasons are Medical / Family / Disability / Administrative). The `status_code != 'T'` rule counts someone on leave on the return-check date as *retained*, which is what we want — and T is the only separation code, so nothing slips through.
- **Cross-year departure dates land in the right year.** A summer-2025 departure is booked to the 2024 academic year (measured to the 9/1/2025 check) with its real date kept.
- The DRY refactor is provably equivalent to the old per-methodology CTEs, and the superseded date-spine design is fully removed.

### Must fix before merge

**1. The view lets anyone slice attrition by race, gender, and ethnicity — and small groups re-identify real people.**
`staff_attrition_summary.yml:49-54` exposes `gender_identity`, `race`, and `is_hispanic` as open breakdowns. We removed a prior view (`staff_summary`) for exactly this, and `src/cube/CLAUDE.md` says demographic breakdowns do not return until small-number suppression ships (#4237). Concretely: grouping by location, race, and termination reason for one year easily yields a group of one — the single Hispanic teacher at a given school who left for a specific reason — which identifies a person and why they left. The view's own comment at line 126 already flags #4237 as the missing guard.
Fix: remove the three demographic fields from this view until #4237 lands.

**2. The permission group the view is locked to does not exist — so nobody can see any data.**
`staff_attrition_summary.yml:127` gates the view on group `cube-access-staff-data`, which is never created anywhere (the real groups are `staff-directory`, `staff-pii-*`, and the forward-compat comp/observations/benefits tiers). Since no one is granted that group, the view returns zero rows for every user. The name came from issue #4275 but was never implemented in `access.js`.
Fix: decide the access tier for aggregate attrition and wire it into `access.js`. This is entangled with #1 — the shortcut of pointing it at `staff-directory` would immediately expose the #1 demographics to every staff viewer, so resolve #1 first.

### Should fix

**3. A comment gives a false reason for skipping a data-model link.** `fct_staff_attrition_weekly.yml:37-40` says no foreign key to `dim_staff` is possible because it is a view. `dim_staff` is a table (`materialized: table`, with a primary key), and sibling facts (`fct_staff_observations`, `fct_staff_observation_goals`) do declare this FK — it feeds the generated model-reference diagram.
Fix: add the `foreign_key` constraint on `staff_key` and correct the comment.

**4. Two load-bearing `DISTINCT`s are not annotated.** `fct_staff_attrition_weekly.sql:74-98` — `year_cohort` and `returner_cohort` use `select distinct` to collapse the assignment-history fan-out to one row per person. Convention (see `dim_assessments.sql`) is to annotate grain-projection DISTINCTs so reviewers can tell an intentional dedup from a band-aid. Correct as-is; add the note.

**5. A join can shift by a day in our timezone.** `staff_attrition.yml:20-25` joins to work-history on `type: time` dimensions. Cube injects a timezone conversion into join predicates (`src/cube/CLAUDE.md`, #4298); under Eastern time the join window shifts about a day. Because it is a `BETWEEN` range join it will not blank out, but a departure on a period boundary can attach the wrong job / location / demographics.
Fix: join on raw DATE columns instead of the time dimensions; confirm on a branch staging deploy under Eastern time.

**6. The rates are only correct one academic year at a time — and nothing enforces it.** This is the biggest analytical risk. Each person keeps the same `staff_key` across years and the measures use `count_distinct(staff_key)`. Within one year that is right. But view the rate *without* pinning `academic_year` — e.g. attrition by region over several years — and a person is counted once instead of once per year, quietly turning an annual rate into a lifetime rate. Today the only protection is a docstring note.
Recommended fix — make it structural, two options:
- **(a) Correct-by-construction (preferred, pure data-model change):** count a per-year key instead of the bare staff key, e.g. `count_distinct(format('%T|%T', staff_key, academic_year))`. Multi-year totals then *sum* the yearly cohorts (person-years) and the rate becomes a proper pooled annual rate. Within-year behavior and the cumulative-collapse property are unchanged.
- **(b) Reject cross-year queries in `queryRewrite`:** throw a clear error if a `staff_attrition` query lacks a single-value `academic_year` filter. Note: RLS was moved out of `queryRewrite` (commit `04c7536d6`) into declarative `access_policy`, but the hook is still live on prod and carries the snapshot-anchor guard (`cube.js:274`; it already throws for mis-grouped period-end measures at `cube.js:316-335`) — so this is a valid home for a correctness guard, though it grows a function the team has been deliberately shrinking. Your call which fits better.

### Minor

- `staff_attrition_summary.yml:71-90` lists measures under `meta.folders`; folders should group dimensions only.
- `staff_attrition.yml:36` — `attrition_methodology` is `public: true` but no view uses it; set `public: false` so a future view cannot accidentally re-enable cross-methodology blending.
- Work-history join fan-safety assumes at most one primary-position period overlaps the exit date per person; counts are protected by `count_distinct`, but work-context breakdowns could double-attribute if that ever breaks — confirm an upstream non-overlap test exists.
- Current-year foundation / recruitment cohorts show zero rows until 9/1 (empty week array) while nj_compliance (opens 7/1) shows earlier — likely intended, worth a sanity check.
- The artifact-reason exclusion (`fct_staff_attrition_weekly.sql:128-130`) is a new behavior change bundled with the window fix — worth a line in the PR description.
- Confirm the "1,014 recovered / null under 1%" figure in CI once the model builds (the 59.2% baseline is confirmed in prod).

Happy to pair on the #1/#2 access design or draft the option-(a) count-key change for #6.

PR #4285, opened 2026-06-30
feat(dbt): idempotent Finalsite-to-Focus imports (diff, codes, 8400 prefix)

### LINKED ISSUE #4208: feat(dbt): Focus crosswalk gaps — enrollment-code E05/E02 + unfilled cells
Follow-up to #4201 (Component 4 of #4073). Combines the outstanding Focus crosswalk gaps.

## 1. Enrollment-code cannot express E05 vs E02

`stg_google_sheets__focus__enrollment_code_crosswalk` keys on `lifecycle_action` (`create` / `re_enroll` / `transfer_out`), so every new entrant maps to a single `create` code. Focus needs E05 for kindergarten entrants and E02 for above-KG new entrants — the crosswalk can't distinguish them.

- Confirm the E05/E02 rule with ops/registrar.
- Branch the enrollment-code derivation on grade (KG vs above-KG) in `rpt_focus__student_enrollment`, or extend the crosswalk grain to include grade.
- Resolve before the Component 5 SFTP transport ships.

## 2. Unfilled crosswalk cells

`school_id` and `drop_code` resolve null where ops has not yet populated the Google-Sheets crosswalk cells (school-id and drop-code crosswalks); `start_date` is null for not-yet-enrolled students. These are ops data-entry gaps that surface as warn-level test results.

- Track ops population of the crosswalk cells.

Refs #4073

### LINKED ISSUE #4279: feat(dbt): idempotent Finalsite-to-Focus imports plus 4208 enrollment/withdraw codes
Make the Finalsite to Focus SFTP imports idempotent so we stop re-sending data Focus already has, resolve the enrollment/withdraw-code gaps from #4208, and apply the required Florida district id prefix.

## Background

The five `rpt_focus__*` extracts (Demographics, Student_Enrollment, Addresses, Contacts, Linked_Students) currently emit every in-scope Finalsite row on every run with no comparison to current Focus state. Desired-state logic lives in `kipptaf` (`kipptaf_extracts.rpt_focus__*`); the `kippmiami` models are thin pass-throughs that the Dagster extract reads to produce the CSVs. The `focus` dbt package (current Focus SIS state) is available only in `kippmiami`.

Join key (export STDT_ID to Focus student), once the 8400 prefix below is applied:

```text
export.student_id ('8400' || focus_student_id) = focus.students.student_id
```

1,376 of 3,088 minted ids already exist in Focus; the rest are genuinely new.

## Requirements

1. **Addresses, Contacts, Linked Students — import once.** Exclude a student once a row exists in the Focus destination table for that extract.
2. **Demographics — diff.** Import only unmatched students and students whose populated values differ from what is in Focus.
3. **Student Enrollment — diff with code exception.** Import only unmatched and changed records, EXCEPT entry and withdrawal codes, which import once and are never overwritten.
4. **Resolve #4208 codes.** `ENROLLMENT_CODE` = `E05` for KG, `E01` for everyone else. `DROP_CODE` sourced from the Finalsite field `fl_state_withdraw_codes_ss`.
5. **8400 id prefix.** Prefix the Finalsite `focus_student_id` with the constant `8400` when emitting STDT_ID / STUDENT_ID to Focus. Required by Florida (district id prefix); Finalsite autoincrement cannot hold the full 10-digit value. Applies to all five extracts. Side benefit: the emitted id then equals Focus `student_id` directly, simplifying the diff join.

## Approach (agreed)

- Keep `kipptaf` as the desired-state build; do reconciliation in the `kippmiami` layer (it has both the `finalsite` and `focus` packages).
- #4 enrollment code (grade-based) and #5 prefix computed in `kipptaf`; drop-code decode plus all diff / import-once filters in `kippmiami`.
- `DROP_CODE` decode: `fl_state_withdraw_codes_ss` is a label (e.g. `(W02) In District Transfer`) that matches `stg_focus__student_enrollment_codes.title` where `type = Drop` at 100 percent, yielding the `short_name`.
- Demographics / Enrollment diff compares only populated export columns against Focus. Focus stores demographics as encoded custom fields, so gender / language / race / ethnicity need value-normalization before comparison.

## Requirement 1 prerequisite (dlt + staging)

The Focus destination tables for Addresses / Contacts / Linked Students are listed in `dlt/focus/config/focus.yaml` but were never materialized in BigQuery: `students_join_address`, `students_join_people`, `people`, and `students_join_students` are absent, and `address` and `people_join_contacts` hold only 2 dummy rows each. They also have no `stg_focus__` staging models. Plan: confirm/repair the dlt pull so these tables materialize, add the staging models, then gate #1 on row-existence.

## Scope

Single combined spec/PR covering all four requirements plus the 8400 prefix and the dlt + staging prerequisite for #1.

Closes #4208

cbini:
## Summary and Motivation

When merged, this makes the Finalsite-to-Focus SFTP extracts idempotent (stop re-sending data Focus already has), resolves the #4208 enrollment/withdraw-code gaps, and applies the Florida-required 8400 student-id prefix. Single PR covering requirements 2-5 of #4279.

**Architecture:** `kipptaf` `rpt_focus__*` build the desired state (no Focus dependency); the `kippmiami` `rpt_focus__*` pass-throughs become the reconciliation layer (they alone have the `focus` package) — joining current Focus state, decoding the withdraw code, and applying the diff / import-once filters.

### Requirement 5 — 8400 prefix (kipptaf)

Every emitted student id is `concat('8400', focus_student_id)` across all five extracts. Florida requires the district prefix; Finalsite autoincrement cannot hold the full 10-digit value. The prefixed id equals Focus `students.student_id` directly, which the diff joins on.

### Requirement 4 — enrollment + withdraw codes (resolves #4208)

- `enrollment_code` = `E05` for kindergarten, `E01` otherwise, null on transfer-out (replaces the lifecycle-action crosswalk).
- Withdraw code sourced from the Finalsite `fl_state_withdraw_codes_ss` field, decoded in kippmiami against `stg_focus__student_enrollment_codes` (type Drop) to the Focus `short_name`. Both the entry and withdraw codes are gated to `transfer_out` where appropriate so a still-enrolled student never emits a drop code.

### Requirement 2 — demographics diff (kippmiami)

Emit only unmatched students and students whose populated values differ from Focus, with per-field normalization (gender bracket-code, Yes/No to Y/N, race flags, birthdate format, language).

### Requirement 3 — enrollment diff with import-once codes (kippmiami)

Match on `(student_id, start_date)`; emit only unmatched or `end_date`-changed rows. Entry and withdraw codes import once and are never overwritten (emitted only when the matched Focus enrollment has none). `school_id`/`grade_id` are excluded from the diff because the export emits Focus import codes while Focus stores internal numeric ids.

### Linked Students — disabled for v1

The `linked_students` asset is removed from the Focus extract config so the CSV stops generating; the dbt models stay dormant.

### Requirement 1 — deferred (tracked in #4279)

Addresses / Contacts / Linked Students import-once is blocked: the Focus linkage tables (`students_join_address`, `students_join_people`) are empty in the Focus source, so dlt has not created them in BigQuery. They materialize automatically once those Focus modules hold rows; #4279 stays open to track it.

## CI note

dbt Cloud CI is kipptaf-scoped, so it builds the self-contained kipptaf `rpt_focus__*` models (green). The kippmiami models are validated by dbt unit tests (in this PR) and ordered after the kipptaf extract at prod runtime via the declared `source()` asset key.

## AI Assistance

Spec, plan, and implementation were drafted by Claude Code under human direction (#4279) and reviewed task-by-task plus a whole-branch review. Human-directed: requirements, the E05/E01 rule, the 8400 constant, the linked-students drop, and all scope decisions.

### dbt

- [x] Properties files updated for every changed model; unit tests cover the prefix, E05/E01, withdraw decode, demographics diff, enrollment diff, and import-once codes.
- [x] **Breaking change:** renames the contracted column `drop_code` to `state_withdraw_label` in the kipptaf `rpt_focus__student_enrollment` carrier; the only consumer is the kippmiami pass-through, updated in this same PR. Rollback: revert this PR.
- No external source added or modified.

## Follow-ups

- Add a `(student_id, start_date)` uniqueness test on `stg_focus__student_enrollment` to guard the diff join grain.
- Retire the two now-orphaned crosswalk models (`stg_google_sheets__focus__enrollment_code_crosswalk`, `stg_google_sheets__focus__drop_code_crosswalk`).

Closes #4208
Refs #4279


src/dbt/kippmiami/models/extracts/focus/rpt_focus__demographics.sql
```diff
@@ -1,3 +1,88 @@
+with
+    desired as (
+        select *, from {{ source("kipptaf_extracts", "rpt_focus__demographics") }}
+    ),
```
cbini: why is this cte necessary?

src/dbt/kippmiami/models/extracts/focus/rpt_focus__demographics.sql
```diff
@@ -1,3 +1,88 @@
...
+        from {{ ref("stg_focus__students") }} as s
+        left join
+            {{ ref("int_focus__students__pivot") }} as p on s.student_id = p.student_id
+    ),
+
+    diffed as (
+        select d.*,
+        from desired as d
+        left join focus_state as f on cast(d.stdt_id as int64) = f.student_id
+        where
```
cbini: can we compare the diff using a surrogate key instead of this complex where statement?

src/dbt/kipptaf/models/extracts/focus/rpt_focus__addresses.sql
```diff
@@ -1,6 +1,6 @@
 -- trunk-ignore(sqlfluff/ST06): column order fixed by Focus ADDRESS contract
 select
-    ida.focus_student_id as student_id,
+    concat('8400', ida.focus_student_id) as student_id,
```
cbini: refactor to int_finalsite__contact_id_attributes

src/dbt/kipptaf/models/extracts/focus/rpt_focus__student_enrollment.sql
```diff
@@ -4,23 +4,38 @@ select
...
+        -- non-digit grade names (e.g. pk) yield null here; Miami is K-9 today
         lpad(regexp_extract(l.grade_canonical_name, r'\d+'), 2, '0')
     ) as grade_id,
 
     format_date('%Y%m%d', l.enrollment_start_date) as start_date,
 
-    ec.focus_enrollment_code as enrollment_code,
+    case
+        when l.lifecycle_action = 'transfer_out'
+        then cast(null as string)
```
cbini: incorrect. enrollment code should not change on transfer out. only drop code should be populated

src/dbt/kipptaf/models/extracts/focus/rpt_focus__student_enrollment.sql
```diff
@@ -4,23 +4,38 @@ select
...
-    dc.focus_drop_code as drop_code,
+    -- fl_state_withdraw_codes_ss is a raw contact custom attribute (NOT gated
+    -- upstream), so gate it to transfer_out here — a withdraw code is only
+    -- meaningful for a withdrawal, and an ungated value would emit a drop code
+    -- for a still-enrolled student downstream.
+    if(
+        l.lifecycle_action = 'transfer_out',
+        cca.fl_state_withdraw_codes_ss,
+        cast(null as string)
+    ) as state_withdraw_label,
```
cbini: this is not the correct import column header. it should be drop_code. recheck headers for all imports

src/dbt/kippmiami/models/extracts/focus/rpt_focus__student_enrollment.sql
```diff
@@ -1,12 +1,48 @@
...
+        select
+            e.*,
+            fe.enrollment_code as focus_enrollment_code,
+            fe.drop_code as focus_drop_code,
+            fe.end_date as focus_end_date,
+        from decoded as e
+        left join
+            {{ ref("stg_focus__student_enrollment") }} as fe
+            on cast(e.student_id as int64) = fe.student_id
+            and e.start_date = format_date('%Y%m%d', fe.start_date)
```
cbini: joins should never contain one-sided calculations. refactor.

src/dbt/kippmiami/models/extracts/focus/rpt_focus__demographics.sql
cbini: Removed in 610586ac7 — it was a `select *` pass-through. Inlined the `source()` into the diff CTE (now a surrogate-key comparison).

src/dbt/kippmiami/models/extracts/focus/rpt_focus__demographics.sql
cbini: Done in 610586ac7 — replaced the OR chain with a `generate_surrogate_key` comparison. To preserve the null-skip semantics (an unpopulated export field must not re-import), the desired side hashes `coalesce(export, focus)` against the raw focus side, so a null export field adopts the Focus value and registers no difference. The existing unit test still passes.

src/dbt/kipptaf/models/extracts/focus/rpt_focus__addresses.sql
cbini: Done in 610586ac7 — added `focus_student_id_prefixed` (the `8400` prefix) as a column on `int_finalsite__contact_id_attributes`; all five extracts now select it instead of repeating `concat('8400', ...)`. Placed it on the kipptaf union wrapper rather than the finalsite package model: `union_relations` is compile-time, so a new package-level column would not surface in kipptaf CI until kippmiami prod rebuilt (per src/dbt/CLAUDE.md).

src/dbt/kipptaf/models/extracts/focus/rpt_focus__student_enrollment.sql
cbini: Fixed in 610586ac7 — enrollment_code is now always `E05`/`E01`; removed the `transfer_out`→null branch. The withdrawal is expressed by `drop_code` + `end_date` only.

src/dbt/kipptaf/models/extracts/focus/rpt_focus__student_enrollment.sql
cbini: Fixed in 610586ac7 — renamed the carrier column back to `drop_code` and moved the decode in place in kippmiami. Rechecked every extract's aliases against main: this was the only renamed header in the PR; the rest match the Focus contract. Confirmed the import header is `DROP_CODE` ← dbt `drop_code` in kippmiami extracts/config/focus.yaml.

src/dbt/kippmiami/models/extracts/focus/rpt_focus__student_enrollment.sql
cbini: Done in 610586ac7 — pre-formatted the Focus join keys (`cast(student_id as string)`, `format_date(start_date)`) in a `focus_enrollment` CTE so the join compares bare columns. Applied the same fix to the demographics join (cast `student_id` in `focus_state`).

src/dbt/kipptaf/models/extracts/focus/rpt_focus__addresses.sql
cbini: move to package

src/dbt/kippmiami/models/extracts/focus/rpt_focus__demographics.sql
cbini: calculate these upstream in a cte

src/dbt/kipptaf/models/extracts/focus/rpt_focus__student_enrollment.sql
cbini, line 1: i'm not so sure about the grain on this. int_finalsite__enrollment_lifecycle is a row per student/year right? student_enrollment should be at the enrollment stint grain. is that a gap in our data model?

src/dbt/kippmiami/models/extracts/focus/rpt_focus__demographics.sql
cbini: Done in 83421da62 — moved both surrogate keys into a `keyed` CTE (`export_key` = each export field coalesced to its Focus value, `focus_key` = current Focus values). The diff filter now reads `where focus_student_id is null or export_key != focus_key`. Also fixed CI: the prior commit only updated the student_enrollment unit-test mock, so the addresses/contacts/demographics/linked_students mocks still referenced the old `focus_student_id` column — all four (and the linked_students join filters) now use `focus_student_id_prefixed`; all 10 focus unit tests pass.

src/dbt/kipptaf/models/extracts/focus/rpt_focus__student_enrollment.sql
cbini: Sized it against prod. `int_finalsite__enrollment_lifecycle` is one row per `finalsite_enrollment_id`, and a Finalsite contact carries a single `school_year_start` + one `enrolled_date` + one `enrollment_end_date` (= `min()` of the three withdrawal-date fields) — so one stint per contact.

Data check:
- Across 1,983 student-years, **0** students have more than one Finalsite contact in the same `school_year_start` (max = 1). So multi-year re-enrollment arrives as separate contacts with distinct `start_date`s, which the `(student_id, start_date)` grain represents correctly as separate stints.
- Only **8 of 4,328 contacts (0.18%)** have more than one withdrawal-date field populated (where `min()` takes the earliest). That is still one stint per contact, not a dropped stint.

So no structural grain gap for the Focus extract today — intra-year withdraw-then-re-enroll under a single contact does not occur in the data. The 8 multi-withdrawal contacts are worth a registrar sanity-check on which date should win, but they don't change the grain. Happy to open a follow-up issue to track that edge if you want.

cbini: Correction — you're right, and my earlier "separate contacts" claim was wrong. Re-enrollment reuses the same `finalsite_enrollment_id`; it does not mint a new one.

Verified against prod: every student maps to exactly one contact and one `school_year_start` (0 of 1,983 students span more than one year or more than one contact). The data holds two cohorts — 124 contacts at `school_year_start` 2025 and 1,864 at 2026 — with no student appearing in both. That is consistent with a stable contact id whose `school_year_start` advances in place, not a new id per year.

So `int_finalsite__enrollment_lifecycle` is one row per student = current Finalsite enrollment state (the contact's latest year), not one row per student-year and not a multi-year stint history.

For this outbound feed it still functions: when a continuing student's contact rolls to the next year, `enrollment_start_date` advances, the diff sees a new `(student_id, start_date)` absent from Focus, and emits it — so re-enrollment is pushed as a new Focus row, and the prior year already lives in Focus. The real limitation is that Finalsite is current-state only, so this extract cannot emit or backfill a prior-year stint. Not a blocker for the push, but Finalsite is not a multi-year enrollment history source — that would be a separate modeling effort (status_report history or a snapshot). I can open a follow-up issue to track that if you want it.

cbini: Added the forward-enrollment protection in dad0f3472. `int_finalsite__enrollment_lifecycle` now only counts a withdrawal date that is on or after the current `enrolled_date` when computing `enrollment_end_date`:

```sql
(select min(d) from unnest([...withdrawal dates...]) as d
 where d >= sr.enrolled_date) as enrollment_end_date
```

A prior-year withdrawal on a reused contact (date < the new start) no longer sets `enrollment_end_date`, so `is_transfer_out` stays false and the forward (re)enrollment emits no `drop_code` / `end_date` — it stays `create`/`re_enroll`. Genuine current-year withdrawals (date >= start) still transfer out unchanged. Added a unit test covering both (`c_stale` → re_enroll/null end, `c_current` → transfer_out). Latent today (the 2025 and 2026 cohorts don't overlap), but protected going forward. Fix is in the SIS-agnostic lifecycle model, so the PowerSchool receiver benefits too.

docs/reference/finalsite-focus-import.md
```diff
@@ -0,0 +1,130 @@
...
+| Kindergarten | `E05`           |
+| All others   | `E01`           |
+
+The entry code reflects how the student **entered** and does not change when a
+student withdraws — a withdrawal is expressed by the drop code and end date, not
+by clearing the entry code.
+
+### Withdraw / drop codes
+
+When a student transfers out, Finalsite records the withdrawal reason as the
```
cbini: name the finalsite field

cbini:
## Code review

Reviewed base `8461ee17` to head `5006967f` against requirements 2-5 of #4279. dbt conventions and the relevant CLAUDE.md files were loaded first.

**Verdict: merge with fixes.** Reqs 4 and 5 are correctly implemented and well-tested; teardown is clean. One likely-blocking idempotency bug in Req 2.

### Strengths

- **8400 prefix is DRY and correct** — applied once at `int_finalsite__contact_id_attributes` as `focus_student_id_prefixed`, consumed by all five extracts. `concat` returns NULL on NULL input, so existing `is not null` guards still hold. No non-prefixed id leaks into an emitted `student_id`.
- **E05/E01 rule correct**, and the entry code correctly persists (is not cleared) on transfer-out.
- **Withdraw-code gating is sound end-to-end** — `drop_code` is gated to `transfer_out`, which is derived from `is_transfer_out` and requires a non-null `enrollment_end_date`, so a still-enrolled student cannot emit a drop code.
- The new `enrollment_end_date` to `enrolled_date` ordering guard in `int_finalsite__enrollment_lifecycle` is a genuine correctness fix for reused Finalsite contacts.
- **No banned dedup idioms.** Diff uses null-safe `is distinct from`. The two deleted crosswalk models have zero dangling refs at head; `linked_students` is dropped only from the Dagster config and its dbt models stay valid, so the DAG does not break.

### Issues

#### Important

1. **Demographics language diff compares a code against a label — likely re-emits every student with a language each run** (`src/dbt/kippmiami/models/extracts/focus/rpt_focus__demographics.sql`, `focus_state`/`keyed` CTEs). The export side emits the FLDOE two-letter `focus_language_code`; the Focus side compares against `language_label` (the decoded human label, e.g. "English"). Different domains, so `coalesce(d.lang, f.language_label)` vs `f.language_label` diverges for any student with a populated language, defeating idempotency for the demographics feed. The unit test masks this by mocking `lang` as a label rather than the code the kipptaf model actually produces, and the kipptaf model description itself flags LANG as unconfirmed ("pending registrar confirmation"). **Fix:** compare like-for-like — decode `language_label` to the FLDOE code on the Focus side, or confirm the Focus LANG field stores the FLDOE code and align the pivot. Verify against production Focus data before merge.

2. **Enrollment import-once join can fan out on a non-unique grain** (`rpt_focus__student_enrollment.sql`, `focus_enrollment` join to `matched`). Joins on `(student_id, start_date)`, but `stg_focus__student_enrollment` is unique only on `id` — no uniqueness test on `(student_id, start_date)`. Two Focus rows sharing a `start_date` (same-day withdraw/re-enroll, grade correction) fan the export row into duplicates. The model's own `unique_combination_of_columns(student_id, start_date)` test would catch this loudly (not silent corruption), but the grain should be confirmed or defended — the PR already acknowledges the missing upstream uniqueness test. **Fix:** confirm `(student_id, start_date)` uniqueness in Focus, or dedupe `focus_enrollment` to one row per key before the join.

3. **Unit tests assert structure but not the production-realistic language value.** The enrollment tests are good (prefix, E05/E01, withdraw decode, import-once null guard, end_date change, unmatched). Once #1 is resolved, add a demographics test row exercising a populated-language match with the real code/label shapes so it would catch #1.

#### Minor

4. **PR description / implementation mismatch on the rename.** The body states it renames the contracted column `drop_code` to `state_withdraw_label`. The actual head keeps `drop_code` in both the kipptaf contract YML and the kippmiami consumer (`git grep` finds no `state_withdraw_label` anywhere). The implementation is internally consistent — this is a stale description, not a code bug. Recommend updating the body so the "breaking rename" framing does not mislead consumers; the column name is unchanged, only its semantics moved from a mapped code to a raw label decoded downstream.

5. **`student_id` carries `not_null` but is documented as nullable** ("null for enrollments where Finalsite has not minted the id"). Pre-existing tension (only `unique` was dropped this PR), but `concat('8400', null)` yields null, which fails `not_null`. Confirm the in-scope status gate guarantees a minted id, or the test will flag legitimately-pending students.

### Recommendations

- Before merge, run the `extracts/focus` unit-test directory in both kipptaf and kippmiami, since the prefix change touched sibling mocks.
- Add a test exercising the language field with genuine code-vs-label shapes to lock in #1's resolution.

The one real blocker is **#1** (language code-vs-label idempotency) — verify against production Focus values before merge. **#2** should be confirmed or defended.

<sub>Reviewed by Claude Code.</sub>

docs/reference/finalsite-focus-import.md
cbini: Done in 742284c46 — the doc now names the field: "Finalsite records the withdrawal reason in its `fl_state_withdraw_codes_ss` field as the full FLDOE label."

cbini:
Worked through the review against prod data (head now `742284c4`):

**#1 (language code-vs-label) — verified not a bug.** The premise that Focus stores a human label doesn't hold: `int_focus__students__pivot.language_label` is `EN` for all 3,453 rows (the FLDOE 2-letter code), and the export emits `EN`/`SP`/`HC`/`ZF`/`AR`. Same domain, so `coalesce(d.lang, f.language_label)` vs `f.language_label` compares like-for-like — a populated language does not re-emit every run. No idempotency defect for the demographics feed.

**#3 (test masking) — fixed in `742284c4`.** The demographics unit test mocked `lang`/`language_label` as labels (`English`/`Spanish`); it now uses the real FLDOE codes (`EN`/`SP`) on both sides, so it reflects production and would catch a code-vs-label regression. Test passes.

**#2 (enrollment grain fan-out) — confirmed unique, defended.** `stg_focus__student_enrollment` is 9,591 rows = 9,591 distinct `(student_id, start_date)` = 9,591 distinct `id`, so the join grain holds today. The kippmiami model's own `unique_combination_of_columns(student_id, start_date)` already guards the output loudly if Focus ever produces a same-day duplicate. Per the no-defensive-dedupe policy I'm not deduping `focus_enrollment`; the source-side uniqueness test on `stg_focus__student_enrollment` stays the tracked follow-up (#4279).

**#5 (`student_id` not_null vs nullable) — pre-existing, warn-level.** The `not_null` test has no explicit severity, so it runs at the project default (warn), not error — it surfaces the count of un-minted students as a soft signal rather than failing CI, consistent with the "null until Finalsite mints the id" documentation. Leaving as-is.

**#4 (stale PR description) — correct, and intentionally not changed yet.** The body's "renames `drop_code` to `state_withdraw_label`" framing is stale (the column stayed `drop_code`); I flagged this earlier and was asked to leave the PR body as-is. Happy to update it if you'd now prefer.

One note beyond this review's base: `drop_code` now updates-on-change while `enrollment_code` stays import-once (generated from grade), and Addresses/Contacts are now import-once against the Focus link tables.

cbini:
## Correction to finding #1 (language diff) — verified against production

Walking back the "likely blocker" framing on #1 after checking prod data. My original finding assumed `int_focus__students__pivot.language_label` holds the human label ("English"). **It does not** — in production it decodes to the FLDOE code:

- `language_label` in prod = `EN` across all 3,453 non-null rows (a code, not a label).
- Export side `lang` = FLDOE codes (`EN`/`SP`/`HC`/`ZF`/`AR`).

So for the English majority (330 of 403 export rows), the comparison is `EN` vs `EN` — it matches and is **not re-emitted**. The feed is idempotent for the dominant case, and the unit-test mock change to `EN`/`SP` correctly reflects what production decodes. **My original #1 was overstated — apologies for the false alarm.**

### Residual (minor, not a blocker)

The language custom-field option table contains a mix of labels — `EN`, `English`, and `Spanish`. Today every student decodes to `EN`, so this is latent, but if a registrar ever sets a student to the `English`/`Spanish` option, the export code (`EN`/`SP`) would diverge from the label and that student would re-emit every run. Clean guard: normalize the Focus side to the FLDOE code (same crosswalk the export uses) before comparison, rather than comparing against the raw label. Suggested follow-up, not a merge stop.

### Still open

Finding #2 (enrollment `(student_id, start_date)` import-once join fan-out) is unaddressed by the latest commit. Lower risk — the model's own `unique_combination_of_columns` test fails loudly rather than corrupting silently — but worth confirming the grain or deduping `focus_enrollment` before the join.

Net: #1 downgraded from blocker to a minor follow-up. With that, no remaining merge blocker from my side.

<sub>Reviewed by Claude Code.</sub>

cbini:
Agreed #1 isn't a blocker. Dispositions (head `c2b47940d`):

**Residual #1 (language label vs code) — fixed.** Confirmed the legacy Focus LANG field (`custom_200000005`) has exactly three options — `EN`, `English`, `Spanish` — and `int_focus__students__pivot.language_label` decodes to the option *label*, so a student set to `English`/`Spanish` would diverge from the export's FLDOE code. The export crosswalk lives in kipptaf and isn't reachable from the kippmiami diff, so I normalize the Focus side inline before comparison (`English`→`EN`, `Spanish`→`SP`, anything else passes through). Confirmed `English`→`EN` / `Spanish`→`SP` against the export crosswalk. The demographics unit test now stores the legacy labels on the Focus side so a matching language no longer re-emits; passes.

**#2 (enrollment `(student_id, start_date)` grain) — kept tracked.** Confirmed unique in prod (9,591/9,591) and the kippmiami model's own `unique_combination_of_columns(student_id, start_date)` already fails loudly on any future same-day duplicate, so no silent fan-out. Per the no-defensive-dedupe policy I'm not deduping `focus_enrollment`; the source-side uniqueness test on `stg_focus__student_enrollment` stays the tracked follow-up in #4279.

cbini:
@claude update your review

PR #4296, opened 2026-07-02
feat(powerschool): year-grain cumulative GPA and needed weighted GPA for 3.0 projection

### LINKED ISSUE #4295: feat(powerschool): year-grain cumulative GPA model and needed weighted GPA for 3.0 projection
Dashboard rebuild for GPA/gradebook data needs two INT-layer additions:

1. **Cumulative GPA by academic year** — new `int_powerschool__gpa_cumulative_year` (powerschool package + kipptaf union), grain `studentid, schoolid, academic_year`. Completed years are recomputed from stored Y1 grades as running sums; the current-year row is joined directly from `int_powerschool__gpa_cumulative` projected columns (flagged `is_projected`) so values match the existing model exactly, by construction.
2. **Needed weighted Y1 GPA for a 3.0 projected cumulative** — additive columns on `int_powerschool__gpa_cumulative`: `potential_gpa_credits_current_year`, `gpa_needed_for_cumulative_3_0`, `is_cumulative_3_0_attainable`, computed from the same rollup sums that produce the existing cumulative and projected GPAs.

Scope: INT layer only — no rpt/mart/extract surfacing in this effort. Validation includes prod-data reconciliation of year-model current-year rows against the original model values.

Design spec to follow on the linked branch.

anthonygwalters:
# Pull Request

## Summary and Motivation

Refs #4295

When merged, this pull request will add two INT-layer GPA capabilities for the gradebook dashboard rebuild:

1. **`int_powerschool__gpa_cumulative_year`** (new, powerschool package) — cumulative Y1 GPA as of the end of each academic year, recomputed from stored Y1 grades, with the in-progress current year as a projected point joined directly from the original model (exact match by construction, per governance requirement).
2. **Needed-GPA columns on `int_powerschool__gpa_cumulative`** — four additive columns: the weighted Y1 GPA needed this year to project a 3.0 cumulative, an attainability flag, and the two auditable credit denominators behind them.

Design spec: `docs/superpowers/specs/2026-07-02-gpa-cumulative-year-needed-gpa-design.md`. Implementation plan: `docs/superpowers/plans/2026-07-02-gpa-cumulative-year-needed-gpa.md`.

### Deployment sequencing — why this PR excludes the kipptaf layer

This is PR 1 of 2. The kipptaf union models are built with `dbt_utils.union_relations`, which resolves columns **at compile time** by querying BigQuery INFORMATION_SCHEMA on the district prod tables — not from dbt code. Until this PR merges and Dagster rebuilds the Newark/Camden/Miami prod tables, the new year-model tables do not exist in the warehouse and the four new columns are absent from prod. A same-PR kipptaf change would therefore fail to compile (the new union has no relations to introspect) or silently compile **without** the new columns (the existing `gpa_cumulative` union). PR 2 (kipptaf union model, sources entries, properties YAML) lands after district prod rebuilds, gated on verifying all three districts materialized. The single-PR fallback (manually seeding `zz_stg_*` schemas per district so CI can introspect the new shapes) touches shared staging datasets and is easier to get subtly wrong, so the repo-documented two-PR pattern is used.

## AI Assistance

Claude Code authored the spec, plan, and implementation (via subagent-driven execution); approach selection, governance constraints (join-not-recompute for the projected point), edge-case policy (raw needed value + attainability flag), and deployment sequencing were human-directed during an interactive design session.

## Self-review

_Complete only the sections relevant to your changes._

### General

      (Claude is advisory — use your judgement, but do not ignore it.)

### dbt

- [ ] Include a `[model_name].yml` properties file for all new models
- [ ] Include (or update) an exposure for all models consumed by a dashboard,
      analysis, or application _(N/A — INT layer only, no external consumers)_
- [ ] **Breaking change?** All model changes are additive; grain and existing
      columns untouched
- [ ] External sources unchanged — no `stage_external_sources` needed

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

### DIFF
--- a/src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_cumulative.sql
+++ b/src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_cumulative.sql
@@ -1,4 +1,10 @@
 with
+    gradescale_max as (
+        select gradescaleid, max(grade_points) as max_grade_points,
+        from {{ ref("int_powerschool__gradescaleitem_lookup") }}
+        group by gradescaleid
+    ),
+
     grades_union as (
         select
             sg.studentid,
@@ -42,6 +48,7 @@ with
             if(
                 sg.excludefromgpa = 0, su.grade_points, null
             ) as unweighted_grade_points_projected,
+            if(sg.excludefromgpa = 0, sg.gpa_points, null) as gpa_points_projected_max,
         from {{ ref("stg_powerschool__storedgrades") }} as sg
         left join
             {{ ref("int_powerschool__gradescaleitem_lookup") }} as su
@@ -81,6 +88,9 @@ with
             null as unweighted_grade_points,
 
             fg.y1_grade_points_unweighted as unweighted_grade_points_projected,
+            if(
+                fg.y1_letter_grade is null, null, gsm.max_grade_points
+            ) as gpa_points_projected_max,
         from {{ ref("base_powerschool__final_grades") }} as fg
         inner join
             {{ ref("base_powerschool__student_enrollments") }} as co
@@ -93,6 +103,7 @@ with
             and fg.course_number = sg.course_number
             and sg.academic_year = {{ var("current_academic_year") }}
             and sg.storecode = 'Y1'
+        left join gradescale_max as gsm on fg.courses_gradescaleid = gsm.gradescaleid
         where
             fg.exclude_from_gpa = 0
             /* ensures already stored grades are excluded */
@@ -132,6 +143,7 @@ with
             null as gpa_points_core,
             null as unweighted_grade_points,
             null as unweighted_grade_points_projected,
+            null as gpa_points_projected_max,
         from {{ ref("base_powerschool__final_grades") }} as fg
         inner join
             {{ ref("base_powerschool__student_enrollments") }} as co
@@ -180,6 +192,9 @@ with
             (
                 potentialcrhrs_projected * unweighted_grade_points_projected
             ) as weighted_points_projected_unweighted,
+            (
+                potentialcrhrs_projected * gpa_points_projected_max
+            ) as weighted_points_projected_max,
         from grades_union
     ),
 
@@ -214,8 +229,52 @@ with
                     potentialcrhrs
                 )
             ) as potential_credits_cum,
+            sum(
+                if(
+                    academic_year < {{ var("current_academic_year") }},
+                    weighted_points,
+                    null
+                )
+            ) as weighted_points_prior,
+            sum(
+                if(
+                    academic_year < {{ var("current_academic_year") }},
+                    potentialcrhrs,
+                    null
+                )
+            ) as potentialcrhrs_prior,
+            sum(
+                if(
+                    academic_year = {{ var("current_academic_year") }},
+                    potentialcrhrs_projected,
+                    null
+                )
+            ) as potentialcrhrs_current,
+            sum(
+                if(
+                    academic_year = {{ var("current_academic_year") }},
+                    weighted_points_projected_max,
+                    null
+                )
+            ) as weighted_points_projected_max_current,
         from with_weighted_points
         group by studentid, schoolid
+    ),
+
+    needed_gpa as (
+        select
+            *,
+
+            safe_divide(
+                (3.0 * (coalesce(potentialcrhrs_prior, 0) + potentialcrhrs_current))
+                - coalesce(weighted_points_prior, 0),
+                potentialcrhrs_current
+            ) as gpa_needed_raw,
+
+            safe_divide(
+                weighted_points_projected_max_current, potentialcrhrs_current
+            ) as gpa_max_current_raw,
+        from points_rollup
     )
 
 select
@@ -225,6 +284,8 @@ select
     potential_credits_cum,
     earned_credits_cum_projected,
     earned_credits_cum_projected_s1,
+    potentialcrhrs_projected as potential_gpa_credits_cum_projected,
+    potentialcrhrs_current as potential_gpa_credits_current_year,
 
     round(safe_divide(weighted_points, potentialcrhrs), 2) as cumulative_y1_gpa,
     round(
@@ -248,4 +309,9 @@ select
     round(
         safe_divide(weighted_points_core, potentialcrhrs_core), 2
     ) as core_cumulative_y1_gpa,
-from points_rollup
+
+    round(gpa_needed_raw, 2) as gpa_needed_for_cumulative_3_0,
+
+    round(gpa_needed_raw, 2)
+    <= round(gpa_max_current_raw, 2) as is_cumulative_3_0_attainable,
+from needed_gpa
--- a/src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_cumulative.sql
+++ b/src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_cumulative.sql
@@ -257,6 +257,14 @@ with
                     null
                 )
             ) as weighted_points_projected_max_current,
+            sum(
+                if(
+                    academic_year = {{ var("current_academic_year") }}
+                    and weighted_points_projected_max is not null,
+                    potentialcrhrs_projected,
+                    null
+                )
+            ) as potentialcrhrs_current_max_known,
         from with_weighted_points
         group by studentid, schoolid
     ),
@@ -271,8 +279,17 @@ with
                 potentialcrhrs_current
             ) as gpa_needed_raw,
 
-            safe_divide(
-                weighted_points_projected_max_current, potentialcrhrs_current
+            /* a course whose gradescale max is unresolvable makes the true max
+               unknowable — null the max (flag reads unknown) rather than
+               understate it by keeping the course's credits in the
+               denominator only */
+            if(
+                coalesce(potentialcrhrs_current_max_known, 0)
+                = coalesce(potentialcrhrs_current, 0),
+                safe_divide(
+                    weighted_points_projected_max_current, potentialcrhrs_current
+                ),
+                null
             ) as gpa_max_current_raw,
         from points_rollup
     )
--- a/src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_cumulative_year.sql
+++ b/src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_cumulative_year.sql
@@ -1,0 +1,130 @@
+with
+    stored_grades as (
+        select
+            sg.studentid,
+            sg.schoolid,
+            sg.academic_year,
+
+            if(sg.excludefromgpa = 0, sg.potentialcrhrs, null) as potentialcrhrs,
+            if(sg.excludefromgraduation = 0, sg.earnedcrhrs, null) as earnedcrhrs,
+            if(sg.excludefromgpa = 0, sg.gpa_points, null) as gpa_points,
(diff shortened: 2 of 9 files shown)


anthonygwalters:
## Prod-data validation results (Task 5, aggregates only)

All validation ran against dev builds whose parents were forced to **live prod relations** via `--defer --state target/prod --favor-state` (a first pass silently read stale March dev copies of the parent tables — the repo's documented stale-dev-shadow trap; rebuilt with `--favor-state` before recording these results).

### 1. Projected-row reconciliation (governance check — required)

Year model `is_projected = true` rows joined to `int_powerschool__gpa_cumulative` on `(studentid, schoolid)`, comparing `cumulative_y1_gpa` vs `cumulative_y1_gpa_projected` and the unweighted pair with `is distinct from` (exact, NULL-safe):

| District   | Projected rows | Unjoined | Weighted mismatch | Unweighted mismatch |
| ---------- | -------------- | -------- | ----------------- | ------------------- |
| kippnewark | 6,802          | 0        | 0                 | 0                   |
| kippcamden | 2,321          | 0        | 0                 | 0                   |
| kippmiami  | 1,504          | 0        | 0                 | 0                   |

Exact match, zero exceptions — the projected point is joined from the original model, so this confirms the join is sound end-to-end.

### 2. Stored-row count spot check

Year model stored rows vs distinct `(studentid, schoolid, academic_year)` in prod `stg_powerschool__storedgrades` (`storecode = 'Y1'`, `academic_year < 2025`):

| District   | Model rows | Expected | Match |
| ---------- | ---------- | -------- | ----- |
| kippnewark | 41,753     | 41,753   | exact |
| kippcamden | 10,960     | 10,960   | exact |
| kippmiami  | 4,714      | 4,714    | exact |

### 3. Test suite (per district: unit tests, uniqueness, reconciliation singular test)

6/6 PASS in kippnewark, kippcamden, and kippmiami on live prod inputs. kipppaterson parses clean with the new model and its tests auto-disabled.

### Deviation from the plan worth reviewer attention

The reconciliation test's tolerance predicates were widened from the plan's `> 0.01` to `> 0.015`. On real data, 28 rows warned at 0.01 — every one a `.xx5` rounding cusp where the two models' different float summation orders round to opposite sides (stored values exactly 0.01 apart, which FLOAT64 `abs()` evaluates as slightly greater than 0.01). The spec explicitly says the tolerance must absorb this case; 0.015 passes the one-cent flip while still catching real drift (0.02+). After the change: 0 rows in all three districts. Documented in the test SQL comment and YAML description.

### Final whole-branch review

Run on the full branch diff: no Critical findings; the one Important finding (dict-format unit-test fixture mocking a same-PR column add — the documented CI introspection trap) is fixed in `697571abd` (`format: sql`). Remaining minors were triaged ship-as-is; notably, the theoretical unmatched-gradescale edge in the attainability ceiling was checked against prod and is currently vacuous (0 unmatched rows out of 99k/33k/45k current-year graded rows per district).

src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_cumulative_year.sql
cbini, line 1: could this just be folded into gpa term?

cbini: Evaluated whether this should fold into `int_powerschool__gpa_term`. Keeping it as a standalone year-grain model — the two differ on two independent axes:

**1. Grain / aggregation.** `gpa_term` is per-term / semester / single-year and not cumulative; this model is a running cumulative across years. (Year-end is effectively a filter on an "as-of" running series, so a term-grain as-of version is conceivable — but the credit-basis point below is the real blocker.)

**2. Credit basis.** `int_powerschool__gpa_cumulative` — which this model must reconcile to, and joins verbatim for the projected current-year point — computes on the **stored grade record's** credit hours (`potentialcrhrs`): the authoritative, post-override "hours actually awarded" basis. `gpa_term` uses the **course-catalog** hours (`courses.credit_hours`) for the Y1 denominator, with an inner join that drops zero-credit courses. Those bases diverge whenever a stored record's credits are overridden away from the catalog default — so a running cumulative built on `gpa_term` would not reconcile with `gpa_cumulative`, breaking the exact-match guarantee the reconciliation test enforces.

Not a bug: the basis difference is intentional — `gpa_cumulative` reflects override-aware earned credits, `gpa_term` is an operational per-period view. Unifying the two would mean aligning their credit bases first, which is a broader refactor of a widely-consumed model and out of scope for this PR.

src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_cumulative.sql
```diff
@@ -214,8 +229,52 @@ with
...
+                    null
+                )
+            ) as potentialcrhrs_current,
+            sum(
+                if(
+                    academic_year = {{ var("current_academic_year") }},
+                    weighted_points_projected_max,
+                    null
+                )
+            ) as weighted_points_projected_max_current,
```
cbini, line 259: Numerator/denominator basis mismatch that can misreport attainability. This sum (`weighted_points_projected_max_current`) drops a row when `gpa_points_projected_max` is NULL, which happens when the left join to `gradescale_max` on `fg.courses_gradescaleid` (line 106) finds no match — an orphaned or non-top-level gradescale id (the lookup is filtered to `parent.gradescaleid = -1`). But `potentialcrhrs_current` (line 252) still counts that course's credits. So `gpa_max_current_raw` is understated, and `is_cumulative_3_0_attainable` (line 316) can flip to false for a student who is actually on track. If every current course is orphaned, the flag goes NULL while `gpa_needed_for_cumulative_3_0` still shows a real number.

Prod is at 0 orphans today per the validation comment, so no live wrong values — but the path is untested (all unit fixtures use `courses_gradescaleid` 10, which always matches). Suggest either restricting the denominator to the same non-null-max rows, coalescing a sane fallback max, or adding a data-quality check that the orphan count stays 0, plus a fixture covering an unmatched `courses_gradescaleid`.

src/dbt/powerschool/models/sis/intermediate/properties/int_powerschool__gpa_cumulative.yml
cbini, line 1: Two things while this model is being modified:

1. (Important) No uniqueness test. Per `src/dbt/CLAUDE.md`, all intermediate models must have a uniqueness test, and this file has no `data_tests` block at all. The model is unique by construction (`points_rollup` is `group by studentid, schoolid`), but this PR now leans on that grain — both the new projected-row join in `int_powerschool__gpa_cumulative_year` and the reconciliation singular test assume one row per (studentid, schoolid). Add `dbt_utils.unique_combination_of_columns` on (studentid, schoolid).

2. (Minor) The model has no model-level `description` and its pre-existing columns carry only `data_type`. The convention asks a modified model to document the model and every column. The four new columns are well described; a model-level `description` at minimum would close the gap (retrofitting all legacy columns can be a follow-up).

src/dbt/powerschool/tests/properties.yml
```diff
@@ -32,3 +32,23 @@ data_tests:
           ref:
             name: int_powerschool__student_enrollment_union
             package: powerschool
+
+  - name: test_int_powerschool__gpa_cumulative_year__reconciles_stored
+    description: >-
+      Drift guard for the year-grain recompute — for each student-school, the
+      latest completed-year row of int_powerschool__gpa_cumulative_year must
+      match int_powerschool__gpa_cumulative within one rounding step (0.01) on
```
cbini: The opening clause says `within one rounding step (0.01)` but the enforced threshold (explained two sentences down) is 0.015. Drop the `(0.01)` parenthetical so a reader who stops at the first clause does not cite the wrong number.

cbini (changes requested):
## Independent review — requesting changes

Fresh dbt analytics-engineering pass, cross-checked against the existing claude-review bot findings and the author's prod validation. CI is green and the core is strong: the needed-GPA algebra is correct (re-derived by hand), the governance join-not-recompute invariant holds, the grain is provably unique, and the unit tests validate real math. Requesting changes to resolve (or explicitly defer) two items before merge, plus two minor polish notes. Details are on the inline comments; summary below.

### Important
1. `is_cumulative_3_0_attainable` can wrongly read false when a current-year course has an orphaned `courses_gradescaleid` — the max-GPA numerator drops the NULL while the denominator keeps the credits, understating the ceiling. Prod is at 0 orphans today, so no live wrong values, but the path is untested and unguarded. Fix the numerator/denominator asymmetry or add a data-quality check that the orphan count stays 0.
2. `int_powerschool__gpa_cumulative` has no uniqueness test, and this PR newly depends on its (studentid, schoolid) grain for the projected-row join and the reconciliation test. Add `dbt_utils.unique_combination_of_columns` on (studentid, schoolid), or explicitly defer.

### Minor
3. The modified model lacks a model-level `description` and its legacy columns are undocumented — the convention expects a modified model to carry a description on the model and every column.
4. The reconciliation-test description opens with `within one rounding step (0.01)` but the enforced threshold is 0.015.

Neither Important item is a hard correctness blocker on today's data, but both are cheap to close while the files are open, and item 1 is a silent-regression risk if left unguarded.

_Separately: the "fold into `int_powerschool__gpa_term`" question on the earlier thread was evaluated and resolved (keep separate — different grain and a deliberate stored-record vs course-catalog credit basis)._

src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_cumulative.sql
anthonygwalters: Confirmed and fixed in 3c41d1dbd. Reproduced it first as a unit fixture (student 6: one course on a known scale, one on an orphaned `courses_gradescaleid`) — pre-fix the flag read `false` for a student whose needed GPA (3.0) was plainly attainable.

On the remedy: restricting the denominator to the non-null-max rows would compare `gpa_needed_raw` (over **all** current credits) against a max computed over a **subset** — still mismatched bases, and neither an upper nor lower bound. And no principled coalesce fallback exists across weighted scales (4.0 vs 4.5 vs 5.0 maxima). So the guard instead nulls `gpa_max_current_raw` whenever the known-max credit basis differs from `potentialcrhrs_current`: an unresolvable max makes the true max unknowable, and the flag now reads unknown (NULL) rather than a misstated false. All other NULL paths already null both sides consistently, so the guard only fires on the genuine orphan case.

Also added the data-quality alert you suggested: a where-scoped `not_null` on `is_cumulative_3_0_attainable` (`where gpa_needed_for_cumulative_3_0 is not null`), warn severity — under the guard, that condition is exactly "an orphan reached prod." Passes today (0 orphans), and unit + reconciliation tests all pass post-fix.

src/dbt/powerschool/models/sis/intermediate/properties/int_powerschool__gpa_cumulative.yml
anthonygwalters: Both added in 3c41d1dbd — `dbt_utils.unique_combination_of_columns` on (studentid, schoolid), passing against a prod-data dev build (19.3k rows), plus a model-level description. Left the legacy per-column retrofit as a follow-up per your note; the four new columns keep their descriptions.

src/dbt/powerschool/tests/properties.yml
anthonygwalters: Dropped the parenthetical in 3c41d1dbd — the sentence now defers the number to the threshold explanation below it.

anthonygwalters:
@cbini responses inline

cbini (approved):

PR #4319, opened 2026-07-02
feat(finalsite): filter Focus enrollment extract to new and changed rows

### LINKED ISSUE #4317: feat(finalsite): filter Focus enrollment extract to new and changed records
## Problem

`rpt_focus__student_enrollment` (kipptaf) is a full snapshot of Finalsite-sourced enrollments pushed to Focus over SFTP. It re-sends every row every run, including students whose enrollment already exists unchanged in Focus.

Reconciliation of the 1,585 SY2026 (26-27) extract rows against `dagster_kippmiami_dlt_focus.student_enrollment`:

| Category | Rows | Should send? |
| --- | --- | --- |
| Not in Focus (new registrations) | 470 | yes (new) |
| Fully identical to Focus | 1,093 | no (redundant) |
| Extract adds `end_date` only (no drop_code) | 19 | yes (change) |
| Extract adds withdrawal (`drop_code` + `end_date`) | 4 | yes (change) |

So ~1,093 of the ~1,585 rows are pure re-sends.

## What is actually comparable between the two sources

Only three columns are directly comparable, and post-floor-fix (#4312) they all match on matched students:

- `student_id` — same 10-digit `8400`-prefixed id space
- `syear`
- `start_date` — all `20260811`, matches Focus `2026-08-11`

`school_id`, `grade_id`, and `enrollment_code` differ only as **import codes vs Focus internal ids** (extract emits `2332A` / `01` / `KG` / `E01` / `E05`; Focus stores `58` / `339` / `1507`). These are translated by the Focus importer on load and are NOT usable for change detection. The only substantive differences on matched rows are the exit fields (`drop_code` / `end_date`).

## Proposed fix

Add an anti-join in `rpt_focus__student_enrollment` against Focus `student_enrollment`, keyed on the comparable columns:

```text
emit a row when:
  (a) no Focus enrollment exists for (student_id, syear, start_date)   -- new, or
  (b) a Focus row exists on that key but has no exit recorded
      (drop_code IS NULL AND end_date IS NULL)
      while the extract carries a drop_code or end_date               -- changed
```

Do NOT compare `school_id` / `grade_id` / `enrollment_code` (different id systems).

Implementation:

1. Add `student_enrollment` to the kipptaf BQ-native Focus source (`models/focus/sources-bigquery.yml`) — the file added in #4312.
2. Add the anti-join to `rpt_focus__student_enrollment.sql` (keep the enrolled-only gate and the first-day floor).
3. Update the model unit tests to mock the new source, covering: new student, identical-so-filtered, end_date-added, withdrawal-added.

## Open question to confirm before shipping

This turns the feed from a full snapshot into a delta. That is only safe if the Focus SFTP import is **additive / upsert** (rows omitted are left untouched), not full-replace (omitted rows deleted). The existing 470-row gap between Finalsite and Focus strongly implies additive, but confirm with the Focus import owner before merge.

Refs #4312

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will make `rpt_focus__student_enrollment` a **delta feed** — it stops re-sending Focus enrollments that already exist unchanged.

Previously the extract emitted a full snapshot every run (1,587 rows on current data) even though most of those enrollments already exist identically in Focus `student_enrollment`. Now it emits only:

- enrollments **absent** from Focus (new registrations), and
- enrollments present in Focus but where the extract carries an **exit** (transfer-out or `end_date`) the Focus row has not recorded yet. `end_date` and `drop_code` are gated independently, so a `drop_code` that arrives after Focus already has the `end_date` still syncs.

Matching is on `student_id`, `syear`, and the floored `start_date` — the only fields comparable across the two systems. `school_id`, `grade_id`, and `enrollment_code` are Focus import codes translated on load (e.g. `2332A` maps to `58`, `01` maps to `339`, `E01` maps to `1507`), so they are not used for change detection.

Validated against current prod data: 1,587 rows drop to 495 (472 new + 23 changed), suppressing 1,092 redundant rows.

Changes:

- Add `student_enrollment` to the BQ-native Focus source (`models/focus/sources-bigquery.yml`).
- Anti-join it in `rpt_focus__student_enrollment.sql`, keeping the enrolled-only gate and the first-day floor.
- Add a unit test covering the delta cases (new, identical-suppressed, end_date-added, withdrawal-added, fully-recorded-suppressed, partial-exit-synced).

**Behavioral note (confirmed):** The Focus SFTP import is an **upsert** — omitted rows are left untouched and re-sent rows update in place. So suppressing already-matched enrollments is safe, and re-emitting a partial-exit row updates the existing Focus enrollment rather than duplicating it. Confirmed with the import owner.

Closes #4317

## AI Assistance

Claude Code authored the investigation, the anti-join, and the unit test under my direction. The delta semantics and the upsert confirmation were human-directed.

### dbt

- [x] Properties file updated for the modified model (description plus new unit test).
- [ ] Exposure — n/a (consumed by the kippmiami SFTP wrapper, not a dashboard).
- [x] Not a breaking schema change — same 28-column contract; the behavioral (row-count) change is noted above.
- [x] Source is BQ-native (`dagster_kippmiami_dlt_focus`, hardcoded schema), not an external table, so `stage_external_sources` does not apply — CI reads prod in all targets.

## CI checks

- [x] Trunk — passing
- [x] dbt Cloud — passing, no warnings (needed a Clone - Staging refresh so the deferred `int_focus__school_year_first_day` from #4312 resolved)
- [x] Dagster Cloud — not triggered (no Dagster code paths changed)

### DIFF
--- a/docs/reference/finalsite-focus-import.md
+++ b/docs/reference/finalsite-focus-import.md
@@ -94,6 +94,41 @@ start date yet) are held back and not sent until they enroll.
 > minted ID; the enrollment record waits for the start date. This is expected —
 > the enrollment flows to Focus once Finalsite records the start date.
 
+### Enrollment start date — aligned to the first day of school
+
+Finalsite often records an enrollment's start date as the **registration or
+contract date**, which can fall weeks before school actually starts. Focus
+records enrollments as of the **first day of school**. To make the two line up,
+the pipeline raises any start date that falls before the first day up to the
+first day of school; a start date already on or after the first day is left
+unchanged (so a genuine mid-year enrollment keeps its real date).
+
+The first day of school is read from **Focus's own attendance calendar** (the
+earliest instructional day of the year), so it always matches what Focus expects
+— for 2026-27 that is August 11, 2026. A school year with no calendar in Focus
+yet leaves its start dates unchanged.
+
+> **This is what lets the enrollment feed match cleanly.** Because the start
+> date is aligned to Focus's first day, a student already enrolled in Focus is
+> recognized as the same enrollment instead of being re-sent with a date that
+> doesn't line up.
+
+### How the pipeline knows Focus already has an enrollment
+
+An enrollment counts as "already in Focus" when Focus has a record for the same
+**student, school year, and start date** — the three values that mean the same
+thing in both systems. Matched enrollments are not re-sent (see the exception
+for withdrawals below).
+
+School, grade, and entry code are **not** used to decide this. Focus stores its
+own internal codes for them and translates the imported values on load, so they
+never match on a literal comparison even when they describe the same school and
+grade — comparing them would make every record look new.
+
+The one thing that will still be sent for a matched enrollment is a **new
+withdrawal**: if the extract now has an end date or drop code and the Focus
+record has none, that exit is filled in (see Withdraw / drop codes above).
+
 ### What gets sent
 
 Every feed follows the same import-once rule — a record is sent only when Focus
@@ -114,6 +149,18 @@ Finalsite ID and moves the start date forward). The pipeline ensures a **new
 enrollment never carries a previous year's drop code or end date** — a
 last-attended date before the current enrollment's start is treated as belonging
 to the prior enrollment and is ignored.
+
+### Address and contact formatting
+
+Household address values are cleaned before they go to Focus so the imports stay
+consistent:
+
+- A blank or space-only field (street, city, state, zip) is sent as **truly
+  empty**, not as a stray space.
+- The **state is upper-cased** (e.g. `fl` becomes `FL`).
+
+This is formatting only — it does not change which addresses or contacts are
+sent, and it follows the same import-once rule as everything else.
 
 ## Where to make corrections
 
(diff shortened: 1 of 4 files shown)


cbini:
Addressed in `fa2dfbbef`.

**Important — partial-exit suppression gap:** Confirmed against the upstream model. `int_finalsite__enrollment_lifecycle` sets `is_transfer_out = (enrollment_start_date is not null and enrollment_end_date is not null)` and derives `enrollment_end_date` from `withdrawal_last_attended_date`, while `drop_code` comes from the independent `cca.fl_state_withdraw_codes_ss` attribute — so a Focus row can hold `end_date` without `drop_code` (in fact ~19 SY2026 rows are `end_date`-only today). Fixed by gating the two exit fields independently:

```sql
where
    fe.focus_student_id is null
    or (e.enrollment_end_date is not null and not fe.has_end_date)
    or (
        e.is_transfer_out
        and cca.fl_state_withdraw_codes_ss is not null
        and not fe.has_drop_code
    )
```

A later-arriving `drop_code` now syncs even when Focus already has the `end_date` (and vice versa). No change on current prod data (kept count stays 1,587 to 495; 0 rows rescued today) — this only affects the future partial-exit sequence you described.

**Minor — missing coverage:** Added two cases to `test_student_enrollment_focus_delta` — a Focus row with both exit fields already recorded (suppressed) and one with `end_date` but no `drop_code` where the extract's `drop_code` still emits (the partial case above).

**Minor — redundant condition:** Removed. The restructure drops `is_transfer_out or enrollment_end_date is not null` — for enrolled-only rows the two are equivalent, so the `end_date` branch now keys on `enrollment_end_date` alone.

**Note — schema types:** Verified via BigQuery. `student_enrollment` is `student_id` INT64, `syear` INT64, `start_date` DATE, `drop_code` INT64, `end_date` DATE — the `e.start_date = fe.focus_start_date` comparison is DATE to DATE.

PR #4322, opened 2026-07-02
fix(finalsite): gate blank records out of Focus import-once feeds

### LINKED ISSUE #4320: fix(finalsite): gate blank records out of the Focus import-once feeds
## Problem

The `kippmiami` Focus import-once feeds (`rpt_focus__addresses`, `rpt_focus__contacts`, `rpt_focus__demographics`) drop a student once a matching row exists in the Focus destination — a **presence-based** anti-join with no check that the payload is populated. So a student first imported with a blank payload gets an empty Focus record created and is then suppressed forever; the real value never syncs.

Concretely, ~124 students currently have a fully-null address block in the desired-state feed. Under presence-based import-once they would import as blank and lock out the real address as Focus address linkage fills in (it is nearly empty today — mid-rollout).

## Fix

Add a completeness gate to the import-once feeds so a blank record is never imported (defer until the value exists), mirroring the enrollment enrolled-only gate:

- **Addresses** — emit only when `address`, `city`, `state`, `zipcode` are all present.
- **Contacts** — emit a contact row only when it has a name (`first_name` or `last_name`).

**Demographics — intentionally not gated** (owner decision). It stays student-presence-only import-once; its optional fields (DOB, race, etc.) can still lock in if null at first import. Revisit separately if that becomes a problem.

Unit tests updated on both models: a complete/new record is kept, an already-in-Focus student is suppressed, and a blank record is suppressed.

Validated against prod (dev-defer build): addresses 1,880 to 1,756 (defers exactly the 124 blank-address students); contacts unchanged (no nameless contacts today — gate is defensive).

## Notes

- District-only change (kippmiami); kipptaf CI is a no-op, so verified with a local `dbt build` against kippmiami.
- Import-once ownership model is unchanged (first complete record wins; Focus owns it after). Later value changes still will not re-sync — that is the separate value-diff question from #4279.

Refs #4279

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request stops the Focus import-once feeds from locking in blank records.

`kippmiami/rpt_focus__addresses` and `rpt_focus__contacts` use presence-based import-once (drop a student once a matching row exists in the Focus destination), with no check that the payload is populated. So a student first imported with a blank address (or a nameless contact) gets an empty Focus record created and is then suppressed forever — the real value never syncs. About 124 students currently have a fully-null address block and would lock in as Focus address linkage fills in (mid-rollout).

Fix — add a completeness gate so a blank record is deferred until it is populated (mirrors the enrollment enrolled-only gate):

- **Addresses** — emit only when `address`, `city`, `state`, `zipcode` are all present.
- **Contacts** — emit a contact row only when it has a name (`first_name` or `last_name`).

**Demographics is intentionally left student-presence-only** (owner decision); its optional fields can still lock in and can be revisited separately.

Validated against prod via a dev-defer build: addresses 1,880 to 1,756 (defers exactly the 124 blank-address students); contacts unchanged (no nameless contacts today — the gate is defensive). Import-once ownership is unchanged; later value changes still will not re-sync (separate value-diff question from #4279).

## AI Assistance

Claude Code found the lock-in, wrote the gates and unit tests, and validated against prod under my direction. The completeness criteria and the demographics-excluded decision were mine.

### dbt

- [x] Properties updated for both models (descriptions plus a blank-record suppression case added to each unit test).
- [ ] Exposure — n/a.
- [x] No schema/contract change — same columns; only the row filter changed.
- [x] No external source change — reads existing `stg_focus__*` and `kipptaf_extracts` sources.

## CI checks

- [ ] Trunk — passing locally.
- [ ] dbt Cloud — this is a district-only (kippmiami) change; the kipptaf-scoped CI job selects no models, so it is a no-op and does not validate this. Verified locally instead: `dbt build` of both models against prod-defer plus both unit tests pass.
- [ ] Dagster Cloud — not triggered (no Dagster code paths changed); kippmiami redeploys on merge and the automation sensor rematerializes the two views.

Closes #4320

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -477,6 +477,13 @@ or mart `facts`/`dimensions`/`bridges`) —
 `cursor=<evaluationId of the oldest record returned>` — not a timestamp or
 opaque token.
 
+- **Prod dbt models are materialized by `<loc>__automation_condition_sensor`
+  runs** (job `__ASSET_JOB`, tag `dagster/from_automation_condition`), NOT dbt
+  Cloud (CI-only) or crons. A merged model SQL change goes stale on CODE and is
+  rematerialized — including view models (distinct from the data-change
+  condition, which skips views) — within minutes of the post-merge location
+  deploy. To confirm a rollout landed: `get_location_load_history` (new commit
+  LOADED) → `list_runs` / `get_asset_materializations` for the asset.
 - **Schedule/sensor-launched runs report `assetSelection: null`** in
   `list_runs`. Read `stepKeysToExecute` and convert `__` → `/` to recover asset
   keys (`kipptaf__tableau__ops_dashboard` → `kipptaf/tableau/ops_dashboard`).
--- a/docs/reference/finalsite-focus-import.md
+++ b/docs/reference/finalsite-focus-import.md
@@ -140,7 +140,9 @@ does not already have it, and nothing is ever overwritten:
   the student is enrolled** (has a start date); a withdrawal (end date + drop
   code) is filled in once when Focus has none yet.
 - **Addresses and Contacts** — a student's address / contacts are sent only if
-  Focus does not already have them for that student.
+  Focus does not already have them for that student **and** the record is
+  complete (a full address; a named contact). Blank or incomplete records are
+  held back until populated (see below).
 
 ### Forward-moving enrollments are protected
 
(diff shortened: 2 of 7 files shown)


cbini:
Addressed in `6e8e042e3`.

**Important — contacts name-gate misses empty strings:** Confirmed and fixed. Verified against the source: `stg_finalsite__contacts` wraps the address fields in `nullif(trim(...), '')` but leaves `first_name`/`last_name` raw, and prod already has 1 empty-string `last_name` today (that row has a first name, so it's correctly a named contact — but an all-blank-name contact would have slipped the `IS NOT NULL` gate and locked in). Fixed the gate to guard blank/whitespace directly:

```sql
and (
    nullif(trim(d.first_name), '') is not null
    or nullif(trim(d.last_name), '') is not null
)
```

Added an empty-string-names unit-test row (`8400000004` with `'' , ''`) alongside the existing null case; both suppress. I kept the fix in the kippmiami gate rather than normalizing names in `stg_finalsite__contacts` to keep this PR district-only — pushing `nullif(trim())` onto the name columns at staging (to match the address fields) is the more consistent long-term fix and a reasonable follow-up, but it widens blast radius to every name consumer and makes this cross-project.

**Minor — retroactive scope:** Confirmed negligible. Focus linkage is at the very start of rollout — only 1 student is currently linked in `students_join_address` and 1 in `students_join_people` — so there are essentially no already-locked blank rows to unstick. The PR doesn't claim a retroactive fix; if the rollout later surfaces locked blanks, that's Ops-side remediation, not this dbt change.

PR #4325, opened 2026-07-03
feat(kipptaf): sheet-driven promotional status policy engine

### LINKED ISSUE #4324: feat(kipptaf): rebuild promotional status as a sheet-driven policy engine
## Goal

Rebuild `int_reporting__promotional_status` so regional leaders can change promotion policies and cutoffs year-over-year by editing a Google Sheet lookup table — no code changes. The current model hardcodes every region-by-grade policy in roughly 600 lines of CASE statements, and must support complex boolean logic, e.g. retained if fails 2+ core classes OR (ADA below 85% AND 3+ grade levels below in ELA).

## Design (OR-of-ANDs rule table)

One sheet row = one condition. Conditions sharing a `rule_group` AND together; groups OR together within a scope (academic year, region, grade range, term, domain). Any fully-satisfied group makes the domain Off-Track. Any boolean policy can be expressed this way (disjunctive normal form).

Sheet columns:

```text
academic_year | region | grade_min | grade_max | term_name | domain | rule_group
| metric | comparator | value | if_missing | detail_label
```

- `metric` comes from a fixed numeric catalog defined once in code (ADA, absence counts, iReady levels-below, DIBELS/STAR/FAST levels, failing counts, projected credits) plus pseudo-metrics `is_off_track_attendance` / `is_off_track_academic` for the `overall` domain.
- `if_missing` (`met` / `not_met`) makes null handling explicit (replaces buried coalesce behavior).
- `term_name` scoping expresses term-scaled thresholds (e.g. Newark HS absences Q1=6 ... Q4=24) as plain rows.
- `detail_label` drives `attendance_status_hs_detail`.

## Changes

- New source + staging model `stg_google_sheets__reporting__promo_status_policies` (contract enforced; accepted-values, uniqueness, grade-range tests at error severity; warn-level coverage and label-consistency tests). Delete the disabled `promo_status_cutoffs` source/staging attempt.
- New `int_students__promotional_status_metrics` — metric layer extracted from the current model (student x year x term grain, all years).
- New `int_students__promotional_status` — evaluation engine; same output contract as today. SPED exemptions and PowerSchool log overrides (Exempt from retention / Retain without criteria) stay in code. Old `int_reporting__promotional_status` is deleted (models move from `reporting` to `students`).
- `rpt_tableau__promo_status` and `rpt_deanslist__promo_status` update refs only.
- Seed the sheet with current-year policies transcribed from the CASE logic; validate by diffing new vs old statuses for the current year before swap.

Years without policy rows emit NULL statuses (not fabricated On-Track).

## Follow-up

- Backfill the previous 2 academic years of policies into the sheet so historical statuses recompute.

anthonygwalters:
## Summary & Motivation

When merged, this replaces the hardcoded ~600-line `int_reporting__promotional_status` model with a Google-Sheet-driven policy engine so regional leaders can change promotion policies and cutoffs year-over-year by editing a lookup table — no code changes. Closes #4324.

Policies are encoded in disjunctive normal form (OR-of-ANDs): one sheet row is one condition; conditions sharing a `rule_group` AND together, rule groups OR together within a scope (academic year, region, grade range, term, domain). Any boolean policy is expressible, e.g. retained if fails 2+ core classes OR (ADA below 85% AND 3+ grade levels below in ELA).

New models (both under `models/students/`, retiring the `reporting/` placement):

- `stg_google_sheets__reporting__promo_status_policies` — the policy rule table (contract-enforced; accepted-values / uniqueness / grade-range tests at error severity; warn-level coverage and label-consistency tests).
- `int_students__promotional_status_metrics` — fixed numeric metric layer extracted from the legacy model (all years, table).
- `int_students__promotional_status` — the evaluation engine (view). Same output contract as the legacy model. SPED exemptions and PowerSchool log overrides stay code-owned.

A `promo_status_metric_columns()` macro is the single source of truth for the metric vocabulary, reused by the unpivot and the validity test. Comparators are stored as spelled-out tokens (`less_than`, `greater_than_or_equal`, `equals`, ...) because Google Sheets treats a leading equals sign as a formula.

`rpt_tableau__promo_status` and `rpt_deanslist__promo_status` change only their `ref()`; the legacy model and the disabled `promo_status_cutoffs` source/staging are deleted.

### AI Assistance

Claude Code authored the models, tests, macro, and seed transcription under human direction (brainstorm, design, and plan reviewed and approved by the author; the comparator-token design change and scope decisions were human-directed). Diff validation against legacy prod and all reviews were AI-run and human-checked.

## Behavior changes (intended)

Validated by a student-by-student diff of the rebuilt model vs legacy prod for AY2025 (aggregates below; grain matches exactly, 0 new-only / 0 old-only rows; `exemption` and `manual_retention` identical). Every divergence falls into one of two buckets:

1. **NULL where no policy matches** (was a silent `On-Track` in the legacy else-branch): Paterson (deliberately unseeded — policies added later) and Miami grade 3 attendance (legacy had no rule). The new warn-level coverage test flags these.
2. **Legacy operator-precedence bug fix**: the legacy NJ academic CASE ended with an `or n_failing_core` at-least-2 clause, which by SQL precedence fired for any region/grade with 2+ core failures. 133 students (NJ grades 9-12; Miami grades 1/3/4) were legacy-Off-Track purely from this leak and are correctly On-Track now. NJ grades 3-8 still flag core failures in both models.

No new false positives: zero cases of `On-Track` becoming `Off-Track` in any seeded scope.

```text
Per-term status diffs (Q1; other terms materially identical):
  attendance     720  = 167 Miami gr3 + 553 Paterson  (all On-Track to NULL)
  academic       686  = 553 Paterson (to NULL) + 133 precedence-bug fix
  overall        625  = Paterson (to NULL) + precedence-bug cascade
  exemption        0
  manual_retention 0
```

Prior years read NULL until the follow-up backfill (below); both `rpt_` consumers filter to the current academic year, so the dashboard is unaffected.

### dbt

- [x] Properties files included for all new models (descriptions on model + every column).
- [x] External source staged with `--target staging` (source renamed from cutoffs to policies; confirmed the `zz_stg` external carries the new name).
- [x] No contracted-column removals. The renamed model is an intermediate (not a contracted stg/rpt/mart); the two `rpt_` consumers keep identical output columns, so no exposure breaks. The `promotional_status_dashboard` exposure is unchanged.
- [x] New Google Sheets source `columns:` declared at source level; reserved word `value` backticked with `quote: true`.

## CI checks

- [x] Trunk — clean locally on all changed files.
- [ ] dbt Cloud — pending (external staged for CI; the policy sheet holds the 61-row seed so error-severity accepted-values / metric-in-catalog tests pass).
- [ ] Dagster Cloud — expected to run (new sheet asset auto-generated from the manifest).

Note: local `state:modified+` fans out to the whole `google_sheets` source (a documented consequence of editing `sources-external.yml`); the promo chain itself builds clean end-to-end (PASS, 0 errors; only the expected coverage + #3900 uniqueness warns). dbt Cloud CI has the defer state for the full fan-out.

## Post-merge (author / ops)

1. Drop the orphaned prod relation `kipptaf_reporting.int_reporting__promotional_status` (and its `zz_stg` copy) — deleting the model leaves the view behind; Claude is DDL-blocked.
2. Confirm Dagster materializes the new `promo_status_policies` sheet asset and the new models after the location deploy.
3. Open the follow-up issue: backfill the previous 2 academic years of policies into the sheet (labels feat, dbt, google-sheets), per #4324.


anthonygwalters:
@claude review

src/dbt/kipptaf/models/students/intermediate/int_students__promotional_status.sql
```diff
@@ -0,0 +1,317 @@
...
+                from metrics
+            -- trunk-ignore(sqlfluff/LT01)
+            ) unpivot include nulls(
+                metric_value for metric in (
+                    {% for m in promo_status_metric_columns() %}
+                        {{ m }}{% if not loop.last %},{% endif %}
+                    {% endfor %}
+                )
+            )
+    ),
```
cbini, line 38: unpivot should be its own model

src/dbt/kipptaf/models/students/intermediate/int_students__promotional_status_metrics.sql
cbini, line 1: probably worth refactoring pivot ctes into their own models

src/dbt/kipptaf/macros/promo_status.sql
cbini, line 1: are these used anywhere? are they used enough across multiple files to merit macro extraction?

cbini (changes requested):

src/dbt/kipptaf/macros/promo_status.sql
anthonygwalters: Used in two files today:

- `promo_status_metric_columns()` — two call sites in `int_students__promotional_status.sql` (the cast list and the `unpivot ... in (...)` list), one in `tests/stg_google_sheets__reporting__promo_status_policies__metric_in_catalog.sql`.
- `promo_status_pseudo_metric_columns()` — one call site in the model, one in that same test.

So the reuse count is thin. The reason for extraction isn't reuse, it's that those two lists have to stay identical. That test is what stops a typo'd `metric` in the policy sheet from silently doing nothing: an unrecognized metric name just fails the join in `conditions`, so no rule fires and the status comes back NULL instead of erroring. If the test's accepted catalog and the unpivot's metric vocabulary drift apart, the test passes on a metric the engine cannot evaluate.

You're right that it's a new pattern here, though — there's no other list-returning macro in the repo, and `kipptaf/macros/` is otherwise general-purpose utility files rather than per-feature ones.

Splitting the unpivot into its own model (your other comment) opens a cleaner option that drops both macros: the unpivot model's explicit column list becomes the single source of truth, and `metric_in_catalog` checks `policies.metric` against `select distinct metric` from that model, with the two pseudo-metrics spelled inline. Trade-off is the test scanning a column instead of comparing against a literal list.

Leaving the macros in place for now — happy to go either way, so tell me which you'd prefer.

Issue #4346, opened 2026-07-08
feat: replace PowerSchool student contacts with Finalsite for NJ regions


cbini:
## Summary

Replace the source of truth for student contact reporting in all NJ regions (Newark, Camden, Paterson) with the Finalsite Contacts API, deprecating the PowerSchool contact models as the reporting source. Miami's contact sourcing stays PowerSchool for now (handled by the Focus migration work) but is mapped into the new surface shape.

## New contact shape (1+4)

- `contact_1` — the top-ranked contact from the Finalsite contacts module (rank field to be discovered from the API payload; person details resolved via `rel_id`)
- `emergency_1` through `emergency_4` — from 4 sets of custom fields on the student contact record (field names to be discovered from data), not the contacts module
- `contact_2` and `pickup_{n}` slots are removed from the surface network-wide

## Design highlights

- Finalsite package `api/` layer gains a rank field on `stg_finalsite__contact_relationships`, emergency field sets in the custom-attributes pivot, and a new SIS-agnostic `int_finalsite__student_contacts` (grain: enrollment id x contact slot)
- kipptaf gains a long-format union intermediate (NJ Finalsite branches + inline PS-mapped CTE for regions not yet cut over and Miami), feeding a redefined pivot, `dim_student_contact_persons`, and `bridge_student_contacts`
- Contact pivot columns move out of `base_powerschool__student_enrollments` (powerschool package) into kipptaf's `int_extracts__student_enrollments` via the new pivot; ~8 downstream extracts update to the 1+4 columns
- Dagster: `contacts` asset + `FinalsiteResource` + daily schedule per NJ code location, Newark first (only region with API credentials today); Camden/Paterson repeat the wiring as credentials arrive
- Per-region cutover gated on a coverage comparison (Finalsite vs PS contact coverage for enrolled students)

Spec to follow in `docs/superpowers/specs/`.

PR #4366, opened 2026-07-10
feat(finalsite): fall back to financial relationship for contact_1 when no primary

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will change `contact_1` selection in `int_finalsite__student_contacts` to fall back to the relationship flagged `financial` when no relationship is flagged `primary`. Previously `contact_1` was exclusively the `primary` relationship, so enrolled students whose Finalsite record has no `primary` flag (a data-entry gap) got no reportable parent contact at all.

Selection is now: the `primary` relationship, else the `financial` relationship, ranked with a stable `relationship_id` tiebreak for the rare case of multiple `financial` relationships (a student never has two `primary`). The grain `(finalsite_enrollment_id, contact_slot)` is unchanged, and `emergency_1..4` are untouched.

This is a value-only change (no new columns). It was validated against an Ops-provided Finalsite parent export: primary-then-financial reproduces the export's parent-1 selection for the students it covers. The financial-tier tiebreak is arbitrary-but-stable — no stored Finalsite field (neither `relationship_id` nor source array order) reproduces the export's ordering for the ~117 multi-financial students, and both candidates are valid financial contacts.

Built and tested through `kippnewark` with defer to prod: grain uniqueness, `not_null`, and `accepted_values` all pass. Network-wide `contact_1` coverage rises from 10,061 to 10,650 students, with exactly one `contact_1` per student.

Refs #4346

## AI Assistance

AI-assisted (Claude Code): the SQL change, docs, coverage validation, and this PR. Human-directed: the selection rule (primary then financial, no relationship-type logic) and the tiebreak decision, chosen after reviewing the Ops export.

### dbt

- [x] Properties file updated for the modified model (description reflects the new selection rule)
- [x] Exposures — no change; downstream column set and grain are unchanged
- [x] Breaking change? No — value-only; grain and columns unchanged, so contracted downstream consumers are unaffected
- [x] External sources — none added or modified

## CI checks

- [ ] Trunk — passes
- [ ] dbt Cloud — passes
- [ ] Dagster Cloud — not triggered (no Dagster path changes)

### DIFF
--- a/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
+++ b/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
@@ -1,13 +1,48 @@
 with
+    contact_1_candidates as (
+        -- contact_1 is the student's ONE reportable parent contact. Finalsite
+        -- has no explicit contact rank, so we take the relationship it flags
+        -- `primary` (its parent1 designation) and fall back to the one flagged
+        -- `financial` when no primary is set — the two flags Ops maintains to
+        -- mark the responsible caregiver. A record with neither flag gets no
+        -- contact_1 (a Finalsite data-entry gap for Ops to resolve).
+        -- `primary`/`financial` are NULL (not false) when unset, so normalize
+        -- to false here to keep the downstream rank ordering deterministic. No
+        -- SIS scoping — downstream receivers filter to enrolled students by
+        -- joining on the student id.
+        select
+            finalsite_enrollment_id,
+            relationship_id,
+            rel_id,
+            rel_name,
+            rel_type,
+
+            coalesce(is_primary, false) as is_primary,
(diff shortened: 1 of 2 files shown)


cbini:
Verified the rank-before-join flag against prod staging data: **0 recoverable losses**. Every student with a `primary`/`financial` candidate that resolves to a contact record receives a `contact_1` (`students_with_resolvable_candidate` = `students_with_contact_1` = 10,650) — no student is dropped because the top-ranked pick failed the join. Keeping the rank-then-join shape (simpler, and confirmed exhaustive).

Kept `is_financial desc` as an explicit primary-then-financial marker.

The two dbt Cloud test warnings (`rpt_tableau__kfwd_career_launch_survey`, `rpt_appsheet__kfwd_career_launch_survey` uniqueness) are on career-launch-survey models unrelated to this change — pre-existing.

PR #4370, opened 2026-07-10
feat(kipptaf): multi-grain topline cascade dashboard (week/month/quarter/YTD)

### LINKED ISSUE #4363: feat(kipptaf): topline cascade dashboard multi-grain aggregates (week, month, quarter, YTD)
## Problem

`rpt_tableau__topline_cascade_dashboard` is costly and only produces week-grain aggregates, while stakeholder requests have evolved to need month, academic-quarter, and YTD views.

Cost driver: `int_topline__dashboard_aggregations` is a **view** — six student-level GROUP BYs plus goal range-joins over `int_topline__student_metrics` (student x week x indicator x layer) recompute on every Tableau extract refresh (daily 5am cron) and any datasource touch.

## Design summary

1. **Periods dimension** (`int_topline__periods`): per school x academic year — weeks (existing spine), exact calendar months, reporting-term quarters (`dim_terms` type RT), and one YTD row. Columns: `period_type`, `period_start`, `period_end`, `period_label`, `is_current_period`, `is_most_recent_complete_period`. In-progress periods included; effective window end is capped at current date.
2. **Per-indicator rollup config**: new `period_rollup` column (`as_of` | `period`) on the topline aggregate goals sheet. As-of indicators (enrollment, GPA, proficiency snapshots, retention, seats staffed, staff metrics) take the last complete week's value within each period — no upstream changes. Period-scoped indicators (ADA, chronic absenteeism, truancy, suspensions, contacts, CA interventions, i-Ready lessons/time-on-task) get period-grain variants computed from date-bearing sources joined to the periods dimension — exact boundaries, roughly 8 models.
3. **Tandem period-goals tab** (long format, new staging model): keyed by the sheet's entity keys plus nullable `academic_year`, `period_type`, `period_label`, with `goal`. Blank = wildcard. Resolution is most-specific-wins via three explicit left joins + coalesce: exact period instance, then grain default, then base `goal` from the config sheet. Supports ramp goals (e.g. ADA October vs November) with no config-row duplication.
4. **One aggregation pass, materialized**: `int_topline__dashboard_aggregations` keeps its shape but groups by `period_type` + period and becomes `materialized: table` (daily Dagster rebuild replaces per-refresh recompute — the main cost fix). The `rpt_tableau__*` extract stays a thin view, adding `period_type`/`period_label`; `term`/`term_end` hold period bounds so existing workbook fields keep resolving.
5. **Rollout/tests**: `period_type = week` rows identical to current output (regression-checked); YTD rows reconcile against current running values for period-scoped indicators; PK uniqueness test on the aggregation table including `period_type`.

Stacked long output lets one Tableau dashboard show multiple grains simultaneously; every worksheet must filter `period_type`.

Spec to follow in `docs/superpowers/specs/`.

anthonygwalters:
# Pull Request

## Summary & Motivation

Closes #4363

When merged, this pull request will rebuild the topline cascade dashboard
pipeline to produce aggregates at **week, month, academic-quarter, and YTD**
grains (config-driven per indicator) and cut its cost by materializing the
aggregation as a **table** (previously a view recomputed on every Tableau
refresh).

This PR bundles two features plus review fixes:

- **Multigrain rebuild** (original scope, #4363).
- **Goal-difference feature** (#4487): per-level difference-from-goal columns
  and enrollment count-vs-target columns.
- **CBini review fixes** (see Review response). The region/org as-of correctness
  fix (must-fix #2) is **deferred to a follow-up (#4514)** — the first approach
  proved unworkable and needs a redesign.

Specs / plans:

- Multigrain rebuild: `docs/superpowers/specs/2026-07-09-topline-multigrain-design.md`
  · `docs/superpowers/plans/2026-07-09-topline-multigrain.md`
- Goal-difference (#4487): `docs/superpowers/specs/2026-07-21-topline-goal-difference-design.md`
  · `docs/superpowers/plans/2026-07-21-topline-goal-difference.md`

### What changed

- **`int_topline__periods`** (new, table): periods dimension — Mon-Sun weeks,
  exact calendar months, reporting-term quarters, one YTD row, per
  school/region/org scope, with `TAF` region alias for central-office staff.
- **Period-scoped student variants** (new): attendance (ADA / truancy /
  chronic absenteeism), suspensions, attendance contacts, i-Ready lessons —
  recomputed within exact period windows from daily/event sources, unioned in
  **`int_topline__student_metrics_periods`**.
- **As-of indicators need no upstream changes**: the aggregation relabels each
  series' last available week inside a period, gated by the
  `seed_topline_period_rollup` config (`as_of` vs `period` per indicator).
- **`int_topline__dashboard_aggregations`** rewritten (multi-grain, one
  aggregation pass, `materialized: table`) with 4-branch most-specific-wins
  period-goal resolution from `seed_topline_period_goals`.
- **Region normalization**: new `region_to_city` macro (values match
  `dim_regions.name`); staff-side models and the goals staging normalize to
  Newark / Camden / Miami / Paterson / TAF. The extract's partial region
  decode is removed.
- **`rpt_tableau__topline_cascade_dashboard`** exposes `period_type`,
  `period_label`, `is_current_period`, `is_most_recent_complete_period`, plus
  the goal-difference columns (see Review response). Tableau contract: **every
  worksheet must filter `period_type`** or aggregates mix grains.
- **Seeds are dev stand-ins** for the future goals-sheet tab and
  `period_rollup` column — sheets are the final home (deferred cutover is
  plan Task 13, separate PR once stakeholders populate real data).

### Region scope: Miami removed (#4943)

Added after the review below, at the dashboard owner's request. Miami's SIS
moved from PowerSchool to Focus and `kippmiami_powerschool` is a frozen
pre-migration archive (final ODBC pull 2026-07-01), so
`int_powerschool__calendar_week` has no Miami AY2026 rows and most Miami tiles
render blank. One path was worse:
`int_topline__student_retention_weekly_aggregations` reads
`int_students__retention_over_time`, which builds its own timeline to
`current_date` and never joins the calendar — so it kept publishing AY2026
Miami rows off the frozen archive. Prod held 72 such rows through 2026-08-17,
dragging network retention roughly 11 points below Newark and Camden. A
plausible-looking wrong number rather than a blank tile.

Miami is therefore removed entirely, all years, including from org-level
network totals, until Focus integration lands. Five predicates, one per union
input, matching the exclusion pattern already used in
`int_tableau__fresh_enrollment_scaffold`. All five carry a `TODO(#4943)` marker
so `git grep` enumerates the whole dam at reversal.

Two simpler placements were tried and rejected: filtering `int_topline__periods`
leaves every Miami week row alive (`flags` joins periods LEFT; only
`as_of_candidates` joins it INNER), and filtering the extract leaves Miami
inside org totals (org rows are a separate aggregation pass keyed `All as
region` that omits region from the group by, and `seats_staffed` and
`student_retention` fan out to org inside themselves).

Consequences worth calling out:

- AY2025 org Total Enrollment at week grain drops from 10,119 to 8,773. The ytd
  figure was already Miami-free, so this also makes the four grains agree with
  each other.
- `FAST` and `Star Diagnostic` are Miami-only and disappear from every scope,
  including their org rows (660 and 636 rows respectively, both to zero).
- `State Assessments` is worse than a value change, correcting an earlier
  version of this section: it goes 876 to 678 rows and **loses all 72 of its
  `org_level = 'org'` rows**. Every surviving row has `org_level`, `has_goal`,
  and `goal` null, so the indicator persists only as an ungoaled remnant.
  Pre-dam, 64 of those 72 org rows carried a value numerically identical to a
  Miami region row at the same period and discipline. This needs an owner
  decision rather than silent acceptance.
- The `accepted_values` list on `int_topline__staff_metrics.region` is
  deliberately left intact — it is a subset check, so it passes with Miami
  absent, and narrowing it is the one edit that would make re-inclusion harder.
- **Paterson is unchanged and still excluded.** Three of the five predicates are
  the lines that already excluded Paterson, so they became deny-lists of both
  regions. Paterson has zero rows in the topline goals sheet, and
  `metric_aggregate_value` is a `case` on the goals-sourced `aggregation_type`,
  so including Paterson before those rows exist would null every Paterson value
  rather than show metrics without goals. Dropping `'Paterson'` is a one-token
  follow-up once the sheet is populated — tracked in #4943.

### Review response (CBini)

- **Region/org/TAF as-of undercount (must-fix #2)** — **deferred to #4514.** The
  first approach (roll up school-grain components) proved unworkable: grade-band
  variants (`0-2` / `0-4` / `5-8` / `0-12` / ...) do not nest across grains, so
  org/region per-band values cannot be reconstructed from school-grain
  components. The correct fix (resolve each school's own as-of week at the
  raw-student level, then re-aggregate) is a larger refactor, tracked in #4514.
  It is dormant during the summer gap; school-level and weekly numbers are
  correct.
- **Staff org totals (must-fix #1)** — fixed: dropped `region` from the staff
  org-level group-by (was: partial per-region rows all labeled `All`,
  colliding). This also removed ~112 duplicate-key rows from the PK-uniqueness
  warn.
- **`region_to_city` leak (should-fix #3)** — fixed: added a fail-loud
  `accepted_values` test (`[Newark, Camden, Miami, Paterson, TAF]`, severity
  error) on the `int_topline__staff_metrics.region` output.
- **Missing uniqueness tests (should-fix #4)** — fixed:
  `dbt_utils.unique_combination_of_columns` on
  `stg_google_sheets__topline_aggregate_goals` (severity error), and on
  `rpt_tableau__topline_cascade_dashboard` (severity warn, mirroring the
  upstream PK grain — warn because pre-existing duplicate-key rows reach the
  extract).
- **`goal_difference_percent` zero-goal guard** — fixed: wrapped the division in
  `safe_divide`, matching `progress_to_goal_pct`.
- **Goal-difference feature (#4487)** — per-level difference-from-goal columns
  (percentage-point for proportions, integer for counts) and enrollment
  count-vs-target columns. The enrollment target-chain gap is pre-existing and
  tracked in #4489.

### AI Assistance

Authored by Claude Code (subagent-driven development) under human direction:
design, per-indicator rollup semantics, region-normalization scope, and seed
strategy were user decisions; implementation, per-task review, and reconciliation
verification were AI-executed.

### Verification

- Full downstream build: PASS=14 / ERROR=0; branch-wide lint clean.
- Week-grain regression vs prod: exact row-count and value match per
  (metric_type, academic_year, org_level, layer, indicator) on all Student
  Metrics groups; staff-side groups differ only via the intentional region
  normalization. **This claim is now superseded for `org` rows (and the
  `region = 'All'` rows) by the Miami removal described above** — those
  intentionally no longer match prod, because prod's network totals still
  include Miami. Newark and Camden rows are unaffected at both school and
  region grain: re-measured against prod, 244/244 week-grain groups match on
  row count, and AY2025-only matches 127/127 on row count with a maximum
  group-sum delta of 0.017, consistent with the comm-log dedup
  non-determinism already noted below.
- YTD reconciliation: period-scoped YTD values match the running weekly
  values (ADA exact for single-school students; suspensions drift
  0.0000-0.0002; contacts AY2024 100% / AY2025 99.98%).
- Goal-difference feature verified in CI. Originally 41,058 non-null
  (36,539 numeric + 4,519 integer); **post-Miami-removal 37,914
  (33,923 + 3,991)**. Both triples are internally consistent.
- CBini review fixes: `dbt build` PASS; the new error-severity tests
  (staff-region accepted_values, goals-config uniqueness) pass; the new rpt
  uniqueness test warns as designed.
- Region/org as-of roll-up fix: deferred to #4514 (see Review response).

### Intentional changes to call out

- **Staff region labels**: long ADP business-unit names normalize to city
  names + `TAF`; `KIPP TEAM and Family Schools Inc.` and `KIPP Paterson`
  previously leaked un-decoded into the extract. Staff region-level
  `aggregation_hash` / `aggregation_display` values shift accordingly.
- **`is_most_recent_complete_week` semantics**: previously date-anchored
  (near-empty in summer); now per-scope "latest complete week of the current
  academic year" (~1,372 rows). The workbook's filters should be revisited when
  grain filters are added.
- **View to table**: dbt's table materialization drops the old view
  automatically — no manual step needed; watch the first post-merge Dagster
  materialization land.

### Known pre-existing issues

The aggregation PK-uniqueness test warns (**130 rows post-Miami-removal**,
previously ~196, down from ~308 after
the must-fix #1 collapse removed ~112 collision rows). The residual are
pre-existing mechanisms (staff-retention `schoolid = 0` / null-`org_level` rows,
Camden formative-assessment rows, `is_current_week` calendar-edge fan-out). Also
pre-existing: comm-log dedup tie-break non-determinism; null-region quarter rows
from unmapped reporting-terms school ids; missing uniqueness test on
`int_topline__iready_lessons_weekly`. A follow-up issue will cover these.

The region/org year-end as-of undercount is tracked separately in #4514
(deferred — see Review response).

### Deferred (Task 13, separate PR)

Sheet cutover: create the period-goals tab + `period_rollup` column
(seed CSVs are the authoring source), swap two staging `from` clauses, delete
seeds. Notes for that PR: CI flips from self-contained (seeds) to
external-dependent (`stage_external_sources --target staging` required);
re-run the goal-join coverage isolation against real sheet data; guard
`period_rollup` coverage of the 7 period-scoped indicators (a blank there
would silently double-represent an indicator at non-week grains, so add a
both-ways guard: every aggregation indicator has a rollup row, plus
accepted_values [as_of, period] on the column).

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Address
      why. (Claude is advisory — use your judgement, but don't ignore it.)

### dbt

      existing `topline_cascade_dashboard` exposure unchanged; extract keeps
      its name and adds columns only
- [x] **Breaking change?** No columns renamed or removed on contracted
      models; the extract adds columns only (period fields + goal-difference
      fields). Staff region VALUES change as documented above. Rollback plan:
      revert the merge commit — no data migration involved.
- [x] External sources: none added or modified — the new staging model reads
      a dbt seed, so CI is self-contained (no `stage_external_sources`
      needed for this PR)

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes


anthonygwalters:
Review response — both actionable findings verified and addressed in 52be1d151.

## High — as-of binding to future spine weeks: confirmed, fixed

Verified before fixing:

- `int_topline__microgoals_assigned_weekly` is itself calendar-spine-based
  (teachers x all AY calendar weeks), so staff weekly rows carry zero-filled
  values on not-yet-elapsed weeks — Newark AY2025 tail weeks read
  `avg = 0.0` across 470 staff.
- Not staff-only: the student weekly spine already holds **40.1M
  future-term rows** (AY2026 out to 2027-06-14), so student as-of series
  (Total Enrollment, i-Ready Diagnostic, GPA, etc.) were equally exposed.
- Scope is broader than flagged: ytd `period_end` is the last day of the
  school year, so the unbounded pick was future-bound for the **entire**
  in-progress year, not just the current month/quarter.

Fix is the suggested bound in `as_of_candidates`:

```sql
and f.term <= current_date('{{ var("local_timezone") }}')
```

Verification (data): A/B same-day dev rebuild with/without the bound on
identical inputs — row counts identical at every grain; every as-of group
byte-identical except two third-decimal flips in Weighted Y1 GPA month
values (Newark Sep/Oct). Those flips also occur between two same-code
builds — `round(avg(float64), 3)` shard-order noise at a rounding boundary,
not the fix (GPA AY2025 has zero future-dated weeks, so its pick cannot
change). The fix is a visible no-op **today** only because AY2025 is fully
elapsed and AY2026 as-of rows relabel to future period starts and drop at
the final date filter; from August it is load-bearing.

## Medium — period-rollup coverage guard: pulled forward from Task 13

New singular test `int_topline__student_metrics_periods__rollup_config_is_period`
(severity error) fails when any (layer, indicator) emitting recomputed
period rows lacks `period_rollup = 'period'` in the config — converting the
silent double-count into a CI failure. Passes on the current seed (7/7
indicators covered). It refs the seed with the same `TODO(#4363)` swap
marker as the model, so the sheet cutover repoints both together.

## Low items

- **Quarter periods keyed by reporting-terms school ids** — already
  captured in #4371 (null-region quarter rows from unmapped reporting-terms
  school ids).
- **`region_to_city` / `dim_regions.name` coupling** — agreed it is worth
  enforcing mechanically; added to #4371 rather than this branch.
- **Paterson rows in `int_topline__periods`** — intentional; no metric
  joins to them.

anthonygwalters:
@claude review and note responses and commits since previous review

anthonygwalters:
@claude why is the dbt run failing?

cbini (changes requested):
**Requesting changes** — but up front: this is strong work. The multi-grain design is sound, the goal-resolution logic is provably safe against fan-out, and it improves test and convention coverage over what was there before. Most items below are small; two are worth fixing before merge.

## Must fix

### 1. Org-level staff totals come out wrong
`int_topline__dashboard_aggregations.sql`, staff org branch (~lines 450-513)

The staff org branch stamps every row `region = 'All'` but still groups by `m.region`. So the org total is really one partial row per region — each holding only that region's staff — all labeled `All`. They collide on the key, and none of them is the true network figure. The student org branch just above already does this right (no region in its group by).

Fix: drop `m.region` from the group by (~line 496). One line.

### 2. Year-end as-of totals silently drop early-ending schools
`int_topline__periods.sql`, region/org/TAF scope (~lines 118-153), feeding the as-of pick in `int_topline__dashboard_aggregations.sql` (~lines 708-773)

At region and org scope, the period window ends on the latest school's last day across the whole network. The as-of pick then takes the last in-session week inside that window. Schools that finished earlier (all of Miami, early June) have no rows that week, so they drop out of the org/region year-end totals — org YTD enrollment would undercount by about 1,346.

This only affects org/region as-of indicators at year-end (YTD, Q4, June), and it is quiet right now because we are in the summer gap. School-level numbers are correct. It was not covered by the school-scope reconciliation in the PR description.

Fix: resolve the as-of pick per school (each school's own last in-session week), then roll up to region and org.

## Should fix

### 3. region_to_city can pass unmapped values through
`macros/region_to_city.sql`

It maps five long names and returns anything else unchanged. Upstream data already holds short codes (MIA, TEAM, KCNA, KTAF, KNJ) that only get filtered out by luck today. If one reaches the macro it becomes a bad region on the staff side, or a silently dropped row in the seats-to-goals join.

Fix: map the short codes too, or add an accepted_values / relationships test on the output so a leak fails loudly.

### 4. Add the two missing uniqueness tests while these files are open
- `stg_google_sheets__topline_aggregate_goals` — no uniqueness test at all (staging models require one, severity error). Its hash is a join key.
- `rpt_tableau__topline_cascade_dashboard` — no uniqueness test, and this PR changes its grain. Upstream coverage makes it low-risk, so either add one or note the deliberate deferral.

## Seeds and period_rollup — for the deferred sheet cutover

The seeds are fine as a temporary bridge. Keeping the mock goals in the repo instead of the live goals sheet is the right call — the sheet is read live, so mock data there would show up on the real dashboard.

`period_rollup` should live in the goals sheet as config, as planned. Two things to add at cutover so a sheet edit cannot quietly break the dashboard:
1. Guard it both ways. Today only the period-side test exists. Add a test that every indicator in the aggregation has a rollup row, plus accepted_values [as_of, period] on the column. Otherwise a blank or mistyped cell makes an indicator disappear at month, quarter, and YTD.
2. Note on the sheet that flipping an indicator to `period` also needs an upstream recompute model. You can move an indicator to `as_of` with a sheet edit alone, but not to `period`.

## Smaller items (not blocking)
- Period goal only applies when its grade band exactly matches the base goal's band; a mismatch silently falls back to the base goal. Consider an orphan test.
- `goal_difference_percent` divides directly while `progress_to_goal_pct` uses safe_divide — a zero goal would error the build.
- i-Ready numerator and denominator are null in the period union (inconsistent with the ADA and contacts branches).
- The seed period_rollup guard tests are warn-level; set severity error if you want them to block CI.

## Scope note
This PR bundles two features — the multigrain rebuild and the separate goal-difference work (each has its own spec and plan). Worth naming both in the title and description so the squash-merge history reflects it.

## Nice work, specifically
- The 4-branch goal resolution cannot fan out — each branch is backed by the uniqueness test on the period-goals key.
- The as-of vs period split is genuinely guarded against double-counting.
- The aggregation model gained a uniqueness test, table materialization in the yml, and explicit column lists it did not have before.

anthonygwalters:
Review dispositions (commit `a37a55be9`):

**Must fix**

- **#1 staff org totals** — Fixed. Dropped `region` from the staff org-level group-by, so the org staff total is now the true network figure instead of partial per-region rows labeled `All`. (Bonus: this removed ~112 collision rows from the PK-uniqueness warn, 308 → 196.)
- **#2 year-end as-of undercount** — **Requesting your OK to defer to #4514.** I implemented the roll-up approach and it does not hold up. The aggregation's grade-band variants (`0-2` / `0-4` / `5-8` / `0-12` / ...) do not nest across grains — `org 0-12` and `region 0-8` have no school-level counterpart, and school carries `0-5` / `(null)` bands the org grain does not — so org/region per-band values cannot be reconstructed from school-grain components. The correct fix is to resolve each school's own as-of week at the raw-student level and re-aggregate per band, which is a materially larger refactor (and further complicated by the two pre-aggregated upstream indicators). It is dormant during the summer gap — school-level and weekly numbers are correct — and reappears at the next divergent year-end close. Full analysis and the Approach-B direction are in #4514. OK to land the rest of this PR with #2 tracked there?

**Should fix**

- **#3 `region_to_city` leak** — Fixed. Added a fail-loud `accepted_values` test (`[Newark, Camden, Miami, Paterson, TAF]`, severity error) on the `int_topline__staff_metrics.region` output.
- **#4 uniqueness tests** — Fixed. `dbt_utils.unique_combination_of_columns` on `stg_google_sheets__topline_aggregate_goals` (severity error) and on `rpt_tableau__topline_cascade_dashboard` (severity warn, mirroring the upstream PK grain — warn because the pre-existing duplicate-key rows already reach the extract).

**Smaller**

- `goal_difference_percent` zero-goal — Fixed (wrapped in `safe_divide`, matching `progress_to_goal_pct`).
- Period-goal grade-band orphan test, i-Ready null numerator/denominator in the period union, and the seed `period_rollup` guards — acknowledged. The `period_rollup` two-way guard is folded into the sheet-cutover follow-up (plan Task 13); I left the orphan-test and i-Ready items noted-not-fixed in this pass.

The PR description is updated to name both features and the #2 deferral. dbt Cloud CI is running on the fixes.

cbini:
Thanks for the thorough response — I checked each fix against the actual diff rather than taking the summary at face value. Also merged main into the branch to clear the conflict (pushed as 9e065f75a; two trivial non-overlapping conflicts, both resolved by keeping both sides' additions).

**Confirmed fixed, matches what was asked:**
- Must-fix #1 (staff org totals) — `region` is gone from the group by. Correct.
- Should-fix #4 (uniqueness tests) — both added: error severity on the goals staging model, warn on the extract with a documented reason. Good call on the warn there.
- `goal_difference_percent` zero-goal guard — now wrapped in `safe_divide` on both branches.

**One gap in should-fix #3:** the `accepted_values` guard landed on `int_topline__staff_metrics.region`, but the same macro also feeds `int_topline__seats_staffed_weekly_aggregations.entity` — the other side of the seats-to-goals join I flagged. That model's yml is still bare, so an unmapped value there still silently drops a row in that join instead of failing loudly. Same fix, same values list — worth adding before merge since it's a one-line addition while the file's fresh in mind.

**Must-fix #2 (year-end as-of undercount) — deferral approved.** Read through #4514 including your follow-up comment. The Approach A postmortem is convincing — you actually built the rollup, hit a real wall (the grade bands genuinely don't nest: region carries `0-8` that org doesn't, school carries `0-5`/null that neither does), and caught it before shipping something wrong. Approach B's plan (resolve as-of at the raw-student level, then run the existing per-band aggregation on the filtered rows) looks like the right direction, and the reconciliation criteria you've written down are concrete enough to hold the re-review to. Fine to land this PR with #2 tracked there.

Once the seats_staffed guard is in, this looks good from my side.

anthonygwalters:
Heads up on a scope addition, pushed as `a305ed037`. This isn't a response to your review — none of your findings are affected, and #2 stays deferred to #4514 as you approved.

The dashboard owner asked for Miami to come out of this dataset entirely until Focus integration lands. While scoping that I found Miami isn't just rendering blank tiles. `int_topline__student_retention_weekly_aggregations` reads `int_students__retention_over_time`, which builds its own timeline to `current_date` and never joins `int_powerschool__calendar_week` — so while every other topline path is starved by the frozen pre-Focus archive, that one kept publishing. Prod currently holds 72 Miami AY2026 rows through 2026-08-17, computed off pre-migration exit dates, dragging network retention roughly 11 points below Newark and Camden. A plausible-looking wrong number rather than a blank tile, and it gains a row every Monday.

The dam is five predicates, one per union input, following the pattern already in `int_tableau__fresh_enrollment_scaffold`. Two tidier placements don't work: filtering `int_topline__periods` leaves every Miami week row alive, because `flags` joins periods LEFT and only `as_of_candidates` joins it INNER; and filtering the extract leaves Miami inside org totals, because org rows are a separate aggregation pass that omits region from the group by, and `seats_staffed` and `student_retention` fan out to org inside themselves. All five predicates carry a `TODO(#4943)` marker so `git grep` finds the whole dam at reversal.

**One thing lands directly on your review.** You checked the week-grain parity claim against the actual diff rather than taking the summary at face value. That claim is now false for `org` and `region` rows — prod's network totals include Miami and ours no longer will. AY2025 org Total Enrollment at week grain goes from 10,119 to 8,773. I've amended the PR body rather than leave a claim you verified standing as written. School-grain Newark and Camden rows are unaffected.

Paterson is deliberately unchanged. Three of the five predicates are the lines that already excluded Paterson, so they became `not in ('Paterson', 'Miami')` rather than a swap. The owner does want Paterson in, but it has zero rows in the topline goals sheet, and `metric_aggregate_value` is a `case` on the goals-sourced `aggregation_type` — so including it today would null every Paterson value rather than show metrics without goals. It's a one-token follow-up once Ops populates the sheet, which is the long-lead item.

Two things worth knowing that aren't mine to decide:

- `FAST` and `Star Diagnostic` are Miami-only, so they disappear from every scope including org. If anyone reads those tiles they should hear it from us first.
- `int_focus__calendar_week`, `int_focus__ada`, and `int_focus__attendance_daily` all landed on main while this was in flight. Re-inclusion may be closer to a repointing job than the four-to-six-month estimate implies, at least for attendance and enrollment.

Full context and the remaining follow-ups — Paterson i-Ready ingestion, the Paterson intervention allow-list, and the other extracts that will blank Miami the same way — are in #4943.

PR #4368, opened 2026-07-10
feat(kipptaf): consume finalsite 1+4 contact surface in extracts, roster, and clever

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request swaps the network student-contact surface from PowerSchool to the 1+4 Finalsite shape at the kipptaf consumption layer:

- **New `int_students__contacts`** — long-format union: NJ Finalsite (Newark, from `int_finalsite__student_contacts` + the contact-id crosswalk) plus a PS-mapped branch for Miami and not-yet-cutover NJ regions. Grain `(student_number, _dbt_source_project, contact_slot)`.
- **New `int_students__contacts_pivot`** — wide one-row-per-student surface reproducing the legacy contact columns (`contact_1`, `contact_2`, `emergency_1..3`, `pickup_1..3`) plus the additive `emergency_4` set. Under 1+4, `contact_2_*`/`pickup_*` emit NULL.
- **Rewired `int_extracts__student_enrollments`** to source contacts from the pivot: the 72 legacy PS contact columns are dropped from `base_powerschool`'s `e.*` via `except()` and re-provided by a LEFT join on the pivot; `emergency_4` threads through `subjects_weeks`.
- **Remapped `rpt_clever__students`** onto the new long model: `contact_1` emits `contact_type='primary'` (the single reportable parent), `emergency_1..4` emit `'emergency'`, and phones are unpivoted to Home/Cell/Work.
- **Verify-only (no edits):** `int_kippadb__roster` and the KIPP Forward / DeansList / student-contact-info gsheet feeds auto-inherit the preserved column surface.

The contact-column surface (names + types) is preserved everywhere an external consumer reads it — retired `contact_2_*`/`pickup_*` simply go NULL — so this is a source swap, not a surface-breaking change. Values move to the strict 1+4 shape **network-wide**: Newark's `contact_1` now comes from Finalsite, and `contact_2_*`/`pickup_*` go NULL for all four regions — including Camden/Miami/Paterson (still on PowerSchool), where ~5,000 students had a 2nd parent and ~840 a pickup contact that now leave the surface (intended per the 1+4 mandate; downstream feed owners to be notified). `emergency_4` is additive.

Depends on #4361 and #4366 (finalsite package foundation), both merged and materialized in prod. Follow-ups (separate PRs): redefine `dim_student_contact_persons` / `bridge_student_contacts`, and remove the now-dead contact columns from the powerschool package.

Refs #4346

## AI Assistance

[...]

PR #4375, opened 2026-07-10
feat(dbt): source dim/bridge student contacts from int_students__contacts

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request redefines `dim_student_contact_persons` and `bridge_student_contacts` to source from the network-wide 1+4 contact surface (`int_students__contacts`, added by #4368) instead of the PowerSchool-only contact models. This is the dim/bridge follow-up (PR A2) to the merged Finalsite NJ student-contacts consumption PR.

- `dim_student_contact_persons` — person grain. `contact_1` keys on the contact's real identity (`_dbt_source_project` plus `finalsite_contact_id` for NJ, `personid` for PowerSchool), collapsed across siblings; emergency contacts have no stable person record under the 1+4 shape, so each keys on (`_dbt_source_project`, `student_number`, `contact_slot`). Source-agnostic typed-phone columns (`phone_mobile`/`phone_home`/`phone_daytime`/`phone_work`/`phone_primary`) replace the single prioritized `phone`.
- `bridge_student_contacts` — grain (`student_key`, `student_contact_person_key`, `contact_slot`); carries `relationship` plus `is_emergency`/`is_pickup`/`is_custodial`/`is_household_member`. The person key is derived identically to the dim.

Coupling verified locally (dev, `--defer`): `bridge_survey_expectations` (which cross-joins every `dim_student_contact_persons` row for family survey expectations) rebuilds against the new dim with all FK and uniqueness tests passing; `bridge_student_contacts` FK relationships and composite uniqueness pass with 0 null FKs.

Data change to note: the new dim is the focused 1+4 surface (45,859 persons) rather than the old all-priorities PowerSchool surface (142,549), so family survey expectations drop from 427,647 to 137,577. This is expected, not a break — but it moves the family-survey response-rate denominator, so the survey/topline owner should be aware. Whether family expectations should be scoped further (contact_1 only) is a separate semantic decision left as a possible follow-up.

`fct_survey_submissions` is unaffected (its `student_contact_person_key` is always null; relationships test passes trivially). No Cube model reads these marts.

## AI Assistance

Authored by Claude Code (Opus 4.8) under human direction: model design, surrogate-key composition, and coupling verification were AI-assisted; the 1+4 design decisions and scope come from the phase-3b plan and #4346.

### dbt

- [x] Properties file included for both redefined models
- [x] Breaking-change review: the dim `phone` column is replaced by typed phones and the surrogate-key hash composition changes. The only consumers are `bridge_survey_expectations` (reads the dim PK only) and `fct_survey_submissions` (null FK) — neither reads the removed columns, and no Cube model or exposure reads these marts by column.

## CI checks

- [ ] Trunk
- [ ] dbt Cloud
- [ ] Dagster Cloud — not triggered (no Dagster paths)

Refs #4346

### DIFF
--- a/src/dbt/kipptaf/models/marts/bridges/bridge_student_contacts.sql
+++ b/src/dbt/kipptaf/models/marts/bridges/bridge_student_contacts.sql
@@ -1,26 +1,56 @@
 with
-    students_union as (
-        select _dbt_source_relation, dcid, student_number,
-        from {{ ref("stg_powerschool__students") }}
+    contacts as (
+        select
+            student_number,
+            contact_slot,
+            relationship,
+            is_emergency,
+            is_pickup,
+            is_custodial,
+            is_household_member,
+            _dbt_source_project,
+
+            coalesce(finalsite_contact_id, personid) as person_identity,
+        from {{ ref("int_students__contacts") }}
+    ),
+
+    keyed as (
+        select
+            student_number,
+            contact_slot,
+            relationship,
+            is_emergency,
+            is_pickup,
+            is_custodial,
+            is_household_member,
+
+            {{ dbt_utils.generate_surrogate_key(["student_number"]) }} as student_key,
+
+            -- keyed identically to dim_student_contact_persons: contact_1 by the
+            -- person's real identity, emergency by student + contact slot
+            if(
+                contact_slot = 'contact_1',
+                {{
+                    dbt_utils.generate_surrogate_key(
+                        ["_dbt_source_project", "person_identity"]
+                    )
+                }},
+                {{
+                    dbt_utils.generate_surrogate_key(
+                        ["_dbt_source_project", "student_number", "contact_slot"]
+                    )
+                }}
+            ) as student_contact_person_key,
+        from contacts
     )
 
 select
-    {{ dbt_utils.generate_surrogate_key(["s.student_number"]) }} as student_key,
-
-    {{ dbt_utils.generate_surrogate_key(["c._dbt_source_project", "c.personid"]) }}
-    as student_contact_person_key,
-
-    c.relationship_type,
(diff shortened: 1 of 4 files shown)


cbini:
## Re: Claude code review

Processed via the receiving-code-review flow. Both MEDIUM findings reduce to "confirm the dim's own `unique` + `not_null` tests pass" — I ran that verification locally against prod `int_students__contacts`; both risks are closed, no code change needed. The LOW finding is a valid but pre-existing observation, left out of scope.

### MEDIUM — `select distinct` grain-projection assumption — verified, closed

Confirmed the `contact_1` attributes are functionally determined by `(_dbt_source_project, person_identity)`: across all 15,026 distinct `contact_1` person keys, the count of distinct attribute tuples per key has `max = 1` (0 keys with varying attributes). The dim built in dev with `unique` and `not_null` on `student_contact_person_key` both PASS.

### MEDIUM — `contact_1` key collision on NULL `person_identity` — verified, closed

`countif(contact_slot = 'contact_1' and finalsite_contact_id is null and personid is null)` = **0** in prod `int_students__contacts`. No placeholder-hash collision today, and the dim `unique` test guards it going forward. Not adding the emergency-style fallback now — it would churn every person-key hash (and require mirroring in the bridge) for a case that does not exist; the `unique` test is the tripwire if it ever does.

### LOW — stale TODO / redundant `distinct` in `bridge_survey_expectations`

Valid observation, but pre-existing rather than newly enabled here: the previous `dim_student_contact_persons` already carried `unique` + `not_null` on its PK, so the `select distinct` and its TODO were already redundant before this PR. Leaving it out of scope so this PR stays a clean dim/bridge source-swap; it can ride a future touch of `bridge_survey_expectations`.

PR #4377, opened 2026-07-10
refactor(dbt): drop dead contact columns from base_powerschool enrollments

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request removes the dead contact/emergency/pickup columns from the powerschool package model `base_powerschool__student_enrollments`. Under the Finalsite 1+4 contact surface, that surface moved to `int_students__contacts_pivot` (consumed by `int_extracts__student_enrollments`) in #4368, so the powerschool package no longer needs to carry it. This is the powerschool dead-column removal (PR B) follow-up to the merged Finalsite NJ student-contacts consumption PR.

Coupled, single-PR cross-project change:

- powerschool package `base_powerschool__student_enrollments.sql`: drop the 72 `scw.*` contact columns and the `int_powerschool__student_contacts_pivot` join; drop the same 72 columns from its contract.
- kipptaf `int_extracts__student_enrollments.sql`: drop the same 72 names from the `e.* except(...)` list. They no longer exist in `base_powerschool`, and `except()` naming a missing column errors, so this edit must ship together with the package change.
- kipptaf union wrapper `base_powerschool__student_enrollments`: drop the 72 columns from its contract, plus a doc comment to force `state:modified` so CI rebuilds it against the narrowed district schema.

`int_powerschool__student_contacts` and `int_powerschool__student_contacts_pivot` are intentionally NOT deleted — raw PowerSchool still loads, and Miami/Focus read them.

## Why single-PR

The coupling is bidirectional: the package provides the columns; kipptaf both `except()`s them and declares them in the wrapper contract. A district-first PR would break kipptaf prod on the next rebuild (`except()` naming missing columns); a kipptaf-first PR would break the wrapper contract (extra columns). So both sides ship together, and the 4 districts' `zz_stg` `base_powerschool` were reseeded with the narrowed schema so kipptaf CI reads it (verified: 0 contact columns in the seeded staging tables).

## AI Assistance

Authored by Claude Code (Opus 4.8) under human direction: the edits, cross-project sequencing analysis, and local verification were AI-assisted; scope and the 1+4 migration come from the phase-3b plan and #4346.

### dbt

- [x] Breaking change reviewed: this removes contract columns from `base_powerschool__student_enrollments` (package and kipptaf wrapper). A full-column grep across all `base_powerschool__student_enrollments` consumers found references only in `int_extracts` (fixed here); the 4 district-level `int_powerschool` children and every other kipptaf consumer read zero of the dropped columns. Built `base_powerschool` plus children in all 4 districts and the kipptaf wrapper plus `int_extracts` locally — all pass.
- [x] Cross-project staging: reseeded `zz_stg` per district so CI reads the narrowed schema.

## CI checks

- [ ] Trunk
- [ ] dbt Cloud
- [ ] Dagster Cloud — not triggered (dbt-only change; `pull_request` paths exclude `src/dbt/`)

Refs #4346


cbini:
## Re: Claude code review

LGTM confirmed. The one confirm-before-merge (all four district `zz_stg` tables reseeded to the narrowed schema) is verified.

Ran the suggested query. The four district schemas (`zz_stg_kippnewark/kippcamden/kippmiami/kipppaterson_powerschool`) all return 0 contact columns (absent from the result set). The only nonzero row is `zz_stg_kipptaf_powerschool` at 72 — that is the kipptaf-level union **wrapper's own** staging copy, not a district source the wrapper reads. CI rebuilds the wrapper (forced `state:modified` via the doc comment) rather than reading that stale copy, so it does not affect the contract check — confirmed by the green dbt Cloud run on this PR.

The five static verifications match my local checks (no consumer reads a dropped column off a `base_powerschool` alias; the `except()` edit is required and correct; contracts drop exactly 72 columns each; the dropped `scw` join was 1:1 on a `severity: error` unique key so no grain change). No action.

Issue #3915, opened 2026-05-13
chore(powerschool): clean up double-written cc records (24 historical groups)


cbini:
Follow-up from #3900.

## Background

PowerSchool wrote two `cc` (course-enrollment) records for the same `(studentid, sectionid, dateleft)` in a small number of historical cases. This causes downstream fan-out in marts that join on date ranges (see `#3900` for the surfacing warn test and mart-side mitigations).

The corpus is **frozen** — no incidents in AY 2024 or later. Concentrated in:

- **Newark**: 19 dup groups (10 from AY 2020 — COVID-era enrollment churn likely)
- **Miami**: 3 dup groups (all AY 2023)
- **Camden**: 1 dup group (AY 2023)
- **Paterson**: 0

Total: **24 groups / 48 offending rows**.

## Identify the records

Run against the data warehouse:

```sql
SELECT
  _dbt_source_relation,
  studentid
  academic_year,
  schoolid,
  sectionid,
  dateleft,
  count(*) AS n_records
FROM `teamster-332318.kipptaf_powerschool.stg_powerschool__cc`
GROUP BY all
HAVING count(*) > 1
ORDER BY _dbt_source_relation, academic_year desc, schoolid
```

This returns the 24 groups with section/school/year/date context but no student identifiers. To locate the underlying records in PowerSchool, extend the query in a non-shared session by adding `studentid` and `cc.dcid` to the projection — keep those results local; do not paste them into issues, PRs, or chat.

## What to fix

For each `(studentid, sectionid, dateleft)` group with `count(*) > 1`:

1. Determine which `cc.dcid` is the correct record (typically the latest written — highest `dcid` — but verify against the audit trail).
2. Delete or supersede the redundant record(s) in PowerSchool.

## Acceptance

- The warn test `dbt_utils.unique_combination_of_columns` on `stg_powerschool__cc(studentid, sectionid, dateleft)` reports 0 failing groups.
- The PK uniqueness test on `fct_grades_assignments` passes at `severity: error` (currently failing pending this cleanup — see #3900).

## Why not fix in code

We considered upstream dedup in `base_powerschool__course_enrollments` or mart-level `dbt_utils.deduplicate()`. Both rejected: they would silently rewrite source data and hide the anomaly from monitoring. Source-of-truth fixes belong in PowerSchool.

Refs #3900.

cbini:
Tracked in Asana: https://app.asana.com/1/913513768672/project/1205971774138578/task/1215030421425063

PR #4394, opened 2026-07-14
fix: surface AP exam scores with no matching course enrollment

### LINKED ISSUE #4391: fix: rpt_tableau__ap_assessment_dashboard drops AP scores with no matching course enrollment
## Problem

`rpt_tableau__ap_assessment_dashboard` (`src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__ap_assessment_dashboard.sql`) silently drops a student's AP exam score whenever they don't currently have a matching AP-tagged course enrollment — instead of showing something like "Took AP exam, not enrolled in course."

The join chain is the cause:

```sql
from {{ ref("int_extracts__student_enrollments") }} as e
left join
    {{ ref("base_powerschool__course_enrollments") }} as s
    on e.studentid = s.cc_studentid
    and e.academic_year = s.cc_academic_year
    and {{ union_dataset_join_clause(left_alias="e", right_alias="s") }}
    and s.rn_course_number_year = 1
    and s.ap_course_subject is not null
    and not s.is_dropped_section
left join
    {{ ref("stg_google_sheets__collegeboard__ap_course_crosswalk") }} as x
    on s.ap_course_subject = x.ps_ap_course_subject_code
    and x.data_source = 'CB File'
left join
    {{ ref("int_assessments__ap_assessments") }} as a
    on e.academic_year = a.academic_year
    and e.student_number = a.powerschool_student_number
    and s.ap_course_subject = a.ps_ap_course_subject_code
    and a.test_subject != 'Calculus BC: AB Subscore'
```

The `a` (assessment score) join predicate references `s.ap_course_subject`. Since `s` only joins when the student has a matching AP-tagged course enrollment (`s.ap_course_subject is not null`), a student with **no** such enrollment gets `s` = NULL, which means the `a` join can never fire either — even though the student genuinely has a score in `int_assessments__ap_assessments` for that subject. The exam score isn't mislabeled; it's completely invisible on the dashboard.

This is a legitimate, expected scenario — College Board doesn't require course enrollment to sit an exam, and not every AP class enrollee tests. The dashboard's own `case` logic (`courses_course_name is null` → `'Not applicable'`, `courses_course_name is not null and test_name is null` → `'Took course, but not AP exam.'`) has no branch for "took the exam, no matching course" because the join structure makes that combination unreachable.

## Evidence

Found while auditing the 2025-2026 AP data pipeline (see #4390): comparing `int_collegeboard__ap_unpivot` against the dashboard for `academic_year = 2025` found 15 exam-score rows present in the former but absent from the latter. 12 traced to a separate PowerSchool course-tagging gap (tracked in #4390). The remaining 3 are this bug — three distinct students (identities kept local per repo PII policy, not included here) each confirmed to have a real AP exam score with either zero AP course enrollments that year, or AP course enrollments in different subjects than the exam they took.

## Suggested direction

Join `int_assessments__ap_assessments` (or the score source) to `e` independently of whether `s` matched — e.g. on `e.academic_year = a.academic_year and e.student_number = a.powerschool_student_number` alone, then bring in `s`/`x` as separate left joins keyed the same way, rather than chaining the assessment join through the course-enrollment join. That lets the `case` logic add a real "Took AP exam, not enrolled in course" state instead of the row disappearing.

## Scope

Distinct from the `int_collegeboard__ap_unpivot` dedup issue also found during the #4390 audit (that one collapses multiple different unresolved-crosswalk students into a single row via `dbt_utils.deduplicate` on a shared NULL `powerschool_student_number` — a separate bug, not tracked here yet).

GabyRangelB:
# Pull Request

## Summary & Motivation

make AP exam scores visible on the AP assessment dashboard for students who took the exam but have no matching AP course enrollment in PowerSchool -- today those rows are silently dropped.

Restructures `rpt_tableau__ap_assessment_dashboard`'s joins around a `subjects` bridge CTE that unions AP subject codes from course enrollments and exam scores independently, instead of chaining the exam join through the course-enrollment join. Adds a `subject_code` grain column, a new `test_subject_area` state (`Took AP exam, not enrolled in course.`), and an `ap_course_name` fallback to the exam's own crosswalk-resolved name.

Verified against current prod data: 245 previously-hidden exam-only rows surface (spanning many historical years, not just the 3 cases cited in the issue), 119 generic "Not applicable" placeholder rows are replaced by real subject rows for students who turn out to have exam-only scores, and a reconciliation check confirms 0 true regressions.

Closes #4391

## AI Assistance

> If Claude Code authored or co-authored this PR, briefly note what was AI-assisted vs. human-directed in the summary above

Claude Code designed and implemented this fix end-to-end: brainstormed the design (spec at `docs/superpowers/specs/2026-07-14-ap-dashboard-exam-only-scores-design.md`), wrote an implementation plan (`docs/superpowers/plans/2026-07-14-ap-dashboard-exam-only-scores.md`), and executed it via subagent-driven development with a task review after each task and a final whole-branch review. All SQL/YAML changes, the dry-run validation against prod data, and the test fixtures were AI-authored; a human directed the design decisions through the brainstorming Q&A captured in the spec and chose the merge path.

## Self-review

### General

### dbt

- [x] Properties file included — `rpt_tableau__ap_assessment_dashboard.yml` updated with the new `subject_code` column, the uniqueness test moved to it, and updated descriptions for `ap_course_name`/`test_subject_area`.
- [x] Exposure — no change needed; the existing Tableau exposure in `models/exposures/tableau.yml` is unaffected (no column renames/removals, only additive changes).
- [x] Breaking change? No — purely additive (new column, new dashboard state, no renames or removals). No downstream dbt model `ref()`s this model (verified via `grep`); only the Tableau exposure references it.
- [x] No new external source — n/a

## CI checks

- [ ] **Trunk** — pending
- [ ] **dbt Cloud** — pending
- [ ] **Dagster Cloud** — not triggered (dbt-only change, no schedule/sensor/integration changes)

---
  - https://app.asana.com/0/0/1216565418570028

### DIFF
--- a/docs/superpowers/plans/2026-07-14-ap-dashboard-exam-only-scores.md
+++ b/docs/superpowers/plans/2026-07-14-ap-dashboard-exam-only-scores.md
@@ -1,0 +1,866 @@
+# AP Dashboard Exam-Only Scores Fix Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use
+> superpowers:subagent-driven-development (recommended) or
+> superpowers:executing-plans to implement this plan task-by-task. Steps use
+> checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Fix `rpt_tableau__ap_assessment_dashboard` so a student's AP exam
+score is no longer silently dropped when they have no matching AP course
+enrollment (issue [#4391](https://github.com/TEAMSchools/teamster/issues/4391)).
+
+**Architecture:** Restructure the model's join chain around a new `subjects`
+bridge CTE that unions AP subject codes from course enrollments and exam scores
+independently, so the exam-score join no longer chains through the
+course-enrollment join. Add a `subject_code` grain column, a third
+`test_subject_area` state (`'Took AP exam, not enrolled in course.'`), and an
+`ap_course_name` fallback to the exam's own crosswalk-resolved name.
+
+**Tech Stack:** dbt (BigQuery), `kipptaf` project, dbt unit tests (dict-format
+fixtures), BigQuery MCP for manual verification queries.
+
+## Global Constraints
+
+- Full design and validated queries:
+  `docs/superpowers/specs/2026-07-14-ap-dashboard-exam-only-scores-design.md`.
+- Never write student names, DOB, or other PII to a committed file, PR, or issue
+  comment — chat/terminal output only (repo `CLAUDE.md` PII policy).
+- All SQL follows `.trunk/config/.sqlfluff` (BigQuery dialect, trailing commas,
+  single-quoted strings, 88-char lines). Run
+  `/workspaces/teamster/.trunk/tools/trunk check --force <file>` from inside the
+  branch checkout before pushing — the pre-commit hook only runs `fmt`, not the
+  sqlfluff/yamllint checks that fire at pre-push/CI.
+- `given`/`expect` dict scalars in dbt unit tests are normally UNQUOTED per repo
+  convention, **except** a scalar containing a literal comma (e.g.
+  `"Smith, Jane"`, `"Took course, but not AP exam."`) — inside a YAML flow
+  mapping (`{ ... }`), an unquoted comma is parsed as a field separator, not as
+  part of the string, and silently produces a bogus extra key. Quote only those
+  values.
+- `dbt` invocations: always `uv run`, never bare `python`/`dbt`. Use
+  `DBT_PROFILES_DIR=/workspaces/teamster/.dbt` and
+  `--project-dir /workspaces/teamster/src/dbt/kipptaf` on every call from the
+  main repo (or the worktree-equivalent absolute paths per repo `CLAUDE.md` if
+  executing from a worktree).
+- Only one other file in the repo references this model:
+  `src/dbt/kipptaf/models/exposures/tableau.yml` (the Tableau exposure
+  definition). No downstream dbt model `ref()`s it — verified via
+  `grep -rl "rpt_tableau__ap_assessment_dashboard" src/dbt`. No exposure edit is
+  needed; this fix is self-contained to the one model + its properties file.
+- Out of scope, do not touch: the `int_collegeboard__ap_unpivot` dedup bug
+  (collapses multiple unresolved-crosswalk students via `dbt_utils.deduplicate`
+  on a shared NULL `powerschool_student_number` — not tracked yet) and the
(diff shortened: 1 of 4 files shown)


GabyRangelB:
Thanks for the review — verified the findings against prod data. Summary of what was checked and what changed:

**🟠 Medium (exam-side grain / severity) — verified clean, no dedupe added.**

Ran your exact exam-side query:

```sql
select academic_year, powerschool_student_number, ps_ap_course_subject_code, count(*) n
from `teamster-332318`.kipptaf_assessments.int_assessments__ap_assessments
where test_subject != 'Calculus BC: AB Subscore' and ps_ap_course_subject_code is not null
group by 1, 2, 3
having count(*) > 1
```

→ **0 rows.** No exam-side fan-out on the join key in current data. The only grain violations in the built model are the **3 course-side** rows (AY2021, Calculus BC) — a student carrying both an "AP Calculus BC" section and a companion "Math IV Calculus AP BC Recitation" section under the same `ap_course_subject`. Those are pre-existing in prod under the old grain key and are the same class as the PowerSchool double-write quirk tracked in #3900 / #3915, so per repo convention we keep the test at `warn` and don't dedupe.

On the suggested `rn_highest = 1` remedy: it would actually cause data loss here. `rn_highest` is `row_number() over (partition by powerschool_student_number, ap_course_name order by exam_score desc)` — **no `academic_year` in the partition**, keyed on `ap_course_name` not the subject code. Applying it to this per-year extract would collapse a student's same-subject exams across years to a single highest-scoring row; there are 9 student/course pairs (18 rows) spanning multiple years that it would wrongly drop. Moot anyway given 0 fan-out, but flagging so it isn't reached for later.

We can't promote the test to `severity: error` while the 3 pre-existing course-side dupes exist — it would fail CI on day one — so `warn` is the deliberate, convention-aligned choice for now.

**🟡 Low (redundant `select distinct`) — fixed.** Dropped both per-branch `select distinct`s; the `union distinct` already dedupes the combined set. Behavior-neutral (verified: identical grain, same WARN 3, 5/5 unit tests pass). Pushed as a follow-up commit.

**🟢 Nits — all confirmed.** Unit tests pass locally (5/5, `dbt build --select rpt_tableau__ap_assessment_dashboard`). Agree on the `union_dataset_join_clause` omission being correct (network-unique `powerschool_student_number`), the coalesce priority, and the ST06 ordering.

GabyRangelB:
@claude - can you review our response?

anthonygwalters (approved):

PR #4396, opened 2026-07-14
fix(dagster): resolve SSH remote_port from environment lazily

cbini:
# Pull Request

## Summary & Motivation

When merged, this PR stops SSH `remote_port` config from reading the environment
at module-load time.

`get_powerschool_ssh_resource()` and `SSH_RESOURCE_LITTLESIS` coerced the port
to `int` via `EnvVar("PS_SSH_PORT").get_value()` /
`EnvVar("LITTLESIS_SFTP_PORT").get_value()`. `.get_value()` resolves **eagerly**
at construction, and because these resources are built at module load (inside
the `Definitions` resources dict / as a module-level singleton), importing a
district `definitions.py` crashed with a `CheckError` whenever the port variable
was unset — e.g. in a codespace. This is the failure the `src/teamster/CLAUDE.md`
note documented.

The fix swaps both call sites to `EnvVar.int(...)` — Dagster's lazy integer
environment-variable wrapper (an `int` subclass that satisfies the `int | None`
field and resolves at resource init, exactly like the sibling string fields).
The now-unused `dagster_shared.check` imports are removed.

### Verification

- New regression test
  `tests/resources/test_resource_ssh.py::test_get_powerschool_ssh_resource_port_is_lazy`:
  red before (raised `CheckError`), green after; asserts lazy construction **and**
  `int` resolution at resource init.
- `kippnewark` / `kippcamden` / `kippmiami` `definitions` now import cleanly with
  the port variable unset (dbt manifest present); `kipptaf.resources` imports
  clean.
- Trunk clean on all changed files.

### Scope

Category A (SSH `remote_port`) only. A separate, larger eager-read pattern
remains in the Illuminate/Zendesk dlt credential specs
(`kipptaf/dlt/*/assets.py`), which build plain dlt credential objects (not
Dagster resources) and so cannot use the lazy wrapper without deferring
construction into asset setup — left as a follow-up.

## AI Assistance

AI-assisted (Claude Code): root-cause diagnosis, the `EnvVar.int` fix, the
regression test, and the `CLAUDE.md` corrections. Human-directed: the scope
decision (Category A only) and fix approach.

### Dagster

- [x] Verified modified code locations import — regression test green plus direct
      module import of `kippnewark`/`kippcamden`/`kippmiami` `definitions` and
      `kipptaf.resources`. Full `dagster definitions validate` was not run in the
      codespace (environment-limited); see Verification above.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — not applicable (no dbt model changes).

### DIFF
--- a/docs/CLAUDE.md
+++ b/docs/CLAUDE.md
@@ -67,8 +67,9 @@ not appear in site navigation.
 adding, removing, or renaming schedules or sensors.
 
 The script imports every code location's `definitions` and silently SKIPS any
-that fail to import — so running it in the codespace (district `definitions`
-fail on unset `PS_SSH_PORT`, etc.) drops those locations from the catalog.
+that fail to import — so running it in the codespace (locations fail to import
+without their dbt manifests, and `kipptaf` additionally on unset
+Illuminate/Zendesk dlt credentials) drops those locations from the catalog.
 Regenerate only in a full environment where all locations load.
 
 ## `superpowers/` Directory
--- a/src/teamster/CLAUDE.md
+++ b/src/teamster/CLAUDE.md
@@ -60,6 +60,13 @@ default "one short line max" rule.
 `check.not_none()` from `dagster_shared` — never `assert isinstance(...)`
 (`assert` is stripped by `-O`).
 
+**Int env-var config**: use `EnvVar.int("NAME")` for `int`-typed resource fields
+(e.g. ports) — it stays lazy and resolves at resource init like a string
+`EnvVar`. Never `int(EnvVar("NAME").get_value())`: `.get_value()` reads eagerly
+at construction and crashes module-load resource wiring when the var is unset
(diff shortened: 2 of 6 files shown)


cbini:
Both non-blocking notes addressed in 5a2dcb477 (fix itself unchanged):

1. **`LITTLESIS_SFTP_PORT` coverage** — added `test_ssh_resource_littlesis_port_is_lazy`, a direct guard on the singleton. It asserts the imported `SSH_RESOURCE_LITTLESIS.remote_port` is still an unresolved `EnvVar.int` marker (a revert to the eager `int(...)` would store a plain `int` and fail this `isinstance` check), then resolves it to `22` via `process_config_and_initialize()`. Both new tests pass locally.

2. **Resolution-path nit** — added a comment on the powerschool test noting that `process_config_and_initialize()` resolves config and environment variables only, not `setup_for_execution`, which is why the absent SSH secret file never blocks the test.

Issue #3807, opened 2026-05-04
feat(powerschool): replace homegrown ODBC sync with dlt


cbini:
## Context

PowerSchool SIS ingestion (`teamster/libraries/powerschool/sis/odbc/`) is a
~1300 LOC homegrown stack (`PowerSchoolODBCResource`,
`build_powerschool_table_asset()`, partition sensors/schedules) with recurring
`DPY-4024`/`DPI-1067` timeouts, a bespoke Avro type map, COUNT staleness probes,
and five per-location partition strategies.

## Decision: dlt (supersedes this issue's original Sling proposal)

Reassessed dlt vs Sling head-to-head in spike **#4398** (full results:
`docs/superpowers/specs/2026-07-14-powerschool-odbc-elt-spike-results.md`). They
tie on row and value fidelity; **dlt wins on correctness-of-defaults and fit**:

- **Timestamp correctness:** Sling silently shifted every `TIMESTAMP` +4h on our
  Oracle connection; dlt preserved the source wall-clock and matched the
  incumbent convention.
- **Robust defaults on real data:** Sling's default CSV staging failed on
  PowerSchool free-text columns (needed parquet); its recommended GCS fast-path
  is broken on our keyless Workload Identity (needed a workaround); its strict
  greater-than cursor risks boundary-tie misses. dlt's pyarrow / keyless-ADC /
  inclusive-cursor defaults handled all of it out of the box.
- **Fit + friction:** dlt needed **1** config adjustment and extends the
  existing Illuminate/Focus/Zendesk dlt pattern; Sling needed **4** and is a new
  dependency.
- Sling's genuine wins (native SSH tunnel, more idiomatic scalar types) do not
  offset these — the tunnel is moot because we already run the wrapper dlt
  reuses, and dlt's types are tunable.

## Design direction

- Per-table `@dlt_assets` (one pipeline per table), multi-asset members matching
  the dbt source layout.
- Keyless ADC to BigQuery (no GCS-staging dependency); pyarrow backend.
- SSH tunnel: reuse `SSHResource.open_ssh_tunnel()` (dlt has no native tunnel);
  separate follow-up to move it to an in-process paramiko forward.

## Ingestion strategy (revised from this issue's original incremental-merge plan)

Spike BigQuery cost analysis (BQ cost is real; Oracle reads are free): row-level
merge bills roughly target-size **every run** (its `DELETE` scans the whole
target, and clustering does **not** prune it), while `LOAD` jobs are free.
[...]

PR #4402, opened 2026-07-14
SPIKE (DO NOT MERGE): dlt vs Sling PowerSchool ODBC head-to-head

### LINKED ISSUE #4398: spike(powerschool): dlt vs Sling head-to-head for Paterson ODBC ingestion
## Context

Paterson's PowerSchool instance has migrated to the same hosts as the other regions, so ODBC (Oracle via SSH tunnel) is now available and the Couchdrop SFTP feed can be retired. #3807 proposed replacing the homegrown ODBC stack with Sling, piloting on kippnewark — but it predates the Paterson migration and its dlt-vs-Sling assessment is outdated.

Reassessment (July 2026, verified against current docs/source) found no hard blockers for either tool, but two questions that cannot be settled from documentation:

1. Oracle NUMBER (no declared scale) type fidelity into BigQuery — Sling's behavior is unverifiable from docs and has recent regression history; dlt maps to decimal with default precision unless adapted.
2. Real throughput through the password-auth SSH tunnel — no credible benchmark exists for either tool.

## Spike

Run both tools against Paterson's Oracle instance in a branch deployment, landing in scratch BigQuery datasets. Three tables chosen for risk coverage: `students` (NUMBER PKs, wide), `storedgrades` (decimal-heavy), `assignmentscore` (largest, incremental on `whenmodified`).

Per tool per table: one full load, one incremental follow-up run. Rubric: row fidelity (counts + PK anti-join), type fidelity (INFORMATION_SCHEMA diff), value fidelity (sampled decimal/date columns), wall-clock, incremental correctness across fresh pods, tunnel stability, config size and failure-mode legibility in Dagster.

## Deliverables

- Spike assets on a branch (not merged): dlt factory + Sling replication config for the same 3 tables
- Results write-up with the rubric filled in and a tool decision
- #3807 rewritten to reflect the winner, a Paterson-first pilot (retiring `sis/sftp/`), parity table scope, and direct-to-BigQuery landing

## Out of scope

- dbt source rework, Couchdrop retirement, and all non-Paterson regions (those land in the rewritten #3807)

Refs #3807

cbini:
## Summary and Motivation

> DO NOT MERGE AS-IS. This is a throwaway spike (issue #4398). The code and config commits are reverted before any merge; only the `docs/superpowers/` spec, plan, and results land on `main`.

Opened non-draft **only** to trigger the kipppaterson branch deployment (the deploy job is gated on `!draft`), so the dlt and Sling spike assets can be run against Paterson's PowerSchool Oracle instance and compared. When merged, this pull request will add nothing to `main` except the three spike docs.

The spike compares two ingestion mechanisms for replacing the homegrown PowerSchool ODBC stack, per a reassessment of #3807: dlt (extends our existing production dlt pattern; SSH tunnel via the current sshpass wrapper) vs Sling (native `ssh_tunnel`). Three tables — `students`, `storedgrades`, `assignmentscore` — each loaded full then incrementally, scored on row/type/value fidelity, throughput, incremental correctness, and operational feel.

Branch-deployment-only surfaces added (all reverted before merge): `dagster-sling` dependency, Paterson PowerSchool credential wiring in `dagster-cloud.yaml`, and the spike asset modules under `powerschool/sis/odbc_spike/`.

### AI Assistance

Human-directed: the decision to spike, tool scope, table selection, and the go/no-go on the run phase. AI-assisted (Claude Code): the research comparison, spec, plan, all spike code, and review loops.

### Dagster

- [x] Definitions import validated for kipppaterson (spike assets registered)
- [x] `tests/spike/test_odbc_spike_defs.py` passes (3/3)
- [x] Spike assets follow the asset-key convention `kipppaterson/powerschool/spike/{dlt,sling}/{table}`

## CI checks

- [ ] Trunk
- [ ] dbt Cloud (no dbt model changes; expected no-op or pass)
- [ ] Dagster Cloud branch deployment

Refs #4398
Refs #3807


cbini:
Thanks — addressing the review:

**Medium 1 (Sling `sid` vs service name):** Fixed in `dd2a6d68`. Changed `sid=` to `service_name=os.getenv("PS_DB_DATABASE")`, matching the production ODBC resource (which passes that variable as `service_name`). Confirmed against Sling's Oracle connection docs that `service_name` is a distinct accepted property. This is exactly the connect-descriptor failure the review predicted, so it's fixed before the Sling side runs.

**Medium 2 (SSH host-key algorithm asymmetry):** Documented as a run-phase watch item in the results doc. Note that Sling's native tunnel uses Go `crypto/ssh` with `InsecureIgnoreHostKey()`, so it skips host-key verification and should sidestep a legacy `ssh-rsa` host key — but if the server only offers `ssh-rsa` as a host-key algorithm and Go's client has dropped it from negotiation, the Sling tunnel could still fail at KEX where dlt (via the `sshpass` wrapper with `HostKeyAlgorithms=+ssh-rsa`) succeeds. If that happens it will be recorded as a tunnel-stack artifact, and "can't tunnel to our host out of the box" is itself a decision-relevant finding for the comparison.

**Low (reuse `get_powerschool_ssh_resource()`):** Acknowledged. The inlined `os.getenv` form was deliberate for codespace import-safety; now that `main` has moved the factory to lazy `EnvVar.int`, the factory would work too. Left as-is since this is throwaway spike code slated for revert.

**Low (dlt host = tunnel-local endpoint) and Nits:** Confirmed. The whole-branch review already verified the `PS_DB_HOST` = localhost convention (the Ops secret mirrors kippnewark, whose `server` key is the tunnel-local endpoint), and the cross-module coupling / hardcoded `teamster-test` bucket are intentional for the branch-deploy-only spike and go away with the revert.

All spike code and config are reverted before merge per the PR body — only the `docs/superpowers/` spec, plan, and results land.

PR #4410, opened 2026-07-15
fix(finalsite): serialize contacts pulls and raise max_runtime to fix 403s

### LINKED ISSUE #4408: fix: finalsite contacts nightly pulls 403 under concurrency; kippnewark exceeds default max_runtime
## Summary

All four district `finalsite/contacts` nightly pulls fail with HTTP 403. Root cause is a shared-gateway concurrency/rate block, not authentication. Separately, `kippnewark` runs longer than the default run timeout (~30 min) once it is allowed to complete.

## Evidence

- All four district schedules are `0 4 * * *` in `America/New_York`, so they fire simultaneously at 08:00 UTC.
- The 403 hits mid-pagination (the failing request carries a `cursor`), well inside the 55-minute JWT lifetime, so it is not token expiry. `kippnewark` reached page 166 (4,150 records) at ~1 req/s before an abrupt 403.
- The 403 response body is a bare nginx page (fronting gateway/WAF), not a Finalsite JSON error and not a `429` with `Retry-After`:

```html
<html>
<head><title>403 Forbidden</title></head>
<body>
<center><h1>403 Forbidden</h1></center>
</body>
</html>
```

- Natural experiment from manual retries: the concurrent retry batch failed on all four districts within ~2.5s of each other; serialized solo retries then succeeded one-by-one (paterson, then miami, then camden). `kippnewark` never completed solo because it is the largest district.
- Because the block is keyed on a shared source (egress IP / shared `fsenrollment.com` gateway), the separate subdomains and separate credentials per district do not help.

## Contributing code factors

- `FinalsiteResource._request` only backs off on `429`; a `403` is a hard failure. See `src/teamster/libraries/finalsite/api/resources.py`.
- Dagster auto-retry re-runs while sibling district runs are still active, so it 403s again.
- The finalsite contacts schedules set no `MAX_RUNTIME_SECONDS_TAG`, so runs inherit the ~30-minute deployment default.

## Planned fix

- Add a shared concurrency pool `finalsite_api` (limit 1, configured in Dagster+ Deployment then Concurrency) to the contacts asset factory so only one district pulls at a time. Mirrors the existing `overgrad_api_limit_kippnewark` and `adp_wfn_api` patterns.
- Add `MAX_RUNTIME_SECONDS_TAG` to the finalsite contacts schedule(s), sized to cover a full `kippnewark` pull plus queue wait.
- Add bounded exponential-backoff retry on `403` in `FinalsiteResource` as defense-in-depth (tenacity).

cbini:
## Summary and Motivation

When merged, this PR stops the nightly Finalsite `contacts` pulls from failing with 403 and lets `kippnewark` finish instead of hitting the default run timeout.

**Root cause.** All four district `contacts` schedules fire at the same instant (`0 4 * * *`, all Eastern) and paginate the shared `fsenrollment.com` gateway from one egress IP. The gateway returns a bare nginx `403` (not a `429`) once the concurrent load from that source trips a per-source ceiling — mid-pagination, well inside the token lifetime, so it is not an authentication or token-expiry problem. Separate subdomains and API credentials per district do not help because the throttle is keyed on the shared source. Confirmed by the retry history: the concurrent retry batch failed on all four within ~2.5s of each other, while serialized solo retries succeeded one-by-one. Separately, `kippnewark`'s strictly sequential pull (~1 req/s) runs past the ~30 min default run timeout.

**Changes.**

- Add a shared `finalsite_api` concurrency pool to the `build_finalsite_asset` factory so only one district pulls at a time (mirrors the `overgrad`/`adp` pooling pattern).
- Add bounded exponential-backoff retry on `403` in `FinalsiteResource` as defense-in-depth, with offline unit tests for the retry / no-retry / exhaustion paths.
- Add `MAX_RUNTIME_SECONDS_TAG` (7200s) to all four finalsite schedules, sized to cover a full `kippnewark` pull plus queue wait behind the serialized pool.

**Tradeoff.** Pool-only (no cron stagger) means the other three districts sit STARTED and blocked on the pool at 4am, which burns the run-timeout clock — hence 7200s. Run-level concurrency or staggered crons would avoid the idle wait; happy to switch if preferred.

## Action required before this fully works

- [ ] Set the `finalsite_api` pool limit to **1** in Dagster+ (Deployment, Concurrency). The `pool=` kwarg does not throttle until the limit is configured — the fix is inert without this step.

Refs #4408

## AI Assistance

Claude Code (systematic-debugging) performed the log/run investigation, wrote the code, and authored the tests. Root-cause interpretation and all decisions (pool vs stagger, adding the 403 retry, opening the issue) were human-directed.

### General

- [ ] Update due date and assignee on the TEAMster Asana project
- [ ] Review the Claude Code Review comment posted on this PR

### Dagster

- [x] Offline retry unit tests pass (`tests/resources/test_resource_finalsite.py`); all four district finalsite submodules import with the pool and tag applied. Full `dagster definitions validate` was not run locally (codespace env limits on kipptaf module-load) — relying on Dagster Cloud CI.
- [x] Change follows the library + config pattern (library factory kwarg + per-location schedule config).

### Docs

- [x] Automations catalog not regenerated: it renders only schedule name / cron / assets, none of which change by adding a `tags` kwarg or an asset `pool` (the existing overgrad `max_runtime` tag likewise is not shown there).

## CI checks

- [ ] Trunk — passes
- [ ] dbt Cloud — n/a (no dbt changes) or passes
- [ ] Dagster Cloud — passes or not triggered

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -582,6 +582,14 @@ wait — no `step_execution_timeout` knob exists. When a run hits `max_runtime`
 having done little work, suspect step-pod `FailedScheduling`, not slow code or
 upstream APIs.
 
+Concurrency-**pool**-blocked runs stay QUEUED, not STARTED (run blocking is the
+Dagster >=1.10 default; repo is 1.13), so pool queue-wait does NOT burn
+`dagster/max_runtime` (it counts from STARTED). With `k8s_job_executor` (all
+locations) each step runs in its own pod (compute is `pid 1`) and a resource's
+`setup_for_execution` runs there only after the op's pool slot is claimed — so a
+pooled resource's short-lived token/session is not aged by queue-wait. Size a
+pooled asset's `max_runtime` for its own run, not for waiting behind siblings.
+
 GKE Autopilot top-of-hour fan-out is the dominant cause of step-pod scheduling
 latency. `FailedScheduling` events trace to "Insufficient cpu/memory" (3-9 min
 waits) while nodes provision. Image pull is ~2s on cached nodes — don't chase
@@ -601,6 +609,12 @@ position expecting type 'String'".
 Authenticates as impersonated service account
 `codespaces@teamster-332318.iam.gserviceaccount.com`. If `PermissionDenied`,
 check the `CodespacesRole` custom IAM role, not user IAM bindings.
+
+`gcloud` via Bash is denied by a `Bash(gcloud *)` deny rule (full-path or
+variable-aliased invocations are classifier-flagged as evasion — don't). Prefer
+the GKE MCP (`list_clusters`/`get_cluster`) and gcp-observability MCP; for
+Compute resources with no MCP coverage (Cloud NAT, routers) or the gcloud
+commands noted elsewhere in this file, hand them to the user to run.
 
 `mcp__gke__query_logs` uses snake_case keys in `time_range` (`start_time`,
 `end_time`), not camelCase. Results cap at 100 — paginate by using the last
--- a/src/teamster/code_locations/kippcamden/finalsite/schedules.py
+++ b/src/teamster/code_locations/kippcamden/finalsite/schedules.py
@@ -1,4 +1,4 @@
-from dagster import ScheduleDefinition
+from dagster import MAX_RUNTIME_SECONDS_TAG, ScheduleDefinition
 
 from teamster.code_locations.kippcamden import CODE_LOCATION, LOCAL_TIMEZONE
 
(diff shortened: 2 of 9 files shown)


cbini:
## Follow-up commit eb12276b — addressing self-review

Ran a code review on the initial fix; two findings, both amplified or exposed by the max_runtime bump.

**1. No per-request timeout (fixed in eb12276b).** Added `request_timeout` (60s) to `FinalsiteResource` and broadened the bounded retry from 403-only to also cover connection errors and connect/read timeouts (per the repo network-retry convention). Without a timeout, the raised 7200s ceiling would let a hung socket stall a run for ~2h. New offline tests cover the network-fault retry and a 429 `Retry-After` regression.

**2. JWT 55-min TTL vs pull duration (verified, no code change).** The token is minted once (`exp = now + 55 min`) and never refreshed. Checked the data: kippnewark is the largest district at ~24k contacts, a sequential ~1 req/s pull of ~20 min — comfortably under 55 min. The token is minted after the pool slot is acquired, so queue-wait does not consume the window. Not implementing token refresh (YAGNI); worth revisiting only if contacts grow ~2.5x.

Also corrected the kippnewark max_runtime comment: its pull is ~20 min (under the 30-min default); it is pool queue-wait behind serialized siblings that can push a run past 30 min, which is what the 7200s ceiling covers.

Note: `claude-review` does not re-run on `synchronize`, so this commit was not auto-reviewed.

PR #4415, opened 2026-07-16
feat(powerschool): migrate kipppaterson SIS ingestion to dlt

### LINKED ISSUE #3807: feat(powerschool): replace homegrown ODBC sync with dlt

### LINKED ISSUE #4413: refactor(ssh): replace sshpass tunnel with in-process paramiko forward
## Context

`SSHResource.open_ssh_tunnel()` (`teamster/libraries/ssh/resources.py`) opens the
PowerSchool Oracle tunnel by shelling out to `sshpass` + `ssh -L` as a
subprocess, reading the password from a mounted password file. Friction:

- External `sshpass` + `ssh` binary dependency baked into the image.
- Subprocess lifecycle management — callers must `.kill()` the `Popen` in a
  `finally`, and there is a readiness race parsing its stdout before the
  forwarded port is up.
- Password sourced from a mounted file rather than an environment variable.

Surfaced by the #4398 dlt-vs-Sling spike: Sling runs its tunnel in-process (Go
`crypto/ssh`) with the password from connection config, which highlighted our
subprocess approach as legacy friction.

## Proposal

Replace the `sshpass` subprocess in `open_ssh_tunnel()` with an in-process
paramiko local forward (or the `sshtunnel` library, which wraps paramiko),
reading the password from the `PS_SSH_PASSWORD` environment variable. The same
`SSHResource` class already uses paramiko — with legacy `ssh-rsa` re-enabled —
for its SFTP methods, so the tunnel would reuse the existing transport and
host-key handling.

## Why this beats Sling's native tunnel

Sling's tunnel uses `InsecureIgnoreHostKey()` (skips verification). A paramiko
forward keeps real host-key verification AND the legacy `ssh-rsa` support — the
in-process simplicity without dropping verification.

## Scope

- `teamster/libraries/ssh/resources.py` and consumers of `open_ssh_tunnel()`
  (`teamster/libraries/powerschool/sis/odbc/`).
- Removes the `sshpass` binary dependency and the mounted password-file for
  PowerSchool SSH.

Refs #3807, #4398

cbini:
## Summary and Motivation

Replaces kipppaterson's Couchdrop-SFTP PowerSchool feed with dlt syncing live
from Paterson's Oracle database (issue #3807), and establishes the dlt template
the ODBC districts migrate onto next.

- New `libraries/dlt/powerschool/` factory (per-table `@dlt_assets`, full-refresh
  `replace`, pyarrow, keyless ADC to BigQuery; `oracle_number_adapter` keeps
  decimals off FLOAT64)
- In-process paramiko SSH port-forward on `SSHResource`
  (`open_ssh_tunnel_paramiko`) replacing the sshpass subprocess for the dlt path
  only (incumbent ODBC districts keep sshpass until they migrate)
- kipppaterson wiring: 48-table config (Newark active set intersected with
  Paterson's Oracle), two cron schedules (15-min intraday, nightly gradebook),
  definitions plus dagster-cloud secret wiring, `enable_legacy_rsa`
- dbt `staging/dlt` variant: all 47 shared models rebuilt from the odbc
  reference; kipppaterson swapped to dlt with gradebook/GPA parity to kippnewark
- Retired the Couchdrop-SFTP PowerSchool ingestion (dagster) and the
  `student_email` consumer chain; Paterson student email now comes from generated
  `@teamstudents.org` logins as Paterson migrates to the KTAF Google instance

Closes #3807

## AI Assistance

Authored by Claude Code (Opus 4.8) under human direction via the superpowers
brainstorm to spec to plan to subagent-driven-execution flow. All design
decisions (dlt over Sling, full-refresh strategy, table scope, gradebook/GPA
parity, `student_email` retirement, Paterson Google-email migration) were
human-directed; each task was implemented, reviewed, and fix-looped.

## Validation

- dbt Cloud CI: passing, zero warnings
- Dagster Cloud branch deployment: builds passing (all arch matrix)
- Trunk / CodeQL / Analyze: passing
- Data parity (dlt vs live SFTP staging, overlap tables): exact or within a
  couple rows (pull-timing). The one delta, `u_studentsuserfields` (876 vs
  1039), is empty null-padding rows the SFTP CSV added per student; Oracle holds
  only populated rows, so the LEFT JOIN result is identical (cleaner).
- All 48 dlt staging + 75 downstream models build green locally; contracts and
  uniqueness tests pass
- Whole-branch review: no correctness bugs across the 47 staging models, parity
  config, and SFTP retirement

## student_email retirement and Paterson Google-email migration

The Paterson-only SFTP `student_email` extract fed `student_email_google` for
Paterson (~504 real addresses on a non-KTAF domain). With the SFTP feed retired:

- The kipptaf `student_email` consumers are removed (kipptaf wrapper model +
  properties + `sources-kipppaterson` registration + the `se`-join in
  `base_powerschool__student_enrollments`, which now uses `sl.google_email` for
  all regions). The powerschool **package** sftp `stg_powerschool__student_email`
  model is retained but disabled (retired package models are disabled, not
  deleted).
- Paterson is added to the `stg_people__student_logins` generation (the
  `_dbt_source_relation not like '%kipppaterson%'` exclusion is removed), so
  Paterson active students get generated `@teamstudents.org` usernames/emails —
  the same convention as the other NJ regions — as Paterson migrates onto the
  KTAF Google instance. This populates `student_email_google` for Paterson from
  the next prod incremental run; the generated addresses align with the accounts
  being provisioned during the migration.

## Go-live checklist (owner action)

- [ ] Set a low concurrency limit on the `dlt_powerschool_kipppaterson` pool in
      Dagster+ deployment settings (UI) before the intraday schedule fans out —
      the per-table pipelines share one dlt state table and will 429 unbounded
- [ ] Confirm Paterson Google accounts are being provisioned to match the
      generated `@teamstudents.org` addresses (transitional timing)
- [ ] Squash merge (CODEOWNERS review on `src/dbt/` required)
- [ ] After a prod soak confirms dlt, tell Ops to stop the Couchdrop PowerSchool
      export
- [ ] Regenerate the automations catalog for the two new schedules

## CI checks

- [x] Trunk passes
- [x] dbt Cloud passes (zero warnings)
- [x] Dagster Cloud branch deployment builds


src/teamster/code_locations/kipppaterson/powerschool/sis/dlt/schedules.py
```diff
@@ -0,0 +1,66 @@
...
+_ungated_intraday = [
+    a
+    for a in config["assets"]
+    if a["schedule_tier"] == "intraday" and a["cursor_column"] is None
+]
+if _ungated_intraday:
+    raise ValueError(
+        "intraday tables require a cursor_column: "
+        + ", ".join(a["table_name"] for a in _ungated_intraday)
+    )
```
cbini: not sure if these are necessary. at most, they should be inside _tier_targets()

src/teamster/code_locations/kipppaterson/definitions.py
```diff
@@ -70,5 +75,13 @@
...
         "ssh_amplify": SSH_RESOURCE_AMPLIFY,
         "ssh_couchdrop": SSH_COUCHDROP,
+        "ssh_powerschool": SSHResource(
+            remote_host=EnvVar("PS_SSH_HOST"),
+            remote_port=EnvVar.int("PS_SSH_PORT"),
+            username=EnvVar("PS_SSH_USERNAME"),
+            password=EnvVar("PS_SSH_PASSWORD"),
+            tunnel_remote_host=EnvVar("PS_SSH_REMOTE_BIND_HOST"),
+            enable_legacy_rsa=True,
+        ),
```
cbini: shouldn't this be defined in resources.py?

src/teamster/libraries/dlt/powerschool/assets.py
```diff
@@ -0,0 +1,266 @@
...
+class PowerSchoolDagsterDltTranslator(DagsterDltTranslator):
+    def __init__(self, code_location: str):
+        self.code_location = code_location
+        super().__init__()
+
+    def get_asset_spec(self, data) -> AssetSpec:
+        asset_spec = super().get_asset_spec(data)
+
+        asset_spec = asset_spec.replace_attributes(
+            key=AssetKey([self.code_location, "powerschool", data.resource.name]),
```
cbini: I think asset key should probably be code_location/powerschool/sis/name

src/teamster/libraries/dlt/powerschool/assets.py
```diff
@@ -0,0 +1,266 @@
...
+
+        return _table_rows
+
+    def _build_source(selected: list[PowerSchoolTable], signatures: dict[str, dict]):
+        @dlt.source(name=source_name)
+        def _src():
+            for table in selected:
+                yield _build_resource(table, signatures.get(table.name))
+
+        return _src()
```
cbini: scoping? should these be defined at module level?

src/teamster/libraries/dlt/powerschool/assets.py
cbini: Verified the ODBC districts use `[code_location, "powerschool", table_name]` (`libraries/powerschool/sis/odbc/assets.py:100`) — no `sis` segment, and no asset key anywhere in the repo carries one. Since Paterson is the template the ODBC districts (newark/camden/miami) migrate onto next, matching their exact key shape keeps that convergence clean — a `sis` segment here would make Paterson the lone divergent district and would also break the schedule `_tier_targets`, which build `kipppaterson/powerschool/<table>`.

Recommend keeping the current 3-part keys. If we want `sis` to become the standard, that's a separate cross-district rename (all four powerschool locations at once), not a Paterson-only change — happy to open that separately, but it's out of this PR's scope. Let me know which way you want it.

src/teamster/code_locations/kipppaterson/definitions.py
cbini: Agreed — moving it into `kipppaterson/resources.py` as a `SSH_POWERSCHOOL` constant (matches the `FINALSITE_RESOURCE` already there and the resource-model convention in `src/teamster/CLAUDE.md`: location-specific resources live in `resources.py`). `definitions.py` will import and reference it.

src/teamster/libraries/dlt/powerschool/assets.py
cbini: Agreed — `_build_resource` / `_build_source` only reference module-level names plus their own parameters, so I'll lift them to module scope. Bonus: that makes them reusable for the ODBC-district migrations that will build on this template, rather than being re-created per factory call.

src/teamster/code_locations/kipppaterson/powerschool/sis/dlt/schedules.py
cbini: Folding both into `_tier_targets()` per your note, rather than standalone module-level blocks. I'd keep them (vs dropping): the tier-typo guard stops a mistyped `schedule_tier` from silently dropping a table out of *both* schedules (it would just never sync), and the intraday-cursor guard stops a no-cursor table from reaching the runtime probe, which would `AttributeError` on `MAX(null_cursor)`. Both are cheap load-time checks. Say the word if you'd rather drop them entirely and rely on the config unit test instead.

Issue #4427, opened 2026-07-17
feat(powerschool): migrate kippnewark and kippcamden SIS ingestion to dlt


cbini:
## Context

Paterson's PowerSchool SIS ingestion migrated from ODBC to dlt in #4415 (merged 2026-07-17), closing the pilot umbrella #3807. The dlt factory (`src/teamster/libraries/dlt/powerschool/`) is proven end-to-end: probe-gated full-replace, parallel `table_rows` extract, dlt-native `resource_state` signatures, progress streamed to the Dagster event log. This issue covers migrating the next two ODBC districts — **kippnewark** and **kippcamden** — onto that template, and tearing down the now-legacy PowerSchool ODBC code for **kippmiami**, which moved its SIS to Focus (`code_locations/kippmiami/dlt/focus/`).

## Template (from #4415)

- Library factory `build_powerschool_dlt_assets` in `libraries/dlt/powerschool/assets.py`: one probe-gated `@dlt_assets` multi-asset over all tables; per-table `PowerSchoolTable(name, cursor_column)`; `oracle_number_adapter` + `remove_nullability_adapter`; asset keys `[loc, powerschool, sis, table]`.
- Per-district wiring mirrors `code_locations/kipppaterson/powerschool/sis/dlt/`: `assets.py`, `config/assets.yaml`, `schedules.py` (intraday 15-min + nightly), `SSH_POWERSCHOOL` in `resources.py` (`enable_legacy_rsa`), `PS_DB_*` / `PS_SSH_*` in `dagster-cloud.yaml`, resource wiring in `definitions.py`.
- dbt: the shared `powerschool` package `staging/dlt` variant already exists; its `powerschool_dlt` source schema templates by `project_name`, so enabling the variant for the district and building it works with no source edits. Asset-key meta already carries the `sis` segment.

## Migrate (near-mechanical), per district: kippnewark, then kippcamden

- [ ] Derive the table scope + `cursor_column` map from the district's current ODBC config (`code_locations/{district}/powerschool/config/*.yaml`).
- [ ] Add `powerschool/sis/dlt/` (assets, config, schedules) + `SSH_POWERSCHOOL` + `definitions.py` wiring.
- [ ] Add `PS_DB_*` / `PS_SSH_*` config to `code_locations/{district}/dagster-cloud.yaml`.
- [ ] Enable the dbt `staging/dlt` variant for the district; disable the ODBC variant.
[...]

Issue #4424, opened 2026-07-17
refactor(powerschool): consolidate SSH resource construction and evaluate sshtunnel reuse


cbini:
## Context

Two related `/simplify` reuse findings on the PowerSchool SSH path (PR #4415), both best handled when the ODBC districts migrate onto the dlt template.

## 1. Centralize the PowerSchool SSH resource

`code_locations/kipppaterson/resources.py` builds `SSH_POWERSCHOOL` by re-specifying the four `PS_SSH_*` constructor args that the shared factory `get_powerschool_ssh_resource()` (`core/resources.py`) already assembles, plus two new fields the paramiko path needs (`password`, `enable_legacy_rsa`).

- Cost: the `PS_SSH_*` wiring now lives in two places; a change must be made twice.
- Direction: extend `get_powerschool_ssh_resource()` with optional `password` and `enable_legacy_rsa` args (defaulted, so the incumbent sshpass districts are unaffected), then call it from the code location.
- Deferred deliberately: centralizing before the ODBC districts migrate is premature; do it as part of that migration so all districts share one construction path.

## 2. Evaluate sshtunnel.SSHTunnelForwarder for the in-process forward

`libraries/ssh/resources.py` hand-rolls ~115 lines for `open_ssh_tunnel_paramiko` (a daemon-thread `_ForwardServer`, a per-connection `direct-tcpip` channel, a bidirectional select/sendall pump, plus load-bearing cleanup ordering). `sshtunnel` (0.4.0) is already a locked dependency, and `dagster_ssh.SSHResource` already exposes `get_tunnel()` returning an `SSHTunnelForwarder` that provides exactly this machinery.

- Real blocker (why it was hand-rolled): `SSHTunnelForwarder` builds its own paramiko transport and does NOT route through this repo's overridden `get_connection()`, so it bypasses the three-layer `enable_legacy_rsa` ssh-rsa re-enable the PowerSchool server requires. A naive swap fails KEX.
- Direction to evaluate: wrap the forwarder's `.start()` in the same `_PREFERRED_KEYS_LOCK` legacy-rsa mutation so the reuse is achievable without losing legacy-rsa handling. Weigh the ~115 lines saved against coupling to sshtunnel internals.

Refs #3807

cbini:
**Partially resolved by #4429.**

[...]

PR #4429, opened 2026-07-17
feat(kippnewark): migrate powerschool SIS from ODBC to dlt

### LINKED ISSUE #4423: refactor(powerschool): inject Oracle DB credentials as a resource in the dlt factory
## Context

Surfaced by a `/simplify` altitude review of PR #4415. The kipppaterson PowerSchool dlt factory (`src/teamster/libraries/dlt/powerschool/assets.py`) takes the SSH tunnel as a properly injected `ssh_powerschool` resource, but reads the Oracle DB connection (including the DB password) straight from the process environment inside `_oracle_connection_url()` via `os.getenv`, bypassing the Dagster resource layer.

## Problem

- The `PS_DB_*` variable names are hardcoded in library code, so DB creds cannot be reconfigured or test-injected through resource config the way every other resource is.
- The connection is invisible to the Dagster resource graph.
- It is asymmetric with the SSH half of the same code path (SSH is a resource; DB is not).
- This factory is the intended template the ODBC districts migrate onto, so the pattern propagates.

The repo already has the canonical shape: `PowerSchoolODBCResource` (a `ConfigurableResource` with user/password/host/port/service_name), wired as `DB_POWERSCHOOL` in `core/resources.py`.

## Proposed direction

Inject a DB-credentials resource into the op, resolve the SQLAlchemy URL once in the op body, and thread that URL string into the per-table resource closures. This respects the real constraint (dlt worker threads cannot see Dagster resources) while keeping connection config in the resource layer.

```python
# op body, once:
url = db_powerschool.sqlalchemy_url()   # or build from the injected resource fields
# then thread url into _build_resource(table, signature, url)
```

## Notes

- Requires wiring the resource in `definitions.py` and re-verifying the branch-deployment E2E (the URL now resolves in the op, not per-resource).
- Companion cleanup on the SSH side is tracked separately.

Refs #3807

### LINKED ISSUE #4439: feat(kippnewark): migrate PowerSchool SIS ingestion from ODBC to dlt
Sub-issue of #4427 — the **kippnewark** leg. Implemented in #4429.

## Done in #4429

- 57-table dlt migration (probe-gated full-replace over the SSH tunnel), intraday + nightly schedules, Newark ODBC path deleted.
- Refactors #4423 (Oracle credentials as a resource) and #4424 part 1 (shared paramiko SSH resource) folded in.
- 9 shared `staging/dlt` models added (disabled for kipppaterson); Newark `u_studentsuserfields` repointed to the `powerschool_dlt` source.

## Template gaps resolved (the "NOT mechanical" items in #4427)

- **Per-table column selection** — proved unnecessary. Newark's incumbent `select_columns` trimming was disproven in the spike; no `included_columns` axis was added to the factory.
- **Large-table replace strategy** — resolved by the paramiko rekey fix (`_persist_legacy_rsa` in `libraries/ssh/resources.py`). The DPY-4011 ceiling was a 512 MiB SSH rekey drop, not tunnel cost; naive full-replace now works at any size (`assignmentscore` 19M loads as one query). No windowing / FY-partitioning needed.

## Validation

- Full 57-table branch-deployment load at default concurrency succeeded (~15 min).
- Prod parity confirmed on assignmentscore / attendance / pgfinalgrades / storedgrades: zero dlt-only keys; only sub-1e-10 float epsilon on double-valued numeric columns (benign).

## Go-live (remaining before this closes)

- [ ] CODEOWNERS review + merge #4429
- [x] `DROP SCHEMA dagster_kippnewark_dlt_powerschool CASCADE` before the first prod run
- [x] Set the `dlt_powerschool_kippnewark` pool limit to 1
- [ ] Regenerate `docs/reference/automations.md` (run in a bootstrapped terminal, not the codespace)

Refs #4429

cbini:
# Pull Request

## Summary & Motivation

Migrates **kippnewark** PowerSchool SIS ingestion from the legacy Oracle **ODBC** path to the shared **dlt** factory (`build_powerschool_dlt_assets`) — 57 tables, probe-gated full-replace over the in-process SSH tunnel → BigQuery, the pattern proven in the Paterson pilot (#4415). Two schedules: intraday (`*/15`) for the 27 cursor tables, nightly (2 AM) for the 30 gradebook / no-cursor tables. Deletes Newark's ODBC ingestion path.

Folds in two queued refactors:

- **#4423** — inject Oracle credentials as a resource (`OracleResource` / `db_powerschool`) instead of inline construction.
- **#4424 (part 1)** — centralize the paramiko SSH tunnel resource (`open_ssh_tunnel_paramiko`).

Also adds 9 shared `staging/dlt/` models (disabled for `kipppaterson`, which doesn't extract them) and repoints Newark's district-specific `stg_powerschool__u_studentsuserfields` from `powerschool_odbc` to `powerschool_dlt`.

### Tunnel fix (uncovered during validation)

A full load of `assignmentscore` (19M) reproducibly died with `oracledb DPY-4011` at ~8.6M rows. Root cause (verified against paramiko 5.0 source): `SSHResource.get_connection` re-enabled the legacy `ssh-rsa` host-key algorithm **only for the initial handshake** and reverted the class-level patch immediately. paramiko renegotiates keys every `Packetizer.REKEY_BYTES` (512 MiB); against the ssh-rsa-only PowerSchool host, that mid-stream rekey's KEXINIT no longer offered `ssh-rsa`, so the transport dropped — surfacing as DPY-4011. `_persist_legacy_rsa` now re-applies `ssh-rsa` scoped to the returned **transport instance** (no process-global `RSAKey.HASHES` weakening) and disables auto-rekey on it.

Result: the apparent large-table ceiling is gone — **naive full-replace works at any table size; no windowing/merge strategy needed.** Regression: `tests/resources/test_resource_ssh_rekey.py` (loopback ssh-rsa-only server forced to rekey; no external creds).

## Validation

**Full 57-table load** on the branch deployment at dlt's default extract concurrency (the exact config that previously crashed at 127s): succeeded, 0 failures, ~15 min step, `assignmentscore` loaded all **19,153,141** rows.

**Prod parity** — dlt landing vs current prod ODBC staging, four highest-leverage tables:

| Table | prod rows | dlt rows | shared | dlt-only | value diffs |
| --- | --: | --: | --: | --: | --: |
| assignmentscore | 19,153,928 | 19,153,141 | 19,153,141 | 0 | 0 |
| attendance | 5,825,608 | 5,825,606 | 5,825,606 | 0 | 0 |
| pgfinalgrades | 4,525,026 | 4,525,026 | 4,525,026 | 0 | 157\* |
| storedgrades | 1,314,797 | 1,314,776 | 1,314,776 | 0 | 1,194\* |

- **Zero dlt-only keys** on every table — the migration never fabricates or duplicates rows; dlt is a clean subset of prod.
- **prod-only keys** (787 / 2 / 0 / 21) are extract-time drift (deletes/churn between the two pulls).
- **\*Value diffs are floating-point epsilon only** (max ~1.2e-11, down to a single ULP of 4.4e-16), confined to the `.double_value`-derived numeric columns (`points`, `percent`, `gpa_points`). No key, integer, string, date, or grade column differs on any shared row. dlt reads the raw Oracle `NUMBER` (NUMERIC) vs the legacy `.double_value` — if anything marginally more precise. Benign; no action.

## AI Assistance

Claude Code authored this end-to-end, human-directed — including the root-cause spike, systematic debugging of the tunnel fix, and the prod-parity validation.

### Dagster

- [x] Branch deployment builds + deploys clean (all `dagster-cloud-deploy` checks green) — stronger than a local `definitions validate`
- [x] Library + Config pattern (factory + YAML), asset keys `kippnewark/powerschool/sis/<table>`

### dbt

- [x] Contract + `.yml` on the 9 new `staging/dlt` models; disabled for `kipppaterson`
- [x] No breaking column renames — the `u_studentsuserfields` repoint preserves the contract
- [x] dbt Cloud CI green against the loaded dlt tables

### CI

- [x] Trunk · dbt Cloud · Dagster Cloud · CodeQL all green

## Remaining before merge / go-live

- [ ] Merge `main` (branch is ~124 commits behind)
- [ ] Regenerate the automations catalog for the 2 new schedules (`uv run scripts/gen-automations-doc.py`)
- [ ] One-time `DROP SCHEMA dagster_kippnewark_dlt_powerschool CASCADE` before Newark's first prod run — branch-deployment runs write the (non-branch-isolated) dlt dataset, so a schema/mode mismatch would otherwise recur at go-live
- [ ] CODEOWNERS review (analytics-engineers) — PR is currently `mergeable_state: blocked` pending approval

Refs #4427, #4423, #4424


cbini:
Review addressed — fixes in `7ecd544a7` and `21c706da5`. Dispositions:

**Fixed**

- **Stale `kippnewark` dbt CLAUDE.md** (Low) — corrected to `dlt` (`dlt.+enabled: true`, `odbc.+enabled: false`). `7ecd544a7`
- **`probe_signature` count coercion** (Low) — added `int(count)`, mirroring the `max_cursor` JSON-safe normalization. `7ecd544a7`
- **`EXTRACT__WORKERS` process-env write** (Low) — `pipeline.extract()` resolves `workers=ConfigValue` from dlt's config system, so this now sets `dlt.config["extract.workers"]` (verified readback) instead of mutating `os.environ`; dropped the now-unused `import`. The `dlt_extract_workers` run-tag knob is retained. `21c706da5`

**Verified — no change**

- **02:00 schedule race** (Medium) — pre-existing, not introduced here: `kipppaterson` runs the identical `*/15` + `0 2` crons in prod, and its most recent nightly run queued ~6.4 min at 02:00 (created 02:00:06 → started 02:06:31) then succeeded — i.e. the `pool` serialized it behind the intraday tick. On overlap the consequence is a self-healing spurious reload, not corruption. Not offsetting the cron: it would diverge from the proven Paterson pattern and wouldn't fully close it anyway (intraday runs every 15 min, so a long tick can overlap any nightly time). The real mechanism is the pool — will confirm `dlt_powerschool_kippnewark` limit = 1 in Dagster+ Concurrency at go-live.
- **`kippcamden` / `kippmiami` `dagster-cloud.yaml` +5 lines** (scope) — intentional and required: refactor #4424 gives the shared SSH resource a `password` field, so those districts (still ODBC, still using the shared resource) need `PS_SSH_PASSWORD` in both config blocks or resource init fails.
- **paramiko private attrs / duplicate YAML read** — acknowledged. The private-attr coupling is inherent to re-enabling a removed algorithm, documented in `libraries/ssh/CLAUDE.md`, and behaviorally guarded by the loopback rekey test (versions pinned). The duplicate config read keeps `assets.py` and `schedules.py` import-decoupled.

**Operational (tracked for go-live)** — the first post-`DROP SCHEMA` intraday tick full-replaces all 27 tables (fits under `max_runtime`; the 57-table validation run was ~15 min); probe cost is 27 `COUNT`/`MAX` every 15 min against Oracle — cheap if `transaction_date`/`whenmodified` are indexed, will glance at Oracle-side load after go-live.

PR #4468, opened 2026-07-20
perf(dbt): materialize assessment-scores star as tables for Cube query performance

### LINKED ISSUE #4333: fix(cube): assessment cube queries time out cold (47-73s) exceeding the 55s MCP poll deadline
## Summary

Cube MCP queries against the assessment cubes intermittently time out. Root cause is performance, not correctness: the mart fact and its dimensions are unmaterialized views re-expanded on every query.

## Evidence

The #4298 repro query (`count_scores` by `academic_year`, filter `module_code = QA1`) runs 47-73s cold in BigQuery (~3 GB scanned, ~7M slot_ms, `cache_hit: false`). The MCP server's `TIMEOUT_SECONDS = 55` poll deadline sits inside that band, so the query intermittently exceeds it; a retry appears to "work" only because the first attempt warmed Cube's cache. Verified: `fct_assessment_scores_enrollment_scoped` and every dimension it joins are views (`type=2`); only `dim_dates` is a table.

This is the timeout that #4298 explicitly deferred ("revisit separately if timeouts persist"). #4299 fixed the separate NULL `academic_year` join-correctness bug, which is confirmed working.

## Planned fix (Approach C)

1. Materialize `fct_assessment_scores_enrollment_scoped` as a table (the `dim_dates` pattern), measure, and materialize the heaviest dimension views only if still needed.
2. Add a standard-level Cube pre-aggregation on the additive primitives (`count_scores`, `_sum_proficient`) for the recurring standard-level-proficiency-by-demographic pattern.

Spec to follow under `docs/superpowers/specs/`.

Refs #4298, #4299.

### LINKED ISSUE #4464: perf(dbt): materialize assessment-scores star for Cube query performance
## Problem

Effectively 100% of Cube's BigQuery spend is the assessment-scores star. Last 14 days (`cube-cloud@` service account): 114 queries, 583 slot-hours, 307 GB billed — 107 of those queries (all of the cost) hit `fct_assessment_scores_enrollment_scoped` and its dims. Typical query: ~3 GB, ~5 slot-hours, 95s average latency, 495s worst. By contrast `fct_student_attendance_daily` queries run in ~3s.

Root cause: the entire chain is views. A single SELECT on the fact expands to ~55 base tables across every assessment source plus per-district PowerSchool enrollments. Per Cube query, BigQuery recomputes from staging up:

- `int_assessments__resolved_section_enrollments` (view): 6-source union, two non-equi date-range joins against course enrollments, anti-join, global dedupe window. Joined 3x by the fact (once per union branch), so the subtree can be computed up to 3x within one query. Standalone: 0.31 slot-hr.
- `int_assessments__response_rollup` (view, 23.4M rows): 0.47 slot-hr standalone.
- `int_assessments__course_enrollments` (view): the date-range join target, expanded up to 6x per query. Already has a commented-out `materialized: table` config.
- `dim_assessments` (6,311 rows) costs 0.52 slot-hr per computation because its view re-walks the Illuminate chain.

Rebuild cadence if materialized: `int_assessments__scaffold` (already a table in this chain) rebuilds 2x/day at 0.28 slot-hr — that is the automation-condition cadence the new tables would inherit. Cube currently burns ~42 slot-hr/day querying.

Dollar cost is small today; the acute problem is latency (1.5–8 min per Cube MCP query) and headroom as MCP usage grows.

## Plan

1. Materialize assessment intermediates as tables: `int_assessments__resolved_section_enrollments`, `int_assessments__response_rollup`, `int_assessments__course_enrollments` (uncomment existing config).
2. Materialize the star's marts as tables: `fct_assessment_scores_enrollment_scoped` (clustered on join/filter keys), `dim_assessments`, `dim_assessment_administrations`.
3. Not incremental — full fact rebuild is ~1.4 slot-hr / ~35s at ~2x/day; nothing for incremental to save yet.
4. Later, optional: a bounded Cube pre-aggregation for the hottest MCP rollup shape, only if step 2 is not fast enough (the prior attempt, reverted in #4462, was unbounded).

Expected outcome: Cube queries join real tables instead of re-deriving the chain; latency drops from ~100s to seconds; view-recomputation slot burn is replaced by ~2 rebuilds/day of a handful of tables (fresh-on-ingest, since the data-change automation condition only fires when assessment/enrollment data actually lands).

cbini:
# Pull Request

## Summary & Motivation

Closes #4464

When merged, this pull request will materialize the assessment-scores star — 3 intermediates plus 11 marts (the fact and its full FK closure of dims) — as tables instead of views, eliminating the per-query recomputation that makes Cube MCP assessment queries cost ~5 slot-hours and run 95–495s each.

Profiling (14 days of `cube-cloud@` query history) showed the assessment star accounts for effectively 100% of Cube's BigQuery spend: every query re-expanded ~55 base tables through the all-view chain, recomputing `int_assessments__resolved_section_enrollments` (two non-equi date-range joins plus dedupe) up to 3x per query. Full detail in #4464.

Scope notes:

- The 11-mart set is the FK-constraint closure from the fact: BigQuery FK DDL requires every referenced relation to be a table with a PRIMARY KEY constraint, so the fact's dims (and their dims) convert together. It coincides exactly with the star's Cube join set.
- `warn_unenforced: false` added alongside `warn_unsupported: false` on the converted models' constraints (they now render into DDL, unenforced).
- Rebuild cost is small and fires only on real data changes: full star rebuild measured at ~2.4 min end-to-end (fact CTAS 17s); the chain's existing table (`int_assessments__scaffold`) rebuilds ~2x/day.
- Staging (`zz_stg_kipptaf_*`) copies of all 14 models were pre-built as tables (with PK constraints) so CI's defer-resolved FK references land on valid targets — dbt resolves FK constraint refs to the deferred relation whenever `--defer` is active, so CI would otherwise fail against the old staging views.

## AI Assistance

Claude Code profiled Cube query costs, identified the hotspot chain, authored the materialization configs, and validated dev + staging builds; scope and rollout sequencing decisions were human-directed.

### General

### dbt

- [x] Properties files updated for all modified models (config-only changes; no new models)
- [x] Exposures unchanged — all models already covered by `cube_semantic_layer`
- [x] No breaking change — materialization conversion only; no column renames/removals, no surrogate-key churn. view→table conversion needs no manual drop (dbt replaces the view).
- [x] No external sources modified; staging copies pre-built via `dbt run --target staging` for the 14 converted models

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

### DIFF
--- a/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__course_enrollments.yml
+++ b/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__course_enrollments.yml
@@ -6,8 +6,8 @@ models:
       Illuminate assessments are joined against. One row per (student, academic
       year, subject area, date enrolled), deduplicated to the latest exit per
       enrollment span.
-    # config:
-    #   materialized: table
+    config:
+      materialized: table
     data_tests:
       - dbt_utils.unique_combination_of_columns:
           arguments:
--- a/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__resolved_section_enrollments.yml
+++ b/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__resolved_section_enrollments.yml
@@ -14,6 +14,8 @@ models:
       homeroom section active on the same anchor date. Scores with no matching
       section in either tier are dropped. The resolution_type column identifies
       which tier resolved the row.
+    config:
+      materialized: table
     columns:
       - name: source_type
         data_type: string
--- a/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__response_rollup.yml
+++ b/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__response_rollup.yml
@@ -9,6 +9,8 @@ models:
       selection. `assessment_id` here is the canonical id, not the per-row
       member id; `assessment_ids` (array) preserves the underlying member ids
       for drilldown.
+    config:
+      materialized: table
     data_tests:
       - dbt_utils.unique_combination_of_columns:
           arguments:
--- a/src/dbt/kipptaf/models/marts/dimensions/properties/dim_assessment_administrations.yml
+++ b/src/dbt/kipptaf/models/marts/dimensions/properties/dim_assessment_administrations.yml
@@ -12,6 +12,8 @@ models:
       scheduling column carry a typed NULL for that column (e.g. Illuminate and
       AP carry NULL administration_period; state branches carry NULL
       source_assessment_id).
+    config:
+      materialized: table
     columns:
       - name: assessment_administration_key
         data_type: string
@@ -23,6 +25,7 @@ models:
         constraints:
           - type: primary_key
             warn_unsupported: false
+            warn_unenforced: false
         data_tests:
           - unique
           - not_null
@@ -41,6 +44,7 @@ models:
             to: ref('dim_assessments')
             to_columns: [assessment_key]
             warn_unsupported: false
+            warn_unenforced: false
         data_tests:
           - relationships:
               arguments:
@@ -59,6 +63,7 @@ models:
             to: ref('dim_dates')
             to_columns: [date_key]
             warn_unsupported: false
+            warn_unenforced: false
         data_tests:
           - relationships:
               arguments:
--- a/src/dbt/kipptaf/models/marts/dimensions/properties/dim_assessments.yml
+++ b/src/dbt/kipptaf/models/marts/dimensions/properties/dim_assessments.yml
@@ -18,6 +18,8 @@ models:
         land in the same bucket (e.g., SAT EBRW and ACT Reading combine to
         EBRW/Reading). Use this for apples-to-oranges comparisons across
         test families.
+    config:
+      materialized: table
     columns:
       - name: assessment_key
         data_type: string
@@ -31,6 +33,7 @@ models:
         constraints:
           - type: primary_key
             warn_unsupported: false
+            warn_unenforced: false
         data_tests:
           - unique
           - not_null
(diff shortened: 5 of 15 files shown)


cbini:
Review response:

1. **`dim_dates` missing `warn_unenforced: false`** — valid, deferring intentionally: a config edit to `dim_dates` marks it `state:modified` and fans `state:modified+` out to nearly every mart on the next CI run, for a parse-warning nicety that's pre-existing on main. Will fold into a future PR that already touches `dim_dates`.
2. **`int_assessments__resolved_section_enrollments` uniqueness test** — real, pre-existing gap. Its grain is unique by construction (`dbt_utils.deduplicate` on `score_grain_key` as the final step), but the key isn't projected, so adding the test needs a small SQL change. Follow-up candidate under #4464 rather than scope creep here.
3. **Conformed-dim blast radius** — intended and reviewed during design: the FK closure was chosen deliberately (it equals the star's Cube join set), the data-change automation condition covers table models by default so refresh stays prompt (fresh-on-ingest), and consumers see the same contract/columns. Rebuild cost measured at ~2.4 min for the full set.

CI warnings triage: all 9 test warnings are pre-existing by construction — this PR changes materialization config only, so compiled SQL is byte-identical to main; the survey/grades/attendance marts appear only because they're downstream of the converted dims under `state:modified+`. The PK-uniqueness warns trace to the documented PowerSchool double-write corpus (#3900 / #3915).

cbini:
Correction to point 1 above: the fan-out claim was overstated — `dbt ls --select dim_dates+` is 39 models, roughly the same scope this PR's CI already built. Implemented instead of deferring: 54a5e88 adds `warn_unenforced: false` to the `dim_dates` primary-key constraint, making the FK closure's constraint config fully consistent. Flag-only change (no DDL difference — it suppresses the parse warning), so the staging copies need no re-seed. Points 2 and 3 stand as written.

Issue #4482, opened 2026-07-21
investigate int_seat_tracker__snapshot grain and add uniqueness test


cristinabaldor:
## Context

Surfaced during #4475 (seat tracker hard-delete tracking). `int_seat_tracker__snapshot` has no uniqueness test, which violates the repo convention that every intermediate model carries one (`src/dbt/CLAUDE.md`).

## Problem

The model is not unique on the seat key, and adding the naive test fails hard. Current deployed grain:

- 128,212 rows
- 7,772 distinct `(staffing_model_id, academic_year)` (~16 rows per seat-year)
- 127,350 distinct `(staffing_model_id, academic_year, valid_from)` — i.e. 862 residual duplicate rows even at the SCD2 version grain

The high row count and per-seat-year multiplicity come from the `stg_google_appsheet__seat_tracker__log_archive` UNION combined with the dbt snapshot history; the 862 residual dupes suggest overlap between the log-archive rows and snapshot versions on the same `valid_from`.

## Action

1. Determine the true intended grain of `int_seat_tracker__snapshot`.
2. Investigate the 862 `(staffing_model_id, academic_year, valid_from)` duplicates (log_archive vs snapshot overlap).
3. Add an appropriate uniqueness test (`dbt_utils.unique_combination_of_columns`), at `severity: error` if the grain can be made clean, or `warn` with a `TODO` if residual dupes are a known upstream issue.

## Out of scope

The #4475 work deliberately does not add the test to avoid shipping a failing/warn test against an unclear grain.

PR #4483, opened 2026-07-21
feat(dbt): track and exclude hard-deleted seat tracker rows

### LINKED ISSUE #4475: feat(dbt): flag hard-deleted seat tracker rows as 'Row Deleted'
## Problem

`snapshot_seat_tracker__seats` uses the default `hard_deletes: ignore`, so rows hard-deleted in the AppSheet seat tracker are never invalidated. Their last version stays open (`dbt_valid_to = 9999-12-31`) indefinitely, so deletions are invisible in the snapshot and every downstream model. A point-in-time row count cannot reveal a deletion.

Confirmed empirically:

- Across 7,541 distinct seats in the snapshot, 0 have ever been closed without an open successor (proof that hard-deletes are not tracked today).
- ~112 seats are currently open in the snapshot but absent from the live source `kipptaf_appsheet.src_seat_tracker__seats` (i.e. deleted at source).
- Backlog breakdown by original status: 45 `Open`, 30 null-status, 28 `Staffed`, 9 `Position Closed`. ~24 had a real teammate assigned; all are either now terminated or still hold other live seats, so none stranded an active teammate.

## Goal

Distinguish hard-deleted rows from legitimate `Position Closed`, preserve the original status, backfill the existing ~112, and capture future deletions automatically. All in dbt, with no AppSheet app or permission changes.

## Approach

Path 1 (backfill + forward, model-side):

- In `int_seat_tracker__snapshot`, flag open snapshot seats whose `(academic_year, staffing_model_id)` is absent from the live staging source as `is_row_deleted = true`, and expose a derived reporting status `Row Deleted`.
- Keep the original `staffing_status` intact in its own column (28 of the backlog are `Staffed` and must not be overwritten).
- Handles both the ~112 backlog and all future deletions on each build.

Path 2 (dated audit trail, forward-looking):

- Set `hard_deletes: new_record` on `snapshot_seat_tracker__seats` (adds the `dbt_is_deleted` meta column, giving a real dated tombstone per future deletion).
- Requires the documented one-time snapshot migration — dbt does not auto-migrate an existing snapshot (see docs link below).
- Adjust downstream active-seat counts to exclude deleted rows.

## Notes / risks

- Snapshot migration required for Path 2: https://docs.getdbt.com/reference/resource-configs/hard-deletes
- Active-seat counts will legitimately drop once deleted rows are excluded from the open set — coordinate with dashboard/report owners before rollout.
- Historical deletion dates are unrecoverable; backfilled rows date to the cutover build, not the true deletion date.

## Out of scope

- Any AppSheet app change, delete-action rework, or permission change.
- Writing rows back into the AppSheet backing table.

A short design note will be added on the feature branch before implementation.

cristinabaldor:
## Summary and Motivation

When merged, this PR makes hard-deleted AppSheet seat-tracker rows detectable and excludes them from staffing reporting, and starts capturing future deletions with a dated audit trail — entirely in dbt, no AppSheet app or permission changes.

Background: `snapshot_seat_tracker__seats` used the default `hard_deletes: ignore`, so rows deleted in the AppSheet seat tracker were never invalidated — their last version stayed open (`dbt_valid_to = 9999-12-31`) forever and kept counting in staffing metrics. ~100 open seats are flagged deleted (all confirmed intentional deletions).

Changes:

- `snapshots/google_appsheet.yml` — `hard_deletes: new_record` on the seat tracker snapshot (dated `dbt_is_deleted` tombstones going forward).
- `int_seat_tracker__snapshot` — new `is_deleted` boolean, derived by anti-join to the live staging source `stg_google_appsheet__seat_tracker__seats`, guarded so `log_archive`-sourced rows are never flagged. Deliberately does NOT reference `dbt_is_deleted`, so it works whether or not the snapshot migration has run (Path 2 decoupled from Path 1).
- `int_topline__seats_staffed_weekly_aggregations` — excludes deleted seats from the current-year staffed fill-rate metric.
- `dim_staffing_positions` — drops deleted positions via inner-join to live keys (nothing FKs into this dimension).
- `rpt_tableau__seat_tracker_snapshot` — unchanged: it already inner-joins the live source, so it never surfaced deleted seats. An earlier redundant `is_deleted` filter was reverted.

AI assistance: authored with Claude Code via subagent-driven per-task implementation and review; design decisions (exclude-everywhere scope, boolean-only flag, dim drops rows) were human-directed.

Impact (dashboard owners): excluding deleted, mostly-unfilled seats RAISES staffed fill rate as a data-quality correction. This affects the topline staffed metric and the Cube staffing dimension — NOT the Tableau seat-tracker snapshot viz, which already excluded them. Please give topline and Cube consumers a heads-up.

Rollout and hand-offs:

1. Prod snapshot migration (before the first `new_record` run): add `dbt_is_deleted` to the existing snapshot table, or results are inconsistent (dbt does not auto-migrate an existing snapshot). See https://docs.getdbt.com/reference/resource-configs/hard-deletes. The `is_deleted` logic does not depend on it, so the reporting fix lands even if the migration lags.
2. Delete the phantom NULL-year seat `a65917ee` from the AppSheet source.

Follow-up: #4482 tracks the `int_seat_tracker__snapshot` grain investigation and uniqueness test (deliberately out of scope here — the model is not unique on the seat key, 862 residual `valid_from` dupes).

Closes #4475

### dbt

- [x] Properties file updated — `is_deleted` documented in `int_seat_tracker__snapshot.yml`.
- [x] Exposures verified — no column renames/removals on contracted models; `rpt` and `dim` output columns unchanged (row filters/drops only); `is_deleted` added only to the intermediate.
- [x] Not a breaking change — no contracted column renamed or removed.

## CI checks

- [x] Trunk passes — checked with `trunk check --force` on all changed files.

Note: dbt Cloud CI builds `state:modified+`, which includes the modified snapshot. The `new_record` mode may require the `dbt_is_deleted` column present in the CI/staging snapshot table; I will monitor CI and coordinate the migration if it flags.


cristinabaldor:
## Prod migration runbook — `hard_deletes: new_record` (rollout step 1)

@cbini — this is the one hand-off needed to complete this PR: the existing `snapshot_seat_tracker__seats` table needs the `dbt_is_deleted` meta column before the first `new_record` run (dbt does not auto-migrate an existing snapshot). It is **dev-validated** — I cloned prod, applied these exact steps, and the first `new_record` run was a clean MERGE that produced 114 correctly-flagged tombstones, 0 NULLs, 0 false positives.

Column type/values confirmed from dbt 1.11 source (`helpers.sql`): `dbt_is_deleted` is a **STRING** holding `'True'` / `'False'`.

### 1. Run this in BigQuery (DDL/DML on the prod table)

```sql
-- backup
create table `teamster-332318.kipptaf_google_appsheet.snapshot_seat_tracker__seats_backup_20260721` as
select * from `teamster-332318.kipptaf_google_appsheet.snapshot_seat_tracker__seats`;

-- add the meta column dbt expects
alter table `teamster-332318.kipptaf_google_appsheet.snapshot_seat_tracker__seats`
add column if not exists dbt_is_deleted string;

-- backfill existing rows to 'False'
update `teamster-332318.kipptaf_google_appsheet.snapshot_seat_tracker__seats`
set dbt_is_deleted = 'False'
where dbt_is_deleted is null;
```

Safe to run before or after merge — the column is inert until `new_record` reads it.

### 2. After merge, let the snapshot run

The next Dagster run of `kipptaf/google/appsheet/snapshot_seat_tracker__seats` runs in `new_record` mode. Expect ~114 tombstone rows (`dbt_is_deleted = 'True'`, open) for the current deleted-from-source backlog, dated at that run. Backlog true-deletion dates are unrecoverable; every deletion from that run onward gets a real dated tombstone.

### 3. Verify

```sql
select dbt_is_deleted, count(*)
from `teamster-332318.kipptaf_google_appsheet.snapshot_seat_tracker__seats`
group by 1;
-- expect ~114 'True', the rest 'False', 0 NULL
```

If anything looks off, the `_backup_20260721` table restores prior state.

cbini (approved):

PR #4488, opened 2026-07-21
feat(dbt): FRESH dashboard scaffold revamp and documentation

### LINKED ISSUE #4451: feat(dbt): FRESH dashboard scaffold revamp and documentation
## Summary

Refactor the FRESH dashboard's enrollment scaffold (school x grade spine) so
its source can be swapped cleanly between the manual
`stg_google_sheets__finalsite__school_scaffold` sheet and a PowerSchool-native
derivation, instead of being permanently hand-maintained.

## Motivation

The scaffold is currently a hand-edited Google Sheet
(`academic_year, region, schoolid, school, grade_level, org, school_level`),
consumed directly by `rpt_tableau__fresh_dashboard_progress_to_goals` and via
`int_google_sheets__finalsite__scaffold` by
`rpt_tableau__fresh_dashboard_aggregated`. Early in a recruitment cycle the
target year has no PowerSchool actuals yet, so the sheet is necessary — but
once a grade/school is live in PowerSchool, the sheet becomes a manual-upkeep
liability with no source-of-truth backing.

## What this delivers

- A new seam model deriving the scaffold from `stg_powerschool__schools`
  (grade-span expansion via `low_grade`/`high_grade`), blended with the sheet
  so sheet rows only fill combos PowerSchool doesn't have yet (new
  schools/grades not yet live).
- A single dbt var controlling source mode (`gsheet` / `powerschool` /
  `blend`), so switching is a one-line change.
- Replace the hardcoded `enrollment_academic_year = 2026` in
  `rpt_tableau__fresh_dashboard_progress_to_goals` with an explicit
  recruitment-year var.
- A living reference doc (`docs/reference/fresh-dashboard-data-model.md`,
  mirroring the gradebook-audit doc pattern) covering the full lineage,
  including an open question on historical/multi-year reporting given
  Finalsite's data model.
- A new `.claude/skills/fresh-dashboard/` skill, tiered for both non-engineer
  (Ops/Analyst self-serve) and engineer (full lineage + swap mechanics)
  audiences.

## Out of scope

- The goals side (`stg_google_sheets__finalsite__goals`) is unchanged — only
  the scaffold/spine is being made pluggable.
- Historical/multi-year scaffold support — flagged as an open question in the
  reference doc, not solved by this issue.

## AI Assistance

AI-assisted: design and implementation via Claude Code, human-directed.

## What Focus needs to provide (Miami)

| # | Gap                                        | Field(s) needed                                                                                                                                                                     |
| - | ------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1 | Equivalent to `stg_powerschool__schools` | School id, name/abbreviation, and (if Focus mixes in non-reporting entities) an exclude-from-reporting flag -- builds the scaffold's school list. Focus school ids are alphanumeric strings (e.g. `2332B`); our scaffold's `schoolid` column is `int64` -- no direct cast. Needs translation via `int_people__location_crosswalk.location_focus_school_id`, already mapped to the integer `location_powerschool_school_id` our scaffold uses. |
| 2 | Equivalent to `stg_powerschool__students`, for FS/PS quality checks | academic_year, enroll_status, school name or abbreviation (the school the student is actually enrolled in per Focus, to check it matches what Finalsite says), grade_level, enrollment start date, enrollment end date, rn_year |
| 3 | Equivalent to `stg_powerschool__calendar`  | First-in-session date per school/year -- needed alongside enrollment start date to compute the first-day-of-school flag |
| 4 | Equivalent to `stg_powerschool__u_studentsuserfields.infosnap_id` | A field on the Focus student record linking back to the Finalsite applicant id. Without it, Focus data can't be joined back to a Finalsite applicant at all. |

Full detail: `docs/reference/fresh-dashboard-data-model.md`, "Open questions".

GabyRangelB:
# Pull Request

## Summary & Motivation

...make the FRESH dashboard's enrollment scaffold source-swappable between the
hand-maintained Google Sheet and a PowerSchool-native derivation, replace the
hardcoded `2026` recruitment year with an explicit var, and ship a living
reference doc + skill for the pipeline. Also adds a Focus-sourced enrollment
comparison model for Miami, since Miami's PowerSchool feed stopped producing
current-year data after its SIS moved to Focus.

- New seam model `int_finalsite__enrollment_scaffold`: derives grade
  membership from **actual enrollment** (`stg_powerschool__students`, active
  status), not `stg_powerschool__schools.low_grade`/`high_grade` — those
  reflect a school's eventual grade span, not what's currently served, and
  produced phantom rows for growing schools (Hatch/Rise/Purpose). Blends in
  sheet rows only for school/grade combos PowerSchool doesn't have yet.
- Single dbt var (`finalsite_scaffold_source`: `gsheet` / `powerschool` /
  `blend`) controls the source mode.
- `finalsite_recruitment_year` dbt var replaces the hardcoded `2026` literal
  at every FRESH-dashboard site (`kipptaf/dbt_project.yml`) — one edit instead
  of several, kept manual (not derived) since Finalsite's recruitment cycle
  timeline is too fluid to key off a fixed date.
- `int_focus__student_enrollments`: Focus-sourced enrollment data for Miami
  (grade level, resolved enrollment/withdrawal codes, PowerSchool-aligned
  school id crosswalk, point-in-time `is_enrolled_*` flags, `enroll_status`),
  mirroring `int_extracts__student_enrollments`'s shape so Miami has real
  comparison data instead of nulls in the scaffold. Promotes three
  `stg_focus__*` kipptaf-level passthroughs (`school_gradelevels`,
  `student_enrollment_codes`, `custom_field_select_options`) to support it.
- Three new tests: a warn-severity scaffold/goals region-mismatch check, a
  `status_order`/`detailed_status_ranking` sync guard, and a single-year guard
  on the status crosswalk sheet.
- `docs/reference/fresh-dashboard-data-model.md` — full lineage reference,
  mirroring the gradebook-audit doc pattern, including open questions on
  Miami/Focus support and historical/multi-year reporting.
- `.claude/skills/fresh-dashboard/` — maintenance skill for this pipeline.

**Follow-up expected**: this PR is scoped to the PowerSchool-side swap plus the
Focus enrollment comparison data. Wiring Miami into the `blend` scaffold mode
itself (removing its 100%-sheet-sourced carve-out) is a natural next step now
that Focus provides live enrollment data, but is a separate change to
`int_finalsite__enrollment_scaffold.sql`.

Refs #4451

## AI Assistance

AI-assisted: design, implementation, and the grade-membership bug fix (derive
from enrollment actuals instead of school config) via Claude Code,
human-directed throughout — including rejecting an initial unconfirmed attempt
at the enrollment-actuals fix before it was scoped and approved.

### General

      why. (Claude is advisory — use your judgement, but don't ignore it.)

### dbt

      analysis, or application — no exposure changes needed; existing FRESH
      exposures already cover the downstream `rpt_tableau__*` models
- [x] Breaking change? No — purely additive models/tests; existing
      `rpt_tableau__fresh_dashboard_*` models are repointed but keep their
      contracted shape
- [x] No new external sources added

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

## Troubleshooting


GabyRangelB:
Thanks for the review. Ran down both 🟠 items against live data:

- **`school_number` uniqueness**: queried `stg_powerschool__schools` scoped to
  `state_excludefromreporting = 0` directly — 0 duplicate `school_number`
  values across all four districts. The guard test holds; no fan-out risk.
- **Goals silently dropped by the blended spine**: compared the old prod table
  (`kipptaf_google_sheets.int_google_sheets__finalsite__scaffold`, 1,719 rows)
  against a dev build of the new `int_finalsite__goals_scaffold` — same row
  count, and every row reconciles 1:1 on
  `(academic_year, region, schoolid, grade_level, goal_type, goal_name)`. The
  earlier full-column join returning a mismatch was a `goal_value`
  precision/formatting artifact, not an identity difference. Nothing dropped,
  nothing added.

Addressed the low-severity nits in `390b38c92`:

- Added `unique` to `int_finalsite__current_academic_year.academic_year`.
- Cross-referenced the `status_order` CASE and its mirror test with comments
  pointing at each other, so a future edit to one is more likely to catch the
  other.

Left as-is (per the review's own "optional/pre-existing" call):

- The `goal_value`-in-uniqueness-key issue on `int_finalsite__goals_scaffold`
  — inherited verbatim from the deleted `int_google_sheets__finalsite__scaffold`,
  and the goals side is explicitly out of scope for this PR (#4451 only makes
  the scaffold/spine pluggable).

This PR is also expected to get another round once Focus provides the fields
listed in #4451's "What Focus needs to provide (Miami)" section — flagging so
that context carries into any later review pass.

GabyRangelB:
@claude

GabyRangelB:
> **Claude finished @GabyRangelB's task in 20s** —— [View job](https://github.com/TEAMSchools/teamster/actions/runs/29869633980)
> 
> ### Todo
> * [x]  Read trigger comment for instructions
> 
> ### Nothing to act on
> The trigger comment (`@claude`) doesn't contain a request — no question, review ask, or implementation instruction after the mention.
> 
> For context, there's already a full static review posted above (from the earlier `@claude` run), and @GabyRangelB has since replied confirming both 🟠 findings are resolved (verified `school_number` uniqueness with no duplicates, and confirmed the goals scaffold row counts/keys reconcile 1:1 old-vs-new) and pushed `390b38c92` addressing the two 🟡 nits.
> 
> If you had something specific in mind — e.g., re-reviewing after `390b38c92`, checking a particular file, or a new implementation ask — reply with the details and I'll pick it up.

review PR please @claude

GabyRangelB:
@claude Thanks for the review — addressed the items that were in the new Focus enrollment work's scope, pushed in `b21210af3`:

- Reverted the five `-- force state:modified...` comments (`stg_focus__students/student_enrollment/schools/attendance_calendar(s)`) — those were a workaround for a transient stale per-PR-schema shadow, which is resolved now, so they'd have merged as inert/misleading.
- Tagged `contains_pii: true` on `int_focus__student_enrollments.yml`'s direct FERPA identifiers (`student_number`, `student_first_name`, `student_last_name`, `fteid`, `student_email`, `dob`), matching `int_extracts__student_enrollments`'s precedent.
- Fixed the one-sided cast in the Finalsite join — added a `finalsite_ids` CTE that precomputes `cast(focus_student_id_prefixed as int)`, joined on the plain column instead of casting inline in the `ON` clause.
- Dropped `is_enrolled_y1` — provably always `true` since `exitdate` is always coalesced to a value, and nothing downstream reads it.
- On `is_same_day_status_duplicate`: after tracing the actual ingestion asset (`status_report`'s `partitions_def` is a school-year file-drop partition, not a reload-date or student-scoped one), we don't think this needs a code fix. `_dagster_partition_key` is the nominal export-file year a row was ingested from — not a student-relevant boundary, since Finalsite tracks a student continuously rather than bounding cleanly by academic year. The existing scoping (`finalsite_enrollment_id` + `enrollment_academic_year`, either explicit in the partition or via the enclosing query's single-year filter) is the correct business key already. Documented that reasoning directly on the column in `int_tableau__finalsite_student_scaffold.yml` rather than leaving it implicit.

Still reviewing the remaining findings (join-guard scope, hardcoded recruitment year, missing column descriptions, `accepted_values` severity) — will follow up separately.

GabyRangelB:
@claude You were right to keep pushing on `is_same_day_status_duplicate` — ran the verification queries this time instead of reasoning further, pushed the fix in `f3d97a2f3`.

Correcting my own last comment: the `_dagster_partition_key` reasoning was real (confirmed `enrollment_academic_year = 2026` does span 2 partition keys, `2025_26` and `2026_27`), but it wasn't actually the mechanism inflating the flag. The real cause: the crosswalk join intentionally fans one `detailed_status` out to multiple `status_group_value` rows (one status can count toward several goal groupings). `countif` ran *after* that fan-out, so any status landing on `latest_status_date` that also happened to be multi-grouped got double-counted as a "duplicate" — verified empirically at a **~30% flag rate** on a fresh dev build, almost entirely fan-out artifacts (traced one specific case: a single "Did Not Enroll" status counted twice because it maps to both `Offers to Accepted` and `Offers to Enrolled`).

Fix: moved the calculation into its own `same_day_duplicate_flags` CTE, computed directly off `int_finalsite__status_report_unpivot` *before* the crosswalk join, deduplicated on `(finalsite_enrollment_id, detailed_status, status_start_date)` to also collapse the partition-recurrence case. Verified flag rate drops to **~7.7%**, and spot-checked both a corrected-to-false case (the fan-out artifact above) and a genuine true case (5 distinct statuses — Applicant through Enrolled — all landing on one backfilled `status_start_date`, matching the documented "Reset Protocol" same-day scenario).

Still working through the remaining findings (join-guard scope, hardcoded recruitment year, missing column descriptions on `int_finalsite__status_report_unpivot.yml` including the `contains_pii` gap you flagged, `accepted_values` severity).

GabyRangelB:
@claude - plz update your review

cbini:
Heads up — I've opened #4582, a **stacked follow-up** based on this branch (not on `main`), rather than pushing changes directly here. This PR is unaffected and merges first.

It restructures three things from this work:

- **Package placement** — the source-cleaning logic in `stg_finalsite__status_report` moves down into the `finalsite` package, and the Focus enrollment derivation moves into the `focus` package (kipptaf keeps thin wrappers for the cross-source joins). The three kipptaf `stg_focus__*` passthroughs go away with it.
- **The scaffold mode switch** — `finalsite_scaffold_source` and the `raise_compiler_error` guard are removed; the model becomes the blend pipeline as plain SQL. `finalsite_recruitment_year` stays as-is.
- **The `grade_level = -1` collision** — PK keeps `-1`; whole-school-total rows become `-9`, recoded where the sheets enter the graph. The sheets themselves keep their `-1` convention, so nothing changes for Ops.

It also carries three fixes from a review of this PR: `region` added to the blend anti-join key, the reference doc corrected to describe the recruitment-year var rather than the hardcoded-literal design it replaced, and the sentinel split above.

**One thing that needs your call before either merges:** the Tableau workbook filters on `grade_level = -1` for school-total rows and will need to move to `-9` in the same window. That one fails silently — the filters just return nothing rather than erroring — so it's worth confirming with SRE explicitly rather than catching it after the fact.

Full detail, including validation numbers and merge ordering, is in #4582.

GabyRangelB:
> Heads up — I've opened #4582, a **stacked follow-up** based on this branch (not on `main`), rather than pushing changes directly here. This PR is unaffected and merges first.
> 
> It restructures three things from this work:
> 
> * **Package placement** — the source-cleaning logic in `stg_finalsite__status_report` moves down into the `finalsite` package, and the Focus enrollment derivation moves into the `focus` package (kipptaf keeps thin wrappers for the cross-source joins). The three kipptaf `stg_focus__*` passthroughs go away with it.
> * **The scaffold mode switch** — `finalsite_scaffold_source` and the `raise_compiler_error` guard are removed; the model becomes the blend pipeline as plain SQL. `finalsite_recruitment_year` stays as-is.
> * **The `grade_level = -1` collision** — PK keeps `-1`; whole-school-total rows become `-9`, recoded where the sheets enter the graph. The sheets themselves keep their `-1` convention, so nothing changes for Ops.
> 
> It also carries three fixes from a review of this PR: `region` added to the blend anti-join key, the reference doc corrected to describe the recruitment-year var rather than the hardcoded-literal design it replaced, and the sentinel split above.
> 
> **One thing that needs your call before either merges:** the Tableau workbook filters on `grade_level = -1` for school-total rows and will need to move to `-9` in the same window. That one fails silently — the filters just return nothing rather than erroring — so it's worth confirming with SRE explicitly rather than catching it after the fact.
> 
> Full detail, including validation numbers and merge ordering, is in #4582.

@cbini - thanks! I am actually in the processes of deprecating the scaffold gsheet and leveraging powerschool schools, focus schools, and finalsite to generate the scaffold. i will also swap the grade level to -9 on the goals sheet for the all gl once these changes are merged. will merging 4582 affect the work im currently doing in the non focus data models? i.e. the models that end up creating the tableau reporting view?

cbini (dismissed):

cbini:
Short answer: yes, it will touch those models — #4582 is mostly *in* the non-Focus path, not beside it. Worth syncing before you go further, because we may be building the same thing.

**What #4582 changes in the scaffold path**

- `int_finalsite__enrollment_scaffold` and `int_finalsite__goals_scaffold` are renamed and moved to `int_tableau__fresh_enrollment_scaffold` and `int_tableau__fresh_goals_scaffold` under `extracts/tableau/intermediate/`. They were blending PowerSchool + Focus + Sheets while filed in the Finalsite source vertical and referencing no Finalsite model.
- Both `rpt_tableau__fresh_dashboard_aggregated` and `rpt_tableau__fresh_dashboard_progress_to_goals` are updated for the renames, and the region-mismatch singular test is renamed with them.
- The scaffold already derives grade membership from three sources: PowerSchool for the NJ regions, Focus for Miami, and the sheet only for what neither SIS can supply (whole-school rows and net-new grades).

So if you have edits in flight on `int_finalsite__enrollment_scaffold`, that file no longer exists on my branch. Easiest path is probably for me to hold, you land what you have, and I rebase — or we split it explicitly. Your call; I'd rather not make you resolve a rename conflict.

**One thing worth checking in your approach**

You mentioned generating the scaffold from PowerSchool schools, Focus schools, and Finalsite. Careful with the *schools* side — #4488 hit this on the PowerSchool half and switched to enrollment actuals because `low_grade`/`high_grade` encodes a school's eventual grade span, not what it currently serves, which produced phantom rows for growing schools.

Focus has the same problem. Comparing `stg_focus__school_gradelevels` against actual active 2026 enrollment:

| School | Configured but not enrolled |
| --- | --- |
| Sunrise | KG, 01-05 |
| Liberty | 05-08 |
| Courage | 04, 05 |
| Miami Tech | 10, 11, 12 |
| Legacy MS | 07, 08 |

Sunrise and Liberty have no 2026 enrollment at all, and Miami Tech and Legacy MS are still growing into their spans. Driving the spine off school config would put all of those into the scaffold as rows with goals attached. Deriving from enrollment actuals instead (`int_focus__student_enrollment` with `enroll_status = 0`) reproduces the current sheet spine exactly — Courage 6-8, Royalty K-5, Miami Tech 9, Legacy ES K-5, Legacy MS 6 — which is what #4582 does today.

**Heads up on the sheets and prod**

Both sheets now read `-9` for the whole-school rows. Because the deployed code still filters `-1`, prod's FRESH views have dropped their school-level rows — `rpt_tableau__fresh_dashboard_progress_to_goals` is at 23,721 rows against 47,073 before. It fails silently; nothing errors, the rows just stop matching.

#4582 restores it, but it merges behind this PR, so the dashboard is wrong until both land. If SRE is looking at school-level numbers this week they should know. If it needs to be right sooner, reverting the sheets to `-1` restores prod immediately and I can re-sequence the code side.

cbini (approved):

PR #4501, opened 2026-07-22
docs(cube): reframe partner integration as shape-first, not API access

### LINKED ISSUE #4455: feat(cube): API-key access layer for external / programmatic clients
## Summary

Design and add an API-key access layer to the Cube semantic layer so external / programmatic clients (developers, partner integrations) can query Cube without an `apps.teamschools.org` staff identity — and without ever holding the deployment master secret.

This issue is a **handoff document for a fresh brainstorming session**. It captures the current architecture, the core insight, a candidate shape, and the open design questions. Nothing here is decided — the brainstorming session should resolve the questions below and produce a spec in `docs/superpowers/specs/`.

## Why

Today there is no way to grant a non-staff, programmatic client scoped access. Access is derived entirely from an `email` claim resolved against HR data, so:

- An external email that is not in `dim_staff_cube_access` authenticates fine but resolves to the empty default-deny context and sees zero rows on every gated view.
- The only lever is which email a token claims, and that email must already be a staff row with the desired scopes.
- Handing an external party `CUBEJS_API_SECRET` is unacceptable: it is the master key for the REST path — a holder can mint a token for any email, including a network-scoped full-PII staff account, and read all student and staff PII. There is no per-token restriction on the REST JWT path.

## Current architecture (post PR #4269, grounded in code)

Auth and RLS after the security redesign:

- `checkAuth` (REST/MCP) receives the raw bearer token string, verifies the HS256 signature against `CUBEJS_API_SECRET`, reads the `email` claim, and sets the request security context to `resolveAccess(email)`.
- `checkSqlAuth` (SQL API) resolves identity from the connecting user.
- `resolveAccess(email)` reads one row from `kipptaf_marts.dim_staff_cube_access` plus reportees from `dim_staff_reporting_chain`, precomputes location-and-department allow-lists, and calls `access.buildSecurityContext(...)`. Fail-closed to an empty context on any error. Cached per-email until next midnight ET.
- `contextToGroups` is a one-line read of `securityContext.groups`.
- RLS lives entirely in per-view `access_policy` blocks that interpolate `securityContext` values. `queryRewrite` retains only the snapshot-anchor guard.

Key files:

- `src/cube/cube.js` — auth hooks, identity reads, cache.
- `src/cube/access.js` — pure, unit-tested access logic (`buildGroups`, `buildSecurityContext`, `computeAllowedAbbreviations`, `computeAllowedDepartmentGroups`).
- `src/cube/CLAUDE.md` — sections "View access policies" and "cube.js security model".

## Core insight

Everything downstream of authentication is identity-agnostic. The pipeline is `checkAuth` then `resolveAccess` then `buildSecurityContext` then `contextToGroups` then per-view `access_policy`. The view policies do not care how the security context was produced.

So an API key is just a **second resolver** that turns a key into the same security-context shape. Every existing view policy then works for key-authenticated clients with no view changes.

## Candidate shape (to validate, not decided)

1. A client registry — new BigQuery mart or dbt seed (e.g. `dim_cube_api_clients`) mirroring the scope columns of `dim_staff_cube_access` (`student_location_scope`, `staff_pii_scope`, `region_key`, `location_abbreviation`, `department_group`, ...), plus a hashed key column, client name, active flag, expiry. Because it carries the same scope enums it feeds the existing `computeAllowed*` / `buildGroups` pipeline unchanged.
2. A `resolveApiKey(key)` function in `cube.js`, sibling to `resolveAccess`: hash the presented key, look up the row, build the same security context, fail-closed on miss/inactive/expired.
3. A branch in `checkAuth`: detect an API key by prefix convention and route to `resolveApiKey`; otherwise fall through to the existing JWT path.
4. Key-management tooling: generate, hash, store, revoke.

## Open design questions for the brainstorming session

1. **What should external clients be allowed to see? (biggest scope driver.)** Today any `student-*` group sees every field including PII, with only row-level location scoping. There is no aggregate-only or non-PII tier — the one aggregate view was removed pending small-cell suppression (#4237). External clients almost certainly need a new restrictive view plus a new scope tier. This is likely the main body of work, not the key plumbing.
2. Where to store the key-to-scope mapping: BigQuery mart vs dbt seed vs a secrets store — and the tradeoffs for update latency.
3. Cache and revocation: current cache is midnight-ET. External keys probably need revocation in minutes, so a shorter TTL or explicit cache-bust.
4. Key format and lifecycle: prefix convention, hashing at rest, rotation, expiry, revocation.
5. Surface scope: REST/MCP only for v1, or SQL API too? (SQL path would need the key in the password slot — awkward.)
6. Audit logging per client — does not exist today; likely required for any data leaving KTAF.
7. Rate limiting / abuse controls — what does Cube Cloud offer per-identity?
8. Governance: policy / legal sign-off before sending KTAF data to an outside party.

## Invariants to preserve

- Cubes private, views public; fail-closed default-deny.
- No change to existing internal identity resolution or view policies.
- Never store raw keys; treat keys as bearer credentials (HTTPS only).
- Do not broaden `canSwitchSqlUser` (the SQL super-user impersonation guard).
- A leaked key's blast radius must be limited to that key's own scope.

## Related

- PR #4269 — the HR-derived security redesign this builds on.
- #4237 — small-cell suppression (blocks any aggregate demographic exposure).
- #4268 — queryRewrite member-strip (relevant to how out-of-tier fields are handled).
- #4342 — post-#4269 follow-ups.

## Suggested first step

Run a `superpowers:brainstorming` session against the questions above, then write a design spec to `docs/superpowers/specs/`.

cristinabaldor:
# Pull Request

## Summary & Motivation

...replace the previous external API-key design spec with a **shape-first
integration design**, **publish the semantic-layer catalog** so a partner team can
start building immediately, and record one empirically confirmed Cube Cloud
configuration trap in `src/cube/CLAUDE.md`.

No Dagster, dbt, cube model, schedule, or sensor changes. One new script.

### The reframe

The partner (MasterBorn) is a **software development agency, not a data
recipient**. The previous spec answered "how does an outside party read our
warehouse." The real question is "how does a team build software against our
semantic layer."

So what they get is **the shape**, not the data:

1. A machine-readable catalog of the semantic layer their code can consume.
1. A behaviorally faithful sandbox holding **zero real records**.
1. A repoint at production marts once the product is built.

**Zero real-data egress, ever** — not just in v1. See the next section.

### The partner confirmed their end users are always KTAF staff

This is the load-bearing fact, and it simplifies the whole design:

- **No new scope model.** Every end user already has a `dim_staff_cube_access`
  row, so the existing HR-derived derivation covers production completely.
- **No new entitlement.** The product shows a staff member data they can already
  reach through Cube today. It changes the interface, not the access.
- **Production identity pass-through is in scope**, not deferred. It was blocked
  only on this question.
- **The de-identified mirror is cut**, not deferred. Its only unique job was
  giving realistic data to a tier that could not be shown PII, and no such tier
  exists.
- **Audit becomes a complete trail.** The accountable party is a staff member we
  already know, and the `surface` field separates a query made through the
  partner's product from one made through Tableau, Superset, or `claude.ai`.

### What it introduces — the part most worth reviewing

Three risks move from our side of the boundary to the partner's, and **none are
enforceable by our access policies**, so each has to be a written contract term:

- **Offline and background queries.** Any feature running while the user is
  absent has no live identity to pass, so it needs a service identity —
  reopening the exact scope model this design avoids.
- **Result caching across users.** A screen cached for a network-scoped director
  and re-served to a school-scoped principal bypasses per-identity filtering in a
  way we cannot see.
- **Token handling in their application.** Pass-through means their product holds
  KTAF staff session tokens, so lifetime, storage at rest, and revocation on
  logout join the security review.

It also **sharpens** one existing constraint rather than relaxing it: the
whole-query denial in #4268 matters more now, because the product serves a
school-based AP and a network-office director on the same screen. That makes the
fidelity rule's narrow personas load-bearing rather than cautious.

### The catalog — generated and published

Generated from the live model, so a partner has something actionable before any
sandbox exists. **6 views, 40 measures, 295 dimensions. No data in any artifact.**

Both artifacts live under `docs/reference/` so the MkDocs site serves them at a
stable URL a partner can fetch:

| Artifact | Purpose |
| --- | --- |
| `cube-catalog-meta.json` | Machine-readable snapshot — what a consumer parses |
| `cube-semantic-catalog.md` | How to query, response envelope, gotchas, samples, per-view reference |

`scripts/cube_catalog_export.py` regenerates both, splicing only the `## Views`
section so the hand-written front matter survives. **Commit its output** — a model
change then surfaces as a reviewable diff, which is how the partner learns the
surface moved without being told.

Two safeguards in the script:

- **Exits non-zero on an empty catalog.** Cube hides every view whose access
  policy the caller's groups do not match, so an access problem and an undeployed
  model both return zero views — and only one is worth writing to disk. Run it as
  a network-scoped identity.
- **First-sentence truncation skips known abbreviations**, so a description
  containing "e.g." is not cut mid-clause. Full text stays in the JSON.

`staff_pii` is listed last with a warning admonition — present so an integrator
sees the view exists and is deliberately outside their scope, rather than finding
a silent gap.

Two of the ten gotchas were found by **calling the API**, not reading the model:
numeric measures come back as JSON strings (`"900"`, not `900`), and the echoed
`query` is normalized rather than a copy of what was sent.

> **Note:** the published URLs go live only when this merges — `mkdocs-gh-deploy`
> runs on push to `main`. The draft reply says so explicitly.

### The partner reply

`docs/superpowers/specs/2026-08-06-cube-partner-reply.md` — their four questions
answered inline with the verbatim text quoted back, the three obligations above
restated as contract terms rather than assumptions, the phased plan for their
approval, and a runnable `curl`. Kept under `docs/superpowers/` so it reviews
alongside the design without publishing to the site.

The spec itself no longer narrates how the design was reached — the sections that
argued toward decisions were removed once the decisions were settled, since a
reader who was not in the room only needs the outcome.

### Other key design points

- **The authorization derivation is already a callable service.** The Cloud Run
  token-exchange service verifies an OIDC identity and returns a short-lived
  token; scope is derived server-side from HR data. The partner passes **an
  identity, never a scope**, and never holds the signing credential.
- **The sandbox is a separate Cube Cloud deployment**, not a branch environment.
  Every model reference is project-unqualified, so one connection variable
  repoints the entire model with **no YAML change**.
- **Its service account gets no IAM on `teamster-332318`**, so a full compromise
  of the sandbox reaches zero real records. Stronger than any policy-level
  control, because it does not depend on the policies being correct.
- **Synthetic data comes from a parameterized generator with a seeded RNG**, not
  hand-authored seeds. Volume was never the de-identified mirror's advantage —
  nothing forces synthetic data to be small.
- **Cube Cloud's built-in audit log covers administrative events, not data
  access** (30-day retention, 10,000-event cap). The data-access trail is ours to
  build.

### The CLAUDE.md addition

Confirmed in the Cube Cloud console: the deployment generates **one signing
credential shared by every environment**, while connection variables are
**per-environment**. The natural inference — "a branch has its own variables, so
a branch is isolated, so it is safe to expose" — is true for the data connection
and false for the credential.

Consequence: a token minted against a branch environment is honored by
production, so the branch route is closed for any external consumer. Documented
because getting it backwards hands out a production key.

### Still needing its own spec

One item, a fallback rather than planned work: if the synthetic generator proves
insufficient, a de-identified mirror would need a stated re-identification-risk
threshold — the same governance decision blocking #4237.

## AI Assistance

Brainstormed and drafted with Claude Code (`superpowers:brainstorming`). The
reframe, the sandbox-runtime choice, the de-identified-default posture, and the
decision to cut the mirror were human-directed by @cristinabaldor; Claude grounded
them in the current `cube.js` / `access.js` / view-policy code, measured the
synthetic-data surface against `kipptaf_marts` (20 tables, 229 columns, zero
nested), generated the catalog from the deployed model, and researched Cube Cloud
Enterprise capabilities against the vendor docs.

The two configuration facts about credential and variable scoping were **verified
by @cristinabaldor in the Cube Cloud console**, not inferred from documentation —
the vendor docs are silent on both.

### General

> `claude-code-review.yaml` excludes markdown but **does** cover `scripts/`, so
> this PR now has a Python file in scope and an auto-review is expected.

N/A — no Dagster changes.

N/A — no dbt changes.

N/A for schedules and sensors. `docs/reference/cube-semantic-catalog.md` is added
to the `mkdocs.yml` nav under Reference; `mkdocs.yml` validated as parseable after
the edit. The specs under `docs/superpowers/` stay outside the nav by design.

## CI checks

- [ ] **Trunk** — passes. Verified with `trunk check --force --no-fix` on every
      changed file: markdownlint clean after converting the generated
      `**Measures**` / `**Dimensions**` lines to real `####` headings per
      `docs/CLAUDE.md`, and the generator was fixed so regeneration does not
      reintroduce them.
- [ ] **dbt Cloud** — not meaningful for this change.
- [ ] **Dagster Cloud** — not expected to run (no Dagster paths, and draft).

## Troubleshooting

Refs #4455


cristinabaldor:
Superseded by #5267 — closing.

Every file from this branch is on `cristinabaldor/feat/claude-cube-sandbox-build` and reviewed there instead. Nothing is dropped: the catalog, the semantic-catalog page, `scripts/cube_catalog_export.py`, the partner reply, and this design doc all moved across intact.

Why consolidate rather than land both: the sandbox build spec on #5267 is the direct child of this design's Deliverable 2, and it changes one of its decisions. Reviewing that decision chain across 2 open PRs meant reading a cross-reference between branches, and this branch had drifted into conflict with `main` in the meantime.

Three things changed during the move, all worth knowing if you had read this branch already:

**The `src/cube/CLAUDE.md` conflict is resolved in main's favor.** `main` relocated most of that file into `.claude/rules/cube-authoring.md` and the `cube-ops` skill after this branch was cut. The resolution takes main's slim version and re-adds only the note this branch genuinely contributed — that `CUBEJS_API_SECRET` is deployment-wide while every other variable is per-environment, so a branch-minted token is mechanically a production credential. That fact exists nowhere else in `main`.

**Deliverable 2 now opens with a pointer to the child spec,** and the superseded "the sandbox deployment tracks `main`" line is marked in place. The sandbox deploys from a `sandbox-YYYY.MM.DD` tag instead, so a model change merged mid-sprint cannot break the partner's build unannounced.

**Three counts in this design were measured and corrected** on #5267: 230 columns rather than 229 (all nullable), 7 scope columns rather than 6, and `dim_staff_reporting_chain` is the 20th table but appears in no cube YAML — only `cube.js` reads it, so anything deriving the table set from `sql_table:` finds 19 and misses it silently.

Refs #5266, #4455

Issue #3142, opened 2025-12-04
perf: add region col to all kipptaf union views


cbini:

cbini:
Partial progress in PR #4024 (assessment domain).

### `_dbt_source_project` materialized on union models

- `int_assessments__scaffold` — projected in all three union legs (`concat('kipp', lower(region))` for internal, replacement K-8, and external assessment branches).
- `int_assessments__response_rollup` — carried through scaffold_responses → tiebroken_attrs (`first_value` over the canonical-grain window) → internal_assessment_rollup (`any_value`) → response_union → final SELECT.
- `int_pearson__all_assessments`, `int_fldoe__all_assessments` — already had `_dbt_source_project` (sourced from staging via `extract_code_location()`).

### Hash slot migrations (`region` → `_dbt_source_project`)

- `dim_assessment_administrations.assessment_administration_key`
- `fct_assessment_scores_enrollment_scoped.assessment_administration_key`
- `bridge_assessment_expectations_enrollment_scoped.assessment_administration_key`
- `bridge_assessment_expectations_student_scoped.assessment_administration_key`
- `bridge_assessment_administration_members.assessment_administration_key`

### Pattern documented

`marts/CLAUDE.md` gained a section on `_dbt_source_project` joins and hashes:

> When a marts fix touches joins or surrogate-key composition involving `_dbt_source_project` (or `_dbt_source_relation`), promote the `extract_code_location()` call up to the union model itself rather than applying it at each consumer. Downstream consumers should join and hash on the materialized `code_location` column.

### Still open

The assessment domain is now on `_dbt_source_project`, but other mart families (`dim_courses`, `dim_student_section_enrollments`, etc. — see #3820) still extract from `_dbt_source_relation` or use the full-relation workaround. #3142 stays open until the full sweep lands.

PR #4502, opened 2026-07-22
refactor(dbt): kickoff #3142 source-project sweep (spec, plan, macro rename)

cbini:
## Summary and Motivation

When merged, this PR kicks off completion of #3142 (replace `union_dataset_join_clause` with a materialized `_dbt_source_project` column). It adds:

- The design spec and implementation plan under `docs/superpowers/`.
- Batch 1a: a mechanical, behavior-preserving rename of the `extract_code_location` macro to `extract_source_project`, now with an optional `relation` arg (default unqualified). `union_dataset_join_clause` is repointed at the new name (it is removed entirely in the final batch).

Verified byte-identical via `dbt compile` — both the producer arg-form and the consumer join-clause path render the same `regexp_extract(...)`. Also normalizes one CRLF file to LF.

Subsequent batches (producer backfill, consumer swaps, macro removal) stack on this branch as separate PRs per the plan.

## AI Assistance

Authored by Claude Code (Opus 4.8), human-directed. The human reviewer decided the design: optional-arg vs argument-less macro, the iReady/renlearn/amplify crosswalk carve-out, convention-only enforcement, and the stacked-PR structure.

### dbt

- [x] Rename is behavior-preserving — no model logic, columns, or contracts changed
- [x] No new models (docs plus macro rename only)
- [x] No external-source changes

## CI checks

- [ ] Trunk
- [ ] dbt Cloud

Refs #3142

### DIFF
--- a/src/dbt/kipptaf/models/students/intermediate/int_students__fldoe_fte.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_students__fldoe_fte.sql
@@ -16,7 +16,7 @@ with
             and att.calendardate between fte.start_date and fte.end_date
             and fte.type = 'FTE'
         where
-            {{ extract_code_location("att") }} = 'kippmiami'
+            {{ extract_source_project("att") }} = 'kippmiami'
             and att.membershipvalue = 1
             and att.attendancevalue = 1
         group by att._dbt_source_relation, att.studentid, att.yearid, fte.name
--- a/docs/superpowers/plans/2026-07-22-dbt-source-project-completion.md
+++ b/docs/superpowers/plans/2026-07-22-dbt-source-project-completion.md
@@ -1,0 +1,446 @@
+# `_dbt_source_project` Completion Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use
+> superpowers:subagent-driven-development (recommended) or
+> superpowers:executing-plans to implement this plan task-by-task. Steps use
+> checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Finish #3142 — every `union_relations` view exposes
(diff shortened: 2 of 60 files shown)


src/dbt/kipptaf/models/students/intermediate/int_students__fldoe_fte.sql
```diff
@@ -16,7 +16,7 @@ with
             and att.calendardate between fte.start_date and fte.end_date
             and fte.type = 'FTE'
         where
-            {{ extract_code_location("att") }} = 'kippmiami'
+            {{ extract_source_project("att") }} = 'kippmiami'
```
cbini, line 19: refactor after extract_source_project added to int_powerschool__ps_adaadm_daily_ctod

cbini: Good catch — and the precondition is already satisfied: `int_powerschool__ps_adaadm_daily_ctod` already materializes `_dbt_source_project` (its line 57), so `att` exposes it here.

This `{{ extract_source_project("att") }} = 'kippmiami'` is a **consumer-side** use (a region filter), so it's slated for the Batch 2 consumer-swap pass rather than this mechanical-rename PR — the plan's Batch 2f explicitly lists this file plus `qa_edplan__powerschool_mismatch` as the two non-producer sites to convert to plain `_dbt_source_project` refs.

Target: `where att._dbt_source_project = 'kippmiami'`, which also drops a row-level `regexp_extract` out of the `WHERE` (a small SQL-guide compliance win).

Kept as just the rename here so 1a stays a pure behavior-preserving rename; it swaps alongside the other consumer sites in Batch 2. Happy to pull only this one forward now if you'd rather see it in this PR.

PR #4510, opened 2026-07-23
fix(google): retry transient errors on individual batch sub-requests in GoogleDirectoryResource

### LINKED ISSUE #4509: fix(google): retry transient errors on individual batch sub-requests in GoogleDirectoryResource
## Summary

`GoogleDirectoryResource` batch methods do not retry transient (429 / 5xx) errors that occur on an *individual sub-request* within a batch. A transient `HttpError 503 Backend Error` on one user-create was recorded as a permanent error and failed the `zero_api_errors` asset check, firing an alert — even though the same request would very likely have succeeded on retry.

## Root cause

`batch_insert_users` (and the other three batch methods) wrap only the outer batch call in the retry layer:

```python
backoff(fn=_retryable_execute(batch_request), retry_on=(_TransientHttpError,))
```

`_retryable_execute` re-raises a transient `HttpError` as `_TransientHttpError` so `backoff` retries it. But with a `BatchHttpRequest`, an individual sub-request failure is **never raised** out of `execute()`. Per `googleapiclient/http.py` `BatchHttpRequest.execute()`:

- The outer batch envelope raises only when the whole batch POST fails (`uri=self._batch_uri`).
- Each sub-request result is delivered to `callback(id, response, exception)`; a sub-request whose status is 300 or above is passed as `HttpError(resp, content, uri=request.uri)`.

So a 503 on one `users().insert()` goes to `callback`, which appends it to `self._exceptions`, which the asset returns as an error, which fails the `zero_api_errors` check (WARN) and fires the alert. `backoff` never sees it.

The single-request `_list` path retries correctly because its `.execute()` does raise `HttpError`.

Evidence: the alert's error URI was the individual users endpoint (`.../directory/v1/users?alt=json`), not the batch endpoint — confirming it came through the per-sub-request callback path.

## Affected methods

All four share the blind spot, in `src/teamster/libraries/google/directory/resources.py`:

- `batch_insert_users`
- `batch_update_users`
- `batch_insert_members`
- `batch_insert_role_assignments`

## Impact

- Self-limiting: `is_create` in `rpt_google_directory__users_import` is derived from directory presence, so a failed create retries on the next daily `user_sync` run (roughly a one-day delay in account provisioning for that student).
- Spurious `zero_api_errors` WARN alerts on any transient Google blip.
- Secondary fragility: `google_directory_user_create` calls `batch_insert_members` for all valid users regardless of whether their create succeeded, so a create that failed transiently also produces a member-insert error in the same run. Noted here as a follow-up; not part of the core retry fix.

## Fix

Retry at sub-request granularity via a shared helper used by all four batch methods:

1. Execute the batch, collecting per-sub-request outcomes via the callback.
2. Re-submit only the sub-requests that failed with a transient code (429 / 5xx) in a follow-up batch, with backoff, up to a max attempt count.
3. Do not re-submit the whole batch — already-succeeded inserts would return 409.
4. Preserve method-specific behavior: `batch_update_users` individual 409 retry, `batch_insert_role_assignments` exponential delay.

## Testing

TDD: add a failing test asserting that a batch whose sub-request returns a transient 503 on the first attempt and 200 on retry produces zero errors (today it produces one). Extend the existing mock harness in `tests/resources/test_resource_google_directory.py`.

cbini:
# Pull Request

## Summary & Motivation

When merged, this fixes spurious `zero_api_errors` alerts (and roughly a
one-day delay in account provisioning) from `GoogleDirectoryResource` batch
methods.

A `BatchHttpRequest` never raises for an individual sub-request failure — the
result is delivered to the callback and `execute()` returns normally — so the
existing `backoff(...)` wrapper only retried a failure of the whole batch
envelope. A transient 503 "Backend Error" on a single `users().insert()` was
therefore recorded as a permanent error, failing the `zero_api_errors` asset
check on `kipptaf/google/directory/user_create` and firing an alert, even
though a retry would very likely have succeeded. (Observed on prod run
`df2f8bec` for one student create.)

This adds a shared `_execute_batch_with_retry` helper that re-submits only the
transiently-failed (429 / 5xx) sub-requests in a follow-up batch with
exponential backoff, up to a bounded attempt count, and wires it into all four
batch methods: `batch_insert_users`, `batch_update_users`,
`batch_insert_members`, `batch_insert_role_assignments`. Already-succeeded
sub-requests are not re-sent (re-sending would return 409). The batch callback
now logs failures at WARNING instead of `exception`, so transients the retry
layer recovers from no longer file false-positive GCP Error Reporting groups.

Method-specific behavior is preserved: `batch_update_users` still retries 409
conflicts individually; a role-assignment 409 ("already exists") is still
treated as non-retryable.

**Also folded in (a follow-up noted on #4509):** the create asset built its
group-membership list from every attempted user, so a user whose create failed
still got a `members().insert` — guaranteed to fail (no account yet) and
double-counting the same student in the check. `batch_insert_users` now returns
structured per-user errors (`{"primaryEmail", "error"}`) instead of error
strings; a new library helper `members_for_created_users` builds membership
payloads only for users that were actually created; and the asset skips the
members call when none remain. The structured error form also keeps the create
payload (which includes the password hash) out of logs and check metadata — the
same payload that leaked into the alert that opened this issue.

Closes #4509.

## AI Assistance

Authored by Claude Code (Opus 4.8) via systematic-debugging plus TDD, directed
by @cbini. The root cause was verified against the `googleapiclient`
batch-execute source; the fix is covered by unit tests written test-first (each
new test was confirmed failing before the implementation landed).

### Dagster

- [x] Shared library resource + asset change, covered by mocked unit tests in
      `tests/resources/test_resource_google_directory.py` — 36 pass (5
      sub-request-retry, 1 no-payload-in-error, 3 `members_for_created_users`);
      no regressions. `dagster definitions validate` for kipptaf does not run in
      the codespace (credential specs resolve eagerly at import), but the
      library module imports cleanly, the asset module byte-compiles, and every
      mocked resource test passes.
- [x] No new integration; existing resource, Library + Config pattern
      unchanged.

## CI checks

- [ ] **Trunk** — `trunk check --force` on both changed files reports no issues
      locally; confirm the CI check.
- [ ] **dbt Cloud** — no dbt changes; expected trivial no-op.
- [ ] **Dagster Cloud** — confirm the branch deploy / deploy passes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -82,6 +82,13 @@ file; domain specifics live in the nearest subdirectory CLAUDE.md.
   path breaks — pass an absolute script path or run it from the main repo.
   Otherwise prefer absolute paths.
 
+- **Bash cwd does NOT persist across calls** — every Bash command (including
+  `run_in_background`) starts at the main repo root, so a prior `cd <worktree>`
+  does not carry over. Tools that resolve relative paths from cwd (`trunk check`
+  with relative paths, `pytest`) must include `cd <worktree> &&` in the SAME
+  command, or they silently operate on the main checkout's (unmodified) copies
+  and report a false "clean". Prefix with `pwd &&` to confirm the directory.
+
 - **Worktree Read/Edit/Write must target the worktree path**, not the main
   checkout: editing `/workspaces/teamster/<path>` instead of
   `/workspaces/teamster/.worktrees/<branch>/<path>` silently leaves the worktree
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -438,7 +438,9 @@ launcher. Package internals: see
 
 - **MCP outages**: If an MCP tool returns "server disconnected" or clearly
   impaired responses, surface to the user before working around with raw `gh` /
-  BigQuery calls.
+  BigQuery calls. Same if an EXPECTED MCP tool is absent from the deferred-tools
+  list (ToolSearch returns "No matching deferred tools") — flag it immediately
+  so the user can reconnect; do not silently fall back.
 
 - **MCP subprocess logs**: stdio MCP stderr captured at
   `~/.cache/claude-cli-nodejs/-workspaces-teamster/mcp-logs-<name>/<ts>.jsonl`.
--- a/src/teamster/CLAUDE.md
+++ b/src/teamster/CLAUDE.md
@@ -246,3 +246,10 @@ the codespace). For a change to either, `py_compile` the edited files and import
 the affected submodule alone (e.g.
 `import teamster.code_locations.kipptaf.finalsite` or
 `teamster.code_locations.kippmiami.extracts`), not the `definitions` module.
+
+Importing a kipptaf **asset submodule** (e.g.
+`teamster.code_locations.kipptaf.google.directory.assets`) transitively imports
+`kipptaf.dbt` and needs the dbt manifest — absent in a fresh worktree, so a unit
+test importing it errors at collection (`FileNotFoundError: .../manifest.json`).
+Put unit-testable pure helpers in `libraries/` (imports cleanly), not a
+code-location asset module.
--- a/src/teamster/code_locations/kipptaf/google/directory/assets.py
+++ b/src/teamster/code_locations/kipptaf/google/directory/assets.py
@@ -20,7 +20,10 @@ from teamster.core.asset_checks import (
     build_check_spec_avro_schema_valid,
     check_avro_schema_valid,
 )
-from teamster.libraries.google.directory.resources import GoogleDirectoryResource
+from teamster.libraries.google.directory.resources import (
+    GoogleDirectoryResource,
+    members_for_created_users,
+)
 
 key_prefix = [CODE_LOCATION, "google", "directory"]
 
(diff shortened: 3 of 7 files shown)


cbini:
Reviewed both passes and verified each item before acting. Fixes in `ce7935f61`.

## Implemented

- **Shadowed `errors` name** (`batch_insert_users`): renamed the accumulator to `exceptions` — it shadowed the module-level `googleapiclient` `errors` import and diverged from the three sibling batch methods.
- **Mixed-batch retry coverage**: added a test with a 2-item batch (item 0 succeeds, item 1 gets a transient 503) asserting the first batch is built with 2 `add()` calls and the retry batch with exactly 1 — so a succeeded create is never re-sent (the 409-on-recreate case).
- **Non-transient short-circuit**: added `assert new_batch_http_request.call_count == 1` to the 409 test, locking in that 4xx sub-requests aren't retried at the batch level.
- **Callback log (Info)**: replaced the raw `(id, exception)` tuple with an f-string.

## Not changing — the premise doesn't hold (verified from source)

- **"Envelope `backoff` can't retry; a second `execute()` raises `BatchError` (executed twice)"** — this pinned `googleapiclient` has no re-execution guard. The suggested check confirms it:

  ```text
  _executed present: False
  twice-guard present: False
  ```

  There is no `_executed` attribute and no "executed twice" `BatchError` in `http.py` (the only `BatchError`s are Content-ID / media / multipart-format). A second `execute()` re-runs `_execute` and re-POSTs the batch, so the envelope-level retry works as the comment describes. Leaving the comment as-is.

- **"role-assignments envelope retry downgraded exponential to fixed delay"** — dagster's `backoff` defaults `delay_generator` to `exponential_delay_generator()` when omitted (`dagster/_utils/backoff.py`), with the same `base=0.1, factor=2.0` the old explicit call passed. The envelope path is still exponential; the old argument was redundant. The sub-request loop instantiates its own generator, so there is no shared-generator-state concern either.

## No change (correctly Info)

- `request_factory: Callable[[dict], object]` — `object` is the pragmatic type for these SDK request objects.

37 mocked tests pass; `trunk check --force` clean on the changed files.

PR #4516, opened 2026-07-23
refactor(dbt): add _dbt_source_project to join-target intermediates (#3142)

cbini:
## Summary and Motivation

When merged, this PR completes **Wave 1** of the #3142 migration from the `union_dataset_join_clause` macro to a materialized `_dbt_source_project` column.

It adds `_dbt_source_project` to the output of 10 intermediate models that are join targets for downstream consumers but did not yet expose the column, and swaps those models' own internal `union_dataset_join_clause` calls to plain column equality (`a._dbt_source_project = b._dbt_source_project`).

The change is **behavior-preserving**: `extract_source_project()` derives exactly the same `regexp_extract(_dbt_source_relation, ...)` region code the macro produced.

This unblocks **Wave 2** (the reporting-leaf swaps), which need these intermediates to expose the column before they can join on it.

Models (all intermediate, no contract enforcement):

- `int_extracts__student_enrollments_subjects_weeks`
- `int_powerschool__final_grades_rollup`, `int_powerschool__gpa_term_lookback`, `int_powerschool__gpa_term_pivot`, `int_powerschool__gpnode`, `int_powerschool__gpprogress_grades`, `int_powerschool__s_nj_stu_x_unpivot`
- `int_students__athletic_eligibility`
- `int_tableau__gradebook_audit_student_scaffold`, `int_tableau__gradebook_audit_teacher_scaffold` (both `enabled: false` dead code; edited only so Batch 3 can delete the macro without a parse error)

Note: 3 files were CRLF on `main` and are normalized to LF here (sqlfmt does this at commit anyway), which inflates their line counts. **Review with whitespace hidden.**

## AI Assistance

AI-assisted (Claude Code), human-directed. The scope (which models need the column) was derived by resolving every macro call's join aliases against the prod schema; the 8 enabled models were built with `--target dev --defer` and confirmed to populate `_dbt_source_project` with real region codes. The 2 disabled scaffolds are parse-validated only.

## Self-review (dbt)

- [x] No new models; no properties/exposure changes needed (intermediate layer, no contract).
- [x] No contracted column renames/removals — purely additive column plus algebraically-equivalent join-predicate swaps.
- [x] No external sources changed.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.
- [ ] **Dagster Cloud** — passes or not triggered.

Refs #3142

PR #4530, opened 2026-07-24
refactor(dbt): swap union_dataset_join_clause for _dbt_source_project in tableau reports

cbini:
# Pull Request

## Summary & Motivation

When merged, this PR advances **Batch 2** (consumer join swaps) of the #3142 `_dbt_source_project` migration — the first directory-scoped PR of that batch.

Following Batch 1 (producer backfill, #4503–#4505) and the join-target intermediates (#4516, now merged and materialized in prod), this swaps the `union_dataset_join_clause` macro for the direct materialized-column equality `a._dbt_source_project = b._dbt_source_project` across the `rpt_tableau__*` report family.

The change is **behavior-preserving**: `_dbt_source_project` equals `regexp_extract(_dbt_source_relation, r'(kipp\w+)_')` by construction (it is derived by that exact expression at the union view), so the new predicate is algebraically identical to the macro it replaces — just without the per-join `regexp_extract`.

**Scope:** 33 files under `models/extracts/tableau/`, 124 macro calls swapped:

- **23 enabled (view) models** — column-resolution gate verified (see Self-review).
- **10 statically-disabled models** (`enabled: false` in properties.yml, unconditional — they never build in dev, staging, or prod) — mechanical swap only, parse-verified. Swapped now so Batch 3 can delete the macro with zero residual call sites. Same apply-but-unverifiable convention as the disabled scaffolds in #4516.

Where a consuming CTE selected explicit columns (it already projected `_dbt_source_relation` for the old macro but not `_dbt_source_project`), the column was threaded through alongside `_dbt_source_relation` from the same source — no new derivation, no `extract_source_project()` re-introduced downstream.

### Deferred to a follow-up PR

Two enabled models were left on the macro (reverted) because a join operand does not yet expose `_dbt_source_project` and needs a producer fix first:

- `rpt_tableau__ops_dashboard` — joins `int_finance__enrollment_targets`, a hand-built non-union intermediate that builds `_dbt_source_relation` via `concat`, so the column was never materialized on it.
- `rpt_tableau__state_assessments_dashboard` — joins `int_pearson__student_list_report`, which drops the column its `stg_pearson__student_list_report` source carries.

Both intermediates will get the column in a follow-up (analogous to the #4516 join-target work), after which their consumers can be swapped.

### Note on line counts

Four files (`rpt_tableau__crdc_roster`, `rpt_tableau__assignment_checks`, `rpt_tableau__college_assessment_dashboard_de`, `rpt_tableau__college_assessment_dashboard_historic`) were CRLF on `main` and normalize to LF here (sqlfmt does this at commit regardless), inflating their diffs. **Review with whitespace hidden.**

## AI Assistance

AI-assisted (Claude Code), human-directed. The swap is mechanical; the producer-gate verification (which CTEs needed the column threaded, which two models to defer) was driven by an `--empty` gate build against prod upstreams.

### dbt

- [x] No new models — no properties/exposure files needed. These are existing reporting views; the swap changes join predicates only, not model output columns.
- [x] **Not a breaking change** — no contracted column renamed or removed; join-predicate-only edits.
- [x] Column-resolution gate: all 23 enabled models pass `dbt build --empty --defer --favor-state --state <prod>` (validates every `_dbt_source_project` reference resolves against prod upstreams, no data scanned). Behavior equivalence holds by algebraic identity.
- [x] `trunk check --force` clean on all 33 files.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Refs #3142

### DIFF
--- a/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__academic_goals_rollup.sql
+++ b/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__academic_goals_rollup.sql
@@ -80,11 +80,11 @@ with
         inner join
             {{ ref("stg_powerschool__u_studentsuserfields") }} as suf
             on f.student_id = suf.fleid
-            and {{ union_dataset_join_clause(left_alias="f", right_alias="suf") }}
+            and f._dbt_source_project = suf._dbt_source_project
         inner join
             {{ ref("stg_powerschool__students") }} as s
             on suf.studentsdcid = s.dcid
-            and {{ union_dataset_join_clause(left_alias="suf", right_alias="s") }}
+            and suf._dbt_source_project = s._dbt_source_project
             and s.grade_level >= 4
         where
             f.assessment_name = 'FAST'
@@ -114,11 +114,11 @@ with
         inner join
             {{ ref("stg_powerschool__u_studentsuserfields") }} as suf
             on f.student_id = suf.fleid
-            and {{ union_dataset_join_clause(left_alias="f", right_alias="suf") }}
+            and f._dbt_source_project = suf._dbt_source_project
         inner join
             {{ ref("stg_powerschool__students") }} as s
             on suf.studentsdcid = s.dcid
-            and {{ union_dataset_join_clause(left_alias="suf", right_alias="s") }}
+            and suf._dbt_source_project = s._dbt_source_project
             and s.grade_level = 3
         where
             f.assessment_name = 'FAST'
@@ -299,7 +299,7 @@ with
             {{ ref("base_powerschool__course_enrollments") }} as cc
             on co.studentid = cc.cc_studentid
             and co.yearid = cc.cc_yearid
-            and {{ union_dataset_join_clause(left_alias="co", right_alias="cc") }}
+            and co._dbt_source_project = cc._dbt_source_project
             and co.illuminate_subject_area = cc.illuminate_subject_area
             and not cc.is_dropped_section
             and cc.rn_student_year_illuminate_subject_desc = 1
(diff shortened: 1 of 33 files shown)


cbini:
Verifications the review flagged — both re-run on this exact SHA (`5e1d17c34`):

- **`--empty` gate**, scoped to the 33 changed files: `PASS=25 WARN=0 ERROR=0` — every `_dbt_source_project` reference resolves against prod upstreams (23 enabled models + 2 tests; the 10 statically-disabled models are excluded by dbt, as expected).
- **`trunk check --force`** on all 33 files: no issues.

Note on the broader `path:models/extracts/tableau` form in your suggested command: it surfaces 2 errors, but both are in models this PR does not touch (`rpt_tableau__gradebook_audit`, `rpt_tableau__gradebook_es_comments`) and neither is a `_dbt_source_project` failure — they fail on a missing deferred upstream (`int_extracts__course_enrollments_by_term`, which postdates the local Jul-10 prod manifest). Pre-existing, unrelated to this swap.

**ST09 nit (`okrts_behavior:67`):** leaving as-is. The operand order (`w` left, `b` right) is carried over verbatim from the macro call's `left_alias="w", right_alias="b"`, so it is unchanged from `main`, and sqlfluff passes on the rendered SQL (trunk clean, 0 issues). Flipping it is a cosmetic edit outside this batch's mechanical scope — better left for a dedicated cleanup if wanted.

Deferred consumers (`ops_dashboard`, `state_assessments_dashboard`) and the eventual macro deletion are tracked in the PR body and #3142.

PR #4531, opened 2026-07-24
refactor(dbt): swap union_dataset_join_clause for _dbt_source_project in extracts reports

cbini:
# Pull Request

## Summary & Motivation

When merged, this PR continues **Batch 2** (consumer join swaps) of the #3142 `_dbt_source_project` migration — the second directory-scoped PR, covering the **extracts reporting layer**: gsheets, deanslist, clever, powerschool autocomm, and littlesis.

It swaps the `union_dataset_join_clause` macro for the direct materialized-column equality `a._dbt_source_project = b._dbt_source_project`. Behavior-preserving: `_dbt_source_project` equals `regexp_extract(_dbt_source_relation, r'(kipp\w+)_')` by construction, so the new predicate is algebraically identical to the macro.

**Scope:** 24 files, 50 macro calls swapped. All build-verified (see Self-review).

Where a consuming CTE selected explicit columns (it already projected `_dbt_source_relation` for the old macro), `_dbt_source_project` was threaded through alongside it from the same source — no re-derivation.

One contract-specific case: `rpt_deanslist__student_misc` needed `_dbt_source_project` threaded into its `co` CTE for the join, but its final `select co.*` would then emit the column, which is not in the model contract. Kept it out of the output with `select co.* except (_dbt_source_project)` (the same idiom used for `source_file_name` in staging) — the internal join key is used but not surfaced, so the contracted output is unchanged.

### Deferred to a follow-up PR

Two reports were left on the macro (fully reverted) because a join operand does not yet expose `_dbt_source_project` and needs a producer fix first:

- `rpt_powerschool__autocomm_students` — joins `int_students__graduation_path_codes`, which does not project the column.
- `rpt_deanslist__promo_status` — joins `int_powerschool__log`, which projects `_dbt_source_relation` only.

Both producers will get the column in a follow-up, after which these two consumers can be swapped atomically.

### Note on line counts

Two files (`rpt_gsheets__gpa_flags_report`, `int_google_sheets__dibels_pm_expectations`) were CRLF on `main` and normalize to LF here (sqlfmt does this at commit regardless), inflating their diffs. **Review with whitespace hidden.**

## AI Assistance

AI-assisted (Claude Code), human-directed. The swap is mechanical; column-threading and the two deferrals were driven by an `--empty` gate build against prod upstreams.

### dbt

- [x] No new models; the swap changes join predicates only. `rpt_deanslist__student_misc` keeps its contracted output identical (the threaded column is `except`-ed out).
- [x] **Not a breaking change** — no contracted column renamed or removed.
- [x] Column-resolution gate: the changed set passes `dbt build --empty --defer --favor-state --state <prod>` with 0 errors (every `_dbt_source_project` reference resolves against prod upstreams, no data scanned; contracts enforced under `--empty`). Behavior equivalence holds by algebraic identity.
- [x] `trunk check --force` clean (no new issues) on all 24 files.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Refs #3142

### DIFF
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__enrollments.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__enrollments.sql
@@ -10,7 +10,7 @@ from {{ ref("stg_powerschool__cc") }} as cc
 inner join
     {{ ref("stg_powerschool__students") }} as s
     on cc.studentid = s.id
-    and {{ union_dataset_join_clause(left_alias="cc", right_alias="s") }}
+    and cc._dbt_source_project = s._dbt_source_project
     and s.enroll_status in (0, -1)
 where
     cc.dateleft >= current_date('{{ var("local_timezone") }}')
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql
@@ -77,16 +77,16 @@ with
         inner join
             {{ ref("stg_powerschool__sectionteacher") }} as st
             on sec.sections_id = st.sectionid
-            and {{ union_dataset_join_clause(left_alias="sec", right_alias="st") }}
+            and sec._dbt_source_project = st._dbt_source_project
         inner join
             {{ ref("stg_powerschool__roledef") }} as r
             on st.roleid = r.id
-            and {{ union_dataset_join_clause(left_alias="st", right_alias="r") }}
+            and st._dbt_source_project = r._dbt_source_project
         inner join
             {{ ref("int_powerschool__teachers") }} as t
             on st.teacherid = t.id
             and sec.sections_schoolid = t.schoolid
-            and {{ union_dataset_join_clause(left_alias="st", right_alias="t") }}
+            and st._dbt_source_project = t._dbt_source_project
         where
             sec.terms_yearid = ({{ var("current_academic_year") - 1990 }})
             and sec._dbt_source_relation not like '%kipppaterson%'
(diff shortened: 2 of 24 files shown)


cbini:
CI green on `0523895c6` (Trunk / dbt Cloud / CodeQL / claude-review). No code changes — all three findings are optional; deferring them to a dedicated cleanup pass rather than churning this focused join-swap PR. Verified each against the files:

1. **regexp value-derivations** — not a one-file fix. Beyond `rpt_clever__sections:42`, the same `regexp_extract(_dbt_source_relation, r'(kipp\w+)_')` derivation is in `rpt_clever__enrollments:5` (both `_dbt_source_project`-equivalent). But `rpt_gsheets__csgf_hs_enrollment:11,138` uses a *different* regex — `kipp(\w+)_` initcap'd for a region label — which is NOT `_dbt_source_project`. A sweep has to distinguish the two, so it belongs in its own pass, not piecemeal here.
2. **`kippfwd_miami_roster` dead `gpa_term` CTE** — confirmed: `gpa_term` appears only at its definition (line 33), zero references. Pre-existing dead code (its line 34 also holds the only untouched `_dbt_source_relation` in the file). Folding into the cleanup pass.
3. **orphaned `_dbt_source_relation` selects** (`designations:4`, `kippfwd_miami_roster:4,16`) — confirmed unreferenced after the swap; harmless unused selections. Will drop with per-file verification in the same pass (most other models still use the column legitimately).

The behavior-equivalence invariant holds — every swapped target is a cross-district PowerSchool/FLDOE union, and the two non-conforming producers were deferred. The `--empty` gate covers resolution; the algebraic identity covers value parity.

PR #4541, opened 2026-07-24
refactor(dbt): expose _dbt_source_project on int_finance__enrollment_targets

cbini:
# Pull Request

## Summary & Motivation

Producer-intermediate wave of the #3142 `_dbt_source_project` migration (follows
#4516 / #4533 / #4537). Exposes `_dbt_source_project` on
`int_finance__enrollment_targets` so its downstream consumer can later be swapped
off the `union_dataset_join_clause` macro.

Consumer still on the macro (not touched here): `rpt_tableau__ops_dashboard` (its
`target_union` -> `t` alias, macro `se`/`t`).

## Approach

`int_finance__enrollment_targets` already synthesizes a `_dbt_source_relation`
from `regexp_extract(sch._dbt_source_relation, ...)`. This adds the matching
materialized `_dbt_source_project` as a pass-through from
`stg_powerschool__schools` (`sch._dbt_source_project`), which equals that same
region by construction — so the exposed column and the synthetic relation stay
consistent.

Also adds a `properties.yml` (the model had none): model description, a grain
uniqueness test, and the `_dbt_source_project` doc entry.

## Verification

- Built into dev against prod-deferred upstreams (reads the finance sheet live
  via ADC): the view creates and `sch._dbt_source_project` resolves.
- Grain confirmed empirically: 641 rows, `(academic_year, schoolid,
  grade_level)` is exactly unique (no fan-out on the `et.schoolid =
  sch.school_number` join), and `_dbt_source_project` is never null (3
  districts). The `dbt_utils.unique_combination_of_columns` test passes on real
  data.

## Scope note

The finance-modeling columns (`grade_band_ratio`, `at_risk_and_lep_ratio`,
`attrition_factor`, etc.) are listed name + type only, mirroring the staging
source yml (`stg_google_sheets__finance__enrollment_targets`), which carries no
column descriptions. Prose descriptions for those need finance-domain input and
are out of scope for this migration PR.

## AI Assistance

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Refs #3142

### DIFF
--- a/src/dbt/kipptaf/models/finance/intermediate/int_finance__enrollment_targets.sql
+++ b/src/dbt/kipptaf/models/finance/intermediate/int_finance__enrollment_targets.sql
@@ -18,7 +18,7 @@ select
     sch._dbt_source_project,
 
     '`{{ target.database }}`.`'
-    || regexp_extract(sch._dbt_source_relation, r'(kipp\w+)_')
+    || sch._dbt_source_project
     || '_powerschool'
     || '`.`base_powerschool__student_enrollments`' as _dbt_source_relation,
 from {{ ref("stg_google_sheets__finance__enrollment_targets") }} as et
--- a/src/dbt/kipptaf/models/finance/intermediate/int_finance__enrollment_targets.sql
+++ b/src/dbt/kipptaf/models/finance/intermediate/int_finance__enrollment_targets.sql
@@ -15,6 +15,8 @@ select
     et.lep_only_ratio,
     et.sped_ratio,
 
+    sch._dbt_source_project,
+
     '`{{ target.database }}`.`'
     || regexp_extract(sch._dbt_source_relation, r'(kipp\w+)_')
     || '_powerschool'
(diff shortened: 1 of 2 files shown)


cbini:
Addressed the review:

- **Finding 1 (reuse):** applied in `8a726388d`. The synthesized `_dbt_source_relation` now concatenates `sch._dbt_source_project` directly instead of re-running `regexp_extract(sch._dbt_source_relation, ...)`. Byte-identical output (both equal the region by construction); rebuilt in dev, grain test still passes, sampled values are well-formed (`kippcamden` / `kippmiami` / `kippnewark`).
- **Finding 2 (join grain):** verified. `school_number` is not globally unique in `stg_powerschool__schools` — a placeholder id appears 4x (one per district) and another 2x — but those don't match real target `schoolid`s, so the join doesn't fan (641 rows = 641 distinct `academic_year` / `schoolid` / `grade_level`), and the `unique_combination_of_columns` test guards a future real collision. Left the description as-is rather than assert a global-uniqueness that doesn't actually hold.

PR #4543, opened 2026-07-24
refactor(dbt): swap union_dataset_join_clause for _dbt_source_project in consumer intermediates

cbini:
# Pull Request

## Summary & Motivation

Consumer wave of the #3142 `_dbt_source_project` migration. Replaces
`union_dataset_join_clause(...)` internal join predicates with direct
`a._dbt_source_project = b._dbt_source_project` equality in 3 enabled consumer
intermediates (7 macro calls). Behavior-preserving — `_dbt_source_project`
equals `regexp_extract(_dbt_source_relation, r'(kipp\w+)_')` by construction, so
each swap is algebraically identical to the macro, and every join partner
already exposes the column in prod.

Files (all gate-verified — built `--empty --defer --favor-state` against prod;
tests pass):

- `int_tableau__state_assessments_demographic_comps`
- `int_assessments__course_enrollments`
- `int_powerschool__state_assessments_transfer_scores`

## Note — disabled gradebook models excluded

An earlier revision also swapped 3 disabled `int_tableau__gradebook_audit_*`
models. Review plus verification found those are stale beyond the macro swap:
their producer `int_powerschool__gradebook_assignments_scores` no longer exposes
`_dbt_source_relation` (prod exposes only `_dbt_source_project`), so their
`assignment_score_rollup` CTE already references a dropped column and would not
build if enabled. They need a holistic `_dbt_source_relation` to
`_dbt_source_project` re-migration (also `sec._dbt_source_relation` from the
scaffold), not a mechanical macro swap. Reverted them here; tracked for a
dedicated pass before the macro is deleted (Batch 3).

## AI Assistance

### dbt

- [x] Column-resolution gate passes for all 3 models (0 errors).
- [x] Join-predicate swap only — no output columns changed; contracts intact.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Refs #3142

### DIFF
--- a/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__course_enrollments.sql
+++ b/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__course_enrollments.sql
@@ -31,7 +31,7 @@ with
             {{ ref("base_powerschool__student_enrollments") }} as co
             on ce.cc_studentid = co.studentid
             and ce.cc_academic_year = co.academic_year
-            and {{ union_dataset_join_clause(left_alias="ce", right_alias="co") }}
+            and ce._dbt_source_project = co._dbt_source_project
             and co.rn_year = 1
         where not ce.is_dropped_course
 
--- a/src/dbt/kipptaf/models/extracts/tableau/intermediate/int_tableau__gradebook_audit_assignments_student.sql
+++ b/src/dbt/kipptaf/models/extracts/tableau/intermediate/int_tableau__gradebook_audit_assignments_student.sql
@@ -164,7 +164,7 @@ left join
     and ce.assignment_category_code = a.category_code
     and a.duedate between ce.week_start_date and ce.week_end_date
     and ce.date_enrolled <= a.duedate
-    and ce._dbt_source_project = a._dbt_source_project
+    and {{ union_dataset_join_clause(left_alias="ce", right_alias="a") }}
     and a.iscountedinfinalgrade = 1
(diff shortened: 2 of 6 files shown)


cbini:
Verified the finding — the `assignment_score_rollup` (`asg`) CTE carries only `_dbt_source_relation`, so the swapped `asg._dbt_source_project` would not resolve. Correct catch; the parse-only gate on disabled models can't see column resolution.

While fixing it I found a deeper problem: the producer `int_powerschool__gradebook_assignments_scores` no longer exposes `_dbt_source_relation` at all — prod `INFORMATION_SCHEMA.COLUMNS` shows only `_dbt_source_project`. So these disabled models are already stale: the `asg` CTE's *existing* `_dbt_source_relation` select/group-by references a dropped column, and threading `_dbt_source_project` in alongside it still would not build. Making them correct is a holistic `_dbt_source_relation` to `_dbt_source_project` re-migration (also `sec._dbt_source_relation` from the scaffold), verified by enabling the whole disabled gradebook subtree — out of scope for a mechanical macro swap.

Reverted the 3 disabled `gradebook_audit` models from this PR (restored the macro). Shipping only the 3 enabled, gate-verified swaps: `int_tableau__state_assessments_demographic_comps`, `int_assessments__course_enrollments`, `int_powerschool__state_assessments_transfer_scores`. The disabled gradebook staleness + macro removal will be handled together before Batch 3.

PR #4547, opened 2026-07-24
refactor(dbt): swap union_dataset_join_clause in big-fanout intermediates

cbini:
# Pull Request

## Summary & Motivation

Consumer wave of the #3142 `_dbt_source_project` migration — PR A of a revised
cut. Replaces `union_dataset_join_clause(...)` join predicates with direct
`a._dbt_source_project = b._dbt_source_project` equality across the six
big-fan-out intermediates (23 macro calls), threading the column through the
CTEs that feed each join.

Behavior-preserving: `_dbt_source_project` equals
`regexp_extract(_dbt_source_relation, r'(kipp\w+)_')` by construction, so each
swap is algebraically identical to the macro it replaces.

### Models

Plain swap plus CTE threading:

- `int_students__contacts` — 3 sites. `ps_person_contacts` is an in-model
  `union_relations` CTE with no `_dbt_source_project`, so it is split into a
  `_union` CTE plus a derive wrapper (pass-through principle: derive at the
  union view, never downstream). Column then threads through
  `ps_methods_ranked` / `ps_typed` / `ps_primary_phone_ranked` /
  `ps_primary_phone` / `students`.
- `int_extracts__student_enrollments_subjects` — 8 sites; threaded through
  `intervention_nj`, both `prev_yr_state_test` UNION ALL branches,
  `magoosh_exempt`, `bucket_programs`, `mtss`.
- `int_reporting__promotional_status` — 6 sites; threaded through `attendance`,
  `credits`, `ps_log`, `exempt_override`, `force_retention`.
- `int_kippadb__roster` — 4 sites, no threading needed.

Union-derive restructures — the join alias reads the raw `union_relations` CTE,
which has `_dbt_source_relation` but not `_dbt_source_project`, so the derive
moves into a CTE between the union and the join:

- `base_powerschool__sections` — derive moved out of the final `SELECT` into a
  `sections` CTE.
- `int_powerschool__ps_adaadm_daily_ctod` — derive moved out of the `calcs`
  select list into a `memberships` CTE; the column is now a plain pass-through
  in `calcs` (and moves to the front of that select per ST06 ordering, since it
  is now a column enumeration rather than a function).

### Incidental cleanup

`int_reporting__promotional_status` had three inline
`regexp_extract(mem._dbt_source_relation, ...)` **value** derivations inside
`hs_at_risk_absences` / `hs_off_track_absences`. Now that `mem` carries the
materialized column these become plain `mem._dbt_source_project` refs — same
values, one less regexp per row.

## Batching change

This PR is the first under a revised cut of the remaining work, documented in
`docs/superpowers/plans/2026-07-22-dbt-source-project-completion.md`. The
directory-based cut (`2a-1` … `2f`) was sized before the Batch 1 producers
merged; that ordering constraint is gone, so the binding constraints are now
build-verifiability and descendant fan-out.

Grouping these six together costs **174** `state:modified+` models versus
**410** if split per-model — `int_powerschool__ps_adaadm_daily_ctod` (157) and
`int_students__contacts` (155) alone share 151 descendants, so the other four
add only 17.

The plan update also **corrects** the snapshot gating note. A `check`-strategy
snapshot only writes columns onto rows it touches, so adding
`_dbt_source_project` upstream never backfills history:
`snapshot_powerschool__gpa_term` lacks the column entirely, and
`snapshot_powerschool__gpa_cumulative` carries it with 98.8% of rows NULL. A
plain swap on the topline GPA models would build green and silently drop most of
the join, so they must derive locally — deferred to their own PR.

## AI Assistance

### dbt

- [x] Column-resolution gate passes for all 6 models: built
      `--empty --defer --favor-state` against the prod manifest, 0 errors.
- [x] Join-predicate swaps and pass-through threading only — no filter, grain,
      or aggregation semantics changed.
- [x] `group by` additions are functionally determined by an existing grouping
      column, so grouping is unchanged.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Refs #3142

### DIFF
--- a/docs/superpowers/plans/2026-07-22-dbt-source-project-completion.md
+++ b/docs/superpowers/plans/2026-07-22-dbt-source-project-completion.md
@@ -398,6 +398,71 @@ non-producer `extract_source_project` filter/alias sites in
 `int_students__fldoe_fte` — done in 2e with `int_students__` — and
 `qa_edplan__powerschool_mismatch` to plain `_dbt_source_project` refs.)
 
+### Revised cut for the remaining work (2026-07-24)
+
+The directory cut above (`2a-1` … `2f`) is **superseded** for whatever is still
+unswapped. It was sized when Batch 1 producers had not merged and ~80 files were
+in flight; that ordering constraint is gone, so every remaining change is
+self-contained within its own file. The binding constraints now are (a) whether
+a file can be build-verified and (b) descendant fan-out, which sets CI cost.
+Counts below come from the prod manifest `child_map`.
+
+#### PR A — big-fanout intermediates
+
+`int_students__contacts`, `int_powerschool__ps_adaadm_daily_ctod`,
+`int_extracts__student_enrollments_subjects`, `int_kippadb__roster`,
+`int_reporting__promotional_status`, `base_powerschool__sections`.
+
+CTE threading plus two union-derive restructures (`base_powerschool__sections`,
+`int_powerschool__ps_adaadm_daily_ctod`) whose join alias reads the raw
+`union_relations` CTE, so `_dbt_source_project` must be derived in a CTE between
+the union and the join. Combined `state:modified+` is **174** models versus
+**410** split per-model — `int_powerschool__ps_adaadm_daily_ctod` (157) and
+`int_students__contacts` (155) alone share 151 descendants, so the other four
+cost `+17`.
+
+#### PR B — leaves (reports, marts, singular test)
+
+`rpt_deanslist__promo_status`, `rpt_powerschool__autocomm_students`,
+`rpt_tableau__state_assessments_dashboard`, `rpt_tableau__ops_dashboard`,
+`int_finance__enrollment_targets`, `bridge_course_section_teachers`,
+`dim_students`, `fct_family_communications`,
+`test_incorrect_student_number_pearson`.
+
+Combined `state:modified+` is **32** versus 35 split — the four reports have
+zero descendant models each. `rpt_tableau__ops_dashboard` additionally drops the
+synthetic `_dbt_source_relation` from `int_finance__enrollment_targets` and
+rewrites `target_union` / `targets` / the join to `_dbt_source_project`,
+including turning its regexp region filter into
+`_dbt_source_project in ('kippnewark', 'kippmiami')`; verify `cal` exposes the
+column first. Marts are contract-enforced — apply Pattern C.
+
+#### PR C — topline (snapshot-derive, not a swap)
+
+`int_topline__gpa_cumulative_weekly`, `int_topline__gpa_term_weekly`,
+`int_topline__iready_diagnostic_weekly`. Combined `state:modified+` is 7. Kept
+standalone because it is the only remaining group where correctness is not
+mechanical — see the corrected snapshot note below.
+
+#### PR D — unverifiable / stale pass
+
+`int_tableau__gradebook_audit_assignments_student`,
+`int_tableau__gradebook_audit_assignments_teacher`,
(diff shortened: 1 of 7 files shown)


cbini:
## Verification follow-up

CI is green (dbt Cloud run `70403218252534`). Two things from the review are
worth recording, plus triage of the 12 test warnings.

### Value-level equivalence (review finding 2)

The `--empty` gate proves column resolution, not values — a fair callout, and
the exact failure mode the plan-doc correction warns about. Rather than a
dev-build row-count diff, I tested the assumption the migration rests on
directly against prod.

Only table-materialized join partners can have a stored `_dbt_source_project`
that drifts from the regexp; views recompute it on read. For all 10 such
partners across the six models, `_dbt_source_project IS DISTINCT FROM
regexp_extract(_dbt_source_relation, r'(kipp\w+)_')` returns **0 mismatches over
239,077 rows**:

`int_extracts__student_enrollments`, `int_overgrad__students`,
`int_powerschool__gpa_cumulative`, `int_powerschool__s_nj_stu_x_unpivot`,
`int_powerschool__spenrollments`, `int_powerschool__terms`,
`stg_powerschool__s_nj_crs_x`, `stg_powerschool__s_nj_stu_x`,
`stg_powerschool__schools`, `stg_powerschool__terms`.

### calendar_week region predicate (review finding 1)

Added in `51a26bb6c`, after confirming it is a no-op on current data rather than
assuming it:

- No `schoolid` in `int_powerschool__calendar_week` maps to more than one
  `_dbt_source_project` (0 rows), so matching on `schoolid` already implies
  matching region.
- Neither side has a null `_dbt_source_project` — `int_powerschool__calendar_week`
  0 of 7,904, `int_powerschool__ps_adaadm_daily_ctod` 0 of 12,840,351. This was
  the load-bearing check: with nulls present the new equality would have dropped
  rows instead of being inert.

### Test warnings — 12, none introduced here

Three land inside this PR's blast radius, so I checked each against prod rather
than assuming:

- `bridge_student_contacts` to `dim_students` — 20 in CI, **20 in prod**.
  Identical, pre-existing.
- `fct_student_attendance_daily` to `dim_student_enrollments` — 2,305 in CI vs
  2,286 in prod. The test's parent resolves to the `zz_stg_kipptaf_marts` defer
  copy, a different vintage than the PR-branch child; drift, not logic.
- `dim_student_section_enrollments` to `dim_course_sections` — 6,133 in CI vs
  **0 in prod**, which looked alarming until traced. The child is byte-identical
  to prod (790,776 rows, 29,425 distinct `course_section_key`), so nothing about
  it changed. The parent is short 1,144 keys, which reconciles exactly against
  stale district staging copies:

  | district | prod | `zz_stg` | delta |
  | --- | --- | --- | --- |
  | kippnewark | 21,520 | 20,806 | -714 |
  | kippcamden | 5,915 | 5,625 | -290 |
  | kipppaterson | 532 | 392 | -140 |
  | kippmiami | 3,433 | 3,433 | 0 |

  Total -1,144, matching the `base_powerschool__sections` and
  `dim_course_sections` shortfalls to the row. `Clone - Staging (On-Demand)`
  refreshes kipptaf-level relations, not the district-level
  `zz_stg_kipp<district>_*` copies, so this is a CI-vintage artifact.

The remaining 9 warnings are on survey, KIPP Forward, AP-assessment and
gradebook models outside the swapped joins; the gradebook ones are the
documented PowerSchool double-write issue (#3900).

### Note for Batch 3

Review finding 4 is a good catch and is now tracked for the macro-removal batch:
`base_powerschool__student_enrollments` derives `_dbt_source_project` and
`code_location` from byte-identical expressions — two names for one value on a
hub model.

PR #4549, opened 2026-07-24
refactor(dbt): swap union_dataset_join_clause in leaf reports and marts

cbini:
# Pull Request

## Summary & Motivation

Consumer wave of the #3142 `_dbt_source_project` migration — PR B of the revised
cut (PR A was #4547). Replaces the remaining `union_dataset_join_clause(...)`
join predicates in the **leaf** models: reports, marts, and one singular test.
28 macro calls, plus one structural cleanup described below.

Behavior-preserving: `_dbt_source_project` equals
`regexp_extract(_dbt_source_relation, r'(kipp\w+)_')` by construction, so each
swap is algebraically identical to the macro it replaces.

**One qualification, surfaced in review.** That claim holds unconditionally for
the 28 join-predicate swaps, but **not** for the `targets` `group by` rewrite in
`rpt_tableau__ops_dashboard` — there it is prod-equivalent and a *fix* for
non-prod targets. The old code grouped on `_dbt_source_relation`, comparing
`int_finance__enrollment_targets`' synthetic literal against the real value from
`int_extracts__student_enrollments`. Measured in prod, those are byte-identical
(e.g. `` `teamster-332318`.`kippnewark_powerschool`.`base_powerschool__student_enrollments` ``
on both sides), so one group forms per key and the rewrite changes nothing.
Under `target=staging`/`dev` the real side resolves to
`zz_stg_<proj>_powerschool` while the synthetic hardcodes `_powerschool`, so
they did *not* match — two groups formed, both matched `se` through the regexp,
and the result fanned out. The rewrite collapses that.

Practical consequence: a dev- or CI-built before/after comparison of this model
will legitimately differ, and only a prod comparison shows parity. Noting it so
the delta isn't later mistaken for a regression.

### Plain swaps

- `rpt_tableau__state_assessments_dashboard` (12) — plus threading
  `_dbt_source_project` through the `assessment_scores` CTE (all three UNION ALL
  branches) and the `schedules` CTE.
- `rpt_powerschool__autocomm_students` (4) — plus threading through
  `grad_path` / `grad_path_pivot`.
- `bridge_course_section_teachers` (3, contracted mart)
- `dim_students` (2, contracted mart)
- `rpt_deanslist__promo_status` (2)
- `fct_family_communications` (1, contracted mart)
- `test_incorrect_student_number_pearson` (1, singular test)

The pearson test is a CRLF file; only the changed line had its carriage return
stripped, so the diff stays at one line.

### `rpt_tableau__ops_dashboard` plus `int_finance__enrollment_targets`

These two are grouped because they are one change.
`int_finance__enrollment_targets` carried a **synthetic**
`_dbt_source_relation` — a string literal assembled as
`'...' || sch._dbt_source_project || '_powerschool' || '...'` — that existed
solely so the macro's `regexp_extract` would find a region to match on. It was
never a real relation path, and `rpt_tableau__ops_dashboard` was its only
consumer.

With the join swapped to `_dbt_source_project` that scaffolding is dead, so this
PR removes it and rewrites the ops_dashboard chain to carry the real column end
to end:

- `att_mem` selects and groups by `_dbt_source_project` (available since #4547
  threaded it through `int_powerschool__ps_adaadm_daily_ctod`).
- Both `target_union` branches project `_dbt_source_project`.
- The Newark/Miami region filter becomes
  `_dbt_source_project in ('kippnewark', 'kippmiami')` rather than a
  `regexp_extract` over the relation string.
- `targets` selects and groups by `_dbt_source_project`.
- `int_finance__enrollment_targets` drops the synthetic column; its
  `properties.yml` entry is removed and the `_dbt_source_project` description
  corrected — it referenced a column this model no longer has.

`rpt_tableau__ops_dashboard` never projected `_dbt_source_relation` in its final
`SELECT`, so its contract is unaffected.

## Verification

- Column-resolution gate: `--empty --defer --favor-state` against the prod
  manifest for all 9 nodes — **PASS=33, ERROR=0**.
- The gate reports only the first unresolved name per model, so rather than
  fix-and-retry blind I traced every join alias in
  `rpt_tableau__state_assessments_dashboard` (`e`, `m`, `sf`, `sf2`, `a`) to its
  producer and confirmed each exposes the column.
- Equivalence rests on the algebraic identity, measured directly against prod
  during #4547 — `_dbt_source_project IS DISTINCT FROM regexp_extract(...)`
  returned 0 mismatches across 239,077 rows on every table-materialized join
  partner.
- Confirmed before editing: `int_powerschool__calendar_rollup` (the `cal` alias)
  exposes `_dbt_source_project`, and `int_finance__enrollment_targets` has
  exactly one consumer.

## Two things left deliberately out of scope

Both are pre-existing and worth a separate look rather than a drive-by fix:

1. `rpt_powerschool__autocomm_students` derives an output column as
   `regexp_extract(se._dbt_source_relation, ...) as code_location`. It could
   collapse to `se._dbt_source_project`, same value — but it is a
   contract-visible output column and the change would force an ST06 reordering
   inside a contracted model. Queued for the macro-removal batch alongside the
   duplicate `_dbt_source_project` / `code_location` expressions on
   `base_powerschool__student_enrollments` flagged in #4547 review.
1. In `rpt_tableau__state_assessments_dashboard`, the `schedules_current` join
   carries **no** region predicate at all — it matches only on
   `students_student_number` and `courses_credittype`. Whether it should have
   one is a real question, but adding it is a behavior change that needs its own
   analysis, not a rider on a behavior-preserving PR.

   (The dead `c._dbt_source_relation` that CTE also selected **was** removed, in
   `890f9b1c1` — dropping a column no consumer reads is not a behavior change,
   unlike adding the predicate.)

## AI Assistance

### dbt

- [x] Column-resolution gate passes for all 9 nodes (0 errors).
- [x] Contracted marts unchanged in output column set — join predicates only.
- [x] `group by` changes swap one functionally-equivalent region key for
      another, so grain is unchanged.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Refs #3142

### DIFF
--- a/src/dbt/kipptaf/models/extracts/deanslist/rpt_deanslist__promo_status.sql
+++ b/src/dbt/kipptaf/models/extracts/deanslist/rpt_deanslist__promo_status.sql
@@ -126,7 +126,7 @@ left join
     on co.studentid = gpa.studentid
     and co.yearid = gpa.yearid
     and term = gpa.term_name
-    and {{ union_dataset_join_clause(left_alias="co", right_alias="gpa") }}
+    and co._dbt_source_project = gpa._dbt_source_project
 left join
     {{ ref("int_reporting__promotional_status") }} as p
     on co.student_number = p.student_number
@@ -142,6 +142,6 @@ left join
     on co.studentid = ss.studentid
     and co.academic_year = ss.academic_year
     and ss.log_type = 'ARFR (Summer School)'
-    and {{ union_dataset_join_clause(left_alias="co", right_alias="ss") }}
+    and co._dbt_source_project = ss._dbt_source_project
 
 where co.academic_year = {{ var("current_academic_year") }} and co.rn_year = 1
--- a/src/dbt/kipptaf/models/extracts/powerschool/rpt_powerschool__autocomm_students.sql
+++ b/src/dbt/kipptaf/models/extracts/powerschool/rpt_powerschool__autocomm_students.sql
@@ -1,12 +1,12 @@
 with
     grad_path as (
-        select _dbt_source_relation, student_number, discipline, final_grad_path_code,
+        select _dbt_source_project, student_number, discipline, final_grad_path_code,
         from {{ ref("int_students__graduation_path_codes") }}
     ),
 
     grad_path_pivot as (
         select
-            _dbt_source_relation,
+            _dbt_source_project,
             student_number,
             s_nj_stu_x__graduation_pathway_math,
             s_nj_stu_x__graduation_pathway_ela,
@@ -77,20 +77,20 @@ from {{ ref("int_extracts__student_enrollments") }} as se
 left join
     {{ ref("stg_powerschool__students") }} as s
     on se.student_number = s.student_number
-    and {{ union_dataset_join_clause(left_alias="se", right_alias="s") }}
+    and se._dbt_source_project = s._dbt_source_project
 left join
     {{ ref("int_powerschool__district_entry_date") }} as de
     on se.studentid = de.studentid
-    and {{ union_dataset_join_clause(left_alias="se", right_alias="de") }}
+    and se._dbt_source_project = de._dbt_source_project
     and de.rn_entry = 1
 left join
     grad_path_pivot as g
     on se.student_number = g.student_number
-    and {{ union_dataset_join_clause(left_alias="se", right_alias="g") }}
+    and se._dbt_source_project = g._dbt_source_project
 left join
     {{ ref("stg_powerschool__s_stu_x") }} as pfs
     on se.students_dcid = pfs.studentsdcid
-    and {{ union_dataset_join_clause(left_alias="se", right_alias="pfs") }}
+    and se._dbt_source_project = pfs._dbt_source_project
 where
     se.academic_year = {{ var("current_academic_year") }}
     and se.rn_year = 1
--- a/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__ops_dashboard.sql
+++ b/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__ops_dashboard.sql
@@ -1,7 +1,7 @@
 with
     att_mem as (
         select
-            _dbt_source_relation,
+            _dbt_source_project,
             studentid,
             yearid,
 
@@ -9,7 +9,7 @@ with
             sum(membershipvalue) as n_membership,
         from {{ ref("int_powerschool__ps_adaadm_daily_ctod") }}
         where membershipvalue = 1
-        group by studentid, yearid, _dbt_source_relation
+        group by studentid, yearid, _dbt_source_project
     ),
 
     target_union as (
@@ -25,7 +25,7 @@ with
             at_risk_and_lep_ratio,
             lep_only_ratio,
             sped_ratio,
-            _dbt_source_relation,
+            _dbt_source_project,
         from {{ ref("int_finance__enrollment_targets") }}
 
         union all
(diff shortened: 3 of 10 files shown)


cbini:
## Review response

Verified each data-dependent claim against prod rather than taking it on faith.
Two findings applied, three declined with reasoning, one question answered.

### 1 — Confirmed, and the PR body was wrong. Body corrected.

This is the finding that mattered. Measured in prod, the two `_dbt_source_relation`
values the old `group by` compared are **byte-identical**:

| side | value | rows |
| --- | --- | --- |
| `int_finance__enrollment_targets` (synthetic) | `` `teamster-332318`.`kippnewark_powerschool`.`base_powerschool__student_enrollments` `` | 432 |
| `int_extracts__student_enrollments` (real) | `` `teamster-332318`.`kippnewark_powerschool`.`base_powerschool__student_enrollments` `` | 97,920 |

(same for kippcamden and kippmiami; kipppaterson has enrollments but no targets
row.)

So in prod the old code produced one group per key and the rewrite is a no-op.
Under `target=staging` the real side resolves to `zz_stg_<proj>_powerschool`
while the synthetic hardcodes `_powerschool` with no prefix — they don't match,
two groups form, and both matched `se` through the regexp. Exactly the latent
fan-out you describe.

My PR body's blanket "behavior-preserving" claim was therefore wrong for
non-prod targets. Corrected: prod-equivalent, and a fix for staging/dev.

### 3 — Applied (`890f9b1c1`).

Verified `schedules` reads only `c.school`, `c.teachernumber`, `c.teacher_name`
plus the join keys, so `c._dbt_source_relation` is genuinely unread. Your
distinction is the right one: dropping an unread column isn't a behavior change,
whereas adding the missing region predicate to that join is — which is why the
predicate stays out. Re-gated, PASS=1 ERROR=0.

### 4 — Declining.

`fct_family_communications` partitions its dedupe on `_dbt_source_relation`
while joining on `_dbt_source_project`. The 1:1 relation-to-project mapping
holds (I verified it for `base_powerschool__student_enrollments` while checking
a `SELECT DISTINCT` in #4552 — 4 projects, exactly 1 relation each), so the
current code is **already correct**; this is a tidiness change to a line the PR
doesn't otherwise touch. Rewriting working code inside a behavior-preserving PR
adds risk without changing behavior. Better placed with the macro-removal
cleanups.

### 5 — Declining, same reasoning.

`dim_students` hand-rolling `extract_region` is real duplication and your read of
the CLAUDE.md caveat is right (it doesn't apply here). But it's an untouched line
in a contracted mart, and the win is stylistic. Queued for the cleanup batch
alongside the `code_location` duplicate in `rpt_powerschool__autocomm_students`
and `base_powerschool__student_enrollments`.

### 2, 6, 7 — Acknowledged as deferrals, stated explicitly.

- **2** (no uniqueness test on the two dashboards) — pre-existing, and I agree a
  follow-up issue is the right home. Not adding a test to a contracted `rpt_`
  in a refactor PR without first confirming the grain actually holds.
- **6** (15 of 16 columns lack `description:`) — deferring deliberately, second
  time this YAML has been touched without them. Flagging rather than silently
  passing again.
- **7** (mixed line endings) — deliberate to keep the diff at one line; a
  `.gitattributes` normalization is the right fix and belongs on its own.

### 8 — Answered.

Deliberate batches, nothing blocked:

- `int_topline__*_weekly` (3) → **#4550**, already open and green. Split out
  because two of them are snapshot-fed and need a local derive, not a swap.
- `qa_powerschool__transfer_records` + `qa_edplan__powerschool_mismatch` (7) →
  **#4552**: both **deleted**. Every QA model was disabled together on
  2025-08-28 (`fa62b5ae8`), nothing references them, and a fifth view disabled
  by that same commit was already converted to a singular test and deleted
  (`aa1cd57bc`).
- `int_tableau__gradebook_audit_*` (5) → **kept as-is by decision**. They sit in
  a closed, fully-disabled subgraph superseded by the AY 2026-2027 rebuild
  (`7fb33d117`), and are unbuildable regardless since
  `int_powerschool__gradebook_assignments_scores` no longer exposes
  `_dbt_source_relation`.
- The 2 `analyses/` sites → **#4552**, migrated and verified by `dbt compile`
  plus a BigQuery dry run of the compiled SQL.

On the macro-removal batch stalling: it won't. I verified that deleting the macro
with the five disabled callers still in place leaves `dbt parse
--no-partial-parse` at exit 0 — they're parsed into `manifest.disabled`, but dbt
doesn't render a disabled model's Jinja deeply enough to fail on an undefined
macro. Macro removal is gated only on this PR and #4550 merging.

Issue #4318, opened 2026-07-02
fix(powerschool): purge dev-relation rows from prod GPA snapshots


anthonygwalters:
A snapshot run on **2025-12-03** ingested rows whose `_dbt_source_relation` points at dev datasets (`zz_cbini_kippnewark/kippcamden/kippmiami_powerschool`) into both prod GPA snapshots in `kipptaf_powerschool`. Because the check columns of those ghost keys never change, the rows are permanently open (`dbt_valid_to = 9999-12-31`) and will never age out.

## Counts (verified 2026-07-02)

| Snapshot | zz version rows | All open? | Injected |
| --- | --- | --- | --- |
| `snapshot_powerschool__gpa_term` | 54,560 | yes | 2025-12-03 only |
| `snapshot_powerschool__gpa_cumulative` | 27,967 | yes | 2025-12-03 only |

## Why it matters

`union_dataset_join_clause` extracts the district prefix via a regex that matches inside dev dataset names — `zz_cbini_kippnewark_powerschool` and `kippnewark_powerschool` both extract `kippnewark`. Consumers that join snapshots to enrollment models on that clause match BOTH the real row and the frozen December ghost row:

- `int_topline__gpa_term_weekly` — every enrollment week since 2025-12-03 can fan out 2x or serve stale December GPA values.
- `int_topline__gpa_cumulative_weekly` — same pattern, same risk (not yet individually verified).
- `int_powerschool__gpa_term_lookback` (new, #4315/#4316) — ships with a defensive relation filter referencing this issue; the filter can be reconsidered once cleanup lands and re-injection is prevented.

## Remediation

1. Delete the ghost rows from both snapshot tables (Claude sessions are SELECT-only; needs a human/ops run):

```sql
delete from `teamster-332318.kipptaf_powerschool.snapshot_powerschool__gpa_term`
where _dbt_source_relation like '%zz_%';

delete from `teamster-332318.kipptaf_powerschool.snapshot_powerschool__gpa_cumulative`
where _dbt_source_relation like '%zz_%';
```

2. Root-cause how a 2025-12-03 run compiled `int_powerschool__gpa_term_current` against `zz_cbini_*` sources while writing to the prod snapshot targets, and prevent recurrence (branch-deployment snapshot config? local run with a prod profile?).
[...]

PR #4550, opened 2026-07-24
refactor(dbt): resolve topline region joins without the macro

cbini:
# Pull Request

## Summary & Motivation

PR C of the revised #3142 cut (PR A was #4547; PR B is the leaf reports and
marts). Removes the last `union_dataset_join_clause` calls from the topline
weekly models.

Two of the three are **not** a plain column swap, which is why they were split
into their own PR.

### Snapshot-fed — derive locally

`int_topline__gpa_cumulative_weekly` and `int_topline__gpa_term_weekly` read
snapshots, so they derive `_dbt_source_project` inside the snapshot-reading CTE
rather than passing a stored column through. A `check`-strategy snapshot only
writes columns onto rows it touches, so it never backfills a new column onto
history. Measured against prod:

- `snapshot_powerschool__gpa_term` does not carry `_dbt_source_project` at all —
  a pass-through swap fails the build.
- `snapshot_powerschool__gpa_cumulative` does carry it, with **2,151,864 of
  2,178,255 rows null (98.8%)**; only rows written since 2026-06-24 are
  populated. A pass-through swap there compiles, builds green, passes the
  empty-resolution gate, and **silently drops most of the join**.

Deriving locally computes exactly what the macro computed, so both models stay
behavior-preserving by construction rather than depending on stored-column
freshness. This is the snapshot-fed exception already documented in
`src/dbt/kipptaf/CLAUDE.md`; the plan doc's older "wait for the snapshot to
re-run" framing was corrected in #4547.

### iReady — plain swap, but verified first

`int_topline__iready_diagnostic_weekly` is a straight swap. iReady is flagged as
a region **carve-out** because the shared `kippnj_iready` schema can make the
region regex disagree with the resolved project — which would have made this
swap a behavior change rather than a no-op. Verified against prod that it does
not apply here:

| `_dbt_source_project` | regexp-derived | rows |
| --- | --- | --- |
| kippnewark | kippnewark | 189,295 |
| kippcamden | kippcamden | 66,809 |
| kippmiami | kippmiami | 54,168 |

Zero divergence — the union is over district-level models, not the shared NJ
schema.

## Verification

- Column-resolution gate: `--empty --defer --favor-state` against the prod
  manifest — **PASS=3, ERROR=0**.
- `_dbt_source_project` stays CTE-internal in all three models, so no output
  column set changes and no `properties.yml` entries are required.
- `trunk check --force` clean.

## AI Assistance

### dbt

- [x] Column-resolution gate passes (0 errors).
- [x] Snapshot-fed models derive rather than pass through, per the documented
      exception.
- [x] Carve-out risk on the iReady join checked against prod before swapping.
- [x] No output column set changes.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Refs #3142

### DIFF
--- a/src/dbt/kipptaf/models/topline/intermediate/int_topline__gpa_cumulative_weekly.sql
+++ b/src/dbt/kipptaf/models/topline/intermediate/int_topline__gpa_cumulative_weekly.sql
@@ -10,6 +10,11 @@ with
 
             cast(dbt_valid_from as date) as dbt_valid_from_date,
             cast(dbt_valid_to as date) as dbt_valid_to_date,
+
+            /* Snapshot-fed, so derive locally rather than passing the stored
+               column through: the check strategy only writes columns onto rows
+               it touches, leaving snapshot history ~99% null. */
+            {{ extract_source_project() }} as _dbt_source_project,
         from {{ ref("snapshot_powerschool__gpa_cumulative") }}
     ),
 
@@ -36,7 +41,7 @@ left join
     on co.studentid = gpa.studentid
     and co.schoolid = gpa.schoolid
     and co.week_start_monday between gpa.dbt_valid_from_date and gpa.dbt_valid_to_date
-    and {{ union_dataset_join_clause(left_alias="co", right_alias="gpa") }}
+    and co._dbt_source_project = gpa._dbt_source_project
 where
     co.is_enrolled_week
     and co.school_level in ('MS', 'HS')
--- a/src/dbt/kipptaf/models/topline/intermediate/int_topline__gpa_cumulative_weekly.sql
+++ b/src/dbt/kipptaf/models/topline/intermediate/int_topline__gpa_cumulative_weekly.sql
@@ -11,9 +11,8 @@ with
             cast(dbt_valid_from as date) as dbt_valid_from_date,
             cast(dbt_valid_to as date) as dbt_valid_to_date,
 
-            /* Snapshot-fed, so derive locally rather than passing the stored
-               column through: the check strategy only writes columns onto rows
-               it touches, leaving snapshot history ~99% null. */
+            /* snapshot-fed: derive locally — the check strategy never backfills
+               the stored column across history (see kipptaf CLAUDE.md) */
             {{ extract_source_project() }} as _dbt_source_project,
         from {{ ref("snapshot_powerschool__gpa_cumulative") }}
     ),
--- a/src/dbt/kipptaf/models/topline/intermediate/int_topline__gpa_term_weekly.sql
+++ b/src/dbt/kipptaf/models/topline/intermediate/int_topline__gpa_term_weekly.sql
@@ -11,6 +11,12 @@ with
 
             cast(dbt_valid_from as date) as dbt_valid_from_date,
             cast(dbt_valid_to as date) as dbt_valid_to_date,
+
+            /* Snapshot-fed, so derive locally. This snapshot does not carry
+               _dbt_source_project at all, and adding it upstream would not help
+               — the check strategy only writes columns onto rows it touches, so
+               snapshot history would stay null. */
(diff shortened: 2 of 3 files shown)


cbini:
## Review response

The structural equivalence argument is better than what I put in the PR body —
particularly the iReady point. I've since also closed the "not verifiable here"
gap with a direct prod comparison. One finding applied, two declined with data.

### Value-level proof, since the warehouse was unavailable to the review job

The PR-branch CI build and prod are **byte-identical** for both GPA models:

| model | prod rows | PR-branch rows | prod distinct keys | PR-branch distinct keys |
| `int_topline__gpa_term_weekly` | 661,831 | 661,831 | 437,323 | 437,323 |
| `int_topline__gpa_cumulative_weekly` | 791,782 | 791,782 | 437,323 | 437,323 |

(`kipptaf_topline` vs `dbt_cloud_pr_70403104388001_4550_topline`, distinct over
`student_number, academic_year, week_start_monday`.) That's a direct before/after
equality, so the behavior-preservation claim no longer rests on the `--empty`
gate alone.

### 1 — Pushing back: the proposed test would fail today.

The suggested grain is **not unique** in the current data — see the table above:
661,831 rows against 437,323 distinct
`(student_number, academic_year, week_start_monday)` in `gpa_term_weekly`, and
791,782 against the same 437,323 in `gpa_cumulative_weekly`. Adding
`unique_combination_of_columns` on that triple at `severity: error` would break
CI immediately.

The duplication comes from the `co` side, not the join being swapped: a student
with multiple enrollment stints in a year yields several
`int_extracts__student_enrollments_weeks` rows per week. So a real uniqueness
test here needs a wider key (stint or school included), which is a modeling
question rather than something to settle in a refactor PR. You did say to verify
before adding at `error` — it doesn't hold.

### 2 — Declining for this PR; the guard belongs on #4318, which is open.

Ran your query. The ghosts are real and district-shaped, so your concern was
well-founded:

| relation | rows | `dbt_valid_from` | still open |
| --- | --- | --- | --- |
| `zz_cbini_kippnewark_powerschool` | 39,445 | 2025-12-03 | 0 |
| `zz_cbini_kippcamden_powerschool` | 10,771 | 2025-12-03 | 0 |
| `zz_cbini_kippmiami_powerschool` | 4,344 | 2025-12-03 | 0 |

Impact is bounded — across dedupe partitions
`(studentid, schoolid, yearid, valid_from_date)`: 2,469 have both a real and a
ghost row, 52,091 are ghost-only. All are closed (`dbt_valid_to` non-null), so
the date-range join limits them to that window.

Declining anyway, for two reasons. It is **not a regression** — the macro read
the same `_dbt_source_relation`, so ghosts matched identically before and after,
which the identical prod/PR counts above confirm. And porting the filter would
be a genuine behavior change (dropping rows) inside a PR whose entire claim is
behavior preservation, needing its own validation. **#4318** ("fix(powerschool):
purge dev-relation rows from prod GPA snapshots") is open and is the right home
— fixing it at the source removes the need for the guard in both consumers
rather than adding a third copy.

### 3 — Applied (`6e5a614af`).

Trimmed both comments to the structural reason and dropped the point-in-time
99% figure, which you're right will drift. Re-gated, PASS=2 ERROR=0.

### On the iReady argument

Your version is stronger than mine and I've noted it: `int_iready__diagnostic_results`
*rewrites* `_dbt_source_relation`, substituting `lc.location_dagster_code_location`
for the `kipp[a-z]+_` prefix, so the shared `kippnj_iready` schema is normalized
to the home region before any consumer reads it. My zero-divergence row counts
are a consequence of that rewrite, not independent evidence — a structural
guarantee beats a point-in-time count. Same for the
`int_powerschool__gpa_term_lookback` precedent, which I hadn't found; good to
know the local-derive pattern was already established rather than introduced
here.

Issue #4388, opened 2026-07-13
fix(renlearn): stg_renlearn__star fiscal-year partition overlap duplicates assessment_id rows


cbini:
## Problem

`stg_renlearn__star` has no uniqueness test and carries duplicate rows. Roughly 2,296 `assessment_id` values appear in two different `_dagster_partition_fiscal_year` partitions with identical data in every column — a fiscal-year re-pull overlap. A smaller set (~144) are genuine same-day multi-attempts.

## Impact

`fct_assessment_scores_enrollment_scoped` (from #3625) dedupes STAR defensively at PK grain with `order_by="scale_score desc, assessment_id desc"`, carrying a `TODO` that points here. Remove that dedupe CTE once staging is unique.

## Fix

Add a uniqueness test on `assessment_id` (or the score grain), then resolve the cross-partition re-pull overlap so each attempt lands once.

Refs #3625

cristinabaldor:
Partially addressed in #4551. The fact-level dedupe in `fct_assessment_scores_enrollment_scoped` now omits `academic_year` from its `partition_by`, so a STAR test pulled under two fiscal-year partitions (same `test_date`, differing pull-derived `academic_year`) collapses to one row instead of both surviving (~2.3k rows: 9,690 → 7,395). This matters because #4551 resolves `academic_year` from the test date, which would otherwise surface the re-pulls as a double-count in real academic-year buckets.

This does NOT fix the underlying `stg_renlearn__star` fiscal-year partition overlap — the `TODO(#4388)` to remove the fact dedupe once staging is deduplicated remains open.

anthonygwalters:
## Root cause found: the partition key comes from the wall clock, not the data

The cross-partition overlap isn't a re-pull artifact — it's structural, and it
recurs every July 1.

`build_renlearn_sftp_sensor` is wired with a partition key pinned to the current
fiscal year, evaluated at module load:

```python
# src/teamster/code_locations/kippmiami/renlearn/sensors.py
partition_key_start_date=CURRENT_FISCAL_YEAR.start.isoformat(),
```

The sensor then stamps every matched file with that key regardless of what the
file contains (`src/teamster/libraries/renlearn/sensors.py`, the
`MultiPartitionKey` construction). Renaissance overwrites a single fixed
filename (`KIPP Miami.zip`) rather than publishing year-scoped files, so when it
re-dropped in July 2026 still holding SY2025-26 results, the mtime check fired
[...]

Issue #4387, opened 2026-07-13
fix(iready): stg_iready__diagnostic_results lacks uniqueness test and has duplicate rows


cbini:
## Problem

`stg_iready__diagnostic_results` has no uniqueness test (the properties YAML declares only a `not_null` on `student_id`), and the data carries duplicates.

At the natural score grain (project, student, academic year, test round, subject, completion date) there are 3,835 duplicate groups (up to 16 rows in a group):

- ~2,500 groups are same-day retests with different scores
- ~1,300 groups are byte-identical duplicate rows

## Impact

`fct_assessment_scores_enrollment_scoped` (from #3625) dedupes iReady defensively at PK grain with `order_by="start_date desc, scale_score desc"`, carrying a `TODO` that points here. Remove that dedupe CTE once staging is unique.

## Fix

Add a `dbt_utils.unique_combination_of_columns` test at the intended grain, then investigate and eliminate the byte-identical duplicates (likely a partition/re-pull overlap) and decide the canonical row for same-day retests upstream.

Refs #3625

cristinabaldor:
Partially addressed in #4551. The fact-level dedupe in `fct_assessment_scores_enrollment_scoped` now omits `academic_year` from its `partition_by`, so an i-Ready test pulled under two fiscal-year partitions (same `test_date`, differing pull-derived `academic_year`) collapses to one row instead of both surviving (~53.8k rows: 299,878 → 246,121). This matters because #4551 resolves `academic_year` from the test date, which would otherwise surface the re-pulls as a double-count in real academic-year buckets.

This does NOT fix the underlying `stg_iready__diagnostic_results` duplication — the `TODO(#4387)` to remove the fact dedupe once staging has a uniqueness guarantee remains open.

Issue #4298, opened 2026-07-02
fix(cube): America/New_York default query timezone breaks every dates join - academic_year returns NULL in all student views


anthonygwalters:
## Summary

The Cube Cloud deployment applies an `America/New_York` query timezone to every query that does not specify one (confirmed: the `/load` response echoes `timezone: America/New_York` when the request carries no timezone — so `CUBEJS_DEFAULT_TIMEZONE` is presumably set on the deployment; neither the repo model, `queryRewrite`, nor the MCP server sets it). Cube bakes its timezone conversion (`convertTz`) into the `dates` cube `date_day` time dimension **inside join predicates**, producing (quote marks omitted around the timezone literal to dodge GitHub sanitizer artifacts):

```sql
LEFT JOIN kipptaf_marts.dim_dates AS dates
  ON TIMESTAMP(DATETIME(dates.date_timestamp, America/New_York))
   = CAST(student_assessment_administrations.administered_date_key AS TIMESTAMP)
```

`date_timestamp` is midnight UTC; the left side shifts it to 19:00/20:00 of the previous day while the right side stays midnight UTC. **The predicate matches 0 of 4,018 calendar dates** (verified in BigQuery), so the `dates` join is dead for every row. Every `dates`-sourced view member (`academic_year`, `academic_year_label`, `month_name`, `month_number`, ...) returns NULL, and any filter on those members returns 0 rows.

The warehouse is clean — this is entirely a Cube-layer failure. Joining the marts directly in BigQuery exactly as the cube declares yields 0 null `academic_year` across all 489,751 grade-5 Math standard-level rows, and QA1 has 236,102 rows in AY 2025-26.

## Proof

Identical `/load` query two ways (measures: `count_scores`, dimensions: `academic_year`, filter `module_code = QA1` on `student_assessment_scores_summary`):

- No timezone (deployment default applies): all 1,365,336 QA1 scores collapse into a single `academic_year: null` bucket.
- Explicit `timezone: UTC`: returns academic years 2017-2025 correctly, AY2025 = 236,100, matching the warehouse.

Under UTC the conversion is a no-op, which is why the joins work with an explicit UTC timezone.

## Affected joins

- `src/cube/model/cubes/student_assessments/student_assessment_administrations.yml` (dates join, line ~10) — equality join, dead
- `src/cube/model/cubes/student_attendance/student_attendance.yml` (line ~8) — equality join, dead
[...]

PR #4551, opened 2026-07-24
fix(cube): resolve academic_year and description defects on assessment scores view

### LINKED ISSUE #4546: fix(cube): student_assessment_scores_view schema and data-quality defects
Four defects on `student_assessment_scores_view` (and its upstream ingestion), surfaced while building the achievement-director assessment direction files (see #4545). Grouped as one tracker: (a) and (b) are Cube-model description fixes; (c) and (d) are upstream data/ingestion issues. All verified against the live connector on 2026-07-24.

## (a) `assessment_type` description does not match the real values — description fix

The dimension description lists `illuminate` / `state_nj` / `state_fl` / `college` / `ap`. The real populated values are granular: `illuminate`, `iready`, `dibels`, `star`, `state_nj_njsla`, `state_nj_njsla_science`, `state_nj_njgpa`, `state_fl_fast`, `state_fl_science`, `state_fl_eoc`. In particular `dibels` / `iready` / `star` are populated but entirely absent from the description, which caused a session to conclude those sources did not exist. Fix: enumerate the actual values in the `description:`.

## (b) `academic_year` recommends a field not exposed on the view — description fix

`academic_year`'s description says to prefer `academic_year_label` over the integer, but `academic_year_label` is not exposed on this view. Either expose the label dimension or drop the recommendation. Also worth stating in the description: `academic_year` is null for state and for `iready` / `dibels` / `star` rows, so `date_taken` is the reliable year-scoping field for those sources.

## (c) i-Ready score rows duplicated ~2x per student per administration — data / ingestion

For i-Ready (e.g., EOY Reading), each student record appears roughly twice, identical in every field except the `assessment_score_key` surrogate key. This looks like an ingestion fan-out (a join emitting two output rows per input). Consumers double-count on `count_scores` unless they dedup on the PK. Fix: investigate the i-Ready ingestion path.

## (d) `is_replacement` not populated for i-Ready — data

`is_replacement` is null throughout i-Ready rows, so a student with genuine multiple attempts cannot be disambiguated to an authoritative score (current workaround: most-recent `date_taken`). If retests are meaningful for i-Ready, populate the flag as for other internal sources; if not applicable, say so in the field description.

Found via the Cube MCP working-group sessions (logs retained locally, no PII). Interim guidance for all four is already documented in the direction files on #4545.

cristinabaldor:
# Pull Request

## Summary & Motivation

Closes #4546. Resolves the four `student_assessment_scores_view` defects.

**(b) `academic_year` null for state/vendor sources — data fix.** The Cube
`dates` join keyed on `dim_assessment_administrations.administered_date_key`,
which is null for every non-Illuminate administration (state, iReady, DIBELS,
STAR record an assessment *window*, not a single date), so `academic_year` came
back null for ~2.9% of score rows. This adds `assessment_date_key` to
`fct_assessment_scores_enrollment_scoped` — `coalesce(administration date, test
date)` — and moves the Cube `dates` join off the administration cube and onto
the scores cube, keyed on `assessment_date_key`. `academic_year` now resolves for
every source; Illuminate values are unchanged (the coalesce falls to the
administration date, which Illuminate always has). `academic_year_label` is now
exposed on the view too.

**(a) `assessment_type` description.** Enumerated the 10 real values; dropped
phantom `college`/`ap` (they have administrations but no scores on this view).

**(d) `is_replacement` description.** Clarified it is Illuminate-only
(replacement curriculum); null for all state/vendor sources by design — there is
no value to populate.

**(c) i-Ready/STAR duplication — now fixed here.** Code review surfaced that (c)
is the fiscal-year re-pull duplication (#4387 i-Ready / #4388 STAR): a physical
test pulled under two fiscal-year partitions survives twice because the fact
dedupe partitioned by `academic_year`. My initial per-administration check (~1.03
rows/student) missed the cross-administration case. Left unfixed, (b)'s
test-date-derived `academic_year` would collapse both copies into one real
academic-year bucket and double-count. The i-Ready/STAR dedupe `partition_by`
now omits `academic_year`, collapsing ~53.8k i-Ready and ~2.3k STAR re-pull rows
to one each (new grain unique). Underlying staging dups remain tracked in
#4387/#4388.

**Why #4298 did not cover (b):** #4298 fixed the `America/New_York` timezone
`convertTz` that zeroed the join for *all* rows (Illuminate now resolves).
#4546(b) is a different root cause — a null join key for the vendor sources,
which were added 11 days after #4298 (in #4393).

### AI Assistance

Investigation, design, and implementation were Claude-assisted; the approach
(coalesced calendar date; keep the administration join for other dimensions) was
human-directed through review.

### dbt

- [x] Properties file updated for the new `assessment_date_key` column (contract +
      FK + relationships test)
- [x] **Breaking change?** No columns renamed or removed. `assessment_date_key` is
      additive; `academic_year` is unchanged for Illuminate and
      newly-populated (was null) for state/vendor. No exposure break.

Validated locally: 0 null `academic_year` across all 10 sources; Illuminate AY
range unchanged (2015-16 to 2025-26); `dbt compile` and `trunk check` clean. The
contract build and relationships test run in dbt Cloud CI; the Cube model
resolves on branch staging.

## CI checks

- [ ] **Trunk**
- [ ] **dbt Cloud**
- [ ] **Dagster Cloud**


cbini (dismissed):

cristinabaldor:

cristinabaldor:
Resolving the three items the updated review left open:

**Finding 8 (build-range floor)** — verified against the warehouse: min `assessment_date_key` is `2015-08-14`, with 0 rows before the `2015-07-01` floor and 0 future-dated. The floor and the coverage claim hold.

**Finding 10 (two FK constraints to `dim_dates`)** — the dbt Cloud contract build is green on the head commit, so BigQuery accepts both foreign keys in the one table DDL. Also confirmed locally: a `dbt run --target dev` of the fact created the table with both constraints.

**Finding 3 (pre-agg academic_year split)** — added a note to the `proficiency_rollup` comment (dbae8dd8). Quantified it: 0 state administrations split across academic years; only ~10 iReady and 1 DIBELS diagnostic *windows* straddle July 1, which is correct per-score attribution (a July sitting belongs to the new academic year), not a fan-out.

Separately, ran a live Cube dev-server SQL-API validation as a network-scoped viewer: all 7 measures compute across all 10 sources, and `academic_year` / `academic_year_label` populate for every vendor and state source with no ambiguity. That validation is how the `dates.academic_year_label` qualification bug was caught and confirmed fixed.

anthonygwalters (approved):

PR #4552, opened 2026-07-24
refactor(dbt): retire disabled qa views and migrate collegeboard analyses

cbini:
# Pull Request

## Summary & Motivation

PR D of the revised #3142 cut (A = #4547 merged, B = #4549, C = #4550). Clears
**this batch's** `union_dataset_join_clause` call sites — the QA and `analyses/`
slice.

To be precise, since an earlier draft of this line overstated it: ~35 enabled
call sites remain on this branch, all of them belonging to **#4549** (reports
and marts) and **#4550** (topline). The macro is clear only once all three of
these merge. Nothing here is blocked on them, and they are not blocked on this.

The seven files in this batch turned out to be three different problems, so they
got three different answers rather than one blanket treatment.

## Deleted — retired QA views

`qa_edplan__powerschool_mismatch` and `qa_powerschool__transfer_records`, plus
their `properties.yml`.

Every QA model in the project was disabled together on 2025-08-28 by
`fa62b5ae8` ("chore: disable qa"), and nothing references these two. The
precedent is explicit: a fifth view disabled by that same commit,
`qa_powerschool__course_enrollment_overlap`, was converted to a singular test
and then deleted (`aa1cd57bc`). Convert-then-delete is the established pattern
for a retired QA view here.

Both `qa/` directories are now empty and removed — those were their only models.
No `dbt_project.yml` config referenced the paths.

One dependency this surfaced: `tests/assets/test_assets_dbt.py` asserted on the
deleted model's Dagster asset key, so that selection entry is dropped; the
remaining four keys still exercise the test. That list also names
`qa_kippadb__hs_enrollment_audit`, which is likewise disabled — pre-existing,
left alone. Neither has failed because `tests/assets/` is not run in CI.

## Migrated — collegeboard analyses

- `collegeboard_ap_downstream_lineage_root_cause` — plain swap.
- `collegeboard_ap_tiered_crosswalk_match` — swap plus threading
  `_dbt_source_project` through the `ps` and `course_bridge` CTEs, and a prose
  comment that named the macro.

These are live tooling, not dead code: created 2026-07-21 (`e88cdca33`) and
referenced by `.claude/skills/collegeboard-ap-data-ingest-protocol/SKILL.md`,
which documents the `dbt compile` command to run them.

`ps` is a `SELECT DISTINCT`, so swapping the relation column for the project
column there could have collapsed rows. Verified against prod first that the
mapping is 1:1 — 4 projects, exactly 1 relation each — so the distinct grain is
unchanged.

**They are also verifiable**, contrary to how this batch was originally framed.
`dbt compile --target prod` followed by a BigQuery **dry run** of the compiled
SQL resolves every column against real prod schemas. Both pass:
`Query successfully validated` for each.

## Kept as-is — the disabled gradebook trio

No change, per direction. For the record on why they are safe to leave: they
belong to a closed, fully-disabled subgraph of seven models
(`int_tableau__gradebook_audit_flags`, both `*_scaffold`s, and
`rpt_tableau__assignment_checks` alongside the three) that reference only each
other. Nothing enabled reaches them — the live `gradebook_audit` exposure
depends solely on `rpt_tableau__gradebook_audit`. They were disabled 2026-06-23
by `7fb33d117`, the AY 2026-2027 rebuild that superseded them, and are
unbuildable regardless: `int_powerschool__gradebook_assignments_scores` now
exposes only `_dbt_source_project` while their `asg` CTE selects
`_dbt_source_relation`.

## Plan-doc correction — this batch does NOT gate macro removal

An earlier revision of the plan (added in #4547) asserted that leaving the
disabled callers in place would block deleting the macro. That is wrong, and the
plan doc is corrected here.

Verified empirically: with the macro definition deleted and all five disabled
callers left in place, `dbt parse --no-partial-parse` **exits 0**. The test is
meaningful — those models are genuinely parsed, appearing in
`manifest.disabled` (1,141 disabled nodes) — dbt simply does not render a
disabled model's Jinja deeply enough to fail on an undefined macro.

Macro removal is therefore gated only on #4549 and #4550 merging.

## AI Assistance

### dbt

- [x] Both analyses compile and dry-run clean against prod schemas.
- [x] Deletions confirmed unreferenced by any model, exposure, or config.
- [x] Distinct-grain safety on the `ps` CTE checked against prod before the
      swap.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Refs #3142


cbini:
## Review response

Three findings applied, one declined. The `SELECT DISTINCT` observation is a
better argument than the one I shipped and I've adopted it.

### 1 — Applied (`2e5523542`). This was the one worth catching.

I had noticed `qa_kippadb__hs_enrollment_audit` was also disabled and
deliberately left it as "pre-existing" — which was the wrong call, for the
reason you give: it leaves `test_dbt_assets_kipptaf` **still red**. Removing one
key because its model vanished while leaving another that fails identically is a
partial fix to a broken list, which is worse than either extreme.

Verified `qa_kippadb__hs_enrollment_audit` is `enabled: false`, and that the
three survivors — `snapshot_people__student_logins`,
`stg_people__student_logins`, `stg_powerschool__students` — are all enabled. The
selection is now correct **by inspection**.

Explicitly not claiming it passes: I did not execute it. `_test_dbt_assets`
calls `materialize()`, which writes snapshots to the warehouse, and
`tests/assets/` is not run in CI — not something to fire off casually to prove a
selection-list edit.

### 3 — Applied (`2e5523542`).

Blank line added between the `cc.` and `p.` groups in `course_bridge`. Agreed
the trigger is that line 338 is a touched line; otherwise it would have stayed
as-is.

### 4 — Applied; PR body corrected.

You're right and the phrasing mattered — "clears the last enabled call sites"
would send someone grepping `main` post-merge straight to a false bug report.
The summary now states that ~35 enabled sites remain, that they belong to #4549
and #4550, and that the macro is clear only once all three merge.

### 2 — Declining.

`collegeboard_ap_course_tagging.sql:31` is commented-out SQL in a file this PR
doesn't otherwise touch, and the column it names still exists, so there's no
correctness or staleness pressure. "Same file family" is a thin reason to widen
a diff — the prose comment I did update was in a file I was already editing, and
that's the line I'd rather hold.

### On the `SELECT DISTINCT` argument — yours is better, and I'm adopting it

You're right that my prod check (4 projects, 1 relation each) is nice-to-have
rather than load-bearing. `base_powerschool__student_enrollments` unions exactly
four `source()`s, so `_dbt_source_relation` takes exactly four values and the
regex maps them injectively — relation-to-project is 1:1 **by construction, in
every target**, not just in today's prod data. My row counts were a
point-in-time shadow of a structural guarantee.

That's the second time in this batch the structural form of an argument beat my
empirical one (the same applied to the iReady carve-out on #4550, where the
model *rewrites* `_dbt_source_relation` to the crosswalk region before any
consumer reads it). I'm carrying the generalization into the plan doc so future
swaps on a four-district union don't re-verify per site.

### Noted for later

`models/kippadb/qa/` being the last surviving `qa/` directory — both models
disabled by the same `fa62b5ae8` sweep — is a good forward pointer. Out of scope
here, but the convert-then-delete precedent applies cleanly whenever someone
picks it up.

PR #4553, opened 2026-07-24
refactor(dbt): retire union_dataset_join_clause macro and dedupe source-project derivations

cbini:
# Pull Request

## Summary & Motivation

Close out #3142 by deleting the `union_dataset_join_clause` macro and updating
every doc that taught it.

PRs #4547, #4549, #4550 and #4552 swapped all **263** call sites across **90**
files to a direct `a._dbt_source_project = b._dbt_source_project` equality.
Zero enabled call sites remain, so the macro is now dead code.

### What is in this PR

**Macro removal** — `union_dataset_join_clause` deleted from
`src/dbt/kipptaf/macros/utils.sql`. `extract_source_project` and
`extract_region` stay.

**Docs rewritten** to teach the column instead of the macro:

- `src/dbt/kipptaf/CLAUDE.md` — the `union_dataset_join_clause (critical)`
  section becomes `Cross-region joins (critical)`.
- `src/dbt/kipptaf/models/marts/CLAUDE.md` — the hash-and-join note.
- `docs/reference/dbt-conventions.md` — the SQL-conventions bullet and the
  `functions.region_join` UDF description.
- The plan doc records Batch 3, plus the by-construction 1:1 argument below.

While in those files I also corrected **four stale references to
`extract_code_location`**, a macro renamed to `extract_source_project` back in
Batch 1a and gone from the codebase since.

**Folded-in cleanups** (same expression-duplication theme, approved for this
batch):

| Model | Change |
| --- | --- |
| `rpt_powerschool__autocomm_students` | `regexp_extract(...) as code_location` becomes `se._dbt_source_project as code_location` |
| `rpt_powerschool__autocomm_students_iep` | same swap; not in the original queue, included so the sibling pair stays consistent |
| `dim_students` | hand-rolled `initcap(regexp_extract(s._dbt_source_relation, ...))` becomes `extract_region('s')` |
| `base_powerschool__student_enrollments` | `_dbt_source_project` and `code_location` were two byte-identical `regexp_extract` calls in one `select` |

### The one judgement call worth reviewing

Five macro calls survive, all inside the **disabled** pre-AY2627
gradebook-audit cluster: `int_tableau__gradebook_audit_assignments_teacher`
(x2), `_categories_teacher` (x2), `_assignments_student` (x1).

They do **not** block this PR — but my first justification for that was wrong,
and review caught it.

I originally argued "`dbt parse --no-partial-parse` exits 0 with the macro gone
and all five calls present, therefore disabled callers are harmless." A probe
kills that: a temporary **enabled** model calling the deleted macro also parses
clean. `dbt parse` renders Jinja only far enough to capture
`ref`/`source`/`config` and never resolves macros at all, so a clean parse says
nothing about whether a macro still has callers.

The actual evidence is two-part: `grep` proves zero enabled call sites, and
compilation — not parse — is where a missing macro fails, which the 195-model
`--empty` build exercises. Disabled models are never compiled, so their stale
calls cannot break a build. Corrected in the plan doc, in
`src/dbt/kipptaf/CLAUDE.md`, and in `src/dbt/CLAUDE.md`, where the same wrong
mechanism had already been merged.

They were left in place by explicit decision. That cluster is the pre-rebuild
lineage superseded by `rpt_tableau__gradebook_audit`, nothing outside it refs
those models, and migrating or deleting it is a gradebook-audit domain change
that deserves its own PR. The trade-off is real: re-enabling any of the three
now hits a compile failure. It is documented in `kipptaf/CLAUDE.md` so that
failure comes with an explanation rather than a mystery.

### Relation to project is 1:1 by construction

Several Batch 2 swaps were argued from a prod query showing one
`_dbt_source_relation` value per `_dbt_source_project`. That is point-in-time
evidence for a structural fact, and the structural version is stronger.

A cross-district union view unions exactly one `source()` per district, so
`_dbt_source_relation` takes exactly four values and
`regexp_extract(_dbt_source_relation, r'(kipp\w+)_')` maps them injectively
onto the four project names. Relation to project is 1:1 in **every** target,
not just in today's prod data, so the macro expansion and the column equality
are the same predicate. The generalization is now in the plan doc so future
swaps on a four-district union need no per-site re-verification.

The documented exceptions are unchanged: non-region unions, where the regex
yields null, and crosswalk-rewritten sources, where the model overwrites
`_dbt_source_relation` before any consumer reads it.

### On the base model reorder

BigQuery cannot reference a select alias within the same `SELECT`, so removing
the duplicate meant moving `code_location` to the final `select` as
`ar._dbt_source_project as code_location`. That places it after `region` in the
output schema instead of before it, and the shift propagates — the immediate
consumer `int_extracts__student_enrollments` is `select e.* except (...)`, so
four further `int_extracts__*` models inherit the new position. My original
claim that consumers enumerate their columns was wrong; most of that chain does
not.

It is still inert, for a different reason: nothing in the chain depends on
ordinal position. dbt contracts match by name, the district
`rpt_powerschool__autocomm_*` wrappers read `code_location` only as a filter
predicate and never project it, and the one order-sensitive construct —
`UNION ALL` — appears in four models reachable from the base model
(`int_extracts__student_enrollments_subjects`,
`int_assessments__course_enrollments`, `rpt_gsheets__gpa_flags_report`,
`rpt_gsheets__award_ceremony_gpa`), every branch of which enumerates its
columns, with zero wildcard selects among them.

### Added scope — dropped guidance for a scaffold that does not exist

Checking a review finding turned up that `scripts/dbt-yaml.py` is
non-functional. It shells out to `dbt run-operation generate_model_yaml`, a
macro from `dbt-labs/dbt-codegen` — a package declared in **no** `packages.yml`
in this repo and installed in no project's `dbt_packages/`. It is not a local
macro either. Confirmed by running it:

```text
dbt could not find a macro with the name "generate_model_yaml" in any package
```

Two docs pages instructed people to use it —
`docs/reference/dbt-conventions.md` (the "Model properties file" section) and
`docs/guides/google-sheets.md` step 6. Both now say to write the properties
file by hand, with an `INFORMATION_SCHEMA.COLUMNS` query for the column list.
The `scripts/CLAUDE.md` catalog entry is marked non-functional with the reason.

This is scope beyond the macro retirement, and normally I would split it. It is
here because `docs/reference/dbt-conventions.md` is already in this diff, the
change is docs-only, and the instructions it removes are actively wrong — anyone
following them hits an error. The alternative was adding `dbt-codegen` as a
dependency, which is a real decision rather than a review fix. The script file
itself is left in place, now unreferenced by any doc.

### Verification

- `grep` — zero enabled `union_dataset_join_clause` call sites. This plus the
  build below is the real proof; `dbt parse` is not (see above).
- `dbt build --empty --defer --favor-state` over
  `base_powerschool__student_enrollments+ dim_students+ autocomm_students+
  autocomm_students_iep+` — **195 models, 0 column-resolution failures**. One
  run reported an error on `rpt_tableau__gradebook_audit`; that was a stale
  local dev relation existing as a table where the model is a view, unrelated
  to this change, and it passed on a `--full-refresh` re-run.
- `trunk check --force --no-fix` on all 9 changed files — clean. The forced run
  caught a duplicate `code_location` alias (sqlfluff AL08 plus ST06) that the
  pre-commit hook's re-check had reported as clean; fixed before push.

### Still queued, deliberately not here

CRLF normalization on `tests/test_incorrect_student_number_pearson.sql`, the
`int_finance__enrollment_targets` yml column descriptions, the disabled
`models/kippadb/qa/` directory, and the disabled gradebook-audit cluster above.

Closes #3142

## AI Assistance

Claude Code authored the change. Human-directed: the decision to fold the three
queued cleanups into this batch rather than split them, and the decision to
leave the five disabled call sites in place rather than migrate or delete that
cluster. Claude flagged the re-enable trade-off before that call was made.

### General

      why.

### dbt

- [x] No new models, so no new properties files needed. `code_location` and
      `_dbt_source_project` already carry yml entries on every touched model.
- [x] No exposure changes — no model gains or loses a column.
- [x] **Not a breaking change.** No contracted column is renamed or removed.
      The only schema-visible effect is the `code_location` position shift on
      `base_powerschool__student_enrollments`, analyzed above.

### Docs

- [x] No schedule or sensor changes, so no automations-catalog regeneration.
- [x] CLAUDE.md updates are the substance of this PR rather than a side effect.

## CI checks

- [ ] **Trunk**
- [ ] **dbt Cloud**
- [ ] **Dagster Cloud**


cbini:
## Review response

Six applied, one declined, pushed in `e45c40421`. Item 7 was the important one —
it turned out to be worse than described, and it invalidated a line already
merged to `main`.

### 7 — Applied, and the mechanism is worse than stated

You said the outcome is right but `dbt parse` exiting 0 is not evidence specific
to disabled nodes. I probed it rather than take it on faith: added a temporary
**enabled** model whose body calls the deleted macro, ran
`dbt parse --no-partial-parse`, and it **exits 0**. So parse does not
discriminate on enabled-vs-disabled at all — it renders Jinja only far enough to
capture `ref`/`source`/`config` and never resolves macros.

That is a bigger correction than a reword. My "verified, not assumed" claim in
the PR body rested on a test that cannot distinguish the two cases. The real
evidence was always the other two things I did: `grep` proving zero enabled call
sites, and the 195-model `--empty` build, which exercises compilation — the
actual gate.

The same wrong mechanism had already merged into `src/dbt/CLAUDE.md` (in
`8138df67e`), where it also recommended `dbt parse --no-partial-parse` as the
confirmation step. That is actively misleading — it would tell a future agent
that a clean parse proves a macro removal is safe. Rewritten there, in
`src/dbt/kipptaf/CLAUDE.md`, in the plan doc, and in the PR body.

### 5 — Applied, and your verification was one step short

You are right that "every downstream consumer names its columns explicitly" is
false — `int_extracts__student_enrollments.sql:104` is `select e.* except (...)`
and the position propagates through four further `int_extracts__*` models.

Your conclusion is also right, but I could not confirm it as written. You said
you "found no `union all` against the base model or any of its `select *`
descendants." There are four: `int_extracts__student_enrollments_subjects`,
`int_assessments__course_enrollments`, `rpt_gsheets__gpa_flags_report`, and
`rpt_gsheets__award_ceremony_gpa` all `ref()` the base model and contain
`UNION ALL`. They are safe for a narrower reason — every branch of all four
enumerates its columns, with zero wildcard selects across the four files. Inert,
but by a checkable property rather than absence.

Both the plan doc and the PR body now carry that version.

### 2, 3, 4, 6 — Applied

- **2** — the macro claim is now "inline `regexp_extract` on a bare `select *`
  view, or the `extract_source_project()` macro otherwise", pointing at
  `kipptaf/CLAUDE.md` for which applies. As written it contradicted the very
  convention it was summarizing.
- **3** — `region_join` row marked deprecated, removed from the code example,
  and a short note added explaining it is the SQL-UDF twin of the retired macro
  with no call sites. Kept the row rather than deleting it, since the table is
  an inventory of what exists in the `functions` dataset and the UDF does still
  exist. Dropping it from the warehouse is warehouse-side, agreed.
- **4** — `docs/models/gradebook-audit-data-model.md:73` fixed. My sweep covered
  `docs/reference/` and missed `docs/models/`; the remaining hits under
  `specs/` and `plans/` are historical and stay.
- **6** — all three yml entries moved to match the new SQL order. Worth doing on
  its own terms, but I could not confirm your stated premise, and checking it
  turned up something else — see below.

### 1 — Declining

`se.code_location` would work — `int_extracts__student_enrollments` does expose
it, as you say. I am keeping `se._dbt_source_project as code_location`.

`code_location` exists on that model only because it is inherited from the base
model, and it exists on the output only because the district wrappers' contract
requires that name. `_dbt_source_project` is the canonical column. Reading the
canonical column and renaming at the output boundary keeps the legacy name
confined to the boundary that actually requires it; reading `se.code_location`
propagates the legacy name one layer further inward.

It also points the right way for the follow-up you identified. When
`code_location` is retired, this line survives as-is until the wrapper's `where`
clause is swapped, whereas `se.code_location` breaks.

The asymmetry you flagged is real but resolves the same way: both files now read
`_dbt_source_project`, and `_iep` happens to have no alternative.

### On the follow-up

Recorded in the plan doc under Batch 3 execution findings — the duplicated
expression is gone but the duplicated column survives, side by side from
`base_powerschool__student_enrollments` through five `int_extracts__*` models
into two extracts, with four district wrappers filtering on `code_location`.
Agreed it is its own PR for the same reason as the gradebook-audit cluster.

### Surfaced while checking item 6 — the yml regeneration path is broken

Your premise was that `scripts/dbt-yaml.py` shells out to `generate_model_yaml`,
which reads relation order, so the next regeneration would emit churn. The
script does call it (`scripts/dbt-yaml.py:59`), but `generate_model_yaml` comes
from `dbt-labs/dbt-codegen`, and that package is **not declared in any
`packages.yml` in this repo** and not installed in any project's
`dbt_packages/`. It is not defined as a local macro either. So the script fails
today with a macro-not-found error, and
`docs/reference/dbt-conventions.md` documents both it and a bare
`dbt run-operation generate_model_yaml` as the way to scaffold model YAML.

I applied the reorder anyway — yml mirroring relation order is right regardless,
and it costs nothing. But the churn you predicted cannot currently happen,
because nobody can run the tool.

Adding `dbt-codegen` to `packages.yml` is a dependency decision rather than a
review fix, so I have not made it here. Flagging for a follow-up issue.

### On the lint note

Worth having, and I verified it: `.trunk/config/.sqlfluff` sets
`templater = jinja` with `ignore = templating, parsing`, so the five dangling
calls stay green under `trunk check`. That reinforces item 7 rather than
softening it — neither parse nor lint catches a deleted macro. Only compilation
does.

PR #4555, opened 2026-07-24
chore(dbt): retire disabled kippadb qa models and document enrollment targets

### LINKED ISSUE #4554: chore(dbt): retire disabled kippadb qa models and complete enrollment-targets descriptions
Two small cleanups surfaced while finishing #3142.

**Remove `src/dbt/kipptaf/models/kippadb/qa/`** — `qa_kippadb__duplicate_enrollments` and `qa_kippadb__hs_enrollment_audit` are both `enabled: false` and have zero references anywhere in the repo. This is the last surviving `qa/` directory; the edplan and powerschool ones were retired in #4552, which also dropped the stale `hs_enrollment_audit` asset key from `tests/assets/test_assets_dbt.py`.

**Complete `int_finance__enrollment_targets` column descriptions** — 15 of its 17 columns carry a `data_type` but no `description`. Only the model itself and `_dbt_source_project` are documented, and the repo convention requires a description on every column of a modified model.

Not included: CRLF line-ending normalization, considered and dropped. It affects 70 tracked `.sql` files (66 models, 1 macro, 3 singular tests) with no `.gitattributes` in the repo. Normalizing changes each file's checksum, which marks the models `state:modified` and bumps their dagster-dbt code versions, so a full pass would trigger a near-total kipptaf CI rebuild and prod rematerialization for zero behavior change. Files get normalized opportunistically when edited for other reasons.

cbini:
# Pull Request

## Summary & Motivation

Clear two small items left over from the #3142 work.

### Remove `src/dbt/kipptaf/models/kippadb/qa/`

`qa_kippadb__duplicate_enrollments` and `qa_kippadb__hs_enrollment_audit` are
both `enabled: false` and have zero references anywhere in the repo — the only
grep hits were their own `name:` lines in their properties files. This was the
last surviving `qa/` directory; the edplan and powerschool ones went in #4552,
which also removed the stale `hs_enrollment_audit` key from
`tests/assets/test_assets_dbt.py`.

### Describe every `int_finance__enrollment_targets` column

15 of its 17 columns carried a `data_type` and no `description`. Only the model
itself and `_dbt_source_project` were documented.

Writing these turned up four things worth having in the docs rather than
rediscovering later, each verified against prod:

- The model's join to `stg_powerschool__schools` is an **inner** join, so a
  sheet row whose `schoolid` has no PowerSchool match is silently dropped. Two
  rows are dropped today.
- `financial_model_enrollment` is **not** a derived total. It matches
  `target_enrollment` plus `adjustment` net of `attrition_factor` on roughly
  half the rows and diverges on the rest, so it is an independent sheet input.
- `region` is a free-typed sheet label, blank on many rows, and is not the
  authoritative district. `_dbt_source_project` is.
- `academic_year` is derived upstream as `fiscal_year` minus one and stores the
  starting year.

### One thing I could not document, and did not guess at

The five financial-model factor columns — `grade_band_ratio`,
`at_risk_and_lep_ratio`, `at_risk_only_ratio`, `lep_only_ratio`, `sped_ratio` —
have no recoverable definition. They are populated on only a handful of rows,
all from a single early academic year, and the introducing commit is
`3c2977f03 mv dbt`, a bulk move with no rationale. Nothing downstream defines
them either. Their only consumer, `rpt_tableau__ops_dashboard`, aggregates them
with `max()` per group and surfaces them to Tableau with no descriptions of its
own.

The column names read like population shares, but the data rules that out —
three of them sum to well above one, and `sped_ratio` holds a single constant
value on every row where it appears. So they are more likely weights or
multipliers than shares.

Rather than invent a plausible-sounding definition, the descriptions state what
each column is and that its definition is maintained in the finance sheet.
**If someone on the Finance side can confirm what these five actually
represent, that is worth a follow-up commit** — the current descriptions are
honest but thin.

Review correctly pushed back on an earlier draft here: it described how often
each column was blank, which snapshots the current contents of a live Google
Sheet rather than stating a rule, and nothing in CI would catch the drift. The
descriptions now state the structural reason (optional sheet input) instead.

### Considered and dropped — CRLF normalization

This was on the cleanup list as a single file. It is actually 70 tracked `.sql`
files (66 models, 1 macro, 3 singular tests) with no `.gitattributes` in the
repo.

Normalizing changes each file's checksum, which marks the models
`state:modified` and bumps their dagster-dbt code versions, so a full pass would
trigger a near-total kipptaf CI rebuild and a prod rematerialization wave for
zero behavior change. The macro is worse — every model calling `parse_html`
would mark modified too. Normalizing only the one file that started the list
would leave 69 and resolve nothing.

Trunk and sqlfluff both pass on CRLF today, so this is cosmetic. Files get
normalized opportunistically when edited for other reasons. Written up in #4554
so the reasoning is recorded rather than rediscovered.

### Added during review — a test for the silent drop

Review flagged that documenting the inner-join drop does not tell anyone when a
_new_ sheet row starts falling out. Added
`int_finance__enrollment_targets__schoolid_resolves`, matching the existing
`*_crosswalk_resolves` singular-test pattern. Validated against prod: warns on
exactly the two known rows, both from a closed school in AY2016.

### On the deleted QA models

Both were `enabled: false`, so they had not been detecting anything — deleting
them loses no active signal. But the condition one of them looked for is still
live: the `qa_kippadb__duplicate_enrollments` predicate currently matches **847
overlapping enrollment pairs across 374 students**, and
`int_kippadb__enrollment_pivot` resolves those with
`row_number() … order by start_date, actual_end_date`, so which enrollment wins
is order-dependent. A real test in the
`dim_student_*__no_overlapping_spans` shape would be the right way to carry that
signal forward, but at 847 pairs it needs a severity decision and likely
upstream Salesforce cleanup first. Deliberately left as a known, untracked gap
for now — recorded here so the deletion is not mistaken for a clean bill of
health.

### A note on CI

The dbt Cloud check on this PR is **trivially green, not validation.** A
properties-yml `description` change does not mark a model `state:modified`, and
the two deleted models were disabled, so `state:modified+` selects nothing.
Verification here was `dbt parse --no-partial-parse` (clean, confirming no
dangling refs to the deleted models) and `trunk check --force` on the yml.

Closes #4554

## AI Assistance

Claude Code authored the change. Human-directed: scoping both items into one PR
and the decision to drop CRLF normalization after seeing the real blast radius.

### General

### dbt

- [x] No new models. The removed pair took their properties files with them.
- [x] No exposure changes — neither deleted model appeared in any exposure, and
      no column was added or removed.
- [x] **Not a breaking change.** Both deleted models were `enabled: false` with
      no references; the other change is descriptions only.

## CI checks

- [ ] **Trunk**
- [ ] **dbt Cloud**
- [ ] **Dagster Cloud**


cbini:
## Review response

Six applied, one verified as already correct, two deferred with reasons. Pushed
in `509fe49fb`. Your question about the deleted QA signal turned out to be the
most valuable item — the answer is not the one I assumed.

### 1 — Applied. Test added and validated

`src/dbt/kipptaf/tests/int_finance__enrollment_targets__schoolid_resolves.sql`,
in the shape you pointed at, with the `properties.yml` entry and
`config.meta.dagster.ref`.

Your caveat was right — it fails today. Ran it against prod
(`dbt test --defer --favor-state`): **WARN=1, ERROR=0, 2 failure rows**, both
grade rows of a single closed school from AY2016. Project default is
`+severity: warn` with `store_failures`, so it lands as a warn and the rows
persist in the audit view, matching the crosswalk-test precedent. No severity
override needed. The standing exception is named in the test description so it
does not read as a broken test.

### 2, 3, 5, 6 — Applied

You are right that I encoded observations of a live Google Sheet as if they were
rules. Nothing guards them, and the factor columns are exactly the case where
Finance resuming data entry would silently falsify the docs.

- **2** — every frequency claim is gone. "Blank on the large majority of rows"
  and friends are now "Optional manual correction… Blank when no correction was
  recorded" — the structural reason, per the
  `int_assessments__resolved_section_enrollments.yml:56` precedent.
- **3** — the `sped_ratio` constant-value claim is deleted outright. Agreed it
  read as a contract; it was a point-in-time observation of 13 rows.
- **5** — "not documented in this repo" is gone everywhere. The model
  description now says the definitions are maintained in the finance sheet.
- **6** — the shared caveat is hoisted into the model `description:`, which
  names the five factor columns once. Each column description is now one
  sentence about its own meaning. No cross-references left.

### 4 — Applied, though the arithmetic did hold

Clause cut. One correction to the reasoning, since it may come up again: if
`at_risk_and_lep_ratio` is the intersection and the other two are the
exclusive-only sets, those three still partition the at-risk ∪ LEP population
and should sum to at most one. They sum to about 1.84, so the shares reading is
ruled out under your interpretation too.

Cutting it anyway, for your item-5 reason rather than this one — a column
description should state what the column means, not carry my inference about
what it does not mean.

### 7 — Verified, no change

Checked: `select distinct grade_level` returns 0 through 12 only, no negatives.
The finance sheet carries no PK rows, so "0 is kindergarten" is complete here
and the fuller `int_extracts__student_enrollments` precedent does not apply.

### 8 — Deferring

Agreed the sheet-input semantics ultimately belong on
`stg_google_sheets__finance__enrollment_targets`. Out of scope for #4554, which
is scoped to the int model, and the staging model has its own wrinkle — TitleCase
sheet-header column names — that deserves its own change rather than being
folded into a cleanup PR. Reasonable follow-up.

### The deleted QA models — you asked the right question, and I had it wrong

I claimed the deletion was safe because nothing references those models. True,
but it dodged your actual question, so I measured the condition instead of
asserting it:

```text
847 overlapping enrollment pairs across 374 distinct students
```

(the `qa_kippadb__duplicate_enrollments` predicate, run against
`stg_kippadb__enrollment` in prod). That is not a dormant historical condition —
it is live and substantial, and `int_kippadb__enrollment_pivot`'s
`row_number() … order by start_date, actual_end_date` is exactly the surface you
identified.

What keeps the deletion defensible is something neither of us said: both models
were `enabled: false`, so they have not detected anything for as long as they
have been disabled. Deleting them loses no active signal — the signal was
already off. But "we deleted a disabled model" and "this condition is fine" are
different claims, and I was implying the second while only having evidence for
the first.

The right resolution is a real test on `int_kippadb__enrollment_pivot` in the
`dim_student_*__no_overlapping_spans` shape you cited, not a resurrected QA
view. I am not adding it here — at 847 pairs it needs a decision about severity
and whether the upstream kippadb (Salesforce) data gets cleaned first, which is
Ops work. Proposing it as a follow-up issue rather than silently dropping it.

### The union trap — real, and pre-existing

Verified your reading of `rpt_tableau__ops_dashboard.sql`. Branch one is
`grade_band, at_risk_only, at_risk_and_lep, lep_only, sped`; branch two is
`grade_band, at_risk_and_lep, at_risk_only, lep_only, sped`. Positions two and
three are crossed, and `UNION ALL` is positional.

Harmless today for the reason you gave — branch two is all `null` and the output
names come from branch one. `git log -L` puts the ordering in `a74a0f8f3`, the
commit that created the model, so it predates my `_dbt_source_project` swap in
#4549 rather than being a regression from it.

Not fixing it in this PR: it is a different file from either item in #4554, and
this PR's dbt Cloud run is currently a trivially-green no-op. Touching a model
with descendants turns that into a real rebuild, which is worth doing
deliberately rather than as a review drive-by. Happy to send it as a one-line
follow-up.

Your `max()` note is also correct and my PR body was sloppy —
`rpt_tableau__ops_dashboard.sql:66-70` aggregates the factor columns with
`max()` per group rather than passing them through. Body corrected.

PR #4560, opened 2026-07-27
perf(dagster): cron automation condition for high-cost eager dbt tables

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will move five expensive dbt tables off the eager (rebuild-on-any-upstream-update) automation condition onto a nightly cron cadence, cutting an estimated ~40% of the July BigQuery analysis bill:

- `int_topline__student_metrics` — 19 eager upstreams drove 44-180 full ~7 GiB rebuilds/day (~15.8 TiB in July, 29% of the bill); its only consumer chain is the topline cascade dashboard, delivered by nightly extract schedules (01:15-03:00 ET).
- `fct_assessment_scores_enrollment_scoped`, `dim_assessments`, `dim_assessment_administrations` — Cube-only consumers, whose sole pre-aggregation refresh key checks once a day; the 125+/day rebuilds were invisible to it.
- `int_assessments__response_rollup` — all 13 consumers are nightly extracts or the models above.

Implementation:

1. `dbt_cron_automation_condition(cron_schedule, cron_timezone)` in `core/automation_conditions.py`, built from the shared `_build_dbt_condition()` skeleton with `cron_tick_passed` as the trigger. Deliberately NOT stock `on_cron()`: its all-deps-updated-since-cron gate never fires when any mixed-cadence upstream (e.g. a weekly model) skips the window. `newly_missing` and `code_version_changed` triggers are retained, so deploys still rebuild immediately.
2. A per-model override in `CustomDagsterDbtTranslator.get_automation_condition`, read from model YAML meta before the materialized-type fallthrough:

   ```yaml
   config:
     meta:
       dagster:
         automation_condition:
           cron_schedule: 0 0 * * *
   ```

3. The five models set to `0 0 * * *` America/New_York — before the extract window and Cube's daily refresh-key check. Same-tick requests batch into one run, so dbt orders `int_assessments__response_rollup` ahead of the star. Downstream eager tables (e.g. `rpt_tableau__topline_cascade_dashboard`) cascade once per tick through the intermediate view.

## AI Assistance

Claude Code performed the cost investigation (INFORMATION_SCHEMA job analysis), authored the condition/translator code, tests, and YAML changes; human-directed scope, model selection, and cadence.

### General

- [ ] Update **due date** and **assignee** on the [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
[...]

Issue #4559, opened 2026-07-27
perf(dagster): cron-based automation condition for high-cost eager dbt tables


cbini:
## Problem

The materialized-as-table automation condition (`dbt_table_automation_condition`) is eager: any upstream data refresh triggers a rebuild. For wide, expensive tables this produces 44–180 full rebuilds/day whose freshness no consumer uses:

- `int_topline__student_metrics`: 19 eager upstreams, ~7 GiB full scan per rebuild, 44–180 rebuilds/day — **29% of the July BigQuery bill (~15.8 TiB, ~$98)**. Sole consumer chain: `int_topline__dashboard_aggregations` to `rpt_tableau__topline_cascade_dashboard`, delivered by nightly extract schedules (01:15–03:00 ET).
- Assessment star (`fct_assessment_scores_enrollment_scoped`, `dim_assessments`, `dim_assessment_administrations`): terminal in dbt, read only by Cube, whose sole pre-aggregation has `refresh_key: every: 1 day` — intraday rebuilds are invisible to Cube consumers. Cost jumped from ~0 to ~500 GiB/day after ec22276ff flipped them from views to tables (Refs #4464).
- `int_assessments__response_rollup`: 13 consumers, all scheduled extracts or the models above. ~224 GiB/day.

The spike showed up as $30–34/day on Jul 22–24 vs a ~$10/day baseline; steady-state residue is roughly +$55/month on the assessments chain plus the pre-existing ~$98/month on `int_topline__student_metrics`.

## Proposal

1. Add `dbt_cron_automation_condition(cron_schedule)` in `src/teamster/core/automation_conditions.py`, built from the existing `_build_dbt_condition()` skeleton with `AutomationCondition.cron_tick_passed(cron_schedule)` as the trigger. NOT stock `AutomationCondition.on_cron()` — its `all_deps_updated_since_cron` gate never fires when any of the 19 mixed-cadence upstreams skips the window. `cron_tick_passed` is unconditional; `newly_missing` and `code_version_changed` triggers and the deps-missing/in-progress guards are retained, so deploys still rebuild immediately.
2. Support a per-model override in `CustomDagsterDbtTranslator.get_automation_condition`, read from dbt model YAML meta, checked before the materialized-type fallthrough:

   ```yaml
   config:
     meta:
       dagster:
         automation_condition:
           cron_schedule: 0 4 * * *
   ```

3. Apply to the five models above, with crons timed before the extract window and Cube's daily refresh-key check.

[...]

PR #4562, opened 2026-07-27
perf(dbt): flip topline weekly views to nightly-cron tables

### LINKED ISSUE #4153: perf(dbt): int_topline__student_metrics fragile to BigQuery straggler re-execution (deep multi-stage plan)
## Problem

`int_topline__student_metrics` (kipptaf `topline`, materialized `table` via `create or replace table kipptaf_topline.int_topline__student_metrics`) is normally a ~43-second BigQuery job, but its query plan is unusually deep (~660 execution stages). On 2026-06-09 a single execution degraded to **39.7 minutes**, consuming the entire 1800s `max_runtime` budget of the kipptaf `__ASSET_JOB` run and forcing a failure + auto-retry.

This was found during the 2026-06-09 day-2 ops review (run `c7669a73-7b8f-4f88-a6e6-a08a74a84adc`).

## Evidence

Six executions of the identical CTAS in one hour (BigQuery `region-us` JOBS):

| Job (UTC start) | Wall | Slot-ms | Bytes scanned | Records read | Shuffle | Spill | Stages |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 10:25:09 | 49s | 5.2M | 9.44 GB | — | — | 0 | — |
| 10:30:37 | 43s | 4.9M | 9.44 GB | 1.039B | 63.5 GB | 0 | 661 |
| **10:34:10 (degraded run)** | **2381s** | **110.3M** | **9.44 GB** | **1.187B** | **73.7 GB** | **0** | **660** |
| 11:05:06 (retry) | 42s | 4.8M | 9.44 GB | 1.029B | 64.3 GB | 0 | 662 |
| 11:09:52 | 110s | 9.6M | 9.44 GB | — | — | 0 | — |
| 11:14:26 | 40s | 5.2M | 9.44 GB | — | — | 0 | — |

Key points:

- **Same query, same plan, same data** — record counts within ~15%, identical bytes scanned, identical ~660-stage plan, zero spill in every run.
- The degraded run consumed **~22x the slot-ms** for the **same output** — i.e. shards did redundant work (BigQuery speculatively re-runs slow shards; re-executions bill slot-ms but the final result is unchanged).
- The job timeline shows `active_units` averaging ~40-160 with no slot starvation, plus a distinct ~460s dead phase at the end (all units complete, job not finalized) — consistent with a straggler / stalled final commit.
- **Slot contention was tested and rejected:** the model runs on-demand (`reservation_id = null`); the degraded 10:34 window had *less* competing slot demand than the 11:05 window where the retry ran in 42s.

Conclusion: the trigger was a transient BigQuery-side slow execution. We cannot prevent the straggler, but a ~660-stage plan amplifies its impact — many sequential stage boundaries multiply the cost of any single slow/re-executed shard, turning a transient blip into a `max_runtime` failure.

## Impact

- Run-level `max_runtime` (1800s) failures whenever this model has a bad execution; the model alone can consume the whole budget.
- Self-heals via auto-retry today, but adds noise and delays the topline dashboard refresh by ~30+ min when it fires.

## Proposed direction

- Inspect the compiled SQL for `int_topline__student_metrics` and flatten / pre-materialize the deepest CTE chains (the 660-stage plan suggests many stacked joins/unions/window steps that could be split into intermediate models).
- Consider materializing the heaviest sub-results as their own intermediate models so a straggler in one stage has a smaller blast radius and the overall plan depth drops.
- Optionally evaluate whether the topline dbt assets warrant a BigQuery reservation/autoscale rather than on-demand, to bound execution-time variance.

## Notes

This issue is about the model's query-shape fragility. The same incident also surfaced a separate operational gap — Dagster terminating the run did **not** cancel the in-flight BigQuery job, which ran ~10 min past the kill and overwrote the retry's output. That is being investigated separately and may get its own issue.

cbini:
# Pull Request

## Summary & Motivation

Flip the 14 view-materialized `int_topline__*` weekly models to `materialized: table` and put all 18 topline weekly models (the 14 flips plus the 4 already-table siblings: star, state assessments, truancy, suspension) on the nightly `dbt_cron_automation_condition` (`0 0 * * *`), matching the cron already on `int_topline__student_metrics`.

Refs #4153

**Why:** `int_topline__student_metrics`' ~660-stage BigQuery plan (the straggler fragility in #4153) comes from view inlining — ~12 weekly views expanded per reference (up to 4x each), each internally re-inlining the `int_extracts__student_enrollments_weeks` spine view (~14 expansions total in one CTAS). With the weeklies as tables, each weekly builds as its own small shallow job and the `student_metrics` CTAS reads tables plus only its 3 direct spine refs — plan depth collapses, and a transient straggler's blast radius shrinks to one small job.

**Why cron and not eager tables:** eager `dbt_table_automation_condition` on 18 new tables would multiply the #4559 cost problem (44–180 rebuilds/day whose freshness nothing consumes). All weeklies share `student_metrics`' midnight tick; the condition's deps-in-progress guard serializes the pass (weeklies build first, then `student_metrics`), finishing well before the 01:15–03:00 ET extract window and the 05:00 Tableau refresh.

**Scope deliberately excludes** (consumer-cadence analysis on #4153):

- `int_extracts__student_enrollments_weeks` / `_subjects_weeks` — also feed the DDI suite (5 refreshes/day incl. afternoon ticks) and iready APM; stay views.
- `int_powerschool__ps_adaadm_daily_ctod` — feeds the hourly ops dashboard and 15:00 attendance dashboard; stays a view.
- `int_topline__attendance_contacts` / `int_topline__school_community_diagnostic` helpers — attendance comm log refreshes at 15:00; stay views (their compute now materializes once nightly inside their weekly children).

**Freshness note:** `rpt_gsheets__school_metrics_extract` (connected sheet, no Dagster schedule) reads `ada_running_weekly` and `gpa_term_weekly`; after this change its data is as-of the last midnight tick — acceptable for a weekly-metrics extract.

## AI Assistance

Claude Code authored the analysis and changes end-to-end (issue diagnosis, consumer-cadence tiering, YAML edits); human-directed scope decisions (sequencing after #4559, cron approach, which models to exclude).

### General

### dbt

- [x] Properties files updated for all 18 models (one created: `int_topline__dibels_benchmark_weekly.yml`)
- [x] No exposure changes needed — no new external consumers; existing topline cascade exposure unchanged
- [x] No breaking change — no column renames/removals; materialization + automation metadata only
- [x] No external sources modified

### Docs

- [x] No schedule/sensor changes (automation condition comes from model YAML meta via the #4560 translator override)

## CI checks

- [ ] **Trunk** — passes (`trunk check --force` on all 18 files: no issues)
- [ ] **dbt Cloud** — passes (`state:modified+` will build all 18 flipped models plus `int_topline__student_metrics` downstream)

### DIFF
--- a/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__ada_running_weekly.yml
+++ b/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__ada_running_weekly.yml
@@ -1,2 +1,11 @@
 models:
   - name: int_topline__ada_running_weekly
+    config:
+      materialized: table
+      meta:
+        dagster:
+          # table + nightly cron flattens the ~660-stage view-inlined plan of
+          # int_topline__student_metrics without eager rebuilds; shares its
+          # midnight tick (deps-in-progress guard serializes). Refs #4153
+          automation_condition:
+            cron_schedule: 0 0 * * *
--- a/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__attendance_contacts_weekly.yml
+++ b/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__attendance_contacts_weekly.yml
@@ -1,2 +1,11 @@
 models:
   - name: int_topline__attendance_contacts_weekly
+    config:
+      materialized: table
+      meta:
+        dagster:
+          # table + nightly cron flattens the ~660-stage view-inlined plan of
+          # int_topline__student_metrics without eager rebuilds; shares its
+          # midnight tick (deps-in-progress guard serializes). Refs #4153
+          automation_condition:
+            cron_schedule: 0 0 * * *
--- a/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__attendance_interventions_weekly.yml
+++ b/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__attendance_interventions_weekly.yml
@@ -1,2 +1,11 @@
 models:
   - name: int_topline__attendance_interventions_weekly
+    config:
+      materialized: table
+      meta:
+        dagster:
+          # table + nightly cron flattens the ~660-stage view-inlined plan of
+          # int_topline__student_metrics without eager rebuilds; shares its
+          # midnight tick (deps-in-progress guard serializes). Refs #4153
+          automation_condition:
+            cron_schedule: 0 0 * * *
--- a/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__college_matriculation_weekly.yml
+++ b/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__college_matriculation_weekly.yml
@@ -1,2 +1,11 @@
 models:
   - name: int_topline__college_matriculation_weekly
+    config:
+      materialized: table
+      meta:
+        dagster:
+          # table + nightly cron flattens the ~660-stage view-inlined plan of
+          # int_topline__student_metrics without eager rebuilds; shares its
+          # midnight tick (deps-in-progress guard serializes). Refs #4153
+          automation_condition:
+            cron_schedule: 0 0 * * *
--- a/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__deanslist_incentives_weekly.yml
+++ b/src/dbt/kipptaf/models/topline/intermediate/properties/int_topline__deanslist_incentives_weekly.yml
@@ -1,4 +1,11 @@
 models:
   - name: int_topline__deanslist_incentives_weekly
-    # config:
-    #   materialized: table
+    config:
+      materialized: table
+      meta:
+        dagster:
+          # table + nightly cron flattens the ~660-stage view-inlined plan of
+          # int_topline__student_metrics without eager rebuilds; shares its
+          # midnight tick (deps-in-progress guard serializes). Refs #4153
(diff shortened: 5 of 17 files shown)


cbini:
Responding to the claude-review findings:

1. **Runtime headroom (medium)** — agreed it's worth measuring, not assuming. Expectation going in: the weeklies' compute previously ran inline inside a 43s `student_metrics` CTAS, so 18 parallel small jobs plus a flat final CTAS should land far under the 1800s run cap; the 900s-per-job worst case would require the same transient-straggler event #4153 documents, which is exactly what this PR shrinks the blast radius of. Will check per-model timings on the first nightly run in the Dagster UI rather than burning a dev build now.

2. **Uniqueness tests (medium)** — valid, and taking the review's own recommendation: follow-up issue #4563 rather than expanding a perf PR into 18 grain validations. The flip actually makes those future tests meaningful — as nightly tables the tests will run on cadence instead of never (views).

3. **Effect lands at first midnight tick (low)** — correct, and fine: no `.sql` checksum changed so `code_version_changed` won't fire at deploy, `newly_missing` won't either (the view relations exist). The BigQuery objects stay views (serving correctly) until the first `0 0 * * *` ET tick tonight, when the CTAS flip happens. Noting it here as the handoff record: don't judge the deploy by BigQuery object types before ~00:30 ET.

PR #4578, opened 2026-07-27
feat(dbt): add contact_2 (Parent 2) slot to finalsite student contacts

### LINKED ISSUE #4569: feat: DeansList contacts.txt — ContactType column and Parent 2 from household 1
Follow-up to #4400. Two change requests from Ops (Asana subtasks under the #4400 task), due together:

1. **New `ContactType` column** ([Asana](https://app.asana.com/1/913513768672/task/1216916256764022)) — DeansList supports an additional field distinguishing guardians from emergency contacts so only guardians receive automated reports/texts/emails. Values: `Parent1` / `Parent2` / `Emergency`, positioned between `Relationship` and `Language`. DeansList is holding off importing the emergency contacts we already send until this field exists.
2. **Include Parent 2 from household 1** ([Asana](https://app.asana.com/1/913513768672/task/1216916256764023)) — today the NJ Finalsite contact feeds carry one parent (the `contact_1` primary/financial pick) plus emergency contacts. Add the second parent, scoped to the student's household 1; households 2+ stay excluded. Applies to all NJ contact models reading `int_finalsite__student_contacts`, not just DeansList.

## Parent 2 definition (validated against Newark data)

Finalsite encodes Parent 1 as the relationship flagged `primary` and Parent 2 as a separate related contact flagged `financial` (not primary), almost always co-resident in household 1. The `is_parent2` custom field lives on the STUDENT record ("this student has a Parent 2"), not on the parent contact. The `Household` API object is id + address only — membership carries no role; parent semantics live on relationships and student custom fields.

Agreed gate (most conservative): **`contact_2` = the non-primary `financial` related contact, required to be a member of the student's household 1, AND the student's `is_parent2` flag is true.** Newark: 3,130 students (vs 3,374 household-only; 171 flagged students have the contact outside household 1 and are excluded).

## Design sketch

- `stg_finalsite__contacts` (finalsite package): surface household ids (household 1 id for the student side; membership array for the contact side) — currently only `households[0]` address fields are kept.
- `int_finalsite__student_contacts` (finalsite package): add a `contact_2` slot per the gate above. Must not double-pick the `contact_1` row (a student with no `primary` gets their `financial` contact as `contact_1`).
- `rpt_deanslist__family_contacts` (kipptaf): add `ContactType` from `contact_slot` (`contact_1` → `Parent1`, `contact_2` → `Parent2`, `emergency_*` → `Emergency`), before `Language`.
- Check downstream consumers of `int_finalsite__student_contacts` for grain/slot assumptions before the new rows land.
- Rollout: two-PR pattern (finalsite package first, then kipptaf consumer) with per-district staging seeding, per #4400.

Open question for DeansList: their sample shows emergency rows with `Relationship = Emergency1`; we currently emit `Emergency 1`. Once `ContactType` exists, the real relationship (`emrg_N_relationship_ss`) could go in `Relationship` instead — confirm with DL before changing.

cbini:
# Pull Request

## Summary & Motivation

Add a `contact_2` (Parent 2) slot to `int_finalsite__student_contacts`, per Ops request for the NJ Finalsite contact feeds (Refs #4569, first of two PRs — the kipptaf `ContactType` consumer change follows).

Finalsite encodes the second parent as an additional relationship flagged `financial` without `primary` (`primary` is a per-student singleton — Parent 1). The pick is scoped to the most conservative read of the request:

- the student's own `is_parent2` custom field must be true, AND
- the related contact must be a member of the student's first household ("household 1") — second parents in other households are intentionally excluded, and
- the row already picked as `contact_1` (the financial fallback when no primary is set) is skipped.

To support the household gate, `stg_finalsite__contacts` now surfaces `household_1_id` and `household_ids` (Finalsite households carry only an id and address — no membership roles — so parent semantics come from relationships and custom fields; validated against the Finalsite Enrollment external API spec).

Validated on Newark dev build: 3,216 `contact_2` rows, one per student; 0 gate violations (every row's student has `is_parent2` true and the contact co-resides in household 1). Grain, contract, accepted-values, and both unit tests (including a new Parent 2 pick test with an outside-household exclusion case) pass.

## AI Assistance

Claude Code authored this PR end-to-end (data archaeology, model changes, unit tests, validation), directed by @cbini (gate definition and scope decisions).

### General

### dbt

- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — n/a, no new external consumers; the DeansList extract change ships in the follow-up kipptaf PR
- [x] **Breaking change?** No — additive only: two new staging columns, one new `contact_slot` value. Downstream `contact_slot = 'contact_1'` branches are unaffected; the kipptaf pivot model already carries an (until now NULL) `contact_2` column surface.
- [x] External sources unchanged — no re-staging needed (`households` is already in the ingested Avro).

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes (expected trivial: package-only PR, kipptaf selects no models).

### DIFF
--- a/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
+++ b/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
@@ -16,6 +16,8 @@ with
             rel_id,
             rel_name,
             rel_type,
+            household_1_id,
+            is_parent2,
 
             coalesce(is_primary, false) as is_primary,
             coalesce(is_financial, false) as is_financial,
@@ -34,6 +36,8 @@ with
             rel_id,
             rel_name,
             rel_type,
+            household_1_id,
+            is_parent2,
             is_primary,
 
             row_number() over (
@@ -62,7 +66,9 @@ with
         -- excluded. rn > 1 skips the ROW picked as contact_1 (the financial
         -- fallback when no primary is set); the rel_id inequality against the
         -- pick additionally skips any other relationship row to the same
-        -- PERSON, so contact_2 can never duplicate contact_1.
+        -- PERSON, so contact_2 can never duplicate contact_1. The student-side
+        -- gate fields (is_parent2, household_1_id) ride on the relationships
+        -- staging grain, so only the related contact's record is joined here.
         select r.finalsite_enrollment_id, r.rel_id, r.rel_name, r.rel_type, r.rn,
         from contact_1_ranked as r
         inner join
@@ -70,17 +76,10 @@ with
             on r.finalsite_enrollment_id = p.finalsite_enrollment_id
             and r.rel_id != p.rel_id
         inner join
-            {{ ref("int_finalsite__contact_custom_attributes") }} as ca
-            on r.finalsite_enrollment_id = ca.finalsite_enrollment_id
-            and ca.is_parent2
-        inner join
-            {{ ref("stg_finalsite__contacts") }} as st
-            on r.finalsite_enrollment_id = st.finalsite_enrollment_id
-        inner join
             {{ ref("stg_finalsite__contacts") }} as cp
             on r.rel_id = cp.finalsite_enrollment_id
-            and st.household_1_id in unnest(cp.household_ids)
-        where r.rn > 1 and not r.is_primary
+            and r.household_1_id in unnest(cp.household_ids)
+        where r.rn > 1 and not r.is_primary and r.is_parent2
     ),
 
     contact_2_ranked as (
--- a/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
+++ b/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
@@ -1,6 +1,6 @@
 with
     contact_1_candidates as (
-        -- contact_1 is the student's ONE reportable parent contact. Finalsite
+        -- contact_1 is the student's FIRST reportable parent contact. Finalsite
         -- has no explicit contact rank, so we take the relationship it flags
         -- `primary` (its parent1 designation) and fall back to the one flagged
         -- `financial` when no primary is set — the two flags Ops maintains to
@@ -34,6 +34,8 @@ with
             rel_id,
             rel_name,
             rel_type,
+            is_primary,
+            is_financial,
 
             row_number() over (
                 partition by finalsite_enrollment_id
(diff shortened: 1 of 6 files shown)


cbini:
Review response — every data-dependent claim was verified against the warehouse. Fixes pushed in f6e05c01.

**Addressed:**

- **1 (same-person contact_2)** — the premise doesn't exist in data (0 duplicate `(student, rel_id)` relationship pairs in Newark, Camden, AND Paterson; 0 same-person slot pairs in the dev build), but the guard is cheap insurance: `contact_2_candidates` now excludes the contact_1 pick's `rel_id` explicitly, and the unit test covers the no-primary case where the exclusion is load-bearing.
- **6 (duplication)** — refactored to the suggested shape: `contact_1_picked` + `contact_2_ranked` feed a shared `parent_picks` → `parents_typed` → `parents` path; the third UNION branch is gone (~100 lines). The `phone_by_type` macro idea is good but touches the four emergency branches too — deferring as out-of-scope churn for this PR.
- **7 (double join)** — the shared typed CTE removes the second parent-branch join; the gate join in `contact_2_candidates` remains (it's what enforces household membership).
- **8 (redundant predicate)** — dropped `r.is_financial` and its projection; comment explains why `not is_primary` suffices.
- Unit test extended with the no-primary fallback case (contact_1 = first financial, contact_2 = next, same-person exclusion exercised).

**Verified non-issues (no change):**

- **3 (`is_parent2` on the contact)** — checked: `is_parent2` exists ONLY on student records in this tenant (every flagged contact has a grade/enrollment_type and no child relationships; 0 of the 3,216 picked contact_2 persons carry the flag themselves). The suggested rel_id-side gate is empty by construction. On rel_type tiering: `financial` is Finalsite's Parent 2 encoding here; rel_type is not a reliable discriminator (plenty of legitimate Parent 2s are `stepparent`/`guardian`), so I'm keeping relationship_id as the tiebreak.
- **4 (contact_2 without contact_1)** — 0 occurrences in the dev build, and post-refactor contact_2 derives from the same ranked set (an rn=1 row always exists). The residual edge (contact_1's rel_id lacking a contact record while contact_2's has one) is real but empty; skipping the data test for now.
- **9 (`h.id is not null` no-op)** — the filter is load-bearing, not cosmetic: BigQuery rejects NULL array elements when writing a table, and this staging model is table-materialized. A single id-less household would otherwise fail the build.
- **5 (household order)** — acknowledged; `households[0]` is the same positional dependency the address block has carried since the Focus feeds, and Finalsite's field groupings model "primary household" as a first-class concept. If the API reorders, addresses already move today — keeping the two consistent rather than inventing a different "household 1" definition for this column.
- **2 (prod behavior before the ContactType PR)** — intended, with one correction to my PR body: this is scoped, not purely additive. Ops requested Parent 2 across ALL NJ Finalsite contact feeds (not just DeansList), so the pivot's `contact_2_*` columns populating and the `guardian_email` fallback firing are the request, not side effects. Ordering in practice: the ContactType view change (#4579) deploys at merge time, while contact_2 rows only appear after the next district prod rebuild — so ContactType lands first when both merge together. The stale pivot comment/description is fixed in #4579 (in the properties yml, per the state:modified fan-out convention).

PR #4582, opened 2026-07-28
refactor(dbt): move FRESH shared models into packages, collapse scaffold

cbini:
# Pull Request

## Summary & Motivation

...restructure #4488's FRESH-dashboard work so shared Finalsite/Focus
transformation logic lives in the source-system packages, replace the
enrollment scaffold's compile-time mode switch with plain SQL, and split the
overloaded `grade_level = -1` sentinel.

**Stacked on `grangel/feat/claude-fresh-scaffold-swap` — merge #4488 first.**

### Structural changes

- **Finalsite package promotion.** Source cleaning (grade decode,
  `enrollment_type` initcap/default, `first_name` initcap,
  `active_school_year_display`) moved from the kipptaf union wrapper into
  `finalsite`'s `stg_finalsite__status_report`. The kipptaf model is now a thin
  `union_relations` + `region` + `_dbt_source_project` + `exclude_ids` filter.
- **Focus package extension.** The Focus enrollment derivation moved into the
  `focus` package's previously orphaned (zero-consumer)
  `int_focus__student_enrollment`. kipptaf's `int_focus__student_enrollments`
  is now a thin wrapper adding the Finalsite-ID crosswalk, the locations
  crosswalk, and region/district constants. Three kipptaf `stg_focus__*`
  passthroughs are deleted — they existed only to feed the misplaced model.
- **Scaffold collapse.** The enrollment scaffold was three alternative SQL
  programs selected at compile time by a `finalsite_scaffold_source` var,
  behind an `exceptions.raise_compiler_error` guard — the only such construct
  in the repo, with only one mode ever deployed. It is now the blend pipeline
  as plain SQL and the var is gone. `finalsite_recruitment_year` stays (read at
  several sites, mirrors the `current_academic_year` pattern).
- **Miami joins the blend.** Miami's grade membership now derives from Focus,
  the same way the NJ regions derive from PowerSchool, retiring the carve-out
  that forced the Google Sheet to carry Miami's entire school x grade spine.
  Focus reproduces that spine exactly, school for school and grade for grade.
- **Scaffolds leave the Finalsite source vertical.** Both scaffolds blended
  PowerSchool + Focus + Sheets while filed under `models/finalsite/` and
  referencing zero Finalsite models. They are now
  `int_tableau__fresh_enrollment_scaffold` and
  `int_tableau__fresh_goals_scaffold` under `extracts/tableau/intermediate/`,
  alongside `int_tableau__finalsite_student_scaffold`. This moves them from the
  `finalsite` dataset to `tableau` and flips them onto that directory's
  enforced contract. Folding the blend into `int_extracts__student_enrollments`
  was considered and rejected — 104 consumers, most PowerSchool-specific.
- **The Finalsite crosswalk left the Focus vertical.** `int_focus__student_enrollments`
  no longer joins Finalsite identity; that blend happens in its consumer. The
  join key is provably safe to move — `int_finalsite__contact_id_attributes`
  has 3,914 non-null Focus ids, all distinct, so it cannot fan out. The
  locations-crosswalk join stays, as id conformance needed by both consumers
  (precedent: `int_extracts__student_enrollments` joins
  `int_people__location_crosswalk` for the same reason).

### Review fixes folded in

- **Region added to the blend anti-join key.** It keyed on
  `(schoolid, grade_level)` only, but PowerSchool assigns `schoolid`
  independently per district, so a sheet row could be silently dropped by
  colliding with another region's PowerSchool row. Latent, not live — no
  collision exists today.
- **Grade sentinel split.** `-1` meant both Pre-K and whole-school-total, and
  `rpt_tableau__fresh_dashboard_aggregated` joins on `grade_level` — so a
  future PK applicant would have silently joined whole-school goal rows. Now
  **PK = `-1`, whole-school total = `-9`**. Ops has recoded both sheets to
  store `-9` natively, so the pipeline reads it straight through with no
  recode. Error-severity `accepted_values` guards on both Google-Sheets staging
  models fail the build loudly if a sheet ever reverts to `-1`, rather than
  silently dropping whole-school rows.
- Reference doc and fresh-dashboard skill rewritten to match shipped code
  (both described a hardcoded-literal design that the var replaced).
- Plus a region-label drift guard, a cross-source dedup-precedence comment,
  corrected `ps_*` column descriptions, and a dropped redundant severity
  override.

### Coordination required before/at merge

- **Production is currently wrong on this dashboard.** The sheets were recoded
  to `-9` ahead of the code, so prod's deployed views — still filtering `-1` —
  have lost 23,352 school-level rows (47,073 down to 23,721). This PR restores
  them, but it merges behind #4488, so the gap persists until both land.
- **Tableau (SRE):** the workbook references `grade_level = -1` for
  school-total rows and must move to `-9`. This fails **silently** — filters
  simply return nothing rather than erroring.
- **One label change:** school 30200805's grade rows now read `Miami Tech`
  (from the locations crosswalk, consistent with every other region taking its
  school name from the SIS) while its whole-school row still reads `MTH` from
  the sheet. Renaming it in the sheet resolves the split.
- **Merge order:** #4488 first, then this PR.
- **Post-merge cleanup:** `int_google_sheets__finalsite__scaffold` in
  `kipptaf_google_sheets` is orphaned by the rename chain and can be dropped.
- **Staging:** the `zz_stg` copies for the four finalsite districts and
  kippmiami focus were rebuilt so dbt Cloud CI can resolve the new package
  columns and the new `kippmiami_focus.int_focus__student_enrollment` source.

### Operational note

The new `not_null` on the package `grade_level` runs at `severity: error` in
**all four districts**, upstream of the `exclude_ids` filter. An unparseable
`application_grade` in the vendor feed will therefore fail finalsite staging
network-wide rather than in one place. That is deliberate — the alternative is
a silent null — but on-call should know what that failure means. It passes in
all four districts today.

### Validation

- Package staging built and tested in all four districts (Newark 17.2k, Camden
  5.2k, Miami 4.6k, Paterson 2.1k rows); the new `grade_level` tests pass in
  each.
- Changed kipptaf descendant graph: 26 nodes, PASS=23, WARN=1, ERROR=0 — the
  single warning is pre-existing and untouched by this branch.
- Row-count parity vs prod: `rpt_tableau__fresh_dashboard_aggregated`
  25,775 = 25,775; `rpt_tableau__fresh_dashboard_progress_to_goals`
  47,073 = 47,073. Compiled SQL was checked for stale dev refs that could fake
  parity.
- Sentinel migration parity against the renamed predecessor
  `int_google_sheets__finalsite__scaffold`: 1719 rows before and after, its
  162 whole-school rows now at `-9`, zero left at `-1`.

Refs #4451

## AI Assistance

AI-assisted end to end via Claude Code, human-directed: the user identified all
three structural problems (package misplacement, the novel mode-switch pattern,
the unjustified project var) and set the sentinel scheme after reviewing the
actual grade-level domains in each source.

Implementation ran as one task per commit, each written by a fresh subagent and
gated by an independent reviewer, plus a whole-branch review at the end. Three
errors in the authored plan were caught during execution and overruled in
favour of the code: a step premised on a contract entry that never existed;
sample SQL that contradicted the plan's own column-ordering constraint; and a
parity check premised on a prod baseline that does not exist (resolved against
the renamed predecessor instead, which was the better check).

The final review caught one real bug before push: the skill's ad-hoc
candidate-row generator still joined `grade_level = -1` against the recoded
staging model, which would have handed an analyst a duplicate whole-school row
for every school to paste into the sheet.

### General

      why. **Note:** base is not `main`, so `claude-review` will not run here.

### dbt

      no exposure changes needed; existing FRESH exposures already cover the
      downstream `rpt_tableau__*` models, whose contracted shapes are unchanged
- [x] **Breaking change?** No contracted column was renamed or removed. The
      `grade_level` VALUE changes for whole-school rows (`-1` to `-9`), which
      is a Tableau-workbook coordination item, not a schema break — see
      Coordination above. Rollback is a revert of this PR plus reverting the
      workbook filter.
- [x] No new external sources added. Existing `zz_stg` copies were rebuilt for
      the four finalsite districts and kippmiami focus so CI resolves the new
      package columns and the new source table.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

## Troubleshooting


src/dbt/kipptaf/models/finalsite/intermediate/int_finalsite__enrollment_scaffold.sql
```diff
@@ -25,9 +13,9 @@ with
     -- Miami's SIS moved to Focus (#4441); stg_powerschool__schools' Miami
     -- rows are a frozen pre-migration snapshot, not a live source of truth.
     -- Deliberate, temporary carve-out -- confirmed with the team to keep
-    -- Miami 100% sheet-sourced regardless of finalsite_scaffold_source
-    -- until Focus is ready as a scaffold source. Remove this filter (and
-    -- update the sheet-side builder's Miami note below) once that happens.
+    -- Miami 100% sheet-sourced regardless of scaffold source until Focus is
+    -- ready as a scaffold source. Remove this filter (and update the
+    -- sheet-side builder's Miami note below) once that happens.
```
cbini: Focus is ready as a scaffold source. What is this supplementing?

src/dbt/kipptaf/models/focus/intermediate/int_focus__student_enrollments.sql
```diff
@@ -1,170 +1,57 @@
...
-                order by academic_year desc, exitdate desc
-            ) as rn_year,
-        from enrollment
+    enrollments as (
+        select *, {{ extract_region("with_source_project") }} as region,
+        from with_source_project
     ),
 
-    with_year_counts as (
+    finalsite_ids as (
```
cbini: this should be blended downstream, not within the focus vertical

src/dbt/kipptaf/models/google/sheets/staging/stg_google_sheets__finalsite__goals.sql
```diff
@@ -1 +1,8 @@
-select *, from {{ source("google_sheets", "src_google_sheets__finalsite__goals") }}
+select
+    * except (grade_level),
+
+    -- -1 in the sheet means "whole-school total"; recoded to -9 (see
+    -- stg_google_sheets__finalsite__school_scaffold).
+    if(grade_level = -1, -9, grade_level) as grade_level,
```
cbini: we will recode in sheet

src/dbt/kipptaf/models/google/sheets/staging/stg_google_sheets__finalsite__school_scaffold.sql
```diff
@@ -1,5 +1,9 @@
 select
-    *,
+    * except (grade_level),
+
+    -- -1 in the sheet means "whole-school total"; recoded to -9 so -1 can
+    -- mean PK everywhere downstream (PK = -1, K = 0, 1-12).
+    if(grade_level = -1, -9, grade_level) as grade_level,
```
cbini: we will recode in sheet

src/dbt/kipptaf/models/finalsite/intermediate/int_finalsite__enrollment_scaffold.sql
cbini: Stale premise — this carve-out is inherited from #4488, which explicitly deferred wiring Miami into the blend. What it supplements today: nothing. The filter excludes Miami's frozen PowerSchool rows, which forces the sheet to carry Miami's full spine (every school × grade) instead of just whole-school rows and net-new grades.

Now that this PR gives the `focus` package a live enrollment intermediate, the replacement is mechanical: derive Miami's grade membership from Focus the same way the other regions derive it from PowerSchool (`stg_focus__schools` + current enrollments), drop this `region != 'Miami'` filter, and relax the sheet's Miami full-spine requirement. Can fold that into this PR or keep it as the follow-up — your call.

src/dbt/kipptaf/models/focus/intermediate/int_focus__student_enrollments.sql
cbini: Verified feasible: the sole consumer of this model is `int_tableau__finalsite_student_scaffold`, which already filters `finalsite_enrollment_id is not null` — the blend model is the only thing that needs the crosswalk. Will move the `int_finalsite__contact_id_attributes` join (and the `finalsite_enrollment_id` column) there, leaving this wrapper pure Focus + location enrichment. The join's fan-out risk moves with it, so the guard becomes the downstream model's uniqueness test rather than this one's.

src/dbt/kipptaf/models/google/sheets/staging/stg_google_sheets__finalsite__goals.sql
cbini: Understood — Ops updates the sheet to store `-9` natively and this recode comes out (model reverts to a passthrough). One sequencing constraint: the removal has to land in the same window as the sheet edit. If the recode is removed before the sheet changes, whole-school rows revert to `-1` and silently vanish from every `-9` filter downstream. Removing the recode also deletes the paste-value-vs-join asymmetry in the fresh-dashboard skill's row generator, which simplifies the docs — sheet and warehouse will finally agree on one value.

src/dbt/kipptaf/models/google/sheets/staging/stg_google_sheets__finalsite__school_scaffold.sql
cbini: Same as the goals model — comes out once the sheet stores `-9` natively, in the same window as the sheet edit (see the sequencing note on `stg_google_sheets__finalsite__goals`). The `school_level` case below is unaffected either way: negative grades produce a null `school_level` under both encodings.

cbini:
Pushed five follow-up commits addressing the review, plus the model-placement question that came out of it.

**Your four inline comments**

- **Miami carve-out** (`9663c3d10`) — Miami's grade membership now derives from Focus, the same way the NJ regions derive from PowerSchool. Verified: Focus reproduces the sheet's Miami spine exactly (Courage 6-8, Royalty K-5, Miami Tech 9, Legacy ES K-5, Legacy MS 6), so the sheet no longer carries Miami's per-grade rows — only whole-school rows, like every other region.
- **Blend downstream, not in the focus vertical** (`00edd5df1`) — the Finalsite crosswalk join moved out of `int_focus__student_enrollments` into its consumer. The join key is provably safe to move: `int_finalsite__contact_id_attributes` has 3,914 non-null Focus ids, all distinct, so it cannot fan out.
- **Recode in sheet** (`b43915d3d`) — both staging recodes removed; the sheets are the source of `-9` now. Added error-severity `accepted_values` guards on both staging models so a sheet reverting to `-1` fails the build loudly instead of silently dropping whole-school rows.

**Placement follow-up** (`367613265`)

Both scaffolds were blending PowerSchool + Focus + Sheets while filed under `models/finalsite/`, the Finalsite source vertical — and referencing zero Finalsite models. They moved to `extracts/tableau/intermediate/` as `int_tableau__fresh_enrollment_scaffold` and `int_tableau__fresh_goals_scaffold`, alongside `int_tableau__finalsite_student_scaffold`. Note this flips them from no contract to the directory-wide enforced contract, and moves them from the `finalsite` dataset to `tableau`.

I looked at folding the blend into `int_extracts__student_enrollments` instead and would advise against it: 104 consumers, most PowerSchool-specific (Illuminate, DeansList, autocomm, NJ state reporting), none of which expect Miami rows.

**Validation** — every commit verified against the same targets: `rpt_tableau__fresh_dashboard_progress_to_goals` 47,073 and `rpt_tableau__fresh_dashboard_aggregated` 25,775, matching prod; enrollment scaffold 172 rows with 24 whole-school; full-graph build PASS=26 WARN=1 ERROR=0 (the one warning is the pre-existing `rpt_gsheets__finalsite__log` duplicate check, untouched here).

**Two operational notes**

1. **Production is currently wrong on this dashboard.** The sheets were recoded to `-9` before the code shipped, so prod's deployed views — still filtering `-1` — have lost 23,352 school-level rows (47,073 to 23,721). This PR restores them, but it merges behind #4488, so the gap persists until both land. Worth telling whoever watches the FRESH dashboard.
1. **One label change for Tableau.** School 30200805's grade rows now read `Miami Tech` (from the locations crosswalk, consistent with how every other region takes its school name from the SIS) while its whole-school row still reads `MTH` from the sheet. Renaming it in the sheet would resolve the split.

Also worth a post-merge cleanup: `int_google_sheets__finalsite__scaffold` in `kipptaf_google_sheets` is left orphaned by the rename chain and can be dropped once this lands.

Issue #4442, opened 2026-07-17
refactor(powerschool): tear down legacy ODBC SIS stack after all districts migrate


cbini:
Sub-issue of #4427 — final teardown, **after** #4429 (Newark), Camden, and the Miami cleanup land and no district builds the ODBC path.

- [ ] Delete `src/teamster/libraries/powerschool/sis/odbc/` (assets, resources, schedules, schema, sensors, utils) + remaining ODBC wiring in each `code_locations/*/powerschool/`.
- [ ] Retire the sshpass `open_ssh_tunnel()` method + the `PowerSchoolODBCResource` / `DB_POWERSCHOOL` wiring (coordinate with #4424 part 2).
- [ ] Remove `sshpass` from the Dockerfile apt install; remove the mounted SSH-password file + its Kubernetes manifest (read only by the sshpass path). Keep `oracledb` + `dagster-ssh` (the dlt path uses both).
- [ ] Remove the dbt `powerschool` package `staging/odbc` variant once no district builds it.

Note: PowerSchool enrollment retirement is tracked separately in #4428.

Depends on: #4429 (Newark), Camden, Miami cleanup.

PR #4586, opened 2026-07-28
fix(dbt): carry the Sumner clause into the dlt schools school_level CASE

### LINKED ISSUE #4583: fix(dbt): dlt stg_powerschool__schools leaves school_level null for Sumner
## Problem

`stg_powerschool__schools.school_level` is `NULL` for KIPP Sumner Elementary
(Camden, `school_number` 179905) in the live `dlt` staging variant.

Sumner is K-6, so `high_grade = 6`. The `dlt` variant's `CASE` only matches
`high_grade` values of 12, 8, and 4/5, with no `else` — so Sumner falls through
to `NULL`.

The archived `odbc` and `sftp` variants both already carry a Sumner clause. It
was never ported when `dlt` became the live path (#3807 / #4442), so
`school_level` silently went from `'ES'` to `NULL` for Sumner and nothing caught
it.

## Evidence

Every school with a null `school_level` in prod
(`kipptaf_powerschool.stg_powerschool__schools`):

| school | high_grade | state_excludefromreporting |
| ------ | ---------- | -------------------------- |
| Sumner (Camden, 179905) | 6 | 0 (reporting) |
| SS (summer school) | 11 | 1 (excluded) |
| Graduated Students (999999) | 13 / 0 | 1 (excluded) |

Sumner is the only **reporting** school affected.

## Fix

Port the `odbc` / `sftp` clause into
`src/dbt/powerschool/models/sis/staging/dlt/stg_powerschool__schools.sql`:

```sql
case
    when high_grade = 12
    then 'HS'
    when high_grade = 8
    then 'MS'
    when high_grade in (4, 5) or abbreviation = 'Sumner'
    then 'ES'
end as school_level,
```

This is a value-only change (no column added or renamed), so per
`src/dbt/CLAUDE.md` it needs no `zz_stg` staging refresh — corrected values land
after the next prod rebuild.

## Why it matters now

Surfaced while building the FRESH dashboard enrollment scaffold (#4451), which
wants the official per-school `school_level` as the base for its
`school_level_alt` override. Downstream consumers of this column that assume
non-null for a reporting school are also affected
(`base_powerschool__student_enrollments`, `base_powerschool__sections`,
`int_powerschool__calendar_week`).

GabyRangelB:
# Pull Request

## Summary & Motivation

...restore `stg_powerschool__schools.school_level` for KIPP Sumner Elementary,
which is currently `NULL` in the live `dlt` staging variant.

Sumner is K-6, so `high_grade = 6` matches none of the CASE's `12` / `8` /
`4, 5` branches, and the CASE has no `else` — so it falls through to `NULL`. The
archived `odbc` and `sftp` variants both already carry the Sumner clause; it was
never ported when `dlt` became the live path (#3807 / #4442), so Sumner's
`school_level` silently went from `'ES'` to `NULL` and nothing caught it.

```sql
when high_grade in (4, 5) or abbreviation = 'Sumner'
then 'ES'
```

**Sumner is the only reporting school affected.** Every other null
`school_level` in prod is `state_excludefromreporting = 1`:

| school                      | high_grade | state_excludefromreporting |
| --------------------------- | ---------- | -------------------------- |
| Sumner (Camden, 179905)     | 6          | 0 (reporting)              |
| SS (summer school)          | 11         | 1 (excluded)               |
| Graduated Students (999999) | 13 / 0     | 1 (excluded)               |

### Verification

Built via a `kippcamden` dev target against the real dlt source:

| school     | high_grade | before | after           |
| ---------- | ---------- | ------ | --------------- |
| **Sumner** | 6          | `NULL` | **`ES`**        |
| LSP        | 4          | ES     | ES (unchanged)  |
| LSM        | 8          | MS     | MS (unchanged)  |
| Hatch      | 8          | MS     | MS (unchanged)  |
| KHS        | 12         | HS     | HS (unchanged)  |
| SS         | 11         | NULL   | NULL (unchanged)|

`unique_stg_powerschool__schools_dcid` passes. `trunk check --force` clean.

### Rollback

Revert the single line. This is a **value-only** change (no column added,
renamed, or removed), so per `src/dbt/CLAUDE.md` it needs no `zz_stg` staging
refresh — corrected values land on the next prod rebuild.

Surfaced while building the FRESH dashboard enrollment scaffold (#4451), which
wants the official per-school `school_level` as the base for its
`school_level_alt` override. Other consumers of this column also benefit:
`base_powerschool__student_enrollments`, `base_powerschool__sections`,
`int_powerschool__calendar_week`.

Closes #4583

## AI Assistance

AI-assisted via Claude Code, human-directed: the null was found while reviewing
an unrelated model, the user directed that the fix be ported from the archived
`odbc` variant rather than reworked, and chose the port-as-is approach over a
range-based rewrite. Claude traced the root cause, confirmed the blast radius
against prod, and verified the dev build.

## Self-review

### General

### dbt

- [x] Properties file — no new models; `school_level` is already declared in
      `src/dbt/powerschool/models/sis/staging/properties/stg_powerschool__schools.yml`
- [x] Exposure — no exposure change needed (value-only fix to an existing
      staging column)
- [x] **Breaking change?** No — value-only. No column renamed or removed, so no
      contracted downstream shape changes. The only behavior change is
      `NULL` becoming `'ES'` for one school.
- [x] No external source added or modified — the dlt source is a BQ-native
      table, so no `stage_external_sources` run is required

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes. Note: dbt Cloud CI builds only `kipptaf`, and this
      is a value-only change to a package model, so it selects no models and
      goes green trivially rather than as validation. The dev build above is the
      real check.

## Troubleshooting


GabyRangelB:
@claude — addressed your review in `908e11681`. Verification notes and two
corrections below.

## Implemented

**🟡 Finding 1 (no test) — agreed, this was the best finding.** It's the actual
reason the odbc-to-dlt port dropped the clause silently, so shipping the value
fix without the guard would have invited the same regression on the next
migration. Added both tests you suggested, at `severity: error` per the
staging-layer convention:

```yaml
data_tests:
  - not_null:
      config:
        severity: error
        where: state_excludefromreporting = 0
  - accepted_values:
      arguments:
        values: [ES, MS, HS]
      config:
        severity: error
```

You flagged this as **needs local verification** — done, in all three consuming
districts (you were right that the package builds in all three):

| district       | build | not_null | accepted_values |
| -------------- | ----- | -------- | --------------- |
| `kippcamden`   | pass  | pass     | pass            |
| `kippnewark`   | pass  | pass     | pass            |
| `kipppaterson` | pass  | pass     | pass            |

No reporting K-6/K-7 school in Newark or Paterson fails the `not_null` — Sumner
was the only reporting school with a null network-wide, confirmed against the
unioned `kipptaf_powerschool.stg_powerschool__schools`.

**🔵 Finding 3 (missing description) — added**, describing the derivation by
logic and noting why the `not_null` is scoped. Per the yml convention I kept the
issue ref out of the description.

I also moved `school_level` to the top of the `columns:` list, since it now
carries per-column tests — matching both the yml convention and this file's own
existing pattern (`dcid` sits first with its `data_tests:` before
`description:`).

## Corrections to your review

**Finding 3's detail is off.** You wrote that `school_level` is "the only column
in that file without one." There are two — `school_level` and
`fee_exemption_status` (50 columns declared, 48 descriptions). I fixed only
`school_level`, since that's the column this PR changes; `fee_exemption_status`
is a pre-existing gap I'm deliberately leaving out of a one-line fix PR.

**Finding 2's range rewrite has more blast radius than the finding implies**, so
I'm keeping the port-as-is, which you also recommended. I ran the check you
asked for:

```text
SS (summer school)          high_grade 11   state_excludefromreporting 1
Graduated Students (999999) high_grade 13   state_excludefromreporting 1
```

Under `when high_grade <= 6 then 'ES' when high_grade <= 8 then 'MS' else 'HS'`
both flip `NULL` to `'HS'`. That isn't cosmetic: `999999` is the
graduated-students sentinel, and `base_powerschool__student_enrollments`
(`sch.school_level`) feeds `int_extracts__student_enrollments`, so every
graduated student's `school_level` would change and propagate to ~20 downstream
models. Too much for a value restoration. Your point about the `else`-less
`case` being a recurring hazard stands, though — the new `accepted_values` plus
scoped `not_null` now converts that failure mode from silent to loud, which
addresses the risk without the behavior change.

## Finding 4 (informational)

Your lineage read matches mine: `school_level` reaches consumers through
**package** models, not the kipptaf union wrapper —
`stg_powerschool__schools` (pkg) to `base_powerschool__student_enrollments`
(pkg, L156) to `int_extracts__student_enrollments` (kipptaf, L429) to
`co.school_level` / `se.school_level` across the topline, assessment, gradebook,
attendance, and attrition models. Only `int_assessments__academic_goals` reads
`school_level` off the kipptaf wrapper. That's also why the fix belongs in the
package rather than at the kipptaf level — a kipptaf-only fix would leave every
consumer downstream of `int_extracts__student_enrollments` still null for
Sumner, and would put two different answers for the same school in one project.

I did not size the Camden ES-attribution flip with a dev build of
`int_extracts__student_enrollments`; agreed it's safe statically, since `rn = 1`
caps the join.

## Unrelated finding worth noting

Separately, while verifying the FRESH scaffold that surfaced this, a real
data-entry error turned up at Purpose (`732514`): one grade-4 student carried
`enroll_status = 0` at an MS-only school. Ops has already corrected it — the
record now reads `enroll_status = 2`. Not part of this PR, just noting it since
the two came out of the same investigation.

PR #4591, opened 2026-07-28
fix(dbt): collapse survey completion fan-out and restore TNTP exclusion

### LINKED ISSUE #4589: fix(dbt): survey completion view has phantom rows, question-grain fan-out, and no open-round flag
## Context

The Personalized Survey Links dashboard was reported as not showing AY2026 assignments. The warehouse side is correct: `rpt_tableau__survey_completion` carries 9,803 AY2026 rows covering all 1,424 active staff, and the reporting-terms sheet has the AY2026 `INFO` window (2026-07-01 to 2027-06-30). The dashboard-side year filter is a separate, Tableau-side problem.

While confirming that, three defects surfaced in this lineage.

## 1. Operator precedence drops the TNTP / Gallup exclusion

`rpt_tableau__survey_completion.sql` filters:

```sql
where
    sr.round_rn is null
    or sr.round_rn <= 1
    and sl.survey not in ('TNTP Insight Survey', 'Gallup Q12 Survey')
```

`AND` binds tighter than `OR`, so this parses as `round_rn is null OR (round_rn <= 1 AND survey not in (...))`. The exclusion added in 52328e7e8 therefore never applies to rows with no response.

Current state: 1,424 TNTP Insight Survey rows in the view, every one with a null `survey_response_id` — one phantom incomplete per active staff member.

Fix: parenthesize the `round_rn` predicates.

## 2. Question-grain fan-out inflates completion counts

The join to `rpt_tableau__survey_responses` is at question-answer grain, and `round_rn` is a `dense_rank` over submission time rather than a per-response dedupe. One assignment therefore matches every question row of the matched response.

Current state: 559,913 matched rows for 21,852 distinct assignments, roughly 25.6x. Every row carries `completion = 1`, so any `count()` or `sum(completion)` in the BI layer is multiplied by that survey's question count (66 for the staff info survey, 10 to 20 for the others). Only `count(distinct survey_response_id)` is currently correct.

Fix: reduce responses to response grain before joining.

## 3. is_current cannot identify the open round

`stg_google_sheets__reporting__terms` sets `is_current` true when the term window covers today, OR when the term is the latest of its type in a past academic year. Full-year survey terms satisfy the second branch in every year, so the staff info survey is currently `is_current = true` in four academic years at once (2023, 2024, 2025, 2026).

Consumers have no field meaning the round that is open right now, which leaves BI-side year selection to hardcoding.

Fix: add a narrower additive flag (window-covers-today only). Do not change `is_current` semantics — other consumers depend on the existing behavior.

## Also noted

The `personalized_survey_links` exposure lists only `rpt_tableau__survey_completion` and `rpt_tableau__survey_responses` in `depends_on`, but the dashboard also renders fields sourced from `rpt_tableau__survey_links` (the assignment label and the prefilled form link). The exposure should include it.

cristinabaldor:
# Pull Request

## Summary & Motivation

...fix two defects in `rpt_tableau__survey_completion` that were inflating
completion counts and surfacing a survey assignment that was supposed to be
excluded, and add a uniqueness test that locks in the corrected grain.

Found while investigating a report that the Personalized Survey Links dashboard
was not showing AY2026 assignments. That symptom turned out to be Tableau-side
(the warehouse view does carry AY2026 for all active staff), but two real
defects surfaced in this model along the way.

**1. Question-grain fan-out.** The join to `rpt_tableau__survey_responses` was
at question-answer grain, and `round_rn` is a `dense_rank` over submission time
rather than a per-response dedupe — so one assignment matched every question row
of its response, roughly 25x. Each of those rows carried `completion = 1`, so
any `count()` or `sum(completion)` in the BI layer was multiplied by that
survey's question count (66 for the staff info survey, 10 to 20 for the
others). Responses are now reduced to one row per
`(employee_number, academic_year, survey_code)` with `dbt_utils.deduplicate`
before the join.

**2. Operator precedence dropped the TNTP / Gallup exclusion.** `AND` binds
tighter than `OR`, so the filter added in 52328e7e8 actually parsed as:

```text
round_rn is null OR (round_rn <= 1 AND survey not in (...))
```

The exclusion therefore never applied to rows with no response, leaving one
phantom incomplete TNTP Insight row per active staff member (1,424 of them).
The `round_rn` disjunction is now redundant — the responses CTE is pre-filtered
to `round_rn = 1` and the LEFT JOIN yields null when nothing matches — so the
exclusion becomes an unconditional filter on the preserved side.

**3. Grain is now enforced.** `dbt_utils.unique_combination_of_columns` on
`(employee_number, survey, academic_year, survey_round)` at `severity: error`.
The whole value of this change is the grain, and a warn-level test would not
stop a regression back to fan-out from silently re-inflating every count in
Tableau. `src/dbt/CLAUDE.md` requires a uniqueness test on every `rpt_` model;
this one had none.

### Consumer-visible behavior change

No columns were added, renamed, or removed, so the contract is unchanged.
**The row grain changed**, which is the whole point of the fix but is visible
downstream: this view now returns one row per assignment instead of ~25.

**Two** Tableau exposures consume this model and both get the identical
collapse — `personalized_survey_links` (LSID `95584fbd-…`) and
`survey_dashboard` (LSID `b11542a2-…`). Any calc in either that was implicitly
compensating for the inflation (dividing by a question count, or leaning on
`COUNTD` to work around it) is worth re-checking. Rollback is a straight revert.

The third consumer, `rpt_extracts__survey_reminder`, is **unaffected** — see
Review follow-ups below.

### Verification

Built against prod upstreams (`--defer --favor-state`; compiled SQL confirmed
reading `kipptaf_tableau.*`, not stale dev copies). The uniqueness test passes
there too, so the grain claim is enforced rather than just asserted:

| Check                       | Before                                    | After                     |
| --------------------------- | ----------------------------------------- | ------------------------- |
| Rows                        | 559,913 matched across 21,852 assignments | 41,800                    |
| Distinct assignments        | 21,852                                    | 41,800                    |
| Rows per assignment         | ~25.6                                     | 1                         |
| TNTP / Gallup rows          | 1,424 (every one with no response)        | 0                         |
| Assignments marked complete | 21,852                                    | 21,852 (0 lost, 0 gained) |

Row total reconciles: 43,224 `rpt_tableau__survey_links` rows minus the 1,424
TNTP rows equals 41,800.

### Review follow-ups

`rpt_extracts__survey_reminder` filters `'TNTP Insight'` where the actual value
is `'TNTP Insight Survey'`, so that filter never matched — and it is now
redundant anyway, since the upstream excludes both surveys unconditionally.
Worth deleting, but **not in this PR** and with no behavior change either way:
all 1,424 phantom TNTP rows sit at `academic_year = 2024` with
`is_current = false`, and that extract filters on `is_current` and
`academic_year = 2026`, so they never reached the feed. Confirmed empirically —
recipient count is identical before and after this change (1,423 both ways,
0 lost, 0 gained).

Per-column `description:` backfill for this properties file (15 columns, none
documented) is a pre-existing gap left out of scope; a model-level
`description:` documenting the grain was added.

### Out of scope

Two findings documented in #4589 and deliberately not addressed here:
`is_current` is true for four academic years at once on full-year survey terms
(a known quirk in `stg_google_sheets__reporting__terms`, intentionally left
alone), and the `personalized_survey_links` exposure omits
`rpt_tableau__survey_links` from `depends_on`.

Separately, and unrelated to this code: the
`kipptaf__tableau__personalized_survey_links_extract_refresh_schedule` Dagster
schedule had been stopped since 2026-06-30, so the dashboard extract was four
weeks stale. That schedule has been restarted.

## AI Assistance

Claude Code investigated the reported symptom, found and quantified both
defects, wrote the fix, and ran the validation above. Human-directed: scoping
this to the completion view only, and the call to leave the `is_current`
behavior alone.

Refs #4589

### General

- [x] Review the **Claude Code Review** comment posted on this PR — all five
      findings addressed or answered with evidence; see the reply comment.

### dbt

- [x] No new models, so no new properties file is needed. The existing
      `rpt_tableau__survey_completion.yml` gains a model `description:` and a
      uniqueness test; the contract column set is unchanged.
- [ ] Exposure unchanged; the missing `rpt_tableau__survey_links` dependency on
      the `personalized_survey_links` exposure is tracked in #4589.
- [x] **Breaking change?** No contracted columns renamed or removed. Row grain
      does change — see "Consumer-visible behavior change" above.
      is required.

## CI checks

- [x] **Trunk** — `trunk check --force` run locally on both changed files: no
      issues.
- [x] **dbt Cloud** — passed on b30d427 with 0 warnings, model built in
      `dbt_cloud_pr_70403104388001_4591_tableau`. Re-running on ab6b731.


cristinabaldor:
Verified each finding against the code and prod data before acting. Four addressed, one corrected. Fixes in `ab6b731f3`.

**#1 uniqueness test — added.** Keyed exactly as suggested, at `severity: error` rather than the project default warn: a warn would not stop a regression back to fan-out from silently re-inflating counts, which is the failure mode this branch exists to fix.

On the caveat, the key was already proven before the finding landed. The dev-build validation counted `count(distinct format('%T|%T|%T|%T', ...))` over those four columns and got 41,800 against 41,800 rows. Worth flagging that the verification query in the finding targets the current **prod** view, which still carries the fan-out, so it returns duplicates and reads as a refutation of the very key it is proposing. The test now passes inside a real `dbt build` against prod upstreams.

**#2 reminder extract — the typo is real, the impact is not.** The string mismatch is confirmed: the value is `'TNTP Insight Survey'` and the filter says `'TNTP Insight'`, so that filter never matched. But the phantom rows never reached that feed. All 1,424 sit at `academic_year = 2024` with `is_current = false`, and the extract filters on `is_current` and `academic_year = 2026`. Ran the extract's own predicate against prod and against the fixed dev build:

```text
prod_recipients 1423 | dev_recipients 1423 | lost 0 | gained 0
```

Nobody drops off the reminder list, so this is not a live-feed behavior change and does not need naming as one. The now-dead filter is still worth deleting; left out of this PR as a separate file and recorded as a follow-up in the description.

**#3 two exposures — correct, description fixed.** `survey_dashboard` (`b11542a2`) does list this model. Both exposures are now named in the behavior-change section.

**#4 descriptions — model-level added,** documenting the grain, which is the non-obvious part. The 15-column backfill is a pre-existing gap and stays out of scope for a bugfix.

**#5 `= 1` — applied.** Confirmed `round_rn` is a `dense_rank` in one union branch and the literal 1 in the other, so never 0 and never null. Row counts identical after the change: 41,800 rows, 41,800 distinct assignments, 21,852 complete.

The blast-radius enumeration and the `partition_by` / `deduplicate` / DISTINCT-insufficiency analysis all check out. The Alchemer-branch and alias-map fan-out points were the useful catch there — both are covered by partitioning on `employee_number` rather than `respondent_email`.

cristinabaldor:
dbt Cloud is red on `ab6b731f3`, and it is **not** this PR. Recording the diagnosis so it is not re-derived.

The failure is `stg_finalsite__status_report`:

```text
CREATE VIEW has columns with duplicate name active_school_year_display
```

The kipptaf wrapper derives that column itself (`regexp_replace(active_school_year, ...) as active_school_year_display`) on top of `select * except (first_name, enrollment_type)`. Whether that collides depends entirely on the column set of the district relations it unions:

| Relation | `active_school_year_display` |
| --- | --- |
| `kippcamden_finalsite` (prod district) | absent |
| `zz_stg_kippcamden_finalsite` (shared staging, last built 2026-07-28 17:44 UTC) | present |

So it compiles clean against prod and fails only under `--target staging`, where the wider staging copy supplies the column that the wrapper also derives. The shared `zz_stg_kipp*_finalsite` copies appear to have been seeded from a branch that moves that derivation upstream to the district or package layer.

Why it is unrelated to this PR:

- This diff touches `rpt_tableau__survey_completion.sql` and its properties yml only. `stg_finalsite__status_report` is an upstream staging model, not a descendant, so `state:modified+` cannot reach it from this change.
- `rpt_tableau__survey_completion` built successfully inside the failing run (`dbt_cloud_pr_70403104388001_4591_tableau` at 19:59:19, within the build step that ended 19:59:32).
- The same commit range passed cleanly 11 minutes earlier on `b30d42756` with 0 warnings. The staging copy predates both runs (17:44), so nothing about it changed between them — the finalsite wrapper simply entered the `state:modified+` selection on the second run via deferring-state drift.

Not fixing it here — it is a different domain and a real latent bug that deserves its own change. It needs the shared district staging copies refreshed from prod (`dbt clone --select stg_finalsite__status_report --target staging --full-refresh` per district, a shared-resource mutation), or the upstream PR to land its kipptaf-side change. Anything whose CI selects that wrapper will hit the same wall until then.

GabyRangelB (approved):

PR #4593, opened 2026-07-28
feat(dbt): port FRESH scaffold revamp onto current main

GabyRangelB:
# Pull Request

## Summary & Motivation

...make `int_tableau__fresh_enrollment_scaffold` fully SIS-derived and retire the
hand-maintained `stg_google_sheets__finalsite__school_scaffold` sheet.

The scaffold now computes every row type that sheet supplied. Ops stops
hand-entering school x grade rows entirely.

### What the scaffold does now

- **`school_directory`** unions PowerSchool (non-Miami, `state_excludefromreporting = 0`)
  with Focus (`int_focus__schools` where `max_syear is null`), joined through
  `stg_google_sheets__people__locations` on `focus_school_id` to reach the
  PowerSchool-space `schoolid` and the school abbreviation. Focus's own
  `school_number` is a Focus code (`2332A`), not a PowerSchool id, so that join is
  what puts Miami in the same id space as every other region.
- **`current_grade_levels`** takes grade membership from live enrollment in both
  SIS. The PowerSchool branch excludes Miami — those rows are a frozen
  pre-migration snapshot that still carries a grade 5 at Courage, which it no
  longer serves.
- **The three row types no SIS produces** are derived: `grade_level = -9`
  whole-school totals, `schoolid = 0` region rollups, and net-new schools/grades
  anti-joined off `int_finalsite__status_report_unpivot` on
  `(region, schoolid, grade_level)`.

### Other changes

- **Mode switch removed.** The `finalsite_scaffold_source` var and its
  `exceptions.raise_compiler_error` guard are gone; the model is plain SQL. The
  net-new branch is gated by a predicate comparing the two year vars, which
  renders as a constant (`2026 != 2026` today) and folds to false until SRE's
  recruitment year runs ahead of PowerSchool's.
- **`academic_year` renamed to `enrollment_academic_year`** through the scaffold,
  the goals scaffold, and the singular region test. Both
  `rpt_tableau__fresh_dashboard_*` views alias it back, so **the Tableau-facing
  column name is unchanged** and their contracts are untouched.
- **Docs** moved to `docs/models/` under the Models nav tab, with
  `mkdocs-redirects` added so the old URL — and the gradebook doc's, which has
  been 404ing since it moved in `427be83af` — forward instead of breaking.
  Documents the rollover process with ownership, a `status_crosswalk` column
  reference, and a goals reconciliation procedure against SRE's target workbook.

### Verification

- **Reproduces the retired sheet exactly**: 172 rows both sides, identical
  `(region, schoolid, grade_level)` key sets, zero orphans either direction.
- **One intentional value change**: schoolid `30200805`'s whole-school row now
  reads `Miami Tech` (from `int_people__location_crosswalk`) instead of the
  sheet's `MTH`. `MTH` existed only in the sheets. This also closes a split
  flagged in #4582 — the grade rows already read `Miami Tech` while the `-9` row
  read `MTH`.
- `int_tableau__fresh_goals_scaffold` unchanged at 1,719 rows; zero goals-sheet
  rows orphaned by the new scaffold.
- Dev build with contracts enforced: PASS=10, ERROR=0 across both scaffolds, both
  rpt views, five uniqueness tests and the singular region test. Re-run after
  merging `main` brought Focus column adds.
- Behaviour-preserving on the Jinja removal, proven by multiset checksum rather
  than row counts: identical 172 rows and identical
  `md5(string_agg(to_json_string(row)))` before and after.

### Rollback

Value-and-shape change to two intermediates and two `rpt_` views. Reverting the
commit restores the prior scaffold; the sheet model and source only need
`enabled: false` removed. No contracted external column was renamed or dropped,
so no downstream exposure changes shape.

### Coordination

- **SRE:** two visible changes. Sumner grades 5-6 continue to report `MS`
  (unchanged), but schoolid `30200805`'s whole-school row label changes from `MTH`
  to `Miami Tech`. No join in the chain keys on the school name.
- **Accepted divergence, documented not fixed:** `school_level` is banded per
  grade using NJ boundaries, so Miami grade 5 reports `MS` here while
  `stg_google_sheets__finalsite__goals` reports `ES` using Miami's real 5/6
  boundary. The goals sheet stays manually entered because some goals are
  standard by grade level across the network rather than by school level. The
  consequence is that `progress_to_goals` (reads the scaffold) and `aggregated`
  (reads the goals sheet) legitimately differ on those rows.
- **Post-merge DDL**, requested in a comment on this PR: three orphaned
  relations in `kipptaf_google_sheets` need dropping. dbt never drops what a
  disabled or renamed model already created.

Refs #4451

## AI Assistance

AI-assisted via Claude Code, human-directed. The user set every design decision:
dropping `school_level_alt`, keeping the per-grade `school_level` band to match
the retired sheet rather than switching to each school's official level, deleting
the mode switch in favour of plain SQL, disabling the sheet rather than deleting
it, and accepting the `Miami Tech` label change. Claude wrote the SQL, ran the
parity and checksum verification, and drafted the docs and skill updates.

Two errors were caught during review and corrected: a `low_grade`/`high_grade`
claim carried forward from an older doc revision (the unreliable field is
`low_grade`, stale-low on three rebanded schools; `high_grade` tracks reality),
and a dropped `region` term in the net-new anti-join key that #4582 had
deliberately added to prevent cross-district `schoolid` collisions.

## Self-review

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Addressed:
      stale Open Questions section rewritten; the `properdocs` supply-chain flag
      verified as a false positive (it is a maintained community fork of MkDocs,
      pulled in by `mkdocs-redirects`' own maintainer) and noted below.

### dbt

- [x] Properties files updated for every changed model, including the column
      rename and the contract column list
- [x] Exposure — no change needed; `fresh_dashboard` already covers both rpt
      views and no new external consumer was added
- [x] **Breaking change?** No contracted external column renamed or removed. The
      internal rename to `enrollment_academic_year` is aliased back to
      `academic_year` in both rpt views, so contracts and Tableau are unchanged.
      One value change (the `MTH` label) is called out under Coordination above.
- [x] **External source modified** — `src_google_sheets__finalsite__school_scaffold`
      is set `enabled: false`. This is a disable, not an addition or a schema
      change, so no `stage_external_sources --target staging` run is required: CI
      simply skips the now-disabled model. Its three workbook siblings
      (`goals`, `exclude_ids`, `status_crosswalk`) are untouched and still
      enabled.

### Docs

- [x] Reference doc moved to `docs/models/` with a nav entry under Models and
      redirects for the old URL
- [ ] No schedule or sensor changes, so the automations catalog is unchanged

## CI checks

- [ ] **Trunk** — passes. `trunk check --force` clean locally across every
      changed file.
- [ ] **dbt Cloud** — passes.

## Notes for the reviewer

- `rpt_tableau__fresh_dashboard_qc` is being actively reworked on this branch
  into an unpivoted one-row-per-student-per-flag view for SRE. It was
  `enabled: false` and referenced by nothing before this branch.
- `detailed_status_branched_ranking` (column E of `status_crosswalk`) has no
  consumer — declared and passed through, read by nothing. Left in place
  deliberately; noted in the doc's Open Questions.
- `int_tableau__fresh_enrollment_scaffold` has no dbt unit test. Worth adding,
  but it needs `format: sql` fixtures because Focus columns changed this cycle,
  so it is cleanly separable from this PR.

---
  - https://app.asana.com/0/0/1216966253150593


GabyRangelB:
@cbini — a small favour when you get a chance, since we don't have DDL access.

Retiring the scaffold sheet in this PR leaves three unreferenced relations in `kipptaf_google_sheets`. dbt stops building a disabled or renamed model but never drops what it already created, so these need a manual drop.

| relation | type | how it got orphaned |
| --- | --- | --- |
| `int_google_sheets__finalsite__scaffold` | view | `6ae7572de` renamed the model to `int_finalsite__goals_scaffold` in March; dbt built the new name and left this behind |
| `stg_google_sheets__finalsite__school_scaffold` | table, 172 rows | disabled in this PR |
| `src_google_sheets__finalsite__school_scaffold` | external | source disabled in this PR |

**The view is safe to drop now** — nothing has built or read it since March:

```sql
drop view if exists `teamster-332318.kipptaf_google_sheets.int_google_sheets__finalsite__scaffold`;
```

**The other two should wait until this merges and deploys.** The disable only exists on this branch, so prod is still building them; dropping early just means the next run recreates them.

```sql
drop table if exists `teamster-332318.kipptaf_google_sheets.stg_google_sheets__finalsite__school_scaffold`;
drop external table if exists `teamster-332318.kipptaf_google_sheets.src_google_sheets__finalsite__school_scaffold`;
```

Not urgent — a 0-row view, a small table, and an external that holds no storage. The reason it's worth doing at all is that a stale table which looks live is easy to trust by mistake: those 172 rows would be frozen at July 2026 while reading as current.

Thanks!

For reference, the first one is the same cleanup noted in #4582's post-merge section that hasn't happened yet.

GabyRangelB:
@claude - please review

GabyRangelB:
@cbini — a second one for you, this time a judgement call rather than a task.

This PR moves the FRESH reference doc from `docs/reference/` to `docs/models/` (per the convention in `docs/CLAUDE.md`, and matching where the gradebook doc already lives). To stop the old URL 404ing I added `mkdocs-redirects` to the docs dependency group. That also fixes the gradebook doc's old URL, which has been dead since `427be83af` moved it — there was no `plugins:` block in `mkdocs.yml` at all, so no redirects existed.

The wrinkle the Claude review caught: **`mkdocs-redirects==1.2.3` depends on a package called `properdocs`, and only 1.2.3 does.** Every earlier release depends on `mkdocs` alone.

I checked it out, because that shape looks like a typosquat:

- `properdocs` is a **community fork of MkDocs** — `github.com/properdocs/properdocs`, BSD-2-Clause, ~2,200 commits, created because upstream MkDocs development stalled. Its version line starts at 1.6.5, continuing MkDocs' own 1.6.x.
- It is PyPI-owned by `oprypin`, who also maintains `mkdocs-redirects`. So the new dependency is that maintainer moving to the fork, not an injected package.

So not a security issue. But it does mean a one-line redirect plugin now installs a **second, near-identical static site generator** (plus click, jinja2, markdown, pyyaml, watchdog) alongside `mkdocs-material`'s own `mkdocs`.

Three options, and I'd rather you pick than decide unilaterally since it touches shared dependency policy:

1. **Leave it.** Docs build verified working, redirects emit correctly. Accepts the duplicate SSG.
2. **Pin `mkdocs-redirects>=1.2,<1.2.3`.** Gets the redirect plugin with `mkdocs` as its only dependency. Smallest footprint, but pins us off the latest release.
3. **Drop the plugin.** No new dependency; the two old doc URLs keep 404ing.

Happy to make whichever change you prefer. Worth noting the broader question this raises, which is bigger than this PR: if the MkDocs ecosystem is migrating to `properdocs`, we may eventually want to decide that deliberately rather than arriving there via a transitive dependency.

Verification, if you want to check independently:

```text
https://pypi.org/pypi/mkdocs-redirects/json   -> requires_dist on 1.2.3
https://pypi.org/pypi/properdocs/json         -> owner, versions
https://github.com/properdocs/properdocs      -> "a fork of MkDocs, aiming to be a complete drop-in replacement"
```

GabyRangelB:
@claude please re-review. Two commits landed since your last pass, and I've actioned or answered each of your findings.

## Actioned

**Finding 1 — the `properdocs` supply-chain flag.** You were right to flag the pattern and right to say it needed verification. Verified: `properdocs` is a community **fork of MkDocs** (`github.com/properdocs/properdocs`, BSD-2-Clause, ~2,200 commits, created because upstream MkDocs stalled), PyPI-owned by `oprypin` — who also maintains `mkdocs-redirects`. So 1.2.3's new dependency is that maintainer moving to the fork, not an injected package. Not a security issue.

The real concern it surfaced is footprint: a one-line redirect plugin now installs a second static site generator alongside `mkdocs-material`'s own `mkdocs`. Raised as a separate comment for @cbini to decide (leave / pin `<1.2.3` / drop the plugin) since it touches shared dependency policy.

**Finding 2 — stale Open Questions.** Fixed. Four of the six entries were answered by this work, so they moved to a "Resolved" subsection rather than being silently deleted: Miami carve-out removed, the Liberty/Sunrise sheet gap moot (both closed, `max_syear = 2025`), Focus id translation built, and both "what Focus must supply" gaps closed — the second of those closed in earlier work, via `int_tableau__finalsite_student_scaffold` reading `int_focus__student_enrollments`. Kept the two genuinely open ones and added `detailed_status_branched_ranking` having no consumer.

**Findings 3 and 4 — both stale PR body, not code.** PR body rewritten to describe what shipped. The external-source checkbox is now checked with the reasoning you suggested (a disable is not an addition or schema change, so no `stage_external_sources` run). `school_level_alt` is gone from the body entirely — it was not unblocked by #4586, it was deliberately deleted, since nothing downstream consumed it.

## New since your review

`rpt_tableau__fresh_dashboard_qc` is now enabled and reshaped into an SRE worklist — one row per student per fired mismatch flag, via `unpivot` over the four `is_*_mismatch` flags. It was `enabled: false` and referenced by nothing before this branch. Contract enforced, uniqueness test on `(enrollment_academic_year, finalsite_id, flag_name)`.

Worth a look at the `select distinct` in its `roster` CTE. The source carries one row per `(goal_type, goal_name)` grouping even inside a single `grouped_status_timeframe` — confirmed on prod, five rows for one student — but every column this view projects is identical across them (24,406 rows collapse to 13,873 distinct, matching the 13,873 student count exactly). So it is grain projection per the convention, not dup-masking. Please sanity-check that reasoning.

Also added a doc section on how Finalsite's `latest_status` maps to an expected enrollment status, because the naming is a trap: `ps_enroll_status` is derived from **Finalsite** despite the `ps_` prefix, while plain `enroll_status` is the SIS value.

## Your minor items

- **No unit test on the scaffold** — agreed, real gap. Deferring rather than skipping: it needs `format: sql` fixtures because Focus columns changed this cycle, which makes it cleanly separable from this PR.
- **`finalsite_new` dormant at `2026 != 2026`** — correct and intentional, documented. Also worth noting the rollover sequence this enables is now written up: SRE enters new schools/grades in Finalsite under the new year, then the year bump brings them in through that branch.

## One correction to my own earlier claim

I had said `enroll_status` going NULL was PowerSchool rollover lag. That was wrong — PowerSchool has 9,994 rows for 2026, all carrying `infosnap_id`. A NULL there means no matching PowerSchool record for that student, not a year lag.

Two things I'd value a second opinion on: whether the `distinct` grain-projection argument above holds, and whether the `is_active_inactive_mismatch` predicate should also treat a NULL SIS `enroll_status` as a mismatch when Finalsite says `Enrolled` — today it only fires against SIS `2`/`3`, so a Finalsite-enrolled student with no SIS record at all is invisible to it.

GabyRangelB:
Third-pass findings actioned in `bdb3fa8e0`. All four addressed, including the two you filed as minor.

## Finding 1 — correct, and it was this PR's own regression

Both halves confirmed. The skill said `rpt_tableau__fresh_dashboard_qc` "is `enabled: false` and referenced by nothing, so it never builds" — accurate when written, then falsified by `347f801a2` in the same PR. And the verify command genuinely missed it; `dbt ls` returned zero matches because `int_tableau__finalsite_student_scaffold` had no `+`.

Rewrote the paragraph to say `qc` **is** a verification site and why (Tableau-exposed, contract-enforced, SRE-facing — a bump that quietly empties it matters). Added the `+`. Re-ran `dbt ls` and it now selects the model and its test.

## Finding 2 — annotation added

`finalsite_new`'s `SELECT DISTINCT` now carries the grain-projection annotation, naming the actual determination rather than restating the convention: `school`/`region` are attributes of `schoolid`, `school_level` is a function of `grade_level`, remainder are literals.

## Finding 3 — fixed at the root, and worse than reported

You were right that the deferral claim doesn't hold, and the underlying rule is stricter than the finding implies. `src/dbt/CLAUDE.md` says descriptions must be present on every column and to **"infer downstream from parents"** — write the description deriving it from the parent, not point at the parent instead. So the pattern was wrong even in principle, independent of the upstream gap.

Documented both layers rather than just satisfying the check: **0 undocumented columns across 34 upstream and 21 in the QC view.** Grounded in the SQL and in prod value domains rather than inferred — `aligned_enrollment_type` really is always `All`, `self_contained` carries a literal `false` alongside Yes/No, `grouped_status_timeframe` is Current/Ever. `is_enrolled_fdos` also now records the Focus caveat, which is where someone debugging it will actually look.

On the `kfwd` precedent: agreed it isn't a novel deviation, but it isn't cover either — the rule's only exception is thin cross-project wrapper `rpt_` models sourcing `kipptaf_extracts`, and neither those nor this model qualify.

## Finding 4 — labelled

Tagged at column level, following the only precedent in `extracts/tableau/` (`rpt_tableau__gradebook_es_comments`): names, `finalsite_id`, `powerschool_student_number` on both models, plus `gender` and `birthdate` upstream. You were right that it isn't an enforced convention — the root `CLAUDE.md` explicitly says to confirm scope with the user for a PII-heavy new model rather than auto-tagging, which is what happened. Worth noting the QC view is the first place this data reaches Tableau rather than staying in an intermediate.

## Two things worth recording for whoever reviews next

**Your withdrawal of the two ST06 findings was the right call**, and the reasoning generalises: any future "fix" to column ordering in this model's `UNION ALL` branches will silently corrupt data rather than fail, because BigQuery matches positionally. Worth being suspicious of that class of suggestion here specifically.

**A verification trap that cost me a false alarm.** Building `int_tableau__finalsite_student_scaffold+` alone fails with `Name enrollment_academic_year not found inside s` — because `--favor-state` resolves the two scaffolds (not descendants of the student scaffold) to **prod**, which still has the pre-rename `academic_year`. It looks like a code bug and isn't. Both roots have to be selected together, which is now what the skill's command does:

```bash
dbt build --select int_tableau__fresh_enrollment_scaffold+ int_tableau__finalsite_student_scaffold+ \
  --target dev --defer --favor-state --state target/prod
```

That run is green post-merge: **PASS=14, ERROR=0** — 5 views, the student-scaffold table, 8 tests, the exposure. `dbt parse` clean, `trunk check --force` clean, and I checked `dbt_project.yml` for the duplicate-top-level-key trap after merging the GPA-goals work in (none).

Nothing outstanding from your three passes. Still open and not review items: @cbini on the `mkdocs-redirects` footprint and the orphaned-relation drops, and a question out to Miami ops about the per-school first-day dates behind the `is_enrolled_fdos` caveat.

mkdocs.yml
cbini, line 1: @GabyRangelB my only question: could we not just reorganize the files rather than adding a new dependency? This one also triggered a potential security flag

GabyRangelB: honestly defer to what you think is best. this got flagged because after you merged the fresh dash PR, the reference doc wasnt showing under the models tab we had talked about. im open to whatever you think is best if we are able to continue posting the reference docs to help with future development.

cbini (approved):

Issue #4400, opened 2026-07-14
feat: DeansList family contacts extract (contacts.txt) for NJ regions from Finalsite


cbini:
Create a new nightly DeansList import file, `contacts.txt`, for the NJ regions (Newark, Camden, Paterson), sourced from Finalsite contacts.

Template: Family Contacts tab of the [DeansList nightly export template](https://docs.google.com/spreadsheets/d/1PBDcnNLVWZmjFIjFM2R-0gi8RVVkEZg1Jpdgv-N1_6c/edit?gid=1172599928#gid=1172599928).

Requirements:

- Single `parent` contact per student (Finalsite `contact_1`: primary/financial pick) — no emergency contacts
- Currently enrolled students only (`stg_powerschool__students.enroll_status = 0`)
- Columns: `StudentID`, `ParentFirstName`, `ParentLastName`, `HomePhone`, `WorkPhone`, `CellPhone`, `Email`, `Relationship`, `Language` (header present, always null)
- File must be named `contacts.txt`; CSV with header; delivered to the existing DeansList SFTP destination on the nightly extract schedule

Design highlights (spec to follow):

- Split `contact_first_name` / `contact_last_name` at the package level in `int_finalsite__student_contacts` (finalsite dbt package)
- New kipptaf `rpt_deanslist__family_contacts` reading the Finalsite intermediates directly (`int_finalsite__student_contacts` union + `int_finalsite__contact_id_attributes` crosswalk)
- New entry in `deanslist-annual.yaml` extract config — no Python changes
- Two-PR rollout: finalsite package column add first, then kipptaf consumer

PR #4594, opened 2026-07-28
feat(dbt): add Focus student contacts models for Miami (Phase 1)

### LINKED ISSUE #4585: feat: replace Miami PowerSchool student contacts with Focus
Miami counterpart of the NJ Finalsite contacts cutover (#4346). The Miami branch of `int_students__contacts` still reads the frozen `kippmiami_powerschool` dataset — stale since the Focus migration. Replace it with live Focus data.

Asana: https://app.asana.com/1/913513768672/project/1215209393570629/task/1216356414391207

**Phase 1 (build now):**

- New `stg_focus__people` staging model in the `focus` package
- New `int_focus__student_contacts` intermediate (grain: `student_id` x `person_id`, all links, no slot cap) from `students_join_people` + `people` + `people_join_contacts` + `address`
- kipptaf plumbing: add to `sources-kippmiami.yml` (focus) + thin union wrapper — inert until Phase 2

**Phase 2 (swap when Focus contacts data lands — today Focus holds ~1 contact link vs 3,874 students; the Finalsite import hasn't run):**

- Replace the PowerSchool branch of `int_students__contacts` with a Focus branch (slotting applied at this layer)
- Retire kipptaf `int_powerschool__contacts` and the frozen contacts source reads
- Verify downstream: `dim_student_contact_persons`, `bridge_student_contacts`, `rpt_gsheets__student_contact_info`

Focus remains system of record for Miami contacts; the Finalsite to Focus import-once feed keeps seeding new enrollees. Design spec to follow on the linked branch.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will add the Focus-side student contacts models
that Phase 2 needs in order to retire Miami's frozen PowerSchool contacts data.

Refs #4585 (Phase 1 of 2 — the issue stays open for Phase 2).

Miami's branch of `int_students__contacts` still reads the frozen
`kippmiami_powerschool` dataset, a snapshot from before the Focus migration, so
Miami contacts in `dim_student_contact_persons`, `bridge_student_contacts`, and
`rpt_gsheets__student_contact_info` are stale and degrading. This is the Miami
counterpart of the NJ Finalsite cutover (#4346, #4400).

**What this adds**

- `stg_focus__people` — contract-enforced staging over the Focus `people` table
  (the guardians/contacts linked to students).
- `int_focus__student_contacts` — one row per live student-to-contact-person
  link (grain `student_id` x `person_id`), with typed phones, residence-scoped
  household membership, and null-preserving custody/emergency/pickup flags.
  Deliberately **unslotted and uncapped**: the `contact_slot` vocabulary
  (`contact_1`, `emergency_N`) is applied at the kipptaf layer in Phase 2, where
  downstream compatibility defines it.
- `focus_unmapped_phone_contact_titles` — warn-level guard test. Focus contact
  titles are free-typed and Focus publishes no type vocabulary, so the pivot's
  regex mapping is inferred; this test surfaces titles that map to nothing, plus
  self-contradictory titles matching both the phone and email vocabularies.
- kipptaf plumbing — one source entry plus a thin `union_relations` wrapper.
  **Intentionally inert**: nothing consumes it until Phase 2.

**Deliberately not in this PR** (Phase 2, gated on the Finalsite-to-Focus
contacts import actually landing in Focus — today Focus holds 1 contact link
against 3,874 students): swapping the PowerSchool branch of
`int_students__contacts`, the slotting logic, and retiring the kipptaf
`int_powerschool__contacts` chain.

**Reviewer notes**

- Every model builds to a couple dozen rows today (`int_focus__student_contacts`
  = 20, against 3,874 students). That is expected, not a failure — the upstream
  Finalsite-to-Focus contacts import has not run at scale. Correctness rests on
  reading the SQL, which is why the guard test and the error-severity grain test
  exist.
- The declared grain is not enforced upstream (`stg_focus__students_join_people`
  is unique on `id` only). The grain test is deliberately error-severity so a
  violation stops the build loudly rather than double-counting into Phase 2
  slotting; the model description documents the remediation.
- Phone values are emitted **raw**, exactly as stored in Focus. An earlier
  revision called `finalsite.clean_phone` for E.164 normalization; that was
  reverted because the `focus` package declares no dependency on `finalsite`, so
  a consumer that installs `focus` without `finalsite` would fail to build.
  Normalization moves to the kipptaf consumer in Phase 2. The prohibition is
  recorded in `src/dbt/focus/CLAUDE.md`.
- `zz_stg_kippmiami_focus` has been seeded, so CI can compile the wrapper. After
  merging `main`, the wrapper's `union_relations` expansion was re-verified
  against the seeded copy — 20 columns, matching the model contract exactly.

## AI Assistance

Claude Code co-authored this PR. Human-directed: the two-phase sequencing, the
decision to keep the intermediate unslotted, and the call to fix all review
findings rather than defer the cosmetic ones. AI-assisted: source-data
profiling, the model and test SQL, the properties files, the design spec and
implementation plan under `docs/superpowers/`, and a multi-agent review pass
that caught four latent correctness bugs (a `phone_primary` that returned NULL
across a filter boundary, email addresses landing in phone columns via titles
like "Home Email", blank values winning the phone rank, and household
membership counting mailing and bus-stop addresses).

## Self-review

### General

No Dagster changes.

### dbt

- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — n/a, no
      external consumer in this PR; the kipptaf wrapper is inert until Phase 2
- [x] **Breaking change?** No. All models are new; no contracted column was
      renamed or removed, and no existing model changed behaviour.
      with **`--target staging`** — n/a, no external sources. The related step
      that *was* required is seeding `zz_stg_kippmiami_focus` (`dbt clone` plus
      a `--target staging` build), because editing `sources-kippmiami.yml` marks
      the whole source `state:modified` and the wrapper's `union_relations`
      resolves its column list at compile time. Done before opening this PR.

No schedule, sensor, or new-integration changes. `src/dbt/focus/CLAUDE.md` was
updated to record that models in this package must not reference another source
package's macros, and that phone values are therefore emitted raw.

## CI checks

## Troubleshooting


src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
```diff
@@ -0,0 +1,165 @@
...
+            person_id,
+            address_id,
+            sort_order,
+
+            student_relation as relationship,
+
+            custody = 'Y' as is_custodial,
+            emergency = 'Y' as is_emergency,
+            pick_up = 'Y' as is_pickup,
+            reunification = 'Y' as is_reunification,
```
cbini: refactor to staging

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
```diff
@@ -0,0 +1,165 @@
...
+    ),
+
+    people as (
+        select
+            person_id,
+            first_name as contact_first_name,
+            last_name as contact_last_name,
+            email,
+
+            nullif(array_to_string([first_name, last_name], ' '), '') as contact_name,
```
cbini: refactor to staging

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
```diff
@@ -0,0 +1,165 @@
...
+            case
+                when regexp_contains(lower(title), r'cell|mobile')
+                then 'mobile'
+                when regexp_contains(lower(title), r'home')
+                then 'home'
+                when regexp_contains(lower(title), r'work|business|office')
+                then 'work'
+                when regexp_contains(lower(title), r'day')
+                then 'daytime'
+            end as phone_type,
```
cbini: refactor to staging

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
```diff
@@ -0,0 +1,165 @@
...
+                order by detail_priority asc nulls last, value asc
+            ) as type_rank,
+
+            row_number() over (
+                partition by person_id
+                order by detail_priority asc nulls last, value asc
+            ) as overall_rank,
+        from phones_filtered
+        where phone_type is not null and value is not null
+    ),
```
cbini: these ctes can be combined

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
```diff
@@ -0,0 +1,165 @@
...
+        where overall_rank = 1
+    ),
+
+    addresses as (
+        select
+            address_id,
+
+            nullif(
+                array_to_string([address, address2, city, state, zipcode], ', '), ''
+            ) as home_address,
```
cbini: refactor to staging

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
```diff
@@ -0,0 +1,165 @@
...
+    ),
+
+    -- read off the unfiltered rank, not phones_typed's type_rank = 1 filter
+    -- the overall_rank = 1 row can belong to a phone_type the type_rank filter
+    -- has already discarded, so this must not read through that filter.
+    primary_phone as (
+        select person_id, value as phone_primary,
+        from phones_ranked
+        where overall_rank = 1
+    ),
```
cbini: don't need this cte, add filter to join

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
```diff
@@ -0,0 +1,165 @@
...
+
+    addresses as (
+        select
+            address_id,
+
+            nullif(
+                array_to_string([address, address2, city, state, zipcode], ', '), ''
+            ) as home_address,
+        from {{ ref("stg_focus__address") }}
+    ),
```
cbini: don't need this cte, add filter to join

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
cbini: Moved to `stg_focus__students_join_people` as null-preserving booleans (`custody = 'Y' as is_custodial`, etc.), replacing the raw `'Y'`/null strings. The `links` CTE is gone entirely — with the derivations upstream it was a pass-through, so the model now drives straight off the staging ref.

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
cbini: Moved to `stg_focus__people`. The `people` CTE is gone; `stg_focus__people` is joined directly and the `contact_first_name`/`contact_last_name` aliases now happen in the final SELECT.

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
cbini: Moved to `stg_focus__people_join_contacts`, along with `is_email_title` and the blank-normalized `value`. Side benefit: the `focus_unmapped_phone_contact_titles` test now reads those same staging columns instead of re-deriving the regexes, so the test can no longer drift from the model.

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
cbini: Done as you described: `not is_email_title` folded into `phones_ranked` and `phones_filtered` dropped. With the typing in staging, `phones` was also redundant, so the three CTEs collapsed to one. Left `phones_typed` separate — the rank has to be computed in one CTE and filtered in the next since `QUALIFY` is off the table.

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
cbini: Moved to `stg_focus__address`, derived inside the existing `source` CTE so the bare `address` reference still resolves to the column rather than the whole-row struct.

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
cbini: Dropped; the rank filter is now in the join — `left join phones_ranked as pp on l.person_id = pp.person_id and pp.overall_rank = 1`. `overall_rank` is a `row_number()` over `person_id`, so it still matches at most one row and cannot fan out.

src/dbt/focus/models/intermediate/int_focus__student_contacts.sql
cbini: Dropped `addresses` and joined `stg_focus__address` directly (`address_id` is its PK, so no fan-out). Kept `student_addresses` with its `select distinct` per your call — it is the structural guard against two residence-flagged rows for one student+address fanning out the join.

cbini:
Verified each finding against the code before acting on it. Three of four are addressed; one is a deliberate open question.

**Important — `INNER JOIN` to `stg_focus__students` can silently drop link rows.** Real, and the mechanism is described correctly. I ran your query against the live dlt source with staging's soft-delete filters applied: **0 orphan links** (of 1 link row today, so weak evidence — the risk is structural, not observed). Kept the `INNER JOIN` and documented it as an intentional scoping filter to students that exist in Focus, since that is what it is (it also supplies `local_student_id`). Your related recommendation — that a `relationships` test belongs on `stg_focus__students_join_people` per the "FK check belongs on the pre-join model" convention rather than on this intermediate — is the right long-term shape and would have caught this for free; noting it for the Phase 2 pass, when real contact data lands and the test would actually have rows to exercise.

**Minor 1 — tracking-issue ref inside a `description:`.** Correct, and it was mine. Removed the `(see issue 4585)` parenthetical from the kipptaf wrapper description; the rest of the sentence stands.

**Minor 2 — PII tagging on `student_id`/`person_id` diverges from the sibling.** Confirmed accurate: `int_finalsite__student_contacts.yml` does not tag `finalsite_enrollment_id`/`finalsite_contact_id`, and upstream `stg_focus__students_join_people.yml` does not tag `student_id`/`person_id`. Deliberately **not** changing this unilaterally — the repo's own PII reference lists student IDs as FERPA direct identifiers while noting the staging-layer precedent is narrower, and it says to confirm scope with a human rather than let it drift. Raised with the owner; left tagged (the safer direction) pending that call.

**Minor 3 — backtick-quoting `` `value` `` may be unnecessary.** Checked by trying it: unquoted does **not** pass. sqlfluff RF04 fires on the alias assignment specifically, which is why the pre-existing unaliased passthrough `value,` needed no quoting while `nullif(trim(value), '') as value` does. Backticks kept, now in the staging model where the derivation moved.

Separately, the owner's inline review moved four derivations (`'Y'`-to-boolean flags, `contact_name`, `phone_type`, `home_address`) into the staging layer and removed five now-redundant CTEs, so the intermediate is down from 166 lines to 91. Output surface is unchanged — verified by rebuilding both versions with an identical `--select` set and diffing the resulting row, after an initial comparison turned out to be contaminated by a stale-`--defer` dev shadow.

PR #4596, opened 2026-07-28
refactor(dbt): surface Focus decode layer to kipptaf

### LINKED ISSUE #4595: Focus decode layer is unreachable from kipptaf
The `focus` package has a complete custom-field decode layer — `int_focus__custom_field_options` plus seven `__pivot` models — with **zero consumers repo-wide**. None of it is exposed to kipptaf, so analysts working there see only coded values and hand-roll translations.

Two changes:

1. Expose the seven `__pivot` models as kipptaf sources in `models/focus/sources-kippmiami.yml`, following the existing deanslist/ADP precedent where pivots are sourced directly. Only decoded views are surfaced — `int_focus__custom_field_options` stays package-internal, since exposing the lookup table itself would just relocate the hand-rolled translation problem into kipptaf. A field not covered by a pivot gets added to the pivot in the package.
2. Deduplicate `int_focus__student_enrollment`, which re-rolls the option lookup against raw `stg_focus__custom_field_select_options` instead of using `int_focus__custom_field_options`. Its join also matches on option id only, never scoping to the `school_level` field, so it would silently mis-decode if Focus stored a code rather than an id.

Verified output-identical for the dedup: 10 schools, 0 mismatches.

Note that `int_focus__schools__pivot.school_level_label` is not a drop-in replacement for the enrollment model's `school_level` — the labels are `E - Elementary` / `M - Middle` / `H - High`, whereas kipptaf's network-wide contract is the `ES` / `MS` / `HS` abbreviation feeding `region_school_level`. Same lookup, different projection.

Reported by @GabyRangelB.

cbini:
# Pull Request

## Summary & Motivation

...make the `focus` package's custom-field decode layer reachable from kipptaf as
complete, already-decoded entity models, and collapse a duplicate school-level
decode onto the pivot that owns it.

**Background.** The `focus` package already contains a complete decode layer —
`int_focus__custom_field_options` plus seven `__pivot` models — with **zero
consumers repo-wide**. None of it was reachable from kipptaf, whose
`models/focus/sources-kippmiami.yml` exposed five `stg_focus__*` models and two
intermediates and nothing else. Analysts at the network level therefore saw coded
values and hand-rolled their own translations.

Reported by @GabyRangelB, after #4582 removed three kipptaf-level passthroughs of
the raw Focus code tables. That PR justified the removal as "they existed only to
feed the misplaced intermediate" — an inference about intent that was wrong. They
had been created to do code-to-value translation at the KTAF level.

### A pivot is not an entity spine

The first cut of this PR surfaced the seven `__pivot` models directly. That was
wrong, and row counts make it obvious. BigQuery `UNPIVOT` drops any entity whose
unpivoted columns are all null, so a pivot is a **decode companion**, correct only
when LEFT JOINed:

| entity | staging rows | pivot rows | dropped |
| --- | --- | --- | --- |
| `master_courses` | 71,334 | 2,166 | 69,168 (97%) |
| `courses` | 16,941 | 363 | 16,578 (98%) |
| `student_enrollment` | 9,866 | 1,124 | 8,742 (89%) |
| `schools` | 10 | 8 | 2 |
| `students` / `course_periods` / `users` | — | — | 0 |

Exposing those as standalone kipptaf sources invited reading them as entity
tables, which on `master_courses` silently returns 3% of the catalog.

### Changes

- **Six new combined package models** — `int_focus__schools`, `__students`,
  `__courses`, `__course_periods`, `__master_courses`, `__users`. Each is its
  staging model's columns plus the decoded `*_label` columns, driving from staging
  and LEFT JOINing the pivot, so the entity set is complete.
- **`int_focus__student_enrollment` absorbs its five labels** rather than gaining
  a sibling, since it was already the surfaced enrollment model. It also now reads
  schools through `int_focus__schools` instead of joining `stg_focus__schools` and
  the pivot separately.
- **The decode itself moved off a duplicate path.** `int_focus__student_enrollment`
  previously re-rolled the option lookup against raw
  `stg_focus__custom_field_select_options`, matching on option id alone and never
  scoping to the `school_level` field — so it would have silently mis-decoded if
  Focus stored a code rather than an id. That duplication predates #4582; it is on
  `main` and was inherited, not introduced, when that PR extended the model.
- **The seven pivots stay package-internal**, the same treatment as
  `int_focus__custom_field_options`. They are the mechanism, not the product.
- **kipptaf gets six `union_relations` wrappers plus source entries** for the
  combined models, following the existing `stg_focus__schools` /
  `int_focus__school_year_first_day` shape. Enrollment already had
  `int_focus__student_enrollments`.
- **New singular test `test_focus__school_level_labels_known`.** The school-level
  mapping key moved from the option `code` to the decoded label, which is the one
  thing this makes more fragile — a Focus label rename would null `school_level`
  silently. The test fails the build instead. `C - Combined` is known and
  intentionally unmapped, matching prior behavior.
- **PII tags on the wrappers.** `contains_pii` does not travel through `source()`,
  so `int_focus__students` and `int_focus__users` re-declare it at model level.
  Declared at model level rather than per column because these are `select *`
  passthroughs whose column docs live on the kippmiami source models.

### Why the raw lookups are not restored

Restoring them would relocate the translation problem into kipptaf rather than
remove it, which is the thing being complained about. For the same reason
`int_focus__custom_field_options` is not exposed — it is the lookup table, not a
decoded view, and surfacing it invites exactly the hand-rolled decode joins this
change eliminates. A field no pivot covers gets added to the pivot, in the package.

### Validation

- `dbt build` of all seven package models via the kippmiami project:
  **PASS=21, ERROR=0**, including the new guard test and a `unique` + `not_null`
  PK test on each new model.
- **Every combined model matches its staging row count exactly — zero drift**, so
  the `UNPIVOT` loss above is gone.
- `int_focus__student_enrollment` parity against prod: **9,866 rows both sides, 0
  `school_level` mismatches, 0 unmatched keys** on a full outer join by
  `student_enrollment_id`; 1,124 populated `prior_state_label`, matching the pivot's
  row count.
- Both decode rewrites were proven equivalent in BigQuery before editing: 10
  schools, 0 mismatches, with `C - Combined` resolving to NULL under old and new
  alike.
- All six kipptaf wrappers compile against `zz_stg_kippmiami_focus` with columns
  resolved — 27/43/40/73/59/38 columns respectively, not an empty expansion.
- `trunk check --force` clean across all changed files.

### Staging seeded

The six new package models did not exist in `zz_stg_kippmiami_focus`, and
`union_relations` resolves columns at compile from `INFORMATION_SCHEMA` — so CI
would have expanded to nothing and failed at build, not at compile. All seven were
built into `zz_stg_kippmiami_focus` with `--target staging` (PASS=21, ERROR=0)
before this was pushed. Six are new tables; the enrollment rebuild is additive.

Closes #4595

## AI Assistance

AI-assisted via Claude Code, human-directed. Four of Claude's proposals were
rejected in sequence and every rejection improved the result: that the deleted raw
passthroughs should be restored (no — decode once in the package, surface the
decoded view); that `int_focus__custom_field_options` should be exposed alongside
the pivots (no — it is the mechanism, not the product); that the enrollment model
could keep its own decode because the pivot's label was not a drop-in (no — the
pivot owns that decode, and the label maps); and that surfacing the pivots was
sufficient (no — combine staging and labels before surfacing, which is what
exposed the 97% row loss above).

Claude did the duplication scan, the equivalence and row-count proofs, and the
edits. The scan also cleared two near-misses left alone as legitimately separate:
`int_focus__custom_field_log__unpivot` filters a different owner class
(`CustomFieldLogColumn`), and kippmiami's `rpt_focus__student_enrollment` joins the
enrollment codes in the encode direction, for write-back into Focus.

A follow-up inventory of what the pivots still do not decode — 61 populated fields,
including `CourseCatalog.course_level` at 43,049 rows — is filed as #4597.

## Self-review

### General

      **Note:** it ran against the first commit only and does not re-run on
      `synchronize`; its one Minor finding targeted a join since deleted. See the
      reply thread.

### dbt

- [x] Properties file included for all twelve new models. The six kipptaf wrappers
      carry a description and the `_dbt_source_project` doc only, per the
      union-passthrough convention; the six package models carry a description and
      a PK uniqueness test, with column docs on their two upstreams.
- [x] No exposure changes. The new models gain no downstream consumers in this PR;
      they become reachable, which is the point.
- [x] **Breaking change?** No. No contracted column is renamed or removed.
      `int_focus__student_enrollment` gains five columns (additive) and its existing
      output is unchanged — see Validation. Rollback is a plain revert plus dropping
      the six seeded `zz_stg` tables.
- [x] No external sources added or modified, so no `stage_external_sources` run is
      needed. Staging was seeded with `dbt build --target staging` instead — see
      above.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

## Troubleshooting


cbini:
Review was against `937509744` only. Two commits have landed since, so recording where its findings stand.

**Minor 1 — bare `custom_100000004` / `SISSchool` literals need an explanatory comment: no longer applicable.** `b0b63763a` removed that join entirely. The model no longer decodes the school-level custom field itself; it reads `school_level_label` from `int_focus__schools__pivot`, which already owns that decode, and maps the label to the `ES`/`MS`/`HS` abbreviation the network contract needs. `grep` confirms neither literal remains in the file. The underlying observation was fair — the fix went further than a comment.

**On the PII assessment.** The claim that package-level `contains_pii` makes kipptaf-side declaration unnecessary is right for the *source entries*, and no entry in `sources-kippmiami.yml` declares PII. But `b0b63763a` also added seven kipptaf `union_relations` wrapper **models**, which are a separate surface the review predates. The package tag does not travel through `source()`, and the wrapper is what consumers read.

Three of the seven wrap PII-carrying pivots — `students`, `student_enrollment`, `users`. `f294c86bd` tags those three at model level, matching `int_focus__student_enrollments`, the existing tagged model in the same directory. Declared at model level rather than per column, because these are `select *` passthroughs whose column docs deliberately live on the kippmiami source models. The four non-people pivots stay untagged, mirroring the package.

Worth noting separately: kipptaf's `stg_focus__students` passes through a `contains_pii`-tagged package model and carries no tag. That looks like a pre-existing gap rather than a precedent, and is out of scope here.

`claude-review` does not re-run on `synchronize`, so the two newer commits are unreviewed by it. Validation for them is in the PR body — dev build PASS=7 with a new guard test, exact row and value parity against prod, and all seven wrappers compiling against `zz_stg_kippmiami_focus`.

PR #4608, opened 2026-07-29
refactor(dbt): normalize Finalsite phones at the staging and pivot layers

### LINKED ISSUE #4599: refactor: normalize Finalsite phone numbers at the staging and pivot layers
`clean_phone` is currently applied at the end of `int_finalsite__student_contacts` (4 call sites). Move it to the earliest point each value exists as a scalar column, so every consumer reads normalized phones.

**Where the cleanup goes**

- `stg_finalsite__contacts` — `phone_1_number`, `phone_2_number`, `phone_3_number`
- `int_finalsite__contact_custom_attributes` — the `emrg_N_phone_M_number` and `additional_phone_number_N_phone_number` fields. These are NOT cleanable in staging: they live inside the `custom_attributes` repeated struct and only become scalar columns at this pivot.
- `int_finalsite__student_contacts` then drops all four `clean_phone` calls.

**Consequence needing confirmation before merge**

`stg_finalsite__contacts.phone_1_number` feeds two live SFTP import payloads — `rpt_focus__contacts` (`contact1_value` / `contact2_value`) and `rpt_focus__addresses` (`phone`). Their format changes from bare 10-digit to `+1NNNNNNNNNN`.

This is desirable in itself (import-seeded Focus contacts then arrive already normalized), but **someone must confirm Focus's import accepts the `+1` form** or imports will fail. The feeds are import-once, so the change affects future imports only.

**Consumers verified unaffected**

`rpt_clever__students` and `rpt_deanslist__family_contacts` both read `int_students__contacts` (already E.164) and strip non-digits, so they are insensitive to moving the cleanup earlier. Checked because `src/dbt/CLAUDE.md` documents a prior incident where an E.164 change broke `rpt_clever__students`.

**Source-format context** (from live data)

Finalsite `phone_1_number` is ~99% bare 10-digit, no punctuation, no extensions. Focus's natively-entered format is `(NNN) NNN-NNNN`, so read-back normalization on the Focus side is still required for registrar-entered numbers regardless of this change — tracked separately in #4585.

Split out of #4594 to keep that PR scoped to the Focus contacts models.

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will move the `clean_phone` E.164 normalization
out of `int_finalsite__student_contacts` and into the earliest point each phone
value exists as a scalar column, so every consumer reads one format instead of
only the contact receivers.

| Model | Change |
| --- | --- |
| `stg_finalsite__contacts` | wrap `phone_1_number`, `phone_2_number`, `phone_3_number` |
| `int_finalsite__contact_custom_attributes` | wrap the 12 `emrg_N_phone_M_number` fields, `additional_phone_number_1_phone_number`, `med_doctor_phone_number`, `med_hospital_phone_number` |
| `int_finalsite__student_contacts` | drop all four `clean_phone` calls |

The custom-attribute phones are not reachable in staging: they arrive inside the
`custom_attributes` repeated struct and only become scalar columns at that
pivot.

The two medical phone fields are beyond what Refs #4599 listed. They are phone
numbers at the same pivot with the same problem, nothing consumes them yet, and
the widened scope was confirmed before implementing.

### Accepted risk - Focus import format

`stg_finalsite__contacts.phone_1_number` feeds two live SFTP import payloads:
`rpt_focus__contacts` (`contact1_value` / `contact2_value`) and
`rpt_focus__addresses` (`phone`). Their format changes from bare 10-digit to
`+1NNNNNNNNNN`.

Whether Focus accepts the `+1` form is unconfirmed, and that risk has been
accepted rather than blocking the merge on it. Quantified exposure:

| Feed | rows pending import | rows carrying a phone |
| --- | --- | --- |
| `rpt_focus__addresses` | 1,402 | 1 |
| `rpt_focus__contacts` | 2,581 | 2,552 |

The ADDRESS feed sources `phone` from the enrolled student's own contact record,
which is almost never populated, so its exposure is a single row. The CONTACTS
feed is the real one. Focus currently holds 3 phone values of its own, all in
the `(NNN) NNN-NNNN` form, so the import-once anti-join has not yet delivered at
scale - this change lands on the first bulk contacts import of roughly 2,552
phone-bearing rows, not an incremental tail.

The in-repo `focus-api-spec.md` names the guardian phone fields but documents no
format constraint, so this cannot be settled from data we hold. Focus stores a
punctuated format natively, which suggests a free-text column that would accept
E.164, but that is inference rather than evidence. If an import does reject the
form, the rollback below restores the prior format. Read-back normalization for
registrar-entered Focus numbers is tracked in #4585.

Rollback plan: revert the commit. No column names or types change, so no
downstream contract breaks and no rematerialization sequencing is required.

### Verification

Behavior was checked against Miami prod data by dry-running the compiled model
SQL and diffing the new output against the currently-materialized tables
(aggregates only, no phone values left the local environment):

| Field | populated | resolve to valid `+1NNNNNNNNNN` | pass through as entered | newly null |
| `stg` `phone_1_number` | 3,824 | 3,759 | 65 | 0 |
| `stg` `phone_2_number` | 131 | 131 | 0 | 0 |
| `stg` `phone_3_number` | 137 | 135 | 2 | 0 |
| `emrg_N_phone_M_number` | 1,279 | 1,263 | 16 | 0 |
| `additional_phone_number_1_...` | 146 | 146 | 0 | 0 |
| `med_doctor_phone_number` | 580 | 549 | 31 | 0 |
| `med_hospital_phone_number` | 115 | 109 | 5 | 0 |

Pass-throughs are values that fail NANP validation or carry no parseable digits.
The macro never nulls a non-null input, which the zero newly-null column
confirms.

Also verified: the result schema of all three models is unchanged (39 / 230 / 18
columns, every phone column still `STRING`), so contract enforcement is
unaffected.

### Consumers checked

`rpt_clever__students` and `rpt_deanslist__family_contacts` read
`int_students__contacts`, which was already E.164, and both strip non-digits, so
they are insensitive to moving the cleanup earlier. Checked because
`src/dbt/CLAUDE.md` documents a prior incident where an E.164 change broke
`rpt_clever__students`.

### Tests

`int_finalsite__student_contacts` no longer normalizes, so its
`test_student_contacts_phone_normalization` unit test could not keep asserting
normalization. It is retargeted as `test_student_contacts_phone_typed_slots`
(typed-slot resolution, pre-normalized inputs), and the normalization coverage
moves to a new `test_contacts_phone_normalization` on `stg_finalsite__contacts`
which also covers the household-address normalization in the same select.

The Focus extract unit-test fixtures in `rpt_focus__contacts` and
`rpt_focus__addresses` are updated to E.164. They passed either way since those
models only pass the value through, but the raw dashed fixtures no longer
reflect reality upstream.

Regression coverage for the 15 `clean_phone` calls at the custom-attributes
pivot is deferred to #4612 - a narrow-column unit test is not possible on a
230-column model, so it needs a singular invariant test instead. Details and
the baseline counts are in that issue.

Unit tests run locally, all passing: 11 in `kippmiami` (finalsite plus every
Focus reconciliation test), 7 in kipptaf `extracts.focus`, 2 in kipptaf
`extracts.deanslist`.

### CI expectations

This is a value-only change to a source-system package plus properties-yml edits
in kipptaf. No column set changes, so no `zz_stg` seeding is needed and the dbt
Cloud CI run will be close to a no-op rather than real validation. The models are
first exercised by Dagster.

## AI Assistance

Claude Code authored the implementation, the prod-data verification queries, and
the unit-test changes. Human-directed: the decision to include the two medical
phone fields beyond the issue scope, and the choice to work in a worktree.

### General

### dbt

- [x] Properties files updated for every changed model (descriptions now state
      where normalization happens)
- [x] No exposure changes needed — no new or removed models
- [x] **Not a breaking change** — no column renames or removals; only the
      formatting of existing `STRING` values changes. The Focus import format
      change above is an accepted risk, not an open blocker.
- [x] No external source added or modified, so no `stage_external_sources` run
      is required

## CI checks

- [ ] **Trunk** — passes. Ran `trunk check --force` on all 8 changed files
      locally: no issues after the commit-time format.
- [ ] **dbt Cloud** — passes.

Refs #4599


cbini:
Response to the Claude Code Review findings.

## Finding 1 — gap confirmed, recommended fix is not workable

The coverage gap is real and every citation checks out: `int_finalsite__contact_custom_attributes.yml` has no `unit_tests:` block, and both student-contacts unit tests do mock the pivot with `rows: []` at the cited lines 223-224 and 412-413. The 15 `clean_phone` calls at the pivot have no test that CI re-runs.

The specific recommendation — a `format: sql` unit test mirroring the new `test_contacts_phone_normalization` on `stg_finalsite__contacts` — does not work on this model, and this was tried before the PR was opened. dbt projects the model's FULL column list onto the `expect` fixture, so a narrow expect on a 230-column pivot fails with:

```text
Unrecognized name: academy_business_ss at [1383:32]
```

Mirroring the staging pattern is only possible there because that model has 39 columns. A working unit test at the pivot would need all 230 columns enumerated in `expect`, roughly 215 of them typed nulls, which then breaks every time a field is added to the pivot.

Deferring the fix on this PR per author decision. The preferred approach when it is picked up is a singular data test asserting the invariant that no populated phone field retains a NANP-valid bare-10-digit or leading-1 11-digit form. That is one file, covers all 15 pivot fields plus the 3 staging fields, re-runs in CI and Dagster, and matches the repo idiom for this shape.

## Finding 2 — could not verify, and it favors the reviewer's reading

Not verifiable with the tooling available here: the dbt MCP exposes no artifact-fetch tool, so `run_results.json` for the CI run could not be read. A local `dbt ls --select state:modified+` does list `test_addresses_shape` and `test_contacts_two_guardians`, but it also fanned out across dozens of unrelated staging models, which is the known target-mismatch artifact documented in `src/dbt/CLAUDE.md` — so that run is not evidence of what CI actually selected. Noting it as unresolved rather than dismissing it.

## New information on the open Focus blocker

Checked the Focus side rather than leaving the `+1` question abstract. It narrows the blast radius to one feed and raises the stakes on that one:

| Feed | rows pending import | rows carrying a phone |
| --- | --- | --- |
| `rpt_focus__addresses` | 1,402 | 1 |
| `rpt_focus__contacts` | 2,581 | 2,552 |

The ADDRESS feed sources `phone` from the enrolled student's own contact record, which is almost never populated, so its exposure is a single row. The CONTACTS feed sources from the guardian record and is the real exposure.

Focus currently holds almost no phone data of its own (3 values, all in the `(NNN) NNN-NNNN` form), so the import-once anti-join has not yet delivered at scale. That means this format change lands on the first bulk contacts import of roughly 2,552 phone-bearing rows, not on an incremental tail — the PR body's "future imports only" is accurate but understates it. The in-repo `focus-api-spec.md` lists the guardian phone field names but documents no format constraint, so this cannot be settled from anything we hold and needs Focus support or a small test import.

Refs #4599

PR #4579, opened 2026-07-27
feat(dbt): DeansList ContactType column and contact_2 mart keying

cbini:
# Pull Request

## Summary & Motivation

> "When merged, this pull request will..."

Add the `ContactType` column to the DeansList `contacts.txt` extract and prepare the kipptaf consumers for `contact_2` (Parent 2) rows — the kipptaf half of #4569 (companion to #4578, the finalsite package half).

- `rpt_deanslist__family_contacts`: new `ContactType` column (`Parent1` / `Parent2` / `Emergency`), positioned between `Relationship` and `Language` per the DeansList template, so DeansList can send automated reports/texts/emails to guardians only. The PK widens to `(StudentID, ContactType, Relationship)` — two parents can legitimately share a `rel_type` (e.g. both `parent`), which would have broken the old `(StudentID, Relationship)` uniqueness test the moment Parent 2 rows arrive.
- `bridge_student_contacts` / `dim_student_contact_persons`: person-key `contact_2` the same way as `contact_1` (a second parent is a real, sibling-shared contact record, unlike the free-text emergency slots). Both models change together so bridge and dim keys stay consistent.
- `int_students__contacts`: admit `contact_2` in the `contact_slot` accepted values; description updates. The pivot model already carried a NULL `contact_2_*` column surface — comment updated.

Everything here is a no-op until #4578 merges and district prod rebuilds produce `contact_2` rows; the two PRs are order-independent but both are required for the Asana asks (Emergency-contact import is blocked in DeansList until `ContactType` lands).

Validated locally: full dev build of the five touched models plus tests — 27 pass, 2 pre-existing warns (8 null `ParentLastName` rows, identical count in prod; 20 bridge `student_key` orphans, known dev-defer drift). Both rpt unit tests updated for `ContactType` and extended with a Parent2 row exercising the new case branch.

Open question flagged to Ops (not changed here): emergency rows currently send `Relationship = 'Emergency 1'` (with a space); DeansList's sample shows `Emergency1`. With `ContactType` now distinguishing emergencies, the real relationship could go in `Relationship` instead — awaiting confirmation with DeansList before changing behavior.

## AI Assistance

Claude Code authored this PR end-to-end, directed by @cbini (gate definition and scope decisions).

### General

- [ ] Update **due date** and **assignee** on the
      [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
- [x] Review the **Claude Code Review** comment posted on this PR.

### dbt

[...]

PR #4611, opened 2026-07-29
fix(dbt): anchor Finalsite Parent 2 to the Parent 1 household

### LINKED ISSUE #4610: fix: Finalsite Parent 2 pulled from the wrong household
Parent 2 (`contact_2`) on the NJ Finalsite contact feeds is being pulled from a household the student's Parent 1 does not live in. Reported by Laszlo de Simon. Follow-up to #4569; PRs #4578 and #4579 are merged and live in prod.

Student identifiers for the cases below are held locally and in the Slack thread, not recorded here.

## Reproduction

A Newark student with two households. Prod `kipptaf_finalsite.int_finalsite__student_contacts` gives `contact_1` = the primary parent, whose only household is in Newark, and `contact_2` = a stepparent whose only household is in Jersey City.

## Root cause

The shipped gate requires the Parent 2 candidate to belong to `households[safe_offset(0)]` of the student. Array position does not identify Finalsite's primary household.

For the student above, the Finalsite UI labels the Newark household `Household1 (Primary)` — and it is offset **1** in the API array. Offset 0 is the Jersey City household, which the stepparent belongs to, so he passes the gate.

The `Household` API object carries only id plus address fields, with no primary flag and no ordinal. Confirmed three ways: the vendor spec at `docs/superpowers/specs/references/finalsite-api-spec.yml`, our Pydantic model in `src/teamster/libraries/finalsite/api/schema.py`, and the ingested Avro external table. The per-student primary-household designation exists in the Finalsite UI but is absent from every field we ingest — the `custom_attributes`, `id_attributes`, and `track_attributes` arrays carry nothing household-related either.

Second defect: the household condition was one-sided. `contact_1` was picked purely on the `primary` / `financial` flags with no household condition at all, so nothing ever required the two parent slots to share a household.

## Agreed rule

Confirmed with Ops:

1. A student's Parent 1 is their primary contact.
2. A student's primary household is the household to which their primary contact is associated.
3. Any other primary or financial contact within that household is Parent 2.

Household matching is on household id. The `is_parent2` student custom field is dropped from the gate — it is unreliable in both directions: false for students who do have a co-resident second parent, and true for students who have none.

## Impact across Newark, Camden and Paterson

| Slot | Now | After |
| --- | --- | --- |
| `contact_1` rows | 15,957 | 14,983 |
| `contact_2` rows | 4,360 | 4,741 |

**The `contact_1` reduction is intended and needs explicit Ops sign-off before merge.** Rule 1 removes the current financial fallback, so a student with no relationship flagged `primary` no longer receives a Parent 1 at all. That is 974 rows, 483 of them enrolled students, who will have no parent contact in the DeansList feed.

For context on the underlying data gap: 673 enrolled students have no contact flagged `primary`, and 192 of those have no contact flagged `primary` or `financial` and therefore already receive no Parent 1 today.

## Known limitation

Duplicate household records are common — the same street address entered two or more times as separate households. Matching on household id therefore still excludes a Parent 2 who lives at Parent 1's address but is attached to a different household record. Roughly 34 students network-wide. Matching on a normalized address instead would resolve them; deferred as a follow-up.

## Out of scope

- The same `households[safe_offset(0)]` assumption feeds the scalar address columns on `stg_finalsite__contacts`, which flow into the Focus ADDRESS feed for Miami. Different region, different consumer — to be tracked separately.
- Whether Finalsite can expose the primary-household designation through the external API. A local probe script exists to capture the raw response and diff it against our declared fields.

cbini:
# Pull Request

## Summary & Motivation

...stop Finalsite Parent 2 from being pulled out of a household that Parent 1 does not live in, and redefine the parent slots per the rule Ops confirmed.

Closes #4610. Follow-up to #4569 (PRs #4578, #4579, both merged and live).

**Root cause.** `contact_2` was gated on `households[safe_offset(0)]` of the student. Array position does not identify Finalsite's primary household — for the reported student the UI's `Household1 (Primary)` is offset **1**, so a stepparent belonging only to the offset-0 household passed the gate. The `Household` API object carries id plus address fields and nothing else (verified against the vendor spec, our Pydantic model, and the ingested Avro), so the designation cannot be read, and no ordering of the array or of the household ids reproduces it. The gate was also one-sided: `contact_1` was chosen on the `primary` / `financial` flags with no household condition, so nothing required the two slots to share a household.

**New rule**, as Ops stated it:

1. A student's Parent 1 is their primary contact.
2. A student's primary household is the household to which their primary contact is associated.
3. Any other primary or financial contact within that household is Parent 2.

The `is_parent2` student custom field is dropped from the gate. It is unreliable in both directions — false for students who do have a co-resident second parent, true for students who have none — so it removed real rows without excluding any wrong ones.

### Impact, and the part that needs sign-off

Across Newark, Camden and Paterson:

| Slot | Now | After |
| --- | --- | --- |
| `contact_1` rows | 15,957 | 14,983 |
| `contact_2` rows | 4,360 | 4,741 |

**The `contact_1` reduction is deliberate and should not be merged without Ops sign-off.** Rule 1 removes the existing financial fallback, so a student with no relationship flagged `primary` no longer gets a Parent 1 at all — 974 rows, 483 of them enrolled students, who will reach the DeansList feed with no parent contact. This was raised as a cost before the rule was chosen and accepted knowingly; flagging it here so the decision is visible at merge rather than discovered downstream.

Underlying data gap for context: 673 enrolled students have no contact flagged `primary`; 192 have neither `primary` nor `financial` and already get no Parent 1 today.

### Known limitations

Both are cases where a student who arguably should get a Parent 2 does not. Neither is a regression against the reported bug; both are consequences of anchoring on Parent 1's household.

1. **Duplicate household records.** One street address is often entered several times as separate households. Because matching is on household id, a Parent 2 living at Parent 1's address but attached to a different household record is excluded. Roughly 34 students network-wide. Normalized-address matching would resolve them; deferred rather than folded into an interim patch.

2. **Parent 1 with no household.** If Parent 1's own `household_ids` is empty, the primary household resolves to nothing and no Parent 2 can match. 300 student-to-Parent-1 links network-wide have this shape (Newark 189, Camden 95, Paterson 16), but only **21** of them have another caregiver sharing the student's household who would otherwise qualify. Raised by the automated review; quantified against prod rather than left as a hypothesis.

### Validation

- `dbt build` of `int_finalsite__student_contacts` through `kippnewark` with prod deferral — model plus 2 unit tests plus 3 data tests all pass.
- The `unique_combination_of_columns` test passing on real Newark data confirms the `primary` singleton assumption the new `contact_1` relies on.
- Newark dev output: `contact_1` 11,169 to 10,543, `contact_2` 3,222 to 3,465. The 10,543 equals exactly the count of Newark students holding a `primary` relationship, which is the rule restated.
- Spot-checked six named students, including the reported case (Parent 2 correctly dropped), a known-good case (unchanged), and a student whose Parent 1 lives outside the UI's primary household (Parent 1 correctly retained — an earlier candidate rule would have removed her).
- Unit tests rewritten: one now asserts a Parent 2 is still selected with `is_parent2` false on every fixture row, the other that a student with no primary yields no rows at all.
- Second commit drops the now-inert `household_1_id` values from those fixtures (they no longer drive any behavior); build re-run clean afterward.

**CI note:** the two dbt Cloud runs on this PR behaved differently and the second one is the meaningful record. The first (commit 1) selected essentially nothing — a 30s no-op, which is the usual outcome for a source-package-only change. The second (commit 2) drew a different `deferring_run_id`, so the state baseline shifted and it rebuilt a wide slice of the `kipptaf` graph over 2m45s. That run reported 9 warn-level test results; all 9 are on descendants of this model, and all 9 match prod's current counts exactly, so they are pre-existing rather than introduced here. Detail in the review reply below.

## AI Assistance

Claude did the investigation, the SQL, and the validation. The rule in the three numbered points was specified by a human after reviewing the Finalsite UI directly; several earlier candidate rules Claude proposed were falsified against that UI evidence and discarded. The decision to remove the Parent 1 financial fallback was human, made with the row-loss figure in hand.

## Self-review

### General

- [x] Include a `[model_name].yml` properties file for all new models — model
      description, the `is_household_member` note, and both unit tests updated
- [ ] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — n/a, no
      new consumers
- [x] **Breaking change?** No columns renamed or removed; the contract is
      unchanged. Row membership changes in both parent slots — see the impact
      table above. Rollback is a revert of this commit; the model is a table
      rebuilt on each run, so no backfill is required.
      with **`--target staging`** — n/a

## CI checks

- [x] **Trunk** — passing. `trunk check --force` also run locally on both changed
      files, clean.
- [x] **dbt Cloud** — passing. The run on the current head rebuilt a wide graph
      (2m45s) and reported 9 warn-level tests; each one was checked against prod
      and matches its current count exactly, so none are introduced by this PR.
      See the CI note above.

### DIFF
--- a/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
+++ b/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
@@ -1,90 +1,83 @@
 with
-    contact_1_candidates as (
-        -- contact_1 is the student's FIRST reportable parent contact. Finalsite
-        -- has no explicit contact rank, so we take the relationship it flags
-        -- `primary` (its parent1 designation) and fall back to the one flagged
-        -- `financial` when no primary is set — the two flags Ops maintains to
-        -- mark the responsible caregiver. A record with neither flag gets no
-        -- contact_1 (a Finalsite data-entry gap for Ops to resolve).
-        -- `primary`/`financial` are NULL (not false) when unset, so normalize
-        -- to false here to keep the downstream rank ordering deterministic. No
-        -- SIS scoping — downstream receivers filter to enrolled students by
-        -- joining on the student id.
-        select
-            finalsite_enrollment_id,
-            relationship_id,
-            rel_id,
-            rel_name,
-            rel_type,
-            household_1_id,
-            is_parent2,
-
-            coalesce(is_primary, false) as is_primary,
-            coalesce(is_financial, false) as is_financial,
-        from {{ ref("stg_finalsite__contact_relationships") }}
-        where is_primary or is_financial
-    ),
-
-    contact_1_ranked as (
-        -- Rank primary above financial; within a tier break on relationship_id.
-        -- Ties occur only among multiple `financial` relationships (a student
-        -- never has two `primary`); relationship_id is an arbitrary but stable
-        -- tiebreak — every candidate in a tier is a valid contact, and
-        -- Finalsite exposes no field that reproduces a caregiver ordering.
-        select
-            finalsite_enrollment_id,
-            rel_id,
-            rel_name,
-            rel_type,
-            household_1_id,
-            is_parent2,
-            is_primary,
-
-            row_number() over (
-                partition by finalsite_enrollment_id
-                order by is_primary desc, is_financial desc, relationship_id asc
-            ) as rn,
-        from contact_1_candidates
-    ),
-
     contact_1_picked as (
+        -- contact_1 is the student's Parent 1: the relationship Finalsite flags
+        -- `primary`. That flag is a per-student singleton and is never `false`
+        -- — it is true or NULL — so a bare `where is_primary` selects exactly
+        -- the Parent 1 row. A student with no primary relationship gets no
+        -- contact_1, and therefore no contact_2: per Ops, a missing primary
+        -- flag is a Finalsite data-entry gap to fix at the source rather than
+        -- something to infer from `financial`. A second primary on one student
+        -- would surface as a duplicate contact_1 and fail this model's
+        -- uniqueness test, which is the intended loud failure. No SIS scoping —
+        -- downstream receivers filter to enrolled students by joining on the
+        -- student id.
         select finalsite_enrollment_id, rel_id, rel_name, rel_type,
-        from contact_1_ranked
-        where rn = 1
+        from {{ ref("stg_finalsite__contact_relationships") }}
+        where is_primary
+    ),
+
+    primary_household_ids as (
+        -- The student's primary household is the household their Parent 1
+        -- belongs to. Finalsite's own household-1 designation is per-student and
+        -- set in the UI, but it is absent from every field the API exposes (the
+        -- Household object is id plus address only), and array position does not
+        -- reproduce it — `households[safe_offset(0)]` is the UI's Household 2
+        -- for confirmed students. Parent 1's membership is therefore what
+        -- defines the household both parent slots must share. Exploded to one
+        -- row per household so the contact_2 match below is a plain join.
+        select p.finalsite_enrollment_id, p.rel_id as contact_1_rel_id, household_id,
+        from contact_1_picked as p
+        inner join
+            {{ ref("stg_finalsite__contacts") }} as cp
+            on p.rel_id = cp.finalsite_enrollment_id
+        cross join unnest(cp.household_ids) as household_id
+    ),
+
+    contact_household_ids as (
+        -- One row per (contact, household) so household co-membership is a join
+        -- rather than an array containment check.
+        select finalsite_enrollment_id, household_id,
+        from {{ ref("stg_finalsite__contacts") }}
+        cross join unnest(household_ids) as household_id
     ),
 
     contact_2_candidates as (
-        -- contact_2 is the student's SECOND reportable parent (DeansList
-        -- "Parent 2"). Finalsite encodes it as an additional relationship
-        -- flagged `financial` without `primary` (`primary` is a per-student
-        -- singleton — Parent 1; candidates are financial-only rows, so
-        -- `not is_primary` alone expresses that and also guards a hypothetical
-        -- two-primary record). Scoped conservatively per Ops: only when the
-        -- student's own `is_parent2` custom field is true AND the related
-        -- contact is a member of the student's first household ("household
-        -- 1") — second parents in other households are intentionally
-        -- excluded. rn > 1 skips the ROW picked as contact_1 (the financial
-        -- fallback when no primary is set); the rel_id inequality against the
-        -- pick additionally skips any other relationship row to the same
-        -- PERSON, so contact_2 can never duplicate contact_1. The student-side
-        -- gate fields (is_parent2, household_1_id) ride on the relationships
-        -- staging grain, so only the related contact's record is joined here.
-        select r.finalsite_enrollment_id, r.rel_id, r.rel_name, r.rel_type, r.rn,
-        from contact_1_ranked as r
+        -- contact_2 is any OTHER contact flagged `primary` or `financial` that
+        -- belongs to the primary household. `primary` is a singleton already
+        -- taken by contact_1, so in practice these are financial-only rows; the
+        -- disjunction states the rule as Ops expressed it and guards a
+        -- hypothetical second primary. The rel_id inequality skips every
+        -- relationship row pointing at the PERSON already chosen as contact_1,
+        -- so contact_2 can never duplicate contact_1. The student's own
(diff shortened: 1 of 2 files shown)


cbini:
Verified each item against the code and against prod before responding.

### Minor 1 — dead fixture column: fixed

Fixed in 7fe7c9c7b. Confirmed first that the model SQL carries no `household_1_id` reference at all — the only remaining hits repo-wide are the two staging models that define the column, plus gitignored `target/` build artifacts. Removed all 12 inert fixture values. `is_parent2 = false` stays, since it is load-bearing for what the test asserts. Build re-run clean afterward, 6/6.

### Minor 2 — `contact_household_ids` explosion: declining

The suggested inline cannot keep the co-membership predicate in an `ON` clause. `cross join unnest(cp.household_ids)` produces a correlated row source, so the `h.household_id = household_id` comparison has to move into `WHERE`. `src/dbt/CLAUDE.md` steers join conditions into `ON` and reserves `WHERE` for row filters on the preserved table, so the restructure trades a documented CTE for a less conventional predicate placement.

Scale does not force that trade: the widest region is roughly 24k contact rows with small household arrays, and the model builds in about 3 seconds. Worth revisiting only if it shows up in `JOBS_BY_PROJECT`.

### Needs local verification — ran it, and it is a real gap

| Region | Parent 1 has no household | Of those, a second caregiver shares the student's household |
| --- | --- | --- |
| Newark | 189 | |
| Camden | 95 | |
| Paterson | 16 | |
| **Total** | **300** | **21** |

So 300 student-to-Parent-1 links have the shape, but only 21 students actually lose a Parent 2 they would otherwise get. Added to the PR body as a second known limitation alongside the duplicate-household case.

One methodology note for anyone re-running it. The suggested query is correct as written against `stg_finalsite__contacts`, because that model builds `household_ids` with `array(select ...)`, which yields an empty array rather than NULL. Reconstructing the same array from the raw `contacts` source with `array_agg(... ignore nulls)` returns NULL for a household-less contact, and `array_length(null) = 0` evaluates to NULL rather than true — so that variant silently reports zero. First attempt here hit exactly that.

### Recommendation — `household_1_id` in the staging contract

Grep claim confirmed: this PR removes the last SQL consumer. Leaving the column in place. Dropping a contracted staging column is a cross-project change requiring the district-then-kipptaf rollout, which is out of scope for an interim patch.

### Recommendation — CI boxes

Ticked. dbt Cloud is confirmed trivially green rather than validating: the `state:modified+` step ran 30 seconds against this branch's SHA with zero warnings from `get_job_run_error warning_only=true`, which is the documented no-op signature for a source-package-only PR.

### On the downstream claim

Independently confirmed. `rpt_deanslist__family_contacts` is driven from this model, inner-joined to the id crosswalk and to enrolled students, so a student with no `contact_1` contributes no Parent1 row rather than erroring.

One detail worth having in front of the merge decision: such a student still appears in the extract if they have emergency contacts, and drops out entirely only if they have neither.

### Merge gate

Agreed that the financial-fallback removal is a business decision rather than a code issue. It is flagged in the PR body with the row counts and needs Ops sign-off before merge.

cbini:
Follow-up on the dbt Cloud warnings, since the run on the current head is not the same shape as the first one.

Commit 1's CI run was a 30s no-op. Commit 2 drew a different `deferring_run_id`, the state baseline shifted, and it rebuilt a wide slice of the graph over 2m45s — surfacing 9 warn-level tests.

My first instinct was to call them unrelated models. That was wrong: `dbt ls --select int_finalsite__student_contacts+` returns 150+ models, and all 9 warned models are in it. This model feeds a student scaffold consumed across most of the kipptaf graph, so "unrelated" was not available as an explanation.

Checked each assertion against prod instead:

| Assertion | Prod | PR branch |
| --- | --- | --- |
| `fct_survey_responses.survey_response_key` | 46,062 | 46,062 |
| `rpt_tableau__gradebook_gpa` combo | 4,332 | 4,332 |
| `fct_survey_submissions.survey_submission_key` | 2,559 | 2,559 |
| `rpt_gsheets__deanslist_mod_audit` combo | 800 | 800 |
| `rpt_tableau__kfwd_career_launch_survey` combo | 127 | 127 |
| `rpt_tableau__student_course_grades` combo | 12 | 12 |
| `rpt_appsheet__kfwd_career_launch_survey.contact_id` | 7 | 7 |
| `rpt_tableau__ap_assessment_dashboard` combo | 3 | 3 |
| `int_students__graduation_path_codes` accepted values | 1 | 1 |

All nine identical, so they are pre-existing and unchanged by this PR rather than something it introduced. Not filing trackers for them here — that is a separate cleanup with its own scope.

Issue #4607, opened 2026-07-29
refactor(cube): remove the CUBE_GROUP_MAP dev bypass and stub BigQuery in its tests


cristinabaldor:
## Summary

Delete the `CUBE_GROUP_MAP` dev bypass from `resolveAccess` in `src/cube/cube.js`
and convert the tests that depend on it to stub BigQuery directly. The stubbing
pattern already exists in `src/cube/cube.test.js`, so this is mechanical rather
than exploratory.

## Why

The bypass short-circuits the BigQuery identity reads and returns a hardcoded
email-to-groups map. Shape of it, with the two deployment variables read from the
process environment shown bare for readability:

```js
if (NODE_ENV !== "production" && CUBE_GROUP_MAP) {
  const map = JSON.parse(CUBE_GROUP_MAP);
  const ctx = {
    ...access.buildSecurityContext(null, []),
    groups: map[email] ?? [],
  };
  ...
}
```

Four problems:

1. **Its only consumer is the test suite.** Nothing else in the repo sets it - no
   script, no VS Code task, no launcher. A production code branch that exists
   only for test convenience is a smell.
2. **It is live on any non-production deployment**, including Cube Cloud branch
   staging environments, because `NODE_ENV` is not `production` there. That is
   the sole reason `docs/guides/cube.md` has to carry a standing "never set this
   in Cube Cloud" warning. Removing the branch makes the warning unnecessary
   rather than load-bearing.
3. **It cannot do what a developer would reach for it to do.** It supplies
   `groups` only. `region_key`, `allowed_abbreviations`,
   `allowed_department_groups`, and `reportee_staff_keys` all come back
   null/empty from `buildSecurityContext(null, [])`, and those are exactly the
   values the per-view `row_level` filters interpolate. So it can answer "does
   this group reach this view" but never "does this viewer see only their own
   region."
4. **It routes around a deliberate safety guard.** `access.buildGroups` refuses
   to emit a `staff-pii-*` group when the remit arrays are empty, because Cube
   throws a hard `Values required for filter` error on an empty array instead of
   cleanly denying. The bypass replaces `groups` directly and never calls
   `buildGroups`, so putting `staff-pii-all_in_scope` in the map reproduces the
   exact failure the guard was written to prevent - a state unreachable in
   production.

## Current state

Measured on the #4526 branch:

- 12 tests in `src/cube/cube.test.js` set `CUBE_GROUP_MAP` to stub identity.
[...]

Issue #4605, opened 2026-07-29
chore(cube): upgrade Cube to 1.7.x and evaluate BigQuery client v8


cristinabaldor:
## Summary

`src/cube` is behind on its runtime and driver packages. Splitting this out of
#4526 deliberately, because that branch's findings are version-specific and it is
building the test coverage this upgrade wants.

## Current state

Measured 2026-07-28 via `npm outdated` in `src/cube`:

| Package | Installed | Available | Kind |
| --- | --- | --- | --- |
| `@cubejs-backend/server` | 1.6.59 | 1.7.14 | minor |
| `@cubejs-backend/bigquery-driver` | 1.6.59 | 1.7.14 | minor |
| `@google-cloud/bigquery` | 7.9.4 | 8.3.1 | **major** |

We are on 1.6.59 only because `src/cube/package-lock.json` is committed and pins
it. `package.json` declares `^1.6.44`, which already permits 1.7.14 — so an
unpinned install would take it.

## Why this is not part of #4526

#4526's evidence base was all measured on Cube 1.6.59, and each item is exactly
the kind of thing a Cube minor release can change:

- `checkAuth` runs in Cube developer mode (contradicting the previous docs).
- `checkAuth`'s `maxAge: '12h'` cap rejects a stale cached Playground token.
- Cube Cloud bypasses `checkAuth` and injects its own unenriched security
  context, with `iss: 'cubecloud'` and no `iat`.
- Tesseract is the planner on both APIs and joining views is supported.

Bumping inside that PR would make it impossible to attribute a behavior change to
our code versus the upgrade, and would invalidate documentation corrections before
they merge.

## Proposed split

Two PRs, not one.

### 1. Cube 1.6.59 to 1.7.14 (minor)

Re-verify each bullet above after the bump, since they are now written down as
fact in `src/cube/CLAUDE.md` and `docs/guides/cube.md`:

- [ ] `checkAuth` still runs in developer mode; the REST Playground still
      resolves a pasted email security context.
- [ ] `maxAge` behavior unchanged (three tests in `src/cube/cube.test.js` assert
      it).
- [ ] Cube Cloud's injected security-context shape unchanged. **If 1.7 changes
      this, it may affect #4526's Cube Cloud enrichment design** — check before
      that work is built.
- [ ] Tesseract planning and multi-fact view joins still work.
- [ ] `src/cube` unit suite passes (`npm test`).
[...]

PR #4614, opened 2026-07-29
feat(cube): admin-gated internal user emulation, and close a Cube Cloud paste escalation

### LINKED ISSUE #4526: feat(cube): internal user emulation for RLS validation
## Problem

The custom `checkAuth` / `checkSqlAuth` in `src/cube/cube.js` derive a viewer's
entire security context from their resolved email (`resolveAccess`), ignoring any
context the caller supplies. That keeps production RLS tamper-proof, but leaves
**no sanctioned way to emulate another user** for validation:

- **Cube Cloud Playground / Explore** reach the deployment with no resolvable
  `email`, so `resolveAccess` returns the empty default-deny context — every view
  is hidden (`/meta` empty, so only source tables show) and queries compile to
  `WHERE (1 = 0)`.
- **Local Playground (REST)** skips `checkAuth` in dev mode, default-denying every
  gated view.
- **Local SQL API** works (connect as the viewer email) and is the current ground
  truth, but there is no cloud equivalent.

This blocks pre-pilot RLS validation ahead of an internal-customer pilot.

## Direction (design spec to follow)

Admin-gated impersonation, honored on any deployment: a caller whose own resolved
identity is an approved impersonator may pass an `act_as` email; `checkAuth`
resolves and returns that target's real internal context. Non-impersonators are
ignored (own scope only). Staged:

1. Make `resolveAccess` fall back to ADC when the driver credentials variable is
   unset (currently `JSON.parse("")` throws and fail-closes to deny-all locally).
2. Document the working local SQL API validation matrix.
3. Verification spike on Cube Cloud (does a Playground security-context paste of
   `{email}` resolve). If yes, cloud emulation needs no new code.
4. Build admin-gated `act_as` only if the spike shows a gap.

Kept compatible with the external-vendor API-key layer (#4501 / #4455): shared
`checkAuth` dispatch, strictly separate group namespaces (internal vs
`external-*`).

Spec to follow under `docs/superpowers/specs/`.

Refs #4455, #4501.

cristinabaldor:
# Pull Request

## Summary & Motivation

When merged, this pull request will give the data team a sanctioned way to see
exactly what an internal user sees in Cube, so a pilot user's row-level security can
be signed off before they get access. It also **closes a privilege-escalation hole
in the Cube Cloud path that predates this work and is live on production** — please
read that section first.

### Read this first: the security fix

`c2212537a` fixes a vector that has nothing to do with the feature and everything to
do with how Cube Cloud behaves.

Cube Cloud **merges a pasted Security Context into the top level** of the object
`contextToGroups` receives. The previous implementation read straight off that
merged object:

```js
contextToGroups: async ({ securityContext }) => securityContext?.groups ?? [],
```

So any Cube Cloud console user could paste this into the Security Context editor and
have it honored verbatim:

```json
{
  "groups": ["staff-pii-all_in_scope"],
  "allowed_abbreviations": ["..."],
  "allowed_department_groups": ["..."]
}
```

That grants the six gated sensitive staff fields with a forged remit, regardless of
the caller's HR-derived scope. Reproduced against the pre-fix code (pasted groups
returned as-is, forged `region_key` intact) and against the fix (groups empty,
`region_key` null).

Why it was never noticed: the July investigation only ever pasted `{email}`, which
yields no groups and looks exactly like default-deny, so the vector sat behind a case
nobody probed.

**Blast radius:** anyone with Cube Cloud console access able to edit a security
context. Believed to be the data team only, worth confirming. Production runs the old
implementation today, so it stays exposed until this merges.

**The fix:** always re-derive the context from the identity Cube Cloud authenticated
and overwrite, never conditionally on whether `groups` is present. `resolveAccess`'s
output covers every key any policy interpolates, so assignment over the top
neutralizes the paste. A `!securityContext.groups` guard there IS the bypass; the
code comment says so, because an earlier commit in this branch had exactly that
guard.

### Trust-model change needing explicit sign-off

Cube Cloud bypasses `checkAuth` entirely, so the enrichment identifies the caller
from `cubeCloud.username` — **asserted by Cube Cloud, not verified by our own
`jwt.verify`**.

Before this change, console identity was not load-bearing for data access: console
users resolved to nothing and saw only source tables. After it, Cube Cloud's console
authentication establishes identity for RLS purposes. That is deliberate and on par
with the trust the SQL API already places in the connecting user, but it is a real
extension of the trust boundary and should be an explicit decision rather than an
implicit one.

### What the feature does

- **Explore now works for real users.** `contextToGroups` resolves
  `cubeCloud.username` against `dim_staff_cube_access` and populates the groups and
  `row_level` values the policies read. No admin gate and no escalation surface, since
  each console user resolves to their own HR-derived scope.
- **Emulation on two surfaces, one rule.** REST/MCP takes an `act_as` claim; Cube
  Cloud takes a pasted `{"email": "..."}`. `access.resolveEmulationTarget` decides
  from the caller's own identity only, so a non-impersonator's target is ignored and
  they keep their own scope. Impersonators come from the `CUBE_IMPERSONATORS`
  deployment variable; unset means emulation is inert, which is the production
  default.
- **Audit trail.** Every real emulation emits one structured `cube_emulation` line
  carrying the two identities and a timestamp, never row data. An ignored `act_as`
  emits nothing, since a line there would implicate a caller who gained nothing.
- **A committed validation tool.** `scripts/cube_rls_matrix.py` opens one SQL
  connection per viewer and runs the same query, so a scope difference is
  attributable to policy alone. Viewer emails are passed in, never hardcoded.
- **403s say which check failed.** `maxAge` stays at 12h so PR #4269's property is
  intact, but a rejection now distinguishes too-old-from-`iat` (with the re-mint
  step), expired-past-`exp`, a bad signature pointing at the deployment's signing
  secret, and a missing `iat`. A single "Invalid token" for all four is what made the
  `maxAge` cap read as an access bug, and it cost an hour of this work to diagnose on
  a branch environment.
- **A local-dev fix.** `resolveAccess` falls back to ADC when the driver credentials
  variable is unset, instead of parsing an empty string, throwing, and fail-closing
  every viewer.

### Verified live, not just unit-tested

| Check | Result |
| --- | --- |
| SQL API, 5 viewer scopes | network sees 4 regions; the region viewer's count exactly equals the network viewer's slice for that region; school viewer a subset; none-scope and unresolved both 0 rows |
| REST `act_as`, local | impersonator resolves the target; a region viewer passing `act_as` for a network viewer keeps their own region; none-scope keeps 0 rows |
| REST `act_as`, Cube Cloud branch deployment | same four outcomes against production data |
| Cube Cloud Explore | emulating a region-scoped viewer returns that region only, with one audit line |
| Paste vector | reproduced before the fix, closed after |

75 unit tests pass. The load-bearing ones are the negatives: a non-impersonator's
target is ignored on both surfaces, pasted groups and scope values are overwritten,
and no audit line is emitted when emulation did not happen.

### Run the RLS sign-off with auth on

An earlier revision of this description said local dev was pinned to Cube **1.6.59**
against Cube Cloud's **1.7.14**, and blamed that version gap for an enforcement
difference: an out-of-tier member request hard-errors on Cube Cloud but silently
returns 0 rows locally. Both halves of that are now superseded by commits on this
branch.

Local dev now runs **1.7.14** too (`5186ec27f`), matching production. And the
enforcement difference was never about the version — measured across all four
combinations of {1.6.59, 1.7.14} × {developer mode, auth on}, the denial shape
tracks the **mode** and is identical across versions (`097790922`). A developer-mode
server downgrades an out-of-tier denial to a quiet 0 rows; with auth on it hard-fails
on both surfaces.

So the pre-pilot RLS matrix must be run against a server started with auth on:

```bash
cd src/cube && NODE_ENV=production CUBEJS_DEV_MODE=false npm run dev
```

Scoped viewers return identical row counts either way, so a developer-mode run
remains authoritative for **which rows** a viewer can reach — only the shape of a
denial needs auth on. The guide, `src/cube/CLAUDE.md`, and the matrix tool's own
docstring are updated accordingly. #4605 stays open for the remaining
`@google-cloud/bigquery` 7 to 8 major.

## AI Assistance

Written with Claude Code across two sessions. Design decisions were human-directed by
@cristinabaldor: the admin-gated trust model, keeping internal emulation separate
from the external-vendor API-key layer, documenting local placeholder config rather
than loosening the secrets hooks, and the scope of the impersonator list.

Claude grounded the work in the actual code and warehouse rather than the design
notes, which mattered — the July-era observation of Cube Cloud's injected context was
wrong for 1.7.14, and building against it would have shipped an enrichment that
silently no-ops. Two Claude-authored missteps are in the history and worth a
reviewer's attention: a first diagnostic probe granted a group and logged an unknown
object to deployment logs, both flagged by an automated security review and reverted
in `e4d044719`; and an intermediate enrichment carried the `!groups` guard that
turned out to be the bypass, fixed in `c2212537a`.

### General

### Dagster

N/A — no Dagster changes.

### dbt

N/A — no dbt model changes. `dim_staff_cube_access` and `dim_staff_reporting_chain`
are read as-is.

### Docs

N/A for the automations catalog — no schedule or sensor changes. `src/cube/CLAUDE.md`
and `docs/guides/cube.md` are updated here.

## Reviewer notes

Deployment configuration is **not** in this PR, because `CUBE_IMPERSONATORS` is read
from the environment per request and nothing needs committing to enable it:

- Setting it on a Cube Cloud deployment enables emulation for the listed emails.
  Leaving it unset means emulation is inert and only self-resolution applies.
- It is a control on Cube Cloud, where the config is not the user's to edit. It is
  **not** a control locally, where the dotenv file belongs to the developer — and
  that is fine, since running the local server already requires ADC access to
  `kipptaf_marts`.
- Suggested selection rule, documented in the guide: prefer callers whose own scope
  already covers anything they could emulate, so emulation is a viewport change
  rather than a grant.

Follow-ups deliberately excluded, both gated on this merging because this branch is
what makes them verifiable:

- #4605 — Cube 1.6.59 to 1.7.14, and the BigQuery client major separately
- #4607 — remove the `CUBE_GROUP_MAP` dev bypass and stub BigQuery in its tests

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — not triggered (no dbt model changes).
- [ ] **Dagster Cloud** — not triggered (no Dagster paths changed).

Closes #4526


cristinabaldor:
All four findings verified against the code before fixing, and all four were real
— no pushback on any of them. Fixed in `aad3021ad`, pushed with a `main` merge as
`dd4731894`.

## 1. PII in a committed file — redacted, but needs a human decision

Confirmed and it was mine: introduced in `73697865d`, copied out of session notes
into the plan's out-of-scope list. Redacted to a query that finds the same rows
without naming anyone.

**@cbini flagging this for you — the review understated it, and redaction does not
retract it.**

The review said it "ships to `main` and GitHub history permanently." Two
corrections:

- **It will not reach `main`.** The repo squash-merges, so `main` only ever gets
  the final, redacted tree.
- **But it is already public and stays that way.** This repo is public and the
  commit is pushed. After merge it remains reachable three ways: the merged PR's
  own commit list (a normal browsing path, not an obscure one), direct fetch by
  SHA, and the fork network's shared object store — there is 1 fork.
  `delete_branch_on_merge` removes the ref, not the object, and there is no
  guarantee GitHub ever garbage-collects it.

So "squash merge keeps `main` clean" and "the address is gone" are different
claims, and only the first is true.

Options, in case you want a different call than the one below:

- **Rewrite history and force-push before merge** — purges it from the branch, but
  GitHub retains force-pushed commits in the PR timeline, so the old SHA may stay
  reachable anyway. Also rewrites every SHA after `73697865d`, churning the commit
  list mid-review.
- **File a GitHub Support request** — the only path that actually purges objects
  and PR references from a public repo.
- **Accept it** — one work address in a guessable `firstinitial+lastname` pattern,
  the kind often already in public staff directories.

**Decision for now: accepting, not blocking the merge.** @cbini please confirm, or
say if you want the Support request filed — happy to do that in parallel since it
does not gate anything.

## 2. Dead export with silent drift — fixed

Confirmed both halves. `git grep -n emulationInputsFromCubeCloud -- src/cube`
returned zero production call sites, and the drift was real: the inline version in
`contextToGroups` had grown a `cubeCloud.userAttributes.email` fallback the
exported helper lacked, so the helper's unit tests were exercising a shadow
implementation.

Took the first suggested fix rather than deleting the export: the fallback moved
into `emulationInputsFromCubeCloud`, and `cube.js` now spreads it, mirroring how
`checkAuth` already does `...access.emulationInputsFromToken(payload)`. Added two
cases pinning the new behavior — the `userAttributes.email` fallback, and
top-level `email` winning over the mirror when both are present. 75 to 77 tests.

On the related recommendation to move those tests to `cube.test.js`: not needed
now. Because `cube.js` calls the helper, `access.test.js`'s cases exercise the
running extraction path, and `contextToGroups`-level coverage already exists
separately (including the `userAttributes`-only case, which is what proved the
fallback moved correctly rather than being duplicated).

## 3. Guide contradicting itself — fixed

Confirmed: the surface table said Cube Cloud emulation was "not possible yet"
while three later sections of the same file explained how to do it. Row updated,
and the paragraph above it no longer says "both local surfaces" now that Cube Cloud
is covered too.

## 4. Duplicate paragraph — removed

Confirmed, and the diagnosis was right about the cause.

---

One process note: `claude-review` only triggers on `opened` / `ready_for_review`,
so this push will not get a re-review. Each claim was checked against the code
before being fixed rather than after, and the verification is in `aad3021ad`'s
message.

cristinabaldor:
@claude update your review

cristinabaldor:
## Heads up: this collides with `kverhoff/feat/cube-individual-access-layer`

@kverhoff — flagging before either of these merges, since we're both editing
`src/cube/access.js` and `src/cube/cube.js` right now. The short version: **the
conflicts are mechanical and the designs are compatible**, but there's one shared
invariant worth knowing about, and one thing on your side I'd double-check.

### The conflict

`git merge-tree` between this branch and yours fails on two files:

- `src/cube/access.js` — we both append new functions immediately before
  `module.exports` and both add entries to the export list, so it collides
  positionally
- `src/cube/CLAUDE.md` — we both rewrote the security-model section

Tedious to resolve, but nothing semantic. Your branch does **not** touch
`contextToGroups`, `resolveAccess`'s control flow, or the paste fix, so there's no
collision on the security-critical path this PR changes.

This PR is open and green, so it'll likely land first and the resolution falls to
you. Happy to take that on instead if you'd rather — say the word.

### The designs are compatible, and here is why

Your `unionAdditionalGrants` is applied **inside `resolveAccess`**, before
`buildSecurityContext`, so the unioned grants are baked into the context that
function returns.

This PR's Cube Cloud fix does:

```js
Object.assign(securityContext, await resolveAccess(target));
```

Because your grants are already inside `resolveAccess`'s output, that assignment
carries them. No interaction, nothing dropped. I checked this specifically because
the fix overwrites `allowed_abbreviations` wholesale, and if grants had been applied
anywhere downstream of `resolveAccess` this PR would have silently discarded them.
They aren't, so it's fine.

### The invariant we now both depend on

The fix neutralizes a pasted Cube Cloud security context by assigning
`resolveAccess`'s output over the top. That is a **complete** overwrite only
because `buildSecurityContext` returns every field any policy interpolates.

Your branch is the first change to test that: it adds
`allowed_student_abbreviations`, a new policy-read field — and it satisfies the
invariant, because you put it in `buildSecurityContext`'s return. That's the right
place, but it held by good design rather than by anything enforcing it.

A policy-read field added anywhere else (computed in a hook, spread in from
elsewhere) would not be overwritten, reopening the paste vector for that one field,
silently, and only on the Cube Cloud surface. I've written the rule into
`src/cube/CLAUDE.md` in this PR (`64a1f6144`) so it's stated rather than implied.

Nothing for you to change — just worth knowing the constraint exists now, since
you're the one adding fields to that return value.

### One thing to double-check on your side

Your `buildSecurityContext` **drops `student_location_scope`** from the returned
context. I went looking and it appears safe: no view `access_policy` interpolates
it, and nothing under `src/cube` reads it off the context. I assume that's
deliberate, since with individual grants a single scope enum no longer describes
student access.

Worth confirming against anything **outside** `src/cube` that consumes the context
shape, if there is any — that's the part I couldn't rule out from here.

### Merge order is flexible

Your new parameters all default (`allowedStudentAbbreviations = []`, and
`buildGroups` defaults all four), so this PR's existing three-argument
`buildGroups(SL, ["ABC"], ["Ops"])` test calls keep working after your change.
Either order works; this one first is simplest only because it's already green.

Also relevant if you're validating locally: local dev is pinned to Cube 1.6.59 and
Cube Cloud runs 1.7.14, and at least one enforcement behavior differs between them
(an out-of-tier member request hard-errors on 1.7.14, silently returns zero rows
locally). Tracked in #4605. This PR adds `scripts/cube_rls_matrix.py`, which may be
useful for checking your grant logic per-viewer once it lands.

cristinabaldor:
@claude update your review

cristinabaldor:
Both review findings addressed in `fcff289d2` plus a description edit.

**1. Stale provenance mentions** — citations were exact. `docs/guides/cube.md:232` and `src/cube/CLAUDE.md:241` each said "verified on Cube 1.6.59" while the *other* mention in the same file had been updated to cover both versions. Fixed both to read 1.6.59 and 1.7.14.

**2. Stale "Known limitation" in the description** — replaced with "Run the RLS sign-off with auth on." It now says plainly what the earlier revision got wrong (the enforcement difference was attributed to the version gap, and is actually a mode difference), and carries the auth-on startup command. Verified the stored body round-tripped without entity double-encoding.

On the passing note about `CUBE_SQL_DEV_EMAIL` staying inert once production mode is set: that holds, and it can be confirmed empirically rather than by reading the diff. The matrix run against an auth-on server returned three **distinct** per-viewer scopes (one school, one region, one network). Had that fallback been pinning every connection to a single identity, all viewers would have returned identical rows — which is the failure mode the tool explicitly warns about.

Worth stating the verification boundary plainly, since the review noted it too: with no `node_modules` or network in the review environment, the 77-test result and the four-way denial-shape measurement were taken from the commit message rather than reproduced. Anyone wanting to re-run the measurement needs a local server started both ways against each version; the commands are in the guide.

Neither outstanding item is touched by these commits: the historical-exposure question awaiting @cbini, and the `kverhoff/feat/cube-individual-access-layer` conflict. Note these commits do edit `src/cube/CLAUDE.md`, so they add slightly to that mechanical conflict surface.

cbini (approved):

Issue #4535, opened 2026-07-24
fix(cube): bind student RLS to employing legal entity, not physical work location


cristinabaldor:
## Problem

Cube student row-level security scopes a viewer by `region_key`, which `dim_staff_cube_access` derives from the staff member's **physical work location** (`dim_work_assignment_locations` -> `dim_locations.region_key`). FERPA binds student-record access to the **legal entity (LEA/charter) that employs the staff member**, not where their desk sits. These coincide today only because `region` <-> `legal_entity` is 1:1 -- the binding is on the wrong attribute.

## Impact (current data)

Staff whose physical region != employing legal entity:

- **KTAF (CMO) employees mis-pinned to a desk city instead of network:** 12 on `region` scope (11 Newark, 1 Camden) + 5 on single-`school` scope -> capped to one city's/school's students. Should be network (KTAF's own entity enrolls no students).
- **KIPP Miami employees whose desk doesn't resolve to a region** (~10): null key -> `region` scope returns **zero** students (wrongful denial).
- **Latent cross-charter leak:** staff sitting in one region but employed by another LEA's charter (3 Paterson-desk / TEAM-employed today, all currently `none`) -- one `region` grant away from exposing another LEA's students.

All 162 KTAF employees resolve to one of three per-city office rooms (Room 9/10/11), each with **0 enrolled students** -- physical location is only a desk city for them. (`is_campus = true` on those rooms is an unreliable school signal.)

## Root cause

1. `dim_staff_cube_access.region_key` is location-derived, not employer-derived.
2. `entity` collapses employer `business_unit_name` to KTAF/Region/unknown; the specific employing legal entity isn't carried as the scope key.

## Proposed direction (to refine in brainstorming)

1. Derive the scope key from employing `business_unit_name` (legal entity), not physical location.
2. `entity = KTAF` implies network scope, independent of desk.
3. Rename region -> legal entity through the RLS chain where it's a security boundary; keep `region` as a reporting dimension.

## Affected surface

- dbt: `dim_staff_cube_access` (key derivation)
- `src/cube/access.js` (`buildSecurityContext`, `computeAllowedAbbreviations`)
- Student views' `access_policy` (`student_enrollments`, `student_attendance`, `student_assessment_scores`, `student_section_enrollments`)
- Scope-enum naming in the `cube_access_role` / `cube_access_department_override` Google Sheets

## Open questions

- Keep enum value `region`, or rename to `legal_entity`/`entity`?
[...]

Issue #4466, opened 2026-07-20
fix(cube): resolveAccess BigQuery client bills jobs to cubejs-cloud, denying every viewer


cristinabaldor:
## Symptom

After the #4269 security-redesign deploy, **every** Cube viewer is denied (all APIs: MCP/REST and SQL/Superset/Tableau). `meta` returns `{"cubes":[]}` even for a network-scope viewer with correct access data.

## Root cause

`resolveAccess` in `src/cube/cube.js` constructs its own `@google-cloud/bigquery` client with a bare `new BigQuery()` and no explicit `projectId`/`credentials`. In the Cube Cloud runtime that client defaults to the host project `cubejs-cloud`, so every identity read fails:

```text
resolveAccess failed for <viewer>: ApiError: Access Denied: Project cubejs-cloud:
User does not have bigquery.jobs.create permission in project cubejs-cloud.
```

The read is wrapped in a try/catch that fails closed to an empty (default-deny) `securityContext`, so the whole network is denied. Cube's own data driver is unaffected because it is configured via the `CUBEJS_DB_BQ_PROJECT_ID` / `CUBEJS_DB_BQ_CREDENTIALS` variables — which the hand-rolled client inherits none of.

Confirmed the data path is healthy: `dim_staff_cube_access` (1,607 rows), `dim_staff_reporting_chain` (9,278), and `dim_locations` (39) are all deployed and populated, and a network viewer's access row is correct. The fault is purely the misconfigured client.

## Fix

Pass `projectId` (from `CUBEJS_DB_BQ_PROJECT_ID`) and `credentials` (base64-decoded from `CUBEJS_DB_BQ_CREDENTIALS`) explicitly to the `new BigQuery({ ... })` constructor, matching Cube's driver config.

Note: an IAM grant alone would not fix it — the table refs are unqualified (`kipptaf_marts.<table>`), so they must resolve under `teamster-332318`.

Prod only redeploys on merge to `main`, so this ships via a squash-merged PR.

PR #4615, opened 2026-07-29
Redesign Cube individual access exceptions with additive grants

kevinverhoff:
# Pull Request

## Summary & Motivation

...redesign the `cube_access_individual_exceptions` Google Sheet and rework
`dim_staff_cube_access` so individual Cube access exceptions actually take
effect — the existing mechanism silently did nothing (the sheet was joined
but never read). The new design is additive: an exception grants one extra
network/region/school on top of a person's normal role-based access, rather
than replacing it, and an employee can have multiple grant rows (e.g. two
extra schools). Location grants and the five sensitive-field remit overrides
(PII, compensation, observations, benefits, department) are gated on
`status`/`grant_date`/`expiry_date`, so revoked/expired/not-yet-effective
rows apply nothing.

This required moving Cube's student-facing row-level security from three
fixed tiers (network/region/school) to a single array-based remit
(`allowed_student_abbreviations`), mirroring the pattern `staff_pii` already
used — a tier can only express "my whole region," not "my region plus one
specific extra school." All 4 student views, `access.js`, and `cube.js`
changed accordingly.

Along the way, found and fixed two pre-existing bugs in `dim_staff_cube_access`
that predated this PR and had never been build-tested (a stale
`dim_staff.employee_number` reference — `dim_staff` had since renamed it to
`staff_unique_id` — and a CTE alias mismatch in the remit coalesce). Neither
was a live production issue: prod's deployed `dim_staff_cube_access` is an
older, simpler version that never attempted this feature at all.

Design doc: `docs/superpowers/plans/2026-07-29-cube-access-individual-exceptions-redesign.md`
(trade-offs, full column reference, worked examples). Ops-facing guide:
`docs/guides/cube-access-individual-exceptions.md`.

**Breaking change to the sheet schema** — `student_location_scope` and
`staff_location_scope` columns are removed from `cube_access_individual_exceptions`;
replaced by `additional_location_type`/`additional_location_name`/`include_student_data`
plus an approval-workflow audit trail. Sheet already migrated to the new
schema and existing test rows validated against a live local build.

## Contractors, and re-keying the sheet on `google_email` (added after the first reviews)

A contract employee holds a KIPP Google login and no ADP employment record, and they need to be scoped like anyone else. The sheet was keyed on `employee_number`, so there was no number to key them on and no way to grant them anything. **`employee_number` is replaced by `google_email` as the sheet's key.**

Re-keying alone does not fix it, which was the non-obvious part. `dim_staff_cube_access` spines on an INNER JOIN to `dim_work_assignment_primary`, so a non-employee produces no row at all, and the exceptions sheet is LEFT JOINed onto that spine — a contractor's grant row would attach to nothing and silently grant nothing. That is the same class of no-op this PR exists to fix. So the mart now unions a second leg:

- `non_employee_grantees` emits a row for every live grant whose address matches no staff member, gated on an INNER JOIN to `stg_google_directory__users` that is neither suspended nor archived. That join is the authorization check — a typo or a deleted account cannot mint an identity. A suspended account keeps its directory row, so it passes the sheet's FK test and is excluded here instead.
- `access_spine` unions the staff leg and the non-employee leg so scope resolution is written once. The non-employee leg carries NULL for every role and org attribute, so all five scopes coalesce to `none` and the person sees only what their own exception rows grant.
- `staff_key` for those rows is `generate_surrogate_key(['non_employee', google_email])` — the literal keeps the two keyspaces from colliding and marks the row's origin. It matches nothing in `dim_staff_reporting_chain`, so reporting-chain scopes resolve empty, which is correct: they manage no one.

Staging lower-cases and trims the address, because Cube's identity read is a case-sensitive `google_email = @email` match against the JWT claim. A hand-typed `Jane@Apps.Teamschools.Org` would otherwise resolve to nothing and deny without saying why.

Four new tests in `access.test.js` pin the contractor shape end to end: a `none` base scope resolving to no abbreviations on either axis, one school grant with student data yielding exactly that school, a grant without student data yielding no student access, and a grantee with no live grants getting the open directory only.

A Google Form request-and-approval workflow that writes this sheet is deliberately **not** in this PR — it will be a follow-up. Dagster already ingests Google Forms responses (dynamic-partitioned by `form_id`, with a polling sensor), and the sheet already carries `requested_by`/`approved_by`/`status`, so it has somewhere to land.

## Merged `origin/main` (six weeks of drift)

Five conflicts, all from main moving under the branch:

- `student_enrollments_view.yml` was deleted on main (enrollment and attendance unified onto one student-day surface). Took the delete, and ported this branch's flat `student` access policy onto `student_attendance_enrollment_periods_view.yml`, which main added after the branch's last push and so still carried the retired three-tier groups. Left alone it would have default-denied every viewer, since `buildGroups` no longer emits `student-region`/`-school`/`-network`.
- `src/cube/CLAUDE.md` — main moved the authoring conventions to `.claude/rules/cube-authoring.md`. Took main's file and ported this branch's two edits into the rules file.
- `dim_staff_cube_access.sql` — main dropped the SQL header docstring as duplicated in YAML. Took the delete.
- `macros/utils.sql` and `tests/properties.yml` — both sides appended; kept both.

Also corrected two published pages: `docs/guides/cube.md`'s group-taxonomy table still documented the retired three-tier student groups (**wrong before this PR**, not a merge artifact), and the model description still said `contextToGroups` builds `queryRewrite` filters, which main retired.

## AI Assistance

Claude Code planned and implemented this end-to-end (schema design, dbt
mart rework, Cube `access.js`/`cube.js`/view changes, docs) in collaboration
with the branch author, who made the key design calls (additive vs. override,
multi-row support, removing the `'All'` sentinel, status/date enforcement) and
drove local validation (`dbt build`, Cube dev server RLS checks against real
staff records).

### General

### dbt

- [x] New/changed models have properties YAML
      (`stg_google_sheets__people__cube_access_individual_exceptions.yml`,
      `dim_staff_cube_access.yml`)
- [x] `dim_staff_cube_access` already has its exposure via `cube.yml`'s
      `cube_semantic_layer.depends_on` (unchanged, verified)
- [x] Breaking change to the individual-exceptions sheet schema — see summary
      above; sheet already migrated, no dashboard/report reads it directly
      (only `dim_staff_cube_access` does)
- [ ] **External source needs re-staging with `--target staging`** — the source's
      column list changed again (`employee_number` → `google_email`), so the
      earlier staging run is stale:
      `google_sheets.src_google_sheets__people__cube_access_individual_exceptions`.
      The sheet's first column must be renamed to the Google address before
      staging, since an explicit-schema Sheets external maps columns by
      position.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes


kevinverhoff:
@claude review again plz

kevinverhoff:
@claude review again por favor

kevinverhoff:
@claude do your thing

kevinverhoff:
@claude one more code review plz

kevinverhoff:
Re: the Minor finding in the latest review (re-review #3) suggesting `is_live_row()` replace the hand-spelled `where` predicate in `stg_google_sheets__people__cube_access_individual_exceptions.yml` (and the two `relationships` tests) — I tried exactly that, including the literal syntax proposed (`where: "{{ is_live_row('status', 'grant_date', 'expiry_date') }}"`), and it doesn't compile:

```text
Compilation Error
  Invalid generic test configuration given in models/google/sheets/staging/properties/stg_google_sheets__people__cube_access_individual_exceptions.yml:
  The stg_google_sheets__people__cube_access_individual_exceptions.None column's "unique_combination_of_columns" test references an undefined macro in its where configuration argument. The macro 'is_live_row' is undefined.
  Please note that the generic test configuration parser currently does not support using custom macros to populate configuration values
```

This is a dbt limitation, not a syntax slip: generic-test `config.where` is parsed before project macros are in scope, so no custom macro — `is_live_row()` included — can be called there. That's why the two pre-existing `relationships` tests in the same file already hand-spell the identical predicate instead of calling the macro; it's forced, not a style choice that was merely left inconsistent. All three occurrences in this file are already the only form that compiles, so there's nothing to consolidate.

Leaving this here so the next re-review pass doesn't re-flag it.

cristinabaldor:
## Ready for review — contractor access, a sheet schema realignment, and an RLS sign-off

Picking this back up after six weeks. Three review rounds had converged on "ready to merge", but the branch had gone stale and a new requirement landed: **a contract employee with a KIPP Google login and no ADP employment record needs to be scoped like anyone else.** That turned out to touch the mart's spine, not just the sheet, so the change is larger than a re-key.

Seven commits, `40090c35d` through `1a6e67061`. All CI green, dbt Cloud included.

The sheet's external source is staged and current: its 16 columns match the source declaration exactly, and dbt Cloud passing at `1a6e67061` confirms it — a stale external fails this model with `Unrecognized name`, which is how the column rename surfaced earlier on this branch.

### 1. Merged `origin/main` (`40090c35d`)

Five conflicts, all from main moving underneath this branch.

The one that mattered: main deleted `student_enrollments_view.yml` (enrollment and attendance were unified onto one student-day surface) and added `student_attendance_enrollment_periods_view.yml` **after** this branch's last push. That new view still carried the retired three-tier `student-region`/`-school`/`-network` groups. Left alone it would have default-denied every viewer on it, because `access.buildGroups` no longer emits those groups — a silent, total denial on a brand-new view. Ported the flat `student` policy onto it.

The rest: main moved Cube authoring conventions to `.claude/rules/cube-authoring.md`, and dropped the SQL header docstring from `dim_staff_cube_access` as duplicated in YAML. Took main's side on both and ported this branch's edits into the new homes.

Also corrected the model description, which still said `contextToGroups` builds `queryRewrite` filters. Main retired `queryRewrite`; no reference to it survives in the JS.

### 2. Contractors, and why re-keying alone was not enough (`6f84ac8a1`)

The sheet was keyed on `employee_number`, so it could only ever reach someone who already had an ADP record. **Re-keying to `google_email` does not fix that by itself** — and this was the non-obvious part.

`dim_staff_cube_access` spines on an `INNER JOIN` to `dim_work_assignment_primary`, so a non-employee produces no row at all, and the exceptions sheet is `LEFT JOIN`ed onto that spine. A contractor's grant row would attach to nothing and silently grant nothing — the exact class of no-op this PR exists to fix.

So the mart now unions a second leg:

- `non_employee_grantees` emits a row for every live grant whose address matches no staff member, gated on an `INNER JOIN` to `stg_google_directory__users` that is neither suspended nor archived. That join is the authorization check: a typo or a deleted account cannot mint an identity. A *suspended* account keeps its directory row, so it passes the sheet's foreign-key test and is excluded here instead.
- `access_spine` unions the staff leg and the non-employee leg so scope resolution below is written once. The non-employee leg carries NULL for every role and org attribute, so all five scopes coalesce to `none` and the person sees only what their own grant rows give them.
- `staff_key` for those rows is `generate_surrogate_key(['non_employee', google_email])`. The literal keeps the two keyspaces from colliding and marks the row's origin. It matches nothing in `dim_staff_reporting_chain`, so reporting-chain scopes resolve empty — correct, they manage no one.

Staging lower-cases and trims the address, because Cube's identity read is a case-sensitive `google_email = @email` match against the JWT claim. A hand-typed `Jane@Apps.Teamschools.Org` would otherwise resolve to nothing and deny without saying why.

### 3. Aligned the sheet's columns with the role sheet (`95d263071`)

The exceptions sheet had drifted from the vocabulary every other access sheet uses.

`additional_location_type` named the network/region/school axis "type", where `cube_access_role` and `cube_access_department_override` both call it `*_location_scope`. And `include_student_data` collapsed two independent axes into one boolean: the role sheet carries `student_location_scope` **and** `staff_location_scope` separately, so it can grant student access without staff access — this sheet could not.

Replaced with `additional_student_location_scope` and `additional_staff_location_scope`. A row still grants ONE location, so a test enforces that the two agree whenever both name a tier; different breadth per axis is two rows. Staging derives `additional_location_scope` as the first non-null of the two, so the grain test, both foreign-key checks, and the mart read one tier instead of re-deriving it three times.

`additional_location_grants` gains `includes_staff_data` beside `includes_student_data`, and `unionAdditionalGrants` takes `{ axis }` in place of `{ studentOnly }`, called once per axis from `resolveAccess`. An absent or unrecognized axis unions nothing and returns the base list, so a future miswiring fails closed rather than widening the wrong axis.

### 4. No blank cells anywhere (`4fd630df1`, `84a9800db`)

Driven by the follow-up that will write this sheet from a Google Form: a form cannot emit "blank means inherit". It also fixed a real hazard.

On the five remit-override columns, blank and `none` were **opposites** — blank fell through the coalesce to the person's role-derived setting, `none` won the coalesce and revoked below it. So typing `none` to be explicit did the opposite of nothing, and reading the sheet could not distinguish "I did not touch this" from "I meant to revoke this". Added `inherit` as the explicit leave-it-alone value; `none` keeps its role-sheet meaning and is now something you can only pick on purpose.

Every typed column is now required at error severity. `additional_location_name` carries the word matching the row's breadth — `all` when the scope is `network`, `none` on a remit-only row — with a test asserting the name and tier agree in all three directions. That catches a school name typed onto a `network` row, which reads as granting that one school while actually granting every location. Both dates are required, with `9999-12-31` for never, reusing the sentinel `dim_work_assignment_status` already uses.

An earlier commit on this branch removed `none` from the location axes on the reasoning that one word should not mean two things. That was the wrong fix — it left the remit columns, where the hazard actually lives, untouched while making the safe columns harder to fill. Reverted.

### 5. A guard for the gap the contractor leg opens (`35b3720a8`)

The non-employee leg requires a live Google account, which creates a failure mode nothing caught: a grantee who is both off the work-assignment spine **and** no longer has an active account resolves to no row, so their grant denies silently. The ordinary way in is a former employee kept on as a contractor — they leave the spine the day their `termination_date` passes, and IT suspends the account separately, often weeks later. In between the grant works; afterwards it stops, with nothing saying so. The `google_email` foreign-key test does not catch it, because a suspended account keeps its directory row.

`test_cube_access_individual_exceptions_grant_reaches_a_viewer` asserts the invariant rather than the causes: every live grant row resolves to a row in `dim_staff_cube_access`. Warn, not error — the state denies rather than over-grants, and it clears on its own once the account is reactivated.

### 6. Docs

Rewrote `docs/guides/cube-access-individual-exceptions.md` for the new key and vocabulary, and corrected two published pages that still documented the retired three-tier student groups. **`docs/guides/cube.md`'s group-taxonomy table was already wrong before this PR** — not a merge artifact.

### 7. Verification: 10 viewers, 3 passes, no failures

Built the chain into a dev schema (`PASS=50 ERROR=0`) and ran the committed RLS matrix against a local Cube server with **auth on** — one SQL API connection per viewer, same query, so any difference is access policy alone. Developer mode was avoided deliberately: it downgrades a real denial to a quiet 0 rows, which would have made the staff pass read as a clean pass while telling us nothing.

| case | result |
| --- | --- |
| contractor, off the spine, network grant | 4 regions / 28 schools — **identical output to two active employees with network grants** |
| cross-entity region grant | exactly 2 regions: base + granted |
| two school grants on one person | exactly 3 schools: base + both grants |
| school scope to network | 4 / 28 |
| region scope to network | 4 / 28 |
| base scope resolving empty, plus a grant | the granted school only |
| remit override, no location grant | **the only viewer able to query `staff_pii`** |
| expired row + revoked row | **nothing** — one school, where an ungranted peer in the same region sees five |
| two ungranted controls | unchanged |

The contractor result is the one worth dwelling on: a terminated employee with **zero rows** in prod's access mart resolved to a full viewer purely off the sheet, and produced byte-identical output to active employees holding the same grant. Downstream of the identity read, Cube cannot tell the two legs apart — which was the design goal. Their `staff_pii` denial is equally correct: the network grant widened their *location* reach on both axes but gave no staff PII, because that is a separate remit column left at `inherit`, falling through to a role they no longer have.

The expired-plus-revoked case is the sharpest negative: two dead rows on one person granted exactly nothing, confirmed against a peer with the same base scope and no grant.

**Axis independence, proven end to end in a follow-up run.** One viewer, one row: a `school` grant with the student axis set and the staff axis `none`, plus `staff_pii_scope = all_in_scope` so the staff side is observable at all. She resolves to **two** schools of student data (her own plus the granted one) and **one** school of staff PII (her own only) — the granted school's staff do not appear. The test is falsifiable by construction: the granted school has 41 active staff, so a leak across the axis would have moved the number and did not. An earlier attempt used a closed school with zero staff, where the negative would have held vacuously.

### Still open

- **Nothing in CI runs the Cube JS tests.** `deploy-cube-mcp.yaml` is the only workflow touching `src/cube`; it is push-triggered, path-scoped to `src/cube/mcp/**`, and builds a Docker image. So the 99 tests covering `access.js` — `buildGroups`, `unionAdditionalGrants`, the fail-closed axis gate — run only when someone runs them by hand, and a regression in the access-control decision layer would ship past every green check. Worth a `node --test` step on PRs touching `src/cube/**`; filing separately. The dbt half is fully covered by CI.
- `staff_compensation_scope`, `staff_observations_scope` and `staff_benefits_scope` remain wired but inert: no view exposes salary, observations or benefits, so a grant on those columns gates nothing today.
- Re-measured #4535 while validating this branch; its counts have drifted and the wrongful denial it predicted is now observable. Commenting there separately.

cristinabaldor:
@claude update your review

cristinabaldor:
## Review round 4 — both findings verified and fixed in `4d23ba543`

### Important 1 — axis-unaware guard test: **confirmed, fixed**

Correct, and it is a regression this PR introduced in `95d263071` rather than a pre-existing gap. Verified the mechanism in the code: `individual_exception_grants` filters on tier (`where iel.location_scope != 'none'`), not on axis, so a student-only grant lands in `additional_location_grants` and satisfies a bare `array_length(...) > 0`.

One correction to the framing — it is **more** reachable than the finding suggests. The non-employee leg sets `staff_location_scope` to `'none'` by construction, so a contractor granted a PII scope plus a student-only location is exactly the failing shape. The feature this PR adds is what makes the state ordinary rather than hypothetical. That is now in the comment above the test.

Took the suggested fix. Proved both directions against a synthetic row in that exact state (`staff_pii_scope = 'all_in_scope'`, `staff_location_scope = 'none'`, one grant with `includes_staff_data = false`):

| expression | flags the bad row? |
| --- | --- |
| old, `array_length(...) > 0` | **false** — passes silently |
| new, `countif(includes_staff_data) > 0` | **true** — caught |

Against real data the new expression returns 0 violations, so it does not false-fire, and the generated test compiles to valid BigQuery with its `unnest` subquery intact.

### Minor 2 — stale `student-<scope>` in `docs/guides/cube.md`: **confirmed, fixed**

Line 452 called `student-<scope>` the current taxonomy while the access-groups table in the same file already said flat `student`. Confirmed it is unchanged from `main`, so missed rather than deliberate. Now reads `student`.

### On the two items listed as not independently verified

Both were run locally, so recording the results here rather than leaving them open:

- **The 99 `src/cube` tests pass** — re-run after this branch merged `main` again in `6d977d4e1`, which bumped `src/cube/package.json` and the lockfile. 99 tests, 99 pass, 0 fail against the new dependency set.
- **The RLS matrix was run**, against a local Cube server with auth on, one SQL API connection per viewer. Results are in the previous comment. The axis-independence case in particular now has an end-to-end result: one viewer with a student-only grant resolves to 2 schools of student data and 1 school of staff PII, with the granted school's 41 staff correctly absent.

Worth noting for future rounds: nothing in CI runs those 99 tests. `deploy-cube-mcp.yaml` is the only workflow touching `src/cube`, and it is push-triggered, path-scoped to `src/cube/mcp/**`, and builds a Docker image. Filed as #5477.

### Not addressed

The three forward-compat remit columns (`staff_compensation_scope`, `staff_observations_scope`, `staff_benefits_scope`) stay wired but inert — no view exposes salary, observations or benefits, so a grant on those columns gates nothing. Unchanged by this PR and out of its scope.

cristinabaldor (dismissed):

.claude/settings.json
cbini: revert -- duplicate

src/dbt/kipptaf/macros/utils.sql
```diff
@@ -20,3 +20,15 @@
         0
     )
 {%- endmacro %}
+
+{% macro is_live_row(status_column, grant_date_column, expiry_date_column) %}
```
cbini: not sure this macro is justified

cbini (changes requested):
## Review at `ac5eab2`: changes requested

This review covers the full diff (`1fec14b..ac5eab2`). I found no way for someone to see student data beyond what a live sheet row grants. I checked NULL handling, email case, duplicate rows from multiple grants, surrogate-key collisions, and leftover `'All'` values. `node --test cube.test.js access.test.js` passes (102/102). The status/date gating in `is_live_row` is solid, and unresolved values deny access. The findings below are behavior changes and new access paths that need a fix or an explicit sign-off.

### Should fix

**1. Network-scoped viewers lose about 16% of student enrollment rows.** `student-network` used to have no row filter. Now `computeAllowedAbbreviations` (`src/cube/access.js:158`) returns every `dim_locations` abbreviation for network scope, and the views filter on it. Any row whose `location_key` doesn't resolve is now hidden. In prod, `dim_student_enrollments` has 21,564 of 133,662 rows with a NULL `location_key` (3,192 of 14,570 in AY2025). Region and school viewers never saw these rows, so only network viewers regress. The attendance and assessment views likely have the same gap; I didn't measure them. Please either confirm with the data owners that hiding these rows is intended and note it in the PR body, or fix the NULL keys upstream first.

**2. A former employee keeps their exception grants until their Google account is suspended.** `non_employee_grantees` (`dim_staff_cube_access.sql`, around lines 225-240) anti-joins to `enriched`, which only holds people with a current work assignment. Once someone leaves, their still-live grant rows qualify as a "non-employee" as long as the account is active. The model's own test description says suspension happens "days or weeks after" termination. Before this PR, grants applied only to people on the staff spine, so this is a new route to student data after offboarding. Suggested fix: anti-join to `dim_staff` on `google_email`, so the check covers anyone with a staff record, not only current assignments. Anyone who truly needs access after leaving can be re-entered on purpose.

**3. A contractor granted staff data at one school sees the whole network's staff directory.** `staff_directory` has no `row_level` filter, so any non-empty `allowedAbbreviations` gives a non-employee the entire directory (`src/cube/access.js:100`). `docs/guides/cube-access-individual-exceptions.md:160` tells maintainers a grant covers "everyone in the locations you named." Either add an abbreviation filter for non-employees (for example, a separate scoped-directory group) or correct the guide.

**4. The staging tests alert on bad sheet data but can't stop it.** The staging model is a table and `dbt build` replaces it before its tests run. `dim_staff_cube_access` is a view, so Cube reads the new rows even when an error-severity test then fails. The one over-grant this allows: scope `network` with a school name typed in. The staging YAML comment itself says that "reads as granting that one school while actually granting every location," and the mart grants network. Suggested fix: fail closed in `individual_exceptions_live`. Treat a row as a location grant only when `(location_scope = 'network' and additional_location_name = 'all') or location_scope in ('region', 'school')`. A lower-impact related case: with student axis `region` and staff axis `school`, the staging `coalesce` gives the staff axis `region` too.

**5. The guide says revocation is immediate, but it takes effect at the next midnight ET.** `resolveAccess` caches each viewer's access until the next midnight ET (`groupCache` in `cube.js`). Setting `status = revoked` mid-day doesn't reach an active viewer until then. Maintainers will rely on this wording for urgent removals (`cube-access-individual-exceptions.md:206-207`). Please change it to "by the next midnight ET" and document how to force it sooner, if there's a way.

**6. Stale `staff-directory` description.** `docs/guides/cube.md:755` and `.claude/rules/cube-authoring.md:367` still say it is emitted to every viewer. It now goes only to employees and to non-employees with a staff-axis grant.

### Minor

- **Duplicate plugin entry.** The Claude settings file lists the `superpowers-dev` marketplace twice, byte for byte. It looks like a merge artifact and doesn't loosen any permissions, but please drop the duplicate.
- **Missing PII tag.** `google_email` in `dim_staff_cube_access.yml` has no `config.meta.contains_pii: true`. The gap predates this PR, but the column now also holds contractor addresses.
- **Missing uniqueness test.** Please add a `unique` test on `dim_locations.abbreviation`. Student access is now keyed on abbreviation alone, so a duplicate across regions would silently give someone access to a second school. It's clean today.
- **Overstated comment.** The `individual_exception_scopes` `max()` is deterministic only while `single_remit_row` passes. Per item 4, that test doesn't block, so "safe deterministic" overstates it.
- **Contradictory YAML comment.** Staging YAML lines 123-124 say a blank axis cell "means the same thing" as `none`, but `not_null` is error severity and the guide says no cell is ever blank.
- **Stale comments.** The mart's `region_key` and `location_abbreviation` descriptions still say cube.js filters on them. They now only feed the allow-list. Also, `access.js:8` points to a `src/cube/CLAUDE.md` "View access policies" section that no longer exists.

### Not flagged

The five sensitive-field settings replace the base value instead of adding to it, so they can narrow access. The plan (section 4.3) and the published guide both define that as deliberate. Location grants are additive, as intended.

### Before merge

Please run the access matrix in `docs/guides/cube.md` against a branch deployment, plus a contractor with a school-only grant and a viewer whose row is revoked. Compare a network viewer's row count to prod to confirm how large item 1 is.

Review prepared with [Claude Code](https://claude.com/claude-code).

cristinabaldor:
Worked through the review at `ac5eab2` one finding at a time. Everything below is addressed in `de972b210b`, the PR body, or a filed issue. Two findings I'm pushing back on, with reasoning.

## Should fix

**1. Network viewers lose student enrollment rows — accepted, with a correction to the size.**

The mechanism is right. The 16% is not what a viewer loses.

Every NULL `location_key` row is PowerSchool's graduate placeholder school, `schoolid` 999999 — 18,861 Newark (AY2006-2025), 2,702 Camden (AY2016-2025), one Miami row with a null `schoolid`. `dim_student_enrollments` left-joins `int_students__schools` on `schoolid` and 999999 has no school row by design, so nothing upstream is broken. These are the alumni placeholder rows the project deliberately retains.

The 16% is `dim_student_enrollments` itself, which no view exposes now that `student_enrollments_view.yml` is gone from main. Measured through the `student_school_enrollments` join on each surviving surface, which also answers the attendance and assessment views you didn't measure:

| view's fact | total rows | now hidden | share |
| --- | ---: | ---: | ---: |
| `fct_student_attendance_enrollment_periods` | 4,402,039 | 0 | 0% |
| `fct_student_attendance_enrollment_daily` | 29,791,485 | 6 | ~0% |
| `dim_student_section_enrollments` | 875,103 | 13,244 | 1.51% |
| `fct_assessment_scores_enrollment_scoped` | 15,080,518 | 16,857 | 0.11% |

Noted in the PR body as intended. Not fixing the keys upstream: the only fix would be giving 999999 a `dim_locations` row, which would newly surface alumni to region- and school-scoped viewers who have never seen them.

**2. A former employee keeps their grants until suspension — declining the code change, documenting the window.**

Confirmed exactly as you describe. `enriched` spines on a current primary assignment, `non_employee_grantees` anti-joins `enriched`, so termination reclassifies someone as a contractor and their live grant rows keep applying until IT suspends the account.

I'm not taking the `dim_staff` anti-join, because it closes a case this PR exists to serve. You wrote "anyone who truly needs access after leaving can be re-entered on purpose" — there's no mechanism for that. A `dim_staff` row is permanent, so anti-joining it means a former employee kept on as a contractor can never be granted access again. Your own test description at `tests/properties.yml:470` calls that "the common way in."

The Google account is the intended authorization anchor, and suspension is the control that ends access. What was missing was anyone saying so, so the guide now has _A grant survives an employee's last day_: set `expiry_date` to the known end date, revoke as part of offboarding, and have IT suspend promptly. Happy to revisit if you think the window is wider than offboarding practice actually allows.

**3. A school-scoped contractor sees the whole staff directory — accepted, correcting the guide.**

Confirmed: `staff_directory` has no `row_level`, so any non-empty `allowedAbbreviations` yields the full network roster.

Taking the guide half of your either/or rather than adding a scoped group. Work-directory information is already internally public across the network, and the `is_employee` split in `2e365189a4` already draws the line that matters — an inert grant row gets nothing. The guide now says plainly that a staff-axis grant is not scoped to the locations named, and to read it as "may see the staff directory," not "at this school."

**4. Staging tests alert but can't stop — accepted, fixed in the mart.**

Confirmed end to end: staging is a table, `dim_staff_cube_access` is a `VIEW` in prod, and `dbt build` replaces the table before its tests run.

`individual_exceptions_live` now resolves `location_scope` to `'none'` for both shapes — `network` beside a real location name, and two axes naming different tiers — which `where location_scope != 'none'` then drops. The staging tests stay as the alert; the mart stops acting on the row. Guide updated to say a contradictory row grants nothing in the meantime.

Your placement was the right one. Fixing it in staging would have made the existing tests describe something other than the sheet.

While verifying this: the two axis scope columns are read for nothing but `!= 'none'` and a coalesce. The tier is written twice and a test forces the copies to agree, because one `additional_location_name` can't be both a region and a school. Filed #5520 to collapse them to one scope column plus two booleans, which makes the second sub-case impossible rather than guarded.

**5. Revocation is not immediate — accepted, and there are two lags.**

`groupCache` is a process-local `Map`, written at `cube.js:88` and `:192`, expiring at `nextMidnightEastern()`, never deleted, with no invalidation path anywhere in the file. So "by the next midnight ET" is the right worst case.

There's a second lag in front of it: the sheet only reaches Cube once the staging table rebuilds. Sibling sheet assets materialize 8-23 minutes apart in bursts, with a 5.9-hour gap between the last two.

The guide now names both, and the only lever there is — a Cube Cloud redeploy drops the process and the cache with it — plus suspending the Google account for anything urgent.

Also filed #5521 on the underlying oddity: a row's liveness is decided by a `status` word plus two dates, and `expired` is just a second copy of what `expiry_date` already says. A `revoked_date` would fold the one non-redundant state into the same date arithmetic.

**6. Stale `staff-directory` description — accepted.**

Fixed at `docs/guides/cube.md:755` and in `.claude/rules/cube-authoring.md`, both the group-taxonomy entry and the _Staff views are split_ prose.

## Minor

All six taken.

| # | what changed |
| --- | --- |
| Duplicate plugin entry | Handed to the branch author to apply — `.claude/settings.json` is Edit-denied to Claude by `permissions.deny`, so it isn't in this commit. |
| Missing PII tag | `config.meta.contains_pii: true` on `google_email` in `dim_staff_cube_access.yml`. |
| Missing uniqueness test | `unique` on `dim_locations.abbreviation`, with a description saying why it matters now. Clean today: 39 rows, 39 distinct, 0 null. |
| Overstated comment | Rewritten to say `max()` is deterministic across runs but not meaningful, and why the test can't prevent the state. |
| Contradictory YAML comment | A blank cell is no longer described as a synonym for `none`; it fails the error-severity `not_null`. |
| Stale comments | The three `dim_staff_cube_access.yml` descriptions now say `resolveAccess` expands these into the allow-lists rather than that `cube.js` filters on them. `access.js:8` repointed at `.claude/rules/cube-authoring.md` — and so were the three other pointers at moved sections in `cube.test.js`, `staff_work_history.yml` and `staff_reporting_relationships.yml`, since leaving three of four stale seemed worse either way. |

## Verification

`node --test cube.test.js access.test.js` — 102/102, same as your run. `trunk check --force --no-fix` clean on all 11 changed files. `dbt compile --select dim_staff_cube_access` succeeds.

Still outstanding from your _Before merge_ list: the access matrix against a branch deployment, a contractor with a school-only grant, and a revoked viewer. Those need the sheet re-staged against the new `google_email` key first, which is the unchecked box in the PR body.

The network-viewer row count you asked for is the table under finding 1.

cristinabaldor:
## Validation run against real data

Ran the access checks from *Before merge*. Summary first, then what was and was not proven, then the one thing that is red.

### Nobody loses their remit

Full population, not a sample. Baseline is prod `dim_staff_cube_access` with `origin/main`'s `access.js`; candidate is a dev build of this branch's mart with this branch's `access.js`. Same location and department universes on both sides. Diffed `groups`, `allowed_abbreviations`, `allowed_student_abbreviations` and `allowed_department_groups` per viewer.

| | viewers |
| --- | ---: |
| compared | 1,574 |
| resolve identically | 1,564 |
| changed | 10 |
| lost a staff location, student location, or department group | **0** |

`student-<tier>` is normalized to `student` so the rename is not counted as a loss. The tier now lives in the abbreviation array instead, and those arrays are compared directly, so a viewer whose breadth actually changed would appear as a location loss. None did.

The 10 that changed all lost the `student` group and nothing else. They are the same 10 whose work assignment carries no location, so their abbreviation array was already empty on main — they were handed a policy whose filter value was null rather than being denied up front. Same outcome, cleaner mechanism. Filed separately as #5524, since it predates this PR.

### Per-path checks

Driven from the real sheet, rebuilt between passes, resolved through the actual `access.js`.

| path | result |
| --- | --- |
| contractor with no employment record | resolves to a viewer, `is_employee` false |
| school grant, both axes | exactly that school on both |
| two live rows for one grantee | union of both, per axis |
| per-axis independence | a student-only grant left the staff array untouched |
| `revoked` row | grants nothing |
| expired row | grants nothing |
| future `grant_date` | grants nothing |
| remit override with no location | scope applied, no location granted |
| additive grant on an employee | base school kept, granted school added |
| network grant | all 39 locations, both axes |
| `network` beside a real location name | grants nothing, guard test fires |
| two axes naming different tiers | grants nothing, guard test fires |
| unresolvable address | default-deny |

The last two are the finding 4 fix. Both halves confirmed together: the row fails closed **and** the error-severity test reports it.

### What this does not prove

This validated identity resolution — that the mart emits the right grants and `access.js` turns them into the right groups and arrays. It did not run a query through Cube, so the policy-application half is still unverified: whether `row_level` on `allowed_student_abbreviations` actually filters rows.

That matters more than usual here, because the student policy is the part that changed shape — three tier policies became one flat policy on a new member. The staff side is unchanged from main. So the remaining pre-merge check is a real query per viewer against a branch deployment, which is what you asked for and what I have not done.

### Making it testable

`resolveAccess` had `kipptaf_marts` baked into both of its `dim_staff_cube_access` queries, and the read is `SELECT *` — so against prod's pre-PR view, `is_employee` and `additional_location_grants` come back undefined rather than erroring, and every grant silently unions nothing. A local run would have looked green while testing none of this.

`d99e5894b5` adds a deployment variable for that dataset, defaulting to `kipptaf_marts`. Only a `zz_` dev schema is honored; anything else warns and falls back, which is the whole safety property given there is no reliable Cube Cloud marker to gate on. Scope is the two `dim_staff_cube_access` reads only — `dim_locations` stays pinned to prod, because a dev copy of the location universe would change every viewer's resolved abbreviations and make a passing matrix meaningless.

Happy to pull it into its own PR if you would rather keep this one focused.

### dbt Cloud is red, and it is my fault

Two failures, both `Got 1 result`, both the finding-4 guard tests. The two deliberately-contradictory rows from the last check are still in the sheet, and CI reads it live. Clearing them turns it green; it needs a fresh build rather than a retry, since `dbt retry` replays against the PR-schema table already frozen with those rows.

Worth flagging as a property rather than an accident: that sheet is now load-bearing for CI. A bad row turns the PR red, and after merge it turns the nightly build red. That is the alert working, but it is new.

src/dbt/kipptaf/macros/utils.sql
cristinabaldor: Fair challenge, and you were right. It was doing nothing a plain comparison would not: `status = 'active'` plus two date checks, with the only fiddly part being the timezone-aware `current_date`. Worse, it standardised three of the *seven* places that rule appears — the other four spell it out by hand in `config.where` on this file — so it carried the indirection without buying the consistency.

Replaced it with a derived `is_live` boolean on the staging model in `db2dc06033`. All seven sites read the column now, the macro is gone from `utils.sql`, and the rule is visible in the table when you want to see why a row is not applying. It also makes #5521 (deciding liveness from dates alone, dropping the `status` word) a one-line change instead of seven.

Two deliberate behavior notes. A null on either date now yields false rather than "unbounded" — the macro read a missing bound as no bound, which fails open on an access grant, and both date columns already carry an error-severity `not_null`. And the value is computed at staging build time rather than per query, which is immaterial here since Cube caches each viewer's access until midnight Eastern anyway.

Verified behavior-preserving across the whole population: 1,574 viewers resolved through both the old and new paths, zero locations or department groups lost, and the same 10 viewers affected as before — all of them the null-location Miami staff in #5524.

.claude/settings.json
cristinabaldor: Done in `54f9ba2e2d`. The `superpowers-dev` marketplace entry appeared twice, byte for byte — a merge artifact from the `origin/main` merge, six added lines against zero deleted. Dropped the second copy; the block now matches `main`.

cbini (changes requested):
## Follow-up review at `6cb3b8d`: changes requested

Thanks for the fixes. Most of the first review is resolved, but two issues came in with the commits after it, and one of them (C1) breaks grant expiry. This review covers `ac5eab2..db2dc06`, plus the final state of the files after the `origin/main` merge in `6cb3b8d`. `node --test cube.test.js access.test.js` passes 102/102. None of those tests cover `d99e589`.

### Original findings

| # | Status |
|---|---|
| 1. Network viewers lose rows | **Documented as intended.** The PR body shows that every hidden row is a school 999999 alumni placeholder. You're right that giving it a location row would show alumni to region and school viewers. The data owners still need to sign off. |
| 2. Former employees keep grants | **Declined; your reason partly holds.** Anti-joining to `dim_staff` would block rehiring a former employee as a contractor, which is a fair objection. The risk is still there, though: a former employee keeps student access until IT suspends the account. A narrower fix covers both cases: treat a live row as a non-employee grant only when its `grant_date` falls after that person's last assignment end date. The guide's fallback (`cube-access-individual-exceptions.md:207-229`, rely on `expiry_date`) depends on C1 being fixed. |
| 3. Contractor sees whole directory | **Fixed.** The guide (`:170-176`) now says the directory is network-wide. |
| 4. Contradictory rows over-grant | **Fixed.** `dim_staff_cube_access.sql:171-185` sets `location_scope` to `none` for `network` next to a real name, and for two axes naming different tiers. Every scope/name combination is covered, and base scope is untouched. |
| 5. Revocation timing | **Fixed for `status` edits** (`:245-264`). Expiry is not fixed; see C1. |
| 6. `staff-directory` "always" | **Fixed.** |
| Minor items | **All fixed**, except that `dim_staff_cube_access.sql:211` still points to `src/cube/CLAUDE.md`. |

### C1 (Critical): expiry no longer takes effect on its date

`db2dc06` moved the `current_date` comparison into `is_live` in the staging model (`stg_google_sheets__people__cube_access_individual_exceptions.sql:35-40`). That model is a table, so the date is checked when the table builds, not when Cube queries it. Before this change, the `is_live_row` macro ran inside the `dim_staff_cube_access` view on every identity read.

The staging model rebuilds only when its dependency updates. That dependency is the Sheets asset, which updates only when the sensor sees the spreadsheet's Drive modified time change. The spreadsheet holds four tabs (`cube_access_role`, `department_override`, `department_rollup`, `individual_exceptions`), so the table rebuilds only when someone edits one of those tabs. No midnight rebuild happens. A sibling model on the same spreadsheet rebuilt 20 times in prod between 2026-09-21 and 09-23, at irregular times, with one gap of about 28.5 hours and no rebuild near midnight. Those were this PR's busiest editing days, and edits will be rarer after merge.

The result is that a row past its `expiry_date` keeps granting student data until someone happens to edit one of those four tabs, plus up to the next midnight ET for the Cube cache. There is no upper bound. The reverse also happens: a future `grant_date` never switches on without an edit. That direction fails closed, but it still contradicts the guide. The guide (`:218-219`, `:239-241`, `:273-274`) says expiry works on its own and makes that the main offboarding control.

**Fix:** keep the `status` check in staging, and move the `grant_date`/`expiry_date` comparison against `current_date` back into `individual_exceptions_live` in the view. The tests don't need query-time dates, so the `config.where` clauses and the singular tests can share a macro. Restoring `is_live_row` gives you that single definition again.

### I1 (Important): `CUBE_ACCESS_DATASET` can point prod identity resolution at a table developers can write to

In `src/cube/cube.js:10-43` (used at `:93` and `:170`), only the process environment sets the override. A request, header, JWT claim, or pasted security context can't reach it, which is good.

The only gate is `^zz_[a-z0-9_]+$`. That pattern accepts exactly the values that matter: any developer's `zz_<user>_kipptaf_marts` and the shared `zz_stg_*` datasets. The comment says a deployment that sets this by accident falls back to `kipptaf_marts`, but that only holds for non-`zz_` values. If someone sets a `zz_` value on the prod deployment, a developer can write a row into their own dev copy and give themselves network-wide student access.

Honoring the override logs nothing; only a rejected value logs. `resolveAccessDataset` isn't exported or tested, and neither `docs/guides/cube.md` nor the `cube-ops` skill mentions the setting. The existing dev pattern puts the `zz_` redirect in code, where review and `grep -r zz_ src/cube` catch it before merge. This moves it into deployment config, which nobody reviews.

**Fix:**
- Honor the override only when `CUBEJS_DB_BQ_CREDENTIALS` is unset. That's the local ADC path; every working deployment sets it (#4466), so it separates local runs from deployments, which the comment says nothing does.
- `console.warn` at startup whenever the override is honored.
- Export `resolveAccessDataset` and test it: unset, a `zz_` value, a non-`zz_` value, and credentials present.

Splitting it into its own PR, as you offered, also works.

### Minor

- `stg_..._individual_exceptions.yml:13-14` still says a null `grant_date`/`expiry_date` means open-ended. The SQL now treats null as not live. Fix it along with C1.
- `staff_work_history.yml:8` and `staff_reporting_relationships.yml:8` cite "Transformation lives in dbt" as if it were a section, but `cube-authoring.md:18` makes it a bullet. It's still findable.

### Checked and fine

- `de972b2` can't remove access someone had before this PR. It drops only the added location grant, never base scope.
- The `unique` test on `dim_locations.abbreviation` is correct.
- The `entity` test removal in `dim_staff_cube_access.yml` comes from the `origin/main` merge (#5530), not from this branch.

### Before merge

Fix C1, and gate or split out I1. Before merge, run one real Cube query per policy on a branch deployment, as your PR body notes.

Review prepared with [Claude Code](https://claude.com/claude-code).

cristinabaldor:
## Follow-up review addressed at `1199fac672`

Both findings fixed. One part of your suggested fix for C1 turns out to be impossible, which changed the shape of the answer — details below.

### C1 — confirmed, and worse than measured

You were right, and the number is bigger than the 28.5-hour gap you cited. Materialization history for the sibling model on that spreadsheet:

| gap | window |
| --- | --- |
| **54 days** | 2026-07-27 → 2026-09-21 |
| 7.2 days | earlier |
| 4.8 days | earlier |
| 28.5 hours | 2026-09-22 → 09-23 |
| 17.3 hours | 2026-09-23 |

The tight clusters are all this PR's editing days. In steady state it went 54 days between rebuilds, so an expired grant would have kept granting for that whole window. My error: I inferred a daily cadence from the burst of materializations during a day of heavy editing and dismissed the build-time risk on that basis, in the commit message no less.

The rule is a macro again, interpolated into `dim_staff_cube_access`. Verified rather than assumed — `current_date` now appears twice in the deployed view definition (`INFORMATION_SCHEMA.VIEWS`), so it evaluates per identity read.

**The macro cannot be shared with the `config.where` clauses.** dbt's generic-test config parser rejects project macros outright:

```
The macro 'is_live_row' is undefined.
Please note that the generic test configuration parser currently does not
support using custom macros to populate configuration values
```

`var()` is a builtin and works there; a project macro does not. So the four `config.where` clauses restate the predicate by hand, and it is three callers rather than seven — a parser constraint, not a choice. Recorded in the macro's own comment so nobody re-attempts it.

That also settles your original "not sure this macro is justified". The justification is neither reuse count nor consistency: a materialized answer to "is this live" is wrong the moment the clock moves, so it cannot be a column. Worth the round trip to find out.

One deliberate change alongside: a null on either date now denies rather than reading as an absent bound. The old form treated a missing `expiry_date` as "never expires", which fails open on an access grant. Both columns carry an error-severity `not_null`, so a null is already a broken row — it now grants nothing while the test reports it. The YAML description that contradicted this is fixed.

### I1 — all three, plus the docs

You were right that the shape check was not a safety property. `zz_` is exactly the prefix of every writable developer schema, so it permitted precisely the dangerous case.

- The override is honored only when `CUBEJS_DB_BQ_CREDENTIALS` is unset. Every working deployment sets it (#4466), so a deployment cannot honor the override at any value. The shape check is now a second gate rather than the only one.
- A honored override logs `cube_access_dataset_override` at startup, with the dataset. Previously only rejection logged, which was backwards.
- `resolveAccessDataset` moved to `access.js` as a pure function taking `(raw, hasDeploymentCredentials)`, exported and covered by four tests: unset, a `zz_` value on the local path, a `zz_` value with credentials present, and non-`zz_` values including `ZZ_UPPER_CASE` and a traversal attempt. 106 tests pass. Keeping it in `access.js` leaves `cube.js`'s documented five-export surface untouched.
- Documented in `docs/guides/cube.md` under _Testing row-level security locally_, including why `dim_locations` deliberately stays on prod.

Happy to still split it out if you would rather review it separately.

### Minor

All three fixed: `dim_staff_cube_access.sql:211`, and the two cube comments now cite the bullet rather than implying a section.

### Verification

`dbt build` on the staging model and mart: 50 nodes, 0 errors, 0 warnings. `node --test`: 106/106. `trunk check` clean on all 12 files. dbt Cloud CI green with zero warnings of any kind. The dev-schema rebuild still resolves the live network grant correctly and leaves the control viewers unchanged.

### Still open

One real Cube query per policy against a branch deployment. I did run the SQL API matrix against a local server pointed at a dev copy of the mart before these two fixes, which exercised the flat `student` policy end to end — a network viewer returned all four regions, a school-scoped viewer only their own, and a `none`-scope viewer got the auth-on hard failure rather than a silent zero. That is not the same as a branch deployment, and it predates `1199fac672`, so I am not counting it.

cbini (approved):
## Approved at `f5af68e`

`1199fac` fixes both issues from the follow-up review.

- **C1 (expiry):** the staging model now passes `status` through unchanged. The `is_live_row` macro is back and runs inside the `dim_staff_cube_access` view (`:187`), so expiry is checked each time Cube reads access. The mart, both singular tests, and all four `config.where` clauses compile to the same condition. A NULL date still grants nothing.
- **I1 (dataset override):** `resolveAccessDataset` returns `kipptaf_marts` whenever `CUBEJS_DB_BQ_CREDENTIALS` is set, and it now has tests. I tried to reach it from a request, header, JWT claim, pasted security context, and through dataset-name injection. None of them worked.
- **Minor items:** all fixed.

I pushed `f5af68e` with the two nits from my last pass:
- When `CUBE_ACCESS_DATASET` is set but ignored, startup now logs `cube_access_dataset_ignored` with the reason.
- `docs/guides/cube.md` names `CUBEJS_DB_BQ_CREDENTIALS` explicitly.

Cube tests pass 106/106.

### Before merge (not blocking this approval)

- Run one real Cube query per access policy on a branch deployment, as your PR body already notes.
- Get the data owners to sign off that network viewers no longer see the school 999999 alumni-placeholder rows (#1).
- #2 (a former employee keeps grants until their account is suspended) is still declined. The guide's `expiry_date` guidance covers it, and that guidance works now that C1 is fixed.

Review prepared with [Claude Code](https://claude.com/claude-code).

cbini (approved):

cristinabaldor:
## Access matrix run against `1199fac672` — and a correction

### Correcting my last comment

I listed "one real Cube query per policy against a branch deployment" as outstanding. That is not this repo's pattern, and I should have checked before repeating it. `docs/guides/cube.md` calls the **local SQL API** ground truth in three places, and the peer-review step says to use *either* the local matrix *or* Cube Cloud emulation:

> To test row-level security behavior, **either** use the local matrix tool — ground truth, one connection per viewer — **or** emulate a viewer in Cube Cloud

Three things make local the right surface here specifically:

- Cube Cloud's only unique coverage is `contextToGroups`, the path that bypasses `checkAuth`. This PR changes zero lines touching it.
- The axis that matters for denials is dev-mode vs auth-on, not local vs cloud. The run below had auth on and produced the production hard failure rather than a dev-mode silent zero.
- A branch deployment now *cannot* exercise the grant logic at all, because of the I1 gate. `CUBE_ACCESS_DATASET` is honored only when `CUBEJS_DB_BQ_CREDENTIALS` is unset, and a branch environment must set those credentials or identity reads fail closed entirely. Either it is set and the override is refused — reading prod's pre-PR mart, where every grant resolves to nothing — or it is unset and every viewer denies. Neither tests anything.

### The run

Local SQL API, auth on (`NODE_ENV=production CUBEJS_DEV_MODE=false`), identity resolved against a dev build of this branch's `dim_staff_cube_access`, one connection per viewer, querying `student_attendance_enrollment_daily_view`. Ten viewers covering every policy branch.

| viewer scope | result |
| --- | --- |
| contractor with a live network grant, no employment record | 4 regions — Camden 5,395 · Miami 3,912 · Newark 16,961 · Paterson 899 |
| network-scoped employee (×3, including one `all_in_scope` PII) | identical 4 regions |
| region-scoped employee | Newark 16,961 |
| school-scoped employee (Camden) | Camden 2,192 |
| school-scoped employee (Miami) | Miami 1,652 |
| region-scoped with an unresolvable location (the #5524 population) | denied |
| `none` on every axis | denied |
| address matching no identity | denied |

The contractor's counts match the network employees exactly, which is the assertion that matters: a grant resolves to the same row set as the equivalent base scope. The three denials are the auth-on hard failure (`Table or CTE with name 'student_attendance_enrollment_daily_view' not found`), which is what production does — a dev-mode server would have reported a silent zero and looked identical to a working narrow scope.

The startup log also confirms the I1 logging on the real path:

```
{"event":"cube_access_dataset_override","dataset":"zz_...","message":"Identity reads are resolving against a dev copy of dim_staff_cube_access, not kipptaf_marts. Expected only on a local run."}
```

### Finding 1

Accepted as intended and documented in the PR body. Getting the data owners to sign off that network viewers no longer see the school 999999 alumni-placeholder rows is ours to close before merge; it does not need anything further from you.

### Finding 2

Still declined, and the reason is stronger than when I first argued it. Your follow-up noted that the guide's `expiry_date` fallback "depends on C1 being fixed." C1 is fixed, and verified — `current_date` now evaluates inside the view on every identity read, so a grant genuinely stops on its expiry date rather than at some unbounded future rebuild.

That makes `expiry_date` a control that works rather than one the guide merely claims. Combined with revoking at offboarding and suspending the Google account — both in the guide's new _A grant survives an employee's last day_ section — the window is bounded by process rather than by code. Your narrower suggestion (treat a live row as a non-employee grant only when `grant_date` falls after that person's last assignment end date) is a reasonable future tightening, but it adds a join to the work-assignment history for a case the expiry date already covers.

PR #4618, opened 2026-07-29
fix(dbt): resolve the Focus ADDRESS feed from the student address of record

### LINKED ISSUE #4613: fix: Focus student address comes from an arbitrary Finalsite household
Split out of #4610, which fixed the same underlying assumption for DeansList Parent 2. This one is Miami and Focus, and unlike Parent 2 there is no correct fix available today.

## The mechanism

`stg_finalsite__contacts` flattens seven scalar columns off `households[safe_offset(0)]`:

```sql
households[safe_offset(0)].id as household_1_id,
nullif(trim(households[safe_offset(0)].address_1), '') as address_1,
nullif(trim(households[safe_offset(0)].address_2), '') as address_2,
nullif(trim(households[safe_offset(0)].city), '') as city,
nullif(upper(trim(households[safe_offset(0)].state)), '') as state,
nullif(trim(households[safe_offset(0)].zip), '') as zip,
households[safe_offset(0)].country as country,
```

Array position does not identify Finalsite's primary household. #4610 established this against the Finalsite UI: for a confirmed Newark student the UI's `Household1 (Primary)` is offset 1, not offset 0. The `Household` API object carries only id plus address fields, and the per-student primary-household designation set in the UI is absent from every field we ingest.

`rpt_focus__addresses` maps those columns directly onto the Focus ADDRESS contract, so whichever household happens to sit at offset 0 becomes the student's address of record in Focus. `rpt_focus__contacts` and `rpt_focus__demographics` read the same staging model.

## Sizing

Enrolled Miami students currently in the Focus feed:

| Shape | Students | Effect |
| --- | --- | --- |
| Single household | 1,161 | No choice to get wrong |
| Duplicate records, same address | 126 | Harmless — different record, same address |
| Genuinely different addresses | **158** | Offset 0 is effectively a coin flip |
| No household at all | **53** | `safe_offset(0)` yields null, so a blank address is pushed |
| **Total** | **1,498** | |

## Why this is harder than the Parent 2 fix

#4610 sidestepped the ordering problem by asking a relational question — does this contact share a household with Parent 1 — which needs no ordinal. An address cannot do that. It has to select exactly one household, so it genuinely requires the primary-household designation.

Two workarounds were considered and rejected:

- **Anchor on Parent 1's household.** Wrong for the case where Parent 1 does not live in the primary household. A confirmed Newark student has an empty primary household and her only parent in Household 2; using that parent's address would push the non-primary address.
- **Cross-reference the SIS address of record.** Finalsite is source of truth for this data, so the SIS cannot arbitrate. Miami is on Focus rather than PowerSchool in any case.

## What would unblock it

Determine whether the Finalsite contacts endpoint returns an undocumented per-household field carrying the primary designation. Our `Household` pydantic model declares seven fields and `fastavro` serializes only declared fields, so anything extra is dropped before it reaches GCS — and the Avro drift check compares top-level record keys only, so a nested subfield is never flagged. The external table therefore cannot answer the question; only a raw response capture can.

A probe script exists locally in the scratch directory. It calls the contacts endpoint, diffs the raw household keys against the declared model, and needs the Finalsite credentials to run.

- If extra keys appear, the fix is small: add the field to the pydantic model, regenerate the Avro schema, re-pull, and select on it.
- If not, this becomes a question for Finalsite support about exposing the designation, and there is no in-repo fix.

## Not in scope

Deciding whether `household_1_id` still earns its place in the staging contract. #4611 removed its last SQL consumer, but dropping a contracted staging column is a cross-project change with its own rollout.

cbini:
# Pull Request

## Summary & Motivation

...resolve the Focus `ADDRESS` feed from the student's own Finalsite household linkage, falling back to their primary contact's only when the student's side is not decisive, and withhold an address entirely when neither side resolves to exactly one complete address.

Closes #4613.

Phase 1 (#4637) is a merged prerequisite: it shipped the `finalsite` package models this PR consumes — `int_finalsite__contacts__households` and `int_finalsite__student_address_of_record` — and widened the `stg_finalsite__contacts` contract with the raw `households` array. Those models are materialized in all four district datasets, and the `zz_stg_*` staging copies are seeded, so dbt Cloud CI can resolve the `union_relations` column list.

### The problem

`stg_finalsite__contacts` flattens its address columns off `households[safe_offset(0)]`. Array position does not identify Finalsite's primary household — that designation is set in the Finalsite UI and is exposed by no API field, confirmed against the vendor spec and by the maintainer. So whichever household happened to sit at offset 0 became the student's address of record.

The feed is import-once with no overwrite path, and almost none of the Miami population had imported yet. Nearly every wrong pick was still prospective and would have become permanent, and unmarked, on first import at scale.

### The change

`households` is an array on **every** contact record, students and adults alike, and the student's own linkage turns out to be the disambiguating signal. Parents are linked to more households than their children — including ones the child does not live in — while the child is linked to the household they live in. Across the students with both records present, for **every** student whose own record resolves to exactly one complete address while their primary contact's record resolves to several, the student's address is one of the parent's. Zero exceptions.

So the rule is: take the student's own household linkage when it yields exactly one distinct complete address; fall back to the primary contact's linkage only when it does not; withhold and flag the student when neither narrows to one. An address is "complete" when street, city, state, and ZIP are all present.

`phone` comes from the primary contact rather than the student. That is load-bearing: a student's own `phone_1_number` is null for 1,497 of 1,498 enrolled Miami students while their primary contacts overwhelmingly carry one, and the import-once gate never checked phone — a student-sourced phone would have baked a permanently blank phone into Focus.

`rpt_focus__addresses` keeps its `status = 'enrolled'` scoping via a join to `stg_finalsite__contacts` used solely for `status`, matching every sibling Focus extract. Its only other filter is `address_source is not null`, which guarantees a complete address by construction — so no completeness filter is added here, and the kippmiami `#4320` anti-join gate stays the single remaining guard. The 12-column `ADDRESS_LAYOUT` contract does not change; only the rows and the values behind `address*` / `phone` do.

### This is a deliberate trade, not a strict improvement

Withholding is recoverable — fix the data in Finalsite and the address flows on the next run. A wrong import-once address is not recoverable. The change therefore trades ~150 fewer rows reaching Focus for the elimination of every arbitrary pick.

Measured against production on 2026-07-30, over the 1,497-student enrolled Miami feed population:

| Measure                                                     | Value         |
| ----------------------------------------------------------- | ------------- |
| Rows the offset-0 rule would export                         | 1,407         |
| Rows this rule exports                                      | 1,257         |
| Net change                                                  | −150          |
| Addresses that actually **change** (wrong picks prevented)  | 10            |
| Newly **enabled** (blank offset 0 masked a linkage that resolves) | 32       |
| Newly **withheld** (cannot be resolved to one)              | 182           |

Resolution breakdown of the exported rows: 1,240 from the student's own household linkage, 17 from the primary contact's. Of the withheld, 170 are flagged `ambiguous` (22 of those hold no complete address anywhere) and 70 are absent from the model for want of a `primary` relationship.

The 10 is the number worth reading twice: those are the students whose permanently-wrong address this prevents. "Eliminates every arbitrary pick" describes risk removed, not addresses changed — the two are easy to conflate.

Duplicate household records in Finalsite are the real cause of the residual, and retiring them is Miami ops work, out of scope here.

### Ops impact

`docs/reference/finalsite-focus-import.md` now documents all three reasons an address can be held back — incomplete, more than one with none marked as the one to use, and no primary contact designated — each with the fix the enrollment team should apply in Finalsite. Without that third entry, Ops would open a family's households, find one complete address, and reasonably conclude the pipeline was broken.

### AI Assistance

Claude Code implemented this under human direction, task by task, with a review pass after each task and a whole-branch review before shipping. Human-directed: the design decision to anchor on the student's own linkage rather than the primary contact (which reverted an earlier shipped approach), keeping the `enrolled` filter, sourcing `phone` from the primary contact, moving the households model to the intermediate layer, and the correction of the ops-doc wording about why duplicate households cannot be resolved. AI-authored: the SQL, the properties YAML, the unit tests, the prose drafts, and the production measurements above.

### General

      feedback; dismiss false positives with a brief reply explaining why. — Its
      check run is green but it posted **no findings**: its comment stalled as a
      partial checklist and there are no inline comments or reviews on this PR.
      See the comment below for the evidence. Treat the advisory review as
      absent, not as a clean bill of health. A separate whole-branch review was
      run before this PR left draft; what it found is described above.

### dbt

- [x] **Breaking change?** No columns renamed or removed. The
      `rpt_focus__addresses` 12-column contract is unchanged; only its rows and
      the values behind `address*` / `phone` change. The kippmiami wrapper needs
      no edit.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes. Editing the four `sources-kipp*.yml` files marks
      the whole finalsite source `state:modified`, so `state:modified+` fans the
      rebuild out across every kipptaf model reading finalsite. Expect a wide
      run, and treat pre-existing warn-level test noise as pre-existing.
      `src/dbt/**`.


cbini:
## Converting to draft — the rule in this PR measures worse than the status quo

Further measurement after this PR went green found that the primary-contact anchor as implemented here **increases** the number of students whose address of record is an arbitrary pick. Correcting the record before anyone approves it.

### What was wrong in the PR body

The PR body says the change resolves 11 of 167 ambiguous students and leaves roughly 150. The 11 is real. The number it breaks was never measured.

On the same 1,377 exported students:

| | Ambiguous |
| --- | --- |
| Old rule, student's own offset 0 | 152 |
| This PR, primary contact's offset 0 | **253** |

112 students became newly ambiguous, 11 became newly determinate. 242 of the 253 are genuinely different street addresses, not formatting drift.

### The mechanism

`households` is an array on every contact record, students and adults alike, and `stg_finalsite__contacts.address_*` is `households[safe_offset(0)]` of whichever contact row is read. So "the primary contact's address" resolves to "the primary contact's `households[0]`" — this PR changed whose offset 0 is read, not the fact that offset 0 is read. #4613's objection still applies, now to the parent's array.

Parents carry more household rows than students:

| Distinct complete addresses on the record | Student | Primary contact |
| --- | --- | --- |
| 1 | 1,245 | 1,145 |
| 2 | 152 | 225 |
| 3 or more | 2 | 30 |
| 2 or more, so the pick is arbitrary | 154 | 255 |

### The finding that changes the design

The student's household linkage is a **subset** of their primary contact's — true for 1,383 of 1,427 students. The parent is additionally linked to households the student is not in; the student is linked to the one they live in.

For all **112** students whose own record resolves to exactly one complete address while their parent's record has several, the student's address is one of the parent's. Zero exceptions.

So Finalsite already records which of the parent's several addresses applies, and records it on the student. This PR discards that signal. Those 112 are not coin flips landing differently — they are cases where a correct answer was available.

This also means #4613's framing was too pessimistic. The issue states that an address "has to select exactly one household, so it genuinely requires the primary-household designation." For 1,245 students the student's own record resolves to one address with nothing to pick, and for 112 more it disambiguates the parent's set. The designation is only genuinely needed for the residual ~143.

### New rule

Use the student's own household linkage when it resolves to exactly one complete address; consult the primary contact only when it does not. Measured on the 1,498-student feed population:

| Rule | Exported | Ambiguous |
| --- | --- | --- |
| Old, student offset 0 | 1,406 | 160 |
| This PR | 1,400 | 255 |
| Student offset 0 when complete, else primary contact | 1,439 | 162 |
| **New rule** | **1,438** | **143** |

### What happens to this PR

Draft until reworked. The new rule needs a per-contact count of distinct complete addresses, which is not available in `stg_finalsite__contacts` today — it exposes the flattened offset-0 columns and `household_ids`, but no addresses per household. That means an upstream change in the shared `finalsite` package across all four regions, so this ships as a district-then-kipptaf sequence rather than a single PR.

Nothing here should merge in its current form. The design spec on this branch will be revised with the subset finding before any code changes.

cbini:
## Heads-up for the reviewer: `claude-review` produced no findings on this PR

Its check run is green, but that green does **not** mean it reviewed the change and found nothing. It means the job exited zero. There is no review content anywhere on this PR.

Evidence:

- Its comment above is still the partial progress checklist. The last three items — apply the reuse/simplification lens, dispatch an independent subagent reviewer, and synthesize and post the final review — are unchecked, and the body has not changed since 19:26:05Z.
- `GET /pulls/4618/comments` and `GET /pulls/4618/reviews` are both empty: no inline comments, no submitted review.
- The Actions run reports `subtype: success`, `is_error: false`, 52 turns, and 4 permission denials, then only "Updated issue comment with job link". So it ran to completion and never posted the synthesis.

Please read the checkbox in the description accordingly — the advisory review is effectively **absent**, not clean. Nothing has been dismissed as a false positive, because nothing was raised.

What did cover the change, for what it is worth:

- A per-task review after each of the three implementation tasks, plus a whole-branch review before this left draft. That review returned one Important finding — the ops guide documented two reasons an address can be held back while this change creates a third, affecting 70 enrolled students with no documented remedy — and four minor ones. All are fixed in `cea59af28` and `7c0f21105`.
- dbt Cloud CI is green with zero warnings, and it genuinely built the new models rather than no-op'ing: the PR schema holds `int_finalsite__student_address_of_record` and `rpt_focus__addresses` plus the grain tests. The built extract is 1,257 rows / 1,257 distinct `student_id`, with no null `address`, `city`, `state`, or `zipcode`.

If you would rather have the bot's own pass on the record, toggling this PR back to draft and then ready re-fires `ready_for_review`, which is the only event that triggers it — it does not re-run on a push.

Issue #4619, opened 2026-07-29
feat(kipptaf): Cumulative GPA Monitor dashboard for KIPP Forward


anthonygwalters:
## Goal

Build the Cumulative GPA Monitor: an HS cumulative-GPA dashboard for KIPP Forward covering goal progress, GPA-band distribution, multi-year trend, equity cuts, and a cusp roster. Replaces a local React mockup with a production Tableau dashboard reading real warehouse data.

## Model scope

One file changes. Extend `rpt_tableau__gpa_cumulative_year` (grain: student by academic year; currently no downstream refs and no exposure) with the current-year enrichment already present on `int_powerschool__gpa_cumulative`, left-joined to the current-year row only via the model's existing `is_projected` flag. Without that gate, current-state values get stamped onto 22 years of history.

New columns:

- actual and projected unweighted cumulative GPA
- both existing integer band columns
- five-band labels, derived in the extract (the upstream bands are four and do not split 3.5 and above; band `4` has existing arithmetic consumers so it is left alone)
- `gpa_needed_for_cumulative_3_0`, `is_cumulative_3_0_attainable`
- a 3.0-cusp flag
- `potential_gpa_credits_current_year`
- `is_latest_graded_year`

Also add the missing exposure, and fix the workbook relationship, which currently matches on `student_number` alone with no academic year and so fans a student across every year they attended.

## Goal source

Goals come from `rpt_tableau__gpa_goals` (#4598): the Foundation 9th-grade goal as two region rows, the college-match goal as one org row. Goal attainment is measured on `cumulative_y1_gpa_projected_unweighted` so the rate reconciles with the topline cascade's `Projected Unweighted Cumulative GPA` indicator, which reads the same column. Current actual is a separate monitoring series and is never compared to a goal.

## Decisions

- Regions are Newark and Camden only. Paterson and Miami have zero HS students in any year on record, so this is structural rather than a pipeline gap. The region dimension stays in place so a future Paterson HS appears automatically.
- Default year is driven by `is_latest_graded_year`, computed from posted Y1 grades. academic_year 2026 (SY26-27) currently has zero posted Y1 grades while carrying loaded schedules, so a default-to-max-year would open on an empty dashboard. Cumulative earned credits cannot serve as the signal because they include prior years.
- Roster and equity cuts render at full granularity. Access is governed by the existing Tableau region and role gates.
- The roster shows credits in progress rather than a course count; no course-count column exists.

## Out of scope

[...]

PR #4598, opened 2026-07-28
feat(dbt): GPA goal scaffold — mixed-grain goals for the gradebook dashboard

anthonygwalters:
## Summary and Motivation

When merged, this adds a dedicated **goal scaffold for the GPA/gradebook
dashboard**, modeled on the proven topline aggregate-goals pattern: goals are
authored in a Google Sheet (one row per org/region/school x grade-band x GPA
metric), joined **upstream** to per-student GPA measures, and rolled up to a
per-grain actual-rate-vs-target.

Chain:

- `stg_google_sheets__gpa_goals` — new Google Sheets source (already staged to
  the staging environment).
- `int_google_sheets__gpa_goals` — adds `grade_band`, the grain
  `aggregation_hash`, `goal_proportion`, and the derived `higher_is_better`.
- `int_gpa__goal_student_metrics` — per-student spine (current-year weighted and
  unweighted Y1, projected unweighted cumulative), joining
  `int_extracts__student_enrollments` and `int_powerschool__gpa_term` on the
  cross-region `_dbt_source_project` key.
- `int_gpa__goal_aggregations` — goal-driven (inner-join) rollup per `org_level`
  (school/region/org), direction-aware numerator, `is_goal_met`,
  `progress_to_goal`.
- `rpt_tableau__gpa_goals` + a dbt exposure.

Design spec and plan:

- `docs/superpowers/specs/2026-07-28-gpa-goal-scaffold-design.md`
- `docs/superpowers/plans/2026-07-28-gpa-goal-scaffold.md`

**Scope:** delivers the threshold metrics (percent at weighted Y1 GPA at or
above 3.0, percent at unweighted cumulative GPA at or above 3.0). The `on_pace`
metric is wired as null placeholders and deferred to a follow-on, gated on the
per-student `is_on_pace_cumulative_3_0` and priority-subset flags.

**Not a breaking change** — all new models; no existing contracted column
renamed or removed.

Post-merge: the prod external for the new source gets created with
`stage_external_sources --target prod` (or the Dagster source asset) so the prod
build reads it.

## AI Assistance

Claude Code (Opus 4.8) authored the spec, plan, and models under human
direction, executed subagent-driven (per-model implement to review to fix) with
a final whole-branch review. The scaffold design, the goals-sheet schema, and
the goal-direction and grain decisions were human-set.

### dbt

- [x] Properties file for every new model with descriptions and a uniqueness
      test; `rpt_tableau__gpa_goals` is contract-enforced (validated `numeric`
[...]

Issue #4581, opened 2026-07-28
feat(dbt): dedicated GPA goal scaffold for the GPA/gradebook dashboard (topline-pattern, mixed grain)


anthonygwalters:
## Summary and motivation

Build a dedicated goal scaffold for the GPA/gradebook dashboard (the
`rpt_tableau__student_course_grades` / `int_powerschool__gpa_cumulative` family),
decoupled from the topline cascade, following the proven topline aggregate-goals
pattern.

GPA goals exist for multiple GPA types at mixed grains and must join **upstream**
(in an intermediate aggregation model), not in the reporting view, so the
dashboard can compare each grain's actual rate to its goal.

## Grain matrix

Goals may be defined at any of: org, region, school, org+grade, region+grade,
school+grade. As in topline, this collapses to `{org / region / school}` crossed
with a grade band, where a single-grade goal is just `grade_low = grade_high`.
The metric model does not group by grade; each goal row picks up exactly the
students its band matches.

## GPA types (goal_type / indicator) in scope

- percent of students at weighted Y1 GPA at or above 3.0 (the mandated leading
  metric)
- percent of students at unweighted cumulative GPA at or above 3.0 (the goal
  outcome)
- percent on pace (per-student unweighted needed-Y1 rolled up over the priority
  subset)
- others TBD on revisit

## Pattern to follow (topline precedent)

Model on `int_google_sheets__topline_aggregate_goals` plus
`int_topline__dashboard_aggregations`:

```text
goals source (Google Sheet) -> stg -> int goals model
  one row per (gpa_type, org_level, entity/schoolid, grade_low..grade_high,
  academic_year), carrying goal value, goal_direction (baseball = higher better
  / golf = lower better), has_goal, aggregation_type (Average/Divide/Sum), and an
  aggregation_hash join key encoding the grain.

int aggregation model
  UNION ALL one block per org_level (org / region / school); group the per-student
  GPA metric to that grain; left join goals on
  grade_level between grade_low and grade_high plus the org_level filter; compute
  is_goal_met and progress_to_goal.

rpt view reads the result (no goal logic in the rpt).
```

## Open questions to resolve on revisit

1. Goal source: a new dedicated Google Sheet vs the existing Ops-maintained
   GPA/SAT goals doc.
2. Join shape: pre-aggregate in dbt (topline style) vs attach the applicable
   goals to the per-student extract row so Tableau does the aggregation. The GPA
[...]

PR #4621, opened 2026-07-29
fix(dbt): exclude unmeasured students from gpa goal rate denominator

### LINKED ISSUE #4620: fix(kipptaf): gpa goal threshold denominator counts unmeasured students as failures
## Problem

`int_gpa__goal_aggregations` computes threshold-metric rates with `count(student_number)` as the denominator, meaning all students in the grain. A student whose measure is NULL is therefore counted as a non-achiever rather than excluded as unmeasured.

This is now visible in production. The single goal row currently loaded targets NCA grade 10 for academic_year 2026 at 58 percent on `y1_gpa_weighted`. Academic year 2026 is SY26-27, which has no posted Y1 grades, so all 239 students carry a NULL measure and `rpt_tableau__gpa_goals` reports:

- `metric_rate` of 0
- `is_goal_met` of false
- `progress_to_goal` of 0

Read literally, that says NCA 10th grade is at 0 percent against a 58 percent target. The true reading is not yet measurable. The same cohort in academic year 2025 was 107 of 208, or 51.4 percent.

This is the same failure mode as the elementary-students-with-no-GPA problem on the gradebook Y1 landing dashboard: a NULL measure silently counting as a non-achiever in a denominator.

## Fix

Change the threshold-metric denominator from `count(student_number)` to a count of students carrying a measurable value. `safe_divide` then yields NULL on a zero denominator, so the rate renders blank rather than 0.

At year end the two denominators converge, because nearly every student has a posted GPA, so the number the goal is actually evaluated against does not change. Mid-year, the current form biases the rate downward and lets it drift upward as grades post, which is an artifact rather than real progress.

The `on_pace` branch already uses `countif(is_on_pace_denominator)`. This change makes the threshold metrics behave the way `on_pace` already does.

## Also in scope

- Add `n_students_in_grain`, `n_students_measured`, and `n_students_met` columns so a genuine coverage gap stays visible instead of being quietly absorbed into the denominator, and so the rate is auditable. Surface them on `rpt_tableau__gpa_goals` too, so a consumer can tell a blank rate caused by no measurement from real zero attainment.
- Compute the rate once rather than in each of the three org-level blocks. The rate expression is currently triplicated across the school, region, and org aggregations; emitting the counts per block and dividing once removes that duplication.
- Update the `metric_rate` description in the properties files for both `int_gpa__goal_aggregations` and `rpt_tableau__gpa_goals`. Both currently describe it as the share of students in the grain and note a null result only for the `on_pace` case.
- Extend the existing `unit_gpa_goal_aggregations` unit test, defined in the model's properties file, with an entirely unmeasured grain plus the new count columns. It already covers a threshold metric at each org level, the grade-band handling, and a lower-is-better direction. It does not cover an unmeasured grain, which is the case this issue is about, nor the `on_pace` path.

Refs #4581
Refs #4619

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop `int_gpa__goal_aggregations` from
counting students with no measurable value as goal non-achievers.

The threshold-metric rate divided by `count(student_number)` — every student in
the grain. A student whose measure is NULL therefore landed in the denominator
as a failure. The `on_pace` branch never had this problem; it already divided by
`countif(is_on_pace_denominator)`.

This was live. The single goal row loaded today targets NCA grade 10 for
`academic_year` 2026 at 58%, on `y1_gpa_weighted`. Academic year 2026 is SY26-27,
which has zero posted Y1 grades, so all 239 students carried a NULL measure and
`rpt_tableau__gpa_goals` reported `metric_rate` 0, `is_goal_met` false,
`progress_to_goal` 0. Read plainly that says NCA 10th grade sits at 0% against a
58% target. The honest reading is "not yet measurable". The same cohort finished
academic year 2025 at 107 of 208, or 51.4%.

**Why the new denominator is the correct one, not merely a prettier one.** At
year end nearly every student has a posted GPA, so both denominators converge —
the number the goal is actually evaluated against does not move. Mid-year, the
old form biased the rate downward and let it drift upward purely as grades
posted, which is an artifact rather than progress.

**Verified against production data.** Rebuilt in a dev schema with upstreams
deferred to prod, the NCA row now reads `metric_rate` null, `is_goal_met` null,
`progress_to_goal` null, alongside `n_students_in_grain` 239 and
`n_students_measured` 0 — so the reason for the blank is legible rather than
guessed at. The extended unit test passes, as do both uniqueness tests, and
`trunk check --force` is clean on all four files.

Three additional changes, all in service of the same fix:

- **The rate is now computed once.** It was triplicated verbatim across the
  school, region, and org blocks. Each block now emits counts and a single
  `rates` CTE divides. Net effect on the three blocks is less SQL, not more.
- **`n_students_in_grain`, `n_students_measured`, and `n_students_met` are
  exposed**, on the intermediate and on `rpt_tableau__gpa_goals`. Excluding
  unmeasured students could otherwise hide a real coverage gap; these make the
  gap visible and the rate auditable. A consumer can now distinguish a blank
  rate caused by nothing being measured from genuine zero attainment.
- **The unit test gained the case that was missing.** `unit_gpa_goal_aggregations`
  already covered a threshold metric at each org level, grade-band handling, and
  a lower-is-better direction, but not an unmeasured grain — the exact case this
  PR is about. Two Camden students with no measures plus a matching school-level
  goal now assert a null rate, and the org row asserts
  `n_students_in_grain` 6 with `n_students_measured` 4, pinning the behaviour
  that under the old code would have produced 2/6 = 0.333 instead of 0.5.

## AI Assistance

Claude Code authored this change end to end — spotting the defect while
validating that #4598 had reached production, the root-cause analysis, the SQL,
the unit-test fixture, the production verification, and this description. The
decision to fix it upstream in the scaffold rather than work around it in the
consuming dashboard was human-directed.

### General

### dbt

- [x] Properties files updated — `metric_rate` descriptions corrected on both
      models, and the three new columns documented on both.
- [x] Exposure — the existing `gpa_goals_dashboard` exposure is unchanged; no new
      consumer is introduced.
- [ ] **Breaking change?** No. No columns renamed or removed and no type changes;
      the contract is additive. `metric_rate`, `is_goal_met`, and
      `progress_to_goal` change in value for grains where students carry no
      measure — from 0 / false / 0 to null. Every consumer of those grains was
      reading a wrong number, so there is nothing to preserve. Rollback is a
      straight revert; the columns are additive and nothing else reads them yet.
- [ ] External source — not applicable, no source changes.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

Closes #4620
Refs #4581
Refs #4619


anthonygwalters:
Verified each finding against the codebase before acting. Dispositions below.

## Important 1 — consolidate the tripled count/countif blocks: declining here

The proposal is technically sound. I checked: the three `joined_*` CTEs are union-compatible today (identical column shape and order, with `region`/`schoolid` null-cast on the region and org branches), `org_level` and `aggregation_hash` fully disambiguate the grains in the `GROUP BY`, and a single aggregation over their union would produce identical output while dropping roughly 34 lines.

I am not doing it in this PR, because the triplication is a deliberate pattern match rather than an oversight. The #4581 design states the model "Follows `int_topline__dashboard_aggregations`: a `UNION ALL` of one block per `org_level`", and that sibling aggregates per block then unions in exactly the same shape — `agg_union_student` and `agg_union_staff` carry a separate `group by` per block. Consolidating only this model makes the pair diverge, which turns a denominator fix into a scaffold refactor and leaves the two goal-aggregation models structurally different for the next reader.

Worth doing as a follow-up applied to both models together, so they stay consistent. Not worth widening this PR's review surface for.

## Minor 1 — no region-grain null assertion: fixed

Correct, and it was a real gap. Added a Camden region-level goal to the unmeasured-grain fixture, so the null rate is now pinned at both the school and region grains rather than school alone. The org grain keeps its indirect assertion via `n_students_in_grain` 6 against `n_students_measured` 4. Unit test passes with the fourth expected row.

## Minor 2 — check the workbook renders null sensibly: not applicable yet

There is no workbook. The `gpa_goals_dashboard` exposure URL is a placeholder ending in `/workbooks/TBD`, carrying a `TODO(#4581)` to fill in the published URL. Nothing consumes `rpt_tableau__gpa_goals` today, so no rendering can break.

The dashboard that will consume it is being designed under #4619, and that spec already requires goal tiles to read the not-yet-measurable state from `n_students_measured` rather than inferring it from a null rate, and to label it explicitly rather than rendering a bare blank tile. That is the reason those count columns are in this PR at all.

## The four items under "needs local verification"

All four have been run, which the review environment could not do:

- Unit test — passes, including the new unmeasured-grain rows.
- Both uniqueness tests — pass, via `dbt build --select int_gpa__goal_aggregations+`, which also exercises contract enforcement on the `rpt_` extract and confirms the three additive `int64` columns are accepted.
- `trunk check --force` on all four changed files — no issues. ST06 ordering was the real risk and is clean.
- The production numbers — re-confirmed by querying the rebuilt dev view whose upstreams were deferred to prod. The NCA row reads `metric_rate` null, `is_goal_met` null, `progress_to_goal` null, with `n_students_in_grain` 239 and `n_students_measured` 0, against 0 / false / 0 before the change.

## Unrelated, for the record

The dbt Cloud failure on the first run of this PR was not the change. CI defers unmodified parents to the shared staging clones, and those had never been seeded with the models #4598 merged an hour earlier, so `int_google_sheets__gpa_goals` was missing from `zz_stg`. Seeded the three missing relations with a scoped `dbt clone`; dbt Cloud is green now.

PR #4622, opened 2026-07-30
feat(dbt): extend gpa cumulative year extract for the Cumulative GPA Monitor

### LINKED ISSUE #4619: feat(kipptaf): Cumulative GPA Monitor dashboard for KIPP Forward

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will extend `rpt_tableau__gpa_cumulative_year` so a single student-year table can drive the Cumulative GPA Monitor — a KIPP Forward dashboard covering goal progress, GPA-band distribution, multi-year trend, equity cuts, and a cusp roster. It also lands the design spec and implementation plan for that dashboard.

One model changes. The year-grain extract gains a LEFT JOIN to `int_powerschool__gpa_cumulative` gated on the model's existing `is_projected` flag, plus derived band labels, a 3.0-cusp flag, and a default-year flag.

### The gate is the load-bearing part

`int_powerschool__gpa_cumulative` is current-state, keyed on student and school with no academic year. Joining it naively would stamp today's values onto 22 years of history. Gating the ON clause with `and gcy.is_projected` confines it to the current-year row. Verified: across 54,271 historical rows spanning 22 years, every current-state column reads zero non-null; the current-year rows carry 3,598 populated.

### Why the workbook reads one table and not two

The existing Tableau data source relates this extract to `rpt_tableau__student_course_grades` on `student_number` alone, with no academic year — so a student fans across every year they attended (current seniors carry up to ten rows here). The Monitor therefore reads this table by itself, which makes the fan-out structurally impossible rather than merely filtered away. The model description previously recommended that relationship; this PR replaces that guidance.

### Verification

- **Contract**: 52 SQL columns against 52 properties entries, exact match both directions.
- **Bands**: all five values span 20 distinct academic years — the check that proves the primary band reads the year-grain spine rather than the current-year-only join. Boundaries exact, no leakage (`3.0-3.49` maxes at 3.49, `below 2.0` maxes at 1.99).
- **Cusp flag**: 1,938 HS rows, every one inside the `2.5-2.99` band, none outside.
- **Default year**: resolves to 2025, the most recent year with posted Y1 grades. Academic year 2026 is SY26-27 and correctly not flagged.
- **Build**: `dbt build --select rpt_tableau__gpa_cumulative_year` passes 7 of 7, including a new singular test asserting exactly one year carries the flag.
- **Lint**: `trunk check --force` clean across all eight changed files.

### Reconciliation against the topline indicator

The same at-3.0 rate is published by the topline cascade, so the two must be explainable against each other. For the latest graded year, HS only, this extract reports 1,849 students at 40.35 percent. The topline `Projected Unweighted Cumulative GPA` indicator reports 46.49 percent over 5,993 students. The gap is scope and grain, not disagreement: topline blends middle-school grades 5 through 8 (which run 50 to 55 percent) with high school, and averages roughly 75 weekly snapshots per student rather than one point-in-time read. Restricting topline to the same population and grain yields 1,851 students at 38.29 percent — within about two points and two students. The residual is definitional, since this extract reports a completed year's actual while topline reports its last weekly projection.

## Known pre-season state — please read before judging the dashboard

`is_latest_graded_year` currently resolves to 2025 while `is_projected` resolves to 2026. They are different years right now, and that has a visible consequence.

On the dashboard's default slice, five columns are entirely null: `cumulative_y1_gpa_unweighted_as_of_today`, `gpa_needed_for_cumulative_3_0`, `potential_gpa_credits_current_year`, `gpa_band_as_of_today_label`, and `is_cumulative_3_0_attainable`. The cusp roster will therefore open with three of its five columns blank, and the students-to-move tile will read zero.

This is not a broken join. Three of those columns derive from `potentialcrhrs_current`, which upstream sums credit hours only where `academic_year = var('current_academic_year')` — that variable was advanced to 2026 in July, and SY26-27 has no course credit hours yet, so `safe_divide` nulls the family. Confirmed 0 non-null across all 27,044 rows of `int_powerschool__gpa_cumulative` while the projected family is populated at 25,674. The two flags coincide from roughly November and everything fills in.

## Five deviations from the design spec

Each is documented with rationale in the plan.

1. **Band labels derive from the year-grain spine, not the current-state join.** This is a correction. The spec sourced both labels from the gated join, which is null on every prior year — that would have left 22 years with a null band and silently emptied the trend, splay, and equity surfaces.
1. **`cumulative_y1_gpa_projected_unweighted` is not added.** On the current-year row the existing `cumulative_y1_gpa_unweighted` already holds that value; a duplicate column invites drift. That column's description now states the semantics explicitly instead.
1. **The two integer band columns are not added.** No dashboard surface consumes them; the labels serve every band need.
1. **The one-latest-year assertion is a singular test, not `expression_is_true`.** That macro compiles to a per-row `where not (...)` and cannot express a cross-row cardinality claim.
1. **`is_on_cusp_3_0` derives from the year-grain spine**, for the same reason as the first item, and uses a half-open `>= 2.75 and < 3.00` which closes the gap the spec's `between 2.75 and 2.999` leaves at values like 2.9995.

## One unrelated file, deliberately

`rpt_tableau__student_course_grades.yml` gets a one-line description fix. Its `gpa_needed_for_cumulative_3_0` described the value as a **weighted** GPA, but the upstream formula uses `unweighted_points_prior` with no weighted term — the description has been stale since PR #4529 corrected that calculation to an unweighted basis, and the authoritative source model documents it correctly. It is included because it is the origin of the same error in this PR's new column, caught in review. A properties description edit does not mark a model `state:modified`, so it causes no CI fan-out.

## AI Assistance

Claude Code produced this branch end to end under human direction: the design spec, the implementation plan, the model changes, the tests, the production verification, and this description. Execution ran task-by-task with a fresh implementer and an independent reviewer per task, plus a whole-branch review at the end. Human direction set the scope, chose to fix the goal-rate denominator upstream rather than work around it, and decided the region coverage question. Three review findings changed the code: the weighted-versus-unweighted description error, a dropped region-scope clause, and two folded YAML scalars that repo convention requires be literal.

### General

### dbt

- [x] Properties file updated — every new column documented, and two stale descriptions corrected.
- [x] Exposure added — `cumulative_gpa_monitor`, depending on both `rpt_tableau__gpa_cumulative_year` and `rpt_tableau__gpa_goals`. Its URL is a placeholder with a `TODO(#4619)` until the workbook is published, matching the existing pattern for unpublished workbooks in that file.
- [ ] **Breaking change?** No. Purely additive — no column renamed, removed, or retyped. Existing columns keep their meanings.
- [ ] External source — not applicable.

## Follow-ups, not in this PR

- The Tableau workbook itself is a manual deliverable; the plan carries step-by-step build instructions including the pre-season caveat above.
- #4621 merged while this branch was being built, so `main` is merged in here and the goal-rate denominator fix is live. The goal tiles now need only Ops populating three rows in the GPA goals sheet.
- Every college-enrollment surface from the original mockup is deferred to a separate spec. Those rest on an assumed constant with no warehouse source; fitting it means joining historical graduating cohorts to their actual kippadb college enrollment.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

Refs #4619
Refs #4621


anthonygwalters:
Verified each finding against the codebase and the warehouse before acting. Two fixed, one refuted by data, one deferred to a human call, one where I am keeping the current behaviour for a factual reason. Fixes are in `8a756cc`.

## Important 1 — default-year signal not region-scoped: FIXED

Correct, and this was also raised independently by the whole-branch review, which I had adjudicated as low-risk. Two reviewers converging on it changed my read.

I ran your check first. All three districts currently max at the same year, so there is no divergence today:

| `_dbt_source_project` | latest graded year | rows with `gpa_y1` |
| --- | --- | --- |
| kippnewark | 2025 | 157,352 |
| kippcamden | 2025 | 42,436 |
| kippmiami | 2025 | 16,512 |

So this was prospective rather than live. I fixed it anyway — the cost is one predicate, and the failure mode is precisely the one the column exists to prevent. The CTE now filters `_dbt_source_project in ('kippnewark', 'kippcamden')`, matching the extract's own region scope, so the two cannot drift.

Verified the change is behaviour-neutral today: `is_latest_graded_year` still flags exactly one year, 2025, and the model still builds 7 of 7 including the singular cardinality test.

## Important 2 — third stale "weighted" description: FIXED

Confirmed. `src/dbt/kipptaf/models/powerschool/intermediate/properties/int_powerschool__gpa_cumulative.yml` carried it in both places you named, while the authoritative source-system model documents both as unweighted and the formula uses `unweighted_points_prior` with no weighted term.

Both corrected. `gpa_needed_for_cumulative_3_0` now reads "Unweighted Y1 GPA the student must average...", and `is_cumulative_3_0_attainable` now reads "credit-weighted maximum unweighted GPA" — the leading "credit-weighted" is correct and stays, since it is a credit-weighted average.

That makes three copies of this drift corrected across two PRs. Good catch that the file my new join actually reads from was still stale.

## Minor 3 — uniqueness test district scoping: REFUTED by data

Ran your query against prod. Zero collisions:

```text
n_colliding_student_numbers: 0
```

No `student_number` appears under more than one `_dbt_source_project` anywhere in the model. The existing `(student_number, academic_year)` test is therefore sufficient, and the Tableau workbook cannot blend two students on that key. No change.

Worth noting the sibling you compared against, `rpt_tableau__student_course_grades`, has a different grain — student x term x course x category — so its PK genuinely needs the extra column. This model is one row per student-year and does not.

## Minor 4 — `contains_pii` tag: deferring to the team, not applying unilaterally

The observation is accurate: the model carries `student_name` and `student_number` plus indirect identifiers, with no model-level tag, and this PR does deepen the surface by tying GPA attainment to a named roster.

I am not adding it in this PR. The repo's PII guidance is explicit that tagging scope for a PII-heavy model — direct identifiers only, versus direct plus indirect — should be confirmed with the team rather than chosen by whoever is editing. Picking a scope silently here would set a precedent for the whole extract family. Flagging it for an explicit decision instead; happy to apply whatever scope is chosen, in this PR or a follow-up.

## Minor 5 — band label text: keeping the current labels, and the documented convention has a factual problem

You are right that `src/dbt/kipptaf/models/students/CLAUDE.md` documents a five-band "KIPP GPA Band" with cut-offs identical to mine, applied to `cumulative_y1_gpa_unweighted` — the exact column this PR bands. So this is the same band, deliberately.

I am not adopting the documented label strings, because the top one is wrong for this data. The doc gives band 5 as `3.50 to 4.00` with max 4.00, but the unweighted scale reaches 4.33 — the doc says so itself two sections earlier in the KTAF Unweighted table. Against production:

```text
rows with a non-null unweighted cumulative : 49,850
rows above 4.00                            :  1,574
max observed                               :   4.33
```

1,574 students sit above the documented band's stated ceiling. A label reading `3.50 to 4.00` misdescribes every one of them, and under a literal reading of the table they fall outside all five bands. `3.5+` is correct for an unbounded top band, which is what the CASE actually produces.

Two further notes. That doc's "implement CASE statements to match these definitions exactly" instruction is scoped to `models/students/`; this model lives in `models/extracts/tableau/`. And the labels here are chart legend text for a dashboard, where `3.5+` is materially more readable than `3.50 to 4.00`.

The drift concern is real though. I would suggest reconciling the doc — either correcting band 5's ceiling to 4.33 or restating it as an open upper bound — rather than propagating a bound that 1,574 rows already exceed. Out of scope for this PR, and it touches a KIPP Foundation definition, so it wants a human decision rather than my edit.

## On the `/simplify` note

Agreed on the duplicated CASE, and on not treating it as a violation. The whole-branch review reached the same conclusion independently, on the same grounds: the direct upstream writes its own two band CASEs out longhand, so per-model banding is the established pattern here rather than something this PR should have centralised on its own.

anthonygwalters:
@claude review

PR #4628, opened 2026-07-30
feat(dbt): assign PM1 to all teachers and scope by ADP job function

cristinabaldor:
# Pull Request

## Summary & Motivation

When merged, this pull request will change two things about who appears in the
SchoolMint Grow PM observation feed
(`rpt_tableau__schoolmint_grow_observation_details`), and document the model.

### 1. Every teacher is eligible for PM1, regardless of start date

The PM1 branch of `pm_round_eligible` previously required one of four
carve-outs -- currently a Teacher in Residence, a prior-year TiR or Para (New
Lead), KIPP Miami, or hired between April 1 and three weeks before the lockbox
date. All of that is gone; PM1 is now simply `true`.

- PM2 and PM3 are unchanged -- the work assignment must have started at least six
  weeks before the term's lockbox date.
- The `tir_previous` CTE and its `left join` are removed. They existed only to
  feed the PM1 branch, and `prior_year_tir` had no other consumer in the repo. The
  CTE resolved to 271 rows across 271 distinct `employee_number`, so removing the
  join is row-count-neutral.

### 2. Teachers are scoped by ADP job function, not job title

The current-year branch previously selected teachers with
`job_title like '%Teacher%' or job_title like '%Learning%'`. It now uses the
roster's `job_function` (the ADP job group, codes `TEACH` and `TIR`), falling back
to the title match only where `job_function` is null.

The fallback is load-bearing: `job_function` is not populated on newly created
work assignments, so without it 23 current teachers would silently drop out of PM
tracking -- all with a July 2026 `effective_date_start`, across all four entities.

On current roster rows this nets 774 to 786:

- **Drops 14 false positives** the title match dragged in -- Teacher Aide,
  Substitute Teacher, Learning Disabilities Teacher Consultant, and Managing
  Director of Teaching and Learning.
- **Adds 26 genuine teachers** the title match missed. 21 are Paterson TiRs
  titled *Instructor in Residence*, which the pattern never caught. The other 5
  are classified into the Teacher job group in ADP despite a non-teaching title
  (3 Assistant School Leader, 1 Dean, 1 Director of Orchestra and Music).

### 3. Documentation

The properties file had no descriptions at all. It now has a model description and
a description for all 56 contract columns. Column names, data types, and order are
unchanged, so contract enforcement is unaffected.

Most descriptions came from `int_performance_management__observations` (the real
origin of the observation columns) and `stg_google_sheets__reporting__terms`;
derived columns came from the upstream model SQL. Three points were resolved
against prod data rather than assumed:

- `tracking_rubric` holds Grow form names (Manager Reflection, Teacher Reflection,
  Coach Comments, Coach Scores), not observation rubric names.
- `grade_taught` ranges 0 to 12 with no negative preschool codes, contrary to the
  upstream description inherited from a student model.
- `etr_score` and `so_score` are populated only on prior-year rows.

Also replaces a leftover `TODO: document distinct` stub on `recent_leave` with the
`grain projection:` annotation the SQL guide requires.

## Impact, measured against prod

Combined effect of changes 1 and 2 on PM1:

| Metric | Value |
| --- | --- |
| Teachers with PM1 tracking rows, after this PR | 790 |
| Eligible in prod today | 170 |
| Newly eligible | 620 |

Measured on the model's own spine (roster joined to the AY2026 `PMS`/`PMC`/`TR`
terms). The same population under the old title filter is 776, so the job-function
change accounts for 14 of the 620.

### A structural finding worth recording

`recent_leave` **cannot** exclude anyone from PM1 this year. The CTE inner joins
`PMS` terms only, and AY2026 has no `PMS` PM1 term row -- AY2024 and AY2025 both
did. So it yields PM2 and PM3 rows only, and
`when r.recent_leave then false` never fires for PM1.

The leave guard is deliberately kept in the `case` so the behavior is correct if
Ops re-adds that term row. This is now commented at the CTE, since it is invisible
from the SQL alone.

## Rollback

Revert the commits. No schema change, so no coordination is needed with downstream
consumers -- only which teachers appear, and the values in `pm_round_eligible`.

## Follow-ups

None block this PR.

- #4631 -- expose `job_function_code` on the roster, so the filter can compare
  stable codes (`TEACH`, `TIR`) instead of display names.
- #4633 -- the model has no uniqueness test, and cannot get one until the
  prior-year archive branch's fan-out is understood (about 685k excess rows on a
  7-column key; the current-year branch is nearly clean at 30).
- The two ADP data-quality gaps behind the null fallback are tracked with Ops for
  an HR follow-up.

## AI Assistance

Human-directed: dropping PM1 round eligibility, keeping the leave exclusion,
switching to the job group, keeping Paterson's Instructor in Residence staff in
scope, and the `lockbox_date` note about BOY coaching conversations.
AI-assisted: the SQL edits, the prod impact and row-count-neutrality queries, the
first draft of the column descriptions, and the PMS/PM1 finding.

### General

      Both findings verified. The TODO stub is fixed. The missing uniqueness test
      is real and pre-existing, but its suggested remedy would fail immediately --
      no near-unique key exists -- so it is tracked in #4633 with the
      measurements. The updated review's request to document the job-function
      change and re-check the impact numbers is what this description now does.

### Dagster

Skipped -- no Dagster changes.

### dbt

- [x] Properties file -- no new models; the existing properties file gains a model
      description and descriptions for all 56 columns.
- [x] Exposure -- two exposures already depend on this model
      (`schoolmint_grow_dashboard` and `coaching_conversation_tool`). No change
      needed, since no columns were added or removed.
- [x] **Breaking change?** Not breaking. No column is renamed, removed, or
      retyped. This is a value-level and row-population change, which the two
      Tableau exposures will see as more teachers marked eligible for PM1.
- [x] External sources -- none added or modified, so no `stage_external_sources`
      run is needed.

### Docs

Skipped -- no schedule, sensor, or integration changes.

## CI checks

- [x] **Trunk** -- passes.
- [x] **dbt Cloud** -- passes.
- [x] **Dagster Cloud** -- not triggered; no Dagster paths changed.

## Troubleshooting

### DIFF
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__schoolmint_grow_observation_details.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__schoolmint_grow_observation_details.yml
@@ -103,7 +103,8 @@ models:
         description: >-
           Date after which the tracking term's data is locked for reporting. It
           also anchors PM2 and PM3 eligibility and the recent-leave lookback
-          window. Null on prior-year rows.
+          window. Null on prior-year rows. For unscored/unlocked BOY rounds,
+          this is the start date for coaching conversations.
       - name: final_score
         data_type: float64
         description: >-
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__schoolmint_grow_observation_details.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__schoolmint_grow_observation_details.yml
@@ -1,115 +1,280 @@
 models:
   - name: rpt_tableau__schoolmint_grow_observation_details
+    description: >-
+      Teacher observation detail feed for the SchoolMint Grow performance
+      management Tableau workbook. UNION of two branches. The first is a
+      current-academic-year tracking scaffold that crosses every active teacher
+      against the PM tracking terms and left-joins whatever observation data
+      exists, so teammates who have not been observed still appear with
+      `is_observed = 0` and a populated `pm_round_eligible`. The second carries
+      actual observation responses from prior years, where every `tracking_*`
+      column and `pm_round_eligible` are null. Grain is one row per teammate per
+      tracking term per observation measurement item, so rows fan out below the
+      observation level — deduplicate on `observation_id` before counting
+      observations.
     columns:
       - name: employee_number
         data_type: int64
+        description: ADP employee number of the observed teammate.
       - name: teammate
         data_type: string
+        description:
+          Formatted name of the observed teammate, from the current staff
+          roster.
       - name: entity
         data_type: string
+        description: >-
+          Home business unit of the teammate — the KTAF central entity or a
+          regional entity.
       - name: location
         data_type: string
+        description:
+          Name of the teammate's home work location (school or office).
       - name: grade_band
         data_type: string
+        description: Grade band of the teammate's home work location.
       - name: department
         data_type: string
+        description: Name of the teammate's home department.
       - name: job_title
         data_type: string
+        description: >-
+          Job title the teammate held on the roster-history row covering the
+          tracking term.
       - name: manager
         data_type: string
+        description: Formatted name of the teammate's manager.
       - name: worker_original_hire_date
         data_type: date
+        description: Date the teammate was originally hired.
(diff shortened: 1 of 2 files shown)


cristinabaldor:
@claude update your review

GabyRangelB (approved):

PR #4629, opened 2026-07-30
chore(dbt,dagster): archive Leadership Development for the Lattice migration

### LINKED ISSUE #4627: chore(dbt,dagster): archive Leadership Development app, dashboard, and tables for Lattice migration
The Leadership Development performance-management workflow is moving to Lattice. Mark its models, dashboard, and tables as archive-only and stop every automated extract or refresh this repo controls.

## Current automated refresh loop

1. The AppSheet app writes to `kipptaf_appsheet.src_leadership_development_output` and `src_leadership_development_active_users`.
2. `kipptaf__google__bigquery__table_modified_sensor` polls those two BigQuery tables every 5 minutes and emits materializations for the matching source assets.
3. Eager dbt automation conditions cascade through the eight-node cluster below.
4. Separately, `kipptaf__google__sheets__asset_sensor` polls the Leadership Development metrics Google Sheet and triggers the metrics staging model.

## The cluster (8 nodes, fully self-contained)

| Node | Kind |
| --- | --- |
| `stg_google_appsheet__leadership_development__output` | staging table |
| `stg_google_appsheet__leadership_development__active_users` | staging table |
| `stg_google_sheets__performance_management__leadership_development_metrics` | staging table, live Sheet external |
| `snapshot_leadership_development__output` | snapshot, holds the rating/note history |
| `rpt_appsheet__leadership_development_roster` | extract view, feeds the AppSheet app |
| `rpt_appsheet__leadership_development_assignments` | extract view, feeds the AppSheet app |
| `rpt_tableau__leadership_development` | extract view, feeds the dashboard |
| `rpt_gsheets__leadership_pm_reminders` | extract view, feeds reminder emails |

Nothing outside the cluster consumes any of these. `int_people__leadership_crosswalk` is a different model (instructional leadership crosswalk, consumed by staff roster, DDI, topline, and gradebook audit) and is explicitly out of scope, as are the SchoolMint Grow observation models and `rpt_gsheets__pm_assignment_roster`.

## Plan

Use the archive pattern documented in `src/dbt/kipptaf/CLAUDE.md`: disable the model, register the frozen relation as a BigQuery-native source, repoint consumers.

1. Set `config.enabled: false` on all eight nodes.
2. Set `config.enabled: false` on the three source entries that drive the polling sensors: the two `google_appsheet` Leadership Development entries in `sources-bigquery.yml` and the metrics entry in `sources-external.yml`. Both sensors build their asset selection from the parsed manifest, so disabling removes them from polling.
3. Register the four frozen tables as BigQuery-native archive sources, using `identifier` so no warehouse rename is needed:
   - new source `google_appsheet_archive`, schema `kipptaf_google_appsheet`: the two staging tables plus the snapshot
   - existing `google_sheets` source, schema `kipptaf_google_sheets`: the metrics table
4. Keep all four exposures, relabel them as archived with a note about Lattice, and repoint `depends_on` from `ref()` to the new archive `source()` calls. An exposure whose `depends_on` names a disabled model breaks parsing, and it would also break the Dagster code location load, since `code_locations/kipptaf/dbt/assets.py` resolves exposure refs against `manifest["nodes"]`.

## What this does not stop

These are outside the repo and need a separate owner:

- **Tableau server-side extract refresh.** The `leadership_development` exposure has no `cron_schedule`, so Dagster has never refreshed this workbook. Any refresh happening today is a Tableau-side schedule on workbook `8f8b058f-2408-470d-95b0-681d6c8b65f4` and must be removed in Tableau.
- **The AppSheet app.** It is the writer into BigQuery. Disabling dbt models stops the reads, but the app keeps accepting entries until it is turned off in the AppSheet console.
- **Reminder emails.** `rpt_gsheets__leadership_pm_reminders` joins live `int_people__staff_roster`, so the view keeps returning rows on every read even once dbt stops managing it. If an Apps Script on the reminders Sheet mails those rows, leaders will keep being nagged about a dead workflow.

The same live-join caveat applies to all four `rpt_` views: three resolve `int_people__staff_roster` transitively through `rpt_appsheet__leadership_development_roster`, and `rpt_gsheets__leadership_pm_reminders` reads it directly. Disabling a view does not freeze its output. A follow-up hard cut, dropping all four `rpt_` views, is the only way to stop those reads; it breaks the app immediately, so it is tracked separately in #4634 with the DDL and a prerequisite checklist rather than run here.

## Archive of record

The four frozen tables are genuine tables and do freeze once disabled. The snapshot carries the full SCD2 history of every rating and note. The `rpt_` views add only joins to live staff-roster attributes, which are reconstructible from git history, so they are not registered as archive sources.

Note that `stg_google_sheets__performance_management__leadership_development_metrics` filters `where not disabled`, so the frozen table holds only enabled metrics; the full metric list stays in the Google Sheet.

Unrelated pre-existing entries in `scripts/bq-cleanup.sh` target older renamed Leadership Development relations in `kipptaf_performance_management` and `kipptaf_appsheet`, not the live relations above.

cristinabaldor:
# Pull Request

## Summary & Motivation

Mark the Leadership Development app, dashboard, and tables as archive-only and
stop every automated extract or refresh that this repo controls. The leader
performance-management workflow is moving to Lattice.

Closes #4627.

Uses the archive pattern documented in `src/dbt/kipptaf/CLAUDE.md`: disable the
model, register the frozen relation as a BigQuery-native source, repoint
consumers. **No relation is dropped and no warehouse DDL is run** — the frozen
tables stay queryable, so this is fully reversible.

### What was refreshing this, and what stops

| Trigger | Stops? |
| --- | --- |
| `bigquery_table_modified_sensor` polling the two AppSheet tables every 5 min | Yes — the two `google_appsheet` source entries are disabled, and the sensor builds its selection from the parsed manifest |
| `kipptaf__google__sheets__asset_sensor` polling the LD metrics Sheet | Yes — that external source entry is disabled |
| Eager dbt automation conditions across the 8 nodes | Yes — disabled models produce no Dagster assets |
| Nightly dbt Cloud prod `dbt build` | Yes |
| Dagster-driven Tableau extract refresh | N/A — this exposure never had a `cron_schedule`, so Dagster has never refreshed this workbook |

### Nodes disabled (8)

`stg_google_appsheet__leadership_development__output`,
`stg_google_appsheet__leadership_development__active_users`,
`stg_google_sheets__performance_management__leadership_development_metrics`,
`snapshot_leadership_development__output`,
`rpt_appsheet__leadership_development_roster`,
`rpt_appsheet__leadership_development_assignments`,
`rpt_tableau__leadership_development`,
`rpt_gsheets__leadership_pm_reminders`

Nothing outside this cluster consumed any of them. `int_people__leadership_crosswalk`
is a **different** model (instructional leadership crosswalk, consumed by staff
roster, DDI, topline, and gradebook audit) and is untouched, as are the
SchoolMint Grow observation models and `rpt_gsheets__pm_assignment_roster`.

### Archive of record (4 frozen tables registered as sources)

New source `google_appsheet_archive` (schema `kipptaf_google_appsheet`) holds the
two staging tables plus the snapshot; the metrics table joins the existing
`google_sheets` BigQuery-native source. Each uses `identifier` to point at the
existing relation, so no rename is needed.

The snapshot is the real archive — SCD2 history of every rating and note.

**The separate source name is load-bearing.** The schema differs from the live
`google_appsheet` source (`kipptaf_google_appsheet` vs `kipptaf_appsheet`), and
`code_locations/kipptaf/google/appsheet/assets.py` builds the specs polled by
`bigquery_table_modified_sensor` from `source_name == google_appsheet`. Reusing
that name would put the archive straight back under a 5-minute poll.

### Exposures

All four kept and relabeled `(archived)` with a description of what moved and
what still needs a human. `depends_on` is repointed from `ref()` to the archive
`source()` calls, because an exposure naming a disabled model breaks dbt parsing
**and** breaks the Dagster code location load — `dbt/assets.py` resolves exposure
refs against `manifest["nodes"]` with `next()`, which raises `StopIteration` for
a disabled node. The `dbt_snapshots` exposure covers live snapshots, so the
frozen one is simply dropped from its list.

**Where the lineage lives, precisely.** Since `refs` only populates from `ref()`,
these three exposures now have `refs == []` and their Dagster `AssetSpec.deps` are
empty — `dbt/assets.py` and `build_tableau_workbook_refresh_asset` both read
`refs`, not `sources`. Lineage is preserved in the dbt manifest and docs
(`exposures[].sources`, 3 archive sources each), **not** in the Dagster asset
graph. This is the existing behavior of the repo's archive pattern, not a
regression: the archive source asset keys are absent from all 982 keys across
every manifest-driven builder, and so is the pre-existing precedent
`stg_google_sheets__people__employee_numbers_archive`. Adding Dagster-graph
lineage for archived sources would mean changing a shared code path for all 71
exposures to depend on undefined asset keys — deliberately out of scope here.

### Not stopped by this PR — needs a human, outside the repo

1. **Tableau extract refresh.** Any refresh still happening on the Leadership
   Development workbook is a Tableau-side schedule and must be removed in
   Tableau.
2. **The AppSheet app.** It is the *writer* into BigQuery. Disabling dbt models
   stops the reads; the app keeps accepting entries until it is turned off in the
   AppSheet console.
3. **Reminder emails.** If an Apps Script on the reminders Sheet mails leaders,
   it has to be stopped in the Sheet.

**Important caveat on the four `rpt_` views:** all four resolve `int_people__staff_roster`
(three of them transitively through `rpt_appsheet__leadership_development_roster`,
and `rpt_gsheets__leadership_pm_reminders` directly), so disabling a view does
**not** freeze its output — it keeps returning current staff data on every read.
A follow-up hard cut (dropping those four views) is the only way to stop those
reads, and it will break the AppSheet app immediately, so it is deliberately left
out of this PR for the team to time. It is tracked in **#4634**, which carries the
DDL and the prerequisite checklist — warehouse DDL stays with the team rather than
being run here.

## AI Assistance

Claude Code authored this change end to end (footprint mapping, the archive
design, and the edits). Human-directed decisions: archive depth (BigQuery-native
archive sources rather than freeze-in-place or full decommission), scope boundary
(Leadership Development only, not the wider performance-management cluster), and
keeping the exposures with an archived relabel.

### General

### dbt

- [x] Properties files updated for every affected model (no new models added)
- [x] Exposures updated for all affected consumers — kept, relabeled `(archived)`,
      `depends_on` repointed to the archive sources
- [x] **Breaking change?** No columns renamed or removed. Models are disabled,
      but nothing outside the disabled cluster referenced them, verified against
      the parsed manifest. Rollback is reverting this commit; the underlying
      relations are untouched, so re-enabling rebuilds in place.
- [x] **dbt Cloud CI emits 7 new warnings — expected, and they are the point.**
      All seven are the same class: `Breaking change to contracted, unversioned
      model ... was disabled`. `staging/` and `extracts/` are contract-enforced,
      so dbt's state comparison flags each disabled contracted model as a
      potential break for downstream consumers. There are none — verified no node
      outside the cluster references any of the 8. Deliberately **not** using
      `deprecation_date` or model versions to silence these: those exist to give
      real consumers a migration window, and this workflow is being retired
      rather than evolved, with zero consumers to migrate. The warnings are the
      accurate signal that 7 contracted models went away on purpose.
- [x] External source change is a **disable**, not an add or modify, so no
      `stage_external_sources --target staging` run is needed — CI will not try to
      read it.

### Docs

- [x] No schedule or sensor definition changed, only the asset selections the two
      table/sheet sensors compute from the manifest. Both sensors are absent from
      `docs/reference/automations.md` (the generator only lists sensors with jobs
      attached), so no regeneration is needed.

### Verification performed

- `dbt parse` clean on the modified project.
- Manifest assertions (17 checks, all passing): the 8 nodes are disabled and
  present in `manifest["disabled"]`; the 3 polling source entries are gone from
  the live sources; neither sensor's computed selection mentions leadership
  development; all 4 archive sources resolve to the expected
  `project.schema.relation`; every LD exposure is relabeled, carries no `ref()`
  deps, and resolves to archive sources; the Tableau refresh asset keeps its LSID
  and gains no cron; `dbt_snapshots` keeps both live snapshots; no live node has
  a dangling dependency into the cluster; `int_people__leadership_crosswalk` is
  still live.
- Exercised the real Dagster spec-building code paths (`dbt.assets`,
  `google.appsheet.assets`, `google.sheets.assets`, `tableau.assets`) — all build,
  and no archived relation appears in either polled selection.
- Trunk clean on all 15 modified files.

## CI checks

- [ ] **Trunk**
- [ ] **dbt Cloud**
- [ ] **Dagster Cloud**


cristinabaldor:
Worked through all four. Two fixed in b0401eac0, two accepted with reasoning below.

### 2. `Refs #4627` in `description:` — fixed

The convention is real (`src/dbt/CLAUDE.md:1118-1120`) and actively followed —
every precedent I found puts the ref in a `#` YAML comment, never in description
prose. Moved all 12 occurrences across 9 files to a comment next to the relevant
`config:` block. Verified against the parsed manifest: zero descriptions in
`nodes`, `sources`, `exposures`, or `disabled` still contain a tracking ref.

### 4. Incomplete `depends_on` — fixed

Correct, and the reasoning matters more than the inconsistency: with the model
nodes gone from the DAG, the exposure `depends_on` is the only remaining lineage
record, so an incomplete list actively loses information. Confirmed both readings
of the now-disabled SQL — `rpt_tableau__leadership_development` read
`active_users` and the metrics directly, and `rpt_gsheets__leadership_pm_reminders`
read `active_users` directly. All three archived exposures now list the same three
archive sources (verified in the manifest: `sources` has 3 entries each).

### 1. Exposure `source()` repoint yields `deps=[]` in Dagster — confirmed, accepting the gap

The factual core is right, and I'd already observed the empty deps; I should have
called it out rather than treating it as unremarkable. Ran the verification you
asked for: `refs` is `[]` and `sources` is populated for all three.

Pushing back on the proposed fix, though — resolving `exposure["sources"]` in
`dbt/assets.py` would point deps at asset keys that **don't exist**. I enumerated
every asset key across all the manifest-driven builders (`core_dbt_assets`,
`adp_payroll_dbt_assets`, `google_sheet_dbt_assets`, exposure specs, appsheet,
sheets, airbyte, tableau — 982 keys) and all 4 archive source keys are absent.
They can't be present: `google/appsheet/assets.py` filters
`source_name == "google_appsheet"` (the archive is deliberately
`google_appsheet_archive`, since reusing the live name would put it back under the
5-minute `bigquery_table_modified_sensor` poll), and `google/sheets/assets.py`
filters on `source.get("external")`, which a BQ-native entry has no reason to
carry.

More decisive: the **pre-existing** archive-source precedent behaves identically.
`stg_google_sheets__people__employee_numbers_archive` — the pattern this PR copies
— is also absent from those 982 keys. So "an archived BQ-native source produces no
Dagster asset" is the established behavior of this repo's archive pattern, not
something this change introduced.

That makes the fix a change to a shared code path affecting all 71 exposures, to
add deps on undefined keys, in service of an archive-only exposure that never
materializes. Declining it as out of scope and disproportionate.

What I'd note instead, and have added to the PR body: lineage is preserved where
exposure lineage actually lives — the dbt manifest and docs, via
`exposures[].sources` — and not in the Dagster asset graph. If Dagster-graph
lineage for archived sources is wanted, that's its own change to the archive
pattern repo-wide, worth a separate issue rather than riding along here.

### 3. Relabeling changes the exposure's Dagster asset key — accepted, no action

Correct, and it's the intended trade for the relabel that was explicitly asked
for. `dbt/assets.py:65` keys off `label`, so the old catalog node is replaced by
`... (archived)`. These are non-materializing description-only specs, and the old
key stops being materialized anyway. Parentheses in a label are already in use
here (`SY25 Junior PTG Tracker (c/o 2026)`, `... (CARAT)`, `... (Schools View)`),
so the key shape isn't novel.

---

One correction to the PR body while I was in there, unrelated to your findings: I
had written "three `rpt_` views" serve live staff data. It is four —
`rpt_gsheets__leadership_pm_reminders` is also an `rpt_` view and joins
`int_people__staff_roster` directly. The PR body and the DDL handoff on #4627 both
say four now; the first commit message still says three.

src/dbt/kipptaf/models/exposures/tableau.yml
cbini, line 1: should we just disable this?

src/dbt/kipptaf/models/exposures/google-sheets.yml
cbini, line 1: should we just disable this?

src/dbt/kipptaf/models/exposures/google-appsheet.yml
cbini, line 1: should we just disable this?

cbini (changes requested):

src/dbt/kipptaf/models/exposures/tableau.yml
cristinabaldor: Done in a779509d — `config: enabled: false`, label restored, `depends_on` back to `ref("rpt_tableau__leadership_development")`.

You were right that this is the simpler shape, and it also removes a hazard the original approach was designed around. Verified:

- A disabled exposure is dropped from `manifest["exposures"]` entirely (not even in `manifest["disabled"]`). Both builders that read exposures — `kipptaf/dbt/assets.py:90` asset_specs and `kipptaf/tableau/assets.py:14` workbook_refresh_assets — iterate that dict, so the exposure produces no asset and the `next()` lookup that raises `StopIteration` on a disabled `ref()` never runs. The `ref()` calls can stay.
- Same shape as the existing `federal_grants_timesheet_approval` exposure (tableau.yml:78), which is disabled and refs a disabled model.
- After the change: `dbt parse` clean; exposure specs 71, tableau refresh assets 46, zero leadership keys in either.
- The previous approach would have left three exposure assets in the Dagster graph with `refs == []` — dead assets that never materialize. Those are gone now.

The extract-refresh note moved to a YAML comment above the exposure, since a disabled exposure's `description` never reaches dbt docs.

src/dbt/kipptaf/models/exposures/google-sheets.yml
cristinabaldor: Done in a779509d — `config: enabled: false` on `leadership_pm_reminders`, label restored, `depends_on` back to `ref("rpt_gsheets__leadership_pm_reminders")`.

Full reasoning is in the tableau.yml thread. Short version: a disabled exposure is dropped from `manifest["exposures"]`, so the `ref()` to a disabled model parses fine and no Dagster asset is built. The Apps Script note moved to a YAML comment above the exposure, since a disabled exposure's `description` never reaches dbt docs.

src/dbt/kipptaf/models/exposures/google-appsheet.yml
cristinabaldor: Done in a779509d — `config: enabled: false` on `leadership_development_app`, label restored, `depends_on` back to the three original `ref()` calls.

Full reasoning is in the tableau.yml thread. This file also drove the rest of the cleanup: the `source("google_appsheet_archive", ...)` repointing existed only to give these three exposures something to depend on, so it is all gone now. The `google_appsheet_archive` source block stays as the archive of record and is annotated registration-only — no dbt node selects from it. The AppSheet-console note moved to a YAML comment above the exposure.

cristinabaldor:
@cbini ready for you again

cbini (approved):

PR #4635, opened 2026-07-30
test(dbt): assert gpa goal coverage so an omitted row is not silent

### LINKED ISSUE #4632: test(dbt): surface missing gpa goal rows instead of dropping the grade silently
## Problem

`int_gpa__goal_aggregations` inner-joins the student spine to the goals sheet, so a grade with no goal row produces no output row at all. That join is correct and should stay — the goal row is what supplies `metric`, `threshold`, and `direction`, and it supplies `aggregation_hash`, which is the model's grain key. A left join would not yield "grade 12 with a null goal"; it would yield a row with no metric to select, nothing to compare against, and no key. The output grain is one row per goal, not one row per student grain.

The problem is not the join. The problem is that a missing goal is **silent**. No row, no null, no zero, no test failure. There is currently no test covering goal coverage at all.

Live example: grade 12 has no `y1_gpa_weighted` goal at any org level, so grade 12 is absent from every `y1_gpa_weighted` row of `rpt_tableau__gpa_goals` in both 2025 and 2026. Nothing anywhere reports that. A dashboard built on this table shows three grades and gives the reader no way to know a fourth was expected.

To be clear about blast radius, because it is narrow: the lineage is a single unbranched chain — `int_gpa__goal_student_metrics` to `int_gpa__goal_aggregations` to `rpt_tableau__gpa_goals` to the `gpa_goals_dashboard` exposure. Nothing else reads either model. Grade 12's 398 AY2025 students are fully present in the upstream spine and in `rpt_tableau__gpa_cumulative_year`. This is about one reporting table being quietly incomplete, not about students disappearing from the warehouse.

## Proposed change

Two singular tests in `src/dbt/kipptaf/tests/`. They do different jobs and want different severities.

### 1. Grain consistency (severity `error`)

Within a given `academic_year` and `metric`, the set of grade levels covered by goals must be the same at every `org_level` that has goals for that metric.

Catches the asymmetric data-entry error: NCA gets a grade 11 goal, NLH does not, and NLH's grade 11 vanishes while NCA's renders. Passes on current data, so it can go in at `error` severity today.

Deliberately keyed on `org_level` combinations that already have at least one goal row, so the absence of region-level `y1_gpa_weighted` rows (a real current state) does not fire.

### 2. Grade coverage (severity `warn`)

For each `academic_year` and `metric` that has any goal row, every grade level present in `int_gpa__goal_student_metrics` for that year should have a matching goal.

This is the one that makes the grade 12 gap visible. **It will warn immediately, on 8 rows** — org and school level, 2025 and 2026, grade 12. That is the intended signal, not a defect in the test, and it clears itself the day the rows are added.

`warn` rather than `error` because of the open question below. Scoped to metrics that already have goals so that `y1_gpa_unweighted` and `on_pace`, which have no goals by design, do not fire.

## Open question for Ops

**Is the absent grade 12 `y1_gpa_weighted` goal deliberate?** It is plausible that leadership set Y1 GPA goals for grades 9 through 11 only and holds seniors to cumulative GPA instead — grade 12 does have a `cumulative_gpa_unweighted` goal. If that is intentional, test 2 is a permanent warning, which is noise, and the better resolution is to record the intent somewhere the test can read rather than to keep warning about it.

Worth answering before this merges, because it decides whether test 2 ships as written, ships at `error` after Ops backfills, or needs an intent column in the goals sheet.

## Also in scope

Add a line to the `int_gpa__goal_aggregations` description naming the silence tradeoff explicitly. The description already says the model is goal-driven; it does not say that a missing goal is indistinguishable from a grade that does not exist.

## Out of scope

Changing the join. See the first section.

Refs #4581

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will make an omitted GPA goal row visible instead of silently deleting a grade from `rpt_tableau__gpa_goals`.

`int_gpa__goal_aggregations` inner-joins the student spine to the goals sheet, so a grade with no goal row produces no output row at all — no null, no zero, no error. Live example: grade 12 has no `y1_gpa_weighted` goal at any org level, so grade 12 is absent from every `y1_gpa_weighted` row of `rpt_tableau__gpa_goals` in both 2025 and 2026. A dashboard built on that table shows three grades and gives the reader no way to know a fourth was expected.

**The join is correct and stays.** The goal row is what supplies `metric`, `threshold`, and `direction`, and it supplies `aggregation_hash`, the model's grain key. A left join would not yield "grade 12 with a null goal" — it would yield a row with no metric to select, nothing to compare against, and no key. The output grain is one row per goal, not one row per student grain. What was missing is any signal that a row is absent, and that is what this PR adds.

Blast radius of the original gap was narrow, which is why this is a test and not a model change: the lineage is a single unbranched chain (`int_gpa__goal_student_metrics` to `int_gpa__goal_aggregations` to `rpt_tableau__gpa_goals` to the `gpa_goals_dashboard` exposure), and grade 12's 398 AY2025 students are fully present in the upstream spine and in `rpt_tableau__gpa_cumulative_year`. One reporting table was quietly incomplete; no students were missing from the warehouse.

### `int_google_sheets__gpa_goals__grade_coverage_consistent`

A grade covered by one grain must be covered by every grain that has goals for the same metric and year.

This catches the likely shape of a data-entry error. Goals are entered per school, so a skipped or fat-fingered row leaves one grain short while its siblings are complete, and only that grain's grade disappears. Grains with no goals at all for a metric are not checked, so the current absence of region-level `y1_gpa_weighted` rows does not fire.

### `int_google_sheets__gpa_goals__every_goal_aggregates`

Every goal row must produce an aggregation row.

The mirror case: a typo in `schoolid`, a misspelled region, or a `grade_low`/`grade_high` range covering nobody produces a goal that exists in the sheet, aggregates to nothing, and appears nowhere downstream — silent in exactly the same way. Scoped to academic years the student spine covers, so a goal entered ahead of its year does not fail before enrollment data exists.

Not expressible as a `relationships` test: `aggregation_hash` alone is not unique across years and metrics, so the check needs the full three-column key.

## What is deliberately not caught

A grade omitted uniformly at **every** grain. Grade 12 having no Y1 GPA goal is intentional — seniors are held to cumulative GPA, and grade 12 does carry a `cumulative_gpa_unweighted` goal. Separating that from an accident would require the intent recorded in the sheet, which does not exist today. A test demanding uniform grade coverage would be a permanent warning, so it is not included.

This is stated in the `int_gpa__goal_aggregations` description as well, so the next reader does not have to rediscover it.

## Test

Both tests pass against production, and both were verified to fire on a simulated omission rather than merely passing:

| Test | Clean run | Simulated defect |
| --- | --- | --- |
| `grade_coverage_consistent` | PASS | removing one school's grade 11 goal flags it |
| `every_goal_aggregates` | PASS | injecting a goal with a non-existent `schoolid` flags it |

```text
Concurrency: 40 threads (target='dev')
PASS int_google_sheets__gpa_goals__grade_coverage_consistent
PASS int_google_sheets__gpa_goals__every_goal_aggregates
Done. PASS=2 WARN=0 ERROR=0 SKIP=0 NO-OP=0 TOTAL=2
```

Run with `--defer --favor-state` against a freshly parsed prod manifest, so both executed against production relations rather than stale dev copies.

`severity: error` on both. They pass on current data, so neither introduces CI noise.

## Asset-check anchoring

Both set `meta.dagster.ref` to **`stg_google_sheets__gpa_goals`**, not to the model each logically describes. `int_google_sheets__gpa_goals` and `int_gpa__goal_aggregations` are both views, and a check anchored to a view does not re-run on data change — only table-materialized models re-materialize. The staging model is the only table in the chain, and it is also the thing that rematerializes when Ops edits the sheet, which is precisely when these checks need to run.

## AI Assistance

Claude Code wrote these tests. The gap was found by a human asking why the aggregation uses an inner join and pointing out that it would exclude grade 12 from reporting — the premise about blast radius turned out to be narrower than feared, but the underlying concern about silent exclusion was correct and is what this PR addresses. The human also confirmed the grade 12 omission is deliberate, which is why the originally-planned grade-completeness test was dropped in favor of the two consistency checks here.

### General

### dbt

- [x] Descriptions added to `tests/properties.yml` for both singular tests, per the singular-test description convention.
- [x] `int_gpa__goal_aggregations` description updated to name the silence tradeoff.
- [x] `severity: error` set explicitly on both — the project default is `warn`.
- [x] `dbt parse --no-partial-parse` run after editing the singular-test properties yml.
- [x] `trunk check --force` clean on all four files.
- [ ] **Breaking change?** No. Tests only, plus two description edits. No model SQL, no columns, no contracts.
- [ ] Exposure — `gpa_goals_dashboard` unchanged; no new consumer.
- [ ] External source — not applicable.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

Closes #4632
Refs #4581

### DIFF
--- a/src/dbt/kipptaf/models/gpa/intermediate/properties/int_gpa__goal_aggregations.yml
+++ b/src/dbt/kipptaf/models/gpa/intermediate/properties/int_gpa__goal_aggregations.yml
@@ -7,7 +7,11 @@ models:
       the resulting rate meets the goal. Goal-driven — a grain surfaces here
       only when it has a matching goal row; students in a grain with no goal
       simply don't aggregate into a row. One row per academic year, metric, and
-      goal aggregation key.
+      goal aggregation key. A consequence worth stating plainly — a grade
+      missing from the goals sheet is indistinguishable in this model's output
+      from a grade that has no students, since both yield no row rather than a
+      null. Two singular tests on stg_google_sheets__gpa_goals assert coverage
+      from each direction so an accidental omission is not silent.
     data_tests:
       - dbt_utils.unique_combination_of_columns:
           arguments:
--- a/src/dbt/kipptaf/tests/int_google_sheets__gpa_goals__every_goal_aggregates.sql
+++ b/src/dbt/kipptaf/tests/int_google_sheets__gpa_goals__every_goal_aggregates.sql
@@ -1,6 +1,14 @@
 with
     /* scoped to years the student spine actually covers, so a goal entered ahead
-       of its academic year does not fail before enrollment exists */
+       of its academic year does not fail before enrollment exists.
+
+       Deliberately NOT scoped by region. TODO(#4581): int_gpa__goal_student_metrics
+       excludes kipppaterson pending the gpa_term / gpa_cumulative unions, and
+       Paterson runs ES and MS only, so a Paterson-scoped goal targets a
+       population of zero and genuinely reaches no dashboard — flagging it is
+       correct, not a false positive. Revisit only if Paterson opens a high
+       school BEFORE that union lands, which would make this fire on a goal that
+       is merely early rather than wrong. */
     measured_years as (
         select distinct academic_year, from {{ ref("int_gpa__goal_student_metrics") }}
     )
--- a/src/dbt/kipptaf/tests/int_google_sheets__gpa_goals__every_goal_aggregates.sql
+++ b/src/dbt/kipptaf/tests/int_google_sheets__gpa_goals__every_goal_aggregates.sql
@@ -1,0 +1,23 @@
+with
+    /* scoped to years the student spine actually covers, so a goal entered ahead
+       of its academic year does not fail before enrollment exists */
+    measured_years as (
+        select distinct academic_year, from {{ ref("int_gpa__goal_student_metrics") }}
+    )
+
+select
+    g.academic_year,
+    g.metric,
+    g.org_level,
+    g.region,
+    g.schoolid,
+    g.grade_band,
+    g.aggregation_hash,
+from {{ ref("int_google_sheets__gpa_goals") }} as g
+inner join measured_years as my on g.academic_year = my.academic_year
+left join
+    {{ ref("int_gpa__goal_aggregations") }} as a
+    on g.academic_year = a.academic_year
+    and g.metric = a.metric
+    and g.aggregation_hash = a.aggregation_hash
+where a.aggregation_hash is null
--- a/src/dbt/kipptaf/tests/int_google_sheets__gpa_goals__grade_coverage_consistent.sql
+++ b/src/dbt/kipptaf/tests/int_google_sheets__gpa_goals__grade_coverage_consistent.sql
@@ -1,0 +1,45 @@
+with
+    goal_grades as (
+        select academic_year, metric, org_level, region, schoolid, grade_level,
+        from {{ ref("int_google_sheets__gpa_goals") }}
+        cross join unnest(generate_array(grade_low, grade_high)) as grade_level
+    ),
+
+    grains as (
+        select distinct academic_year, metric, org_level, region, schoolid,
+        from goal_grades
+    ),
+
+    metric_grades as (
+        select distinct academic_year, metric, grade_level, from goal_grades
+    ),
+
+    /* every grade any grain covers for a metric is expected at every grain that
+       has goals for that same metric */
+    expected as (
+        select
+            gr.academic_year,
+            gr.metric,
+            gr.org_level,
+            gr.region,
+            gr.schoolid,
+
+            mg.grade_level,
+        from grains as gr
+        inner join
+            metric_grades as mg
+            on gr.academic_year = mg.academic_year
+            and gr.metric = mg.metric
+    )
+
+select e.academic_year, e.metric, e.org_level, e.region, e.schoolid, e.grade_level,
+from expected as e
+left join
+    goal_grades as gg
+    on e.academic_year = gg.academic_year
+    and e.metric = gg.metric
+    and e.org_level = gg.org_level
+    and e.grade_level = gg.grade_level
+    and e.region is not distinct from gg.region
+    and e.schoolid is not distinct from gg.schoolid
+where gg.grade_level is null
--- a/src/dbt/kipptaf/tests/int_google_sheets__gpa_goals__grade_coverage_consistent.sql
+++ b/src/dbt/kipptaf/tests/int_google_sheets__gpa_goals__grade_coverage_consistent.sql
@@ -21,9 +21,10 @@ with
     ),
 
     /* pooled WITHIN org_level, never across it — a school is compared against
-       what other schools cover, not against network-level coverage. Otherwise a
-       school legitimately serving a narrower grade band than the network (a new
-       high school still adding a grade a year) reads as an omission. */
+       what other schools cover, not against network-level coverage, so a school
+       band narrower than the network's is not itself an omission. This does NOT
+       cover a school newer than its peers and still adding a grade a year; that
+       still flags, because intent is recorded nowhere the test can read. */
     org_level_grades as (
         select distinct academic_year, metric, org_level, grade_level, from goal_grades
     ),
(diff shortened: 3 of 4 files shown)


anthonygwalters:
Verified each finding against the code and the warehouse before acting. Two implemented, one left as-is with reasoning. Pushed in 1176441e0 and 934cf344d.

## 2 and 3 — both correct, both fixed

**Finding 2 (no measured-years guard).** Correct, and the inconsistency between the two tests was not deliberate. The spine holds `academic_year` 2007 through 2026, so a 2027 goal entered school-by-school would have fired this at `severity: error` mid-rollout. The same `measured_years` CTE is now on both tests.

**Finding 3 (grade coverage pooled across org levels).** Correct. `metric_grades` is now `org_level_grades`, pooled *within* `org_level` — a school is compared against what other schools cover, never against network-level coverage.

Re-verified after both changes: passes against production, and the simulated omission (one school's grade 11 row removed) still produces exactly one failing row, at the school grain only.

The verification query you asked for:

```text
org_level  region  schoolid  metric                     academic_year  bands
org        —       —         cumulative_gpa_unweighted  2025/2026      9,10,11,12
region     both    —         cumulative_gpa_unweighted  2025/2026      9,10,11,12
school     all 3   all 3     cumulative_gpa_unweighted  2025/2026      9,10,11,12
org        —       —         y1_gpa_weighted            2025/2026      9,10,11
school     all 3   all 3     y1_gpa_weighted            2025/2026      9,10,11
```

Every grain of a given metric and year shares an identical band today, so the scoping change alters no current result — it prevents a future false positive rather than fixing a live one.

One thing your finding 3 got right that I initially under-fixed: scoping by `org_level` handles school-vs-network, but a school **newer than its peers** and still adding a grade a year still flags against them. There is no way to separate that from a forgotten row without intent recorded in the sheet — the same limit as grade 12. That residual is now stated in the test description, and 934cf344d fixes an inline comment that had overclaimed the fix.

## 1 — Paterson: citation correct, conclusion is not. Leaving as-is.

Your file:line is accurate — `int_gpa__goal_student_metrics.sql:38` does carry `and sr._dbt_source_project != 'kipppaterson'` with a `TODO(#4581)`. But two facts change the conclusion:

**Paterson has no high school.** `int_extracts__student_enrollments` for Paterson, 2025 and 2026:

```text
region    school_level  academic_year  rows
Paterson  ES            2025           371
Paterson  ES            2026           543
Paterson  MS            2025           188
Paterson  MS            2026           299
```

No HS rows in either year, and zero Paterson HS rows ever in `rpt_tableau__gpa_cumulative_year`. The `school_level = 'HS'` filter on line 25 already removes every Paterson student before line 38 is reached — that exclusion is currently a no-op for this population, a guard against a future union rather than an active filter.

**The test is not misfiring today.** `select count(*) from int_google_sheets__gpa_goals where region = 'Paterson'` returns **0**; goal regions are Camden and Newark only. So the PASS in the PR body is a real pass, not a pass that happens to dodge the case.

Given that, a Paterson-scoped GPA goal entered today would target a population of zero students and genuinely reach no dashboard. Flagging it is the test working, not a false positive — that is exactly the "goal exists in the sheet, appears nowhere downstream" class the test was written for. Suppressing it via a region scope or a Paterson carve-out would suppress a true positive.

It becomes a real false positive only in one sequenced window: Paterson opens a high school **before** the `gpa_term` / `gpa_cumulative` unions land. That is now written into the test as an inline `TODO(#4581)` naming the condition to revisit, and into the description.

Taking your Assessment's second option — an explicit author decision, documented the way grade 12 already is — rather than option (a) or (b).

## On the shared-fix recommendation

Findings 1 and 2 do share the "have we ever seen data" vs "have we seen data for this grain" root, but they resolve in opposite directions: 2 wants the guard *added*, 1 wants it deliberately *not* narrowed to region. So the shared `measured_years` CTE is the common piece, and the region dimension is intentionally absent from it. Not a divergence in design so much as one axis being load-bearing and the other not.

## Note on the dbt Cloud warning check

Unrelated to your findings: the dbt MCP server is not connected in my session, so I could not run `get_job_run_error(warning_only=true)` against the passing dbt Cloud run. Flagged to the author rather than skipped silently. Low exposure — this PR adds two tests and edits descriptions, no model SQL — but it is unverified, not verified-clean.

PR #4637, opened 2026-07-30
feat(dbt): add the Finalsite contact-households and address-of-record models

### LINKED ISSUE #4613: fix: Focus student address comes from an arbitrary Finalsite household

cbini:
# Pull Request

## Summary & Motivation

...add the two `finalsite` package models that resolve a student's address of
record, so the Focus `ADDRESS` feed can stop reading an arbitrary household.

Today `stg_finalsite__contacts` flattens seven address columns off
`households[safe_offset(0)]`, and `rpt_focus__addresses` maps those onto the
Focus ADDRESS contract. Array position does not identify Finalsite's primary
household — that designation is set in the UI and is absent from every field the
API exposes. Whichever household sits at offset 0 becomes the student's address
of record. The feed is import-once with no overwrite path, so a wrong pick
becomes permanent on first import.

This PR is the package half of a two-PR change (district-first, since kipptaf
consumes these via `source()`). It adds:

- **`stg_finalsite__contact_households`** — flattens the `households` array to
  one row per (contact, household), the only place per-household addresses are
  exposed. Grain verified unique: 8,473 rows, zero duplicate pairs, zero null
  household ids. Carries the same blank-to-null and state-uppercasing
  normalization `stg_finalsite__contacts` already applies.
- **`int_finalsite__student_address_of_record`** — resolves the address from the
  student's own household linkage when that yields exactly one distinct complete
  address, falls back to their primary contact's when it does, and otherwise
  emits no address with an `ambiguous` flag. Address identity is `address_1`,
  `address_2`, `city`, `state`, `zip`; complete means `address_1`, `city`,
  `state`, and `zip` are all non-null.

The key finding behind the rule: the student's household linkage is a subset of
their primary contact's, and for every student whose own record resolves to
exactly one complete address while their primary contact's record has several,
the student's address is one of the parent's. Zero exceptions. So Finalsite
already records which of the parent's addresses applies, and records it on the
student. Anchoring on the parent instead would move the pick onto the record
carrying more competing addresses.

No behavior changes in this PR — nothing consumes these models yet. The Focus
feed switches over in the kipptaf PR (#4618), which closes #4613.

Design: `docs/superpowers/specs/2026-07-29-finalsite-address-of-record-design.md`
Plan: `docs/superpowers/plans/2026-07-29-finalsite-address-of-record.md`

Refs #4613

## AI Assistance

Both models were authored by Claude Code from the design spec, executed
task-by-task with a fresh review pass per task. Human-directed decisions along
the way: keeping the three-level `nullif(upper(trim(state)))` nesting as a
verbatim copy of the existing `stg_finalsite__contacts` line rather than
splitting it to satisfy the repo's max-one-level rule; adding
`primary_contact_phone` to the intermediate, which the design argued for but did
not plumb; and strengthening the completeness data test to a biconditional
against `address_source` after review found the originally specified version
held trivially.

## Self-review

### General

### dbt

- [ ] Include (or update) an

      Not applicable — neither model is consumed by a dashboard, analysis, or
      application. The downstream consumer is the Focus SFTP extract, whose
      exposure lives in kipptaf and is unchanged.

      summary above.

      Not a breaking change — both models are new, and no existing column is
      renamed or removed. `stg_finalsite__contacts` is untouched.

      See

      Not applicable — no external source added or modified. The new staging
      model reads the existing `finalsite.contacts` source.

## Verification

Both models were built and tested against a dev BigQuery schema through the
`kippmiami` project (the Focus consumer):

- `stg_finalsite__contact_households` — builds; grain uniqueness, both
  `not_null` tests, and an `expression_is_true` asserting the
  `is_complete_address` derivation all pass.
- `int_finalsite__student_address_of_record` — builds; `unique` and `not_null`
  on the grain, `accepted_values` on `resolution_status` and `address_source`,
  and the completeness biconditional all pass. Four unit tests cover each branch
  of the resolution rule: the student's own linkage decisive, the student's
  ambiguous but the primary contact's decisive, both ambiguous, and a student
  with no households at all.
- Resolution distribution on the unscoped model (all four regions, prospects and
  applicants included): 2,004 resolved from the student's own linkage, 56 from
  the primary contact's, 1,124 ambiguous. Every resolved row carries a complete
  address and every ambiguous row carries none. The design's narrower
  enrolled-Miami figures are validated in the kipptaf PR.

## CI checks

- [ ] **dbt Cloud** — passes. Expect a trivially-green no-op run: dbt Cloud CI
      builds only the `kipptaf` project, and this PR touches no kipptaf model
      and no kipptaf-consumed source column set, so it selects zero models.
      These models are first exercised by Dagster's dbt step.
- [ ] **Dagster Cloud** — passes or not triggered. Note this emits one
      same-named check-run per code location; all must reach a terminal
      conclusion before the deploy counts as green.

## Troubleshooting

### DIFF
--- a/src/dbt/finalsite/models/api/intermediate/int_finalsite__contacts__households.sql
+++ b/src/dbt/finalsite/models/api/intermediate/int_finalsite__contacts__households.sql
@@ -4,7 +4,7 @@ with
         -- address columns: Finalsite emits empty strings (not null) and
         -- mixed-case states. Blank -> null; uppercase the state code.
         select
-            c.id as finalsite_enrollment_id,
+            c.finalsite_enrollment_id,
 
             h.id as household_id,
             h.country,
@@ -14,7 +14,7 @@ with
             nullif(trim(h.city), '') as city,
             nullif(upper(trim(h.state)), '') as state,
             nullif(trim(h.zip), '') as zip,
-        from {{ source("finalsite", "contacts") }} as c
+        from {{ ref("stg_finalsite__contacts") }} as c
         cross join unnest(c.households) as h
     )
 
--- a/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_address_of_record.sql
+++ b/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_address_of_record.sql
@@ -99,6 +99,9 @@ select
 
     coalesce(s.address_source, 'ambiguous') as resolution_status,
 from sourced as s
+-- address_contact_id is only ever set to a contact whose candidate_count is
+-- exactly 1, so this join cannot fan out; when it is null (an unresolved
+-- address) nothing matches and the address fields stay null.
 left join address_candidates as a on s.address_contact_id = a.finalsite_enrollment_id
 left join
     {{ ref("stg_finalsite__contacts") }} as pc
--- a/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_address_of_record.sql
+++ b/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_address_of_record.sql
@@ -12,25 +12,35 @@ with
         where is_primary
     ),
 
-    address_candidates as (
-        -- One row per (contact, distinct complete address). Address identity is
-        -- the five mailing fields including address_2 — an apartment difference
-        -- is a different mailing address. country is not part of the identity
-        -- (Finalsite stores it unnormalized and no consumer reads it), so a
-        -- single min() carries it through rather than letting it split one
-        -- address into two candidates.
+    -- trunk-ignore(sqlfluff/ST03): referenced via dbt_utils.deduplicate below
+    complete_households as (
         select
             finalsite_enrollment_id,
+            household_id,
             address_1,
             address_2,
             city,
             state,
             zip,
-
-            min(country) as country,
-        from {{ ref("stg_finalsite__contact_households") }}
+            country,
+        from {{ ref("int_finalsite__contacts__households") }}
         where is_complete_address
-        group by finalsite_enrollment_id, address_1, address_2, city, state, zip
+    ),
+
+    address_candidates as (
+        -- One row per (contact, distinct complete address). Address identity is
+        -- the five mailing fields including address_2 — an apartment difference
+        -- is a different mailing address. country is NOT part of the identity,
+        -- so it comes from the canonical lowest-household_id row rather than
+        -- being aggregated across rows, which would blend values from
+        -- different households.
+        {{
+            dbt_utils.deduplicate(
+                relation="complete_households",
+                partition_by="finalsite_enrollment_id, address_1, address_2, city, state, zip",
+                order_by="household_id asc",
+            )
+        }}
     ),
 
     candidate_counts as (
--- a/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_address_of_record.sql
+++ b/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_address_of_record.sql
@@ -1,0 +1,105 @@
+with
+    student_primary_contacts as (
+        -- One row per student record. `relationships.primary` is a per-record
+        -- singleton that is true or NULL (never false), and only child/student
+        -- records carry it, so a bare `where is_primary` selects exactly the
+        -- student rows and `rel_id` is that student's Parent 1. A second primary
+        -- on one student would surface as a duplicate and fail this model's
+        -- uniqueness test, which is the intended loud failure. No SIS scoping —
+        -- receivers filter to enrolled students downstream.
+        select finalsite_enrollment_id, rel_id as primary_contact_id,
+        from {{ ref("stg_finalsite__contact_relationships") }}
+        where is_primary
+    ),
+
+    address_candidates as (
+        -- One row per (contact, distinct complete address). Address identity is
+        -- the five mailing fields including address_2 — an apartment difference
+        -- is a different mailing address. country is not part of the identity
+        -- (Finalsite stores it unnormalized and no consumer reads it), so a
+        -- single min() carries it through rather than letting it split one
+        -- address into two candidates.
+        select
+            finalsite_enrollment_id,
+            address_1,
+            address_2,
+            city,
+            state,
+            zip,
+
+            min(country) as country,
+        from {{ ref("stg_finalsite__contact_households") }}
+        where is_complete_address
+        group by finalsite_enrollment_id, address_1, address_2, city, state, zip
+    ),
+
+    candidate_counts as (
+        select finalsite_enrollment_id, count(*) as candidate_count,
+        from address_candidates
+        group by finalsite_enrollment_id
+    ),
+
+    counted as (
+        select
+            spc.finalsite_enrollment_id,
+            spc.primary_contact_id,
+
+            coalesce(sc.candidate_count, 0) as student_candidate_count,
+            coalesce(pcc.candidate_count, 0) as primary_contact_candidate_count,
+        from student_primary_contacts as spc
+        left join
+            candidate_counts as sc
+            on spc.finalsite_enrollment_id = sc.finalsite_enrollment_id
+        left join
+            candidate_counts as pcc
+            on spc.primary_contact_id = pcc.finalsite_enrollment_id
+    ),
+
+    sourced as (
+        -- The student's household linkage is a subset of their primary contact's
+        -- and is the disambiguating signal, so it is tried first. Parents carry
+        -- more household rows than students, so anchoring on the parent
+        -- unconditionally would move the pick onto the record with more
+        -- competing addresses.
+        select
+            finalsite_enrollment_id,
+            primary_contact_id,
+            student_candidate_count,
+            primary_contact_candidate_count,
+
+            case
+                when student_candidate_count = 1
+                then 'student_household'
+                when primary_contact_candidate_count = 1
+                then 'primary_contact_household'
+            end as address_source,
+            case
+                when student_candidate_count = 1
+                then finalsite_enrollment_id
+                when primary_contact_candidate_count = 1
+                then primary_contact_id
+            end as address_contact_id,
+        from counted
+    )
+
+select
+    s.finalsite_enrollment_id,
+    s.student_candidate_count,
+    s.primary_contact_candidate_count,
+    s.address_source,
+
+    a.address_1,
+    a.address_2,
+    a.city,
+    a.state,
+    a.zip,
+    a.country,
+
+    pc.phone_1_number as primary_contact_phone,
+
+    coalesce(s.address_source, 'ambiguous') as resolution_status,
+from sourced as s
+left join address_candidates as a on s.address_contact_id = a.finalsite_enrollment_id
+left join
+    {{ ref("stg_finalsite__contacts") }} as pc
+    on s.primary_contact_id = pc.finalsite_enrollment_id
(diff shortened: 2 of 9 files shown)


cbini:
Verified all four findings against the code and the warehouse. Two are actioned, one recommendation taken, two dismissed as false positives with evidence below.

## Actioned

**`not_null` on `resolution_status`** — added (`456fceff2`). Worth recording why it isn't redundant: `coalesce(s.address_source, 'ambiguous')` does make null unreachable, but dbt's `accepted_values` compiles to `where value not in (...)`, and a NULL never satisfies `not in`, so a null would pass `accepted_values` silently. The `not_null` closes that hole.

**PII tagging divergence** — retrofitted rather than deferred (`523e54dc0`). `address_1`, `address_2`, `city`, and `zip` on `stg_finalsite__contacts.yml` now carry `contains_pii: true`, matching the new models. `state` and `country` are deliberately left untagged on both sides.

Residual gap, flagged deliberately and not addressed here: `first_name`, `last_name`, `email`, `birth_date`, and the three phone columns on that same file remain untagged. That is a larger question than address parity — it turns on whether the model tags direct identifiers only or direct plus indirect — and widening the retrofit to cover it belongs in its own change.

**Fan-out comment at the final `left join`** — taken (`456fceff2`). The recommendation was correct that the invariant is non-obvious to a future editor.

## Dismissed

**Duplicated `CASE` predicates — the proposed simplification does not compile on BigQuery.** The suggestion references the `address_source` alias from within the same `SELECT` list. BigQuery has no lateral column aliases:

```text
select 1 as a, case when a = 1 then 'x' end as b
-- Error 400: Unrecognized name: a at [1:26], invalidQuery
```

Lateral column aliases work in Snowflake and DuckDB, not BigQuery. Removing the duplication would require an additional CTE whose only purpose is to make one `CASE` reference another, which costs more in readability than the duplication does. The two expressions are adjacent and share their `WHEN` predicates verbatim, so a divergent edit is visible in a single screenful. Left as-is.

**`min(country)` preferring an empty string — not reproducible, and the premise does not hold.** Measured on the built table:

| Check | Result |
| --- | --- |
| Rows with `country = ''` | 0 |
| Rows with `country is null` | 2,514 |
| Grouped addresses with more than one distinct `country` | 0 of 5,755 |

The Avro source yields NULL rather than an empty string, and `min()` ignores NULLs, so there is no case where it selects between competing values. Adding normalization to `country` would also break the deliberate parity with `stg_finalsite__contacts`, which passes `country` through unnormalized. Left as-is.

## Note on CI coverage for this PR

For anyone reading the green checks: neither model was actually built by CI. dbt Cloud builds only the `kipptaf` project and this PR touches no kipptaf model or consumed source column set, so it selected zero models. No Dagster Cloud branch deployment fired either, since the `pull_request` path filters exclude `src/dbt/**`. These models are first exercised in prod after merge.

Pre-merge validation is therefore the local dev build described in the PR body: both models built for real against a dev BigQuery schema, 6 data tests and 4 unit tests passing, grain and completeness invariants measured across 8,473 household rows.

src/dbt/finalsite/models/api/intermediate/int_finalsite__student_address_of_record.sql
```diff
@@ -0,0 +1,108 @@
...
+        -- address into two candidates.
+        select
+            finalsite_enrollment_id,
+            address_1,
+            address_2,
+            city,
+            state,
+            zip,
+
+            min(country) as country,
```
cbini: should dedupe, not blend address records

src/dbt/finalsite/models/api/staging/stg_finalsite__contact_households.sql
```diff
@@ -0,0 +1,39 @@
...
+
+            h.id as household_id,
+            h.country,
+
+            nullif(trim(h.address_1), '') as address_1,
+            nullif(trim(h.address_2), '') as address_2,
+            nullif(trim(h.city), '') as city,
+            nullif(upper(trim(h.state)), '') as state,
+            nullif(trim(h.zip), '') as zip,
+        from {{ source("finalsite", "contacts") }} as c
```
cbini: this should pull from staging contacts

src/dbt/finalsite/models/api/intermediate/int_finalsite__contacts__households.sql
cbini, line 1: rename to int?

src/dbt/finalsite/models/api/intermediate/int_finalsite__student_address_of_record.sql
cbini: Agreed, and this is the right call for a reason broader than the one column. `src/dbt/CLAUDE.md` requires a canonical **row** pick rather than per-column aggregation — "Use `first_value(... order by <pk>)` for every attribute, not separate `min()` calls" and "never `DISTINCT` when a projected column varies within the partition (`min()`, `first_value()`) — use `dbt_utils.deduplicate()`".

Replacing the `group by` plus `min(country)` with a `dbt_utils.deduplicate` over the five address-identity fields, ordered by `household_id`, so `country` comes from a canonical row instead of being blended across households.

Worth noting this also supersedes a dismissal I posted earlier in this PR. The bot flagged the same line for a different reason — that an empty-string `country` would sort ahead of a populated one — and I dismissed it on the evidence that no such row exists and no grouped address has competing values. That measurement stands, but it answered the wrong question: the problem isn't whether the blend currently picks a wrong value, it's that per-column aggregation is the wrong construct for producing one row per address. Your framing catches what mine missed.

src/dbt/finalsite/models/api/staging/stg_finalsite__contact_households.sql
cbini: Doing this, together with the rename below — they're coupled, since an intermediate should read a staging model rather than a source.

Flagging what it costs, since as written the change isn't a one-liner: `stg_finalsite__contacts` doesn't currently expose the `households` array. It projects `households[safe_offset(0)]` as scalar `address_*` columns plus `household_ids`, which is ids without addresses. So per-household addresses can't be derived from it as it stands, and making this work means adding the repeated `households` STRUCT to that model's enforced contract.

I checked that widening before committing to it:

- Exact `data_type` pulled from `INFORMATION_SCHEMA.COLUMNS` rather than hand-transcribed: `ARRAY<STRUCT<id STRING, address_1 STRING, address_2 STRING, city STRING, state STRING, zip STRING, country STRING>>`.
- Only two consumers do `select *`. The kipptaf union wrapper is uncontracted, so it just gains the column after the next prod rebuild. `stg_people__student_logins`'s `select *` is from a local CTE, and it reads only `first_name`, `last_name`, and `birth_date` from contacts. Neither breaks.
- `test_contacts_phone_normalization` on that file has a `format: sql` `expect` block enumerating every output column, so it needs `households` added with the *unnormalized* literal, since the array is a raw passthrough.

The one thing I'd have argued against on its own: ~9 models in this repo flatten a source array at the staging layer, including this model's own sibling `stg_finalsite__contact_relationships`, plus `stg_coupa__users__roles`, `stg_alchemer__survey_response__survey_data`, `stg_tableau__workbook__views`, `stg_zendesk__ticket_audits__events`, and `stg_google_directory__orgunits`. Going to intermediate is a departure from that pattern, and it leaves `stg_finalsite__contact_relationships` flattening straight from the source — so the package ends up inconsistent in the opposite direction. Happy to follow your call here; noting it so the inconsistency reads as a known trade rather than an oversight, and in case you want a follow-up to move `contact_relationships` too.

src/dbt/finalsite/models/api/intermediate/int_finalsite__contacts__households.sql
cbini: Renaming to `int_finalsite__contacts__households` and moving both files into `api/intermediate/`.

Two notes on the target name. The repo's dominant convention for a flattened child array is `<layer>_<source>__<parent>__<child>` — `stg_coupa__users__roles`, `int_deanslist__incidents__actions`, `int_focus__users__pivot` — and the name I originally shipped followed `stg_finalsite__contact_relationships`, which itself deviates from that. So this rename fixes two things at once: the layer and the naming form.

The move also drops contract enforcement, since `api/intermediate` sets only `+materialized: table` while `+contract: enforced: true` is scoped to `api/staging`. The grain uniqueness test, both `not_null` tests, and the `is_complete_address` assertion all stay, and I'm keeping `data_type` on every column even though it's no longer enforced, matching `int_finalsite__student_address_of_record`.

PR #4639, opened 2026-07-30
fix(libraries/ssh): banner timeout inversion and lost transient retry break SFTP sensors

### LINKED ISSUE #4636: fix(libraries/ssh): banner timeout inversion and lost transient retry break SFTP sensors
Two open GCP Error Reporting groups (`CPuG15Xz8J-doAE`, `CO7X67XY7MG3YA`) are the same event: the chained traceback paramiko's `Transport.run()` logs at ERROR severity, split into two groups by Error Reporting.

Affected code-location server pods: `kippnewark`, `kippcamden`, `kippmiami`. First seen 2026-05-07, still firing 2026-07-30.

## Trigger

`sftp.titank12.com` was slow to send its SSH banner. Confirmed identical on two pods on 2026-07-30:

- Camden: `ssh_titan` init 15:13:11 -> banner timeout 15:13:26 (15.0s)
- Newark: `ssh_titan` init 15:13:37 -> banner timeout 15:13:52 (15.0s)

## Root cause 1: timeout inversion

| Setting | Value | Set where |
| --- | --- | --- |
| `SSHResource.timeout` | 10s | `dagster_ssh` default; `SSH_TITAN` does not override |
| `Transport.banner_timeout` | 15s | paramiko hardcoded (`transport.py:530`); only settable via a `connect()` kwarg that `dagster_ssh` never passes |

Because `timeout` (10s) is less than `banner_timeout` (15s), `start_client()` hits its deadline first and **breaks out of its wait loop without raising** (`transport.py:695-706`). `connect()` then calls `get_remote_server_key()`, which raises `SSHException("No existing session")` (`transport.py:880`).

The transport thread is left orphaned and keeps reading for 5 more seconds before logging the ERROR traceback. This is why the Camden tick at 15:13:10 shows `skipReason: "No existing session"` and ended at 15:13:21 — before the 15:13:26 traceback was even emitted.

Net effect: a slow banner surfaces as a misleading `"No existing session"` instead of a proper banner-read failure, plus an abandoned thread that logs at ERROR after the caller gave up.

## Root cause 2: transient retry regression

`c4541aa1b` (2026-05-18) narrowed the `get_connection` retry predicate:

```python
- retry=retry_if_exception_type((SSHException, TimeoutError, socket.gaierror))
+ retry=retry_if_exception_type(
+     (NoValidConnectionsError, TimeoutError, socket.gaierror, ConnectionResetError)
+ )
```

The intent was correct — `IncompatiblePeer` / `BadHostKeyException` / `BadAuthenticationType` are deterministic config failures that should not burn 5 attempts. But those are all **subclasses** of `SSHException`, so dropping the base class also dropped every transient bare-`SSHException` case.

Verified: `isinstance(SSHException(...), (NoValidConnectionsError, TimeoutError, socket.gaierror, ConnectionResetError))` is `False`. The retry built for exactly this failure mode never fires.

## Blast radius

Only 1 of 5 SFTP sensors degrades gracefully:

| Sensor | Behavior on `SSHException` |
| --- | --- |
| `titan` | catches -> `SkipReason` |
| `iready`, `renlearn`, `edplan`, `amplify/mclass` | uncaught -> tick FAILS |

Observed impact on 2026-07-30 was bounded: one skipped tick, self-healing 10 minutes later. But `titan/person_data` carries a `FreshnessPolicy` with a 1:00am deadline and a 1-hour window, so a sustained slow-banner window would trip it, and the four uncaught sensors would hard-fail rather than skip.

## Planned fix

1. Pass `banner_timeout` / `auth_timeout` explicitly so the banner phase fails inside `start_client()` with a real exception, instead of falling through to `"No existing session"` with an orphaned thread.
2. Restore transient retry with a class-based predicate that accepts `SSHException` but excludes the deterministic subclasses. No message matching, per the tenacity convention in `src/teamster/CLAUDE.md`.
3. Demote the paramiko transport-thread traceback below ERROR. Without this, a **recovered** attempt still refiles both Error Reporting groups. Same trap already documented in `core/CLAUDE.md`: "Don't `log.exception` inside retry-wrapped helpers ... creates false-positive error groups", here coming from a vendored logger.
4. Make `iready` / `renlearn` / `edplan` / `amplify/mclass` degrade to `SkipReason` like `titan`, rather than failing the tick.

Parts 1, 2, and 4 are the reliability fix; part 3 is what actually closes the Error Reporting groups. Part 3 alone would hide a real bug.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will fix two independent defects that turned a slow SFTP banner into a failed sensor tick plus a pair of false-positive GCP Error Reporting groups.

The two open groups (`CPuG15Xz8J-doAE`, `CO7X67XY7MG3YA`) are not two bugs — they are the two halves of one chained traceback that paramiko's `Transport.run()` logs at ERROR severity, which Error Reporting filed separately. Confirmed on two pods on 2026-07-30: Camden `ssh_titan` init 15:13:11 to banner timeout 15:13:26, Newark 15:13:37 to 15:13:52. Both exactly 15.0s.

### Root cause 1: timeout inversion

`SSHClient.connect` passes `timeout` to `Transport.start_client()`, whose wait spans the banner read (`banner_timeout`, 15s) **and** the key exchange that follows (`handshake_timeout`, 15s). dagster-ssh defaults `timeout` to 10s, so it expired first — and `start_client()` breaks out of its wait loop **without raising** when its own deadline passes. `connect()` then fell through to `get_remote_server_key()`, which raises the misleading `SSHException('No existing session')` while the real negotiation kept running on an abandoned transport thread that logged its ERROR traceback seconds after the caller had given up.

That ordering is visible in the incident: the Camden tick at 15:13:10 skipped with reason `No existing session` and ended at 15:13:21 — before the 15:13:26 traceback was even emitted.

The default is now 45s, clearing paramiko's 30s negotiation ceiling. Three resources also pinned `timeout=30`, which sits exactly **on** that ceiling and would have stayed broken: `SSH_EDPLAN`, `SSH_RESOURCE_CLEVER`, `SSH_RESOURCE_ILLUMINATE`. All three now inherit the default.

### Root cause 2: lost transient retry

`c4541aa1b` narrowed `get_connection`'s tenacity predicate:

```python
- retry=retry_if_exception_type((SSHException, TimeoutError, socket.gaierror))
+ retry=retry_if_exception_type(
+     (NoValidConnectionsError, TimeoutError, socket.gaierror, ConnectionResetError)
+ )
```

The intent was correct — `IncompatiblePeer` / `BadHostKeyException` / `BadAuthenticationType` are deterministic config failures that should not burn 5 attempts. But those are all **subclasses** of `SSHException`, so dropping the base class also dropped every transient bare `SSHException`, including the banner read failure the retry existed for.

The predicate now **subtracts** the deterministic subclasses from `SSHException` instead of dropping the base class. The non-retryable set is `AuthenticationException` (which covers `BadAuthenticationType` and `PartialAuthentication`), `BadHostKeyException`, `IncompatiblePeer`, and `ProxyCommandFailure` — the latter two beyond what the original narrowing named, both also deterministic. A `stop_after_delay` ceiling bounds the sequence, since each attempt can now run longer; note it is evaluated between attempts, so the bound is soft (worst case nearer 165s), still well inside the 600s shortest sensor interval.

### Root cause 3: the ERROR-severity noise itself

Even with the retry working, a **recovered** attempt still refiles both Error Reporting groups, because `Transport.run()` logs from a background thread regardless of what the caller does. A logging filter demotes `paramiko.transport` ERROR records to WARNING. No signal is lost: `run()` also stores the exception in `saved_exception` for `start_client()` to re-raise on the calling thread, so an unrecovered failure still propagates and is logged by Dagster at the run/tick level. This is the vendored-library form of the rule already in `core/CLAUDE.md` about not logging tracebacks inside retry-wrapped helpers.

Fixing only this part would have hidden a real bug, so it ships with 1 and 2.

### Blast radius

Only 1 of 5 SFTP sensors degraded gracefully. `iready`, `renlearn`, `edplan`, and `amplify/mclass` let `SSHException` escape and **failed the tick**; `titan` alone caught it.

All five now route through `SSHResource.listdir_attr_r_or_skip()`, which triages the failure instead of blanket-catching it:

| Failure | Example | Outcome |
| --- | --- | --- |
| Transient | host slow or down, DNS blip, reset peer | already retried, then `SkipReason` — next tick picks the files up |
| Deterministic | rotated credential, changed host key, algorithm mismatch | **re-raised, tick FAILS and alerts** |

Two separate bugs sat here, both found during review of this PR:

- `except SSHException` missed `NoValidConnectionsError` / `TimeoutError` / `socket.gaierror` / `ConnectionResetError`. None are `SSHException` subclasses, and `reraise=True` re-raises the *original* type once the retry budget is spent, so a host simply being **down** failed the tick.
- Widening to the full transient net then swallowed `AuthenticationException`, `BadHostKeyException`, `IncompatiblePeer`, and `ProxyCommandFailure`. That is the worse direction: `SkipReason` is indistinguishable from "no new files" in the UI, so a rotated credential would stop ingestion **silently and indefinitely** with nothing alerting. `titan` had this behavior before this PR.

The triage is centralized rather than repeated per sensor — a per-sensor `try/except` is exactly the shape that got this wrong twice, and the logic is no longer trivial enough to copy. Each sensor is now a call plus an `isinstance` check (net -95/+55 across the five). This also closes the review's duplication finding.

## AI Assistance

Human-directed: the two Error Reporting URLs and the decision to fix all three root causes plus the four sensors.

AI-assisted: the investigation (log correlation, reading paramiko's `transport.py` / `client.py` control flow to find the silent `start_client` fall-through, bisecting the retry regression to `c4541aa1b`), the tests, and the implementation.

Findings were verified against the source rather than assumed — the exception hierarchy, the paramiko timeout defaults, and the demotion were each checked at runtime, and the "unaffected" test failures were baselined against `main` rather than dismissed.

## Self-review

### General

- [x] Run `uv run pytest tests/test_dagster_definitions.py` for any modified code location — 4 passed (`kippcamden`, `kippnewark`, `kipppaterson`, `kipptaf`). `kippmiami` fails at `dlt/focus/assets.py:12`, where `resolve_configuration` eagerly needs a `FOCUS_DB` credential that is unset in a codespace; `test_definitions_all` fails only because it includes `kippmiami`. Both are the pre-existing local limitation documented in `src/teamster/CLAUDE.md`, unrelated to SSH. (The fresh worktree also needed `dagster-dbt project prepare-and-package` first — without it every location fails on a missing `target/manifest.json`.)
- [x] New integrations follow the Library plus Config pattern — n/a, no new integrations

Note for review: the one wiring risk worth a look is the `timeout` field override on `SSHResource`, since it redeclares a field inherited from `dagster_ssh.SSHResource`. Every configured resource across `core` / `kipptaf` / `kippmiami` constructs fine under test, and the Dagster Cloud branch deploy exercises the rest.

## Testing

New regression suite: `tests/resources/test_resource_ssh_banner_timeout.py` (26 tests). It uses a loopback server that accepts the TCP connection and then stays silent, so it needs no external credentials.

Written failing first, for exactly the defects being fixed. On the 14 tests that existed at that point, 7 failed. The suite has since grown: the 3 resource-guard cases added after finding the `timeout=30` pins also fail pre-fix (`kippmiami` inherits the old 10s default), and the 5 sensor cases below fail 4-of-5 pre-fix — so against the final suite the pre-fix count is 10, not 7. (An earlier revision of this description said 7 against 17 tests, conflating the two runs.)

- `test_default_timeout_outlasts_paramiko_negotiation` reads paramiko's `banner_timeout` / `handshake_timeout` **live** rather than hardcoding 15/30, so a paramiko upgrade that raises them fails the test instead of silently reintroducing the inversion.
- `test_no_configured_ssh_resource_undercuts_the_negotiation_budget` scans every configured `SSHResource` across `core`, `kipptaf`, and `kippmiami` — raising the default is only half the guarantee when a per-resource override can undercut it, which is how the three `timeout=30` pins were caught.
- Deterministic failures are asserted to fail fast at 1 attempt, so this fix cannot undo what `c4541aa1b` set out to do.
- `test_sftp_sensor_skips_rather_than_fails_on_any_transient_failure` and `test_sftp_sensor_fails_the_tick_on_deterministic_failures` drive a **real** sensor through each exception type, asserting opposite outcomes. Both verified to catch their bug rather than pass vacuously: reverting the sensor to `except SSHException` fails 4 of the 5 transient cases, and removing the deterministic re-raise fails all 4 deterministic cases.

Also verified the **real** runtime path, not just synthetic log calls: an actual failed handshake against a silent-banner server now raises `SSHException: Error reading SSH protocol banner` (not `No existing session`) and produces **zero** paramiko records at ERROR or above.

Full SSH suites pass: `test_resource_ssh_banner_timeout.py`, `test_resource_ssh_rekey.py`, `test_ssh_paramiko_tunnel.py` (22 passed).

`test_resource_ssh_dir_mtime_propagation.py` fails on live-SFTP mtime assertions — baselined on `main` at 10 failures vs 11 here; the one delta (`littlesis`) passes on re-run on both branches, so it is live-server drift, not this change.

## CI checks

- [ ] **Trunk** — `trunk check --force` on the changed files reports no new issues locally.
- [ ] **Dagster Cloud** — shared library change, so expect all five locations to redeploy.

Refs #4636

### DIFF
--- a/src/teamster/CLAUDE.md
+++ b/src/teamster/CLAUDE.md
@@ -250,6 +250,14 @@ the affected submodule alone (e.g.
 `import teamster.code_locations.kipptaf.finalsite` or
 `teamster.code_locations.kippmiami.extracts`), not the `definitions` module.
 
+`tests/test_dagster_definitions.py` fails all 6 in a FRESH worktree — run
+`prepare-and-package` for all 5 locations first or each dies on a missing
+`target/manifest.json`. The pytest failure is a bare
+`subprocess.CalledProcessError` that hides the cause; re-run
+`dagster definitions validate -m <module>` directly to see it. With manifests,
+conftest's 1Password bootstrap resolves kipptaf's dlt credentials too, so only
+`kippmiami` (`FOCUS_DB`) and `test_definitions_all` still fail.
+
 Importing a kipptaf **asset submodule** (e.g.
 `teamster.code_locations.kipptaf.google.directory.assets`) transitively imports
 `kipptaf.dbt` and needs the dbt manifest — absent in a fresh worktree, so a unit
--- a/src/teamster/code_locations/kipptaf/resources.py
+++ b/src/teamster/code_locations/kipptaf/resources.py
@@ -89,7 +89,9 @@ SSH_RESOURCE_CLEVER = SSHResource(
     remote_port=22,
     username=EnvVar("CLEVER_SFTP_USERNAME"),
     password=EnvVar("CLEVER_SFTP_PASSWORD"),
-    timeout=30,
+    # No `timeout` override: 30s sits exactly ON paramiko's banner + handshake
+    # ceiling, which reintroduces the inversion in #4636. The `SSHResource`
+    # default (45s) already exceeds the 30s this was raised to.
 )
 
 SSH_RESOURCE_COUPA = SSHResource(
@@ -118,7 +120,9 @@ SSH_RESOURCE_ILLUMINATE = SSHResource(
     remote_port=22,
     username=EnvVar("ILLUMINATE_SFTP_USERNAME"),
     password=EnvVar("ILLUMINATE_SFTP_PASSWORD"),
-    timeout=30,
+    # No `timeout` override: 30s sits exactly ON paramiko's banner + handshake
+    # ceiling, which reintroduces the inversion in #4636. The `SSHResource`
+    # default (45s) already exceeds the 30s this was raised to.
 )
 
 SSH_RESOURCE_IDAUTO = SSHResource(
--- a/src/teamster/core/resources.py
+++ b/src/teamster/core/resources.py
@@ -119,10 +119,11 @@ SSH_EDPLAN = SSHResource(
     remote_port=22,
     username=EnvVar("EDPLAN_SFTP_USERNAME"),
     password=EnvVar("EDPLAN_SFTP_PASSWORD"),
-    # Upstream default is 10s. secureftp.easyiep.com TCP handshakes from GKE
-    # occasionally exceed that during peak periods, causing transient TimeoutError
-    # on connect. 30s tolerates those without masking a truly unreachable host.
-    timeout=30,
+    # No `timeout` override: secureftp.easyiep.com TCP handshakes from GKE
+    # occasionally exceed the old 10s upstream default during peak periods, which
+    # is why this used to set 30s. `SSHResource` now defaults to 45s, which
+    # covers those AND clears paramiko's 30s negotiation ceiling — 30s would sit
+    # exactly ON that ceiling and reintroduce the inversion in #4636.
     # secureftp.easyiep.com (GlobalSCAPE EFT 8.1) advertises only `ssh-rsa`
     # (SHA-1 RSA) host keys. paramiko 5.0 dropped ssh-rsa from defaults; opt
     # back in for this host only.
--- a/src/teamster/libraries/amplify/mclass/sftp/sensors.py
+++ b/src/teamster/libraries/amplify/mclass/sftp/sensors.py
@@ -18,9 +18,11 @@ from dagster import (
     sensor,
 )
 from dagster_shared import check
-from paramiko.ssh_exception import SSHException
 
-from teamster.libraries.ssh.resources import SSHResource
+from teamster.libraries.ssh.resources import (
+    TRANSIENT_CONNECT_EXCEPTIONS,
+    SSHResource,
+)
 
 
 def build_amplify_mclass_sftp_sensor(
@@ -68,7 +70,7 @@ def build_amplify_mclass_sftp_sensor(
                     sftp_client=sftp_client,
                     remote_dir=remote_dir_regex,
                 )
-        except SSHException as e:
+        except TRANSIENT_CONNECT_EXCEPTIONS as e:
             # `get_connection` already retried the transient cases; an
             # unreachable host is not a code error, so skip the tick instead of
             # failing it and let the next one pick the files up (#4636).
--- a/src/teamster/libraries/amplify/mclass/sftp/sensors.py
+++ b/src/teamster/libraries/amplify/mclass/sftp/sensors.py
@@ -13,10 +13,12 @@ from dagster import (
     RunRequest,
     SensorEvaluationContext,
     SensorResult,
+    SkipReason,
     define_asset_job,
     sensor,
 )
 from dagster_shared import check
+from paramiko.ssh_exception import SSHException
 
 from teamster.libraries.ssh.resources import SSHResource
 
@@ -57,14 +59,20 @@ def build_amplify_mclass_sftp_sensor(
         run_requests = []
         cursor: dict = json.loads(context.cursor or "{}")
 
-        with (
-            ssh_amplify.get_connection() as connection,
-            connection.open_sftp() as sftp_client,
-        ):
-            files = ssh_amplify.listdir_attr_r(
-                sftp_client=sftp_client,
-                remote_dir=remote_dir_regex,
-            )
+        try:
+            with (
+                ssh_amplify.get_connection() as connection,
+                connection.open_sftp() as sftp_client,
+            ):
+                files = ssh_amplify.listdir_attr_r(
+                    sftp_client=sftp_client,
+                    remote_dir=remote_dir_regex,
+                )
+        except SSHException as e:
+            # `get_connection` already retried the transient cases; an
+            # unreachable host is not a code error, so skip the tick instead of
+            # failing it and let the next one pick the files up (#4636).
+            return SkipReason(str(e))
 
         for asset in asset_selection:
             asset_identifier = asset.key.to_python_identifier()
--- a/src/teamster/libraries/amplify/mclass/sftp/sensors.py
+++ b/src/teamster/libraries/amplify/mclass/sftp/sensors.py
@@ -19,10 +19,7 @@ from dagster import (
 )
 from dagster_shared import check
 
-from teamster.libraries.ssh.resources import (
-    TRANSIENT_CONNECT_EXCEPTIONS,
-    SSHResource,
-)
+from teamster.libraries.ssh.resources import SSHResource
 
 
 def build_amplify_mclass_sftp_sensor(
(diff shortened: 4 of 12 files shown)


cbini:
Verified each finding against the code before acting. Pushed as `b9931cac1` + `93296e3ba`.

## 1 (Important) — narrow sensor `except` clauses: confirmed, fixed at the root

Confirmed the mechanism: none of `NoValidConnectionsError`, `TimeoutError`, `socket.gaierror`, `ConnectionResetError` are `SSHException` subclasses, and `reraise=True` re-raises the original type once the budget is spent. So a downed host or reset peer escaped every sensor — including `titan`, which had the gap before this PR.

Took the "export the retryable tuple" option rather than hand-copying a wider tuple into five files. `TRANSIENT_CONNECT_EXCEPTIONS` is now public in `ssh/resources.py`, the retry predicate consumes it, and all five sensors catch it — so the retry boundary and the skip boundary are the same object and cannot drift.

Added `test_sftp_sensor_skips_rather_than_fails_on_any_transient_failure`, which drives a real `edplan` sensor through each transient type. It is a genuine regression test, not a vacuous one: reverting the sensor to `except SSHException` fails 4 of its 5 cases (`host-down`, `timeout`, `connection-reset`, `dns`), and only `banner-read` passes.

## 2 (Minor) — centralize the try/except block: partly taken, rest deferred

The drift risk you identified is real and is what the shared tuple fixes — a sixth sensor copying the block now inherits the correct exception set automatically.

Not extracting a full `open_sftp_or_skip` helper, for a specific reason rather than by omission: the five sensors pass different arguments to `listdir_attr_r` (`remote_dir="Reports"`, a regex, or nothing) and a context manager cannot return a `SkipReason` in place of yielding. The realistic shape is a callback-taking helper, and each caller would still need its own `if isinstance(result, SkipReason): return result` — so it trades six lines of explicit `try` for two lines plus an indirection, without removing the per-sensor branch. Happy to do it if you disagree, but it reads as a separate refactor rather than part of this fix.

## 3 (Minor) — test setup duplication: fixed

Extracted `_paramiko_negotiation_budget()`; both tests call it.

## 4 (dignified-python) — import-time side effect: fixed

Moved `_demote_paramiko_transport_errors()` out of module scope and into `get_connection`. Still early enough — the transport thread that logs the traceback is not started until `connect()` runs — and the existing idempotency check makes the repeated call free.

## 5 (Informational) — `stop_after_delay` is a soft bound: correct, comment fixed

You are right that it is only evaluated between attempts, so an attempt starting just under the threshold still runs its full 45s. Rewrote the comment to state the true worst case (nearer 165s) and why that is still fine against the 600s shortest sensor interval.

## 6 (Very minor) — undocumented additions to the non-retryable set: fixed in the PR body

`AuthenticationException` (covering `BadAuthenticationType` and `PartialAuthentication`) and `ProxyCommandFailure` are now called out in the description, not just the code comment.

## 7 (Low confidence) — the "7 failed" count: you were right, description corrected

Your 9-10 estimate beats my 7. The 7 was accurate when I ran it, against the 14 tests that existed at that point; I then added the 3 resource-guard cases after finding the `timeout=30` pins and did not restate the number. All 3 guards fail pre-fix — `kippmiami`'s `SSH_FOCUS` sets no override, so it inherits the old 10s default, which is under the 30s budget just as the two explicit `timeout=30` pins are. With the sensor cases added, pre-fix on the final suite is 10. Description updated.

## On the environment limitation

Noted, and it cut the right way here — the one finding you could not test (#1) was the one that mattered, and reading the exception hierarchy was enough to find it. For the record on this end: 22 tests pass, `trunk check --force` reports no new issues, and definitions validate for 4/5 locations (`kippmiami` fails on an unset `FOCUS_DB` credential in a codespace, which is the pre-existing local limitation documented in `src/teamster/CLAUDE.md`, not an SSH issue).

PR #4654, opened 2026-07-30
docs: add Overgrad SFTP extracts design spec and project plan

### LINKED ISSUE #4649: feat: build Overgrad SFTP extracts for roster, GPA, SAT, and AP
Overgrad now offers SIS Sync over SFTP for student rostering, GPA, and standardized test scores. Build the outbound extracts for Newark and Camden so these three file types (plus AP, limited to the five subjects Overgrad accepts) flow nightly instead of by manual CSV upload.

Rostering is gated on the student already having a Salesforce contact id, since Overgrad's Student ID is the Salesforce contact id (`int_extracts__student_enrollments` joins `salesforce_contact_id` to `external_student_id`). Students without one cannot be sent.

Target: pipeline established and tested by mid-September 2026. Design spec to follow on the linked branch.

anthonygwalters:
# Pull Request

## Summary & Motivation

Add the Overgrad SFTP extracts design spec and project plan at
`docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md`.

Docs-only. No models, no Dagster wiring, no schema changes. The spec covers the
four planned extracts (roster, GPA, SAT, AP), the layering decision, 18
stakeholder and vendor questions, and a milestone plan targeting mid-September
2026.

Refs #4649

**Opened as a draft to host a design review.** The spec is CGibson17's work; this
PR exists so review findings can be anchored to specific lines. See the review
comments on the diff.

## AI Assistance

The spec was AI-assisted. This PR wrapper and the attached line-anchored design
review were produced by Claude Code at a reviewer's request; every claim in the
review was verified against the codebase before posting.

## Self-review

### General

N/A — docs only.

N/A — docs only. The models this spec describes are not built yet.

N/A — no schedule, sensor, or integration changes. Spec document only.

## CI checks

- [ ] **Trunk** — passes. Markdownlint is the only relevant linter for a
      docs-only change.
- [ ] **Dagster Cloud** — not triggered (docs-only, and this PR is a draft).

## Review status

Blocking items raised in review, in priority order:

1. Q12 (upsert vs create-only) gates the roster model's shape, not just the file
   count — the Aug 10–21 build window currently treats roster as unblocked.
2. The AP College Board ID crosswalk gap is an unmentioned, known silent-drop
   path that `rpt_overgrad__ap_scores` would inherit.
3. Three competing `graduation_year` definitions exist in the repo; the spec
   picks one without saying which or why.

Plus four citation corrections where the spec asserts existing-codebase facts
that do not hold. Details inline.


docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+
+AP scores are in scope as a fourth file, riding on the same test-score endpoint,
+limited to the five AP subjects Overgrad accepts.
+
+### The Salesforce gate
+
+Overgrad's `Student ID` is the **Salesforce contact id**, not `student_number`.
+This is load-bearing and already established in the codebase:
+`int_extracts__student_enrollments` joins Overgrad on
+`e.salesforce_contact_id = ovg.external_student_id`.
```
anthonygwalters: **The join quote drops its region predicate.** The actual join at `int_extracts__student_enrollments.sql:487-489` also carries `and e._dbt_source_project = ovg._dbt_source_project`.

That predicate is load-bearing here, not incidental. With separate per-region Overgrad accounts, an unscoped anti-join would treat a Newark-to-Camden transfer as already in Overgrad and never create them in Camden's account.

Since the roster is specified further down as an anti-join against `int_overgrad__students.external_student_id` alone, this is the detail most likely to get lost in implementation. Please carry the source-project predicate into the spec text in both places.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+
+The Overgrad API assets we ingest are `students`, `custom_fields`, `admissions`,
+`followings`, and `universities`. There is **no** test-score or GPA endpoint. So
+"what is already in Overgrad" is available at these grains:
+
+| File   | Comparison available                                         | Grain                             |
+| ------ | ------------------------------------------------------------ | --------------------------------- |
+| Roster | `stg_overgrad__students.external_student_id`                 | exact, per student                |
+| GPA    | `academics__unweighted_gpa` / `academics__weighted_gpa`      | current value only                |
+| SAT    | `academics__highest_sat`, `sat_superscore`, `highest_psat_*` | aggregate only, not per test date |
```
anthonygwalters: Column names are missing the `academics__` prefix. The actual columns on `stg_overgrad__students` are `academics__sat_superscore`, `academics__highest_sat`, `academics__highest_psat_nmsqt`, `academics__highest_psat_10`, and `academics__highest_psat_8_9`.

Trivial on its own, except this table is doing column-contract work.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+
+This follows the existing `rpt_powerschool__autocomm_students` pattern. The
+alternative — diffing inside each district project against its own `overgrad`
+staging models — halves join depth but duplicates the comparison logic twice per
+file type, giving eight places to keep in sync instead of four.
+
+`code_location` comes from `_dbt_source_project`, which
+`int_extracts__student_enrollments` already passes through via its
+`e.* except (...)` block. This is cleaner than the `regexp_extract` on
+`_dbt_source_relation` that `rpt_powerschool__autocomm_students` uses.
```
anthonygwalters: **This citation does not hold.** `rpt_powerschool__autocomm_students` already uses `_dbt_source_project` — line 30 of that model is `se._dbt_source_project as code_location`. There is no `regexp_extract` on `_dbt_source_relation` anywhere in the file.

The contrast being drawn is invented, which undercuts an otherwise correct choice. Suggest replacing it with: this follows the existing `rpt_powerschool__autocomm_students` pattern. That is the stronger argument anyway — matching the house pattern beats being cleaner than it.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+### Dagster wiring
+
+Mirrors `src/teamster/code_locations/kippnewark/extracts/`. Each district code
+location gets a `config/overgrad.yaml` with four asset entries, built by
+`build_bigquery_query_sftp_asset()`, plus entries in `jobs.py` and
+`schedules.py`.
+
+Newark and Camden have separate Overgrad accounts today, evidenced by separate
+API keys (`pool="overgrad_api_limit_kippnewark"`, and a distinct
+`OVERGRAD_RESOURCE` per code location). Each therefore needs its own SSH
```
anthonygwalters: **Both cited facts are wrong, though the conclusion is right.**

- `OVERGRAD_RESOURCE` is a single shared object, defined once at `src/teamster/core/resources.py:102` and imported identically by both districts' `definitions.py`. It is not distinct per code location.
- The pool name is auto-derived at `src/teamster/libraries/overgrad/assets.py:22` as an f-string over `code_location`. It is a Dagster concurrency pool, and it would be per-location even with one shared Overgrad account. It carries no information about API keys.

The real evidence is in `dagster-cloud.yaml`: the two districts reference distinct 1Password secret items (`op-overgrad-api-kippnewark` vs `op-overgrad-api-kippcamden`).

Worth correcting precisely rather than quietly, because the 8-manual-uploads plan, the two SSH resources, and Q13 all rest on this claim being true.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+#### `rpt_overgrad__students`
+
+| CSV header        | Source column        | Notes                                        |
+| ----------------- | -------------------- | -------------------------------------------- |
+| `student_id`      | `salesforce_id`      | `where salesforce_id is not null` — the gate |
+| `email`           | `student_email`      | required to create the account               |
+| `first_name`      | `student_first_name` |                                              |
+| `last_name`       | `student_last_name`  |                                              |
+| `high_school`     | `school`             | abbreviation; stable across rebrands         |
+| `graduation_year` | `graduation_year`    | `yyyy`                                       |
```
anthonygwalters: **Three competing definitions of graduation year exist in this repo. The spec should name which one it means and why** — Overgrad's match algorithm keys on it.

- `int_extracts__student_enrollments.graduation_year` resolves to `adb.graduation_year`, the Salesforce contact field (`base_powerschool__student_enrollments.sql:100`).
- `salesforce_graduation_year` in that same base model, derived from actual or expected HS graduation date.
- `rpt_powerschool__autocomm_students` derives its own: `academic_year + (13 - grade_level)`.

Picking the Salesforce-backed one is consistent with the compare-against-Salesforce principle stated elsewhere in this spec — just say so explicitly rather than leaving it to the column name.

Heads up for whoever implements this: that column's own YAML `meta.source_model` claims `stg_powerschool__studentcorefields`, which is wrong — the SQL is `adb.graduation_year`. Pre-existing repo doc bug, not this spec's, but it will mislead.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+contact field `college_match_display_gpa__c`. It is labeled `Unweighted` in
+Overgrad because the spec states the match algorithm requires an unweighted GPA.
+
+**Assumption pending Kyla and Diane.** Both the source field and the type label
+need confirmation, as does whether the value is already on a 4.0 scale. This is
+the model most likely to change after stakeholder input.
+
+#### `rpt_overgrad__test_scores`
+
+From `int_assessments__college_assessment` where `scope = 'SAT'`, joined to
```
anthonygwalters, line 244: Worth verifying before relying on this filter alone: `int_assessments__college_assessment` also unions `int_collegeboard__psat_unpivot`, and `scope` is that model's `test_type` passed straight through.

Confirm no PSAT row can carry `scope = 'SAT'`, otherwise PSAT scores leak into the SAT extract. I could not resolve the concrete `test_type` values from the model source, so flagging it to check rather than asserting a problem.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+what `int_assessments__college_assessment` already unions for
+`scope in ('ACT', 'SAT')`. This keeps the extract consistent with the "compare
+against Salesforce" principle used for GPA.
+
+**Full resend every run**, per decision. A `TODO` and a follow-up issue will
+track moving to a sent-row ledger if Overgrad rejects duplicates in practice.
+
+#### `rpt_overgrad__ap_scores`
+
+From `int_collegeboard__ap_unpivot`, joined through
```
anthonygwalters: **Missing risk: the College Board ID crosswalk gap.** `int_collegeboard__ap_unpivot` reaches `powerschool_student_number` by joining `stg_google_sheets__collegeboard__ap_id_crosswalk` on `ap_number_ap_id`. Unresolved IDs drop out silently.

This is known and recurring, not hypothetical. There is a dedicated data test for exactly this failure — `int_collegeboard__ap_unpivot__crosswalk_resolves`, documented at `src/dbt/kipptaf/tests/properties.yml:169-177` — whose description states that unresolved students have their AP scores dropped. The repo also carries a `collegeboard-ap-data-ingest-protocol` skill because these gaps keep recurring.

`rpt_overgrad__ap_scores` inherits that path, and it is independent of the Salesforce gate the spec does cover. It belongs in Schedule risks, and probably wants a row-count assertion before send.

One thing to confirm while building: that model's `deduplicate` partitions by `powerschool_student_number`, so any NULL-keyed rows from a crosswalk miss would collapse to a single row.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+`int_extracts__student_enrollments` on `powerschool_student_number` to reach
+`salesforce_id` and `school`.
+
+Two problems, both requiring a decision that is recorded here rather than
+buried:
+
+**No test date.** `int_collegeboard__ap_unpivot` carries only `admin_year`, a
+two-digit year parsed from the College Board file. Overgrad requires
```
anthonygwalters: `admin_year` on `int_collegeboard__ap_unpivot` is already a 4-digit INT64, not a two-digit year. The model applies `extract(year from parse_date(...))` during the unpivot, and also derives `academic_year` as that value minus one.

The raw College Board field is two-digit; the model has already parsed it. Minor, but it changes how the synthesized May 15 date gets written.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+  near-zero marginal cost.
+- **Q11.** AP Physics: Overgrad has one `AP Physics` against College Board's
+  four exams. This spec keeps the highest score and drops the rest. Acceptable,
+  or should Physics be excluded entirely so nothing is silently lost?
+
+### Vendor — blocks setup (`integrations@overgrad.com`)
+
+- **Q12.** Does the SIS Sync `StudentSisFile` mapping support upsert, or is it
+  fixed to create-only? The manual importer forces a choice between "Creating"
+  and "Editing" at mapping time, which suggests one SFTP mapping cannot do both.
```
anthonygwalters, line 397: **Q12 gates more than the file count.** `rpt_overgrad__students` is specified above as one row per student *to create*, built on an anti-join. If `StudentSisFile` supports upsert, the anti-join is the wrong design — you would send the full roster and let Overgrad reconcile — and that changes the model's grain, its tests, and part of the rationale for diffing in `kipptaf` at all.

So this is not just whether we need one roster file or two. It decides the shape of the roster model. See the note on the Aug 10–21 milestone row.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+The credential request is the one item that can be front-loaded, and it sits on
+the critical path from day one with a turnaround we do not control. It goes
+first.
+
+### Milestones
+
+| Week         | Work                                                                                                                                                                                | Gates                                  |
+| ------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------- |
+| Aug 3–7      | Email `integrations@overgrad.com` for SFTP credentials on both accounts, with Q12–Q18 attached. Hold the GPA and roster-scope meeting. Record decisions in this spec.               | none                                   |
+| Aug 10–21    | Build 4 network models and 8 district passthroughs. One PR per file type. `rpt_overgrad__gpas` waits on Q1–Q5; the other three do not.                                              | GPA model gated on stakeholder answers |
```
anthonygwalters: This row says only `rpt_overgrad__gpas` waits on stakeholder answers and the other three do not. Per the note on Q12, the roster model's shape depends on whether the importer supports upsert, so it is gated too.

Two ways out, either is fine:

- Gate `rpt_overgrad__students` on Q12 alongside the GPA model, and accept that two of four models wait on answers.
- Build it upsert-shaped from the start, so the anti-join is a filter you can drop rather than a structural assumption you have to unwind.

The second keeps the schedule intact and is probably the better bet, given vendor turnaround is already flagged as the top risk.

docs/superpowers/specs/2026-07-30-overgrad-sftp-extracts-design.md
```diff
@@ -0,0 +1,378 @@
...
+| Aug 24–28    | Generate sample CSVs from the built models. Manual upload of each file type in both accounts to establish header and value mappings — 8 uploads.                                    | models merged; credentials received    |
+| Aug 31–Sep 4 | Dagster wiring: SSH resources per code location, `config/overgrad.yaml`, assets, jobs, schedules at `0 4 * * *`. Dry-run extracts to `couchdrop` instead of Overgrad, inspect CSVs. | dry run needs no vendor involvement    |
+| Sep 7–11     | Flip destination to Overgrad. Live send on a small subset, verify against the Overgrad History page, then full send. Confirm row counts and spot-check students in the UI.          | mappings established                   |
+| Sep 14–15    | Monitor a full week of nightly runs. Hand off welcome-email ownership. Update docs. File follow-up issues.                                                                          | target met                             |
+
+The dry-run-to-`couchdrop` step is worth the extra day. It shows exactly what
```
anthonygwalters, line 449: Agreed, and this is the strongest part of the plan.

Confirmed feasible as written: `SSH_COUCHDROP` is already wired into both the `kippnewark` and `kippcamden` definitions, so the dry run needs no new resource and no vendor involvement.

anthonygwalters (commented):
## Design review

The design is sound and the layering choice is right. Diffing in `kipptaf` with thin district passthroughs is the correct call, and the reasoning about why (four places to keep in sync instead of eight) holds up.

Two categories of finding below.

**Three design gaps worth resolving before building:**

1. Q12 (upsert vs create-only) gates the roster model's *shape*, not just the file count. The Aug 10–21 milestone currently treats roster as unblocked.
2. The AP College Board ID crosswalk gap is a known, recurring silent-drop path that `rpt_overgrad__ap_scores` would inherit. Not mentioned anywhere in the spec.
3. Three competing `graduation_year` definitions exist in this repo; the spec picks one without saying which or why. Overgrad's match algorithm keys on it.

**Four citation corrections.** The spec asserts several facts as already established in the codebase. Four of them do not hold on checking. None invalidate the design, but they are the kind of thing that makes a reviewer distrust the parts they did not verify themselves.

### Verified sound

Checked and correct, worth stating explicitly: the Salesforce gate and `external_student_id` as a native `stg_overgrad__students` column; the absence of any GPA or test-score endpoint (assets are `students`, `custom_fields`, `admissions`, `followings`, plus `universities` at kipptaf, with `schools` commented out); all the `academics__` comparison columns; the district passthrough shape, which matches `rpt_powerschool__autocomm_students` exactly; `contract enforced: true` on district extracts; `build_bigquery_query_sftp_asset()` at `src/teamster/libraries/extracts/assets.py:164`; the `config/[source].yaml` layout; `0 3 * * *` as the current extract cron; the SAT score types and the legacy-type exclusion matching `rpt_gsheets__college_assessments_long`; and the absence of any exam date in the College Board file.

### On the stakeholder questions

Calling out Q16 (delivery feedback) as the one to push hardest on is the right instinct. I would add that Q9 (welcome emails) is the one most likely to get nodded at and then owned by nobody. It is the only item in the plan with no engineering artifact attached to it, which is exactly how it ends up unassigned.

Refs #4649

CGibson17:
## Review response

All 11 threads verified against the codebase before acting. Fixes are in `fa7be6ff6` and `e8a757b93`.

### Accepted and fixed

| Thread | Resolution |
| --- | --- |
| Region predicate on the join | Corrected. The join quote now shows both predicates with the `:487-489` citation, and the roster anti-join is specified on `external_student_id` **and** `_dbt_source_project`. The transfer-suppression failure mode is written into the spec. |
| `academics__` prefix | Corrected on all five SAT comparison columns. |
| `autocomm` uses `_dbt_source_project` | Confirmed, my citation was wrong. Replaced with the house-pattern citation at `:30`. Root cause below. |
| `OVERGRAD_RESOURCE` and pool name | Confirmed, both citations were wrong. Replaced with the distinct 1Password items in `dagster-cloud.yaml` plus `.k8s/1password/items.yaml`, and added an explicit note that the shared resource object and the auto-derived pool name are **not** evidence, so nobody re-derives the claim from them later. |
| Three `graduation_year` definitions | Named the Salesforce-backed one explicitly with the reasoning, listed the two alternatives, and recorded the wrong `meta.source_model` as an implementation warning. |
| AP crosswalk gap | Added as a documented failure path and a schedule risk, with the test citation. Also drove a source change, below. |
| `admin_year` is 4-digit `INT64` | Corrected. |
| Q12 gates roster shape | Agreed. Took the build-upsert-shaped option so Q12 changes a filter rather than the grain. Q12 is now marked highest-priority vendor question, the model inventory row and the Aug 10-21 milestone gate are both updated. |
| Aug 10-21 milestone row | Updated per the above. |

### One pushback

The PSAT leak flag resolves as a non-issue, and the spec now documents why so it does not get re-raised. `test_type` on the PSAT branch is a closed `case` in `stg_collegeboard__psat.sql` over the Dagster partition key, yielding only `PSAT NMSQT`, `PSAT 8/9`, or `PSAT10` — never `SAT`. The only hardcoded `'SAT' as test_type` in the project is `int_collegeboard__sat_unpivot.sql:77`. So `scope = 'SAT'` cannot admit a PSAT row.

Flagging it as unverified rather than asserting a problem was the right call — it just happens to be answerable.

### Your AP comment changed the design

Following the crosswalk thread led to `int_assessments__ap_assessments`, and `rpt_overgrad__ap_scores` now sources from it instead. It wraps `int_collegeboard__ap_unpivot` for 2018+, unions Salesforce AP history for earlier years, and applies `where powerschool_student_number is not null` — discarding crosswalk misses rather than letting `deduplicate` collapse NULL-keyed rows into one arbitrary survivor, which was your point.

Three consequences are documented on the model:

- It carries no `admin_year`, so the synthesized date becomes `date(academic_year + 1, 5, 15)`. Equivalent for the CB branch, which derives `academic_year` as `admin_year - 1`. The Salesforce branch's convention needs confirming before we rely on it.
- Its `rn_highest` partitions by `(powerschool_student_number, ap_course_name)` across all years, so it must not be reused for the per-date dedupe — a retake in a later year is legitimately two Overgrad records.
- `exam_score` replaces `exam_grade`, and the two union branches use different subject vocabularies, so the five-value mapping has to cover both or pre-2018 scores fall through silently.

### Root cause of my two bad citations

The `regexp_extract` contrast described code that existed in my working checkout but not on `main` — `61ef3970d refactor(dbt): dedupe source-project regexp derivations` had already replaced it. I read a stale branch rather than the spec branch's actual base.

The API-key error was worse and not excusable that way: I inferred per-district keys from a hardcoded pool string and a code comment without opening `core/resources.py`. The conclusion happened to be right, which is exactly why the bad evidence needed correcting rather than quietly dropping.

### Note on the two line citations in the review

Substance was right in both cases, line numbers drifted: the crosswalk test is at `src/dbt/kipptaf/tests/properties.yml:228-241`, not `:169-177`. No action needed, just so the spec cites the right place.

One item deliberately left open: the `academic_year` convention on the Salesforce AP branch is marked confirm-while-building rather than resolved, since it needs a data check rather than a code read.

PR #4656, opened 2026-07-31
feat(dbt): uniform identity, location and entity contract across the 13 gated tableau extracts

### LINKED ISSUE #4638: feat(tableau): rebuild RLS permissions calcs for the Entra ID identity migration
## Context

KIPP Tableau Server is migrating identities from Active Directory usernames to email addresses. Tableau `USERNAME()` will return `mail` after cutover, breaking every row-level-security calculated field that compares it against a `samaccountname`-style column.

An external audit surfaced 85 flagged content items network-wide. 21 belong to cbaldor; 8 are already tagged `entra-broken-accepted`, leaving **13 Production workbooks**. Tracked in Asana (due 2026-08-07).

The identity break is the trigger, not the whole problem. The flagged calcs also carry:

1. **Stale org model** — retired locations and groups referenced, 5 active locations missing entirely, and no `KIPP Paterson` branch in any entity gate, so 96 active staff resolve to deny regardless of group membership.
2. **Hardcoded individuals** — 12 individual usernames granting access directly, bypassing groups.
3. **Drifted role gates** — six different DSO/SL/AP variants, including a dead job title matching zero staff and a title predicate that hides all 9 Miami ESE teachers from their APs.
4. **A wrong user function** — Federal Grants Timesheet Approval calls `USERATTRIBUTE('username')` instead of `USERNAME()`. Since `USERATTRIBUTE` returns NULL without a Connected App supplying the attribute, that field has likely been inert.

## Design

Full design doc: `docs/superpowers/specs/2026-07-30-tableau-rls-entra-migration-design.md`

Tableau groups remain the access mechanism. The calc is restructured into five tiers driven by standardized extract columns. `dim_staff_cube_access` (#4269) informs the structure — separating location and department axes — but is not a dependency; it keys on `google_email`, which differs from `mail` for 1,460 of 1,627 active staff.

## Scope

**dbt** — standardize location and entity field names and values across the models backing the 13 workbooks, and add the identity columns the dual-match needs:

```text
rpt_tableau__leadership_development
rpt_tableau__schoolmint_grow_observation_details
rpt_tableau__schoolmint_grow_goals
rpt_tableau__teacher_observations
rpt_tableau__pm_outlier_detection
rpt_tableau__content_team
rpt_tableau__survey_responses
rpt_tableau__survey_completion
rpt_tableau__grants_timesheets
rpt_tableau__stipend_and_bonus_app
rpt_tableau__operations_pm
rpt_tableau__operations_ekg
rpt_tableau__manager_survey_details   (new)
```

Plus a new `rpt_tableau__manager_survey_details` extract so the two manager survey workbooks stop pointing at an intermediate model, exposures added for `teacher_goals` / `content_team_dashboard` / `manager_survey_rollup`, and `manager_survey_reports` repointed.

**Tableau** — one canonical Permissions block pasted across all 13, including Federal Grants Timesheet Approval once its `USERATTRIBUTE` call is corrected to `USERNAME()`. Edits are manual in Desktop; the Tableau MCP is read-only, with no tag write and no publish.

## Decisions

| Item | Decision |
| --- | --- |
| Identity attribute | `mail`, dual-matched with `sam_account_name` through cutover |
| UPN hedge | Also match `user_principal_name` (differs from `mail` for 3 staff) |
| Individual grants | Gut entirely, no replacement |
| Location value source | `int_people__location_crosswalk.location_clean_name`, unchanged |
| Entity value source | `int_people__staff_roster.home_business_unit_name`, never `location_region` |
| Entity gate semantics | Reads group membership, not the viewer's own entity, so cross-entity supervision needs no calc change |
| Paterson entity group | `KNJ-SG-Tableau All Staff Paterson` |
| Group naming rule | `KNJ-SG-Tableau All Staff ` plus exact location name |
| Rooms | Excluded from location scoping; central office is department-based |
| AP role predicate | `job_function`, with a NULL fallback to `job_title` for 25 affected teachers |
| Tier 4 on norming and pulse blocks | Left ungated; cross-region norming is deliberate |
| Removed | Syndicate from all-access, 3 campus groups, 2 legacy Group Staff groups |
| Kept | `Group Staff NJ Regional`, `Group Staff MIA Regional` |

## Blocking pre-check

All 12 formerly-hardcoded individuals need their group coverage confirmed before any edit lands. Gutting an individual grant assumes the person is covered by a group; if any were hardcoded because they were not, gutting silently removes their access. Tableau group membership is not readable through the MCP, so this is a manual check.

## Out of scope

- The 64 flagged items owned by other authors.
- Production/Coaching Conversation Tool ownership has changed and may leave this list.
- Renaming canonical clean names upstream, which would ripple into 33 consuming models and three external lookup sheets. Tracked as a follow-up.
- Populating `job_function` for the 25 unclassified teachers, and the 5 staff whose `job_function` of Teacher conflicts with an Assistant School Leader or Dean title. Both are People Operations data fixes.

cristinabaldor:
# Pull Request

## Summary & Motivation

When merged, this pull request will give all 13 permission-gated Tableau
extracts an identical set of eleven identity, location, and entity columns under
identical names, so one Permissions calculated field works across every workbook
unmodified — before and after the Entra ID identity cutover.

Tableau Server is migrating identities from AD usernames to email. After cutover
`USERNAME()` returns `mail`, which breaks every row-level-security calc that
compares it against a `samaccountname`-style column. An external audit flagged
85 items network-wide; 21 belong to this author, 8 are already tagged
`entra-broken-accepted`, leaving 13 Production workbooks.

The identity break was the trigger, not the whole problem. Reading the flagged
calcs surfaced three compounding defects, all addressed here on the data side:

1. **Identity** — extracts carried no email column to dual-match against.
1. **Stale org model** — `KIPP Paterson` appeared in no entity gate anywhere, so
   96 active staff resolved to deny regardless of group membership.
1. **Hardcoded individuals** — 12 usernames granted access directly, bypassing
   groups.

Two bugs were found and fixed along the way that were already live in
production:

- **`schoolmint_grow_observation_details` was emitting 276,048 un-normalized
  entity rows** (`TEAM` 189,391, `KCNA` 79,525, `MIA` 7,132) and its
  `accepted_values` test was failing. Any entity gate on that data was matching
  nothing. All 13 views now normalize entity with a `case` in the select list.
- **`operations_ekg` left 235 rows ungated** — one respondent whose address
  resolves on `sam_account_name` rather than `google_email`. Those rows were
  invisible to every location- and entity-gated viewer, including that person's
  own School Leader. Now 0, with all 32 respondents resolving.

### What is not in this PR

The Tableau-side work is human-executed in Desktop and tracked separately: 13
Permissions blocks, gutting the individual grants, and repointing the two
Manager Survey workbooks at the new extract. The design doc
(`docs/superpowers/specs/2026-07-30-tableau-rls-entra-migration-design.md`) is
the hand-off artifact and has been synced to what actually shipped.

### Rollback plan

Every model change is a view definition — no rematerialization, no backfill.
Reverting the branch restores the prior column sets on the next Dagster run. The
one item not reversible by revert is the two Manager Survey workbook data
sources, which are not repointed until after merge, so a pre-repoint revert is
clean.

## AI Assistance

Claude Code authored the dbt changes under close human direction.
Human-directed: every design decision — `mail` as the identity attribute,
gutting individual grants, keeping Tableau groups rather than moving row-level
decisions into the data, real source column names with no aliasing, deleting
every legacy alias, normalizing entity with a `case` per view rather than a
shared crosswalk model, skipping the `teacher_goals` exposure, and the two
per-model carve-out rulings (`operations_pm` keeps `respondent_job_title`;
`grants_timesheets` loses its four duplicates). AI-assisted: the SQL and YAML
edits, the verification queries, and the review passes.

Each of the 18 tasks was implemented by one subagent and reviewed at depth by a
second, followed by a whole-branch review. That process caught defects worth
naming, because they are the kind a single pass misses:

- A **dormant duplication bug** in `pm_outlier_detection`: widening an
  aggregate's `GROUP BY` by seven columns split three teachers' history rows.
  Row counts matched production exactly (4,733 and 3,326 distinct), so it looked
  clean — the splits are invisible only because that upstream holds 2023-25 and
  the affected rows are 2018. Redesigned to resolve the teacher's attributes at
  an outer join, making the failure structurally impossible rather than dormant.
- A **stale-dev-shadow measurement error**: a diagnosis was 25x wrong because
  `--defer` silently preferred a 20-day-old personal dev copy of the roster over
  production. Every subsequent command uses `--favor-state`.
- A **test that could never pass**, added on a column that legitimately is not
  unique. Replaced with a composite test on the real grain.
- `manager_survey_details` emitting the **gated person's own contract values a
  second time** under `subject_*` names — a name-collision check had passed
  because the duplication was in the values, not the names.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models —
      `rpt_tableau__manager_survey_details.yml` added
      `manager_survey_reports` repointed off the intermediate;
      `manager_survey_rollup` and `content_team_dashboard` added with verified
      workbook LUIDs. Both omit `cron_schedule` deliberately, since one there
      becomes a real Dagster refresh schedule and neither workbook is
      Dagster-refreshed today.
- [x] **Breaking change?** Yes, deliberately — see below.
      n/a, no external sources touched.

#### Breaking change detail

Renames are the point of this PR, not a side effect: uniform names are what let
one Permissions block work everywhere. Every rename breaks the Tableau field
references in the affected workbook, which is expected and accepted, and the
affected workbooks are enumerated in the plan's appendix.

Deleted or renamed, by model: `entity`, `` `location` ``, `department`,
`business_unit`, `region`, `legal_entity`, singular
`report_to_sam_account_name`, `manager_email`, `respondent_location`,
`respondent_job_title` (in `operations_ekg` only), `teacher_location`,
`teacher_department`, `teacher_job_title`, `home_work_location_name` as an
output column, four `respondent_*` duplicates in `grants_timesheets`, and seven
`subject_*` duplicates in the new `manager_survey_details`.

Coordination: all 13 workbooks are owned by this PR's author, and the Desktop
edits are sequenced after merge. No other team consumes these extracts — the
only dbt-side consumers are `rpt_appsheet__seat_tracker_roster` and
`rpt_gsheets__leadership_pm_reminders`, both built and confirmed unaffected.

### Verification

Full build over the 13 models plus 3 new tests, with `--favor-state` so every
upstream resolves to production:

```text
PASS=50 WARN=8 ERROR=0 TOTAL=58
```

- **Cross-model uniformity proven, not asserted.** `INFORMATION_SCHEMA.COLUMNS`
  over all 13 built relations: the eleven names present in all 13, all `STRING`,
  contiguous, in the mandated order in every model, no near-miss variants.
- **Zero fan-out.** Every one of the 13 matches its deployed production view
  row-for-row.
- **8 remaining warns are all `not_null`**, and all are roster-join misses rather
  than unmapped locations — proven by the absence of any row where a location is
  null while a roster-sourced column is populated. Shares:
  `teacher_observations` 0.55%, `leadership_development` 0.11%, `content_team`
  1.60%, `stipend_and_bonus_app` 0.01%, all others zero.
- **Three new invariant tests**, all passing against production: 0 unresolved
  active-staff locations, the active location set matches the expected 30
  exactly, and 0 staff holding a `sam_account_name` without a `mail`. These
  catch the failure this rebuild exists to eliminate — a location that resolves
  to no Tableau group gates its rows to nobody rather than erroring.
- **No PII** in any committed file; the 12 formerly-hardcoded usernames are
  referenced by count only.

## CI checks

- [ ] **Trunk** — `trunk check --force --no-fix` is clean locally on all changed
      files
- [ ] **dbt Cloud**
- [ ] **Dagster Cloud** — expected to be skipped, no Dagster paths changed

Refs #4638


cristinabaldor:
@claude update your review

cristinabaldor:
Verified each finding against the code before acting. Four fixed, one disputed with evidence, one out of scope. Pushed as `5569bd6c9`, `739a326d7`, `bcc70d281`, `2b5ab32bc`.

## Fixed

**1. `pm_outlier_detection` score_aggs grain — correct, and the fix went further than suggested.** Confirmed: the `score_aggs` join carried no `primary_indicator`, its `GROUP BY` held five roster columns, and three of them were dead. Rather than only deleting the dead three, I collapsed the aggregate entirely — `teacher_name` and `teacher_manager` now come from the `trh` outer join (which does filter `primary_indicator`), so `score_aggs` groups by exactly four columns with no roster columns at all. `overall_score` is now one true average per key instead of one per roster-attribute combination.

This changed row counts, in the right direction, and I verified it is a pure merge rather than any loss:

```text
                          new     old (prod)
rows                     4607           4733
distinct grain keys      3326           3326
keys lost                   0
keys gained                 0
keys increased              0
keys decreased             78
rows removed              126
```

So 78 grain keys were carrying duplicate rows from mid-window roster changes, double-counting those teachers' scores in any viz that aggregates. Zero keys disappeared and none split.

You were right that the PR body's "structurally impossible" claim was too strong — it held for the eleven contract columns, which the earlier redesign moved to `trh`, but not for `teacher_name` / `teacher_manager` / `overall_score`, which still rode the widened grain. The claim is accurate now.

**4. `survey_responses` duplicated alias — correct.** `eh.reports_to_formatted_name` was selected twice, as `manager` (line 21) and `manager_name` (line 24). Deleted `manager`, kept `manager_name` for parallelism with `respondent_name`, and removed its properties entry.

**3. `operations_ekg` OR-join — fair, and made permanent.** The concern is right that the code alone cannot prove two independently-sourced identity columns always agree on one person. Empirically there is no duplication today: `count(*)` = 3577 and `count(distinct (response_id, question_id))` = 3577. Rather than rely on that, a `dbt_utils.unique_combination_of_columns` on `(response_id, question_id)` now guards it permanently, so an OR-join that ever matched two roster rows would fail the test instead of silently duplicating.

**Minor — 33 missing column descriptions on `manager_survey_details`.** Correct, and a real gap: the implementer copied `data_type` from the upstream but not `description`. All 38 columns now carry one, with the 27 inherited descriptions taken verbatim from `int_surveys__manager_survey_details` rather than reworded.

## Disputed

**5. ST06 suppression comments — this one is a false positive, and acting on it would break CI.** The factual half is right: 3 of 13 models lack the comment. The conclusion that they "will likely trip sqlfluff ST06 at `trunk check --force`/CI" is not:

```text
trunk check --force --no-fix rpt_tableau__leadership_development.sql \
  rpt_tableau__schoolmint_grow_observation_details.sql \
  rpt_tableau__teacher_observations.sql
→ Checked 3 files, No issues
```

ST06 does not fire in those three, and the Trunk Check on this PR passed over all 33 changed files. Adding a suppression where nothing fires trips `trunk/ignore-does-nothing`, which is itself a CI failure — this repo's own CLAUDE.md documents that. So the three files are correct as they stand, and the suppression is present in exactly the ten models where the rule actually fires. Not changing them.

## Out of scope

**2. Missing `primary_indicator` in `content_team`, `schoolmint_grow_goals`, and `schoolmint_grow_observation_details`.** The gap is real and I am not disputing the risk. But those join lines are pre-existing — `git diff` against the merge base shows this branch never touched any of them:

```text
git diff <merge-base>...HEAD -- <each of the three files> | grep 'staff_roster_history|primary_indicator|assignment_status'
→ no join-line changes on this branch
```

Adding the filter would change row counts on three production Tableau feeds, in a PR whose purpose is an identity-column migration. That deserves its own change with its own before/after verification rather than riding along here. Filing it separately.

Also leaving alone, both of which you correctly identified as pre-existing rather than introduced here: the missing `url:` on the new exposures (only 3 of the file's 47 carry one, so the new ones follow local precedent), and the absence of model-level uniqueness tests on 11 of the 13.

**Minor — test severity.** The critique lands: two tests describe their failure mode as a launch risk while sitting at the project's default `warn`. I am keeping `warn` deliberately rather than promoting to `error`, because the network opens campuses on purpose and `error` would block every dbt run until someone updated the expected set — turning a routine school launch into a broken pipeline. The rationale is now written into the test descriptions so the choice is explicit rather than inherited by default.

## Note on one thing you flagged in passing

The residual duplication you would find in `pm_outlier_detection` after this fix (~1281 rows against 3326 distinct keys) is a separate pre-existing issue in `score_dates`, where `stg_google_sheets__reporting__terms` holds multiple rows matching the same code/type/name filter. It reproduces identically on production and is unrelated to `score_aggs`. Not fixed here; it goes in the same follow-up as finding 2.

GabyRangelB (dismissed):

cristinabaldor:
@claude update your review

cbini (approved):

PR #4667, opened 2026-07-31
fix(clever): scope all six extract feeds to Newark, Camden and Paterson

### LINKED ISSUE #4653: fix(clever): Paterson and Miami leaks make the six extract feeds region-inconsistent
## Summary

`kipptaf__extracts__clever__asset_job_schedule` (hourly) is **STOPPED** in prod.
Before restarting it, the six Clever feeds need to agree on which regions they
cover. Today they do not: `students.csv`, `schools.csv` and `staff.csv` omit
regions that `enrollments.csv`, `sections.csv` and `teachers.csv` still
reference, so a live sync would push orphaned rows.

Counts below are prod `kipptaf_extracts.rpt_clever__*` as of 2026-07-30.

| feed          | Newark | Camden | Miami | Paterson | not in `schools.csv`     |
| ------------- | -----: | -----: | ----: | -------: | ------------------------ |
| `students`    | 24,927 |  5,934 |     0 |        0 | 0                        |
| `enrollments` | 24,024 |  5,865 | 1,114 |        0 | 0                        |
| `schools`     |     12 |      5 |     2 |        0 | n/a (+1 District Office) |
| `sections`    |  1,563 |    511 |    10 |        9 | 0                        |
| `teachers`    |    834 |    271 |    97 |       88 | 37                       |
| `staff`       |  3,520 |    766 |   308 |        0 | 37                       |

## Problem 1: Paterson is deliberately excluded, but leaks into two feeds

Paterson PowerSchool is live and healthy on the dlt-synced instance: 842 AY2026
students with `rn_year = 1` in `int_extracts__student_enrollments`, and ~88
Paterson staff now flow through `int_people__staff_roster` since 5dc08de3d
removed the roster spine exclusion.

Clever still gates Paterson out per the 2026-06-22 staff-propagation design
(all six feeds drop the exclusion together when Paterson's SIS is ready). Four
feeds honor that. Two do not:

- **`rpt_clever__teachers`** has no Paterson filter at all. It reads
  `int_people__staff_roster` and `int_people__temp_staff` unfiltered, so 88
  Paterson teacher rows now ship, pointing at `school_id` 2 and 1234, which
  `schools.csv` omits. This regressed silently when the roster spine exclusion
  was removed.
- **`rpt_clever__sections`** filters Paterson in the PowerSchool branch
  (`_dbt_source_relation not like '%kipppaterson%'`) but not in the
  auto-generated ENR branch. That branch reads `dsos` off the roster and
  cross-joins `stg_powerschool__schools` with no source-relation filter, so 9
  Paterson ENR sections ship (4 at `school_id` 2, 5 at 1234), again referencing
  schools that `schools.csv` omits.

**Decision needed:** either finish the gate (add the two missing filters) or
lift it network-wide (drop the exclusion from all six). #4193, the stated gate,
is an attendance-conversion-item config gap that affects ADA only, not
enrollment or rostering, so it may no longer be the right blocker for Clever.

## Problem 2: Miami students silently disappeared, the rest of Miami did not

Miami has migrated to Focus. `kippmiami_powerschool` is frozen at AY2025:

| region  | AY2025 eligible | AY2026 eligible |
| ------- | --------------: | --------------: |
| Newark  |           5,590 |           6,808 |
| Camden  |           1,832 |           2,142 |
| Miami   |           1,114 |         **0**   |

`rpt_clever__students` filters `academic_year = current_academic_year` (2026),
so Miami drops to zero. Nothing else does:

- **`rpt_clever__enrollments`** reads `stg_powerschool__students` with only
  `enroll_status in (0, -1)` and no academic-year filter, so it ships 1,114
  stale AY2025 Miami rows, exactly the frozen count. Every one references a
  `student_id` absent from `students.csv`.
- **`rpt_clever__sections`** ships 10 Miami sections, `schools.csv` ships 2
  Miami schools, `staff.csv` 308 and `teachers.csv` 97 Miami rows.
- 37 staff rows and 37 teacher rows point at Miami `school_id` 30200805,
  30200806 and 30200807, which are not in `stg_powerschool__schools` and so are
  missing from `schools.csv` entirely.

**Decision needed:** is Miami supposed to roster through Clever at all now that
it is on Focus? If yes, the Miami side of all six feeds has to be re-sourced
from the `focus` project. If no, Miami should be filtered out of all six the
way Paterson nominally is. Either way the current state, live staff and teachers
against zero students, is wrong.

## Recommendation

Do not restart the schedule until both are resolved. Suggested order:

1. Confirm with Ops and Miami whether Clever is still the rostering path for
   Miami post-Focus.
2. Confirm whether Paterson should go live in Clever now that its dlt sync and
   staff roster are healthy.
3. Apply the resulting region filters consistently across all six models, and
   add a test asserting that every `school_id` in `enrollments`, `sections`,
   `teachers` and `staff` exists in `schools`, and every `student_id` in
   `enrollments` exists in `students`. That test is what would have caught both
   of these.
4. Restart `kipptaf__extracts__clever__asset_job_schedule`.

## Verification queries

```sql
-- rows per feed per region
with sch as (
  select cast(school_number as string) as school_id,
         regexp_extract(_dbt_source_relation, r'(kipp\w+)_') as loc
  from `teamster-332318.kipptaf_powerschool.stg_powerschool__schools`
  group by 1, 2
)
select coalesce(sch.loc, 'not in schools.csv') as loc, count(*) as n
from `teamster-332318.kipptaf_extracts.rpt_clever__teachers` as t
left join sch on cast(t.school_id as string) = sch.school_id
group by 1
order by 1;
```

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will make all six `rpt_clever__*` extract feeds
agree on which regions they cover — Newark, Camden and Paterson in, Miami out —
and add tests that fail if they ever disagree again.

Each feed carried its own region filter, so they diverged and shipped orphaned
rows. Per the decisions recorded on #4653, the filters invert: Miami rosters into
Clever directly from Focus so it leaves all six feeds; Paterson's dlt sync and
staff roster are healthy so its gate comes off.

Feed contents before and after, from a local dev build against live prod
upstreams:

| feed          | Newark | Camden |    Miami |     Paterson |
| ------------- | -----: | -----: | -------: | -----------: |
| `students`    | 24,915 |  5,942 | 0 (same) | 0 to 3,016   |
| `enrollments` | 27,505 | 10,974 | 1,114 to 0 | 0 to 4,796 |
| `schools`     |     12 |      5 | 2 to 0   | 0 to 2       |
| `sections`    |  1,687 |    565 | 10 to 0  | 9 (same)     |
| `teachers`    |    834 |    271 | 97 to 0  | 90 (same)    |
| `staff`       |  3,436 |    740 | 308 to 0 | 0 to 240     |

Paterson's `school_number` 2 and 1234 both carry
`state_excludefromreporting = 0`, so they appear in `schools.csv` and Paterson's
existing teacher and section rows now resolve. The Miami ids `30200805`,
`30200806` and `30200807` that 35 staff and 35 teacher rows pointed at do not
exist in `stg_powerschool__schools` at all; the Miami exclusion clears them.

Also in this change:

- `rpt_clever__students`: collapses the now-dead
  `if(region = 'Miami', fleid, state_studentnumber)` branch.
- `rpt_clever__sections`: the auto-generated ENR branch joined
  `stg_powerschool__schools` on `school_number` alone. Adds the code-location
  predicate (a latent cross-district collision that turning Paterson on would
  widen) and the `state_excludefromreporting` filter, so an ENR section can no
  longer reference a school `schools.csv` omits. No cross-region
  `school_number` collisions exist in the post-change school set, so Paterson's
  ids are unambiguous.

Two new singular tests assert cross-feed referential integrity: every
`school_id` in enrollments, sections, teachers and staff resolves to schools,
and every `student_id` in enrollments resolves to students. Both run at
`severity: error`.

**Please confirm one side effect before merging.** Filtering the staff roster on
`home_work_location_dagster_code_location != 'kippmiami'` drops Miami-homed
people from the NJ rows they also appeared in, not just from Miami rows:

- 5 Miami-homed network staff (KTAF Inc. business unit, Teaching and Learning /
  Data / Executive) lose their `staff.csv` rows for every Newark and Camden
  school — 85 rows, via the CMO cross-join branch.
- 60 Miami-homed staff lose their `teachers.csv` row at `school_id` 0, the
  District Office.

That follows the decision as written, and everything removed is a Miami-homed
person. If those people should keep NJ access, the surgical alternative is to
keep the roster-side Miami filter only on the primary-school assignment branch
and let the CMO and all-NJ branches keep filtering via the already
Miami-excluded `schools` CTE. Say the word and I'll adjust.

Restarting `kipptaf__extracts__clever__asset_job_schedule` is deliberately not
part of this PR — handing that back per the issue.

## AI Assistance

Claude-authored. Human-directed: both gating decisions (Miami out, Paterson in)
were made by Charlie and recorded on #4653 before implementation. Claude
re-verified the issue's row counts against prod, wrote the model changes and
tests, and ran the verification below.

### General

### dbt

- [x] Properties files exist for all six models; the two new singular tests have
      `tests/properties.yml` entries with descriptions and `meta.dagster.ref`.
- [x] No exposure change needed — these feed the Dagster BigQuery-to-SFTP
      extract assets, not a dashboard.
- [x] **Not a breaking change** — no columns renamed or removed. `state_id` keeps
      its name and type; only its derivation is simplified. Rollback is a plain
      revert, since nothing downstream reads these views.

## Verification

Local dev build of all six models plus their tests, deferred to prod
upstreams with `--favor-state`:

```text
Done. PASS=11 WARN=0 ERROR=0 SKIP=0 NO-OP=0 TOTAL=11
```

Negative control — the same two tests run against the unfixed prod views fail,
confirming they catch the bug rather than passing vacuously:

```text
Failure in test rpt_clever__enrollments__student_id_resolves_to_students_feed
  Got 1114 results, configured to fail if != 0
Failure in test rpt_clever__school_id_resolves_to_schools_feed
  Got 10 results, configured to fail if != 0
```

Predicted post-merge orphan count for Paterson was checked ahead of the change:
840 students in the enrollments branch against 840 in the students feed, zero
orphans.

`trunk check --force` on all nine changed files reports no issues after the
commit-time `fmt` pass.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes. This PR is kipptaf-only, so CI's
      `state:modified+` selection genuinely builds these six models and both new
      tests.
      `src/dbt/`.

Refs #4653

### DIFF
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__enrollments.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__enrollments.sql
@@ -14,7 +14,8 @@ inner join
     and s.enroll_status in (0, -1)
 where
     cc.dateleft >= current_date('{{ var("local_timezone") }}')
-    and cc._dbt_source_relation not like '%kipppaterson%'
+    -- Miami rosters into Clever directly from Focus; excluded from all six feeds
+    and cc._dbt_source_relation not like '%kippmiami%'
 
 union all
 
@@ -30,4 +31,4 @@ select
 
     student_number as student_id,
 from {{ ref("stg_powerschool__students") }}
-where enroll_status in (0, -1) and _dbt_source_relation not like '%kipppaterson%'
+where enroll_status in (0, -1) and _dbt_source_relation not like '%kippmiami%'
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__schools.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__schools.sql
@@ -22,7 +22,9 @@ select
 from {{ ref("stg_powerschool__schools") }}
 where
     /* filter out summer school and graduated students */
-    state_excludefromreporting = 0 and _dbt_source_relation not like '%kipppaterson%'
+    state_excludefromreporting = 0
+    -- Miami rosters into Clever directly from Focus; excluded from all six feeds
+    and _dbt_source_relation not like '%kippmiami%'
 
 union all
 
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql
@@ -2,6 +2,7 @@ with
     dsos as (
         select
             sr.powerschool_teacher_number,
+            sr.home_work_location_dagster_code_location,
 
             coalesce(
                 ccw.powerschool_school_id, sr.home_work_location_powerschool_school_id
@@ -13,12 +14,33 @@ with
             and not ccw.is_pathways
         where
             sr.assignment_status != 'Terminated'
+            -- Miami rosters into Clever directly from Focus; excluded from all
+            -- six feeds
+            and sr.home_work_location_dagster_code_location != 'kippmiami'
             and sr.job_title in (
                 'Director of Campus Operations',
                 'Director Campus Operations',
                 'Director School Operations',
                 'School Leader'
             )
+    ),
+
+    schools as (
+        -- Matches rpt_clever__schools' filters so an auto-generated ENR section
+        -- can never reference a school that schools.csv omits.
+        select
+            abbreviation,
+            low_grade,
+            high_grade,
+            school_number,
+
+            regexp_extract(
+                _dbt_source_relation, r'(kipp\w+)_'
+            ) as dagster_code_location,
+        from {{ ref("stg_powerschool__schools") }}
+        where
+            state_excludefromreporting = 0
+            and _dbt_source_relation not like '%kippmiami%'
     ),
 
     teachers_long as (
@@ -89,7 +111,9 @@ with
             and st._dbt_source_project = t._dbt_source_project
         where
             sec.terms_yearid = ({{ var("current_academic_year") - 1990 }})
-            and sec._dbt_source_relation not like '%kipppaterson%'
+            -- Miami rosters into Clever directly from Focus; excluded from all
+            -- six feeds
+            and sec._dbt_source_relation not like '%kippmiami%'
 
         union all
 
@@ -132,8 +156,9 @@ with
             'Homeroom/advisory' as `subject`,
         from dsos
         inner join
-            {{ ref("stg_powerschool__schools") }} as s
+            schools as s
             on dsos.school_id = s.school_number
+            and dsos.home_work_location_dagster_code_location = s.dagster_code_location
         cross join unnest(generate_array(s.low_grade, s.high_grade)) as grade_level
     ),
 
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__staff.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__staff.sql
@@ -17,8 +17,9 @@ with
         where
             not is_prestart
             and worker_status_code != 'Terminated'
-            -- Paterson gated from Clever until #4193; drop with the sibling feeds
-            and home_work_location_dagster_code_location != 'kipppaterson'
+            -- Miami rosters into Clever directly from Focus; excluded from all
+            -- six feeds
+            and home_work_location_dagster_code_location != 'kippmiami'
 
         union all
 
@@ -37,7 +38,7 @@ with
             department as home_department_name,
             title as job_title,
         from {{ ref("int_people__temp_staff") }}
-        where dagster_code_location != 'kipppaterson'
+        where dagster_code_location != 'kippmiami'
     ),
 
     schools as (
@@ -52,7 +53,7 @@ with
         from {{ ref("stg_powerschool__schools") }}
         where
             state_excludefromreporting = 0
-            and _dbt_source_relation not like '%kipppaterson%'
+            and _dbt_source_relation not like '%kippmiami%'
     ),
 
     assignments as (
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__students.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__students.sql
@@ -50,6 +50,9 @@ select
     sr.gifted_and_talented as ext__gifted,
     sr.cumulative_y1_gpa as weighted_gpa,
     sr.cumulative_y1_gpa_unweighted as unweighted_gpa,
+    -- every region left in this feed is NJ, so the FL education id (fleid) that
+    -- Miami reported here is no longer reachable
+    sr.state_studentnumber as state_id,
 
     c.contact_name,
     c.relationship as contact_relationship,
@@ -79,7 +82,6 @@ select
 
     if(sr.lep_status, 'Y', 'N') as ell_status,
     if(sr.spedlep in ('SPED', 'SPED SPEECH'), 'Y', 'N') as iep_status,
-    if(sr.region = 'Miami', sr.fleid, sr.state_studentnumber) as state_id,
     if(sr.grade_level = 0, 'Kindergarten', cast(sr.grade_level as string)) as grade,
 from {{ ref("int_extracts__student_enrollments") }} as sr
 left join
@@ -91,4 +93,5 @@ where
     and sr.rn_year = 1
     and not sr.is_out_of_district
     and sr.enroll_status in (0, -1)
-    and sr._dbt_source_relation not like '%kipppaterson%'
+    -- Miami rosters into Clever directly from Focus; excluded from all six feeds
+    and sr._dbt_source_relation not like '%kippmiami%'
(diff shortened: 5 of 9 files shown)


cbini:
## Review response

### dbt Cloud failure was unrelated to this change, and is resolved

The failure was `rpt_clever__students` not finding
`zz_stg_kipptaf_students.int_students__contacts`. That relation was created by a
concurrent staging-clone job at 19:31:46.322Z; the failing query ran just before
it landed and the build step reported at 19:31:56.668Z. It is a clone-vs-build
race on an unmodified upstream, not a defect in this diff — the compiled SQL
already pointed at the correct relation, which now holds 46,191 rows.

Retried the run (`70403219134522`): **passes, with zero warnings**
(`warning_only` returns `total_warnings: 0`).

### Important 1 — Miami-homed staff losing NJ rows

Agreed this is a decision rather than a code defect; it is the sign-off being
requested and nothing is merged until it is answered.

### Important 1 sub-note — pushing back on "85 may undercount"

The 85 figure is measured, not estimated. It comes from a row-level diff of the
current prod feed against a local dev build, grouped by region:

| region | rows removed | distinct staff |
| ------ | -----------: | -------------: |
| Newark |           60 |              5 |
| Camden |           25 |              5 |

Paterson contributes **zero** removed rows because Paterson had no `staff.csv`
rows before this change — its rows are added, not removed. So the post-change
19-school cross-join width is not the right multiplier for a removal count.

The arithmetic that reconciles both numbers: those 5 people previously produced
95 rows total — 85 at NJ schools plus 10 at the two Miami schools. The 10 Miami
rows are part of the intended 308-row Miami removal. "85 rows lost at Newark and
Camden schools" is therefore exact, and it is the figure the access decision
turns on.

### Minor 2 — colon-space in the description: not a defect, leaving as-is

The claim is that this breaks YAML parsing and that it is the only description in
the file that does it. Both are incorrect:

- Parsing: loaded the edited file with a YAML parser — it parses, and the
  description round-trips to 541 characters intact. The local `dbt build` also
  parsed all 823 data tests without error.
- Prevalence: **285 descriptions across 115 yml files under `src/dbt/` already
  contain a colon-space.** It is not an outlier, so there is no consistency
  argument for rewording either.

Colons are literal inside a folded block scalar, which is what every description
in this file uses. The `src/dbt/CLAUDE.md` rule concerns plain (non-block)
multi-line scalars, where a colon-space genuinely is ambiguous. No change.

### Minor 3 — duplicated region literal

Agreed, and confirmed the count: 12 occurrences across the six models
(`_dbt_source_relation not like` seven times, `dagster_code_location !=` five).
Consolidating into a var or macro is the right direction and this PR is evidence
of the cost of not having one — but it is a refactor of all six feeds beyond the
scope of the fix that unblocks the schedule restart. Happy to open a follow-up
issue.

### Minor 4 — sections `schools` CTE duplication is deliberate

Correct, and for exactly the reason given: `rpt_clever__schools` emits
`low_grade` as a display string (`'Kindergarten'`) and carries the hardcoded
District Office row, either of which would break the `generate_array` call. The
CTE already carries a comment explaining it mirrors that model's filters; I will
extend it to name the `generate_array` hazard explicitly if the file is touched
again for the Important 1 decision.

### Recommendation — teacher_id cross-check

Noted, and agreed it is out of scope: sections-to-teachers was not one of the
failure modes in #4653, and adding it here would widen the change without
evidence of a live gap.

### Independent re-verification

Re-ran the filter sweep myself rather than relying on the review's grep: 12
`kippmiami` predicates across the six models, zero remaining Paterson gates.

cbini:
## Decision on the open question — resolved, no code change

The open question in the PR body (and Important 1 in the review) is answered:
**ship as-is with the single simple rule — anyone based in Miami is excluded from
all six feeds.**

Accepted trade-off, deliberately, with eyes open:

- 5 Miami-based network employees (KTAF Inc. business unit — Teaching and
  Learning, Data, Executive) lose their `staff.csv` rows at all 12 Newark and all
  5 Camden schools: 85 rows.
- 60 Miami-based staff lose their `teachers.csv` row at `school_id` 0, the
  District Office.

Rationale for the simple rule over the surgical alternative: one filter that
keys off the person's home region is easier to reason about and harder to get
wrong on the next region change than two rules with a carve-out — which matters
given the drift across six hand-maintained filters is precisely what caused
#4653.

The surgical alternative, for the record if this is revisited: apply the
person-based Miami filter only to the own-school assignment branch of
`rpt_clever__staff` (where it is load-bearing, since Miami school ids like
`30200805` are absent from `schools.csv` and would orphan), and let the
network-wide and all-NJ branches rely on the `schools` CTE, which already
excludes Miami.

Nothing further to change on this branch. All 8 checks are green, dbt Cloud
passed on retry with zero warnings, and this is ready for CODEOWNERS review.

PR #4668, opened 2026-07-31
feat(extracts): add KIPP Newark ParentSquare outbound feed

### LINKED ISSUE #4480: feat: ParentSquare (KIPP Newark) outbound extracts
## Summary

Build a **KIPP Newark** outbound extract feed for ParentSquare (regional mass/emergency communication to Newark families). Follows the Clever precedent (`kipptaf` BigQuery to SFTP via `build_bigquery_query_sftp_asset`). This is the "Rebuild Student Contacts" successor to the Finalsite contact workstream.

Tracks Asana task 1216377560708697 (Summer 2027 : SIS Portfolio). Scope resolved via the KIPP NJ + ParentSquare Integration Planner and the ParentSquare SFTP Integration spec.

## Scope (rescoped from NJ-wide to Newark)

**KIPP Newark only** this phase (not Camden/Paterson). Four files:

- `schools`, `students`, `parents`, `emergency_contacts`

Excluded: attendance, staff (deferred to a later phase, `school_id = 0`), sections/rosters/terms, parent/student logins.

## Approach

- New `rpt_parentsquare__*` reporting views in `src/dbt/kipptaf/models/extracts/parentsquare/`, filtered to Newark
- New `parentsquare.yaml` extract config + job/schedule in `kipptaf/extracts/` (early-evening send, shipped paused)
- New `ssh_parentsquare` SFTP destination resource (server `sftp3.parentsquare.com`, district-level)

Contacts sourced from Finalsite via `int_students__contacts` (Household 1 parents to `parents.csv`; `emergency_1..4` slots to `emergency_contacts.csv`). Keys are SIS-native: `student_id = student_number`, `school_id = school_number`; emergency `contact_id = hash(finalsite_contact_id, contact_slot)`.

## Blocker

SFTP credentials — set up in ParentSquare (issues username; we set password/SSH key), pending app access.

Design spec: `docs/superpowers/specs/2026-07-29-parentsquare-newark-extracts-design.md`.

cbini:
# Pull Request

## Summary & Motivation

...build the KIPP Newark ParentSquare outbound feed — **seven** `rpt_parentsquare__*`
models delivered as full CSV snapshots to a new `parentsquare` SFTP destination
via the existing `build_bigquery_query_sftp_asset` factory, following the Clever
precedent.

**The schedule ships STOPPED.** SFTP credentials are still pending from
ParentSquare, so nothing can send yet. Cron is 6pm ET per ParentSquare's
early-evening recommendation, with
`default_status=DefaultScheduleStatus.STOPPED` spelled out explicitly rather
than relying on the Dagster default.

Design spec:
`docs/superpowers/specs/2026-07-29-parentsquare-newark-extracts-design.md`,
updated here with a _File set_ section and _Implementation notes_ recording every
correction below.

### File set — all required files, not the four originally scoped

The issue and the Integration Planner scoped four files. Reading ParentSquare's
actual SFTP spec (page 4) changed that: it names **seven required** files, and
classifies `emergency_contacts.csv` as *optional*.

| ParentSquare spec        | Status                                                 |
| ------------------------ | ------------------------------------------------------ |
| `schools.csv`            | required — built                                       |
| `students.csv`           | required — built                                       |
| `parents.csv`            | required — built                                       |
| `staff.csv`              | required — built (regional Operations leaders only)    |
| `terms.csv`              | required *only if terms is enabled* — omitted          |
| `sections.csv`           | required — built (synthetic per-grade sections)        |
| `rosters.csv`            | required — built (student to per-grade section)        |
| `emergency_contacts.csv` | **optional** per the spec — built anyway, Ops wants it |

The Planner marked Staff and Sections as not needed, so the two roster-side files
are shaped to satisfy ParentSquare **without contradicting the Planner's other
decisions**:

- **`staff.csv` carries only the regional Operations leadership**
  (Planner Q4: "No school staff, no teachers"), each emitted once per operating
  school. ParentSquare's staff file is per-school and its spec permits a staff
  member at more than one, so the fan-out is what grants them school-level access
  everywhere and what makes every `sections.staff_id` resolve at its own school.
  Membership comes from the hand-curated **`TS-DL-Regional Ops Leaders`** LDAP
  distribution list, scoped to active Newark regional-office Operations staff, so
  Ops owns the roster and role changes need no code change. See _Identifying the
  Operations leaders_ below.
- **`sections.csv` holds one synthetic section per (school, grade)**, owned by an
  Operations leader, rather than real course sections — the same pattern
  `rpt_clever__sections` already uses for its auto-generated `ENR` sections. This
  is the load-bearing design choice: it satisfies the required file at exactly
  the granularity Planner Q5 specifies ("School + Grade Level only") **without
  importing a single teacher**. That matters because a staff row with an email is
  synced as a ParentSquare staff user, so real teacher-led sections would have
  provisioned accounts for every Newark teacher — the opposite of Q4.
- **`rosters.csv` places each student in their (school, grade) section.**

`terms.csv` is omitted: the spec makes it required only if terms is enabled, and
`sections.term_id` required only when sending `terms.csv`, so both stay empty.

Sections derive from the (school, grade) pairs students are actually enrolled in,
so no section is empty and every roster row resolves.

### Identifying the Operations leaders

`staff.csv` sources membership from **`TS-DL-Regional Ops Leaders`**, scoped to
active Newark regional-office (`school_id = 0`) Operations staff — **8 people**.

The group is hand-curated (no `idautoGroupIncludeFilter`, no sync interval, so a
human sets membership) and was last modified 2026-06-25, so it is actively
maintained. Two alternatives were rejected:

- **A hardcoded list of individuals** matched on KIPP mail. Reproduces the
  Planner's six exactly but rots silently — a new Operations leader is absent
  until someone edits SQL. This is what the earlier commits did.
- **A `job_function` rule** (`Chief Level`, `KTAF or Regional Director`,
  `EDs, HOSs, MDOs`). Also returns exactly six today, but only because
  `job_function` is **NULL** for one Managing Director of School Operations who
  shares a `job_title` with two included peers. `job_function` is null for ~3% of
  active Newark staff, so that is a data gap, not a signal, and a routine backfill
  would silently add them. Rejected because the failure mode is silent
  over-provisioning.

**The group yields 8 where Planner Q4 names 6.** The extras are a Deputy Chief and
the MDSO above. Both are in the curated leaders list *and* in the automated
`TS-SG-Okta Parentsquare All Ops` group, so both look intended — but this needs an
Ops confirmation, and if they disagree the fix belongs in the group rather than in
SQL so the list stays Ops-owned. Tracked as an open question in the spec.

For the record, `TS-SG-Okta Parentsquare All Ops` is unusable here despite its
name: it is idauto rule-generated on a 4-hour sync and resolves to all 12 regional
Operations staff at the same scoping (including Facilities Managers); without the
`school_id = 0` filter it reaches 124 Newark staff, 88 school-based, contradicting
"no school staff". It governs app SSO, not the roster feed.

### Cross-file referential integrity is enforced, not assumed

Every cross-file reference has a `relationships` test at `severity: error`:
`students` / `staff` / `sections` / `rosters` all point at `schools.school_id`;
`rosters` points at both `sections.section_id` and `students.student_id`; and
`sections.staff_id` points at `staff.staff_id`.

### `int_students__contacts` gains two columns

ParentSquare requires discrete `first_name` / `last_name` per contact, and the
model exposed only the combined `contact_name`. Rather than split that string —
wrong for the ~4% of Newark contacts with 3+ name tokens — the model now carries
`contact_first_name` / `contact_last_name`. Both branches supply them natively
(Finalsite via `int_finalsite__student_contacts`, PowerSchool via
`int_powerschool__contacts`), so no name is ever derived by splitting.

Purely additive. All four consumers (`int_students__contacts_pivot`,
`rpt_clever__students`, `bridge_student_contacts`,
`dim_student_contact_persons`) enumerate columns explicitly and are unaffected.

### Corrections to the design spec

- **`contact_id` keys on `(student_number, contact_slot)`**, not
  `finalsite_contact_id`. The spec's formula was broken:
  `int_finalsite__student_contacts` sets `finalsite_contact_id` to explicit NULL
  for every emergency row, so the original hash would have collapsed to 4
  distinct values network-wide.
- **`grade_level` needs no mapping table.** Newark's domain is `0`..`12`, which
  already satisfies ParentSquare's `-4..12` scale with K = 0 — independently
  confirmed against the spec's own documented domain.
- **Principal names come from `int_people__staff_roster`**, joined on
  `lower(principalemail) = lower(mail)`, not from splitting PowerSchool's single
  combined `principal` field, which carries honorifics. All 12 reporting schools
  match; `mail` is unique on the roster so the join cannot fan out.
- **The `dagster-cloud.yaml` variable mappings are deferred to the credentials
  phase.** Those entries are `secretKeyRef`s into a 1Password-synced Secret
  (`op-parentsquare-sftp`) that does not exist yet, and a `secretKeyRef` to a
  missing Secret fails container creation — taking the **whole `kipptaf` code
  server** down on deploy, not just this feed. `EnvVar` resolves lazily, so
  unmapped variables are inert while the schedule stays stopped.
- **No dbt exposure.** No SFTP extract feed in this repo has one, and the extract
  asset already depends on the dbt model's asset key.

### Phone validation

Contact models strip formatting, drop a leading US country code, truncate to 10
digits, then null anything that is not exactly 10 digits rather than send a
number ParentSquare would reject along with the rest of the row. This drops **6
emergency-contact rows** whose only phone is a 9-digit typo and that carry no
email — an upstream Finalsite data-entry fix for Ops.

### Verified output (current academic year, prod data)

| File                 | Rows   |
| -------------------- | ------ |
| `schools`            | 12     |
| `students`           | 6,796  |
| `parents`            | 9,259  |
| `emergency_contacts` | 15,399 |
| `staff`              | 96     |
| `sections`           | 53     |
| `rosters`            | 6,796  |

`staff` is 8 leaders x 12 schools with no missing email, first_name, or last_name. `sections` is 53 distinct
(school, grade) pairs; every `section_id` is unique, at most 11 characters, and
free of spaces and periods. `rosters` is one row per student.

Referential integrity holds in both directions across the whole set — no orphan
`school_id`, `student_id`, `section_id`, or `staff_id`; no section without
students; no student without a roster row. No contact row lacks both email and
phone, no phone is other than 10 digits, every school has a resolved principal
name, and `state_student_id` / `student_email` are unique as the spec requires.

## AI Assistance

Claude-authored, human-directed. The scope expansion from four files to all
required files was a human decision, as was scoping `staff.csv` to the
Operations leaders group. The corrections above were found during implementation by
validating the design spec's claims against ParentSquare's actual SFTP spec and
against prod data.

### General

### Dagster

- [ ] `dagster definitions validate` for kipptaf — **not run locally.**
      `kipptaf.definitions` fails at module load in a codespace on the
      Illuminate/Zendesk dlt credential specs, unrelated to this change. Verified
      by importing the `extracts` submodules directly: 7 assets built, each
      requiring `ssh_parentsquare` and each depending on the correct
      `kipptaf/extracts/rpt_parentsquare__*` key; all 7 in the package `assets`
      list; job `kipptaf__extracts__parentsquare__asset_job`; schedule
      `0 18 * * *` / `America/New_York` / `DefaultScheduleStatus.STOPPED`.
      Additionally, the branch deployment for an earlier commit loaded the full
      `Definitions` cleanly in a real environment — see note below.
- [ ] `uv run pytest tests/test_dagster_definitions.py` — **not run**, same
      module-load blocker. Needs a full environment.
- [x] Follows the Library plus Config pattern with the correct asset key format —
      adds a destination to the existing `extracts` module. Asset keys confirmed
      empirically by running the real `CustomDagsterDbtTranslator` against a
      parsed worktree manifest.

### dbt

- [x] Properties file for every new model, with `description` on the model and
      every column, PII columns tagged `contains_pii`, a uniqueness test per
      model, and cross-file `relationships` tests.
- [x] Exposure — intentionally omitted; rationale above.
- [x] **Not a breaking change.** `int_students__contacts` is additive only (not
      contract-enforced); the seven models are new relations. No column renamed
      or removed anywhere.
      is needed.

### Docs

- [ ] Regenerate the automations catalog
      (`uv run scripts/gen-automations-doc.py`) — **not run, deliberately.** The
      script silently skips code locations that fail to import, and `kipptaf`
      does not import in a codespace, so running it here would drop kipptaf from
      the catalog entirely. Needs a full environment. This PR adds one schedule,
      so the catalog is stale until then.
- [x] No code-location `CLAUDE.md` update needed — kipptaf's already lists
      `extracts` with a schedule; this adds a destination, not a new integration
      module.

### Local validation

`dbt build` of the modified upstream plus all seven models: **44/44 nodes pass, 0
warnings**, contracts enforced. Re-run after the formatter rewrote the SQL, since
sqlfmt reflows these models. `trunk check --force` over all 23 changed files: **no
issues**.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

> Note on an earlier red `dagster-cloud-deploy / deploy`: that run's code server
> was `Preempted` mid-import on a capacity-starved GKE Autopilot cluster and its
> replacement came up clean ~90s after the deploy CLI's wait window expired. The
> re-run passed on the first attempt. Not a code failure.

## Open items for a human

1. **Confirm the reduced roster shape with ParentSquare / Ops before un-pausing.**
   The synthetic per-grade sections deliberately import no teachers; confirm that
   grade-level sections owned by an Operations leader give the audiences expected.
2. **Regenerate the automations catalog** in a full environment.
3. **6 emergency contacts with 9-digit phone typos** — upstream Finalsite fix.

Refs #4480


cbini:
## Review response

Verified each finding against the codebase and warehouse before acting. Three applied in `02a9f42a9`, two declined with reasoning.

### 1. Missing `relationships` test on `school_id` — applied

Correct, and the important part is that it exposed an overclaim in my PR body. I wrote that `students.csv` "cannot reference an absent `school_id`"; the accurate statement is that two independently filtered queries agree today and nothing in the DAG forces them to keep agreeing. Added on `rpt_parentsquare__students.school_id` at `severity: error`:

```yaml
- relationships:
    arguments:
      to: ref('rpt_parentsquare__schools')
      field: school_id
```

Also confirmed your reasoning that this one test covers the contacts models: `parents` and `emergency_contacts` inner-join their own copy of the same enrollment filter, so their `school_id` domain is a subset of the students feed's. Adding the test to them too would be vacuous per the pre-join convention, so I did not.

### 2. Possible stale `union_relations` defer — already resolved, no change

This one is empirically closed, and by exactly the check you asked for: **dbt Cloud CI passed on this PR** (run `70403219134777`) — it was still running when your review was written, which is why the checklist looked unchecked.

The mechanism makes it decisive rather than circumstantial: `int_students__contacts` **is** `state:modified` in this PR, so CI rebuilt it, and its Finalsite branch selects `fc.contact_first_name` / `fc.contact_last_name`. Had the `zz_stg_*` defer copies carried the pre-#4611 schema, that build would have failed `Name contact_first_name not found`. It succeeded. Independently, `INFORMATION_SCHEMA.COLUMNS` on `kipptaf_finalsite.int_finalsite__student_contacts` lists both columns in prod.

Good call flagging it though — the failure mode is real and would have been invisible in my local build.

### 3. `grade_level` domain not defensively filtered — applied

Verified and you're right that it's reachable: `grade_level = 99` is not merely a sentinel in a neighbouring CTE, it flows to this model's output grain (`int_extracts__student_enrollments.sql:273` branches on `e.grade_level = 99`). Warehouse check: **22,565 rows across 3 regions, AY2006–2025** carry it, though **zero** survive this model's filter combination — so the risk is latent, not live.

Added `not_null` plus `accepted_values` over `-4`..`12` rather than a filter or a mapping. Reasoning: silently dropping the row would cut a family out of an emergency-notification feed, and inventing a grade mapping is a business decision for Ops, not a modelling one. The test makes the assumption self-policing; a hit is the signal to agree a real mapping.

### 4. Inconsistent region filter — applied

Switched `rpt_parentsquare__schools.sql` to `_dbt_source_project = 'kippnewark'`. Confirmed the column exists on `stg_powerschool__schools`, and `src/dbt/kipptaf/CLAUDE.md` is explicit that `_dbt_source_relation` does not always encode region, so the materialized column is the correct one. Output unchanged at 12 rows.

### 5. Extract the phone pipeline into a macro — declining

Two reasons beyond the precedent you already noted:

The pipeline isn't a single expression — it spans CTEs and yields an intermediate column the models depend on. Both contact models gate deliverability on the **pre-truncation** digit column (`phone_mobile_digits` / `phone_digits`), not the final 10-digit value, because a value that reduces to `'1'` would truncate to a non-empty-but-invalid string and slip past an `is not null` gate. A macro returning the final value wouldn't serve that, and one returning both columns would be more contorted than the inline CTEs.

It also wouldn't remove the duplication it targets. `rpt_clever__students` is a different feed with a different phone-type shape (it `UNPIVOT`s home/cell/work into rows); converting it is a change to a live Clever feed and belongs in its own PR, not this one. Converting only my two call sites leaves three variants instead of four.

Happy to file a follow-up issue for a repo-wide phone-normalization macro across both feeds if that's wanted.

---

**Verification after changes:** 24/24 nodes pass (both new tests included); output unchanged at 12 / 6,796 / 9,259 / 15,399 rows; `trunk check --force` clean.

One correction to your scope note, for the record: the earlier `dagster-cloud-deploy / deploy` failure was not code — the code server was `Preempted` mid-import at `19:52:50` (GKE Autopilot, capacity-starved cluster) and its replacement came up clean at `19:54:46` with `Started Dagster code server for module teamster.code_locations.kipptaf.definitions`. The re-run passed. That branch-deployment load is also the real-environment `Definitions` validation I couldn't run locally.

PR #4671, opened 2026-07-31
feat(dbt): swap Miami student contacts to Focus (Phase 2)

### LINKED ISSUE #4585: feat: replace Miami PowerSchool student contacts with Focus

cbini:
# Pull Request

## Summary and Motivation

Phase 2 of #4585. Swaps the Miami branch of `int_students__contacts` from the
frozen `kippmiami_powerschool` snapshot to live Focus data, and retires the
kipptaf PowerSchool contacts chain that branch was the sole consumer of.

Refs #4585 (Phase 1 was #4594).

> [!WARNING]
> **DRAFT — do not merge yet.** The code is complete and verified; the data is
> not there. Merging today would take Miami contacts from stale-but-present to
> effectively empty.

| Miami contacts in `int_students__contacts` | rows      | students  |
| ------------------------------------------ | --------- | --------- |
| prod today (PowerSchool, frozen)           | **9,382** | 3,867     |
| this branch (Focus)                        | **32**    | 12        |

That is a 99.7% drop, because the Finalsite-to-Focus contacts import has not run
at scale — Focus holds 20 contact links against 3,873 students. Merge gate: Focus
`students_join_people` link count reaching the same order of magnitude as
enrolled students, per the design spec.

## What changed

- `int_students__contacts` — the `ps_*` CTE chain (base, slotting,
  person-contacts enrichment, primary-phone ranking, frozen-students join) is
  replaced by a `focus` branch reading the Phase 1 kipptaf wrapper. The NJ
  Finalsite branch is untouched.
- Deleted the kipptaf `int_powerschool__contacts` union wrapper plus its
  properties file, and dropped the `int_powerschool__contacts` /
  `int_powerschool__person_contacts` entries from the powerschool
  `sources-kippmiami.yml`. The wrapper unioned Miami only (the NJ chain was
  already disabled in #4346), so nothing else loses a source. The frozen
  `kippmiami_powerschool` dataset stays — it still feeds
  `int_fldoe__all_assessments` and others.
- Added a unit test pinning the slotting rules.

## The student_number derivation

This resolves open question 2 in the design spec, which asked which
`student_number` form the Miami spine uses. The answer, with evidence:

Focus stores the KIPP student number `8400`-prefixed in `local_student_id`.
Stripping the prefix derives a `student_number` for **all 3,873** Focus students,
and it agrees with `int_finalsite__contact_id_attributes.powerschool_student_number`
on **all 1,377** rows where that column is populated, with zero disagreements.

The obvious alternative — joining that crosswalk on `focus_student_id_prefixed` —
is a trap worth recording. `powerschool_student_number` is null for 4,490 of
5,868 Miami crosswalk rows, because Focus-native students have no pre-migration
PowerSchool record. Joining on it covers only 35.6% of Focus students and would
have silently dropped roughly two thirds of Miami while appearing to work.

## Decisions carried from review

- **Slotting keeps the spec rule as written.** `contact_1` is the `sort_order` 1
  link; `emergency_N` ranks every emergency-flagged link by `sort_order`,
  including that same row. So the primary contact deliberately occupies both
  `contact_1` and `emergency_1`. This does not break the grain test, which keys
  on `contact_slot` alongside `student_number` and `_dbt_source_project`.
  Confirmed in the build: 12 `contact_1`, 12 `emergency_1`, 8 `emergency_2`.
- **Capped at `emergency_4`.** Not cosmetic — there is an `accepted_values` test
  on `contact_slot` permitting only `contact_1`, `contact_2`, and
  `emergency_1..4`, so an uncapped rank fails the build. It also matches the
  outgoing branch and the fixed slot list in `int_students__contacts_pivot`.
- **Phone values stay raw, no E.164 normalization.** The design spec flags this
  as a required obligation, but that note is stale. `rpt_clever__students`, the
  one consumer it names as previously broken, now strips all non-digits and a
  leading `1` before slicing 10, so it is format-agnostic. The other four
  consumers pass phones through with no digit assumptions. The real cost is
  cosmetic: mixed formats in one column, visible where the Sheets extract
  concatenates phones into display strings.
- **`person_id` breaks `sort_order` ties** in the emergency rank so slot
  assignment is stable across rebuilds. The outgoing PowerSchool branch had no
  tiebreak.

## Deliberately not in scope

The Newark, Camden, and Paterson `sources-kipp*.yml` entries for
`int_powerschool__contacts` and `int_powerschool__person_contacts` are left in
place. They had no consumer before this change either, since the wrapper unioned
Miami only — pre-existing dead declarations, not something this PR orphaned.
Removing them is a separate cleanup.

## Verification

All run locally against BigQuery; see the CI note below on why that matters.

- `dbt build --select int_students__contacts` — 6/6 pass: the model plus the
  grain test, both `not_null`s, and the `contact_slot` `accepted_values` test.
- `dbt build --empty --select int_students__contacts+` — **162 nodes pass**, zero
  errors, covering all 168 downstream models including
  `dim_student_contact_persons`, `bridge_student_contacts`,
  `rpt_gsheets__student_contact_info`, `rpt_clever__students`, and
  `int_students__contacts_pivot`. This proves every consumer still resolves its
  columns after the branch swap.
- New unit test `test_focus_slotting_contact_1_and_emergency_ranks` passes.
- `dbt parse` clean after the deletions: 771 models (was 772), 1,161 sources
  (was 1,163).
- `trunk check --force` clean on all changed files.

## AI Assistance

Claude Code co-authored this PR. Human-directed: the decision to build Phase 2
ahead of the data landing, keeping the spec slotting rule as written, and the
call not to normalize phones. AI-assisted: the id-form investigation that
resolved the spec open question, the branch SQL, the retirement scope analysis,
the unit test, and the downstream verification sweep.

### dbt

      models; the modified model's properties file gained the unit test.
- [x] Include or update an exposure for models consumed by a dashboard, analysis,
      or application — n/a, no new external consumer.
- [x] **Breaking change?** Behaviourally yes for Miami, which is the point of the
      change, and the reason this is a draft. No contracted column was renamed or
      removed; the output column set is unchanged.
      with `--target staging` — n/a, no external sources.

## CI checks

Note that the dbt Cloud CI job builds kipptaf, so it does exercise this change —
unlike a district-only PR. Local verification above is still the stronger signal
for the downstream sweep.

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.


cbini:
Processed with `superpowers:receiving-code-review` — verified each cited line against the file before changing anything. Fixed in 8aaccaba2.

## The finding is correct, including the part that matters most

Confirmed all six citations resolve to real stale text, and confirmed the additive-only claim: `git diff HEAD~3..HEAD` on that file was `+102/-0`, so the Focus swap left every description untouched. That was a genuine miss — I edited the SQL and appended the unit test without revisiting the docs.

The important half of this finding is not the terminology. It is that line 9 documented the **old** slotting rule — "priority-3-and-higher emergency contacts" — which the retired branch enforced via `contactpriorityorder >= 3`. The Focus branch ranks every emergency-flagged link **including** `sort_order` 1. So the docs asserted a disjointness the code deliberately does not have, and a reader debugging a duplicate contact would have concluded the code was broken. That is worse than a mislabeled source system, and the review was right to call it out as a behavior change the docs did not mention.

## What changed

- **Model description** — Miami now sources from Focus, `sort_order` replaces `priority`, and the emergency rule is stated as written: it includes the `sort_order` 1 contact, so Miami's primary contact occupies both `contact_1` and `emergency_1`. Added that the grain still holds because `contact_slot` differs, and that Miami never produces `contact_2` — both suggested in the fix note.
- **`personid`** — Focus person id for Miami. Also recorded that the id space is per-source-system and the region is always part of the dimension key, which is the reassurance your `dim_student_contact_persons` / `bridge_student_contacts` trace established.
- **`finalsite_contact_id`** — null for Miami (Focus-sourced) rather than "PowerSchool regions".
- **`is_emergency`** — a Miami `contact_1` carries its own Focus emergency flag and is usually true, which is *why* it also appears as `emergency_1`.

## One citation deliberately not changed

Line 7's "PowerSchool student number" stays. It describes the NJ Finalsite branch crosswalking to `powerschool_student_number` via `int_finalsite__contact_id_attributes`, which is still exactly what happens. It is not drift. (Line 228's `powerschool_student_number` is also untouched — that is a column name in the unit test's zero-row mock.)

## Verification

- `trunk check --force` clean on the file.
- Unit test `test_focus_slotting_contact_1_and_emergency_ranks` still passes after the edits.
- All 18 columns still carry a `description`, and the YAML parses.

Per the repo convention that a properties `description` change does not mark a model `state:modified`, this push should be a near no-op for dbt Cloud CI rather than a rebuild.

## On the assessment

Agreed that the merge blocker is the data gate, not this review. Miami is 9,382 rows in prod today against 32 on this branch, so it stays unmergeable until the Finalsite-to-Focus contacts import runs at scale — tracked in #4585. Noting for the record that the draft flag came off to trigger this review; the gate is unchanged.

cbini:
## CI warning worth recording: 13 new orphan rows in `bridge_student_contacts`

The doc-fix run (`70403219140619`) passed but carried 10 test warnings, where the prior run carried none. Nine are unrelated noise from the `state:modified+` fan-out. One is attributable to this change and is worth writing down, because it is data-dependent and therefore outside what the static review could see.

### The finding

`relationships_bridge_student_contacts_student_key__ref_dim_students_` warned with 33 results. That decomposes exactly:

| region    | rows in PR | students | orphan rows | orphan students |
| --------- | ---------- | -------- | ----------- | --------------- |
| Paterson  | 3,029      | 872      | **20**      | 6               |
| Miami     | 35         | 13       | **13**      | 5               |
| Newark    | 27,160     | 8,013    | 0           | 0               |
| Camden    | 6,620      | 2,494    | 0           | 0               |

Paterson's 20 are **pre-existing** — prod bridge against prod `dim_students` returns the same 20 rows / 6 students, so that test already warns on `main`. Miami's 13 rows across 5 students are **new with this PR**.

### Root cause

`bridge_student_contacts.student_key` is `generate_surrogate_key(["student_number"])`, and `dim_students` is built from PowerSchool enrollment. A Focus-native Miami student — enrolled after the migration, with no pre-migration PowerSchool record — derives a perfectly valid `student_number` from `local_student_id`, but has no row in `dim_students` for it to point at. So the contact rows land as orphans.

This is the same root cause family as the crosswalk trap already described in the PR body: Miami's student identity has moved to Focus while the PowerSchool-derived spine has not caught up.

### Scale when the data gate opens

Today this is 13 rows because only 20 contact links exist. Measured against the full Focus roster:

- 3,881 Focus students total
- 3,453 resolve to `dim_students`
- **428 do not — 11%**

So once the Finalsite-to-Focus contacts import runs at scale, expect this warn to report contacts for roughly 428 Miami students rather than 5.

### Why this is not being "fixed" here

The contacts model is doing the right thing. The alternative — filtering Miami contacts to students present in `dim_students` — would silently discard real contact data for 11% of Miami families, which is strictly worse than a warn-level orphan. The gap belongs to `dim_students` coverage of current Miami enrollment, not to this model, and it lands in the same territory as the enrollment-spine question carried as open question 2 in the design spec.

Leaving it as a warn. Flagging it here so that whoever opens the data gate is not surprised by the orphan count jumping two orders of magnitude, and so it is on the record as a known consequence rather than a regression discovered later.

### Note on review coverage

The static review explicitly stated it had no BigQuery/dbt/Dagster access and made no data-dependent claims. This finding is precisely data-dependent, so it complements that review rather than contradicting it — and it is not something the local `dbt build --empty` sweep could have surfaced either, since an empty build produces no rows and therefore no orphans.

PR #4673, opened 2026-07-31
docs: publish the Tableau permissions guide

cristinabaldor:
# Pull Request

## Summary & Motivation

...publish the Tableau row-level-security model as a guide at
`docs/guides/tableau-permissions.md`, so there is a referenceable place where
all Tableau permissions live.

One page, two halves, per request:

- **Part 1 — Who can see what.** Plain language, no calculation syntax. The five
  routes to seeing a row, what central office can and cannot see, the
  senior-leader shield, why Rooms do not grant access, why you might see less
  than a colleague, and how to get access (ask for the group, not a workbook
  edit).
- **Part 2 — Implementation reference.** The five calculated fields with the
  reasoning behind each shape: why `ISMEMBEROF` needs one branch per location
  (literal-string argument, and a parameter degenerates to all-rows-or-none), the
  five group-naming bridges, the `ISNULL(job_function)` fallbacks and what
  removing them early would silently break, per-workbook variants, and the
  testing personas to run after editing a workbook.

Part 2 also records the region-scoped KTAF entity branch and the leak it fixed —
two central office senior leaders at the same level able to see each other via
Tier 4, because the KTAF branch was unconditional and Rooms are absent from the
location gate, so Tier 4 is central office's only route. The coupling between
that branch and the Tier 2 senior-leader shield is called out explicitly, since
it is the thing a future edit is most likely to break.

Also adds the nav entry in `mkdocs.yml` and a row in the guides routing table.

Content was drafted through this migration's testing and carries no staff names,
usernames, or employee numbers — the individual by-name grants that used to live
in these workbooks were removed by the migration and the guide states they are
not permitted.

## AI Assistance

Claude-assisted, human-directed. The permissions model, the trade-off decisions
(region-scoped KTAF branch, keying seniority on `job_function`), and the leak
diagnosis came out of live workbook testing in this session; Claude drafted and
structured the document from that work.

## Self-review

### General

      (`claude-code-review` excludes markdown, so it will not fire on a
      docs-only PR.)

No Dagster changes.

No dbt changes.

### Docs

- [x] No schedule, sensor, or integration changes — automations catalog not
      regenerated.
- [x] `uv run mkdocs build --strict` passes.
- [x] Nav entry added to `mkdocs.yml` and a row added to
      `docs/guides/index.md`.

## CI checks

- [x] **Trunk** — `trunk check --force --no-fix` clean on all three files.
- [x] **dbt Cloud** — not applicable; no dbt changes.
- [x] **Dagster Cloud** — not triggered; no matching paths change.

Refs #4656


anthonygwalters:
## Field notes from running this guide end to end

I followed `docs/guides/tableau-workbook-editing.md` on a live workbook today — read the RLS calcs, then built five new calculated fields plus two parameters into a test copy of the Academic and Gradebook Health Suite in `TEMP-Walters`, repackaged, and dry-run published. The guide held up: the `shared-view` filter location, the CRLF trap, the internal-name-vs-caption warning and the three-layer filter reference stripping were all exactly as described, and `hidden_views` behaved as documented (4 live views in, 4 out).

Four things I hit that the guide does not cover. All reproduced, not theoretical.

### 1. A publish also drops embedded datasource credentials

This is the one I would fix first — it belongs as a fifth row in the table at L215, and it partly contradicts **What is not affected** at L238.

Measured on the same workbook, before and after a `CreateNew` publish of a byte-verified repackage:

```text
original          5 BigQuery connections, embed_password=True   -> all 4 views render
republished copy  5 BigQuery connections, embed_password=False  -> render fails
```

The render failure is:

```text
400074: Bad Request
There was a problem querying the image for view ...
ExportViewException: Tableau needs an unexpired OAuth refresh token to
connect to the data. Authorize refresh tokens or ask the datasource owner
for help.
```

Same service account on both, credential flag flipped. This is worse than the `hidden_views` trap the guide already documents, because it is not subtle degradation — the dashboard stops rendering at all. And a revision rollback would not obviously fix it, since the credential is server-side state that is not in the `.twb`.

Practical consequence for the sequence at **The sequence that worked**: step 7 needs either `connections=[...]` with credentials supplied at publish time, or a post-publish re-embed, or the whole thing routes through Desktop. Right now the guide reads as though an `Overwrite` is credential-neutral, and it is not.

I stopped short of the `Overwrite` on my own target for exactly this reason.

### 2. The repackage snippet at L195-L203 corrupts the source archive

`zout.writestr(item, data)` mutates the `ZipInfo` object it is handed, and those objects are the ones backing the source archive's central directory. So `zin` is progressively invalidated as you write, and a later read in the same loop fails:

```text
zipfile.BadZipFile: Bad magic number for file header
```

It surfaces on the second or third entry, so it looks like a corrupt download rather than a bug in the loop. Every entry reads fine individually, which sends you chasing the wrong thing.

Fix is to read all payloads before opening the output, and write fresh `ZipInfo` objects:

```python
payloads = {n: zin.read(n) for n in zin.namelist()}
payloads[twb_name] = edited.encode("utf-8")

with zipfile.ZipFile(out, "w", zipfile.ZIP_DEFLATED) as zout:
    for item in zin.infolist():
        fresh = zipfile.ZipInfo(item.filename, date_time=item.date_time)
        fresh.compress_type = item.compress_type
        fresh.external_attr = item.external_attr
        fresh.create_system = item.create_system
        zout.writestr(fresh, payloads[item.filename])
```

The SHA-identity assertion the guide recommends still passes afterwards, so that check does not catch this on its own.

### 3. Publishing needs `skip_connection_check=True`

Without it, both publish calls (L212 dry-run and L251) fail:

```text
403132: Forbidden
<workbook name> failed to establish a connection to your datasource.
```

The server probes the connection at publish time and there are no embedded credentials to probe with. The extracts are self-contained, so the flag is safe here — but nothing in the guide tells you to set it, and the error reads like a permissions problem rather than a missing argument.

### 4. Minor: listing workbooks is paginated at 100

Not shown in the guide, but it belongs near **Four parsing traps** because it is the same class of silent-wrong-answer failure. `srv.workbooks.get()` returns the first page only:

```text
first_page=100  total_available=450
```

Our site has 450 workbooks, and `SchoolMint Grow Dashboard` is not on the first page. My first pass fell through to `wbs[0]` and cheerfully audited an unrelated workbook in `Archive` while reporting success. Use `list(tsc.Pager(srv.workbooks))`.

### One suggestion on the credentials section

The throwaway `tests/**/test_zz_*.py` pattern works well and I used it throughout. Worth adding one line that it must be deleted in the same session and never committed — otherwise a live workbook-publishing path ends up sitting in the test tree. I checked and no workflow runs pytest in CI, so the risk is a committed publish path rather than an automated one, but the sentence is cheap.

---

None of this changes the shape of the doc — it is the most useful thing I have read on this repo's Tableau surface, and the `hidden_views` finding alone justifies it. Happy to push these as a commit on this branch if you would rather not hand-merge them. Note the branch currently conflicts with main: the sole conflict is in CLAUDE.md, and the paragraph it edits has since been restructured away, so dropping that hunk resolves it.

docs/guides/tableau-permissions.md
```diff
@@ -0,0 +1,453 @@
...
+Being in the central office group grants visibility into TEAM, KIPP Cooper
+Norcross, KIPP Miami, and KIPP Paterson rows. It does **not** grant visibility
+into other central office rows. Those are reachable only by being the person
+themselves, being their direct manager, or belonging to one of the network-wide
+groups in route 2.
+
+!!! note "A known consequence"
+
+    A central office director does not see their reports' reports. Route 1 covers
+    only the _direct_ manager. This is accepted rather than accidental — the
```
cristinabaldor, line 80: is this true now that we have the department scope? i don't think there's any real case where this data would be obscured. grow dashboards are only PM for regional employees, which ktaf has access to, surveys are scoped by department, not by reports.

docs/guides/tableau-permissions.md
```diff
@@ -0,0 +1,453 @@
...
+On the workbooks that carry performance and development data, rows belonging to
+senior leaders are restricted to that person, their manager, and the
+network-wide groups. Peers at the same level do not see each other, including
+peers who otherwise have broad access.
+
+Seniority is read from the ADP job function rather than from job title text, so
+a newly created senior title is covered automatically without anyone editing a
+workbook. Where the job function is missing, a job-title fallback applies —
+which currently over-reaches by one role, shielding executive assistants as
+though they were executives.
```
cristinabaldor: let's fix this and then not mention it in the guide.

docs/guides/tableau-permissions.md
```diff
@@ -0,0 +1,453 @@
...
+| A teacher or learning specialist                 | Your manager, your school's assistant principals, your school leader and director of school operations, your regional leadership, and the Employee Relations, Recruiting and Leadership Development teams |
+| An assistant school leader                       | Your manager, your school leader and director of school operations, your regional leadership, and those three teams — **not** other assistant principals                                                  |
+| A school leader or director of school operations | Your manager, your regional leadership, and those three teams — **not** other school leaders or DSOs                                                                                                      |
+| A departmental director                          | Your manager, your region's senior leadership, and those three teams — **not** other directors in your department                                                                                         |
```
cristinabaldor, line 114: let's dig into some examples here, i'm wondering if we should gate on department as well as region

docs/guides/tableau-permissions.md
```diff
@@ -0,0 +1,453 @@
...
+| Regional leadership                              | Your manager and those three teams — **not** other regional leaders                                                                                                                                       |
+| Central office staff                             | Your manager and those three teams. No regional leader sees central office answers                                                                                                                        |
```
cristinabaldor, line 116: does TEAM Council not also see everything except each other here?

docs/guides/tableau-permissions.md
```diff
@@ -0,0 +1,453 @@
...
+These are accepted, not undiscovered. Each is a place the gate is approximate.
+
+- **The TEAM Council shield hides chief-level titles, not council membership.**
+  Tableau can only ask which groups the _viewer_ belongs to, never the
+  respondent, so route 7's exclusion has to be inferred from job title. A
+  council member whose title is senior but not chief level therefore stays
+  visible to fellow members.
+- **Miami's KIPP Forward staff have no departmental viewer.** Route 6 is scoped
+  by region, and Miami has KIPP Forward respondents but no KIPP Forward director
+  of its own, so those answers reach only their manager and the route-2 teams.
```
cristinabaldor, line 176: can we not scope this and other regional folks to to department (KIPP Forward) + Director roles?()

docs/guides/tableau-permissions.md
```diff
@@ -0,0 +1,475 @@
...
+
+Ask for the one matching the department whose feedback you need to read.
+
+| Department rated                | Group                                               |
+| ------------------------------- | --------------------------------------------------- |
+| Compliance                      | `KNJ-SG-Tableau All Compliance`                     |
+| Data                            | `KNJ-SG-Tableau All Data`                           |
+| Development                     | `TS-DL-Development`                                 |
+| Finance, including Purchasing   | `TS-SG-R9 Finance` or `TS-SG-R9 Purchasing`         |
+| Human Resources Operations      | `Group Staff Employee Relations`                    |
```
cristinabaldor: Change to Human Resources - Employee Relations

docs/guides/tableau-permissions.md
```diff
@@ -0,0 +1,475 @@
...
+
+#### If you are a viewer, what you see
+
+Eight routes, and the peer exclusion differs on each because "your own level"
+means something different depending on where you sit.
+
+| Route | Who                                                                                                            | Reaches                                          | Minus                                                                                                                  |
+| ----- | -------------------------------------------------------------------------------------------------------------- | ------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------- |
+| 1     | You, and the manager recorded on the response                                                                  | that response                                    | nothing — a manager sees their report even when both are director-rank                                                 |
```
cristinabaldor: where does this actually happen? we don't typically show reports their manager responses anywhere

Issue #4670, opened 2026-07-31
refactor(dbt): replace hand-copied frozen-region exclusion filters in current-state extracts with a shared var


cbini:
## Problem

#4653 was caused by six independently-maintained region filters drifting apart:
`rpt_clever__teachers` had none at all, and `rpt_clever__sections` had one in its
PowerSchool branch but not its auto-generated ENR branch. #4667 fixed the
symptom, but it left the underlying shape intact — the exclusion is now spelled
out **12 times** across the six models, in two different forms:

- `_dbt_source_relation not like '%kippmiami%'` — 7 occurrences
- `dagster_code_location != 'kippmiami'` (or
  `home_work_location_dagster_code_location`) — 5 occurrences

Per-model counts after #4667: enrollments 2, schools 1, sections 3, staff 3,
students 1, teachers 2.

Adding or removing a region from these feeds means finding and editing every one
of those spots, in two syntaxes, across six files. That is precisely the failure
mode #4653 documents, so the same bug class remains available.

## Why not just drop the region from the union views

The obvious-looking fix — remove `kippmiami` from the powerschool
`union_relations` views so consumers need no filter — is **not** safe.
Those views carry Miami's historical PowerSchool data (AY2025 and earlier
attendance, assessments, enrollment, alumni). Marts and topline reporting depend
on it, and Miami PowerSchool being frozen is exactly why the history has to stay
readable. Dropping Miami at the union would break historical reporting to fix a
current-state feed.

The distinction that matters is **current-state consumers** versus
**historical consumers**. Only the former need to exclude a frozen region.

## Proposal

Introduce one shared definition of "regions excluded from current-state feeds"
and have consumers reference it instead of hardcoding a literal. Options, roughly
in increasing order of effort:

1. A `dbt_project.yml` var (e.g. a list of excluded code locations) referenced in
   each `where` clause. Smallest change, keeps the predicate visible at each
   site.
2. A macro wrapping both predicate forms, so the `_dbt_source_relation` and
   `dagster_code_location` spellings derive from the same list.
3. A small seed or intermediate model mapping code location to
   "is current-state rosterable", joined by consumers. Heaviest, but makes the
   set queryable and testable.

Preference is probably 1 or 2 — option 3 adds a join to feeds that are already
wide.

## Scope

Start with the six `rpt_clever__*` models, then audit for other current-state
consumers that hardcode a region. Worth checking at minimum:

- `models/extracts/` more broadly (google, deanslist, powerschool)
[...]

PR #4675, opened 2026-07-31
fix(clever): exclude terminated staff from the teachers extract

### LINKED ISSUE #4674: fix(clever): rpt_clever__teachers has no termination filter and ships terminated staff
## Problem

`rpt_clever__teachers` gates its roster branch on `uac_account_disable = 0`, plus
`not is_prestart` and non-null `employee_number` / school id. It has **no worker
status filter**, so a terminated employee keeps shipping in `teachers.csv` for as
long as their AD account stays enabled.

Its sibling feeds both filter:

| feed                       | termination filter                    |
| -------------------------- | ------------------------------------- |
| `rpt_clever__staff`        | `worker_status_code != 'Terminated'`  |
| `rpt_clever__sections` DSO | `assignment_status != 'Terminated'`   |
| `rpt_clever__teachers`     | none                                  |

## Evidence

As of 2026-07-31, prod `kipptaf_extracts.rpt_clever__teachers` carries **10 rows
for staff whose `worker_status_code` is `Terminated`**:

- Termination dates range from 2019-01-04 to 2026-07-31 (two of the ten
  terminated on 2026-07-31, which is ordinary processing lag; the other eight are
  stale, the oldest by more than seven years).
- 8 of the 10 point at `school_id` 0, the District Office sentinel.
- **2 point at live schools** — one Newark, one Camden — so terminated staff hold
  teacher records against active school rosters.
- All 10 have `uac_account_disable = 0`.

Row-level detail is deliberately not reproduced here (staff PII); it is available
locally to the data team.

Reproduce the count:

```sql
select count(*) as terminated_rows
from `teamster-332318.kipptaf_extracts.rpt_clever__teachers` as t
inner join
  `teamster-332318.kipptaf_people.int_people__staff_roster` as sr
  on t.teacher_id = sr.powerschool_teacher_number
where sr.worker_status_code = 'Terminated'
```

## Two distinct causes

1. **Missing dbt filter** (this issue). `teachers.csv` should exclude terminated
   staff the way `staff.csv` already does.
2. **Offboarding gap, out of scope here.** All 10 have enabled AD accounts after
   termination, one since 2019. That reaches beyond Clever to anything keyed on
   the account (mail, SSO), so it is a People Operations and IT question rather
   than a data fix. Flagging it so it is not lost, not proposing to solve it in
   dbt.

## Fix

Add `worker_status_code != 'Terminated'` to the `int_people__staff_roster` branch
of `rpt_clever__teachers`, matching `rpt_clever__staff`.

The `int_people__temp_staff` branch needs no change — that model carries no
worker-status or termination column (only `idauto_status` and
`uac_account_disable`), so there is no termination concept to filter on, and
`rpt_clever__staff` likewise filters only its roster branch.

## Note on test coverage

The two referential-integrity tests added in #4667
`rpt_clever__enrollments__student_id_resolves_to_students_feed`) do **not** cover
this — they assert that school and student ids resolve across feeds, not that the
people shipped are currently employed. A terminated employee at a valid school id
passes both.

## Gate

`kipptaf__extracts__clever__asset_job_schedule` is still STOPPED after #4667.
This should land before it is switched back on, otherwise the first hourly sync
pushes all 10 terminated staff — including the two at live schools — into Clever.

Surfaced while reviewing the #4667 drop-off list. Refs #4653.

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will stop `teachers.csv` and `staff.csv` shipping
terminated and offboarded staff to Clever.

`rpt_clever__teachers` gated only on `uac_account_disable = 0`, with no worker
status filter, so a terminated employee kept shipping for as long as their AD
account stayed enabled. It was the outlier among the people feeds:

| feed                       | termination filter                    |
| -------------------------- | ------------------------------------- |
| `rpt_clever__staff`        | `worker_status_code != 'Terminated'`  |
| `rpt_clever__sections` DSO | `assignment_status != 'Terminated'`   |
| `rpt_clever__teachers`     | none, until this change               |

As of today that shipped **10 rows for terminated staff**, termination dates
reaching back to 2019. Two of the ten terminated today, which is ordinary
processing lag; the other eight are stale. Two point at live schools rather than
the District Office sentinel, so terminated staff held teacher records against
active school rosters.

The fix adds `worker_status_code != 'Terminated'` to the
`int_people__staff_roster` branch, matching `rpt_clever__staff`.

### Temp staff: also fixed here, after a correction

An earlier version of this description said `int_people__temp_staff` "carries no
worker-status or termination column (only `idauto_status` and
`uac_account_disable`)". **That was wrong**, and `claude-review` caught it. The
model carries `idauto_person_term_date` and `idauto_person_end_date`. My
`INFORMATION_SCHEMA` probe filtered column names on `%terminat%` / `%status%` /
`%disable%` / `%active%` — none of which match `term_date` or `end_date` — so a
too-narrow filter produced a false negative I then asserted as fact.

It was a live leak, not a theoretical one: of the 21 temp-staff rows in
`teachers.csv`, one carried a term date from October 2022 while `idauto_status`
was still `A` and the AD account still enabled. `int_people__temp_staff` gates
only on `idauto_status` and `uac_account_disable`, neither of which flips on
offboarding.

Both Clever temp branches are now filtered on `idauto_person_term_date is null`,
so `teachers.csv` and `staff.csv` cannot disagree about the same person.

Two scoping decisions worth stating:

- **Not filtered in `int_people__temp_staff` itself**, though that is the tidier
  altitude. That model also feeds `rpt_illuminate__roles`,
  `rpt_illuminate__users`, `rpt_powerschool__autocomm_teachers` and
  `rpt_littlesis__enrollments`. Changing four unrelated outbound integrations from
  inside a Clever-scoped PR is a worse trade than one duplicated predicate.
  Normalizing the string to a DATE upstream and applying it network-wide stays
  scoped in #4674.
- **Presence of a term date is the signal**, rather than parsing it and comparing
  to today. The field is sparsely maintained — 1 of 26 rows populated, and
  `idauto_person_end_date` entirely null — so a date comparison would add parsing
  plus a CTE restructure for no behavioral difference on any current row. Noted in
  #4674.

### Not fixed here, and not a data problem

All 10 removed rows have `uac_account_disable = 0`, meaning enabled accounts
after termination, one since 2019. That reaches beyond Clever to anything keyed
on the account, so it is a People Operations and IT question. Written up in
#4674 rather than papered over in dbt.

Found while reviewing the drop-off list from #4667, where one terminated staff
member was removed by the Miami region filter — incidentally, for the wrong
reason.

## AI Assistance

Claude-authored, human-directed. Charlie spotted that the person in question was
terminated, which is what surfaced the missing filter; Claude traced the cause,
quantified the population, and made the change.

### dbt

- [x] No new model, so no new properties file. No contract or column change —
      this is a `where`-clause predicate only.
- [x] No exposure change; this feeds the Dagster BigQuery-to-SFTP extract asset.
- [x] **Not a breaking change** — no columns added, renamed or removed. Rollback
      is a plain revert.

## Verification

Local dev build of all six clever models plus their tests, deferred to prod
upstreams:

```text
Done. PASS=11 WARN=0 ERROR=0 SKIP=0 NO-OP=0 TOTAL=11
```

Row-level diff of the current prod feed against the dev build, which confirms the
change is surgical:

| feed       | prod rows | dev rows | removed                                  |
| ---------- | --------: | -------: | ---------------------------------------- |
| `teachers` |     1,460 |    1,449 | 10 roster (10 people) + 1 temp (1 person) |
| `staff`    |     4,416 |    4,404 | 12 temp rows, one person across 12 schools |

Rows added: **0**. Terminated roster rows remaining in either feed: **0**. Temp
rows carrying a term date remaining in either feed: **0**.

NULL-safety of `!=` checked as well: **0 rows** in `int_people__staff_roster`
have a NULL `worker_status_code` while meeting this model's other predicates, so
the `!=` comparison drops nothing beyond the intended `Terminated` set.

`trunk check --force` on the changed file reports no issues.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes, zero warnings (run `70403219144192`).
- [x] **Dagster Cloud** — not triggered; the `pull_request` paths exclude
      `src/dbt/`.

## Re-review

The temp-staff commit landed after `claude-review` passed, and that check does not
re-fire on `synchronize`. Toggling draft state would re-trigger it if a second
pass is wanted before merge.

## Gate

`kipptaf__extracts__clever__asset_job_schedule` is confirmed still STOPPED. This
should land before it is switched on, otherwise the first hourly sync pushes all
10 terminated staff — including the two at live schools — plus the offboarded temp
staffer.

Closes #4674
Refs #4653

### DIFF
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__staff.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__staff.sql
@@ -38,7 +38,12 @@ with
             department as home_department_name,
             title as job_title,
         from {{ ref("int_people__temp_staff") }}
-        where dagster_code_location != 'kippmiami'
+        where
+            dagster_code_location != 'kippmiami'
+            -- int_people__temp_staff gates on idauto_status and the AD account
+            -- flag, neither of which flips on offboarding. A populated
+            -- idauto_person_term_date is the only termination signal it carries.
+            and idauto_person_term_date is null
     ),
 
     schools as (
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__teachers.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__teachers.sql
@@ -17,6 +17,10 @@ from {{ ref("int_people__staff_roster") }}
 where
     uac_account_disable = 0
     and not is_prestart
+    -- an enabled AD account does not imply active employment; offboarding can
+    -- leave the account enabled indefinitely, so gate on worker status the way
+    -- rpt_clever__staff does
+    and worker_status_code != 'Terminated'
     and employee_number is not null
     and home_work_location_powerschool_school_id is not null
     -- Miami rosters into Clever directly from Focus; excluded from all six feeds
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__teachers.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__teachers.sql
@@ -46,4 +46,9 @@ select
 
     null as `password`,
 from {{ ref("int_people__temp_staff") }}
-where dagster_code_location != 'kippmiami'
+where
+    dagster_code_location != 'kippmiami'
+    -- int_people__temp_staff gates on idauto_status and the AD account flag,
+    -- neither of which flips on offboarding. A populated
+    -- idauto_person_term_date is the only termination signal it carries.
+    and idauto_person_term_date is null


cbini:
## Review response

### Important 1 — correct, and the PR description is now fixed

Verified: `int_people__temp_staff` does carry `idauto_person_term_date` and
`idauto_person_end_date`. My claim that it carried neither was wrong.

The cause is worth naming, because it is a repeatable trap: I probed
`INFORMATION_SCHEMA.COLUMNS` filtering column names on `%terminat%`, `%status%`,
`%disable%` and `%active%`. None of those match `term_date` or `end_date`, so a
too-narrow filter returned a false negative that I then stated as fact instead of
listing the columns and reading them.

PR description corrected with the real reasoning. The temp branch is still
unfiltered here, on narrower grounds:

- Sparse: 1 of 26 rows in `int_people__temp_staff` has a `term_date`; 0 have an
  `end_date`.
- Badly typed for the purpose: `idauto_person_term_date` is a STRING in
  `M/D/YYYY`, `idauto_person_end_date` is `FLOAT64` and entirely null. Either
  needs parsing plus a data-quality call, not a one-line predicate.
- `rpt_clever__staff` also leaves its temp branch unfiltered. Filtering only here
  would make the two feeds disagree, which is the class of inconsistency this
  workstream exists to close.

**Your instinct that the failure class is structurally open was right.** Of the
21 temp-staff rows currently in `teachers.csv`, one carries a `term_date` from
October 2022 with `idauto_status = 'A'` and an enabled account. So this is a live
leak, not a theoretical one — just one that needs both feeds fixed together.
Folded into #4674 with scope rather than filed separately.

### Minor 1 — NULL semantics: verified, no change

You were right to flag the semantics and right that it needed data to settle. It
settles clean:

```text
0 rows in int_people__staff_roster have a NULL worker_status_code
while meeting this model's other predicates (1,439 rows meet them)
```

So `!=` drops nothing beyond the intended `Terminated` set. Leaving the plain
comparison rather than adding a `coalesce` guard — it matches the four sibling
files you identified, and hardening only this one would be inconsistent for no
present benefit. Added the verification to the PR body.

### Minor 2 — DRY: agreed, informational, no action

Confirmed there is no existing macro or `is_active_worker` boolean, so this
follows the repo's actual pattern rather than deviating. Not centralizing it here.
The analogous region-literal duplication from #4667 is already tracked in #4670;
if that refactor happens, the worker-status predicate is a natural second
candidate for the same treatment.

### On the test-coverage section

Agreed with your reasoning for not flagging a missing "no terminated rows" test —
`rpt_clever__staff` has carried this filter far longer at the same coverage tier,
so requiring one here would invent a requirement the tier does not carry.

PR #4677, opened 2026-07-31
fix(dbt): resolve the Focus CONTACT address from the guardian household linkage

### LINKED ISSUE #4651: fix: Focus CONTACT feed still sends an arbitrary Finalsite household address
#4613 fixed the arbitrary-household problem for the Focus `ADDRESS` feed only. The same root defect still governs the Focus `CONTACT` feed.

`src/dbt/kipptaf/models/extracts/focus/rpt_focus__contacts.sql` (lines 21-25) projects each guardian's `households[safe_offset(0)]` address into the CONTACT feed. Array position does not identify Finalsite's primary household — that designation is set in the Finalsite UI and is exposed by no API field — so whichever household sits at offset 0 becomes the contact's address in Focus.

Two consequences:

- The CONTACT feed is import-once with no overwrite path, same as ADDRESS, so a wrong pick becomes permanent.
- A family can now end up with a guardian CONTACT address in Focus that disagrees with the student's ADDRESS of record, because the two feeds resolve the address by different rules. Guardians are typically linked to more households than their children, so they are the records most exposed to an arbitrary pick.

Surfaced by the final review of the Phase 2 ADDRESS work (PR #4618). The plan for that work explicitly scoped this out — `rpt_focus__contacts` and `int_finalsite__student_contacts.household_address` are why `stg_finalsite__contacts` keeps its scalar `address_*` columns — so this is a deliberate follow-up, not a regression.

Worth deciding as part of any fix: whether a contact's address should be resolved by an analogous rule, inherited from the student's resolved address of record, or omitted from the CONTACT feed entirely. Related: #4616 (asserting the `is_primary` singleton at its source) and #4617 (students with no `primary` relationship).

cbini:
# Pull Request

## Summary & Motivation

...resolve each Focus `CONTACTS` guardian address from that guardian's own
Finalsite household linkage instead of an arbitrary array position, and put both
Focus address feeds on one shared resolution rule.

Closes #4651. Follows #4613 / PR #4618, which fixed the same defect for the
`ADDRESS` feed only.

### The problem

`rpt_focus__contacts` took each guardian's address from
`stg_finalsite__contacts`, whose scalar address columns flatten
`households[safe_offset(0)]`. Array position does not identify Finalsite's
primary household — that designation is set in the UI and exposed by no API
field. Whichever household sat at offset 0 became the contact's address in
Focus. The feed is import-once with no overwrite path, so a wrong pick is
permanent. Only 13 students are currently in Focus, so roughly 99% of the
population was still prospective.

### The change

New `finalsite` package model `int_finalsite__contact_address_of_record`, one
row per contact. A household is a candidate when it carries a street line;
address identity is an **exact match on the five mailing fields** — no
case-folding, punctuation-stripping, or ZIP+4 truncation; a contact resolves
only when exactly one distinct address remains, and the exported values are the
raw text of the lowest `household_id` in that group.
`int_finalsite__student_address_of_record` was refactored onto it so the two
feeds cannot drift apart again — that was the actual defect #4651 describes.

Two spellings of one address stay distinct candidates and the contact is
withheld rather than guessed at. A looser key was considered and rejected:
collapsing on `address_1` alone would gain about 12 rows from 8 guardians, of
whom 5 differ by city — that is choosing a household by surrogate-key order,
the same class of defect as `safe_offset(0)`.

Completeness is deliberately not a candidacy gate, so an address missing its
city, state, or ZIP can still resolve. Measured at the guardian-feed level that
change is a **no-op today: zero rows gained, zero lost.** It is a forward-looking
rule change, not one that rescues rows now. Households carrying only a
city/state/ZIP fragment with no street are still excluded — Miami holds 94 of
them, and each would otherwise count as its own candidate and manufacture
ambiguity that is not real.

### Measured against production

| Measure | CONTACTS | ADDRESS |
| --- | --- | --- |
| Rows or students exported | 2,601 rows / 2,071 with an address | unchanged |
| Exported values changed | 0 | 0 |
| Newly enabled | 42 | 0 |
| Newly withheld | 436 | 0 |

The `ADDRESS` feed is a genuine no-op: across all 3,428 student records there
are zero mailing-field changes, zero feed-membership changes, and zero phone
changes. Exactly one contact's internal `address_source` label flips, and
`rpt_focus__addresses` never observes it because that model only tests
`address_source is not null`.

### Scope note — this also touches the ADDRESS feed's gate

The kippmiami `rpt_focus__addresses` wrapper carried a #4320 gate requiring a
complete address, which silently defeated the loosened candidacy rule on that
feed. It is now a street-line gate. Zero rows change today. The advisory review
recommended keeping this in the same PR rather than splitting it, since it is a
three-line change sharing the root cause being fixed here; happy to split it out
if a reviewer disagrees.

### The trade being made, stated plainly

436 guardian rows lose an address they previously had. They are **not withheld**
— every contact row still ships with name, relationship, email, and phones; only
the address fields are blank. Because CONTACTS import-once keys on the student
rather than the contact, those blanks are permanent, exactly as the arbitrary
offset-0 addresses they replace were permanent. The judgement is that a blank
prompts a human while a wrong address looks correct.

225 students end up with no guardian contact row carrying an address. All 225
still receive an address through the `ADDRESS` feed on their own student record,
so **zero students end up with no address anywhere in Focus.**

### Advisory review findings

All four findings from the automated review were verified against the code
rather than taken on trust:

1. **Stale properties description (Important) — fixed.** The kippmiami
   `rpt_focus__addresses.yml` description still documented the completeness gate
   this PR deletes; `git log` confirmed no commit here had touched it. It now
   describes the street-line gate, and its unit test gained a case proving a
   street-only address is included.
1. **NULL field groups as a distinct candidate (Important) — real, zero
   incidence, now tested.** A household missing a ZIP does not match one that
   has it, so two households on the same street could count as two candidates.
   Measured against production: zero such contacts exist today. A unit test now
   documents the behavior deliberately rather than leaving it undefined.
1. **ZIP+4 exports raw (Minor) — pre-existing, not fixed here.** True before this
   PR and still true. Worth its own issue if Focus wants a fixed-width ZIP;
   widening this PR for it is not warranted.
1. **Doc lint unverified (Minor) — already satisfied.** `trunk check --force`
   was run on both new documents and is clean. The review sandbox had no trunk
   binary.

### Staging environment note for reviewers

`stg_finalsite__contacts` was missing the `households` array in all four
`zz_stg_*_finalsite` schemas — PR #4618 added that column but its staging copies
were never seeded, because it cloned the finished
`int_finalsite__student_address_of_record` table, which never needed the
upstream column. Newark, Camden, and Paterson were two further columns behind.
All four are now reseeded from prod, plus both address-of-record models built
into each. This was pre-existing breakage that any PR touching this chain would
have hit.

### AI Assistance

Claude Code implemented this under human direction, task by task, with a review
after each task and a whole-branch review before this PR. Human-directed: the
resolution rule (anchor on the guardian's own linkage; keep the key a strict
exact match; drop the completeness gate), the decision to loosen the kippmiami
ADDRESS gate rather than document around it, and accepting the permanent-blank
trade. AI-authored: the SQL, properties YAML, unit tests, ops documentation, and
the production measurements above.

### General

- [x] Review the **Claude Code Review** comment posted on this PR — all four
      findings verified and addressed or explained above.

### dbt

- [x] **Breaking change?** No columns renamed or removed. The 50-column
      `CONTACTS_LAYOUT` and 12-column `ADDRESS_LAYOUT` contracts are unchanged;
      only rows and values move. `int_finalsite__student_address_of_record`
      gains `is_complete_address`, an additive change on a non-contracted
      intermediate.
- [x] No external sources added or modified. The four district staging schemas
      were seeded manually — see the staging note above.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — expect a wide run. Editing the four `sources-kipp*.yml`
      files marks the whole finalsite source `state:modified`, so
      `state:modified+` fans out across every kipptaf model reading finalsite.
      Treat pre-existing warn-level test noise as pre-existing.
      `src/dbt/**`.

### DIFF
--- a/docs/reference/finalsite-focus-import.md
+++ b/docs/reference/finalsite-focus-import.md
@@ -163,13 +163,18 @@ does not already have it, and nothing is ever overwritten:
   drop code) is filled in once onto that student-year's **open** Focus
   enrollment when it has neither yet — never onto one Focus already shows as
   closed.
-- **Addresses and Contacts** — sent only once the student is **enrolled**, Focus
-  does not already have the record for that student, **and** Finalsite points to
-  a single address for them (a named contact, for Contacts). For Contacts, it's
-  the **student's** enrolled status that gates the feed, not the guardian
-  contact's own. An address Finalsite cannot narrow to one is held back; a
-  partial address is now sent rather than held, so it can be spotted and
-  corrected in Focus.
+- **Addresses** — sent only once the student is **enrolled**, Focus does not
+  already have the address record for that student, **and** Finalsite points to
+  a single address for them. An address Finalsite cannot narrow to one is held
+  back entirely; a partial address (missing city, state, or ZIP) is sent rather
+  than held, so it can be spotted and corrected in Focus.
+- **Contacts** — sent only once the student is **enrolled**, Focus does not
+  already have the record for that student, **and** the contact has a name. It
+  is the **student's** enrolled status that gates the feed, not the guardian
+  contact's own. A guardian's address does not gate whether the contact is sent:
+  when Finalsite can't narrow it to one, or the guardian has no street address
+  on file at all, the contact still goes out with the rest of their details and
+  the address is simply left blank.
 
 ### Forward-moving enrollments are protected
 
@@ -197,7 +202,8 @@ Because addresses and contacts are import-once, a record sent before Finalsite
 can resolve it would be locked in — a student imported with no address at all
 would keep that gap in Focus even after a real one is entered, because
 import-once never sends them again. To prevent that, the pipeline **holds a
-record back when it cannot tell which address to send**:
+student's address record back when it cannot tell which address to send**. A
+guardian's address works differently — see the Contacts bullet below.
 
 - **Addresses** — a student's address is sent once Finalsite points to a single
   address for them. The pipeline reads the households the student is linked to,
@@ -209,11 +215,19 @@ record back when it cannot tell which address to send**:
   back flows the first run Finalsite resolves to a single address.
 - **Contacts** — a contact is sent only once it has a name. A nameless contact
   is skipped and flows once the name is filled in. A guardian's address is
-  resolved the same way a student's is, but from the guardian's own households:
-  when Finalsite links them to more than one address, the contact is still sent
-  with the rest of their details and the address is left blank rather than
-  guessed. Guardians are usually linked to more households than their children,
-  so this is more common on the contact record than on the student.
+  resolved from the guardian's own households only, with no fallback: when
+  Finalsite links them to more than one address, or to no street address at all,
+  the contact still goes out with the rest of their details and the address is
+  simply left blank. Guardians are usually linked to more households than their
+  children, so this is more common on the contact record than on the student.
+
+  **A guardian's blank address is not held back the way a student's is — it is
+  permanent.** A student's address record waits until Finalsite resolves it,
+  then sends a real address the first run it can. A guardian's contact record
+  does not wait: it imports as soon as it has a name, blank address and all, and
+  once that student's contacts have imported, a later Finalsite fix — retiring
+  the extra household, or adding a street where none existed — is never re-sent.
+  A guardian's address gap has to be filled in Focus by hand.
 
 > **A student can be enrolled in Focus with no address yet.** That is expected
 > when Finalsite has no street address on file for them, when it has several and
--- a/docs/reference/finalsite-focus-import.md
+++ b/docs/reference/finalsite-focus-import.md
@@ -164,10 +164,12 @@ does not already have it, and nothing is ever overwritten:
   enrollment when it has neither yet — never onto one Focus already shows as
   closed.
 - **Addresses and Contacts** — sent only once the student is **enrolled**, Focus
-  does not already have the record for that student, **and** the record is
-  complete (a full address; a named contact). For Contacts, it's the
-  **student's** enrolled status that gates the feed, not the guardian contact's
-  own. Blank or incomplete records are held back until populated (see below).
+  does not already have the record for that student, **and** Finalsite points to
+  a single address for them (a named contact, for Contacts). For Contacts, it's
+  the **student's** enrolled status that gates the feed, not the guardian
+  contact's own. An address Finalsite cannot narrow to one is held back; a
+  partial address is now sent rather than held, so it can be spotted and
+  corrected in Focus.
 
 ### Forward-moving enrollments are protected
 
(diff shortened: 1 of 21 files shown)


cbini:
Verified each finding against the code before acting on it. Outcomes below, plus one change that came from outside the review.

**1. Stale `rpt_focus__addresses.yml` description (Important) — confirmed, fixed.**

Confirmed exactly as described: the description asserted the mailable address must be complete, and `git log 6702fede3..HEAD -- <that file>` returned nothing, so no commit in this PR had touched it. The description now describes the street-line gate. Its unit test `addresses_import_once_suppresses_existing` gained a case for a student whose address has a street but null city, state, and zipcode, asserting it is now included — the behavior this PR actually changes.

**2. NULL completeness field grouped as a distinct candidate (Important) — logic confirmed, zero incidence, now covered by a test.**

The reasoning holds. `GROUP BY` treats a NULL `zip` as its own group, so two households on one street where only one carries a ZIP count as two candidates and the contact goes ambiguous — where the old completeness gate would have excluded the incomplete one and resolved.

Measured against production rather than left open. Comparing the old rule (completeness-gated) to the new one (street-gated) across all Miami contacts: **1 contact regresses from resolved to ambiguous, and 0 of those are the same-street-differing-by-a-null-field case** you identified. The single regression is a household with a person's name typed into the street field and no city, state, or ZIP — two genuinely different streets, not a null-grouping artifact. At the guardian-feed level the gate removal is a wash: 0 rows gained, 0 lost.

Added `test_contact_address_null_field_is_a_distinct_candidate` documenting the behavior deliberately rather than leaving it undefined.

**3. ZIP+4 exported raw (Minor) — accurate, pre-existing, deliberately not fixed here.**

Correct that `int_finalsite__contacts__households` never truncates and the projected `zip` carries whatever Finalsite holds. Also correct that it is pre-existing rather than a regression. Leaving it: this PR is already touching two feeds, and a fixed-width ZIP question belongs with whoever owns the Focus import contract. Worth a separate issue.

**4. Doc lint unverified (Minor) — already satisfied.**

`trunk check --force --no-fix` was run on both new documents from inside the worktree; clean. Noted that your sandbox had no trunk binary.

---

**One change that supersedes part of the review.** After this review posted, the dedup key was reverted from a normalized comparison to a **strict exact match** on the five raw mailing fields — no case-folding, no punctuation-stripping, no ZIP+4 truncation. Any observation you made about normalization no longer applies to the current head. Consequences: the contact feed now carries 2,071 addresses rather than 2,081, and **zero exported values change on either feed** (previously two rows changed punctuation). The `ADDRESS` feed is now a strict no-op — zero mailing-field, membership, or phone changes across 3,428 student records.

**On your scope recommendation:** agreed, the kippmiami ADDRESS-gate change stays in this PR. The PR body flags it explicitly so it is not a surprise.

PR #4679, opened 2026-07-31
fix(focus): map Postgres interval columns to int64 microseconds

### LINKED ISSUE #4676: fix(focus): map Postgres interval columns so gradebook_assignments dlt load stops failing
## Problem

`kippmiami/dlt/focus/gradebook_assignments` fails deterministically. Focus added an `interval` column named `time_limit` to `public.gradebook_assignments` mid-day on 2026-07-31, and dlt cannot map it.

Bottom of the error chain:

```text
dlt.common.libs.pyarrow.UnsupportedArrowTypeException: Arrow type `duration[us]`
for field `time_limit` in table `gradebook_assignments` is unsupported by dlt
```

Failing runs (both on commit `c0663b51a`):

- [5baaace7-4a07-4dcf-a4bd-cbe110f4ac16](https://kipptaf.dagster.cloud/prod/runs/5baaace7-4a07-4dcf-a4bd-cbe110f4ac16) — 76 of 77 steps succeeded
- [4ed9f0fa-ab52-4d51-8f11-05d0cbb1863e](https://kipptaf.dagster.cloud/prod/runs/4ed9f0fa-ab52-4d51-8f11-05d0cbb1863e) — auto-retry, failed identically, `will_retry: false`

Not a code regression: the same asset succeeded at 08:04 UTC the same day (371,227 rows) on the same commit, then failed at 22:31 UTC. The BigQuery table has 39 columns and no `time_limit`.

## Root cause

1. Postgres `interval` reflects as `postgresql.INTERVAL`, which matches none of the branches in dlt's `sqla_col_to_column_schema`. It falls through to the `else`, logs a warning, and emits a column schema with no `data_type`.
2. With no type hint, `row_tuples_to_arrow` infers the arrow type from the Python `timedelta` values, producing `duration[us]`.
3. `_compute_tables` calls `py_arrow_to_table_schema_columns`, whose `get_column_type_from_py_arrow` has no `duration` branch, so it raises.

`time_limit` is consumed by nothing — `stg_focus__gradebook_assignments.sql` does not select it. This is purely an ingestion-layer type-mapping gap.

## Fix

Add a `type_adapter_callback` to `build_focus_dlt_assets` mapping `interval` to `sqltypes.BigInteger`, so the column lands as BigQuery INT64 microseconds. This is the same hook Illuminate already uses for `unbounded_numeric_adapter`. Focus currently passes only `table_adapter_callback=remove_nullability_adapter`; the two arguments are independent, so both coexist.

Generic by table and column, so any future Focus `interval` column is covered rather than just this one.

Verified locally against dlt's conversion path (the Focus DB itself is unreachable from a Codespace):

| adapter target | arrow type | value for a 45-minute interval | verdict |
| --- | --- | --- | --- |
| none (current) | `duration[us]` | n/a | reproduces the prod exception |
| `BigInteger` | `int64` | `2700000000` | lossless microseconds |
| `Text` | `string` | `'2700000000'` | microseconds as a string |
| `Time` | `time64[us]` | `time(0, 45)` | unsafe past 24h or negative |

`BigInteger` chosen: lossless, no 24-hour ceiling, and correct for negative intervals. `Time` would convert via a raw buffer reinterpret, so an out-of-range interval corrupts silently.

Gotcha worth recording: `isinstance(INTERVAL(), sqltypes.Interval)` is `False`. The correct check is `sqltypes._AbstractInterval`.

No manual table drop needed — `remove_nullability_adapter` makes the new column NULLABLE, and adding a NULLABLE column is permitted under the `replace` write disposition.

## Scope

- `src/teamster/libraries/dlt/focus/assets.py` — add the adapter, wire it into `sql_database`
- `src/teamster/libraries/dlt/focus/CLAUDE.md` — document the mapping and its microsecond unit

Deliberately out of scope: adding the same adapter to Illuminate (no `interval` column observed there), and adding `time_limit` to the dbt staging contract (no consumer asked for it).

cbini:
# Pull Request

## Summary & Motivation

...stop `kippmiami/dlt/focus/gradebook_assignments` from failing on every nightly run, by mapping Postgres `interval` columns to BigQuery INT64 microseconds.

Focus added an `interval` column named `time_limit` to `public.gradebook_assignments` mid-day on 2026-07-31. dlt cannot map it, so the load dies at the extract step:

```text
dlt.common.libs.pyarrow.UnsupportedArrowTypeException: Arrow type `duration[us]`
for field `time_limit` in table `gradebook_assignments` is unsupported by dlt
```

Not a code regression — the asset succeeded at 08:04 UTC and failed at 22:31 UTC on the same commit `c0663b51a`. Failing runs: [5baaace7](https://kipptaf.dagster.cloud/prod/runs/5baaace7-4a07-4dcf-a4bd-cbe110f4ac16) (76 of 77 steps succeeded) and its auto-retry [4ed9f0fa](https://kipptaf.dagster.cloud/prod/runs/4ed9f0fa-ab52-4d51-8f11-05d0cbb1863e), which failed identically with `will_retry: false`.

**Root cause chain**, traced through the installed dlt:

1. Postgres `interval` reflects as `postgresql.INTERVAL`, matching none of the branches in dlt's `sqla_col_to_column_schema`. It falls to the `else`, logs a warning, and emits a column schema with no `data_type`.
1. With no type hint, `row_tuples_to_arrow` infers the arrow type from the Python `timedelta` values, producing `duration[us]`.
1. `_compute_tables` calls `py_arrow_to_table_schema_columns`, whose `get_column_type_from_py_arrow` has no `duration` branch, so it raises.

**The fix** adds `interval_to_microseconds_adapter` as the Focus `type_adapter_callback` — the same hook Illuminate already uses for `unbounded_numeric_adapter`. Focus previously passed only `table_adapter_callback=remove_nullability_adapter`; the two arguments are independent, so both coexist. The adapter keys off the reflected type rather than a column name, so any future Focus `interval` column is covered rather than just this one.

`BigInteger` over the alternatives, measured against dlt's actual conversion path:

| adapter target | arrow type | 45-minute interval | verdict |
| --- | --- | --- | --- |
| none (current) | `duration[us]` | n/a | reproduces the prod exception |
| `BigInteger` | `int64` | `2700000000` | lossless microseconds |
| `Text` | `string` | `'2700000000'` | microseconds as a string |
| `Time` | `time64[us]` | `time(0, 45)` | unsafe past 24h or negative |

`Time` looks natural for a duration but dlt converts it by reinterpreting the underlying buffer, so an interval of 24 hours or more, or a negative one, corrupts silently. Microseconds are lossless for anything Postgres can hold. Consumers divide by `1e6` for seconds; the unit is documented in the docstring and in `focus/CLAUDE.md`.

Gotcha recorded in both the docstring and `focus/CLAUDE.md`: `isinstance(INTERVAL(), sqltypes.Interval)` is `False`. `_AbstractInterval` is the only base shared by `postgresql.INTERVAL` and `sqltypes.Interval`, so the obvious check silently matches nothing.

**No manual table drop needed** — `remove_nullability_adapter` makes the new column NULLABLE, and adding a NULLABLE column is permitted under the `replace` write disposition. `time_limit` is consumed by nothing today (`stg_focus__gradebook_assignments.sql` does not select it), so there is no downstream contract to update.

Deliberately out of scope: the same adapter for Illuminate (no `interval` column observed there) and adding `time_limit` to the dbt staging contract (no consumer asked for it).

Refs #4676

## AI Assistance

Claude-authored, human-directed. The user supplied the failing run URL; Claude did the diagnosis, chose the candidate set, and wrote the fix, tests, and docs. The user made three calls: `BigInteger` microseconds as the target type, issue-plus-worktree for branch setup, and skipping a formal brainstorming pass.

Claude verified the root cause by reproducing the exact prod exception locally (Focus itself is unreachable from a Codespace — IP allowlist) and confirmed all three candidate mappings against dlt's real conversion functions rather than reasoning about them.

**Not yet verified against the live Focus database.** The reproduction drives dlt's `sqla_col_to_column_schema` / `row_tuples_to_arrow` / `py_arrow_to_table_schema_columns` with a hand-built `INTERVAL` column, which is the code path that raised, but it does not prove the real `time_limit` column reflects as `postgresql.INTERVAL`. It is the only Postgres type that infers to `duration[us]`, so this is very likely, but a branch deployment materializing the asset would settle it — see CI checks below.

## Self-review

### General

      code location

**The first two are left unchecked because they cannot pass in a Codespace, not because they were skipped.** `kippmiami.definitions` fails at module load there regardless of this change — it resolves the Focus credential spec eagerly and the variable is unset locally, a pre-existing constraint documented in `src/teamster/CLAUDE.md`. Verified instead by:

- `uv run pytest tests/libraries/test_dlt_focus_type_adapter.py` — 8 passed, including a guard that reproduces the prod `UnsupportedArrowTypeException` and a case asserting 48-hour and negative intervals survive the mapping exactly.
- Calling `build_focus_dlt_assets` with the real `postgresql+psycopg` dialect: constructs cleanly, asset key `kippmiami/dlt/focus/gradebook_assignments`.
- Asserting `sql_database` propagates `type_adapter_callback` and `table_adapter_callback` together, with `reflection_level` and `backend` intact.
- `uv run pytest tests/libraries/` — 82 passed, 3 failed. All 3 failures are in `tests/libraries/powerschool/sis/odbc/` and reproduce identically on a clean `main` checkout, so they are pre-existing and unrelated.

No schedule, sensor, or new integration, so the Docs section does not apply. No dbt changes.

## CI checks

`trunk check --force` was run on all four changed files after the format hook: no issues. dbt Cloud selects zero models here.

**Recommended pre-merge validation.** This PR is opened ready-for-review specifically so the branch deployment builds. Focus is IP-allowlisted and unreachable from a Codespace, so materializing `kippmiami/dlt/focus/gradebook_assignments` in the branch deployment is the only way to confirm the real `time_limit` column loads. Note that branch-deployment dlt runs write to the **prod** dataset `dagster_kippmiami_dlt_focus` — no isolation — which here is the desired outcome, since it repairs the broken table directly.

## Troubleshooting

### DIFF
--- a/src/teamster/libraries/dlt/CLAUDE.md
+++ b/src/teamster/libraries/dlt/CLAUDE.md
@@ -27,6 +27,9 @@ backend.
 - Uses `reflection_level="full_with_precision"` + `remove_nullability_adapter`
   (forces all columns `NULLABLE` so upstream `NOT NULL` changes don't break the
   `replace` load — see `focus/CLAUDE.md`)
+- `interval_to_microseconds_adapter` maps Postgres `interval` to INT64
+  microseconds; without it dlt rejects the inferred `duration[us]` (see
+  `focus/CLAUDE.md`)
 - Factory:
   `build_focus_dlt_assets(sql_database_credentials, code_location, table_name)`
 
--- a/src/teamster/libraries/dlt/focus/CLAUDE.md
+++ b/src/teamster/libraries/dlt/focus/CLAUDE.md
@@ -48,6 +48,17 @@ obvious check silently matches nothing. It imports only from
 Unlike the nullability adapter, this one needs no table drop: the new column
 arrives NULLABLE, which `replace` may add.
 
+**An all-NULL `interval` column loads fine and is silently dropped**, so absence
+from BigQuery does NOT mean the column is new. pyarrow infers arrow `null`,
+which dlt maps to no `data_type` and omits from the destination; the load breaks
+only when the FIRST non-null value lands. That is why #4676's asset succeeded at
+08:04 UTC and failed at 22:31 on the same commit — one assignment populated
+`time_limit` at 13:23. Two consequences: don't date a schema change from the
+BigQuery column set, and other Focus tables may hold unpopulated `interval`
+columns that break on first use, which you cannot enumerate from BigQuery. Hence
+the adapter keys off the reflected type for every table rather than naming
+columns.
+
 ## Empty source tables
 
 The PyArrow backend writes NO BigQuery table for a 0-row source extract (the
--- a/src/teamster/libraries/dlt/focus/CLAUDE.md
+++ b/src/teamster/libraries/dlt/focus/CLAUDE.md
@@ -7,12 +7,12 @@ default).
 
 ## Differences from Illuminate
 
-| Aspect              | Illuminate                              | Focus                        |
-| ------------------- | --------------------------------------- | ---------------------------- |
-| Schema dimension    | Multi-schema (asset key includes it)    | Single `public` schema       |
-| Type adapters       | `unbounded_numeric_adapter`             | None                         |
-| Query callbacks     | `filter_date_taken_callback` (optional) | None                         |
-| Nullability adapter | `remove_nullability_adapter`            | `remove_nullability_adapter` |
+| Aspect              | Illuminate                              | Focus                              |
+| ------------------- | --------------------------------------- | ---------------------------------- |
+| Schema dimension    | Multi-schema (asset key includes it)    | Single `public` schema             |
+| Type adapters       | `unbounded_numeric_adapter`             | `interval_to_microseconds_adapter` |
+| Query callbacks     | `filter_date_taken_callback` (optional) | None                               |
+| Nullability adapter | `remove_nullability_adapter`            | `remove_nullability_adapter`       |
 
 ## Nullability adapter (required)
 
@@ -25,6 +25,28 @@ mode. BigQuery forbids both adding a `REQUIRED` column and relaxing an existing
 same fix Illuminate uses. Adding/removing this adapter against existing tables
 that already have `REQUIRED` columns requires dropping those tables first
 (`replace` repopulates them on the next run).
+
+## Interval adapter (required)
+
+Postgres `interval` matches none of the branches in dlt's
+`sqla_col_to_column_schema`, so the reflected column carries no `data_type`, the
+PyArrow backend infers `duration[us]` from the `timedelta` values, and the load
+dies with `UnsupportedArrowTypeException` (#4676 —
+`gradebook_assignments.time_limit`).
+`type_adapter_callback=interval_to_microseconds_adapter` declares `BigInteger`,
+so **every Focus `interval` column lands as BigQuery INT64 microseconds** —
+divide by `1e6` for seconds when consuming one downstream.
+
+Not `Time`: dlt converts duration to `time64` by reinterpreting the buffer,
+which silently corrupts intervals of 24 hours or more and negative intervals.
+
+`_AbstractInterval` is the isinstance check —
+`isinstance(postgresql.INTERVAL(), sqltypes.Interval)` is `False`, so the
+obvious check silently matches nothing. It imports only from
+`sqlalchemy.sql.sqltypes`.
+
+Unlike the nullability adapter, this one needs no table drop: the new column
+arrives NULLABLE, which `replace` may add.
 
 ## Empty source tables
 
--- a/src/teamster/libraries/dlt/focus/assets.py
+++ b/src/teamster/libraries/dlt/focus/assets.py
@@ -49,8 +49,9 @@ def interval_to_microseconds_adapter(col_type: TypeEngine) -> TypeEngine | None:
 
     ``BigInteger`` rather than ``Time``: dlt converts duration to ``time64`` by
     reinterpreting the underlying buffer, which silently corrupts intervals of
-    24 hours or more and negative intervals. Microseconds are lossless for any
-    interval Postgres can hold.
+    24 hours or more and negative intervals. int64 microseconds spans roughly
+    292,000 years — lossless for any realistic interval, though narrower than
+    Postgres ``interval``'s full +/-178,000,000-year range.
 
     Note the check is ``_AbstractInterval``, the only base shared by
     ``postgresql.INTERVAL`` and ``sqltypes.Interval`` —
--- a/src/teamster/libraries/dlt/focus/assets.py
+++ b/src/teamster/libraries/dlt/focus/assets.py
@@ -7,6 +7,9 @@ from dlt.common.configuration.specs import ConnectionStringCredentials
 from dlt.common.runtime.collector import LogCollector
 from dlt.destinations import bigquery
 from dlt.sources.sql_database import remove_nullability_adapter, sql_database
+from sqlalchemy import BigInteger
+from sqlalchemy.sql.sqltypes import _AbstractInterval
+from sqlalchemy.types import TypeEngine
 
 
 class FocusDagsterDltTranslator(DagsterDltTranslator):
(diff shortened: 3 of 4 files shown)


cbini:
Verified all three findings against the code before acting. Two were correct and are fixed in `bfc31ff66`; the third is accurate but deliberate, so I left it.

### 2. Losslessness claim — correct, fixed

Confirmed the overstatement by computing the bounds rather than reasoning about them:

| bound | value |
| --- | --- |
| int64 microseconds | 9,223,372,036,854,775,807 µs ≈ **292,271 years** |
| Postgres `interval` range | ±178,000,000 years |
| Python `timedelta.max` | 999,999,999 days ≈ 2,737,851 years |

So int64 µs is narrower than both, and "lossless for any interval Postgres can hold" was wrong as written. The docstring now states the ~292,000-year span and names the Postgres range it falls short of. No behavior change — immaterial for an assignment time limit, but a docstring should not assert more than is true.

### 3. No factory-wiring test — correct, and a sharper gap than described, fixed

This one is worth more than nice-to-have. Every existing test calls `interval_to_microseconds_adapter` directly, so **all of them would still pass if `build_focus_dlt_assets` dropped `type_adapter_callback=`** — the regression guard did not cover the wiring that actually makes the fix work in production.

My first attempt to reach the factory-built source failed: `AssetsDefinition` exposes no `dlt_source`, the op has none, and the decorated function's closure has no free variables. I nearly pushed back that any such test would be tautological — restating the arguments to `sql_database` rather than testing the factory. That was wrong: `dagster_dlt/asset_decorator.py:48` stores the source in the asset spec metadata under `META_KEY_SOURCE`, so `test_factory_wires_both_adapters_into_the_dlt_source` now asserts against the real factory output.

Proved it discriminates rather than passing vacuously — building the same source *without* the callback reports `type_adapter_callback` as `None`, so the assertion fails. It also pins that the interval adapter did not displace `remove_nullability_adapter`, since those two arguments are the thing most likely to be broken by a careless edit.

`defer_table_reflect=True` means no connection is opened, so this needs no live Focus database. Credentials use a password-free connection string so gitleaks has nothing to flag.

### 1. Private `_AbstractInterval` import — accurate, no change

Flagging it was right, and I want the tradeoff visible rather than buried in a docstring. I checked the public alternatives before settling on it:

- `sqlalchemy.types` does not re-export it — `hasattr(sqlalchemy.types, "_AbstractInterval")` is `False`, so there is no public import path.
- `postgresql.INTERVAL` alone is public but narrower: it would miss a generic `sqltypes.Interval`, and `isinstance(INTERVAL(), sqltypes.Interval)` is `False`, so neither public type covers both.
- `col_type.python_type is timedelta` is public and dialect-agnostic, but `python_type` raises `NotImplementedError` on some types, so it needs a `try`/`except` wrapper around every column — more moving parts than the isinstance check for the same result.

`_AbstractInterval` is the only base the two interval types share. `test_adapter_maps_generic_interval_to_bigint` exercises `sqltypes.Interval()` directly, so a SQLAlchemy release that restructures that hierarchy fails a test rather than silently matching nothing in prod.

### On the environment note

Since the review job had no `uv`, for the record: 9 tests pass locally (8 before this commit, plus the new wiring guard), `trunk check --force` is clean on all changed files, and the 3 failures in the full `tests/libraries/` run reproduce identically on a clean `main` checkout — they are in `powerschool/sis/odbc/` and unrelated.

### Still open, and it is the load-bearing assumption

The review correctly routed the live-Focus question to the branch deployment. Restating it so it does not get lost at merge: nothing verified so far proves the real `time_limit` column reflects as `postgresql.INTERVAL`. It is the only Postgres type that infers to `duration[us]`, so this is very likely, but the reproduction uses a hand-built `INTERVAL` column. Materializing `kippmiami/dlt/focus/gradebook_assignments` in the branch deployment is the only way to close that loop — and it repairs the prod table directly, since branch-deployment dlt writes to `dagster_kippmiami_dlt_focus` unisolated.

cbini:
## Verified against the live Focus database

The open item from the self-review is now closed. Materialized `kippmiami/dlt/focus/gradebook_assignments` in branch deployment `e75f2cc082c5503ecb4c19a1b849fd9b3f0722d6` (location `LOADED` at `bfc31ff66`):

[run 3308de25-b8f3-439f-b717-0a6a39e65d07](https://kipptaf.dagster.cloud/e75f2cc082c5503ecb4c19a1b849fd9b3f0722d6/runs/3308de25-b8f3-439f-b717-0a6a39e65d07) — **SUCCESS**, 371,228 rows loaded, 17s.

`time_limit` now exists in `dagster_kippmiami_dlt_focus.gradebook_assignments` as **INT64** at ordinal position 40. The prod table is repaired, since branch-deployment dlt writes to the prod dataset unisolated.

### The run corrected the root cause

Only **1 of 371,228** rows has a non-null `time_limit`, and it is `assignment_id` 423278, created **13:23 UTC** — between the 08:04 success and the 22:31 failure, and exactly the +1 row versus the last good load of 371,227.

That prompted a check I had not run, and it changes the mechanism. An all-NULL `interval` column does **not** fail:

| interval column state | inferred arrow type | dlt result |
| --- | --- | --- |
| all-NULL | `null` | no `data_type` → column **silently dropped**, load succeeds |
| one non-null value | `duration[us]` | raises `UnsupportedArrowTypeException` |

`get_column_type_from_py_arrow` handles `is_null` by returning `{}`, so dlt omits the column from the destination entirely. **So the trigger was the first non-null value, not the column being added.** `time_limit` may have existed in Focus Postgres for some time, silently dropped from every load — which is why BigQuery showed 39 columns and no `time_limit`, and why I originally read its absence as "the column is new."

Two consequences, now recorded in `focus/CLAUDE.md` (commit `e90428f18`):

1. A schema change cannot be dated from the BigQuery column set — an absent column may be present-but-unpopulated upstream.
1. Other Focus tables may hold unpopulated `interval` columns that will break on the first value a user enters, and that exposure **cannot be enumerated from BigQuery** because dropped columns leave no trace there.

The second point is why the adapter keying off the reflected type across every table matters more than I credited when I wrote it off as "generic rather than just this one" — it is pre-emptive coverage for landmines that are currently invisible.

### One honest limitation

The single real value is `0` microseconds, so **prod data does not exercise a non-zero conversion** — a zero interval maps to 0 under essentially any candidate, including the rejected `Time`. Value fidelity rests on the unit tests instead, which assert exact round-trips for 45 minutes (2,700,000,000 µs), 48 hours, and -30 minutes. What the prod run proves is that the real column takes the interval path and loads without error; what the tests prove is that the arithmetic is right.

PR #4682, opened 2026-08-01
fix(finalsite): pick the best address instead of withholding it

### LINKED ISSUE #4680: fix(dbt): pick the best household for the Finalsite address of record instead of withholding
## Problem

The Focus ADDRESS and CONTACTS feeds are import-once, so a missing address is permanent for that student. Measured against production on 2026-08-01, **158 KIPP Miami students would receive no address in either feed**.

Neither `contacts_csv` nor `addresses_csv` has ever materialized, so nothing is locked in yet. This is fixable before first delivery.

Breakdown of the 158:

| Cause | Students |
| --- | --- |
| Several candidate addresses, so the address is withheld | 100 |
| No Parent 1 designated | 38 |
| Parent 1 has no valid address | 20 |

Only the last two match the intended rule: a student should have no address only when they have no primary contact, or their primary contact has no valid address. The 100 are withheld purely because `int_finalsite__contact_address_of_record` requires exactly one distinct candidate and refuses to choose between them.

A related gap: the model resolves the student's own household linkage first and falls back to Parent 1. That ordering exists because the old withhold rule punished the parent's larger household count. Once the model picks, that rationale no longer applies.

## Change

Two commits, each independently revertable.

1. **`int_finalsite__contact_address_of_record` picks instead of withholding.** Among a contact's street-bearing households, choose by `is_complete_address` descending, then `household_id` ascending. `resolution_status` gains `picked` alongside `resolved` and `no_street`; `resolved` keeps its current meaning of exactly one candidate.

2. **`int_finalsite__student_address_of_record` reorders and widens.** Spine changes from `where is_primary` to contacts carrying a workflow status, so a student with no Parent 1 still resolves. Resolution order flips to Parent 1's household first with the student's own as fallback. A new `is_picked_address` boolean reports whether the address was chosen from several; `resolution_status` continues to name the tier.

Commit 1 lands first because the pick belongs in the model that owns address identity. It also un-blanks a large share of the guardian rows in the CONTACTS feed.

## Expected impact

| Metric | Before | After |
| --- | --- | --- |
| Miami students with no address anywhere | 158 | 46 |
| Students whose shipped address changes | n/a | 113 |
| Guardian rows with a blank address in CONTACTS | 436 | materially lower |

The 113 changes come from the tier reorder, for students whose own household and their Parent 1's household differ. Nothing has shipped, so these carry no import-once cost.

The remaining 46 are Finalsite data gaps: 20 families with no address recorded anywhere and 21 with no household at all. No code change reaches them; they need Ops cleanup in Finalsite.

## Safety bar

No student who has an address today may lose one. The only permitted address changes are the 113 from the tier reorder. Anything else is a defect.

Refs #4651

cbini:
# Pull Request

## Summary & Motivation

...finish the Finalsite address-of-record rework: the pipeline now picks the
most complete candidate address for both contacts and students instead of
withholding whenever more than one household is on file, and the student model
prefers the household linked to Parent 1, falling back to the student's own
household when Parent 1 has none. This validates the change against
production, updates the enrollment-team-facing doc, and closes #4680.

## AI Assistance

This PR (and the two it builds on) was authored by Claude Code end to end:
model logic, unit tests, validation queries, and this doc update. Human
direction set the requirements and reviewed each step.

## Validation

Built the full Focus chain into a dev schema (`--defer --favor-state` against
each project's prod manifest) across all four district packages plus the
kipptaf layer — all builds passed with zero errors or warnings.

Compared dev `rpt_focus__addresses` (kipptaf layer, where this change lands)
against the deployed `kipptaf_extracts.rpt_focus__addresses`:

- `newly_losing`: **0** (no student who has an address today loses one —
  the safety bar for an import-once feed)
- `address_changed`: **56** (at most 113 expected; these are the intended
  consequence of preferring Parent 1's household over the student's own)
- `newly_gaining`: 194 (students who previously had no resolvable address now
  get one)

For the Miami-specific stranded-student count (computed from the sendable set
the kippmiami wrapper produces, rewritten against the dev relations since that
wrapper always reads prod regardless of target): **45** students remain
stranded, down from a prod baseline of **158**. The remaining stranded
students are Finalsite data gaps (no usable street address for the student or
their Parent 1) that need Ops cleanup in Finalsite, not a further pipeline
change.

### dbt

- [x] No new models — column addition (`is_picked_address`) on existing
      `int_finalsite__student_address_of_record`
- [x] Existing properties files updated with the new column's description
- N/A breaking change — this widens a spine and changes which address is
      picked; it does not remove or rename a contracted column

### Docs

- [x] `docs/reference/finalsite-focus-import.md` rewritten to describe
      picking-the-most-complete-address behavior in place of the old
      withhold-on-ambiguity behavior, across all four passages the
      requirements brief called out plus additional stale prose surfaced by
      grepping for "ambiguous" / "more than one address" / "several"

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — see note below; this PR alone triggers a real
      `state:modified+` rebuild in kipptaf because the column ADD marks the
      union wrapper modified.
- [ ] **Dagster Cloud** — not expected to trigger (no code-location paths
      touched by this PR beyond dbt)

## Staging-seed requirement (needs authorization)

`is_picked_address` is a column ADD on `int_finalsite__student_address_of_record`,
a package model that the kipptaf union wrapper reads via `source()`. dbt Cloud
CI reads the `zz_stg_<district>_finalsite` copies, which will not carry the new
column until each district's staging copy is rebuilt. That rebuild is:

```text
dbt build --select int_finalsite__student_address_of_record --target staging
```

...run once per district (`kippmiami`, `kippcamden`, `kippnewark`,
`kipppaterson`). This recreates shared `zz_stg` relations, so it needs explicit
authorization from a maintainer before it's run, and it must happen before
merge.

**Green CI on this PR is not evidence the column landed.**
`dbt_utils.union_relations` resolves its column list at compile time from the
`zz_stg_kipp*_finalsite` copies, and since nothing downstream selects
`is_picked_address` yet, a stale (unrebuilt) copy makes the union silently drop
the column rather than fail — the PR-built relation would compile to 15
columns instead of 16, and CI goes green regardless. The real check is
querying `INFORMATION_SCHEMA.COLUMNS` on the PR's built
`int_finalsite__student_address_of_record` relation and confirming it reports
16 columns, not just that the build succeeded.

Closes #4680

### DIFF
--- a/docs/reference/finalsite-focus-import.md
+++ b/docs/reference/finalsite-focus-import.md
@@ -163,18 +163,22 @@ does not already have it, and nothing is ever overwritten:
   drop code) is filled in once onto that student-year's **open** Focus
   enrollment when it has neither yet — never onto one Focus already shows as
   closed.
-- **Addresses** — sent only once the student is **enrolled**, Focus does not
-  already have the address record for that student, **and** Finalsite points to
-  a single address for them. An address Finalsite cannot narrow to one is held
-  back entirely; a partial address (missing city, state, or ZIP) is sent rather
-  than held, so it can be spotted and corrected in Focus.
+- **Addresses** — sent only once the student is **enrolled** and Focus does not
+  already have the address record for that student. The address itself comes
+  from the households the student's Parent 1 is linked to, falling back to the
+  student's own households when Parent 1 has none; when Finalsite points to
+  several, the most complete one is sent rather than none. A household with no
+  street line is not treated as an address at all; a partial address (missing
+  city, state, or ZIP) is sent rather than held, so it can be spotted and
+  corrected in Focus.
 - **Contacts** — sent only once the student is **enrolled**, Focus does not
   already have the record for that student, **and** the contact has a name. It
   is the **student's** enrolled status that gates the feed, not the guardian
   contact's own. A guardian's address does not gate whether the contact is sent:
-  when Finalsite can't narrow it to one, or the guardian has no street address
-  on file at all, the contact still goes out with the rest of their details and
-  the address is simply left blank.
+  when Finalsite links the guardian to several addresses the most complete one
+  is sent, and the address is left blank only when the guardian has no street
+  address on file at all — either way the contact still goes out with the rest
+  of their details.
 
 ### Forward-moving enrollments are protected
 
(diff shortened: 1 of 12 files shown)


cbini:
## Verification record

**Green CI alone did not prove this change works.** On the first run, `dbt_utils.union_relations` resolved its column list at compile time from the `zz_stg_kipp*_finalsite` copies, which lacked `is_picked_address`. Nothing downstream selects that column, so it was **silently dropped** and CI passed with 15 columns instead of 16.

Fixed by rebuilding all four district staging copies (`dbt build --select int_finalsite__student_address_of_record --target staging`, once per district). Each went from 13 to 14 columns. After re-firing CI:

```text
dbt_cloud_pr_70403104388001_4682_finalsite.int_finalsite__student_address_of_record
  16 columns, is_picked_address present
```

7/7 checks pass and `get_job_run_error(warning_only=true)` reports zero warnings.

Note for anyone re-running the staging seeds: each district build emits one WARN on the `address_source implies address_1` invariant. That is a staging-**data** artifact, not a code defect. The staging contact model is still the old cloned withholding logic, which leaves 511 rows with `candidate_count >= 1` and a null address — exactly the `picked` count. The new student model sets `address_source` for those, tripping the invariant. The rebuilt dev contact model has 0 such rows, CI builds the contact model fresh because it is `state:modified`, and prod deploys both together.

## Validation

Re-derived independently from production tables, calibrated by reproducing the deployed feed at 1264/1264 rows and the 158 stranded baseline exactly.

| Metric | Before | After |
| --- | ---: | ---: |
| Miami students with no address anywhere | 158 | 45 |
| Students losing an address | n/a | 0 |
| Students gaining an address | n/a | 194 |
| Students whose address changes | n/a | 56 |
| Guardian rows with a blank address, CONTACTS | 526 | 84 |

The 56 changes come from the tier reorder. Of the 113 students whose own household and their Parent 1's resolve differently, 37 are prospects the feed filters out and 20 had no prior address so they count as gains — 56 + 20 + 37 = 113, all accounted for.

## Review triage

`claude-review` ran against `1d4caa451`, before the fix wave in `c87c70f3b`.

- **Stale kipptaf `ambiguous` descriptions** — already fixed by that commit. A grep at HEAD finds only the ordinary word "unambiguous".
- **A NULL `status` would be silently dropped from the spine** — real and correctly reasoned: `status != 'not_in_workflow'` evaluates to NULL rather than TRUE, so such a row leaves the spine with no signal. Verified as non-occurring: **0 NULL values across all four districts, 42,547 contacts**. Shipping as-is by decision; no `not_null` test currently guarantees it stays that way, so this is a known latent fragility rather than an oversight. Worth revisiting if a NULL ever appears.
- **No model-level test tying `is_picked_address` to `address_source`** — deferred. The invariant is covered by unit tests, not a generic test.

PR #4684, opened 2026-08-01
feat(dbt): add teacher manager and cumulative GPA slide columns to rpt_tableau__student_course_grades

### LINKED ISSUE #4683: feat(dbt): add teacher manager and cumulative GPA slide flag to rpt_tableau__student_course_grades
## Summary

Three additions to `rpt_tableau__student_course_grades`, supporting the school-level Grade and Failure Analysis dashboard.

1. Teacher manager columns, so a teacher list can be filtered by manager.
2. A flag for whether a student's projected unweighted cumulative GPA sits at least one KIPP GPA Band below their actual unweighted cumulative GPA at the end of the prior year.
3. The raw signed difference behind that flag, for roster-level filtering.

## 1. Teacher manager

The model already joins `int_people__staff_roster` as `r` on `teachernumber = r.powerschool_teacher_number`. Add two columns from that existing join, following the pattern in `rpt_tableau__pm_overall_scores`, `rpt_tableau__schoolmint_grow_goals` and `rpt_tableau__leadership_development`:

- `reports_to_formatted_name as manager`
- `reports_to_sam_account_name as report_to_sam_account_name`

No new join required.

Note that `rpt_tableau__ddi_dashboard` is not the pattern to copy here. It uses `reports_to_sam_account_name as homeroom_teacher_name`, which is a different concept.

## 2 and 3. Cumulative GPA slide

Compares two values:

- **A, projected.** `cumulative_y1_gpa_projected_unweighted`, already on the model.
- **B, prior-year actual.** `cumulative_y1_gpa_unweighted` from `int_powerschool__gpa_cumulative_year` at the prior academic year, where `is_projected` is false. This is a new join.

Proposed columns, names to be settled in review:

```text
cumulative_y1_gpa_unweighted_prior_year  float64  B
gpa_band_projected                       int64    1-5, band of A
gpa_band_prior_year                      int64    1-5, band of B
gpa_band_change_from_prior_year          int64    gpa_band_projected - gpa_band_prior_year
is_gpa_band_slide                        boolean  gpa_band_change_from_prior_year <= -1
gpa_unweighted_change_from_prior_year    float64  A - B, negative means falling
```

### Band definition

KIPP GPA Band, five bands, as documented in `src/dbt/kipptaf/models/students/CLAUDE.md`. **This is the first implementation of these bands anywhere in the repo**, so it sets precedent.

```text
5   3.50 and above
4   3.00 to 3.49
3   2.50 to 2.99
2   2.00 to 2.49
1   below 2.00
```

The documented ceiling for band 5 is 4.00, but unweighted GPA reaches 4.33. Band 5 is implemented open-ended so values above 4.00 are not left unbanded.

The KTAF four-band system also applies to unweighted GPA and was considered. It collapses everything above 3.00 into one band, so a student sliding from 3.60 to 3.10 would not be flagged. Five bands were chosen for sensitivity at the top, where the 3.0-goal population sits.

### Grain and persistence

Both sides are student-level. Joining on `studentid` plus `_dbt_source_project` replicates the new columns across every quarter row and the Y1 row, which satisfies the filtering requirement without extra work.

Consistent with every existing `cumulative_*` column on this model, the new columns are NULL on prior-year rows.

## Verified before writing

- `int_people__staff_roster` carries `reports_to_formatted_name`, `reports_to_sam_account_name`, `reports_to_user_principal_name` and `reports_to_employee_number`.
- `cumulative_y1_gpa_projected_unweighted` is populated on 35,544 of 49,138 AY2026 Q1 rows and already repeats across Q1 and Y1.
- `int_powerschool__gpa_cumulative_year` holds one row per student-year with an `is_projected` flag. AY2025 has 6,237 rows, all not projected, all with unweighted populated. AY2026 holds the projected rows.
- Neither documented band system is implemented anywhere in the repo today.

## Notes for implementation

- The model is contract-enforced, so each new column needs a `properties.yml` entry with `data_type` and `description`.
- The model has a live Tableau consumer but no dbt exposure. Adding one is worth considering alongside this change.
- Validation runs against AY2026 as the current year, since the projected side only exists for the current year.

## Out of scope

Weighted band equivalents, and the KTAF four-band system.

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will add eight columns to
`rpt_tableau__student_course_grades`, supporting the school-level Grade and
Failure Analysis dashboard.

**Teacher manager**, from the staff-roster join the model already makes — no new
join:

- `manager`
- `report_to_sam_account_name`

**Cumulative GPA slide**, comparing the in-progress projected unweighted
cumulative GPA against the student's actual unweighted cumulative GPA at the end
of the prior year:

- `cumulative_y1_gpa_unweighted_prior_year`
- `gpa_band_projected_unweighted`
- `gpa_band_unweighted_prior_year`
- `cumulative_y1_gpa_unweighted_change_from_prior_year`
- `gpa_band_change_from_prior_year`
- `is_gpa_band_slide`

Bands use the KIPP Foundation five-band unweighted scale documented in
`models/students/CLAUDE.md`. `rpt_tableau__gpa_cumulative_year` already bands on
the same thresholds as string labels (`gpa_band_label`); these are integer ranks
because `gpa_band_change_from_prior_year` is arithmetic and string labels cannot
be subtracted. Band 5 is open-ended rather than capped at the
documented 4.00, because unweighted GPA reaches 4.33 and a closed bound would
leave roughly 1,574 rows unbanded.

The prior-year baseline comes from a new join to
`int_powerschool__gpa_cumulative_year` filtered to `not is_projected`, which is
the actual end-of-year value rather than a projection of it. Both sides are
student-grain, so every new column repeats across each quarter row and the Y1
row and is filterable at any marking period.

### Verification

- **No fan-out.** Duplicate groups on the model's composite key are unchanged
  from prod — 12 groups, 36 extra rows, both before and after. That is the
  pre-existing `#3915` storedgrades warn, not a regression.
- **Contract passes.** `dbt build --target dev` creates the view with all eight
  columns declared.
- **Band arithmetic validated** on the AY2024-to-AY2025 pair, where real
  year-over-year movement exists: every negative band change flags, every flat
  or positive change does not, and the 211 students missing either band return
  null rather than false.

### Two things a reviewer should know

**The flag reads all-false today.** AY2026 has no posted grades, so the
projection still equals the prior-year actual for essentially every student —
35,458 of 35,538 populated rows have a change of exactly zero, and the remaining
80 differ by 0.01 from rounding drift between the two source models. It begins
producing signal once Q1 grades post. Nothing to fix; worth not mistaking for a
broken join.

**A band comparison cannot see movement inside bands 1 and 5**, which are
open-ended. A student falling from 1.99 to 0.10 never trips
`is_gpa_band_slide`. That is inherent to banding, and it is why
`cumulative_y1_gpa_unweighted_change_from_prior_year` ships alongside the flag
rather than behind it.

### Known gap, deliberately not in this PR

This model is read by two Tableau dashboards but has **no dbt exposure** — the
existing `gradebook_and_gpa_dashboard` exposure covers the older pre-rebuild
models. Adding one is worth doing, but the consuming workbook is not published
yet so there is no URL to point at. Flagged in #4683 rather than guessed at
here.

## AI Assistance

Claude Code authored the SQL, properties entries, and validation queries. The
three requirements, the choice of the five-band scale over the KTAF four-band
scale, and the decision to scope this to buildable-today columns were
human-directed.

### General

### dbt

- [x] Properties file updated with `data_type` and `description` for all eight
      new columns
- [ ] Exposure — not added; see "Known gap" above
- [x] **Not a breaking change.** Additive only; no renames or removals, so
      existing consumers are unaffected

## CI checks

- [ ] **Trunk** — `trunk check --force` clean on both changed files before push
- [ ] **dbt Cloud**
      `pull_request` paths

Refs #4683

### DIFF
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
@@ -295,6 +295,44 @@ models:
         description: >-
           True when `gpa_needed_for_cumulative_3_0` is at or below the
           credit-weighted max achievable this year (current-year rows only).
+      - name: cumulative_y1_gpa_unweighted_prior_year
+        data_type: float64
+        description: >-
+          Final unweighted cumulative GPA at the end of the prior academic year,
+          the actual value rather than a projection of it. Student-grain, so it
+          repeats across every quarter and the Y1 row. Current-year rows only.
+      - name: gpa_band_projected_unweighted
+        data_type: int64
+        description: >-
+          KIPP GPA Band of `cumulative_y1_gpa_projected_unweighted`, on the
+          five-band unweighted scale — 5 at or above 3.50, 4 from 3.00, 3 from
+          2.50, 2 from 2.00, 1 below 2.00. Band 5 is open-ended, since
+          unweighted GPA exceeds the scale's documented 4.00 ceiling. NULL when
+          the projection is NULL.
+      - name: gpa_band_unweighted_prior_year
+        data_type: int64
+        description: >-
+          KIPP GPA Band of `cumulative_y1_gpa_unweighted_prior_year`, on the
+          same five-band unweighted scale as `gpa_band_projected_unweighted`.
+      - name: cumulative_y1_gpa_unweighted_change_from_prior_year
+        data_type: float64
+        description: >-
+          Projected unweighted cumulative GPA minus the prior year's final
+          unweighted cumulative GPA. Negative means the projection sits below
+          where the student finished last year.
+      - name: gpa_band_change_from_prior_year
+        data_type: int64
+        description: >-
+          `gpa_band_projected_unweighted` minus
+          `gpa_band_unweighted_prior_year`. Negative means the student is
+          projected into a lower band than they finished last year.
+      - name: is_gpa_band_slide
+        data_type: boolean
+        description: >-
+          True when the projected unweighted cumulative GPA falls at least one
+          KIPP GPA Band below where the student finished the prior year. NULL
+          when either band is missing, since an unbanded student is unknown
+          rather than known to be holding steady.
       - name: sectionid
         data_type: int64
         description: PowerSchool section id of the course enrollment.
@@ -331,6 +369,16 @@ models:
       - name: teacher_tableau_username
         data_type: string
         description: Teacher Tableau username (row-level security).
+      - name: manager
+        data_type: string
+        description: >-
+          Formatted name of the teacher's manager, from the staff roster's
+          reports-to relationship. NULL when the teacher has no roster match.
+      - name: report_to_sam_account_name
+        data_type: string
+        description: >-
+          Account name of the teacher's manager, for filtering a teacher list to
+          one manager's reports.
       - name: tutoring_nj
         data_type: boolean
         description:
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
@@ -301,6 +301,11 @@ models:
           Final unweighted cumulative GPA at the end of the prior academic year,
           the actual value rather than a projection of it. Student-grain, so it
           repeats across every quarter and the Y1 row. Current-year rows only.
+          Cumulative GPA accumulates per student-school upstream, which splits
+          the middle- and high-school series — for a grade 9 student this
+          baseline is their middle-school cumulative, not a high-school one, so
+          any comparison against it spans two series rather than measuring
+          movement within one.
       - name: gpa_band_projected_unweighted
         data_type: int64
         description: >-
@@ -332,7 +337,12 @@ models:
           True when the projected unweighted cumulative GPA falls at least one
           KIPP GPA Band below where the student finished the prior year. NULL
           when either band is missing, since an unbanded student is unknown
-          rather than known to be holding steady.
+          rather than known to be holding steady. Bands 1 and 5 are open-ended,
+          so movement inside them is invisible here — a student falling from
+          1.99 to 0.10 never trips this. Read it alongside
+          `cumulative_y1_gpa_unweighted_change_from_prior_year`, which does not
+          have that blind spot. Grade 9 carries the middle-to-high-school series
+          caveat on `cumulative_y1_gpa_unweighted_prior_year`.
       - name: sectionid
         data_type: int64
         description: PowerSchool section id of the course enrollment.
--- a/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__student_course_grades.sql
+++ b/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__student_course_grades.sql
@@ -75,6 +75,27 @@ with
         from prior_year_gpa_rollup
     ),
 
+    prior_year_cumulative as (
+        /* prior-year FINAL unweighted cumulative GPA — the baseline the
+           in-progress projection is measured against. is_projected is true only
+           for the year in progress, so filtering it out yields the actual
+           end-of-year value rather than a projection of it.
+
+           Grain is one row per student per year from AY2023 on. AY2019-AY2022
+           carried a second row for 11-29 students a year (dual-school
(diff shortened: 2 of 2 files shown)


anthonygwalters:
Verified each finding against the code and the warehouse before acting. Two implemented, two dismissed with evidence, two acknowledged. Pushed in 4cf8f303d.

## 2 — Correct about the comment, and it led somewhere better

You're right that the inline comment's safety claim is wrong, and the reasoning is right: upstream's key is `(studentid, schoolid, academic_year)` at `severity: error`, so two rows for one student-year with different schoolids are legal there; and this model's own key test is `severity: warn` for the unrelated #3915 storedgrades issue, so a new fan-out would inflate the warn count rather than fail CI. "Catches that loudly" was an overclaim. Rewritten to say exactly that and to point at a prod count comparison instead of a green build.

For the live risk specifically: AY2025 — the year the CTE reads — has 6,237 rows against 6,237 distinct `(studentid, _dbt_source_project)` pairs, so zero duplicates. AY2023 and AY2024 are also zero. The 6,237-vs-6,031 gap that looks like duplication is `studentid` colliding across districts, which the `_dbt_source_project` join key already resolves. So no fan-out today, and the dev build confirms it: 12 duplicate groups / 36 extra rows, identical to prod.

**Chasing your point surfaced something neither of us had flagged.** The same per-school accumulation splits the middle- and high-school series — the upstream description says so outright. A grade 9 row therefore pairs a *high-school* projection against a *middle-school* cumulative. Today that is invisible because the projection is null for 580 of 633 grade 9 students, so the flag reads null; once high-school grades post, the comparison spans two series rather than measuring movement within one. Documented on both affected column descriptions rather than silently excluding grade 9, since whether they belong in this monitoring view is a product call, not a code one. Thanks for the thread that led there.

## 3 — Correct, and my PR body was wrong

`rpt_tableau__gpa_cumulative_year.sql` does already carry `gpa_band_label` and `gpa_band_as_of_today_label` on identical thresholds. My "first implementation in the repo" claim was wrong, and I can see why I missed it: I grepped for the label strings as written in `students/CLAUDE.md` (`2.50 to 2.99`, `< 2.00`) and that model emits different ones (`2.5-2.99`, `below 2.0`). PR body corrected.

Keeping the integer form here rather than reusing the strings, deliberately: `gpa_band_change_from_prior_year` is arithmetic, and you cannot subtract `'3.0-3.49'` from `'3.5+'`. The thresholds match the existing model exactly, so the two agree on band boundaries; only the output type differs. Consolidating both onto a shared macro is worth doing, but it would touch a model outside this PR's scope, so I would rather it be its own change than a drive-by.

## 1 — Dismissing. The build disproves it.

The backtick rule in `src/dbt/CLAUDE.md` applies to **plain** scalars, where a leading backtick is a reserved YAML indicator. These descriptions are folded block scalars (`>-`), where content is taken literally. Three pieces of evidence:

- There are **14 existing folded descriptions in this repo whose first content line starts with a backtick** — `dim_assessments.yml`, `int_assessments__assessments_canonical.yml`, `int_assessments__assessments_members.yml` among them. All shipped.
- A local `dbt build --target dev` created the view successfully. A YAML parse failure would have stopped it before the model ran.
- **dbt Cloud CI passed on this PR** with the file as-is. A parse break is not survivable there either.

Happy to reword for consistency if the team prefers, but as a Critical build-breaker it is a false positive.

## 6 — Dismissing. I ran the linter you could not.

`.trunk/tools/trunk check --force` on both changed files from inside the worktree: **no issues**, before and after this commit. sqlfluff does not flag ST06 on the placement. The `rpt_tableau__iready_apm.sql` precedent is real but is not the only valid ordering — the linter is the authority here and it is satisfied.

## 4 — Correct, fixed

Eight columns, not seven. PR body updated.

## 5 — Fair, and out of scope by agreement

Agreed on the substance: `manager` and `report_to_sam_account_name` are staff-identifying, and this model carries untagged student and staff PII that predates the PR. Also agreed it is not this PR's gap to close. A `contains_pii` pass on this model needs a scope decision — direct identifiers only, or direct plus indirect — which root `CLAUDE.md` says to confirm with the team rather than assume.

## Note on CI

dbt Cloud passed with one warning: the `#3915` uniqueness warn at 12 results. That is unchanged from prod, which is the check that matters for finding 2.

anthonygwalters:
@claude review

Issue #4340, opened 2026-07-07
feat(kipptaf): rebuild gradebook GPA and grades Tableau views, decoupled from audit


anthonygwalters:
## Goal

Replace the GPA/grades side of the gradebook Tableau reporting family with two new views built for the from-scratch dashboard rebuild, fully decoupled from the gradebook-audit/fidelity lineage (no assignment-level data, no audit flags, no comment/citizenship columns).

## New models (both in `models/extracts/tableau/`)

1. **`rpt_tableau__student_course_grades`** — the main flat extract. Grain: one row per student x term x course section x grade category (long categories). Window: current + prior academic year (fall back to current-only if the extract bloats; ~1.8M rows projected for 2 years). Carries term grades, category grades, term GPA, GPA lookback (1/2/4-week), cumulative + needed-GPA columns, the full demographics block from today's `rpt_tableau__gradebook_gpa` plus `student_slideback`, and `need_60/70/80/90`. Citizenship and comment columns are dropped (they live on the audit/comments side).
2. **`rpt_tableau__gpa_cumulative_year`** — small companion, one row per student x academic year, over `int_powerschool__gpa_cumulative_year` (#4295) plus that year's enrollment attributes. Related to the main extract in Tableau on `student_number` (single Tableau data source; year-over-year cumulative trend without forcing year grain into the main view).

## Key decisions (from design session)

- One flat Tableau data source, not related tables, for the main extract (LOD/window-calc limitations with relationships); the year-grain companion is the one low-risk relationship.
- Prior-year rows source quarter/Y1 grades from `stg_powerschool__storedgrades` (`base_powerschool__final_grades` is current-year only) and categories from `int_powerschool__category_grades` (history verified present).
- Cumulative, needed-GPA, and lookback columns are as-of-today student-grain measures: populated on current-year rows only, NULL on prior-year rows.
- Paterson included, gated on an implementation-time check that Paterson gradebook data is populated.
- Existing four GPA-family models are untouched and retire with the old dashboard.

## Dependencies and sequencing

- **Hard dependency: PR #4316** (`int_powerschool__gpa_term_lookback`) must merge before this ships — the lookback columns come from that model.
- **Not a dependency: PR #4132** (gradebook audit revamp). Explicit follow-up commitment: once #4132 merges, re-platform the main view's enrollment spine onto `int_extracts__course_enrollments_by_term` (extended to the 2-year window). A follow-up issue gets opened when #4132 lands.

## Out of scope

[...]

PR #4686, opened 2026-08-01
feat(dbt): add need_next columns to student course grades extract

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will consolidate the `need_60` / `need_70` /
`need_80` / `need_90` ladder into a single "what do you need for the next
letter" answer, without removing the four. Three columns:

| Column | Type | What |
| --- | --- | --- |
| `need_next_letter_grade` | string | next WHOLE letter above the student's in-progress Y1 standing |
| `need_next_cutoff_percent` | float64 | that letter's cutoff on the course's own scale |
| `need_next` | float64 | percent required this term to get the Y1 grade there |

Whole letters only, A/B/C/D/F, per an explicit product decision — a student at
`C-` and a student at `C+` are both working toward the same `B`.

### Why the rungs come from the grade scale rather than a fixed ladder

The scales genuinely differ. Newark Q1 alone spans five:

| Scale | Rows | Quirk |
| --- | ---: | --- |
| KIPP NJ 2019 (5-12) Unweighted | 18,801 | full ladder, `D` at 63 |
| NCA 2011 | 1,465 | **no `A-`**, `D` at 60 |
| KIPP NJ 2016 (5-12) | 1,425 | **no `D+`/`D-`** — `D` at 60, jumps to `C-` |
| KIPP NJ 2019 Weighted | 759 | full ladder |
| KIPP NJ 2024 Weighted-Honors | 250 | no `F*` |

Roughly one Newark row in seven is on a non-default scale, so a hardcoded
60/70/80/90 `CASE` would be quietly wrong for them — the same class of error as
the `= 'F'` bug fixed in #4685.

### Two things this deliberately does not require

**No `src/dbt/powerschool/` change.** `courses_gradescaleid` is already
projected on `base_powerschool__final_grades`, so the extract passes it through,
and `int_powerschool__gradescaleitem_lookup` already exists at kipptaf level
carrying `_dbt_source_project`.

**No new point columns.** `need_*` is affine in its target percent — it is
`(points_still_needed * target - points_banked) / (term_points / 100)`, whose
other three terms are row constants — so the need for any target is exactly
recoverable from two of the four existing columns:

```text
need(t) = need_60 + (t - 60) / 10 * (need_70 - need_60)
```

At `t = 70` that returns `need_70` identically. The CTE-internal point columns
never have to cross the package boundary.

### Two traps this avoids, both verified

**The lookup's own bands are unusable here.** Its window partitions by scale id
alone, so a scale carrying two items at one cutoff makes `lead()` repeat the
value and collapses the band to `max = min - 0.1` — 119 of its 976 rows.
Restricting to whole letters removes every duplicate cutoff (135 rows, 135
distinct by both letter and cutoff), and the bands are recomputed locally.

**Scale ids collide across districts.** 976 rows but only 550 distinct
`gradescaleid|letter`, so a bare `gradescaleid` join fans out about 1.8x. The
join and the window both carry `_dbt_source_project`, which makes the key
exactly unique at 976.

### Verification

- **Rung coverage is total and unambiguous.** Every percent from 0 to 100 lands
  on exactly one rung for every scale — 0 violations across 3,232 scale-percent
  pairs. The KIPP NJ 2019 chain reads `F` to `D` at 63, `D` to `C` at 73, `C` to
  `B` at 83, `B` to `A` at 93, `A` to top.
- **No fan-out.** Dev and prod both return 854,245 rows and 854,209 distinct
  composite keys — exactly equal. The uniqueness test still warns at 12, the
  pre-existing `#3915` storedgrades count.
- **Contract passes** on a dev build, before and after formatting.
- **Trunk clean** on both files.

### What is not verified, and why

`need_*` is null across the entire extract today — it is a live-gradebook
calculation and AY2026 has no posted grades, so `need_next` is null everywhere
too. The arithmetic follows from the formula's linearity rather than from
observed values. **The produced numbers want a spot-check against real students
once Q1 grades land**; the rung logic and the join are verified now, the
multiplication is not.

### One documented sharp edge

Dropping the minus rungs widens each whole-letter band downward. On KIPP NJ 2019
the `F` rung therefore runs to 62.9, so a student at 61 — a passing `D-` — shows
a next letter of `D`. That is the correct next whole letter for them, but it is
not a pass-fail signal; `is_quarter_course_failing` is. Called out on the column
description.

## AI Assistance

Claude Code found the scale heterogeneity and the two join traps, derived the
linearity identity that avoids a package change, and wrote the SQL,
descriptions, and validation queries. The three product decisions were
human-directed: whole letters rather than plus-minus, a failing student's target
being `D`, and the Y1 framing.

### General

### dbt

- [x] Properties file updated with `data_type` and `description` for all three
      new columns
- [ ] Exposure — still absent on this model; pre-existing, see #4685
- [x] **Not a breaking change.** Additive only; the four existing `need_*`
      columns are untouched

## CI checks

- [x] **Trunk** — `trunk check --force` clean on both changed files before push
- [ ] **dbt Cloud**
      `pull_request` paths

### DIFF
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
@@ -503,8 +503,15 @@ models:
           negative when the grade is already secured regardless of remaining
           work, and can exceed 100 when the rung is out of reach in the time
           left — consumers should handle both rather than displaying a raw
-          percent. NULL wherever need_60 is null, meaning prior-year rows and
-          any course with no next rung.
+          percent.
+
+          NULL from two independent causes, and the second one surprises people.
+          Either need_60 is null, which is every prior-year row since need_* is
+          a live-gradebook calculation; OR there is no next rung, which is a
+          student already at the top whole letter and any course on a scale with
+          no letter ladder. The second case does NOT require need_60 to be null
+          — a student at 95 percent has a perfectly good need_60 and still gets
+          a null need_next, because there is nothing above an A to need.
       - name: category_name_code
         data_type: string
         description: Gradebook category code (e.g. F/S/W; NULL on Y1 rows).
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
@@ -462,6 +462,49 @@ models:
       - name: need_90
         data_type: float64
         description: Percent needed to reach a 90 (current-year rows only).
+      - name: need_next_letter_grade
+        data_type: string
+        description: >-
+          The next WHOLE letter grade above where the student's in-progress Y1
+          grade currently sits — one of A, B, C, D or F. Plus and minus rungs
+          are deliberately excluded, so a student sitting at C- and a student
+          sitting at C+ are both working toward the same B. NULL once the
+          student is at the top whole letter, and NULL on courses whose grade
+          scale has no letter ladder, such as the Camden Pass Fail Scale.
+
+          Read from the course's OWN grade scale, not a fixed ladder, so the
+          same letter means different cutoffs across courses — see
+          need_next_cutoff_percent.
+
+          Do NOT read the implied current rung as a pass or fail signal.
+          Dropping the minus rungs widens each whole-letter band downward, so on
+          KIPP NJ 2019 the F rung runs all the way to 62.9 and a student sitting
+          at 61 — a passing D- — shows a next letter of D. That is the correct
+          next WHOLE letter for them, but they are not failing. Use
+          is_quarter_course_failing for pass or fail.
+      - name: need_next_cutoff_percent
+        data_type: float64
+        description: >-
+          The percent cutoff for need_next_letter_grade on that course's grade
+          scale. It varies by scale and will NOT generally match the 60/70/80/90
+          targets behind need_60 through need_90 — on KIPP NJ 2019, which covers
+          most enrolments, a whole-letter D is 63 and a C is 73, while 60 and 70
+          are the D- and C- cutoffs. The two sets of columns are answering
+          different questions and are expected to disagree by a few points.
+      - name: need_next
+        data_type: float64
+        description: >-
+          Percent required in the CURRENT term for the student's YEAR-TO-DATE
+          course grade to reach need_next_cutoff_percent. It is not the percent
+          needed to earn that letter for the quarter alone.
+
+          Derived from need_60 and need_70, which are exactly linear in their
+          target, so this is an exact value rather than an interpolation. Can be
+          negative when the grade is already secured regardless of remaining
+          work, and can exceed 100 when the rung is out of reach in the time
+          left — consumers should handle both rather than displaying a raw
+          percent. NULL wherever need_60 is null, meaning prior-year rows and
+          any course with no next rung.
       - name: category_name_code
         data_type: string
         description: Gradebook category code (e.g. F/S/W; NULL on Y1 rows).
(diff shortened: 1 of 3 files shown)


anthonygwalters:
Verified each finding against the code and the warehouse before acting. Four implemented, one declined with evidence, two acknowledged as out of scope. Pushed in `56ddff880`.

## 2 — Right that it needs a guard. The mechanism is different from what you described.

Added the test you proposed, on `int_powerschool__gradescaleitem_lookup`, scoped to whole letters, at `severity: error`.

But the failure mode is not fan-out. Two whole letters sharing a cutoff make `lead()` return the earlier row's **own** cutoff, so its band collapses to `max = min - 0.1` and matches nothing, while the later row absorbs the whole range. Concretely, if a scale ever carried `C@73` and `C2@73`, the `C` row would go dead and every student from 73 to 82.9 would keep matching `C2` and getting the correct next rung — but if the duplicate sat at the **bottom** rung, students below the second cutoff would silently get the duplicate letter as their next rung instead of the true one. Silently wrong next letter, not multiplied rows.

Row multiplication needs two **fully identical** rows, same letter and same cutoff. Your proposed key catches that case too, so the test is right either way — the reason to want it is just different from the reason given.

Evidence the construction behaves that way rather than overlapping: the unscoped lookup already carries 119 rows at `max = min - 0.1`, all from duplicate cutoffs among plus-minus and special-code items. Degenerate, never overlapping.

Two things worth recording about the test itself:

- **It is not vacuous.** In scope: 0 duplicate groups. Unscoped: 10 groups across 129 rows. The `where` clause is load-bearing, and the unscoped version would fail today on items no consumer range-matches against.
- **`error`, not the intermediate-layer default of `warn`**, exactly for the reason you gave — both failure modes are silent, and this model's consumer has a `warn`-level key test sitting at a baseline of 12 for the unrelated `#3915` issue, so a regression there raises a number nobody reads. The model is also `materialized: table`, so the check actually re-runs on the data-change condition rather than going stale.

## 3 — Correct, and the wording was worse than you said

The description read "NULL wherever `need_60` is null, meaning prior-year rows and any course with no next rung," which folds two independent causes into one clause and implies the second is a subset of the first. It isn't. Reworded to name both, and to state the case explicitly: a student at 95 percent has a perfectly good `need_60` and still gets a null `need_next`, because there is nothing above an A to need.

## 1 — Correct, moved

Not an ST06 violation — sqlfluff passes on both placements, since arithmetic and boolean comparisons land in the same "complex" bucket and the rule does not order within it. But the file had already established a de facto order, with `cumulative_y1_gpa_unweighted_change_from_prior_year` and `gpa_band_change_from_prior_year` grouped ahead of the `coalesce`, the `if`, and the logicals, and `need_next` is the same kind of expression. Moved to sit with them.

## 6 — Good catch, and it would have been a silent null

Confirmed: the kipptaf union covers `kippnewark`, `kippcamden` and `kippmiami` only, and `TODO(#4340)` at line 307 is where Paterson gets turned on. Extended that comment so the future PR sees it — without a `kipppaterson` source on the lookup, the ladder join finds no scale and `need_next_*` stays null for every Paterson row with nothing failing.

## 4 — Declining. No precedent, and you found that yourself.

Confirmed your grep independently: **zero** named `WINDOW` clauses across every model in `src/dbt`. The upstream model this reads from repeats its own `over (partition by parent.id order by items.cutoffpercentage)` rather than naming it.

Four adjacent, identical `OVER` clauses are scannable, and the repetition makes each column's partitioning legible at the point of use. Introducing a BigQuery construct that appears nowhere else in sixteen dbt projects, in a leaf extract, to save three lines, is a worse trade than the duplication — and it would be the kind of novel idiom a future reader has to stop and check sqlfmt's handling of. If the team wants named windows it is a repo-wide convention call, not a drive-by in this PR.

## 5 — Agreed, and deliberate

The tradeoff is reasoned through in the PR body and you read it that way. Promoting the rung logic onto `int_powerschool__gradescaleitem_lookup` is the right move the moment a second consumer appears; today it would mean a `src/dbt/powerschool/` change and the two-PR cross-project dance for one caller. The new test lands on that model regardless, so half the coupling is already there if someone wants to finish the job later.

## 7 — Agreed, and out of scope by the same agreement as #4684

`contains_pii` on this model needs a scope decision, direct identifiers only or direct plus indirect, which root `CLAUDE.md` says to confirm with the team rather than assume. Unchanged from the same finding on #4684. The new columns add no PII surface — they are grade metrics in the same class as the existing `need_60` through `need_90`.

## Verification after the change

Rebuilt both models on a dev target against prod defer. The new test passes, the consuming model's contract passes, and its uniqueness test still warns at 12, unchanged from prod. `trunk check --force` clean on all three files. No logic change to `need_next` itself.

Issue #4382, opened 2026-07-12
feat(kipptaf): add prior-quarter and prior-year Y1 GPA columns to grades extract


anthonygwalters:
Tableau handles lag/lead-style calculations poorly, so `rpt_tableau__student_course_grades` should carry two precomputed comparison columns, populated only on current-year rows:

- `gpa_y1_prior_quarter` — weighted Y1 GPA as of the previous quarter (Q2 row shows the Q1 value; Q1 and Y1 rows NULL). Sourced from the live-gradebook branch of `int_powerschool__gpa_term`.
- `gpa_y1_prior_year` — the student's **final** weighted Y1 GPA from the prior academic year. PowerSchool only stores end-of-year Y1 grades, so a true "as of the same quarter last year" value does not exist; the final value was chosen deliberately over an approximate reconstruction. Blended across schools by credit hours when a student attended more than one school that year.

Prior-year rows keep both columns NULL, consistent with the extract's other as-of-today measures.

Follow-up to the gradebook rebuild (#4340, #4341, #4344).

PR #4688, opened 2026-08-02
feat(dbt): resolve category driver at Y1 and backfill prior-year running grades

### LINKED ISSUE #4687: feat(dbt): resolve category driver at Y1 and backfill prior-year running grades
Two changes to `rpt_tableau__student_course_grades`, shipping together but with **different lifespans**. One is permanent. One must be deleted later, and this issue is the authoritative instruction for deleting it.

Spec: `docs/superpowers/specs/2026-08-01-prior-year-backfill-and-category-drivers-design.md`
Plan: `docs/superpowers/plans/2026-08-01-prior-year-backfill-and-category-drivers.md`
Shipped in: #4688

---

# REMOVAL RUNBOOK

**Read this whole section before editing anything.** It is written for an agent or engineer who has no other context and has been pointed here to undo the temporary half.

## The selector

Every temporary object is named with a `backfill_` prefix. Nothing else in this model uses that token.

```bash
grep -n "backfill_\|TODO(#4687)" src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__student_course_grades.sql
```

Six hits at merge, one per temporary CTE, and zero false positives. Verified collision-free before the work shipped. Read the "Also restore" section below for the one column this grep deliberately does not reach.

## Trap 1 — do NOT grep for `prior_year_`

**This is the single most dangerous mistake available here.** Three CTEs in this model are named `prior_year_gpa_rollup`, `prior_year_gpa` and `prior_year_cumulative`. They came from #4684, they are permanent, and they are load-bearing for `is_gpa_band_slide`, `cumulative_y1_gpa_unweighted_prior_year` and `gpa_band_unweighted_prior_year`.

Deleting them silently breaks the GPA-slide monitoring columns. The temporary CTEs were deliberately renamed away from that prefix so a grep cannot conflate them. Use `backfill_`, never `prior_year_`.

## Trap 2 — Change A is permanent, do NOT remove it

The four category-driver columns and their two CTEs ship in the same PR and look like part of the same change. They are not.

Keep all of this:

- CTEs `category_ranked` and `category_drivers`
- The `category_drivers as cd` join
- Every column matching `lowest_category_*` and its properties-yml entry. At time of writing that is four: `lowest_category_y1_name`, `lowest_category_y1_percent`, `lowest_category_recent_term_name`, `lowest_category_recent_term_percent`. Trust the glob over this list if they ever disagree.

Removing them re-breaks `Category Driving Gap` at the Y1 marking period, which is the level the dashboard is primarily viewed at. None of them carry a `TODO(#4687)` marker, which is how you tell them apart.

## Trap 3 — restore, do not delete, inside the UNION

`quarter_grades` branch 3 is one arm of a three-branch `UNION ALL`, which binds **by position**. Deleting the two backfilled projections misaligns the union against branches 1 and 2.

Restore them to exactly this, in this order, in the same slot:

```sql
            cast(null as float64) as y1_course_in_progress_percent_grade_adjusted,
            cast(null as string) as y1_course_in_progress_letter_grade_adjusted,
```

Then delete the `backfill_running_course` join from that branch. If branch 3 was aliased to `sg` to support that join, un-aliasing is optional cleanup, not required.

## Trap 4 — leave `courses_gradescaleid` alone

Branch 3 contains:

```sql
            cast(null as int64) as courses_gradescaleid,
```

This predates the backfill. It came from #4686 and it is **load-bearing**: the `grade_scale_ladder` join keys off it, and its being null on prior-year rows is the only thing keeping `need_next_letter_grade` and `need_next_cutoff_percent` null there. Populating it, or deleting it while tidying up prior-year nulls, produces a nonsense pairing where a next letter grade appears with no `need_next` percent beside it.

## Trap 5 — restore the GPA expression verbatim

In `student_roster`, the backfill wrapped the original in a coalesce. Restore the original exactly:

```sql
            if(term.quarter = 'Y1', gty.gpa_y1, gtq.gpa_y1) as gpa_y1,
```

Then delete the `backfill_running_gpa` join. **Do not touch `gpa_for_quarter` or `gpa_n_failing_y1`** on the neighbouring lines — neither was ever modified by this work, and an over-broad revert of "the GPA backfill" is the likely way they get damaged.

## Also restore

The prose in three column descriptions in the properties yml, which currently describe the reconstruction: `y1_course_in_progress_percent_grade_adjusted`, `y1_course_in_progress_letter_grade_adjusted` and `gpa_y1`. Each says the values are reconstructed and temporary. That wording becomes false on removal.

**A fourth column changes behaviour and the grep will NOT find it.** `y1_course_letter_grade_adjusted` is a coalesce of the stored Y1 letter with `qg.y1_course_in_progress_letter_grade_adjusted`. Because the stored side only matches on Y1-quarter rows, this column was null on every prior-year Q1-Q4 row before the backfill — 0 non-null across 661,429 AY2025 non-Y1 rows. After it, all of them carry the reconstructed letter. It has no `backfill_` name and no `TODO(#4687)` marker, so the selector above does not surface it, and it reverts to null on its own once the backfill CTEs are deleted.

That reversion is correct and needs no code change. But its description still says the in-progress fallback applies "for the current year", which is false while the backfill is live, and its behaviour will visibly change at removal. Update that description at removal, and do not be surprised when a column nobody edited starts returning nulls again.

## Verify the removal

Build to dev and confirm all four:

1. **Current-year data unchanged.** Sum `gpa_y1`, `gpa_for_quarter` and `gpa_n_failing_y1` grouped by academic year and quarter. Every current-year figure must be byte-identical to prod. This is the invariant that matters most.
1. **Row count unchanged.** Total rows and distinct composite keys must match prod exactly. Removal is subtractive and must not move either. Take the prod read in the same session — prod drifts, and a stale remembered number will look like a regression that is not one.
1. **Category drivers still work.** `lowest_category_y1_name` must still be non-null on Y1 rows. If it went to zero, Trap 2 was hit.
1. **`need_next_*` still null on prior years.** If it became non-null, Trap 4 was hit.

The model's composite uniqueness test warns at 12 — the pre-existing #3915 storedgrades count. That number must be unchanged before and after.

---

# What was built, and why

## 1. Category driver at Y1 — permanent

`Category Driving Gap` read `not available` at every Y1 row, because the category join is `s.quarter = c.term` and there is no Y1 term in the category data. All 38,281 Y1 rows had a null `category_name_code`.

Four student-course-grain columns, present on every row including Y1: the lowest year-running category and the lowest recent-term category, each with its percent. Scalar columns rather than attaching category rows to Y1, which would have fanned Y1 out from 38K to roughly 150K rows on a contracted model with a live uniqueness test.

The recent-term driver reads from the most recent term holding any usable category percent, quarter-level or year-running. Its name is null whenever that term has no quarter-level percent, so a category name never appears without a value behind it. A null there does **not** mean the section has no quarter category data anywhere — measured on Newark and Camden AY2025, 1,547 of the 1,549 nulls have an earlier term that does carry one.

**Correction on the naming rationale.** The rename from `latest_quarter` to `recent_term` was originally justified by a claim that a Q4 row could carry a Q2-sourced driver. That was wrong: the code does not fall back, it blanks, and genuine earlier-term sourcing occurs 3 times network-wide. The name is still the better of the two, since the driver's term is whichever most recently held data rather than Q4 by definition — but do not repeat the fallback story.

Deletes two Tableau calcs: `Lowest Category % (course)` and the `MIN(IF ...)` form of `Category Driving Gap`.

## 2. Prior-year running backfill — temporary

`y1_course_in_progress_*` was 100 percent null on prior years, and `gpa_y1` was the final year value repeated on every quarter row (see #4382). Pointing M1 and M2 at the running columns would have blanked the drill-down on the only year that currently holds data, which blocks summer stakeholder training.

- **Course grades** — running average of stored quarter percents. Q1 exact, Q4 anchored to the stored Y1 and exact, Q2 and Q3 approximate. Letter derived by banding the reconstructed percent.
- **GPA** — running from `weighted_gpa_points_term` over `total_credit_hours_term`, deliberately **not** anchored. Anchoring would insert a step at Q4 that the data does not contain. The real AY2025 trajectory rises — 45.1, 46.7, 46.6, 48.2 percent of high school students at or above 3.0, against a stored Y1 of 48.9.

The asymmetry between those two is intentional. Do not "fix" it.

Accuracy was explicitly not the goal; populated and labelled was. The course reconstruction lands within half a point of the stored Y1 on 97.0 percent of courses. The GPA reconstruction tracks the distribution closely but moves 222 to 748 individual students across the 3.0 line.

**On the numbers above.** An earlier version of this issue cited 98.2 percent and 173 to 645, and described the GPA reconstruction as showing a year-long decline followed by a fake Q4 recovery. Those were AY2024 figures, produced by querying `int_powerschool__gpa_term` at `yearid = 34` when the convention is `yearid = academic_year - 1990`. The corrected AY2025 data rises rather than declining. The decision not to anchor was re-confirmed against the corrected numbers and stands, but its original justification was wrong and is not preserved.

## Isolation

Course grades are isolated **structurally** — `quarter_grades` branch 3 filters to the prior year while branches 1 and 2 filter to the current year, so they are disjoint by `WHERE`.

GPA is **not**. It is assembled in `student_roster` from a `gtq` join carrying no year predicate, so its isolation comes only from an `ON`-clause gate on `academic_year`. That gate is the reason current-year rows receive null and the coalesce falls through untouched. It was verified at row level, not asserted: a complete anti-join over all 141,036 AY2026 rows returned zero mismatches on `gpa_y1`, `gpa_for_quarter` and `gpa_n_failing_y1`.

## Out of scope

- Moving the category code decode into the model
- Backfilling `gpa_n_failing_y1` — nothing that reads it is marking-period filtered
- The as-of versus period-only labelling work, tracked in Asana

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will make `Category Driving Gap` resolve at the
Y1 marking period, and temporarily populate prior-year running course grades and
GPA so the drill-down is not blank during summer stakeholder training.

Two changes to `rpt_tableau__student_course_grades`, shipping together with
**deliberately different lifespans**. Issue #4687 is the authoritative removal
runbook for the temporary half.

### 1. Category drivers at Y1 — permanent

`Category Driving Gap` read `not available` on every Y1 row, which is the level
the dashboard is primarily viewed at. Cause: the category join is
`s.quarter = c.term` and there is no Y1 term, so all 38,281 Y1 rows carried a
null `category_name_code`.

Four student-course-grain columns, present on every row including Y1:

| Column | Meaning |
| --- | --- |
| `lowest_category_y1_name` / `_percent` | lowest year-running category |
| `lowest_category_recent_term_name` / `_percent` | lowest category in the most recent term holding data |

Scalar columns rather than attaching category rows to Y1, which would have
fanned Y1 out from 38K to roughly 150K rows on a contracted model with a live
uniqueness test.

Deletes two Tableau calcs: `Lowest Category % (course)` and the `MIN(IF ...)`
form of `Category Driving Gap`.

### 2. Prior-year running backfill — temporary

`y1_course_in_progress_*` was 100 percent null on prior years, and `gpa_y1` was
the final year value repeated on every quarter row (#4382). Pointing M1 and M2
at the running columns would have blanked the drill-down on the only year that
holds data.

- **Course grades** — running average of stored quarter percents. Q1 exact, Q4
  anchored to the stored Y1 and exact, Q2 and Q3 approximate.
- **GPA** — running from the per-term components, deliberately **not** anchored.

That asymmetry is intentional and documented. Accuracy was explicitly not the
goal; populated and labelled was.

**Six CTEs are temporary**, each carrying a `backfill_` prefix and a
`TODO(#4687)` marker. `grep -n "backfill_\|TODO(#4687)"` finds all of them and
nothing else — verified collision-free.

## Three defects found during implementation

Recorded because each was caught by a guardrail rather than by luck, and two
were mine.

**The plan's `backfill_y1_stored` join omitted `storecode` from its key.** The
first build produced WARN=130 and 166 extra rows — a real fan-out, caused by
roughly 96 conflicting duplicate Y1 rows from the known #3915 double-write
leaking onto previously-clean Q1-Q3 rows. Caught because the plan made
`warn = 12` and the exact row count hard invariants rather than "the build
passes". Fixed with `dbt_utils.deduplicate` on the join input; the tracked #3915
warn signal is untouched at 12, not masked.

**The plan's yearid expression was off by one.** `current_academic_year - 1 - 1991`
yields 34, but the convention is `yearid = academic_year - 1990`, so AY2025 is
35. Left in, the join would never have matched — and the first build proved it,
going flat across all five marking periods, which is the exact failure signature
the plan names.

**The same off-by-one contaminated the design analysis.** Every GPA figure in
the original spec was AY2024. Corrected in `e1862585f`: course reconstruction
accuracy 98.2 to 97.0 percent, students crossing the 3.0 line 173-645 to
222-748, and the trajectory from `50.1/48.4/46.1/47.7` against `50.4` to
`45.1/46.7/46.6/48.2` against `48.9`. The rejected-anchoring argument had
claimed a year-long decline; the real data **rises**. The decision not to anchor
was re-confirmed against corrected numbers and stands, but the spec now records
that its original justification was wrong rather than quietly restating it.

## Verification

Current-year isolation was the first invariant, because the GPA half is not
structurally isolated the way the course half is — it depends on a single
`ON`-clause year gate.

- **Isolation proven at row level**, not aggregate: a complete anti-join over
  all 141,036 AY2026 rows returned **0 mismatches** on `gpa_y1`,
  `gpa_for_quarter` and `gpa_n_failing_y1`. Review independently confirmed from
  the compiled SQL that the gate emits `enr.academic_year = 2025` as a hardcoded
  literal, so a current-year row cannot match.
- **No fan-out.** Dev and a same-session prod read are identical in both years:
  854,263 / 854,227 overall, AY2025 713,227 / 713,191, AY2026 141,036 / 141,036.
  Note prod drifted 18 rows mid-execution; comparisons are same-session for that
  reason.
- **Hard invariants all zero** — Q1 exactness, Q4 anchor, letter coverage.
- **`need_next_*` still null across all 713,227 AY2025 rows.** That property
  depends on `courses_gradescaleid` staying null in branch 3; reusing the
  existing grade-scale ladder would have lit up a next-letter with no percent
  beside it.
- Composite uniqueness test still warns at **12**, the pre-existing #3915 count.
  `int_powerschool__gradescaleitem_lookup`'s uniqueness test still passes.

## AI Assistance

Claude Code designed and implemented this under human direction, executed
task-by-task with a fresh subagent and an independent review per task. Four
decisions were human-made: whole letters rather than plus-minus, nulling a
driver name when its percent is null, renaming `latest_quarter` to `recent_term`,
and leaving the GPA reconstruction unanchored — the last re-confirmed after the
year error was found.

### General

### dbt

- [x] Properties file updated with `data_type` and `description` for all new and
      changed columns
- [ ] Exposure — still absent on this model; pre-existing, unchanged by this PR
- [x] **Not a breaking change.** Additive columns plus a `coalesce` that falls
      through to the original expression outside the prior year

## CI checks

- [x] **Trunk** — clean on the full branch diff before push
- [ ] **dbt Cloud**
      `pull_request` paths

Refs #4687

### DIFF
--- a/docs/superpowers/plans/2026-08-01-prior-year-backfill-and-category-drivers.md
+++ b/docs/superpowers/plans/2026-08-01-prior-year-backfill-and-category-drivers.md
@@ -613,8 +613,8 @@ description: >-
   progress this is the real running value from the gradebook. For the prior year
   it is RECONSTRUCTED from stored quarter grades — Q1 is exact, Q4 is anchored
   to the stored Y1 percent and is exact, Q2 and Q3 are approximations that agree
-  with the year value to within half a point on roughly 98 percent of courses.
-  The reconstruction is temporary; see TODO(#4687) in the model.
+  with the year value to within half a point on 97.0 percent of courses. The
+  reconstruction is temporary; see TODO(#4687) in the model.
 ```
 
 For `y1_course_in_progress_letter_grade_adjusted`:
--- a/docs/superpowers/plans/2026-08-01-prior-year-backfill-and-category-drivers.md
+++ b/docs/superpowers/plans/2026-08-01-prior-year-backfill-and-category-drivers.md
@@ -1,0 +1,978 @@
+# Category drivers at Y1 and prior-year running backfill — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use
+> `superpowers:subagent-driven-development` (recommended) or
+> `superpowers:executing-plans` to implement this plan task-by-task. Steps use
+> checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make `Category Driving Gap` resolve at the Y1 marking period, and
+temporarily populate prior-year running course grades and GPA so the drill-down
+is not blank during summer stakeholder training.
+
+**Architecture:** Two independent changes to one model,
+`src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__student_course_grades.sql`.
+Change A adds four student-course-grain category-driver columns via two new CTEs
+and one ungated join. Change B adds prior-year reconstructions via new CTEs
+consumed by the existing prior-year union branch (course grades) and by a
+year-gated join (GPA).
+
+**Tech Stack:** dbt on BigQuery, kipptaf project. No Python, no unit tests —
+this repo validates dbt models with `dbt build` plus BigQuery MCP queries.
+
+## Global Constraints
+
+- Spec:
+  `docs/superpowers/specs/2026-08-01-prior-year-backfill-and-category-drivers-design.md`.
+  Read it before Task 1.
+- Worktree:
+  `/workspaces/teamster/.worktrees/anthonygwalters/feat/claude-backfill-and-category-drivers`.
+  Branch `anthonygwalters/feat/claude-backfill-and-category-drivers`. Use
+  `git -C <worktree>` for every git call.
+- Every dbt call:
+  `uv run dbt <cmd> --project-dir <worktree>/src/dbt/kipptaf --target dev --defer --favor-state --state /workspaces/teamster/src/dbt/kipptaf/target/prod`.
+  Never bare `dbt`.
+- Lint before every push:
+  `/workspaces/teamster/.trunk/tools/trunk check --force --no-fix <files> </dev/null`,
+  run with cwd set to the worktree.
+- **Current-year isolation is the first invariant.** Any non-zero difference in
+  a current-year value of `gpa_y1`, `gpa_for_quarter`, `gpa_n_failing_y1` or
+  `y1_course_in_progress_*` between the dev build and prod means the change does
+  not ship.
+- **Do NOT populate `courses_gradescaleid` in `quarter_grades` branch 3.** The
+  `grade_scale_ladder` join keys off it, and it being null on prior-year rows is
+  the only thing keeping `need_next_*` null there. Populating it would light up
+  `need_next_letter_grade` while `need_next` stayed null.
+- SQL conventions in `src/dbt/CLAUDE.md` are binding: no `QUALIFY`, no
+  `ORDER BY`, no subqueries against CTEs, max one level of function nesting,
+  ST06 column ordering, ST09 join-predicate ordering, trailing commas.
+- Contract is enforced. Every new column needs a `data_type` and a `description`
+  in
+  `src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml`.
+- The model's composite uniqueness test warns at **12** — the pre-existing
+  `#3915` count. It must still read 12 after every task.
+- Prod baseline as of 2026-08-01: **854,245 rows, 854,209 distinct** composite
+  keys.
+
+## File Structure
+
+| File                                                    | Responsibility                          | Change                             |
+| ------------------------------------------------------- | --------------------------------------- | ---------------------------------- |
+| `.../rpt_tableau__student_course_grades.sql`            | all CTEs, joins, projections            | modified in Tasks 2, 3, 4          |
+| `.../properties/rpt_tableau__student_course_grades.yml` | contract + descriptions                 | modified in Tasks 2, 3, 4          |
+| `.claude/scratch/4687-baseline.md`                      | pre-change numbers for isolation checks | created in Task 1, never committed |
+
+No new model files. Both changes belong in this model because both are
+extract-shaped presentation concerns, and the spec establishes that neither
+needs a `src/dbt/powerschool/` change.
+
+---
+
+### Task 1: Capture the pre-change baseline
+
+Every later isolation check compares against these numbers. Capture them before
+touching anything, because after the first edit the dev relation is no longer a
+clean control.
+
+**Files:**
+
+- Create: `.claude/scratch/4687-baseline.md` (gitignored, do not commit)
+
+**Interfaces:**
+
+- Consumes: nothing
+- Produces: the baseline numbers referenced by Tasks 2, 3, 4 and 5
+
+- [ ] **Step 1: Record the current-year GPA control set**
+
+Run via BigQuery MCP:
+
+```sql
+select
+  academic_year,
+  quarter,
+  count(*) as rows_,
+  round(sum(gpa_y1), 4) as sum_gpa_y1,
+  round(sum(gpa_for_quarter), 4) as sum_gpa_for_quarter,
+  sum(gpa_n_failing_y1) as sum_n_failing,
+  countif(y1_course_in_progress_percent_grade_adjusted is not null) as has_running_pct
+from `kipptaf_tableau.rpt_tableau__student_course_grades`
+group by 1, 2
+order by 1, 2
+```
+
+Paste the full result into the scratch file under a heading
+`## Current-year GPA control set`.
+
+- [ ] **Step 2: Record row-count and grain baseline**
+
+```sql
+select
+  count(*) as rows_,
+  count(distinct format('%T|%T|%T|%T|%T|%T',
+    _dbt_source_relation, studentid, academic_year, quarter,
+    course_number, category_name_code)) as distinct_key
+from `kipptaf_tableau.rpt_tableau__student_course_grades`
+```
+
+Expect `854245` / `854209`. If it differs, prod has moved — record the new
+numbers and use those as the control from here on.
+
+- [ ] **Step 3: Record the Category Driving Gap ground truth**
+
+This is the regression oracle for Task 2. It captures what the current Tableau
+logic produces at Q4, so the new column can be checked against known-good
+behaviour.
+
+```sql
+with
+ranked as (
+  select
+    studentid,
+    sectionid,
+    category_name_code,
+    category_quarter_percent_grade,
+    row_number() over (
+      partition by studentid, sectionid
+      order by
+        (category_quarter_percent_grade is null) asc,
+        category_quarter_percent_grade asc,
+        category_name_code asc
+    ) as rn
+  from `kipptaf_tableau.rpt_tableau__student_course_grades`
+  where
+    academic_year = 2025
+    and quarter = 'Q4'
+    and category_quarter_percent_grade is not null
+)
+select
+  category_name_code,
+  count(*) as student_sections
+from ranked
+where rn = 1
+group by 1
+order by 1
+```
+
+Record the distribution. Task 2 must reproduce it.
+
+- [ ] **Step 4: Verify the letter-coverage gate**
+
+**This gate can fail the plan.** The spec flags full-scale band coverage as
+unverified. If it fails, stop and report before writing any code — Task 3's
+letter derivation needs redesigning.
+
+```sql
+with
+probe as (select p from unnest(generate_array(0, 100)) as p),
+scale as (
+  select _dbt_source_project, gradescale_name, min_cutoffpercentage, max_cutoffpercentage
+  from `kipptaf_powerschool.int_powerschool__gradescaleitem_lookup`
+)
+select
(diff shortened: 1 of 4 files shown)


anthonygwalters:
Verified each finding against the code and the warehouse. Two were already fixed before this posted, two are dismissed with evidence.

## 1 and 2 — both correct, both already fixed in `d0d270f62`

This review ran against an earlier commit. A whole-branch review found the same two defects and they were fixed in `d0d270f62`, along with two minors:

- `98.2 percent` and `roughly 98 percent` are now `97.0 percent` in both shipped files.
- The `lowest_category_recent_term_name` description no longer claims a fallback.

**Your diagnosis of finding 2 is better than the one we acted on**, and worth recording. We knew the description was wrong and had measured it — 1,549 null names across Newark and Camden AY2025, of which 1,547 have an earlier term that does carry a quarter percent, against 3 genuine earlier-term fallbacks network-wide. What we had not pinned down was the mechanism. You did: `category_y1_percent_grade_running` is a running average that stays non-null once any earlier term had data, because `AVG` skips nulls in its window, so a term whose own `percent_grade` is null everywhere still wins `rn_latest_term`. That is exactly it.

**On your suggested alternative.** You offer two fixes — decouple the ranks so `rn_lowest_quarter` uses its own "latest term with quarter data" rank, or correct the docstring. We took the docstring, because nulling the name was an explicit product decision made earlier in this PR's history and decoupling would reverse it. But the decoupling is the better product outcome: it would give those ~1,547 sections a real driver instead of a blank. That is a behaviour change and a decision for the dashboard owner rather than a drive-by, so it is going on the follow-up list rather than into this PR.

## 3 — Dismissing. The linter says otherwise, and adding it would be its own violation.

`trunk check --force --no-fix` on the model returns **No issues**. That is three independent runs: the implementer, a whole-branch reviewer, and again just now.

More decisive than the absence of a fire: the implementer **tested adding** the ignore comment during Task 3, and trunk flagged it as not suppressing anything. Adding it would trade a rule that does not fire for a `trunk/ignore-does-nothing` violation that does.

The precedent you cite is real — `int_finalsite__contact_address_of_record.sql:2` carries that exact comment and lints clean. I cannot fully explain why one needs it and the other does not; it may be a templater difference between the two dbt projects. But the empirical answer for this file is settled, and it is the linter's to give, not ours to infer from a sibling.

## 4 — Dismissing. Zero incidence.

Good catch on the mechanism: `if(r.storecode = 'Q4', y1.y1_stored_percent, r.running_percent)` does take the LEFT JOIN's value unconditionally, so a missing stored Y1 row would null the Q4 running value rather than falling back.

It never happens. Across all of AY2025:

| quarter | rows with a graded quarter percent | of those, missing a running value |
| --- | ---: | ---: |
| Q1 | 121,499 | **0** |
| Q2 | 123,754 | **0** |
| Q3 | 125,774 | **0** |
| Q4 | 122,457 | **0** |
| Y1 | 32,839 | **0** |

Every graded Q4 row has its stored Y1 counterpart. The dedup you wondered about as a possible cause picks a survivor rather than dropping a key, so it cannot orphan one either. Worth leaving as a latent note on temporary code rather than guarding.

## On the `/simplify` lens and the "verified as sound" section

Both were useful, and the second one saved time — independently confirming the `yearid` fix against `prior_year_gpa_rollup`'s identical expression, and confirming ST09 ordering on every new join, matched what we had found separately.

One correction to it. You note that "`backfill_quarter_running`'s own Q1-Q4 source is not separately deduplicated, consistent with the PR's finding that the double-write is Y1-row-specific." The double-write is not Y1-specific — Q4 carries 14 duplicate groups in the raw source. They are all `kipppaterson`, which `student_roster`'s `region in ('Newark', 'Camden')` filter excludes before the join sees them, so in-scope exposure is zero. The conclusion holds; the reason is structural exclusion rather than absence.

anthonygwalters:
@claude re-review

PR #4691, opened 2026-08-03
feat(dbt): rework the FRESH QC worklist definitions, field names and reference doc

GabyRangelB:
# Pull Request

## Summary & Motivation

...rework the FRESH QC worklist so it can be defended in front of the SRE team:
rename field names that actively mislead, implement the three definition changes
that came out of the AY2026 definitions review, and correct the published FRESH
reference doc, which had drifted from the shipped models — including one place
where it contradicted itself.

Context: we are reviewing the new `rpt_tableau__fresh_dashboard_qc` definitions
with SRE. That review surfaced the misleading names first, then produced real
definition changes, so this PR grew past its original rename-and-doc-fix scope.

### The renames

| before                         | after                              | why                                                                                                                |
| ------------------------------ | ---------------------------------- | ------------------------------------------------------------------------------------------------------------------ |
| `ps_enroll_status`             | `finalsite_expected_enroll_status` | Derived entirely from Finalsite, reads nothing from PowerSchool. The yml had to apologize for the prefix in prose.  |
| `ps_grade_level`               | `sis_grade_level`                  | These are SIS values, but Miami's come from Focus, not PowerSchool.                                                 |
| `ps_schoolid`                  | `sis_schoolid`                     | Same.                                                                                                              |
| `ps_school`                    | `sis_school`                       | Same.                                                                                                              |
| `is_active_inactive_mismatch`  | `is_enroll_status_mismatch`        | "Active/inactive" is SIS jargon; the flag is about enrollment status disagreeing.                                   |
| `is_same_day_status_duplicate` | _deleted_                          | Renamed to `is_same_day_status_tie` first, then dropped outright by the definition changes below. Net: it is gone.  |

### The definition changes

Three changes from the review, all in
`int_tableau__finalsite_student_scaffold`:

- **A third direction on `is_enroll_status_mismatch`** — six pending Finalsite
  statuses (`Accepted`, `Assigned School`, `Did Not Enroll`,
  `Campus Transfer Requested`, `Parent Declined`, `Enrollment In Progress`) now
  carry an expectation, so a student mid-enrollment whom the SIS already has
  active gets flagged. `finalsite_expected_enroll_status` was renumbered to
  `0` / `2` / null in the process: the values deliberately mirror the SIS's own
  `enroll_status` domain so a comparison between the two columns means the same
  thing on both sides. An earlier revision used `1` for withdrawals, which
  collided with the SIS's `2`.
- **`is_same_day_status_tie` is deleted**, along with its two CTEs and
  `latest_status_date`. The pending-status comparison replaces it. The underlying
  same-day-tie behavior is unchanged and still documented — it just no longer
  gets its own worklist row, and the Reset Protocol is still the fix.
- **`is_enrolled_fdos` is now computed here**, against a first-day-of-school date
  hardcoded per region (Newark and Paterson Aug 28, Camden Aug 24, Miami Aug 14 —
  month and day hardcoded, year from `var("finalsite_recruitment_year")`). It no
  longer passes through either SIS's own flag: Focus computes one network-wide
  first day, and PowerSchool's is per-school; neither matches the regional date
  the enrollment team reports against. Adds `sis_entry_date` and
  `custom_fdos_date` to the contract so a reader can see what a row was judged
  against.

`is_enrolled_fdos` is a **bare comparison, not `if(cmp, true, false)`** — the
wrapped form would report every student with no SIS record as `false` instead of
null. Its sibling comparison flags keep the wrapper deliberately; both behaviors
are now documented.

### Two deliberate scope limits

- **`enroll_status` is renamed only at the QC view**, as `sis_enroll_status`. The
  shared `int_tableau__finalsite_student_scaffold` keeps the bare name because
  Progress to Goals and Aggregated expose it that way across 11 sites, and
  renaming it there would break their Tableau field references too. Both
  properties ymls now record why the layers differ, so nobody "fixes" the
  inconsistency later.
- **`int_focus__student_enrollments.ps_schoolid` is untouched.** There `ps_`
  means PowerSchool-space school id — a different and correct meaning. An
  initial sweep did rename the scaffold's reference to it; `dbt build` caught it
  and it was reverted. `int_extracts__student_enrollments` and
  `int_focus__student_enrollments` also keep their own `is_enrolled_fdos` for
  their other consumers; neither model is modified by this PR.

### Documentation

- Corrected a self-contradiction: the doc claimed Miami's point-in-time
  enrollment flags are always NULL while its own Resolved section said that gap
  was closed. The code agrees with the Resolved section — `enrollment_lookup`
  unions `int_focus__student_enrollments` for Miami and supplies all five
  columns, bridged via `int_finalsite__contact_id_attributes`.
- Documented `rpt_tableau__fresh_dashboard_qc`, which had zero coverage despite
  being contract-enforced and in the `fresh_dashboard` exposure. Includes a
  plain-language section on all four flags for non-technical readers, the ten
  Finalsite statuses that drive them, and the questions still open with SRE.
- **`is_missing_sis_record` is derived in the QC model rather than upstream**, and
  the two comparison flags read `false` — not NULL — when no SIS record matched,
  because they are `if()`-wrapped. Either way a student with no SIS record
  surfaces once rather than three times, since the view keeps only rows where a
  flag fired. `flag is null` is therefore not a usable missing-SIS proxy; an
  earlier revision of this doc got that backwards.
- Documented the hardcoded FDOS date, including that the dates came from SRE
  directly and that only the year rolls itself forward — so re-confirming them is
  now a required step of the recruitment-year rollover, not an optional one.
- Presented `is_enroll_status_mismatch` as **two** directions rather than three.
  The "left" and "not finished enrolling" status sets make one comparison
  (expected `2` against SIS `0`), so documenting them separately overstated what
  the check does. They are now one direction with two named halves; the follow-up
  still differs between the halves, which `latest_status` is what distinguishes.
- Rebuilt the lineage diagram, which omitted four edges — most materially
  `int_google_sheets__finalsite__goals_pivot`, without which goals appear to
  reach only the Aggregated view.
- Added a `CLAUDE.md` line distinguishing the published MkDocs pages from the
  working specs and plans under `docs/superpowers/`. That ambiguity is what led
  to auditing a stale plan while giving the published page a pass.

### Verification

`uv run dbt build --select int_tableau__finalsite_student_scaffold rpt_tableau__fresh_dashboard_qc --target dev --defer --favor-state --state target/prod`
— both models build, both contracts enforce, both uniqueness tests pass, 52.9k
rows on the int model (unchanged after deleting the tie-flag join).
`trunk check --force` clean on every changed file.

Worklist volumes in that dev build: 360 rows — 192 grade, 97 missing SIS, 38
enroll status, 33 school. Expected-status distribution: `0` for 11.3k students,
`2` for 1.9k, null for 4.4k.

**On the FDOS change specifically, a dev-vs-prod comparison moves zero students
today, and that is the expected result.** At rollover both SISs stamp every
enrolled student with one bulk entry date (Jul 1 in NJ, Aug 12 in Miami), all of
it before any first day, so the flag reads `true` for every student with a record
and null for every student without one in both versions. The change only begins
to discriminate once real per-student entry dates land after school starts.

The evidence therefore comes from settled prior-year data, where it is decisive:
applied to AY2025, the regional date moves **990 of 1,514 Miami students** from
`false` to `true` (Focus's single network-wide cutoff fell on Aug 11, before
Miami's own first day). The cost side is real and accepted — **83 NJ students**
move the other way, having enrolled after the regional date at a school that
started later, because one date per region is coarser than PowerSchool's
per-school date. Both figures apply this cycle's dates to a prior year, so they
measure how far a single regional date can diverge from per-school reality, not
what prod would have shown.

## AI Assistance

Claude Code authored the rename sweep, the definition changes, the doc
corrections, the dev-vs-prod analysis and this PR body. Human-directed: the
decision to rename, the specific naming scheme (reviewed and approved before
implementation), the three definition changes and their semantics, the scope
limits, the decision to merge two documented directions into one, and the choice
to leave the spec/plan documents alone as historical records. The four FDOS dates
came from SRE.

### General

- [x] Review the **Claude Code Review** comment posted on this PR — two passes,
      findings from both verified against the code and addressed.

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — both
      affected models already had one; descriptions updated for every renamed,
      added and deleted column.
- [x] Include (or update) an
      — no change needed; the `fresh_dashboard` exposure references model names,
      not column names, and no model was renamed.
- [x] **Breaking change?** **Yes.** On the contracted
      `rpt_tableau__fresh_dashboard_qc`, the Tableau workbook's references to
      `ps_grade_level`, `ps_schoolid`, `ps_school`, `ps_enroll_status`,
      `enroll_status` and `is_active_inactive_mismatch` break, and
      `is_same_day_status_duplicate` disappears entirely rather than being
      renamed. `sis_entry_date` and `custom_fdos_date` are added upstream. The
      workbook owner remaps those fields after deploy. Chosen now deliberately:
      the view is new and its definitions are still under review with SRE, so
      this is the cheapest moment to do it.

      **Rollback plan:** revert the commits. The columns return to their previous
      names on the next materialization — these are view and table column names
      with no stored state to migrate, so there is no data backfill or cleanup
      involved.

- [ ] External sources — not applicable, no external source added or modified.

### Docs

Not applicable — no schedule, sensor, or integration changes.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

---

  - https://app.asana.com/0/0/1217117628173683


GabyRangelB:
Both findings were valid and are fixed in 7dc408b4c. @claude for visibility.

## Important — comparison flags read `false`, not NULL

Correct, and the error was mine. I verified the semantics against BigQuery
rather than reasoning about them:

| expression                               | result    |
| ---------------------------------------- | --------- |
| `if(a != b, true, false)` with `b` NULL  | **false** |
| bare `a != b` with `b` NULL              | NULL      |
| `if(null, true, false)`                  | **false** |

So `if()` takes its else branch on a NULL condition exactly as on FALSE, and the
columns materialize as `false`. The behavior the doc described still holds —
`where flag_value` drops `false` and NULL alike, so a student with no SIS record
still surfaces once under `is_missing_sis_record` rather than three times — but
the stated mechanism was wrong.

Corrected at all three sites (the reference doc's implementation notes and both
column descriptions), and the doc now carries an explicit warning against the
failure mode you named: do not treat `is_grade_level_mismatch is null` as a
missing-SIS proxy, because it never fires.

**On the fix option: I kept the SQL as-is and did not drop the `if()` wrapper.**
Reasoning, for the record rather than as a rebuttal:

- `if(cond, true, false)` is the established idiom in this model — the sibling
  `is_enroll_status_mismatch` uses the same shape.
- Switching to a bare comparison would change the values a contract-enforced
  column emits. No downstream behavior would change today, since the only
  consumer filters on `where flag_value`, but it is a semantics change and
  belongs in a deliberate decision rather than as a side effect of a doc fix.
- Whether these flags *should* carry NULL, to distinguish "could not compare"
  from "compared, no mismatch", is now an open question logged in the reference
  doc alongside a related one about splitting "absent from the SIS" from
  "present but unlinked". If we adopt NULL propagation, that is the change to
  make it under.

## Minor — CTE name reintroduced the word the rename removed

Genuine loose end. `same_day_duplicate_flags` is now
`same_day_status_tie_flags`, with both SQL references and the yml cross-reference
updated.

## Recommendation — dbt Mesh versioning

No action taken. Noting that you verified the exception reasoning against
`git log` independently and reached the same conclusion, so there is nothing
further to add.

## Verification

`dbt build --select int_tableau__finalsite_student_scaffold rpt_tableau__fresh_dashboard_qc --target dev --defer --favor-state`
— both models build, both contracts enforce, both uniqueness tests pass.
`trunk check --force` clean on the changed files.

One note on the run itself: the BigQuery MCP tools were not registered in the
session where this PR was authored, so the flag-distribution checks that would
normally accompany a change like this were done through `dbt show` and `dbt
build` instead of ad-hoc warehouse queries.

GabyRangelB:
@claude - can you do another review pass, plz

GabyRangelB:
@claude - ok, check again, please?

GabyRangelB:
@claude one more time review plz

GabyRangelB:
@claude - please review

cbini (approved):

PR #4695, opened 2026-08-03
fix(renlearn): keep fast_star sch as string

### LINKED ISSUE #4694: fix: stg_renlearn__fast_star casts sch to int64 but 2332A is a valid school id
## Problem

`kippmiami` renlearn dbt runs have been failing since Fri Jul 31 09:08 ET:

```text
Database Error in model stg_renlearn__fast_star (models/staging/stg_renlearn__fast_star.sql)
Bad int64 value: 2332A
```

Failing runs: `f8071bc5-0d27-460e-b07b-48447aa5c12a` (auto-retry `b7f94a7a-02e8-4596-8b7b-96e5ed47e127`, also failed). The asset is currently DEGRADED.

## Root cause

`stg_renlearn__fast_star.sql:111` applies `cast(sch as int) as sch`. The `sch` column is `STRING` in the source and Renaissance delivers an alphanumeric school id for KIPP MIAMI-LIBERTY CITY:

```sql
select sch, schname, count(*)
from `teamster-332318.kippmiami_renlearn.src_renlearn__fast_star`
where regexp_contains(sch, r'[^0-9]')
group by 1, 2
-- sch=2332A, schname=KIPP MIAMI-LIBERTY CITY, n=2663 of 11451 rows
```

`2332A` is a legitimate school id, not bad data. The hard `cast` (not `safe_cast`) fails the whole model.

## Fix

Keep `sch` as a string:

- `src/dbt/renlearn/models/staging/stg_renlearn__fast_star.sql` — drop the `cast(sch as int)` and select `sch` as-is
- `src/dbt/renlearn/models/staging/properties/stg_renlearn__fast_star.yml` — `data_type: int64` to `string`

## Blast radius

No downstream model reads `sch` from `stg_renlearn__fast_star`. The kipptaf copy is a `dbt_utils.union_relations` passthrough over the kippmiami staging model and inherits the type; it has no SQL consumers. `kippnewark` also includes the local `renlearn` package, so the column type changes there too, with no consumers either.

## Related

Surfaced in the day-2 operations check for Fri Jul 31 to Mon Aug 3.

cbini:
# Pull Request

## Summary & Motivation

...unblock the kippmiami renlearn dbt build by carrying `stg_renlearn__fast_star.sch` through as a string instead of casting it to `int64`.

`stg_renlearn__fast_star.sql` applied `cast(sch as int) as sch`. Renaissance delivers an alphanumeric school id for KIPP MIAMI-LIBERTY CITY — `2332A` — so the hard cast (not `safe_cast`) failed the entire model:

```text
Database Error in model stg_renlearn__fast_star (models/staging/stg_renlearn__fast_star.sql)
Bad int64 value: 2332A
```

`2332A` is a legitimate school id, not bad data. The asset has been DEGRADED and failing since Fri Jul 31 09:08 ET (runs `f8071bc5-0d27-460e-b07b-48447aa5c12a` and its auto-retry `b7f94a7a-02e8-4596-8b7b-96e5ed47e127`).

Changes:

- `stg_renlearn__fast_star.sql` — drop the `cast`, select `sch` as-is in the plain-select block
- `properties/stg_renlearn__fast_star.yml` — contract `data_type` from `int64` to `string`

**Rollback plan:** revert the commit. The column reverts to `int64` and the model resumes failing on the same rows, so there is no state to unwind.

## AI Assistance

Claude Code authored this PR end to end — root-cause analysis, the diagnostic queries, the two-line change, and the dev-target validation build. Human-directed: the diagnosis originated from a `/dagster-day2` operations check, and Charlie identified `2332A` as a valid Miami school id and directed the string fix over a `safe_cast`.

## Self-review

### General

- [x] Properties file updated alongside the model
- [x] **Breaking change?** Yes — `sch` changes type on a contracted `stg_` model. Blast radius verified as nil: no downstream model reads `sch` from `stg_renlearn__fast_star`. The kipptaf copy is a `dbt_utils.union_relations` passthrough over the kippmiami staging model and inherits the type; it has no SQL consumers. `kippmiami` is the only project this builds in at all — `kippnewark` imports the local `renlearn` package but disables both `stg_renlearn__fast_star` and `src_renlearn__fast_star` in its `dbt_project.yml`, so the model never exists there. (Corrected from an earlier version of this description, which claimed the type also changed in kippnewark.)
- [x] No external source added or modified — the underlying `src_renlearn__fast_star` column is already `STRING`, so no re-staging is needed.

## Validation

Built against the dev target with contract enforcement (which only runs inside `dbt build`/CTAS):

```text
1 of 1 OK created sql table model zz_cbini_kippmiami_renlearn.stg_renlearn__fast_star
  [CREATE TABLE (11.5k rows, 4.2 MiB processed) in 4.43s]
Done. PASS=1 WARN=0 ERROR=0 SKIP=0 NO-OP=0 TOTAL=1
```

Both school id forms now survive the build:

```text
sch=2332,  sch_name=KIPP MIAMI-LIBERTY CITY, n=8788
sch=2332A, sch_name=KIPP MIAMI-LIBERTY CITY, n=2663
```

`trunk check --force` on both changed files: no issues.

## Notes for reviewers

Two things surfaced during the fix that are out of scope here:

1. The same school reports under two ids in the vendor feed — `2332` and `2332A`. Both are preserved faithfully now, but if downstream reporting ever groups by `sch`, that split will need reconciling.
1. `stg_renlearn__fast_star` has no uniqueness test, which the staging-layer convention calls for. Pre-existing gap, left alone rather than guessing at the grain.

Refs #4694


cbini:
Verified the kippnewark note and it's correct — `src/dbt/kippnewark/dbt_project.yml` disables both `stg_renlearn__fast_star` (L92-93) and `src_renlearn__fast_star` (L121-122), so the model never builds in that project and its column type doesn't change there. PR description corrected.

Net effect is that the blast radius is smaller than described, not larger: `kippmiami` is the only project that builds this model.

Also confirming the `FastStar` Pydantic `SCH: str | None` cross-check — that's the stronger version of the "already STRING in the source" claim, since it establishes the type at ingestion rather than just at the external table.

Issue #4698, opened 2026-08-03
fix: deliver minted student emails to Miami students already imported into Focus


cbini:
## Parent

Follow-on to #4512 (Deliverable 1 of #4511). Related to #4513.

## Problem

#4512 shipped the Miami student-email generator, and it works — every enrolled Miami student has a stable `@teamstudents.org` email minted in `stg_people__student_logins` and surfaced in the kipptaf desired-state extract `rpt_focus__demographics.stdt_email`.

But **420 of 1,584 currently-enrolled Miami students (AY2026) still have no email in Focus.** The value exists and is correct; it just never gets delivered.

Cause: the kippmiami `rpt_focus__demographics` wrapper is import-once. It anti-joins students already present in `stg_focus__students`, so it only sends demographics for students Focus has never seen. Every one of these 420 was already imported, so the wrapper skips them. It currently emits **0 rows**.

This is the exact risk called out in #4511 and #4512 ("the ~1,497 already-enrolled students need emails, which the current new-enrollee-only feed will not send"). It is now confirmed in production data.

## Impact

**Corrected 2026-08-03.** This issue originally claimed the gap was the binding constraint on #4513 (Google Workspace provisioning). **That was wrong, and it is now settled.**

We provision Google accounts ourselves, from the warehouse, via `rpt_google_directory__users_import` — not from Focus, and not via Clever. The email-source decision on #4513 has resolved to reading `stg_people__student_logins.google_email` directly, so provisioning does not round-trip the address through Focus.

**This issue therefore does not block #4513.** Account creation can proceed for the 412 of 420 affected students who already have a minted address available.

What this issue still affects:

- **Clever rostering.** Clever reads Miami's roster from Focus. A student with no `stdt_email` there is missing an address in every app Clever fronts.
- **Focus UX and any downstream consumer** that treats the Focus record as authoritative.

Current AY2026 state:

| Focus email state | Students | Has Google account |
| --- | --- | --- |
| Correct `teamstudents.org` | 1,163 | 1,130 |
| Staff domain (see below) | 1 | 1 |
| No email | 420 | 0 |
| **Total enrolled** | **1,584** | **1,131** |

Note the 0 in the bottom row is a consequence of the same PowerSchool-era key divergence described in #4513, not of the Focus gap — those students have no account because the provisioning model has emitted nothing for Miami since the PowerSchool archive froze.

## Scope of the fix

Of the 420 missing:

[...]

PR #4700, opened 2026-08-03
feat(dbt): provision Google Workspace accounts for Miami students from Focus

### LINKED ISSUE #4513: feat: provision live Google Workspace accounts for Miami (Focus) students
## Parent

Part of #4511 — Deliverable 2 (target 8/6).

## Goal

Restore Google Workspace account provisioning for Miami (Focus) students.

## Status — corrected 2026-08-03

The original framing of this issue was wrong in two ways. Both are corrected below; the original Options A/B are superseded.

**All four decisions are settled and the group prerequisite is confirmed. This is ready to build.**

One assumption is deliberately left untested — that Miami has no out-of-district concept, so the branch omits that org-unit routing term. It fails quietly if wrong. See "Note for later" at the end of Decisions.

**Clever does not provision Google accounts. We do.** Option A (let Clever provision from Focus) is not a real option and is removed.

**Provisioning for Miami is not new work — it is a pipeline that went dark.** `rpt_google_directory__users_import` and the `google_directory_user_create` asset already created Miami student accounts during the PowerSchool era. Verified in production: 3,442 Miami accounts carry an `externalIds` entry of type `custom` whose value is that student's bare PowerSchool number, which is exactly what the create asset writes.

## Why it stopped

`rpt_google_directory__users_import` reads `int_extracts__student_enrollments`, which is PowerSchool-backed. Miami's PowerSchool is a frozen archive ending AY2025, and `student_email` is null on every Miami row in that model.

The cause is a key divergence, not an exclusion. The model excludes only Paterson. But `stg_people__student_logins` now keys Miami rows by the Focus-prefixed student id, while the PowerSchool hub model carries bare student numbers. Nothing joins, so `where student_email is not null` silently drops all of Miami.

## Verified: no cross-student collisions

Worth recording because the generator's dedupe could not have prevented one. The live `stg_people__student_logins` retains only 1 of the 2,048 PowerSchool-era bare Miami rows — the rest live in `stg_google_sheets__people__student_logins_archive`, not the live table. So the dedupe predicate `username not in (select username from this)` never saw them.

Across all 3,905 minted Miami emails that match a live Google account:

| Account externalId belongs to | Accounts |
| --- | --- |
| the same student | 3,442 |
| a different student | 0 |
| no externalId present | 463 |

The deterministic name-plus-dob algorithm re-derived identical usernames for identical people, so the new minting re-adopted each student's own existing account. Correct outcome, but by construction rather than by guard. Any future change to the username algorithm loses that property.

## Work required

Give `rpt_google_directory__users_import` a Focus-backed source for Miami. `int_focus__student_enrollment` already carries `student_email`, `student_number`, `schoolid`, `school_title`, `academic_year`, `rn_year`, and `school_level` — the columns the model needs.

The Miami OU tree already exists and is correctly shaped: `/Students/Miami/{school}` with descriptions matching school names, which is the join the model already performs (`school_name = orgunit.description`). No OU work appears necessary.

## Decisions

**1. externalId key form — RESOLVED: use the prefixed Focus id.**

New Miami accounts write the Focus-prefixed student id as their `externalIds` value.

An earlier revision of this issue claimed the prefixed form would break `is_update` matching against the 3,442 existing accounts. **That was wrong.** Matching is by email throughout: `is_matched` in the model joins `student_email_google` to `primary_email`, and `batch_update_users` keys each request on `user["primaryEmail"]`. externalIds are never a match key.

The real consequence is narrower but still worth planning for. The update asset pops `student_number` and never sends externalIds, so the 3,442 existing accounts would otherwise keep their **bare** PowerSchool values indefinitely.

**Resolved: run a one-time migration** of those 3,442 accounts to the prefixed form, so Miami ends on a single scheme. This is a separate task from the model change and does not gate it — new accounts can start writing prefixed ids before the migration runs. Note the migration cannot go through the existing update path, which never sends externalIds; it needs its own one-off job.

**2. Email source — RESOLVED: read direct from `stg_people__student_logins.google_email`.**

Provisioning does not round-trip the address through Focus. This decouples account creation from the import-once feed and means work can start now: 412 of the 420 students currently missing an email in Focus already have a minted address available.

Note the Miami branch still needs `int_focus__student_enrollment` for the non-email attributes — `school_title` (org unit), `academic_year` / `rn_year` (current enrollment), grade level (for `changePasswordAtNextLogin`), and an enrollment-status equivalent (for `suspended`). Only the address comes from the generator, which also supplies `default_password`.

**3. Initial password and change-on-login policy — RESOLVED: carry the NJ behavior over unchanged.** Miami derives a default password the same way and sets `changePasswordAtNextLogin` for grade 3 and above. `stg_people__student_logins.default_password` already supplies the value for Miami students.

**4. Suspension semantics — RESOLVED: `int_focus__student_enrollment.enroll_status` already exists and uses the PowerSchool numeric domain.**

Focus has no scalar enrollment-status field. Status lives in `student_enrollment.drop_code` — a foreign key to `student_enrollment_codes` (Florida state W-codes, `type = 'Drop'`) — paired with `end_date`. But `int_focus__student_enrollment` already derives the translation:

```sql
case
    when dc.grad_type = 'graduated' then 3
    when e.drop_code is not null then 2
    else 0
end as enroll_status,
```

That maps 1:1 onto PowerSchool's domain — 0 currently enrolled, 2 transferred out, 3 graduated. **So the NJ expression `if(enroll_status = 0, false, true) as suspended` carries over unchanged.** No new derivation is needed.

Verified in production 2026-08-03:

| Academic year | status 0 | status 2 | status 3 |
| --- | --- | --- | --- |
| 2026 | 1,512 | 72 | 0 |
| 2025 | 14 | 1,486 | 21 |

The underlying columns agree perfectly — of 1,584 AY2026 rows, 1,512 have both `drop_code` and `end_date` null and 72 have both non-null, with no mixed rows. Dropped students never hold a second active stint, so the active and dropped sets are cleanly exclusive.

`int_focus__student_enrollment` also already exposes everything else the Miami branch needs: `grade_level` (for `changePasswordAtNextLogin`), `school_title` (for the org-unit join), `student_email`, `dob`, `entrycode` / `exitcode`, and `rn_year`.

Two caveats, neither blocking:

- **`grad_type` is narrower than its name suggests.** W07 "Graduated W/ Spcl Diploma", W08 "Left W/ Cert of Completion", W10 "Graduated W/ GED" and W27 are all tagged `non_graduate`, so a special-diploma graduate lands on status 2, not 3. Harmless for suspension — both are non-zero and suspend — but anything keying on status 3 specifically would undercount graduates.
- **Duplicate `short_name` values across ids** (W23 is both 1534 and 3699; WGD is 1504 and 3737; WD1 is 4257 and 4419). The model joins on `drop_code = c.id`, which is correct. A join on `short_name` would fan out.

## Note for later: out-of-district has no Miami equivalent

Recording this deliberately as an open question rather than a decision, because it may never matter.

The NJ path routes to `/Students/Disabled` on `suspended or is_out_of_district`. **`int_focus__student_enrollment` has no `is_out_of_district` equivalent**, and Focus surfaces no obvious analog — the nearest fields are `educational_choice_label` and `student_offender_transfer_label`, neither of which means out-of-district placement.

**Build the Miami branch without that term**, routing on `suspended` alone. Miami may or may not have an out-of-district concept; nobody has needed one so far.

How this would surface if the assumption is wrong: a Miami student placed out of district would keep an active, correctly-org-unit'd Google account instead of being routed to `/Students/Disabled`. It fails quietly — the account stays live and nothing errors. If someone reports a Miami student retaining access after an out-of-district placement, this is the cause, and the fix is to find the Focus field that encodes it and add the term back.

Who owns Miami offboarding operationally is likewise still open. It does not gate the model.

## Prerequisite — confirmed

The create asset fires a group-membership insert for every new account, against `groupKey` built as `'group-students-' || region || '@teamstudents.org'`, so Miami resolves to `group-students-miami@teamstudents.org`. There is no groups staging model, so this could not be confirmed from the warehouse.

**Confirmed present via the Admin console (2026-08-03).** No action needed. Recorded because a missing group would have let every Miami create succeed while its membership insert failed with "resource not found" — surfacing only as a warn-level asset check.

## Relationship to #4698

**Resolved by decision 2: #4698 does not block this issue.** Provisioning reads the minted address directly, so it does not wait on Focus delivery. #4698 remains needed for Clever rostering and for any consumer treating the Focus record as authoritative.

## Key files

- `src/dbt/kipptaf/models/extracts/google/directory/rpt_google_directory__users_import.sql`
- `src/teamster/code_locations/kipptaf/google/directory/assets.py` (create and update assets, externalIds)
- `src/dbt/focus/models/intermediate/int_focus__student_enrollment.sql`
- `src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments.sql`
- `src/dbt/kipptaf/models/people/staging/stg_people__student_logins.sql`

## Verification

Counts only, no PII. Confirm the provenance of existing Miami accounts with:

```sql
with miami_minted as (
  select
    sl.student_number as prefixed_id,
    cast(sl.student_number - 8400000000 as string) as bare_id,
    sl.google_email
  from `teamster-332318.kipptaf_people.stg_people__student_logins` as sl
  where sl.student_number >= 8400000000
)
select
  count(*) as minted_emails_matching_a_live_account,
  countif(x.value = m.bare_id) as account_belongs_to_same_student,
  countif(x.value != m.bare_id) as account_belongs_to_different_student
from miami_minted as m
inner join `teamster-332318.kipptaf_google_directory.stg_google_directory__users` as u
  on m.google_email = u.primary_email
left join unnest(u.external_ids) as x
```

Note on ids: `stg_focus__students.student_id` and `int_focus__student_enrollment.student_number` carry the Focus-prefixed form; `stg_focus__students.powerschool_id` carries the bare number. Joining the two forms against each other yields a silent zero-match rather than an error.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will restore Google Workspace account provisioning for KIPP Miami students, extend it to Paterson, and replace the fragile org-unit name matching that blocked Paterson in the first place.

Miami provisioning went dark when its PowerSchool became a frozen archive. `rpt_google_directory__users_import` reads `int_extracts__student_enrollments`, which is PowerSchool-backed, and every Miami row there has a null `student_email` — the login generator keys Miami on the Focus-prefixed student id while the PowerSchool hub carries bare numbers, so nothing joins and the `student_email is not null` filter silently drops all of Miami. The model never excluded Miami; the join key diverged.

This is not new provisioning. The same model and the `google_directory` assets already created 3,442 Miami accounts during the PowerSchool era — each carries an `externalIds` entry written by `user_create`.

### Three enrollment sources instead of one

- `powerschool_students` — the NJ regions.
- `focus_students` — Miami. Ranks every stint rather than the current year only; the PowerSchool branch already uses `rn_all`, so a current-year filter would hide a student whose last stint predates it and their account could never be suspended.
- `finalsite_students` — Miami students not yet imported into Focus. Finalsite is the admissions front door, so a new enrollee appears there first; sourcing only from Focus meant no account existed at their first import.

Precedence matters between the last two. A returning student whose last Focus stint is a prior year, but who is enrolled again in Finalsite for the current year, would otherwise be suspended by the stale Focus row. `focus_students` drops those rows so `finalsite_students` claims them as active — **147 students** who would have had working accounts disabled.

Addresses and default passwords come from `stg_people__student_logins` directly rather than through the Focus DEMOGRAPHICS feed, so provisioning does not wait on the delivery gap in #4698.

Suspension needed no new derivation: `int_focus__student_enrollment` already derives `enroll_status` on the PowerSchool numeric domain from the Florida state withdrawal codes.

### Org units are named, not matched

The org-unit join matched a student's school name against the Google org unit's free-text `description`. That required three school-name vocabularies to agree with one field nobody owned — Focus and Finalsite resolve to the locations sheet's name, PowerSchool keeps its own (`KIPP Hatch Academy` vs the sheet's `KIPP Hatch Middle`). It worked only because the descriptions happened to carry PowerSchool's names for NJ and the sheet's for Miami.

Paterson was the case where nobody filled it in: all three of its student org units have an empty description, so joining on it emitted 840 rows with a null `orgUnitPath`. That — not the email gap, which was already fixed — is why Paterson was excluded.

Ops now names the path directly in the `people__locations` sheet (`Google Student orgUnitPath`, named range v3, 28 of 42 rows populated). The extract resolves it through `int_people__location_crosswalk`, so every branch reaches the same row from whichever spelling its source uses — each row there is one alias. `stg_google_directory__orgunits` is still joined, now on the path itself, so a path that does not exist in Google resolves to null rather than being sent as a bad `orgUnitPath`. Rows with a null org unit are dropped explicitly instead of being provisioned at the domain root.

**Paterson is no longer excluded.**

## Validation

Ran the compiled logic against prod with the sheet staged to a dev copy. Aggregates only, no student rows.

| Branch | Region | Creates | Updates | Dropped for null org unit |
| focus | miami | 445 | 386 | 0 |
| finalsite | miami | 0 | 14 | 0 |
| powerschool | paterson | 389 | 454 | 0 |
| powerschool | newark | 10 | 10 | 0 |
| powerschool | camden | 6 | 5 | 0 |

Every region resolves an org unit. The v3 named range is column-only — 42 rows before and after, verified against prod.

Newark and Camden creates are warehouse drift since an earlier run in the same session (the directory pull refreshed); confirmed independently against `stg_google_directory__users`, not an effect of this change.

**The first prod run is large: roughly 1,700 API operations**, and the create path also fires a group-membership insert per new account. Notable inside the Miami updates: **68 withdrawn students whose Google accounts are still active** get suspended, and 15 suspended accounts belonging to currently-enrolled students get reactivated. The remainder are org-unit corrections and a one-time name-case normalization that settles after the first run.

## AI Assistance

Claude Code authored the model changes, the properties documentation, and the prod validation queries. Design decisions were human-directed — provisioning mechanism, `externalIds` key form, address source, password policy, the Finalsite-extends-Focus choice, and naming org units explicitly were all decided by the reviewer.

### General

### dbt

- [x] Properties updated for every touched model, including the new column on `stg_google_sheets__people__locations` and `int_people__location_crosswalk`.
- [x] Added `config.meta.contains_pii` and a `primaryEmail` uniqueness test to the extract, plus `unique` and `not_null` on `stg_people__student_logins.student_number` — the merge key two branches now join on, previously untested. Both pass against prod.
- [x] No exposure change — the consumer is the `google_directory` Dagster assets.
- [x] Not a breaking change: the extract's output column set is unchanged.
- [ ] **External source modified.** `sheet_range` moves to `src_google_sheets__people__locations_v3` and declares one new column. It has been staged to a dev copy for validation, but **`stage_external_sources --target staging` still needs to run** before dbt Cloud CI can see the column.

## CI checks

- [x] **Trunk** — `trunk check --force` clean on all six changed files.
- [ ] **dbt Cloud** — will fail until the external is staged to `--target staging` (see above).
- [ ] **Dagster Cloud** — not triggered; no Dagster paths changed.

Refs #4513

### DIFF
--- a/src/dbt/kipptaf/models/extracts/google/directory/rpt_google_directory__users_import.sql
+++ b/src/dbt/kipptaf/models/extracts/google/directory/rpt_google_directory__users_import.sql
@@ -14,10 +14,57 @@ with
 
             if(enroll_status = 0, false, true) as suspended,
         from {{ ref("int_extracts__student_enrollments") }}
+        /* Paterson stays excluded: its student org units carry an empty
+        description, which is the column the org-unit join below matches on, so
+        including it would emit rows with a null orgUnitPath. Re-enable once Ops
+        populates those descriptions -- see #4513. */
         where
             rn_all = 1
             and student_email is not null
             and region not in ('Paterson', 'Miami')
+    ),
+
+    finalsite_enrolled as (
+        /* Finalsite is the admissions front door, so a newly enrolled student
+        appears here before they are imported into Focus. Sourcing them here too
+        means their account exists by the time that first import runs. */
+        select
+            l.finalsite_enrollment_id,
+            l.assigned_school,
+            l.is_transfer_out,
+
+            cast(ida.focus_student_id_prefixed as int64) as student_number,
+
+            regexp_extract(l.grade_canonical_name, r'\d+') as grade_digits,
+        from {{ ref("int_finalsite__enrollment_lifecycle") }} as l
+        inner join
+            {{ ref("int_finalsite__contact_id_attributes") }} as ida
+            on l.finalsite_enrollment_id = ida.finalsite_enrollment_id
+        where
+            l.school_year_start = {{ var("current_academic_year") }}
+            and l.assigned_school is not null
+            and l._dbt_source_project = 'kippmiami'
+    ),
+
+    focus_ranked as (
+        /* Rank every stint, not just the current year's. The PowerSchool branch
+        uses rn_all, so restricting Focus to the current academic year would hide
+        a student whose last stint predates it -- their account could never be
+        suspended. Ordering matches int_powerschool__student_enrollment_union. */
+        select
+            student_number,
+            student_first_name,
+            student_last_name,
+            school,
+            grade_level,
+            region,
+            enroll_status,
+            academic_year,
+
+            row_number() over (
+                partition by student_number order by academic_year desc, exitdate desc
+            ) as rn_all,
+        from {{ ref("int_focus__student_enrollments") }}
     ),
 
     focus_students as (
@@ -26,11 +73,11 @@ with
         Focus, so provisioning does not wait on the import-once DEMOGRAPHICS
         feed that still owes ~420 enrolled students an address (#4698). */
         select
-            e.student_number,
-            e.student_first_name as first_name,
-            e.student_last_name as last_name,
-            e.school as school_name,
-            e.grade_level,
+            r.student_number,
+            r.student_first_name as first_name,
+            r.student_last_name as last_name,
+            r.school as school_name,
+            r.grade_level,
 
             sl.google_email as student_email_google,
             sl.default_password as student_web_password,
@@ -41,14 +88,61 @@ with
             /Students/Disabled -- see the note on #4513. */
             false as is_out_of_district,
 
-            lower(e.region) as region,
+            lower(r.region) as region,
 
-            if(e.enroll_status = 0, false, true) as suspended,
-        from {{ ref("int_focus__student_enrollments") }} as e
+            if(r.enroll_status = 0, false, true) as suspended,
+        from focus_ranked as r
+        inner join
+            {{ ref("stg_people__student_logins") }} as sl
+            on r.student_number = sl.student_number
+        left join finalsite_enrolled as fe on r.student_number = fe.student_number
+        /* A stale Focus stint must not outrank a live Finalsite enrollment. A
+        returning student whose last Focus stint is a prior year, but who is
+        enrolled again for the current year, is dropped here and picked up by
+        finalsite_students as active -- otherwise the prior year's withdrawal
+        would suspend an account that should stay open. */
+        where
+            r.rn_all = 1
+            and (
+                r.academic_year = {{ var("current_academic_year") }}
+                or fe.student_number is null
+            )
+    ),
+
+    finalsite_students as (
+        select
+            e.student_number,
+            e.is_transfer_out as suspended,
+
+            c.first_name,
+            c.last_name,
+
+            /* assigned_school carries the Finalsite spelling, which does not
+            always match the Google org unit description (KIPP Miami Tech vs
+            KIPP Miami Technical High). The crosswalk normalizes it. */
+            x.location_clean_name as school_name,
+
+            sl.google_email as student_email_google,
+            sl.default_password as student_web_password,
+
+            false as is_out_of_district,
+
+            /* the lifecycle model is scoped to kippmiami above */
+            'miami' as region,
+
+            if(e.grade_digits is null, 0, cast(e.grade_digits as int64)) as grade_level,
+        from finalsite_enrolled as e
+        inner join
+            {{ ref("stg_finalsite__contacts") }} as c
+            on e.finalsite_enrollment_id = c.finalsite_enrollment_id
+        inner join
+            {{ ref("int_people__location_crosswalk") }} as x
+            on e.assigned_school = x.location_name
         inner join
             {{ ref("stg_people__student_logins") }} as sl
             on e.student_number = sl.student_number
-        where e.academic_year = {{ var("current_academic_year") }} and e.rn_year = 1
+        left join focus_students as f on e.student_number = f.student_number
+        where f.student_number is null
     ),
 
     students as (
@@ -79,6 +173,21 @@ with
             region,
             suspended,
         from focus_students
+
+        union all
+
+        select
+            student_number,
+            first_name,
+            last_name,
+            school_name,
+            grade_level,
+            student_email_google,
+            student_web_password,
+            is_out_of_district,
+            region,
+            suspended,
+        from finalsite_students
     ),
 
     with_google as (
--- a/src/dbt/kipptaf/models/extracts/google/directory/rpt_google_directory__users_import.sql
+++ b/src/dbt/kipptaf/models/extracts/google/directory/rpt_google_directory__users_import.sql
@@ -14,14 +14,7 @@ with
 
             if(enroll_status = 0, false, true) as suspended,
         from {{ ref("int_extracts__student_enrollments") }}
-        /* Paterson stays excluded: its student org units carry an empty
-        description, which is the column the org-unit join below matches on, so
-        including it would emit rows with a null orgUnitPath. Re-enable once Ops
-        populates those descriptions -- see #4513. */
-        where
-            rn_all = 1
-            and student_email is not null
-            and region not in ('Paterson', 'Miami')
+        where rn_all = 1 and student_email is not null and region != 'Miami'
     ),
 
     finalsite_enrolled as (
@@ -191,6 +184,16 @@ with
     ),
 
     with_google as (
+        /* The org unit is named explicitly in the people__locations sheet rather
+        than matched against the org unit's free-text description. The three
+        branches above carry three different school-name vocabularies -- Focus
+        and Finalsite resolve to the sheet's name, PowerSchool keeps its own
+        (KIPP Hatch Academy vs the sheet's KIPP Hatch Middle) -- and the
+        crosswalk maps every one of them to the same row, since each row there
+        is one alias. Joining stg_google_directory__orgunits on the sheet's path
+        keeps it honest: a path that does not exist in Google resolves to null
+        and the student drops out below rather than being sent with a bad
+        orgUnitPath. */
         select
             s.*,
 
@@ -208,9 +211,11 @@ with
             {{ ref("stg_google_directory__users") }} as u
             on s.student_email_google = u.primary_email
         left join
+            {{ ref("int_people__location_crosswalk") }} as x
+            on s.school_name = x.location_name
+        left join
             {{ ref("stg_google_directory__orgunits") }} as o
-            on s.school_name = o.description
-            and o.org_unit_path like '/Students/%'
+            on x.location_google_student_org_unit_path = o.org_unit_path
     ),
 
     final as (
@@ -249,4 +254,7 @@ select
 
     if(grade_level >= 3, true, false) as `changePasswordAtNextLogin`,
 from final
-where is_create or is_update
+/* A null org unit means the school has no student org unit named in the
+locations sheet. Dropping those students is deliberate -- provisioning an
+account with no orgUnitPath would file it at the domain root. */
+where (is_create or is_update) and org_unit_path is not null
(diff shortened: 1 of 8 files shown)


src/dbt/kipptaf/models/extracts/google/directory/rpt_google_directory__users_import.sql
```diff
@@ -14,7 +14,71 @@ with
 
             if(enroll_status = 0, false, true) as suspended,
         from {{ ref("int_extracts__student_enrollments") }}
-        where rn_all = 1 and student_email is not null and region != 'Paterson'
+        where
+            rn_all = 1
+            and student_email is not null
+            and region not in ('Paterson', 'Miami')
```
cbini: Paterson should be included now

src/dbt/kipptaf/models/extracts/google/directory/rpt_google_directory__users_import.sql
```diff
@@ -14,7 +14,71 @@ with
...
+            /* Focus surfaces no out-of-district placement field; routing is on
+            suspension alone. If a Miami student is ever placed out of district
+            they will keep an active account instead of moving to
+            /Students/Disabled -- see the note on #4513. */
+            false as is_out_of_district,
+
+            lower(e.region) as region,
+
+            if(e.enroll_status = 0, false, true) as suspended,
+        from {{ ref("int_focus__student_enrollments") }} as e
```
cbini: this actually needs to come from finalsite. in order for students to be imported with emails, it has to be on the first run

src/dbt/kipptaf/models/extracts/google/directory/rpt_google_directory__users_import.sql
cbini: Confirmed the premise — all 842 AY2026 Paterson students have minted emails since `8d9db5bd1`, so the email-side reason for the exclusion is gone.

But I left it excluded, because enabling it emits **840 rows with a null `orgUnitPath`**. The org-unit join matches on `o.description`, and the Paterson student org units have an empty description:

| org_unit_path | description |
| --- | --- |
| `/Students/KIPP Paterson` | *(empty)* |
| `/Students/KIPP Paterson/Paterson Prep Elementary` | *(empty)* |
| `/Students/KIPP Paterson/Paterson Prep Middle` | *(empty)* |

Two things need to happen before this flips on. Ops has to populate those descriptions, and they have to match the student-side school names exactly — which are `Paterson Prep Elementary School` and `Paterson Prep Middle School`, carrying a `School` suffix the org unit leaf names don't have.

Left a comment on the exclusion pointing at this so it isn't mistaken for the stale email-era reason. Happy to open a separate issue for the Ops task if you want it tracked.

src/dbt/kipptaf/models/extracts/google/directory/rpt_google_directory__users_import.sql
cbini: Done in `289e5bbe4` — added a `finalsite_students` branch off `int_finalsite__enrollment_lifecycle` covering current-year Finalsite enrollees that Focus doesn't yet carry, so an account exists before the first Focus import.

One thing worth knowing if you touch this again: `assigned_school` can't join to org unit descriptions directly. Finalsite says `KIPP Miami Tech`, the org unit says `KIPP Miami Technical High` — a direct join resolves **0 of 99** students at that school. Routing through `int_people__location_crosswalk` resolves all 1,773. Focus only worked because its wrapper already normalizes via the locations sheet.

The bigger catch was an interaction with review finding 1. Widening the Focus branch to rank every stint (rather than current-year only, so departed students can still be suspended) collided with this change: **147 students** are enrolled in Finalsite for AY2026 but their most recent Focus stint is a prior year. The widened Focus branch marked them withdrawn and would have suspended working accounts. `focus_students` now drops a stale row when Finalsite has a live current-year enrollment for that student, so `finalsite_students` claims them as active.

Validated: focus 445 creates / 385 updates, finalsite 0 creates / 14 updates, no null org unit or address on any row.

PR #4710, opened 2026-08-03
feat(dbt): add strand-level assessment rows and unify response_type (1 of 2)

### LINKED ISSUE #4708: feat(dbt): add DIBELS subtest and i-Ready domain scores to the assessment scores fact
## Problem

`fct_assessment_scores_enrollment_scoped` exposes only the summary score for the
two vendor benchmark diagnostics:

- **DIBELS** is filtered to `measure_standard = 'Composite'`, so subtest
  (strand) results — Nonsense Word Fluency, Phoneme Segmentation Fluency, Oral
  Reading Fluency, Letter Naming Fluency, etc. — never reach the mart or Cube.
- **i-Ready** reads only the subject-level diagnostic row from
  `int_iready__diagnostic_results`, so the 14 per-domain results (Phonics,
  Vocabulary, Comprehension, Algebra and Algebraic Thinking, etc.) are absent.

Analysts want strand/domain-level cuts for both. Internal Illuminate
assessments already model exactly this shape via the `response_type` family, so
the same pattern should carry the vendor breakdowns rather than inventing a
parallel structure.

## Approach

Populate the existing `response_type` / `response_type_description` columns for
the new breakdown rows, keeping `module_code` anchored to the summary grain so
FKs to `dim_assessments` and `dim_assessment_administrations` still resolve.
Those two dims are subject-grain and stay unchanged.

Naive alternative rejected: dropping the `Composite` filter and letting
`measure_standard` flow into `module_code`. Both dims independently re-derive
their DIBELS rows with the same `Composite` filter, so every subtest row would
be an FK orphan.

Mapping follows the real values already in the fact (verified against prod):

```text
summary row    response_type = NULL,    response_type_description = NULL
breakdown row  response_type = 'group', response_type_description = <subtest or domain name>
```

`'group'` is the established value for a named category-level breakdown with no
standards code; `response_type_code` and `response_type_root_description` stay
NULL, matching that precedent. Keeping the summary row at NULL makes the change
purely additive — the documented `response_type notSet` filter for vendor
sources keeps working.

Both sources ship in one PR so there is a single surrogate-key migration event.

Design spec and implementation plan to follow on the linked branch.

## Dependency

**Resolved as of #4709 (merged), with one carry-over.** That PR landed
`test_round` and `_dbt_source_project` on `int_iready__domain_unpivot`, but not
`illuminate_subject` — which the vendor branch needs to join the resolver
(`va.illuminate_subject = sr.subject_area`), so domain rows without it would all
drop. That pass-through is now folded into this work as the first commit of the
implementation PR rather than tracked as an external dependency; it is an
additive upstream edit, which the marts convention permits.

#4709 also added an upstream `relative_placement is not null` filter, which
changes this design's domain CTE. See the dependency-update comment below for
both, plus re-verified row counts. The DIBELS half never had a dependency and
could still ship alone.

## Follow-up items for a data engineer

Surfaced during design; none are blockers for this PR, all warrant a look.

1. **Rejected alternative worth revisiting — dedicated columns.** Instead of
   reusing the `response_type` family, the breakdown could get its own
   `assessment_component_code` / `assessment_component_name` pair. Rejected here
   because reuse needs no new fact columns or Cube members and `'group'` is a
   genuine semantic fit, but dedicated columns are cleaner if `response_type`
   should stay strictly Illuminate-scoped.

2. **i-Ready `is_mastery` decode is unguarded.** Domain rows have no ordinal
   placement column upstream (only the subject-level
   `overall_relative_placement_int` exists), so mastery is derived by testing
   `relative_placement` against two literal labels
   (`Early On Grade Level`, `Mid or Above Grade Level`) — the same threshold
   `stg_iready__diagnostic_results` applies at subject level. A new or reworded
   i-Ready label would silently read as *not mastered* rather than failing.
   Suggested guard: an `accepted_values` test on `relative_placement`. Note the
   vocabulary already has a 6th value, `Not Assessed`, present in
   `phonological_awareness` and `high_frequency_words` only, which this change
   filters out as an explicit not-administered marker.

3. **`pct_proficient` changes meaning for these two sources.** It is documented
   as the source-agnostic headline metric. A query filtering only
   `assessment_type = 'dibels'` or `'iready'` currently returns summary-only
   proficiency; afterwards it blends summary and breakdown rows unless
   `response_type notSet` is added. Same for `count_scores` and
   `avg_scale_score`. This PR updates the Cube MCP knowledge docs so the agent
   filters correctly, but **saved Tableau workbooks and Superset queries reading
   `student_assessment_scores_view` need a human audit** — they cannot be
   enumerated from the repo.

4. **Surrogate-key values churn on STAR too.** Adding an eighth hash input to
   the shared vendor-branch `assessment_score_key` changes the key value for
   every vendor row, including STAR rows that gain no data. Verified that no dbt
   model and no Cube join consumes this column (the only `ref()` is the Cube
   exposure; Cube uses it solely as a `primary_key` dimension). Worth confirming
   no external system persisted those key values.

5. **`int_iready__domain_unpivot` has no uniqueness test** — a pre-existing gap
   also noted as out-of-scope in #4706. The fact-level `unique` PK test becomes
   the de facto guard; a test on the intermediate would be better.

6. **i-Ready `placement` column goes unused.** #4706 adds an absolute
   (non-grade-relative) `placement` alongside `relative_placement`. This change
   consumes only `relative_placement`, matching the existing subject-level
   branch. Surfacing the absolute scale would need its own fact column.

## Out of scope

- DIBELS progress-monitoring rows (the `assessment_type = 'Benchmark'` filter
  stays).
- State-assessment subclaim or strand breakdowns.
- Unifying `response_type` across all sources — state rows would need to change
  too, and it would break the documented `notSet` filter.
- Domain-level `national_percentile`; i-Ready publishes percentiles only at
  subject level.

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will add strand-level assessment rows to `fct_assessment_scores_enrollment_scoped` and give `response_type` a single, non-nullable vocabulary across every source.

**This is the first of two merges.** It was one PR until 2026-09-10; the Cube half now lives in a separate PR based on this branch, and merges after this one. See _Reviewer Notes_ for why the order matters.

Today the fact carries only the summary score for our two vendor diagnostics. DIBELS is filtered to the Composite measure, and i-Ready reads only the subject-level row. Teachers and school leaders cannot see which skill a student is actually behind on. This change adds those breakdown rows:

- **DIBELS subtests** — eight measures per sitting alongside the Composite, roughly 269,600 rows.
- **i-Ready domains** — twelve domain names across Reading and Math, four to seven per sitting, roughly 1,207,200 rows. The `comprehension_overall` rollup is excluded so the comprehension construct is not counted three times.

Breakdown rows use the existing `response_type` columns. `module_code` stays anchored to the summary grain, so the foreign keys to `dim_assessments` and `dim_assessment_administrations` still resolve.

The change also fixes a long-standing problem with `response_type` itself. It used to be null for every non-Illuminate source **and** for every Illuminate assessment a student was assigned but never took. Those are two different facts sharing one value, and neither could be queried on its own. `response_type` is now non-nullable with four values:

| Value | Meaning |
| --- | --- |
| `overall` | The summary score. Every source. |
| `group` | A named breakdown — Illuminate reporting groups, i-Ready domains, DIBELS subtests. |
| `standard` | A single standard-level response. Illuminate only. |
| `not_taken` | Assigned but no recorded response. Illuminate only. |

**What is in this PR:** the fact model and its properties file, the i-Ready domain unpivot and its properties file, and the design spec, implementation plan, baseline capture and rollback runbook.

**What is not:** every Cube change. Those are in the stacked Cube PR, which must not merge until this one has and the fact has rebuilt in production.

Refs #4708.

**Rollback plan.** The fact is `materialized: table`, so a plain revert fully restores the prior output on the next build. The runbook at `docs/superpowers/plans/2026-08-28-strand-scores-rollback.md` gives the ordered steps for both merges, and asserts every check against the captured baseline rather than against numbers typed into a document.

## Reviewer Notes

**Merging this PR alone changes no metric value, but it does change a filter that consumers use.** No Cube file moves with this PR, so `pct_proficient` and `count_scores` keep their current definitions. What changes is the data: `response_type` stops being null. **Anything outside this repository that filters for a null or not-set `response_type` to reach vendor or state rows returns nothing from the moment this merges** — not when the Cube PR merges. It must use `response_type = 'overall'` instead. Superset charts, Tableau workbooks on the Cube SQL API, and direct Cube callers all sit inside Cube, so no dbt exposure covers them. This is #4708 follow-up 3, and it is a merge gate on **this** PR.

**Merge gate owner: @anthonygwalters.** The rollback runbook (`docs/superpowers/plans/2026-08-28-strand-scores-rollback.md`) requires a named owner for the consumer announcement before merge, and it cannot be closed from inside the repo — the audience it targets is reachable only by a human. @anthonygwalters owns sending it. This is the gate that binds **this** PR; the claude.ai Project upload gate, also @anthonygwalters, sits on the stacked Cube PR #5248.

**Row counts rise, so `count_scores` and `pct_proficient` move even though their definitions do not.** Measured 2026-09-10 on a local dev server, with `main`'s cube model reading a dev-schema build of this branch: global `count_scores` roughly 14,959,733 (up from 13,485,050) and global `pct_proficient` roughly 45.21% (down from 45.66%). Treat those two as indicative, not exact — that run was served by a pre-aggregation whose partitions had mixed build times.

**Why the split, in one line.** Cube Cloud redeploys on merge to `main` while the fact rebuilds hours later, so a single merge always puts the new Cube model in front of the old fact. Measured 2026-09-10: in that state DIBELS, i-Ready, STAR and all five NJ/FL state sources each read exactly 0 — 1.32M rows, silently, with no error. Merging the fact first removes the window.

**Breaking change, twice over.** `assessment_score_key` changes value on every vendor row including STAR, because `response_type_code` joins the surrogate key. And every Illuminate `not_taken` row's key changes too, because the 4th hash input flips from null to `'not_taken'`. Both are pure key-value changes, not new or removed rows. No column is renamed or removed, so the model contract is unaffected.

**Known limitation, accepted.** 5,565 DIBELS subtest rows have a null `is_mastery` because `measure_standard_level_int` is null upstream, concentrated in the K-2 phonics measures. Those rows sit in a proficiency denominator without entering the numerator. Tracked on #4708, not fixed here.

**#4708 closes when this merges, which is early.** This branch is linked to #4708 through the Development sidebar, so merging closes the issue while the Cube PR is still open. Either detach the linkage before merging, or reopen #4708 afterwards. #4708 also still holds open follow-ups beyond this work.

### General

- [x] No Asana update needed — tracked on #4708

Not applicable — no Dagster changes.

- [x] Properties files updated for both changed models (`fct_assessment_scores_enrollment_scoped.yml`, `int_iready__domain_unpivot.yml`), including `accepted_values` and `not_null` on `response_type` and `accepted_values` on `relative_placement`
- [x] Exposure unchanged — `models/exposures/cube.yml` already covers this fact, and no columns were added or removed
- [x] **Breaking change?** Yes, in two ways, both documented above and in the rollback runbook.
- [x] Both models build clean in a dev schema: `PASS=15 WARN=0 ERROR=0` on 2026-09-10, with `--defer --favor-state` against a freshly regenerated prod manifest

Not applicable. The spec, plan, baseline and rollback documents live under `docs/superpowers/`, which is excluded from the MkDocs nav on purpose.

## CI checks

- [x] **Trunk** — passes on the current head.
- [ ] **dbt Cloud** — this PR selects models, so this check is real validation and not a no-op. It must be green before merge, and its warnings must be pulled with `warning_only=true`: the `relationships` test on `assessment_administration_key` is `warn` severity and is the guard against orphaned breakdown rows.

<details>
<summary>For Claude</summary>

## AI assistance

Claude-authored spec, plan and implementation; human-directed design. Every decision point was raised as an explicit choice and settled by the requester: combined-design scoping, reusing the `response_type` family over dedicated columns, decoding i-Ready placement labels for `is_mastery`, retaining placement-only domain rows, and unifying `response_type` across every source rather than leaving vendor rows null. The two-PR split was decided on 2026-09-10 after dev-server validation found the deploy-ordering hazard.

An adversarial design review with three parallel reviewers ran against the spec on 2026-08-28 and found five blockers, all since fixed. A verification pass on 2026-09-02 re-checked every finding against the implemented branch. Both are posted as PR comments and predate the split.

## Detail behind the Reviewer Notes flags

**Why `module_code` stays anchored.** Letting `measure_standard` flow into `module_code` orphans every subtest row. `dim_assessments` and `dim_assessment_administrations` are siblings of this fact, not descendants, and each re-derives its DIBELS rows with the same Composite filter. Cube's proficiency measures gate on `student_assessments.assessment_key IS NOT NULL`, so orphans would vanish from those measures while still inflating `avg_scale_score` — and `module_code` is not exposed on the scores cube, so no consumer could scope around it. Note the fact itself has no `assessment_key` column; Cube reaches it by traversal.

**The i-Ready dedupe needed the discriminator.** Its `partition_by` includes `module_code`, which domain rows deliberately share with the anchor. `response_type_code` had to join the partition key or every domain row for a sitting, plus its anchor, would collapse to one row, silently. Even with the fix the partition still collapses 319,380 of 1,572,858 eligible domain rows; 95.8% of that is the `TODO(#4387)` fiscal-year re-pull collapse and is intended.

**`response_type_code` is populated for the first time.** i-Ready gets `domain_name`, DIBELS gets `measure_standard`, and `response_type_description` gets `measure_name`. Verified 1:1 across all Benchmark rows. It remains null for Illuminate `group` rows, where the code is null across all 2,867,069 rows while the description varies.

**Enrollment scoping is asymmetric, and that is recorded rather than fixed.** `int_assessments__resolved_section_enrollments` still filters DIBELS to Composite and i-Ready to a non-null `overall_scale_score`. A DIBELS subtest row for a student with no Composite row, and an i-Ready diagnostic with domain placements but no subject-level scale score, both have no resolver row and are dropped by the inner join. Widening the resolver is deferred so the row reconciliation is not misread as a defect.

**i-Ready ingestion.** The spec and plan describe a July ingestion stall as an open dependency. That is stale: both regions materialized on the 2026 partition on 2026-09-01, and `int_iready__domain_unpivot` holds 73,464 FY27 rows. #4951 was closed unmerged, so it is no longer the tracking PR for it.

**Row counts against the 2026-08-28 estimates.** The plan estimated roughly 252,300 DIBELS and 1,139,444 i-Ready breakdown rows. A dev-schema build on 2026-09-10 produced 269,609 and 1,207,216. The gap is the i-Ready FY27 partitions landing on 2026-09-01, after the estimates were taken. The runbook carries the newer figures.

**Measurement method, for anyone re-running it.** The dev-server numbers above came from the local Cube server (`npm --prefix src/cube run dev`) with a locally-chosen API key, a minted HS256 token carrying an `email` claim, and the scores cube's `sql_table` temporarily pointed at `zz_cristinabaldor_kipptaf_marts`. Every measure read is served by the `proficiency_rollup` pre-aggregation rather than the fact — confirmed through the `/sql` endpoint — so per-source counts and the global count can disagree within a single capture when partitions have different build times. Two such gaps were seen (670 rows and 84,237 rows) and were not isolated further. Order-of-magnitude findings are unaffected; treat sub-0.1% deltas as noise from that layer.

</details>


docs/superpowers/specs/2026-08-03-assessment-strand-level-scores-design.md
```diff
@@ -0,0 +1,353 @@
...
+grows; its partitioning and build range are unchanged.
+
+Documentation is a required deliverable, because this change redefines an
+existing metric rather than only adding a new capability. `pct_proficient` is
+documented as the source-agnostic headline metric; a query filtering only
+`assessment_type = 'dibels'` or `'iready'` currently returns summary-only
+proficiency and afterwards blends summary and breakdown rows unless
+`response_type notSet` is added. `count_scores` and `avg_scale_score` shift the
+same way. The Cube MCP knowledge docs currently assert these sources are
+summary-only, which would actively steer the chat agent away from the filter it
```
cristinabaldor, line 300: rather than relying on mcp instructions for this would it be better to split these into a summary cube or model and a by standard cube or model? or is it not worth it if someone asking for the breakdown in the same chat would mean the LLM has to join on the assessment id anyway?

anthonygwalters:
## Adversarial design review

Claude-run review against **current production** (`main` @ `99c4cbabe`), not the PR base. The branch is **1054 commits behind** `4c7dde53d`, and several load-bearing claims have drifted.

Three reviewers with separate lenses (staleness/drift, semantic-layer + consumer blast radius, SQL implementability), plus coordinator verification against prod data. **Claims marked "verified" were run against the warehouse**; the rest are static code review.

**Verdict:** the design reasoning is largely sound. The SQL is not implementable as written, three of the five Validation bullets cannot detect what they exist to detect, and there is a data-modeling defect that no Cube tuning fixes.

---

## Blockers

### 1. Implemented literally, the spec ships a no-op

Two reviewers found this independently. `vendor_all` (`fct_assessment_scores_enrollment_scoped.sql:309-358`) is three explicit 12-column `SELECT` blocks, and the vendor final `SELECT` hardcodes `cast(null as string) as response_type` at `:612-615`. **Neither is mentioned in the spec.**

Ten edits are required; the spec supplies five:

| # | Edit | Location | In spec? |
|---|---|---|---|
| 1 | `dibels_scores` rewrite | `.sql:284-307` | yes |
| 2 | `iready_scores_raw` + 2 literals | `.sql:178-202` | prose only, no SQL, no position |
| 3 | `iready_domain_scores_raw` (new) | — | yes |
| 4 | `iready_all_raw` (new) | — | **no SQL at all** |
| 5 | `iready_scores` dedupe re-target | `.sql:214-228` | yes |
| 6 | `vendor_all` i-Ready branch +2 | `.sql:310-323` | no |
| 7 | `vendor_all` STAR branch +2 casts | `.sql:327-340` | no |
| 8 | `vendor_all` DIBELS branch +2 | `.sql:344-357` | no |
| 9 | Final vendor `SELECT` 612, 614 | `.sql:612,614` | no |
| 10 | Vendor key + 8th input | `.sql:551-563` | yes |

Marts are `+contract: enforced: true` (`dbt_project.yml:212-215`), so the literal implementation compiles, passes the contract, passes `unique`, and emits every breakdown row with `response_type IS NULL` — indistinguishable from summary rows in Cube. That is precisely the quiet failure the spec's own *Rejected approach* section argues against.

Note the spec says "`star_scores` and its dedupe are untouched" — true of the CTE, false of the union. STAR has no response-type columns anywhere in its lineage and `star_scores` is the output of `bigquery__deduplicate`'s `select unique.*`, so the literals must be written into the `vendor_all` STAR block.

### 2. `iready_all_raw` is referenced but never defined

The dedupe block passes `relation="iready_all_raw"`, described in one sentence with no SQL. The two CTEs it must union are written in **different column orders**:

- `iready_scores_raw` (13 cols): `... proficiency_level, score_source, scale_score, national_percentile, is_mastery`
- `iready_domain_scores_raw` (15 cols): `... proficiency_level, response_type_description, score_source, response_type, scale_score, national_percentile, is_mastery`

The spec says the anchor "gains only the two NULL discriminator literals" without saying **where**. Appended at the end, a positional union puts `score_source` against `response_type_description` — both STRING, so it unions silently with values swapped, then errors one column later on a NUMERIC/STRING mismatch. This is the most guess-dependent part of the spec and the one place it supplies no code.

Also: `int_iready__domain_unpivot` emits `rn_subject_test`, which `iready_all_raw` must drop or the anchor must fake.

### 3. i-Ready has a parent-child domain hierarchy — the spec asserts it does not

The spec's justification for `'group'`: "a named category-level breakdown carrying no standards code **and no parent-domain rollup** ... `'standard'` does not — neither source has a CCSS-style code **or a parent hierarchy**."

**Verified against prod:**

| domain_name | years | rows |
|---|---|---|
| `comprehension_literature` | 2022-2026 | 119,756 |
| `comprehension_informational_text` | 2022-2026 | 119,756 |
| **`comprehension_overall`** | 2022-2026 | **119,756** |
| `reading_comprehension_literature` | 2020-2021 | 17,338 |
| `reading_comprehension_informational_text` | 2020-2021 | 17,338 |

`comprehension_overall` is the rollup of its two siblings — identical row counts across the same five years. `_sum_proficient` and `count_scores` are a plain `SUM`/`COUNT` over rows, so every 2022-2026 Reading sitting would contribute the comprehension construct **three times** to both numerator and denominator. Any `pct_proficient` cut at `response_type = 'group'` is silently comprehension-weighted, and summing across domains double-counts.

The fact already has the column for this: `response_type_root_description`.

One reviewer claimed *two* rollup families plus a possible 6x if both co-occur. **Corrected by the data:** `reading_comprehension_*` is the pre-rename 2020-21 naming, temporally disjoint from `comprehension_*`, and its `_overall` produces zero rows (dropped by the upstream `relative_placement is not null`). So there are **13 live domains, not 14**, and exactly one parent. Real, but smaller than first reported.

**Required before build:** either exclude the `*_overall` rows, or populate `response_type_root_description` on the children and document that summing `'group'` rows double-counts. Do not ship all of them as flat siblings.

### 4. The orphan argument and FK validation cite a column that does not exist

The spec's *Rejected approach* section and its Validation bullet 4 both reference `assessment_key` **on this fact**. It has never existed — not on `main`, and not at the spec's own base commit `4c7dde53d`. It lives on `dim_assessment_administrations`; Cube reaches it by traversal (`student_assessment_scores` -> `student_assessment_administrations` -> `student_assessments`).

Two consequences:

- The mechanism description is wrong. The fact hashes only the administration key, so it is a single-FK orphan — and it is caught by the existing `relationships` test the spec never mentions (`properties/...yml:55-59`). "The failure would be quiet" survives only because `dbt_project.yml:23` sets `+severity: warn`; it would be a CI **warning**, not silence.
- Validation bullet 4 is **not runnable and vacuous**. `assessment_administration_key` is `generate_surrogate_key` output, which never returns NULL — the macro coalesces every input to `'_dbt_utils_surrogate_key_null_'` before hashing. A 100%-orphaned breakdown population would show a 100% non-null rate and pass. And under the spec's own anchoring design the key is byte-identical between summary and breakdown rows, so the null-rate reads 0% in both the success **and** failure cases.

Replace with: "the existing `relationships` test on `assessment_administration_key` must stay green — it is `warn` severity, so pull it with `get_job_run_error(warning_only=true)`."

### 5. i-Ready ingestion has been stalled for 41 days

**Verified in Dagster 2026-08-28:** `kippmiami/iready/diagnostic_results` and `kippnewark/iready/diagnostic_results` both last materialized **2026-07-18**. Open PR #4951 documents the cause — Curriculum Associates renamed the FY27 SFTP exports; the FY27 partition holds FY26 rows.

Building the i-Ready half now ships domain rows for a source that is not ingesting current-year data. The DIBELS half has no such dependency and could ship alone.

---

## Major

### The `not_null` test the spec calls "the real proof" was deleted

`d51920ff8 perf(dbt): remove vacuous not_null tests from marts primary keys` (2026-08-07, four days after the spec) removed it. Current state is `unique` only (`properties/...yml:41-42`). The spec's parenthetical was also self-contradictory on day one — it cites the repo convention ("never add `not_null` to `generate_surrogate_key` output") and then names the violation of that convention as proof. Validation bullet 2 repeats it. As written this walks an implementer into adding a forbidden test.

### The DIBELS `is_mastery` claim is false

Spec: "`measure_standard_level_int` is a pass-through column populated on **every** `measure_standard` row ... so the existing `>= 3` threshold applies per subtest unchanged."

**Verified against prod:**

| bucket | rows | pct mastery | null `measure_standard_level_int` |
|---|---|---|---|
| Composite (today) | 58,124 | 45.2% | **0** |
| Subtests (new) | 255,144 | 48.0% | **5,565** |

Nulls concentrate in the K-2 phonics measures: NWF-WRC / NWF-CLS / WRF at 1,321 each, LNF / PSF at 801. `NULL >= 3` yields NULL, so those rows get `is_mastery = NULL` — counted in the denominator of `pct_proficient`, never in the numerator. One reviewer marked this claim as holding on structural grounds; the data overrides it.

### The dedupe does more than the spec admits

The spec's fix (adding `response_type_description` to the partition) is correct and is the sharpest catch in the document. But **verified:** even with the fix, the partition still collapses **319,380 of 1,572,858 domain rows (20.3%)**.

| collapse signature | groups |
|---|---|
| differ only by `academic_year_int` (intended re-pull collapse) | 274,292 |
| differ in `scale_score` | 9,286 |
| differ in `relative_placement` | 4,321 |
| differ in `start_date` | 1,761 |

95.8% is the `TODO(#4387)` fiscal-year re-pull collapse — correct behavior. But ~13.6K groups carry genuinely different scores or placements for the same student, subject, round, date, and domain, where `order by start_date desc, scale_score desc` silently keeps the higher score. Pre-existing behavior, now replicated 13x per test.

Also unmodelled: `bigquery__deduplicate` materializes an `array_agg` whole-row struct per input row. Feeding it ~14x the rows is a real cost change on a table rebuilt 5x/day.

### `proficiency_rollup` cannot serve any query this feature exists for

The rollup's dimension list omits `student_assessments.assessment_type`. Cube only serves a query from a rollup when every requested and filtered member is present — and `assessment-cube-reference.md:14-19` names `assessment_type` as the canonical way to select i-Ready or DIBELS, while the orchestrator doc makes explicit `response_type` filtering a per-session non-negotiable.

So the exact query shape this feature exists to serve (`assessment_type = 'iready'` + `response_type = 'group'`) falls back to the fact **every time**. "New rows enter the existing pre-aggregation automatically" is true about contents and false about benefit: ~1.8M new rows get materialized into a rollup that can never answer a question about them.

`assessment_type` is near-functionally-determined by `module_code` (already in the rollup), so adding it costs close to zero rows. Highest-value single Cube edit in this change.

### The resolver's filters are narrower than the new CTEs

`int_assessments__resolved_section_enrollments.sql:138-156` still filters DIBELS to `measure_standard = 'Composite'` ("PM probes and subskill measures are out of scope"), and its i-Ready CTE filters `overall_scale_score is not null`.

- A DIBELS subtest row for a student/period with **no** Composite row has no resolver row and is dropped by the INNER JOIN. `int_amplify__all_assessments.sql:181-200` explicitly coalesces missing composites to `'No data'`, so such students exist.
- An i-Ready diagnostic with domain placements but a null `overall_scale_score` passes the spec's domain filter but has no resolver row and is dropped.

The scoping gate is **asymmetric** between anchor and breakdown rows. The spec's 1,572,858 figure is an upper bound on a population it never bounds from below. Whether to widen the resolver is a design question the spec never asks.

### Nothing guards either vocabulary

No `accepted_values` on `response_type` anywhere in the fact yml, the Cube models, or `models/assessments/`. None on `relative_placement` either.

The i-Ready threshold translation is correct — `stg_iready__diagnostic_results.sql:163-175` maps five labels to 1-5 with `Early On Grade Level -> 4` and `Mid or Above Grade Level -> 5`, so `overall_relative_placement_int >= 4` is equivalent to the spec's two-label set. But the **failure semantics differ**: the upstream `case` has no `else`, so an unknown label yields NULL; the spec's `IN (...)` yields FALSE. A reworded i-Ready label would be silently scored as *not proficient* rather than *unknown*, across ~1.5M rows feeding the headline metric. #4708 follow-up 2 is the wrong severity for that.

### The DIBELS breakdown vocabulary is a hand-maintained Google Sheet

`int_amplify__all_assessments.sql:45-52` INNER JOINs `int_google_sheets__dibels_expected_assessments` on `measure_standard = expected_measure_standard`, keyed on (academic_year, region, assessment_grade_int, admin_season). So the set of values reaching a public Cube dimension and the pre-aggregation varies by year x region x grade x season and changes when someone edits a spreadsheet. `measure_standard` is undocumented in the model contract, and that sheet model has no uniqueness test and no declared columns.

### The spec's DIBELS subtest names are from the wrong column

The spec's prose promises "Nonsense Word Fluency, Phoneme Segmentation Fluency, Oral Reading Fluency, Letter Naming Fluency." Those are `measure_name` values. The code writes `measure_standard`. **Verified actual vocabulary:**

| measure_standard | rows |
|---|---|
| Composite | 58,124 |
| Reading Fluency (ORF) | 50,824 |
| Reading Accuracy (ORF-Accu) | 43,644 |
| Reading Comprehension (Maze) | 43,101 |
| Decoding (NWF-WRC) | 29,117 |
| Letter Sounds (NWF-CLS) | 29,117 |
| Word Reading (WRF) | 29,117 |
| Letter Names (LNF) | 15,112 |
| Phonemic Awareness (PSF) | 15,112 |

Neither "Phoneme Segmentation Fluency" nor "Letter Naming Fluency" exists in either vocabulary, and NWF splits into **two** measures the spec does not anticipate. Subtest total is 255,144 — a 5.4x multiplier on DIBELS rows.

This also matters against the spec's own design table, which defines `'group'` as code-NULL / description "full text". `'Decoding (NWF-WRC)'` is a parenthesised-code label, closer to the `'standard'` shape the spec rejects. Decide `measure_standard` vs `measure_name` and make prose and code agree.

### DIBELS uniqueness at the widened grain was never re-verified

The spec's rewrite deletes the `-- DIBELS benchmark composites are unique at this grain upstream (verified); no dedupe needed` comment without re-verifying at the new grain. Upstream, `int_amplify__all_assessments.sql:158-168` dedupes on `(surrogate_key, round_number, measure_standard)` — and `round_number` is **not** in the fact's key inputs.

**Verified: it holds.** 313,268 DIBELS rows produce 313,268 distinct 8-column keys. No dedupe needed. But this was luck rather than analysis — the risk was real, and the failure mode would have been a post-build `unique` failure on a table Cube already reads.

### Cube pre-aggregation migration window

`proficiency_rollup` is ~12 yearly partitions with no `incremental: true` and no `update_window`, each served from its last successful build until its replacement lands. A change that flips `response_type` on every historical row means some partitions carry the old value and others the new during the refresh sweep — a multi-year query returns **partial history with no error**. Mitigate by bumping the pre-agg name to force a clean rebuild, or drop/ship/rebuild/re-add.

`src/cube/CLAUDE.md:124-138` already mandates validating a pre-agg build on a branch staging deployment before merge (incident #4460 -> revert #4462). The spec's Validation section has no pre-aggregation step at all.

---

## On unifying `response_type` across all sources

The requester has since directed that summary rows be promoted to `'overall'` across every source rather than left NULL, with the reference docs and saved queries updated. Findings re-scoped for that.

### The spec's premise for keeping NULL does not hold

The spec's value table says NULL means "every non-Illuminate source." **Verified — it does not:**

| source | rows with `response_type IS NULL` |
|---|---|
| **illuminate** | **1,073,422** |
| iready | 240,749 |
| dibels | 57,546 |
| state (all NJ + FL) | 62,378 |
| star | 7,349 |

So `response_type notSet` — the filter the spec is protecting — already returns 74.5% Illuminate rows. It is not selective today, and the "purely additive, keeps the documented filter working" argument is defending something already broken. This strengthens the case for unification.

### Those Illuminate NULL rows are a deliberate record, and need their own token

They are **not** a join artifact. `int_assessments__scaffold` produces the canonical "expected to take" grain by cross-joining the assessment catalog against course enrollments; `int_assessments__response_rollup.sql:44-47` left-joins responses onto it. A surviving NULL is an **assessment assigned but not taken**.

Verified: all 1,073,422 carry NULL `scale_score`, `percent_correct`, `is_mastery`, and `response_type_description`, across every module type (QA 210,563; CRQ 160,317; TP 95,118; MQQ 90,023; ...).

So NULL currently carries three unrelated meanings, and the not-taken population is **un-queryable in isolation** — as is the vendor summary population. Proposed target:

| meaning | today | unified |
|---|---|---|
| expected to take, not taken | `NULL` | `not_taken` |
| summary score, **all** sources | `overall` (Illuminate) / `NULL` (vendor, state) | `overall` |
| named category breakdown | `group` | `group` (vendor breakdowns join here) |
| standards breakdown | `standard` | `standard` |

**A blanket `NULL -> 'overall'` would relabel 1.07M not-taken records as scores.** The branch-aware version is safe.

Note there is no existing token for this anywhere in dbt or Cube (`is_completion_row` appears only in open Tableau PR #3576, on a different surface), so naming needs a decision.

### Keep it scoped to the fact's UNION branches

`int_assessments__response_rollup` is read by ~15 downstream models filtering the lowercase vocabulary — `rpt_tableau__power_standards`, `rpt_deanslist__mod_standards_domains`, `rpt_gsheets__assessment_roster`, `int_topline__formative_assessment_weekly`, `rpt_tableau__qbl`, and others. Changing the intermediate breaks all of them.

Worth flagging: the repo already runs **three** `response_type` vocabularies — lowercase here, Title-case (`Overall`/`Group`/`Subject`/`Total`) in the CARAT lineage, plus one-offs. "Consistent across every data source" read literally is much bigger than this PR.

### Populate `response_type_code` while you are in there

The pre-agg comment claiming `response_type` and `response_type_description` are "functionally determined by `response_type_code` (zero added rows)" is **already false** — Illuminate `'group'` rows carry a NULL code with populated, varying descriptions. Both designs make it worse; unification makes it worst.

Setting `response_type_code` to the token (i-Ready `domain_name`, DIBELS `measure_standard`) and `response_type_description` to the human label makes code -> (type, description) genuinely 1:1 for the first time, makes the comment true rather than aspirational, and fixes the fact that the spec currently puts codes into a column documented as "human-readable description." Same edit, three problems.

### The doc surface is eight locations, not two

`assessment-cube-reference.md` asserts summary-only at `:23-25` (global vocabulary), `:26-31` (the `notSet` instruction), `:143`, `:269` (i-Ready), `:350` (DIBELS), `:372` (STAR), `:391` (NJ state), `:431` (FL state). The spec's doc table names i-Ready and DIBELS only.

Also missed: `:108-112` hard-codes "null on all **302,907** i-Ready, DIBELS, and STAR rows" (goes ~6x stale), and `:317-325` gives an i-Ready dedup recipe — "one row per student per window" — that at the new grain collapses 14 rows into 1 and **destroys the feature silently** for any analyst following it.

Plus `student_assessment_scores_view.yml:9-10` ("Illuminate only"), the four `response_type*` descriptions in the fact yml (each ends "Null for state assessments"), the `avg_scale_score` Grain clause (its documented safe boundary — source/subject/grade — stops being safe when Composite pools with ORF words-per-minute), and the pre-agg comment.

### Sequencing is the real risk, not the decision

`src/cube/mcp/project_knowledge/README.md:3-6`: the repo is source of truth, the claude.ai Project is the **deployment target** — changed files must be **manually re-uploaded**. Meanwhile the fact is `materialized: table` on `0 0,10,13,15,17 * * *`, so the data change lands within hours of merge, ungated.

The moment `'overall'` lands, every agent following the still-uploaded `notSet` protocol gets zero rows for i-Ready, DIBELS, STAR, NJ and FL state — silently. **The re-upload has to be sequenced ahead of the model merge with a named owner in the PR body**, not left as a doc-table row.

And "we're in a good place to update queries" holds only for what the repo can enumerate. Repo-enumerable consumers: the 13 files above. Not enumerable: Superset saved charts, Tableau workbooks on the Cube SQL API, direct Cube REST/GraphQL callers, ad-hoc BigQuery. Those sit **inside** Cube, so no dbt exposure covers them by construction. The #4708 follow-up 3 audit becomes a **prerequisite**, not a follow-up.

### Add `accepted_values` on `response_type` in this PR

Under unification the column becomes non-nullable in practice, which makes the test tractable for the first time. It is the cheapest item in this review and the guard the change most needs — a typo (`'Overall'`, `'group '`) currently ships clean, and the only loud guard is `unique` on the PK, which catches a *missing* discriminator, not a *wrong* one.

---

## What the spec got right

Verified, no change needed:

- **The `'group'` shape table is exact.** 2,867,069 rows: `response_type_code` 0% populated, `response_type_description` 100%, `response_type_root_description` 0%. Reusing `'group'` is well-founded.
- **The sibling-dim orphan mechanism is real.** `dim_assessments.sql:294-295` and `dim_assessment_administrations.sql:290-291` both re-derive DIBELS with the same `Composite` filter.
- **The i-Ready dedupe catch** is the sharpest thing in the document and correctly reasoned. NULL groups cleanly for anchor rows, so the `TODO(#4387)` semantics are preserved per-domain.
- **The row counts reproduce exactly.** `int_iready__domain_unpivot` = 1,635,032; after filters = 1,572,858. Both match the spec, which had flagged them as unconfirmed. The stated blocker ("prod had not yet rematerialized") no longer applies.
- **The surrogate key genuinely discriminates.** Vendor key inputs number exactly 7, so "eighth and last" is right, and appending *last* is load-bearing — `'Reading Accuracy (ORF-Accu)'` contains the `-` delimiter, harmless only in final position. Verified: 313,268 DIBELS rows -> 313,268 distinct keys.
- **No fan-out on the resolver join.** `int_assessments__resolved_section_enrollments` is deduped to one row per `score_grain_key`; the vendor join is exactly that key minus the NULL `canonical_assessment_id`, so all domains match one row each.
- **`illuminate_subject` is still absent** from `int_iready__domain_unpivot` — zero commits under `models/iready/` since base. Still the required first commit, and "do not re-derive downstream" is right.
- **The upstream `relative_placement is not null` rule is correctly not repeated** in the fact.
- **`term_key` / `reporting_terms` / the singular test** are unaffected — breakdown rows carry the same `assessment_date_key` and school as their summary row, so the new singular test evaluates identically. No new failures.
- **DIBELS `module_code`** genuinely does not move: `measure_standard as module_code` under the Composite filter emits exactly the literal.
- **RLS is unaffected.** Policies filter on `region_key` / `abbreviation` via `student_section_enrollment_key`; breakdown rows share their summary row's section enrollment. No new PII columns.
- **All four `response_type*` members are on the public view**, and the blast-radius claim is right — the only non-doc consumer of this fact is Cube, plus one singular test.
- **Contract is unaffected** — adding rows changes no column shape or `data_type`.

Corrections to the spec's own claims where it was *more* right than it thought: the `unique` test alone is sufficient proof of the discriminator, and the i-Ready mastery threshold translation is exactly equivalent to the upstream ordinal.

---

## Two corrections to the spec's other claims

- **"No dbt model reads this fact. The only `ref()` is `models/exposures/cube.yml`"** is now false. `d19af9567` added `tests/fct_assessment_scores_enrollment_scoped__term_covers_assessment_date.sql:13`, which refs the fact. The corrected claim: the only model-graph consumers are the Cube exposure and one singular test.
- **The unit-test failure mode is overstated in the direction that matters.** The `given` block must gain `illuminate_subject` (loud). But dbt derives the compared column set from `expected_rows[0].keys()` only, so omitting the column from **all 12** expect rows compiles, runs, and **passes** with the column silently uncompared. The real risks are partial edits (loud UNION mismatch) and inconsistent key order (silent — `illuminate_subject`/`domain_name`/`placement` are all STRING, so BigQuery unions positionally into the wrong columns).

---

## Process

- **Merging this docs-only PR will auto-close #4708.** The issue carries a `connected` timeline event to this PR (2026-08-03T21:55:08Z, one second after creation — the `gh issue develop` linkage). There is no closing keyword in the body or commits, but the Development linkage closes it on merge. #4708 still holds six open follow-up items and is the tracking issue for the implementation.
- **`claude-review` never ran here.** The `claude` check-run on `b26969c2` is `completed/skipped`, and there are zero issue comments. The self-review checklist item referencing it is unsatisfiable as-is. To trigger it, toggle draft state via GraphQL.
- CI is otherwise green (CodeQL, Trunk Check, dbt Cloud). `mergeable_state: blocked` is the CODEOWNERS review requirement, not a CI failure.
- The branch is 1054 commits behind `main`; merge `origin/main` before further work.

---

## Suggested minimum bar before implementation

1. Fix the domain hierarchy (blocker 3) — a modeling defect independent of the vocabulary decision, and it will not get easier once 1.5M rows are in prod.
2. Write the missing SQL: `iready_all_raw`, all three `vendor_all` branches, and `.sql:612`/`:614`. Pin the anchor CTE's literal positions explicitly.
3. Decide the internal-branch NULL policy (`not_taken` vs. coalesce) and correct the key-churn blast-radius claim accordingly.
4. Fix Validation bullets 2, 3, and 4 — as written they cannot detect the failures they exist to detect. Bullet 3 ("rows where `response_type is null` must be unchanged") becomes meaningless under unification; replace with a per-`score_source` before/after comparison.
5. Add `accepted_values` on `response_type` and on i-Ready `relative_placement`.
6. Add `student_assessments.assessment_type` to `proficiency_rollup`.
7. Sequence the knowledge-doc re-upload ahead of the merge, with a named owner.
8. Validate the pre-aggregation build on a branch staging deployment before merge, per `src/cube/CLAUDE.md:124-138`.
9. Resolve the i-Ready ingestion stall (or ship the DIBELS half alone).

### AI assistance

Review run by Claude with three parallel reviewers on separate lenses, plus coordinator verification. Prod queries were run for every claim marked "verified"; one reviewer's blocker was corrected against the data (two rollup families reduced to one), and one reviewer's "claim holds" verdict on `measure_standard_level_int` was overturned by it. Static-only claims are cited to `file:line` and should be spot-checked before acting.

docs/superpowers/specs/2026-08-03-assessment-strand-level-scores-design.md
anthonygwalters: (wrapped into adversarial review pass)

anthonygwalters:
## Response to the adversarial design review

Re-verified every finding against the current head (`6d18076f`), not the state the
review ran against. Warehouse claims below were re-run today (2026-09-02); the rest
is code reading at that commit.

**Two things changed since the review:** the branch merged `origin/main` and then
implemented the design, so this PR is no longer docs-only — it now carries the fact
model, its properties, the i-Ready unpivot, both Cube files, and the three knowledge
docs. The PR body has not caught up and still says "Docs only" (see New findings
below).

---

## Blockers

| # | Finding | Verdict |
|---|---|---|
| 1 | Implemented literally, the spec ships a no-op — 10 edits needed, 5 supplied | **Fixed.** All three `vendor_all` branches now carry the discriminators (`.sql:441-443`, `:462-464`, `:482-484`), and the vendor final `SELECT` sources them from `va.*` (`.sql:740-742`) instead of hardcoding NULL. The STAR branch got the literals in the union, as the review specified. |
| 2 | `iready_all_raw` referenced but never defined | **Fixed.** Defined at `.sql:265-303` with an explicit 16-column list on **both** union branches, so the positional-union swap the review predicted cannot occur. `rn_subject_test` is dropped by the explicit list. |
| 3 | i-Ready has a parent-child domain hierarchy | **Fixed by exclusion.** `and domain_name != 'comprehension_overall'` (`.sql:262`), with the reason written into the CTE comment. Row reconciliation restated as 1,453,102 after the exclusion. |
| 4 | Orphan argument and Validation bullet 4 cite `assessment_key`, which is not on this fact | **Fixed.** Spec now states the fact has no such column and never has; Validation bullet 4 replaced with "the existing `relationships` test on `assessment_administration_key` must stay green — `warn` severity, pull it with `get_job_run_error(warning_only=true)`." |
| 5 | i-Ready ingestion stalled 41 days | **No longer true — resolved, not deferred.** Both `kippmiami/iready/diagnostic_results` and `kippnewark/iready/diagnostic_results` materialized within the last day on the `2026` partition (Newark ELA 7,381 records; Miami ELA 1,491). `int_iready__domain_unpivot` holds 73,464 FY27 rows with `max(completion_date) = 2026-09-01`. The design and plan still describe this as an open dependency — that text is now stale. Note #4951 was **closed unmerged** on 2026-09-01, so the reference to it as the tracking PR needs re-pointing. |

---

## Major

| Finding | Verdict |
|---|---|
| The `not_null` test called "the real proof" was deleted | **Fixed.** Spec cites `d51920ff8` and states convention forbids re-adding it. Separately, the `not_null` now added on `response_type` (properties `:172`) is a different and correct thing — a plain column, not surrogate-key output. |
| DIBELS `is_mastery` claim is false; 5,565 subtest rows have null `measure_standard_level_int` | **Documented, not fixed.** Recorded on the `is_mastery` column description (properties `:203`) and in the deferred list. Those rows still sit in a proficiency denominator without ever entering the numerator. Accepting as a known limitation is defensible; it should be a named follow-up on #4708 rather than prose in a column description. |
| The dedupe collapses 20.3% even with the fix | **Documented.** Numbers written into the CTE comment with the 95.8% re-pull split. Behavior unchanged, which is the right call for a `TODO(#4387)` item. |
| `proficiency_rollup` cannot serve the queries this feature exists for | **Fixed.** `student_assessments.assessment_type` added to the rollup (`:398`) with the reasoning in-comment. |
| Resolver filters narrower than the new CTEs; scoping gate asymmetric | **Accepted and recorded, not fixed.** Spec `:614-621` states the asymmetry explicitly and marks widening the resolver as deferred, so the row counts are not read as a defect. Reasonable — but it means the i-Ready and DIBELS breakdown populations are silently gated on the presence of a summary row. |
| Nothing guards either vocabulary | **Fixed.** `accepted_values` on `response_type` (`[not_taken, overall, group, standard]`) plus `not_null`; `accepted_values` on `relative_placement` with all six labels. The i-Ready `IN (...)` vs upstream-`case`-with-no-`else` semantic gap is now guarded by that test. |
| DIBELS vocabulary is a hand-maintained Google Sheet | **Unaddressed.** The sheet dependency still shapes a public Cube dimension and the pre-aggregation, and `int_google_sheets__dibels_expected_assessments` still has no uniqueness test or declared columns. Not a blocker, but it belongs on #4708 rather than dropped. |
| Spec's DIBELS subtest names came from the wrong column | **Fixed, and better than either option the review offered.** `response_type_code` takes `measure_standard`, `response_type_description` takes `measure_name` — both populated rather than choosing one. Verified: `measure_name` is 1:1 with `measure_standard` across all 327,795 Benchmark rows (zero `measure_standard` values map to more than one name, zero nulls in either), so the rollup's "code determines description" claim genuinely holds for the vendor rows. |
| DIBELS uniqueness at the widened grain never re-verified | **Fixed.** Comment replaced with the re-verified figure at the new grain. |
| Cube pre-aggregation migration window | **Fixed.** Renamed `proficiency_rollup` to `proficiency_rollup_v2` (`:379`) with the partial-history reasoning in-comment, forcing one clean rebuild. Plan Step 7 adds the branch-staging validation `src/cube/CLAUDE.md:124-138` requires. |

---

## Unification items

| Finding | Verdict |
|---|---|
| A blanket NULL to `'overall'` would relabel 1.07M not-taken records as scores | **Fixed.** Internal branch uses `coalesce(rr.response_type, 'not_taken')` (`.sql:28`), with the scaffold/left-join reasoning in-comment. Vendor and state summary rows get `'overall'` explicitly. |
| Keep it scoped to the fact's UNION branches | **Honored.** `int_assessments__response_rollup` untouched, so the ~15 downstream consumers of the lowercase vocabulary are unaffected. |
| Populate `response_type_code` while you are in there | **Fixed.** i-Ready gets `domain_name`, DIBELS gets `measure_standard`. The rollup comment was also corrected rather than left aspirational — it now states the mixed determination honestly (vendor: determined; Illuminate `'group'`: not, code is NULL across all 2,867,069 rows) instead of asserting "zero added rows". |
| The doc surface is eight locations, not two | **Fixed.** `assessment-cube-reference.md` rewritten (the `notSet` instruction now reads as historical, the stale 302,907 figure is gone), plus `README.md` non-negotiable 2, the orchestrator's rule 3, the view description, and all four `response_type*` descriptions in both the fact yml and the cube. |
| Sequencing is the real risk | **Partly.** The rollback runbook has a real merge gate (`:256-284`) and correctly splits the claude.ai Project update into its two distinct mechanisms. But the gate's own requirement — "name an owner for both halves in the PR body" — is not satisfied; see below. |
| Add `accepted_values` on `response_type` in this PR | **Fixed** (above). |
| `avg_scale_score` grain boundary | **Fixed, and extended past what the review asked.** The description now walks it to `response_type_code` as the only safe single-breakdown boundary, because all eight DIBELS subtests share `response_type = 'group'`. |

---

## Three new findings

**1. The PR body and title are stale, and the merge gate depends on the body.**
The body still says "**Docs only. No model, YAML, or Cube changes.**" and the title
is `docs:`. The PR now changes two dbt models, two properties files, both Cube
files, and three knowledge docs — so the squash-merge commit will be typed `docs`
for a change that alters a mart fact and a public Cube view. The body also says
"dbt Cloud CI expected to be a trivial no-op: this PR selects zero models. Green
here is not validation of anything," which is no longer true and would mislead a
reviewer into discounting the one CI signal that now matters. And the rollback
runbook's merge gate requires a **named owner** for both halves of the claude.ai
Project update **in the PR body** — there is no owner named. The gate is currently
unsatisfiable by its own terms.

**2. Validation bullet on `not_taken` pins a literal count that has already drifted.**
The spec asserts `count(response_type = 'not_taken')` must equal **1,073,422**.
Measured in prod today, the equivalent population
(`response_type is null` plus null `scale_score`, `percent_correct` and
`is_mastery`) is **1,073,302** — 120 rows lower after five days. Total
null-`response_type` rows are 1,451,610 today against 1,441,444 the day of the
review, and DIBELS Benchmark rows are 327,795 against 313,268. The fact rebuilds
5x/day, so a literal equality will fail on a correct implementation. This is the
same defect class the review corrected in bullets 2, 3 and 4 — assert
`count('not_taken')` equals the pre-change `count(response_type is null)` captured
in the baseline doc **in the same session**, not a number typed into the spec.

**3. `count_students` is the one public measure that still counts not-taken rows.**
Six measures gained `response_type != 'not_taken'`; `count_students` (cube `:181`)
did not, and it carries no filters at all. The behavior is unchanged from `main`, so
this is not a regression — but its description still reads "distinct students with a
score in the filtered slice," and this change is precisely what makes "has a score"
a nameable distinction. `count_students` and `count_scores` now disagree about what
counts as a score, and only one of them says so. Either add the filter or say in the
description that it counts students expected to take, not students who scored.

---

## Process items

- **Merging still auto-closes #4708** via the Development linkage, and #4708 is still
  open with follow-ups (now including the 5,565 null `measure_standard_level_int`
  rows and the DIBELS Google Sheet guard). Detach the linkage or reopen after merge.
- **`claude-review` still has not run.** No `claude` check-run on `6d18076f`. The
  self-review checklist item remains unsatisfiable without toggling draft state via
  GraphQL — and it matters more now than it did when the PR was docs-only.
- CI is green on the current head (CodeQL, Trunk Check, both Analyze jobs).
  `mergeable_state: blocked` is still the CODEOWNERS requirement.
- The branch is 40 commits behind `main` again after the 2026-08-29 merge.

### AI assistance

Verification run by Claude. Warehouse figures re-run against prod on 2026-09-02;
Dagster materialization state read live. Line citations are against `6d18076f`.

cristinabaldor:
@anthonygwalters — heads up, I restructured this PR today. Summary of what changed and why, so nothing is a surprise.

_Edited 2026-09-10 to cover a second round of changes since the original note: the comment and description cleanup, two new tracking items, and a full row-level validation._

**What I found.** I brought the branch up to date with `main` and ran a before-and-after validation of the Cube measures on a local dev server. The new measure filter was written `{CUBE}.response_type != 'not_taken'`. Against the current production fact, where `response_type` is null on every non-Illuminate row, `null != 'not_taken'` evaluates to null rather than true — so the row is dropped. In that state DIBELS, i-Ready, STAR and all five NJ/FL state sources each returned **exactly 0**. That is 1.32M rows, silently, with no error.

That state is not hypothetical. Cube Cloud redeploys on merge to `main` and the fact rebuilds hours later, so a single merge always produces it for a period.

**What I changed.**

1. Switched the six measure filters to `IS DISTINCT FROM 'not_taken'`, which counts null rows instead of dropping them.
2. Split the change in two. This PR is now the dbt half — the fact model, the i-Ready unpivot, both properties files, and the docs. The Cube half is #5248, based on this branch. Merge this one, let the fact rebuild, then merge #5248. That way `proficiency_rollup_v2` builds against the correct fact on its first build.
3. Revised the rollback runbook for the two-merge shape. The ten branch commit SHAs it listed are unreachable from `main` after a squash merge, so they are replaced by the two squash commits, and the revert order is now Cube-then-dbt with the reason stated.
4. Corrected one number. The body said global unfiltered `pct_proficient` moves 45.80% → 49.54%. That pair was Illuminate-only rather than global, and it has drifted. Measured 2026-09-10: global is 45.66% → 48.24%, Illuminate-only is 46.17% → 49.73%.
5. Trimmed the SQL comments on the fact, at the requester's direction. Three dated row censuses removed (they go stale silently and a reader cannot tell a moved number from a broken one), an 8-line paragraph that had been copy-pasted into both the i-Ready and STAR dedupe blocks stated once, and several lines cut that restated a `not_null` test or pointed at other documentation. The file now carries **50 comment lines against `main`'s 57**, while adding 73 lines of code. Every mechanism explanation survives — only the restatements went.
6. Moved consumer-facing content out of SQL comments and into descriptions. The i-Ready row-inclusion rules (why `comprehension_overall` is excluded, why placement-only domains are kept) are now in the model description. The per-source mastery thresholds are now in the `is_mastery` description in both the dbt properties and the cube — that definition previously existed **only** in an SQL comment, so no BI user or chat-agent caller could reach it.
7. Removed a row count from the `is_mastery` description ("Null on 5,565 DIBELS subtest rows"). Same staleness problem as the SQL censuses, and `.claude/rules/dbt-yaml.md` asks descriptions to use qualitative language.

**Validation.** Posted a full row and measure reconciliation as a separate comment on this PR: https://github.com/TEAMSchools/teamster/pull/4710#issuecomment-5626084245

Short version — **zero rows lost, 1,476,429 gained and all of them the intended breakdown rows, and the summary grain is bit-identical to `main`.** Every vendor and state source at `response_type = 'overall'` matches `main` on row count, mastery count, and rate to four decimals. Illuminate's scored population is identical at 12,173,456 rows and 6,053,158 mastery; the rate moves only because the never-taken rows leave the denominator.

One thing worth knowing before you verify any of this yourself: reading the measures through Cube gave a 670-row and a 146-row Illuminate discrepancy that turned out to be `proficiency_rollup` serving partitions built at different times. Against the fact tables those deltas are exactly zero. Validate against the fact, not the semantic layer, until `_v2` has built once cleanly.

**Two things that are yours to decide.**

- The consumer audit gate moved to **this** PR. The null-`response_type` filter breaks when the fact changes, not when the Cube model changes, so anything outside the repo filtering on it returns nothing from this merge onward.
- This branch is still linked to #4708 in the Development sidebar, so merging this closes the issue while #5248 is still open. Either detach the linkage first or reopen #4708 afterwards.

The claude.ai Project upload gate still lists you as the owner and now sits on #5248. One wrinkle: the `response_type` vocabulary flips when **this** PR merges, but the knowledge files ship on #5248, so uploading right after this merge is slightly better than waiting.

**Two follow-ups came out of the review, neither blocking this PR.**

- #5256 — Cube dimension descriptions have drifted from their dbt mart columns, 113 of 172 comparable dimensions. `sync_cube_descriptions.py` cannot correct drift by design: it skips any dimension that already has a description, copies once rather than linking, ignores measures, and nothing runs it. Worth a look because it means a dbt description does not reach BI. It also carries one open design question that touches #5248 — see below.
- #5258 — adds the concrete cuts and a length ceiling to the inline-SQL-comment rule in `.claude/rules/dbt-sql.md`. The rule already existed and was read; this PR still shipped a 16-line comment above a 5-line macro call, so the rule needed teeth rather than restating.

On #5256 and #5248 together: the `is_mastery` cube description I added carries one sentence the dbt one does not — that `pct_proficient` counts this, so a cross-source rate compares different definitions. That references a Cube measure with no dbt equivalent, which is exactly the case #5256 has to settle. Either we align the two before #5248 merges, or it becomes the worked example. Your call, and it does not block this PR.

Your two earlier review comments predate all of this and still stand — nothing in them was reverted.

cristinabaldor:
## Row and measure validation against `main`

Ran the reconciliation this PR needed but did not have: does the branch lose any rows, gain any it should not, and do the existing measures survive?

**Answer: zero rows lost, 1,476,429 gained and all of them intended, and the summary grain is bit-identical to `main`.**

### Method

Compared the PR CI build (`dbt_cloud_pr_70403104388001_4710_marts`, built 2026-09-10 22:01 UTC) against prod `kipptaf_marts`, both joined to the same prod `dim_assessment_administrations` and `dim_assessments`. Each pair was read inside a single query so both sides see the same moment.

This is deliberately a direct table comparison rather than a Cube read. An earlier pass through Cube showed a 670-row and a 146-row discrepancy on Illuminate that I could not account for; against the tables those are exactly zero, so that gap was the `proficiency_rollup` pre-aggregation serving partitions of different vintages, not the data. Worth knowing for anyone validating this through the semantic layer.

### Rows: nothing lost, nothing unexpected

Every source's `null` count on `main` becomes its `overall` count on the branch, exactly:

| Source | `main` (null) | branch `overall` | new `group` rows |
| --- | ---: | ---: | ---: |
| dibels | 61,076 | 61,076 | +269,545 |
| iready | 255,339 | 255,339 | +1,206,884 |
| star | 7,349 | 7,349 | — |
| state_nj_njsla | 35,108 | 35,108 | — |
| state_fl_fast | 17,741 | 17,741 | — |
| state_nj_njsla_science | 6,313 | 6,313 | — |
| state_nj_njgpa | 2,720 | 2,720 | — |
| state_fl_science | 702 | 702 | — |
| state_fl_eoc | 257 | 257 | — |

Illuminate: 939,982 nulls become 939,982 `not_taken`, and `group` / `overall` / `standard` come back with deltas of **0, 0, 0**.

Total gained: **1,476,429 rows, every one `response_type = 'group'`** — the DIBELS subtests and i-Ready domains this PR sets out to add. Total lost: **0**.

### Measures: the summary grain does not move

Filtering the branch to `response_type = 'overall'` reproduces `main` exactly for every vendor and state source — row count, mastery count, and rate to four decimal places:

| Source | rows | mastery | rate |
| --- | --- | --- | --- |
| iready | 255,339 = 255,339 | 63,613 = 63,613 | 24.9132% = 24.9132% |
| dibels | 61,076 = 61,076 | 27,755 = 27,755 | 45.4434% = 45.4434% |
| state_nj_njsla | 35,108 = 35,108 | 10,180 = 10,180 | 28.9962% = 28.9962% |
| state_fl_fast | 17,741 = 17,741 | 4,117 = 4,117 | 23.2061% = 23.2061% |
| star | 7,349 = 7,349 | 2,469 = 2,469 | 33.5964% = 33.5964% |
| state_nj_njsla_science | 6,313 = 6,313 | 445 = 445 | 7.0489% = 7.0489% |
| state_nj_njgpa | 2,720 = 2,720 | 1,065 = 1,065 | 39.1544% = 39.1544% |
| state_fl_science | 702 = 702 | 229 = 229 | 32.6211% = 32.6211% |
| state_fl_eoc | 257 = 257 | 148 = 148 | 57.5875% = 57.5875% |

Illuminate's scored population is identical on both sides: **12,173,456 scored rows and 6,053,158 mastery**. The rate moves 46.16% to 49.72% purely because the 939,982 never-taken rows leave the denominator. The numerator does not move at all, which is what makes this a fix rather than a redefinition.

### What a consumer sees

- Filtering `response_type = 'overall'`: **numbers unchanged from today.**
- Querying unfiltered: `count_scores` rises by the 1.48M breakdown rows, and Illuminate proficiency rises 3.56 points.
- Still filtering `response_type IS NULL` to reach vendor or state rows: **returns nothing.** That is the merge gate already flagged in the PR description, and it bites at **this** merge, not the Cube one.

### Caveat

Prod is live and rematerialises, so these are same-moment snapshots. A re-run tomorrow will show different absolute counts with the same deltas. The reconciliation holds because each comparison read both tables in one query.

Measure definitions themselves (`pct_proficient`, `count_scores`) live in the stacked Cube PR, so this validation covers both halves.

cristinabaldor:
@claude update your review

cristinabaldor:
## Verdict on the updated review

Verified both findings against the files rather than the commit messages. One was sharper than it read, one is partly a convention question, and one item is not mine to close.

### Finding 1a — hardcoded `not_taken` assertion: **valid, fixed**

`design.md` asserted `count(response_type = 'not_taken')` must equal the literal **1,073,422**. Confirmed present, and confirmed wrong: measured against the branch build 2026-09-10 the count is **939,982**.

The review called this "the same defect class the spec explicitly fixed one bullet up." That understates it — it is the same defect *in the same bullet*. The bullet is marked `**[revised]**` precisely because the draft's version was unassertable, and the revision made the `overall` half baseline-relative while leaving the `not_taken` half a typed literal.

Fixed in `5488afaf8`: both halves are now baseline-relative, with the reason recorded so the number does not get written back. The population shrinks through the year as students sit the assessments they were assigned — 12% in thirteen days — so any literal here is wrong by the time it is read.

### Finding 1b — spec documents `!= 'not_taken'`: **valid, fixed, and worth more than "documentation-accuracy"**

The review classed this as a doc gap because the code lives in #5248 and is already fixed there. Agreed on impact, but the failure mode is worse than staleness: the spec is the record someone consults to understand *why* the filter reads as it does, and it currently instructs the reader toward the form that returned exactly 0 rows for every vendor and state source. Someone "restoring consistency with the spec" would reintroduce the bug.

Fixed at both sites (`design.md`, `plan.md`) with the measurement that forced the change and an explicit "do not simplify it back."

### Finding 1c — i-Ready ingestion stall: **valid, fixed by annotation not rewrite**

Marked resolved 2026-09-01 at both sites, keeping the original text as the record of why the design reads as it does.

### Not changed, deliberately

The `1,073,422` figures in the spec's Problem section and the plan's order-of-magnitude block stay. Those describe prod as of 2026-08-28, and the plan explicitly labels them **"NOT the assertion basis — every before/after assertion compares against that file."** They are record, not instructions. `.claude/rules` guidance is that a stale spec is expected; a stale *assertion* is not, which is why 1a was fixed and these were not.

### Finding 2 — unsatisfied merge gate: **valid, and open**

The runbook (`:363`, `:370`) requires a named owner for two things before merge. Current state:

- **claude.ai Project update** — @anthonygwalters is named, but in **#5248's** body, since the three knowledge files ship there. The runbook asks for it in the PR body; it is satisfied in substance, on the other half of the split.
- **Consumer announcement** — **not named anywhere.** This is genuinely open. The runbook treats it as a blocking prerequisite, not a follow-up, and it cannot be closed from inside the repo: it targets BI consumers reachable only by a human announcement.

Flagging rather than assigning — picking who announces to the analytics/BI audience is not a call I should make. It needs an owner named in this PR's body or a linked tracking issue before merge.

### On the two "could not verify" items

Both now have answers:

- **dbt Cloud CI** — green on `986bbe20e`, with the 15 warnings pulled via `warning_only=true`. All 15 are pre-existing and none belong to this change; the two whose names resemble this fact (`fct_assessment_scores_student_scoped`, `bridge_assessment_expectations_student_scoped`) are different models. Note the check failed twice before that on a `Clone - Staging` race — 259 `zz_stg_kipptaf_*` tables were recreated inside a two-minute window while the build read them. Different tables each time, disjoint error sets, and all present afterwards.
- **Live warehouse** — the row and measure reconciliation is posted above: zero rows lost, 1,476,429 gained and all intended, summary grain bit-identical to `main`.

CI will re-run on `5488afaf8`.

cristinabaldor:
@anthonygwalters — the runbook's last open merge gate is now assigned to you in the PR body.

**Consumer announcement**, before this PR merges. The runbook treats it as a prerequisite rather than a follow-up, and it cannot be closed from inside the repo: the audience is Superset, Tableau-on-the-SQL-API, and direct Cube callers, none of which any dbt exposure covers.

What it needs to say: **anything filtering `response_type IS NULL` (or "not set") to reach vendor or state rows returns nothing once this merges.** The replacement is `response_type = 'overall'`. Worth stressing the timing — this breaks at **this** merge, when the fact changes, not at the Cube merge. That is the opposite of what the original single-PR framing implied, so anyone who read the earlier version has the wrong date.

Your other gate, the claude.ai Project upload, is unchanged and sits on #5248. One wrinkle already noted above: the `response_type` vocabulary flips when **this** PR merges while the knowledge files ship on #5248, so uploading right after this merge beats waiting for that one.

Reassign either by editing the PR body if someone else should carry it.

cristinabaldor (dismissed):

cristinabaldor:
@cbini ready for you, i could not get the comments inline in sql any shorter, submitting a side issue about claude following that guidance.

src/dbt/kipptaf/models/iready/intermediate/int_iready__domain_unpivot.sql
```diff
@@ -4,6 +4,7 @@ with
             _dbt_source_relation,
             student_id,
             `subject`,
+            illuminate_subject,
```
cbini, line 7: pre-existing, but int_iready__diagnostic_results is not the right place to define this. it belongs downstream in blended models

cbini (changes requested):

src/dbt/kipptaf/models/iready/intermediate/int_iready__domain_unpivot.sql
cbini: Confirmed, and it is wider than i-Ready. Six source models each define their own `illuminate_subject`: `int_iready__diagnostic_results`, `int_amplify__all_assessments`, `stg_renlearn__star`, `int_pearson__all_assessments`, `int_fldoe__all_assessments`, `stg_cambium__njgpa`.

Only two models read the column, and both are blended: `int_assessments__score_anchors` takes it as `subject_area`, and this fact joins `int_assessments__resolved_section_enrollments` on it. So the mapping has no source-layer consumer at all — your read is right.

Not fixing it here. It spans four dbt projects, and two of the six are package models kipptaf reads through `source()`, which makes each a two-PR column removal under `.claude/rules/dbt-models.md`. That exceeds the additive-upstream-edits-only rule in `src/dbt/kipptaf/models/marts/CLAUDE.md`. Filed as #5340 with the model list and the two gotchas.

`int_iready__domain_unpivot` keeps selecting the column through for now. Dropping it just from the i-Ready path would move the mapping into the fact for one source and leave the other five where they are, which is worse than the current consistency.

One note on #5340 — I could not add it to project board #4. Both the Codespace token and the OAuth fallback are missing `read:project`.

cbini (approved):

PR #4720, opened 2026-08-04
feat(dbt): add completion flag to rpt_tableau__survey_links

cristinabaldor:
# Pull Request

## Summary and Motivation

When merged, this pull request will let the Personalized Survey Links workbook
report whether a person has completed a survey without a Tableau data source
join. `rpt_tableau__survey_links` now carries a `completion` column derived the
same way `rpt_tableau__survey_completion` derives it — the deduped latest
response per employee, academic year, and survey round.

`rpt_tableau__survey_completion` already computed this flag but does not carry
`link` or `assignment`, which is why the workbook still had to join two views.
Nothing was removed or renamed, so the existing Tableau join keeps working until
the workbook is repointed (owner will do that right after merge).

Untracked surveys read NULL rather than 0. TNTP Insight and Gallup Q12 are
administered outside Google Forms and never produce a trackable response here, so
a 0 would report every person as confirmed-incomplete forever instead of unknown.

Two in-passing cleanups on the same model: the final `select` is now enumerated
(it was `select *`, which the repo SQL conventions prohibit in an `rpt_` model,
and the new join makes a star ambiguous anyway), and the model gained the
uniqueness test every `rpt_` is meant to carry plus model and column
descriptions, which the properties file was missing entirely.

Not a breaking change — additive column only, no rename or removal, so no
downstream exposure coordination is needed.

### Verification

Built into a personal dev schema with upstreams resolved to prod
(`--defer --favor-state`), then compared against prod data.

| Check | Result |
| --- | --- |
| Row count | 44,030 — unchanged grain, no fan-out |
| `completion = 1` / `= 0` | 21,768 / 20,812 |
| NULL flags | 1,450 — exactly the TNTP Insight plus Gallup Q12 rows, no others |
| Agreement with `rpt_tableau__survey_completion` | 42,580 rows compared, 0 disagreeing |
| New uniqueness test | PASS |
| Downstream `rpt_tableau__survey_completion` | rebuilt, all 4 of its tests PASS |
| Trunk | clean on all 3 changed files |

The join cannot fan out by construction — the `dbt_utils.deduplicate` partition
key is exactly the join key, so at most one response row matches per assignment.

## AI Assistance

Claude Code authored the SQL, properties file, and exposure change. Human-directed
throughout: the approach (compute completion on `survey_links` rather than
consolidate the two views), the NULL-for-untracked-surveys decision, and the
exposure update were all chosen by the author. Claude found and reported that
`rpt_tableau__survey_completion` already had the derivation before any code was
written.

### General

### Dagster

Skipped — no Dagster changes.

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — updated
      `rpt_tableau__survey_links.yml` with the new column, a model description,
      descriptions for all 21 columns, and a uniqueness test
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — added
      `rpt_tableau__survey_links` to the `personalized_survey_links` exposure
- [x] **Breaking change?** No — additive column only. `rpt_tableau__survey_completion`
      selects named columns from this model and was rebuilt clean.
- [x] External sources — not applicable, no source changes

### Docs

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

## Post-merge follow-up

Owner will edit the Personalized Survey Links workbook to drop its Tableau data
source join and read `completion` off `rpt_tableau__survey_links`. If that edit
also stops the workbook reading `rpt_tableau__survey_completion` or
`rpt_tableau__survey_responses`, those refs should come off the exposure in a
follow-up.


cristinabaldor:
Verified each finding against the code before acting. Three taken, two not — reasoning below.

## Taken — pushed in `1fab168c5`

**Alias collision.** Confirmed: `sr` is the roster inside `eligible_roster` (line 66) and was the response dedup in the final `SELECT` (line 519). Renamed the dedup alias to `resp` rather than renaming the roster alias — the roster alias has ~30 references inside `eligible_roster`, so renaming the new one is the smaller, lower-risk edit that achieves the same disambiguation.

**`severity: error` on the new uniqueness test.** Correct that the convention permits omitting it but that the sibling sets it explicitly. Added. This model's grain now carries the completion join, so a break should fail rather than warn.

**Cross-reference note.** Added to `rpt_tableau__survey_completion`'s description, explaining why two derivations coexist so the next reader does not read the duplication as an oversight.

## Not taken — consolidating the dedup CTEs

The equivalence claim is correct and the numbers hold. Not taking it, for a reason beyond scope: **it was an explicit design decision, not a deferral.** The two derivations are independent so that neither view's consumers are coupled to the other's filters.

Concretely, `rpt_tableau__survey_completion` filters out TNTP Insight and Gallup Q12; `rpt_tableau__survey_links` must keep them, which is why its `completion` is a nullable superset rather than the same column. Having `survey_completion` select `sl.completion` would make its output depend on a nullability contract owned by a different view — so a future change to how `survey_links` treats untracked surveys would silently alter `survey_completion`, and through it `rpt_extracts__survey_reminder`, which drives outbound reminder emails.

The duplication is real and the note above now documents it.

## Not taken here — view inlining depth

Measured rather than assumed, since the review flagged this as unverifiable without warehouse access.

A query reading `rpt_tableau__survey_links` and `rpt_tableau__survey_completion` joined together — the heaviest realistic shape, and strictly more than Tableau's hourly refresh of `survey_links` alone — came in at 794 MB processed, 6.8 s elapsed, no nesting or query-complexity error. That is roughly half a cent per run, about 3 dollars a month at hourly cadence.

No materialization flip warranted. Worth revisiting only if the workbook's actual refresh shows otherwise.

## Deferred to a human — `config.meta.contains_pii`

Correctly identified as pre-existing on both models rather than something this diff introduced, and correctly left alone.

One detail that strengthens the case for eventually tagging it: `link` embeds the person's race/ethnicity, gender identity, languages, education level, and alumni status directly in the prefilled URL for the Staff Info and Certification Update form. That is indirect-identifier PII in a string column, which is easier to miss than the `mail` and `employee_number` columns named in the finding.

Per the root CLAUDE.md, tagging scope — direct-only versus direct plus indirect — needs a human decision, so it stays out of this PR. Flagged to the author for a follow-up covering both models together.

GabyRangelB (approved):

Issue #4220, opened 2026-06-18
feat: Focus SIS (KIPP Miami) — Phase B residuals: staging/intermediate for tables empty at initial load


cbini:
Tracks the Focus SIS source tables that were configured for extraction but **empty at initial load**, because Focus was not yet fully in production in KIPP Miami. Each yielded zero rows at first extraction, so no `stg_focus__*` model could be contract-enforced against it.

These build incrementally: when Focus brings a domain online, its table starts carrying rows, at which point a `stg_focus__*` staging model (and any intermediate model) can be added.

Split out of #4213 so that issue could close on the 54 populated tables plus `int_focus__student_enrollment`; this one stays open as a long-running tracker.

## Rewritten 2026-08-14 — the framing changed

Two things invalidated the previous body, which described 11 tables as having "no BigQuery table at all":

1. **#3757 is fixed.** Empty source tables now materialize in BigQuery with a schema and zero rows. All 77 configured tables exist in `dagster_kippmiami_dlt_focus`. "No BigQuery table" is no longer a state any table on this list can be in, so absence is no longer the signal — row count is.
2. **Miami's school year started 2026-08-12** and Focus attendance began flowing on day one.

The composition of the list is unchanged: the same 13 tables from the original 22 still lack a staging model. What changed is that **8 of the 13 now carry rows**, up from 2.

## Current state — measured 2026-08-14

64 of the 77 configured tables have a `stg_focus__*` model. These 13 do not.

### Populated and buildable now — 8

| Table | Rows | Notes |
| --- | --- | --- |
| `attendance_period` | 9,010 | 1,579 students, 2026-08-12 to 2026-08-14, growing daily |
| `attendance_day` | 4,538 | 1,579 students, 2026-08-12 to 2026-08-14, growing daily |
| `gradebook_grades` | 130 | |
| `discipline_referrals` | 25 | was 1 row when previously checked |
| `student_groups` | 14 | |
| `students_join_groups` | 5 | thin |
| `students_join_students` | 4 | thin |
| `scheduling_teams` | 1 | thin |

**Attendance is the priority and is no longer a tracker item.** `attendance_day` and `attendance_period` both carry live AY2026-27 data for 1,579 students and grow every school day. Their staging models are the gate on #4865, which is in turn what #4803 needs for its option 3 — modeling Miami attendance from Focus directly, described there as the real fix.

[...]

PR #4725, opened 2026-08-04
feat(dbt): staging and intermediate models for newly-populated Focus tables

### LINKED ISSUE #4724: feat: Focus SIS (KIPP Miami) — staging models for 6 newly-populated tables
Six Focus source tables tracked as empty in #4220 are now populated in `teamster-332318.dagster_kippmiami_dlt_focus` and have no `stg_focus__*` model. Build them.

Population verified 2026-08-04 with `count(*)`; see https://github.com/TEAMSchools/teamster/issues/4220#issuecomment-5184108244 for the full re-check.

## Staging models to add

| Source table | Rows | Model |
| --- | --- | --- |
| `schedule` | 16,771 | `stg_focus__schedule` |
| `student_report_card_grades` | 11,631 | `stg_focus__student_report_card_grades` |
| `students_join_users` | 2,039 | `stg_focus__students_join_users` |
| `test_history_administrations` | 1,133 | `stg_focus__test_history_administrations` |
| `test_history_scores` | 1,103 | `stg_focus__test_history_scores` |
| `co_teachers` | 28 | `stg_focus__co_teachers` |

Each follows the package convention in `src/dbt/focus/CLAUDE.md`: select from `source("focus", ...)`, drop `_dlt_*` bookkeeping and the audit quad, apply `where deleted is null` where the table has the column, `contract: enforced: true`, and a PK `unique` + `not_null` at `severity: error` in `properties/`.

Naming note: `stg_focus__report_card_grades` already exists over the separate `report_card_grades` source table. The new model must be `stg_focus__student_report_card_grades`.

## Intermediate models unblocked

- `int_focus__schedule` (needs `schedule`)
- `int_focus__report_card_grades` (needs `student_report_card_grades`)

Whether these land in the same PR or a follow-up depends on how much grain work they need — decide after the staging models are profiled.

## Explicitly out of scope

- `discipline_referrals` and `scheduling_teams` — 1 row each; not enough to validate a contract against. Stay on #4220.
- The 11 tables that still have no BigQuery table at all. Stay on #4220.

Refs #4220
Refs #3584

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will add a contract-enforced staging model for each of the six Focus source tables that were tracked as empty in #4220 and now carry rows, plus the two intermediate models their population unblocks.

Re-checked population on 2026-08-04 with `count(*)`: 11 of the 22 residual tables now exist, and 3 of those (`people`, `students_join_people`, `students_join_address`) already got staging models via the contacts/address work. `discipline_referrals` and `scheduling_teams` hold 1 row each — enough for dlt to create the table, not enough to validate a contract against — so they stay on #4220. That leaves the 6 here.

### Staging models

| Model | Source rows |
| --- | --- |
| `stg_focus__schedule` | 16,771 |
| `stg_focus__student_report_card_grades` | 11,631 |
| `stg_focus__students_join_users` | 2,039 |
| `stg_focus__test_history_administrations` | 1,133 |
| `stg_focus__test_history_scores` | 1,103 |
| `stg_focus__co_teachers` | 28 |

Three things worth a reviewer's attention:

- **`co_teachers` has no surrogate id.** Its key is the composite `(syear, school_id, course_id, course_period_id, staff_id)`, verified unique, tested with `dbt_utils.unique_combination_of_columns`.
- **The two wide tables project only what Miami populates** — 38 of 184 columns on `schedule`, 28 of 113 on `student_report_card_grades`. The remainder is Florida adult-education, clock-hour and functioning-level reporting that is empty network-wide today. Populated custom slots are aliased to their catalog titles, matching the slugs `stg_focus__course_periods` already uses (`fefp_number`, `basic_skills_exam`, and so on).
- **None of the 6 tables has a `deleted` column**, so no soft-delete filter applies to any of them.

### int_focus__schedule

Resolves section, course and marking-period context. Course period and course are inner-joined (0 orphans, and a schedule row without them is not interpretable); `stg_focus__schedule` carries `relationships` tests on both keys so an orphan fails there rather than silently vanishing here.

Focus stores `0` rather than null for a full-year schedule, paired with `mp = 'FY'`. Left unhandled that reads as 13,399 unresolvable marking-period ids, so the sentinel is normalized with `nullif` before the join.

### int_focus__report_card_grades

**This one departs from the original Focus integration design, deliberately.** That design called for joining `student_report_card_grades` to `report_card_grades` and `report_card_grade_scales`. That is not buildable: every Miami row is imported course history (`course_history = 'Y'`, `syear` 2013 through 2025) whose `report_card_grade_id`, `grade_scale_id`, `course_period_id` and `course_id` are 100% NULL. There are no keys to join on, and the letter grade, GPA points and credits are already denormalized onto the row.

What the model does instead:

- Joins `marking_periods`, the one foreign key that resolves — all 48 distinct values match, 0 orphans. Focus stores it as text on this table alone, so the staging model casts it to an integer once.
- Decodes Course Flag 1 and 2 from the custom-field option crosswalk. Both are resolved with two joins in the model rather than a dedicated `__pivot` model, since they are the only two decodable fields on the table and both read the same option list. `District` and `School` have no option rows at all, and `Gradelevel` decodes to its own code (an identity mapping), so none of those three is decoded.

The model description records this so the next person does not re-derive it. Revisit once Focus posts live report-card grades in Miami.

## AI Assistance

Claude Code authored the models, properties files and this description. Human-directed: the decision to break this out as a sub-issue of #4220, the PR scope (6 staging plus 2 intermediate in one PR), and the call to decode the report-card custom fields rather than ship the thinnest possible join or defer the model. Claude did the profiling that surfaced the null-foreign-key problem and proposed the three options.

### General

### Dagster

Skipped — no Dagster changes. All 6 source tables were already declared in `sources-bigquery.yml` with their `meta.dagster.asset_key`, so lineage needs no edit.

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — all 8 have one, with a description on the model and every column
- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — not applicable, nothing consumes these yet; both intermediate models are documented internal-only, needing an `rpt_` view before any external consumer
- [x] **Breaking change?** No. Every model is new; no existing column is renamed or removed
- [x] If adding or modifying an external source, run `stage_external_sources` — not applicable, these are dlt-loaded BQ-native sources, not external tables

## Validation

Local `dbt build` through the kippmiami project with `--favor-state --defer`: **55 pass, 0 warn, 0 error, 0 skip** — 8 models and 47 data tests.

Row counts match source exactly, so neither intermediate model fans out: `int_focus__schedule` 16,771 and `int_focus__report_card_grades` 11,631. Spot-checked the parts a green build would not catch:

- Course flag labels resolve rather than coming back all-null — 35 and 32 rows labeled, matching the populated source rows exactly, to values like `H - Honors` and `9 - Ninth Grader`
- The FY sentinel nulls on exactly 13,399 rows, and `marking_period_title` populates on the complementary 3,372
- Every schedule row resolves a course title
- All 13 `relationships` FKs were checked against the warehouse before the tests were added — 0 orphans on each

`trunk check --force` over all 16 files: no sqlfluff or yamllint findings; the only output was prettier line-wrapping, which the pre-commit `fmt` hook fixed (re-check reported no issues).

**Note on dbt Cloud CI:** this PR is confined to a source-system package, so `state:modified+` selects zero kipptaf models and the dbt Cloud check goes green trivially rather than as validation. A dbt-only PR also gets no branch deployment. The local build above is the only pre-merge validation; these models are first exercised in prod after merge.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes (trivially; see note above)
- [ ] **Dagster Cloud** — not triggered; `src/dbt/**` is excluded from the `pull_request` paths

Refs #4724
Refs #4220


cbini:
Both findings verified and fixed in e4bfa2564.

## 1. `mp` alias collision — fixed

Renamed the joined marking-periods alias to `mkp` in both intermediate models, with a comment at the join naming why. `git grep` found no existing model that joins `marking_periods` at all, so there was no precedent to preserve either way — these two are the first.

Renamed it in `int_focus__report_card_grades` too, even though that model does not project `mp` and so has no collision. Sibling models joining the same table under different aliases is worse to read than either uniform choice.

## 2. `not_null` on `marking_period_id` — fixed, with one correction to the rationale

The mechanic is correct, and I confirmed it from the macro source rather than by inference — `dbt/include/global_project/macros/generic_test_sql/relationships.sql` line 6 filters `where {{ column_name }} is not null` before the orphan join. A fully-null column passes `relationships` trivially.

The data supports locking the guarantee:

```sql
select count(*) as total_rows,                                    -- 11,631
  countif(marking_period_id is null) as source_null,              -- 0
  countif(trim(marking_period_id) = '') as source_blank,          -- 0
  countif(safe_cast(marking_period_id as int) is null) as uncastable,  -- 0
  count(distinct marking_period_id) as distinct_vals              -- 48
from `teamster-332318.dagster_kippmiami_dlt_focus.student_report_card_grades`
```

Added `not_null` at `severity: error` and noted in the column description why both tests are there, so the next person does not read the pair as redundant.

**One correction.** The finding says this locks a guarantee "that `int_focus__report_card_grades`'s join depends on." The join is a LEFT join, so it does not depend on non-nullity — a null FK would yield null marking-period context on that row, not a dropped row or a broken join. The real value of the test is catching silent source drift, which is a good enough reason on its own. Worth being precise about, because the overstated version implies a correctness risk that is not there.

Also relevant and not mentioned in the finding: the staging model uses `cast`, not `safe_cast`, so a non-numeric marking-period value fails the build loudly rather than coercing to null. A null source value was the only silent path to a null column, and that is exactly what `not_null` now closes.

## Verification

`dbt build` on the three changed models: **13 pass, 0 warn, 0 error**, including the new `not_null_stg_focus__student_report_card_grades_marking_period_id`. `int_focus__report_card_grades` still returns 11,631 rows, confirming the alias rename changed nothing about the output. `trunk check --force` on the three files: no issues.

## On the review's scope note

The point about row counts and orphan claims being un-re-runnable without warehouse access is fair and worth keeping in mind for this PR generally. Two of the static corroborations were the useful kind: confirming the three join keys carry `unique` tests does bound the fan-out question independently of my local run, and checking the decode idiom against the existing `__pivot` models is a real cross-check. No other findings to dispute.

PR #4726, opened 2026-08-04
chore(dbt): disable rpt_tableau__grants_timesheets and its exposure

cristinabaldor:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop an hourly Dagster failure by disabling
`rpt_tableau__grants_timesheets` and the `federal_grants_timesheet_approval`
exposure. Nothing is deleted — both stay in the repo with `enabled: false`.

The Federal Grants Timesheet Approval workbook now reads a live Google Sheet, so
it is no longer a dbt consumer.

### The hourly failure

The exposure declared `cron_schedule: 0 * * * *`, which
`code_locations/kipptaf/tableau/schedules.py` turns into a real full extract
refresh. The workbook has no extract, so Tableau rejected every run:

```text
403180: Bad Request
Full extract refresh operation for the workbook
'c7c821a4-a236-4da8-b41e-52b2fd62cbed' is not allowed.
(0x5CE10192 : Full refresh extract operation for the workbook is not allowed.)
```

Confirmed against the Tableau API: filtering `hasExtracts:eq:false` returns this
workbook among the 122 extract-less workbooks on the site.

Per `src/dbt/kipptaf/CLAUDE.md`, `cron_schedule` belongs on an exposure only when
Dagster owns the refresh. Nothing here was Dagster's to refresh.

**Already stopped in Dagster.** The schedule was RUNNING on 2026-08-04 and
failed 7 consecutive hourly ticks (most recent run `7e4734e7`, each with an auto
retry) before being stopped manually. The failures have ended regardless of when
this merges; this PR keeps the schedule from being regenerated on the next
deploy.

### Why disable rather than delete

Review decision on this PR. Disabling achieves the same operational outcome
while keeping the code available if the certification process moves back into
dbt.

It is also the safer default for the data: these are federal grant compliance
records subject to retention requirements. The underlying Alchemer archive
(`kipptaf_alchemer.base_alchemer__survey_results`) is untouched, and the prod
view `kipptaf_tableau.rpt_tableau__grants_timesheets` lingers until someone
drops it. Dropping it needs DDL and is a separate decision, not in this PR.

### The model was already dead

Alchemer is a disabled integration read from a frozen BigQuery-native archive,
so the model has not received new data in about 26 months:

|                           |                          |
| ------------------------- | ------------------------ |
| Rows                      | 600 (339 people)         |
| Signature range           | 2003-02-02 to 2024-06-07 |
| Latest approver signature | 2024-06-20               |

That is why the certification process moved to a Sheet.

### Verification

`dbt parse --no-partial-parse --target prod` is clean, and in the resulting
manifest:

- `exposure.kipptaf.federal_grants_timesheet_approval` and
  `model.kipptaf.rpt_tableau__grants_timesheets` both land in `disabled`, so
  `schedules.py` (which iterates the manifest exposures collection) generates no
  refresh schedule
- the 3 column tests on the model carry the off flag with empty deps, so they do
  not run
- no active node or exposure depends on the model
- `grants_timesheet_roster_feed` is a separate exposure over
  `int_people__staff_roster` and is unaffected

`trunk check --force` is clean on all 3 files. Net diff against `main` is 3
added lines and 0 deletions.

### Known gap in this PR

`docs/reference/automations.md` still lists
`kipptaf__tableau__federal_grants_timesheet_approval_extract_refresh_schedule`
and was deliberately not touched. Per `docs/CLAUDE.md` that file must never be
hand-edited, and it must not be regenerated in a Codespace because
`gen-automations-doc.py` silently skips code locations that fail to import
there, which would drop whole sections from the catalog. **It needs regenerating
in a full environment where all five locations load** — worth folding into the
next schedule change rather than doing blind.

## AI Assistance

Claude Code authored the change. Human-directed: the author identified that the
workbook moved to a Sheet and that the model is dead, and the reviewer directed
disabling rather than deleting. Claude verified the extract-less workbook, the
absence of other consumers, the manifest disabled-state, and the live hourly
failures in Dagster, and found the `automations.md` regeneration constraint.

### General

- [ ] Review the **Claude Code Review** comment posted on this PR — the check
      reports success; findings not yet read by the author

Skipped — no Dagster code changes. Disabling the exposure does remove a
generated Tableau refresh schedule, which is the intended effect.

- [x] Properties file — retained, with `config: enabled: false` added
- [x] Exposure updated — disabled, since the workbook is no longer a dbt
      consumer
- [x] **Breaking change?** No downstream consumer exists. `dbt parse` is clean
      and the manifest shows no active node or exposure depending on the model.
      Rollback is reverting the 3 added lines.
- [x] External sources — not applicable

- [ ] Regenerate `docs/reference/automations.md` in a full environment — see
      _Known gap_ above

## CI checks

- [x] **Trunk** — passes (`check --force` run locally on all 3 changed files)
- [ ] **dbt Cloud** — passed on the prior commit; pending re-run on `b764834dd`

## Open question for the reviewer

The disabled exposure still carries `cron_schedule: 0 * * * *`, which is inert
while disabled. Since that cron is what failed against an extract-less workbook,
it may be worth dropping the line so re-enabling the exposure later does not
restore the hourly failure. Left in place rather than making an unrequested
change.

## Follow-up, not in this PR

The certification process now lives in a Sheet with no dbt model behind it. Two
options, and this PR does not decide between them: build a Sheets-sourced model
and repoint the workbook so lineage is honest again, or leave it in Sheets and
accept that this dashboard sits outside dbt.

Separately, the workbook carries a `userattribute` calculated field calling
`USERATTRIBUTE('username')`, which returns NULL without a Connected App and has
been failing closed. The author is removing it in Tableau directly.


cristinabaldor:
No findings to act on. Recording the one informational note so the next person to open those plans finds it.

**Task 11 of `2026-07-30-tableau-rls-dbt-contract.md` and the Federal Grants Timesheet section of `2026-07-31-tableau-workbook-remediation.md` are done-and-retired.**

Verified the "already landed" half of that claim independently: the deleted SQL carries the full RLS contract column set those plans describe adding — `location_clean_name`, `campus_name`, the normalized `home_business_unit_name` case expression, `mail`, `user_principal_name`, `sam_account_name`, and both `reports_to_*` columns. So the dbt-side work landed in an earlier PR and no unfinished task is being silently dropped by this deletion.

What does remain from the remediation plan's section on this workbook is the workbook-side edit, and it is being handled outside dbt: the `userattribute` calculated field calling `USERATTRIBUTE('username')` is being deleted from the workbook directly rather than converted to `USERNAME()` plus the standard Permissions block. The block in that plan references dbt columns (`[mail]`, `[location_clean_name]`, `[home_business_unit_name]`) that a live Google Sheet data source does not carry, so applying it as written is no longer possible for this workbook.

Not editing the plan documents in this PR. Per `docs/CLAUDE.md` they are working documents excluded from the MkDocs nav, and a plan describing a superseded design is expected rather than a defect — so this comment is the durable record instead, reachable from the commit that retired the model.

On the unverified data claims: fair to flag. For the record, the extract-less finding came from the Tableau REST API (`hasExtracts:eq:false` returns the workbook; 342 others return under `hasExtracts:eq:true`), and the staleness figures from `kipptaf_tableau.rpt_tableau__grants_timesheets` itself (600 rows, 339 people, signatures 2003-02-02 through 2024-06-07). Both are re-runnable by anyone with warehouse and Tableau access.

cbini (changes requested):
disable these instead of deleting

cristinabaldor:
Switched from delete to disable in b764834dd.

All three files are restored. The model and the exposure each carry `enabled: false`. Net diff against `main` is now 3 added lines, 0 deletions:

```text
src/dbt/kipptaf/models/exposures/tableau.yml                        | 1 +
.../extracts/tableau/properties/rpt_tableau__grants_timesheets.yml  | 2 ++
```

The hourly failure still goes away: a disabled exposure drops out of the parsed manifest exposures collection, and `code_locations/kipptaf/tableau/schedules.py` builds schedules by iterating that collection, so no refresh schedule is generated.

Verified via `dbt parse --no-partial-parse --target prod`:

- `exposure.kipptaf.federal_grants_timesheet_approval` and `model.kipptaf.rpt_tableau__grants_timesheets` both land in manifest `disabled`
- the 3 column tests on the model carry the off flag and empty deps, so they will not run
- no active node or exposure depends on the model
- `grants_timesheet_roster_feed` is a separate exposure over `int_people__staff_roster`, untouched

Trunk clean on all 3 files with `check --force`.

One open question. The disabled exposure still carries `cron_schedule: 0 * * * *`, which is inert while disabled. Since that cron is the thing that failed against an extract-less workbook, do you want the line dropped so re-enabling later does not restore the hourly failure? I left it in place rather than making a change you did not ask for.

Separately, that schedule was RUNNING in Dagster earlier today and failed 7 consecutive hourly ticks with `403180 Full extract refresh operation for the workbook is not allowed`. I stopped it, so the failures have ended regardless of when this merges.

Note the PR description above still describes the delete approach and is now stale.

PR #4728, opened 2026-08-04
feat(dbt,tableau): department-scoped row access for the Survey Dashboard support views

### LINKED ISSUE #4721: feat(dbt,tableau): department-scoped row access for the Survey Dashboard support views
The Survey Dashboard has no department-scoped gate. `Permissions - Support` scopes by entity, region, and school, so any viewer who passes it sees every support department's feedback — including a blanket grant to `KNJ-SG-Tableau All Staff KTAF`.

The rated department exists only as a prefix inside `question_shortname`, across three inconsistent naming schemes, and departments have combined over time so the mapping is not one-to-one.

## Plan

Add `rated_department_code` and `rated_department_name` to the existing `src_google_sheets__google_forms__form_items_extension` sheet, which is already keyed on `abbreviation` = `question_shortname`. Carry them through `int_surveys__survey_responses` into `rpt_tableau__survey_responses`. A new `RLS - Department Gate` field in the workbook then matches group membership against the code, with an all-access head and per-group department lists.

Questions that rate no department get a blank code and fall through to the all-access and cross-department leadership branches only.

## Also in scope

Remediation that does not depend on the sheet:

- `support_open_ended` currently has neither the department filter nor the small-cell suppression the other four support sheets carry
- Scope the blanket KTAF grant to the four regions
- Delete the dead `Permissions - Support (Preview)` field, which contains an individual by-name grant
- Remove `The Syndicate`
- Apply the PR #4656 column renames, still unapplied in this workbook

Design doc to follow under `docs/superpowers/specs/`.

Refs #4638

cristinabaldor:
# Pull Request

## Summary & Motivation

...carry the department each survey question rates through to
`rpt_tableau__survey_responses`, so the Survey Dashboard can scope
support-survey rows to the department being rated.

Today `Permissions - Support` scopes by entity, region, and school and never by
department, so every viewer who passes it sees every department's feedback —
including a blanket grant to `KNJ-SG-Tableau All Staff KTAF`. The rated
department is not a column: it exists only as a prefix inside
`question_shortname`, across three inconsistent naming schemes, and departments
have merged over time, so a string parse cannot express the mapping. It has to
be data.

The Survey Dashboard is the last of the 13 workbooks in the Tableau RLS rollout
still awaiting its edit.

### Redesigned after review

The first version of this PR appended `rated_department_code` and
`rated_department_name` to the existing
`src_google_sheets__google_forms__form_items_extension` sheet, then projected
them to one row per abbreviation with `select distinct` before joining.

Review rejected that, correctly. The department a question rates is a property
of the question abbreviation alone, while that sheet's grain is
`(form_id, item_id)` — 356 rows carrying 309 distinct abbreviations, with 33
abbreviations recurring across up to 4 rows. Storing the mapping there forced
Ops to hand-enter the same value up to four times, made contradictory entries
representable, and needed a `distinct` projection plus a bespoke guard test to
hold the join together.

The mapping now lives on its own sheet at its own grain. What that removed:

- `select distinct` and the `question_departments` grain-projection CTE.
- The `one_department_per_abbreviation` guard test — 24 lines of SQL and 21 of
  justification, which existed only to prove the collapsed dimension was never
  real.
- The `enriched_keyed` CTE and its `except (question_shortname_key)` — `lower()`
  in the join predicate does the same job.

Net across the redesign commit: **146 insertions, 95 deletions.**

### What ships here

- A new `src_google_sheets__google_forms__question_department_crosswalk` source
  and its staging model, keyed on `abbreviation`. 309 rows, one per abbreviation.
- `abbreviation`, `rated_department_code`, and `rated_department_name` joined
  onto `int_surveys__survey_responses` and published on
  `rpt_tableau__survey_responses`.
- The `form_items_extension` source and staging contract reverted to their
  original 6 columns.
- Two singular tests replacing the deleted guard test.
- The design spec and the implementation plan, rewritten to match what shipped.

Sheet sources need no Dagster Python change — the asset comes from the
`asset_key` meta in `sources-external.yml`. 26 of the 88 sheet sources in that
file are already standalone crosswalks or lookups, two of them in the surveys
domain.

### The tests, which are the load-bearing detail

`unique_lowered_abbreviation` (error) asserts one row per **lowered**
abbreviation. It replaces a column-level `unique` rather than joining it: the
join lowers both sides, so `A1` and `a1` would both satisfy `unique` and then
collide, fanning out every response for that question. Uniqueness on the lowered
value subsumes uniqueness on the raw value.

`covers_all_abbreviations` (warn) asserts every non-null `abbreviation` in
`form_items_extension` has a crosswalk row. This is the one failure mode a
separate sheet introduces — a question on a new form arrives with no mapping. It
fails safe, because a null department routes to the most restricted audience, so
it warns rather than blocking a build. Passing at 309 of 309.

Neither model carries `not_null` on `abbreviation`. The named range is
row-unbounded, so BigQuery can surface fully-blank phantom rows below the data —
that is what makes the sibling `form_items_extension` sheet stage roughly 525
null abbreviations out of 881 rows. Every consumer filters
`where abbreviation is not null` instead.

### Also decided

The spec's open question on respondent identity is resolved: no support-view
viewer sees respondent names, including the rated department. That is what lets
the support surveys be described to staff as anonymous, and it depends on
revoking Download Data and Web Edit — Tableau has no column-level security, so
sheet edits alone do not deliver it. Both are workbook tasks in the plan, not
code here.

### What is NOT in this PR

- Every Tableau change: the `RLS - Department Gate` field, the remediation slice
  (`support_open_ended` protection, scoping the blanket KTAF grant, deleting
  `Permissions - Support (Preview)`, removing `The Syndicate`, the #4656
  renames), and the respondent-name removal. Plan Tasks 7 through 10.

### Rollback plan

Additive at the consumer end — two new nullable columns on
`rpt_tableau__survey_responses`, no renames and no removals, so no downstream
exposure breaks. Reverting the commit and re-materializing restores the prior
column set. The crosswalk sheet and its tab can stay in place harmlessly.

## AI Assistance

Claude Code wrote the spec, the plan, and the dbt changes. Human-directed: the
decision to put the mapping in a Google Sheet rather than a dbt seed, the
decision that no support-view viewer sees respondent names, the scope call to
ship the data layer while the workbook edits follow, and the review that sent
the mapping to its own sheet.

Claude found and corrected two real flaws during the work. In the original
design it caught its own fan-out — the spec called for joining on `abbreviation`
directly — and added the projection plus guard test. In the redesign it caught
that the spec's own phantom-row finding made `not_null` on `abbreviation` a
latent false-failure, and filtered every consumer instead.

### General

### dbt

- [x] Properties entries with `description:` for all new columns
- [x] Exposure — `rpt_tableau__survey_responses` already has one; this is an
      additive column change, so no exposure edit is needed
- [x] **Breaking change?** No. Two additive nullable columns downstream; the
      `form_items_extension` contract returns to its original shape
      with **`--target staging`** — done for the new crosswalk source. No
      `ext_full_refresh` needed: the table is new, and the macro only skips
      tables that already exist

### Verification

- `dbt parse --no-partial-parse --target staging` — clean. A stale
  partial-parse manifest surfaces as
  `'model...' depends on 'snapshot...' which is not in the graph!` on the next
  `run-operation`; `--no-partial-parse` clears it.
- `dbt build --target staging` across the crosswalk, both singular tests,
  `int_surveys__survey_responses`, and `rpt_tableau__survey_responses`:
  **PASS=11, WARN=1, ERROR=0.**
- **No fan-out, proven by the model's own guard.**
  `dbt_utils_unique_combination_of_columns` on
  `(survey_id, survey_response_id, survey_question_id, question_shortname, answer)`
  passed — that is the exact test a fan-out breaks.
- Row counts corroborate. Staged `rpt_tableau__survey_responses` held 3,619,668
  rows against prod's 3,619,735. Diffing the response keys found 1 row prod-only
  and **0 rows staging-only** — one survey response submitted after `zz_stg` was
  cloned, times its 67 question rows. A left join cannot drop rows, and nothing
  appears in staging that is absent from prod.
- 115,434 rows carry a department code, across 13 of the 14 codes.
- The sheet was audited directly: 309 rows, 309 distinct abbreviations, 0 nulls,
  66 mapped, all 14 agreed codes in use, none outside the list, and each code
  mapping to exactly one display name. Blank cells stage as `NULL` rather than
  empty string, so `accepted_values` passes on the 243 unmapped rows without an
  entry for them.
- `trunk check --force` on all changed files, docs included — no issues.

### The one build warning is pre-existing

`dbt_utils_expression_is_true_int_surveys__survey_responses_...` reports 480
rows. It reports the same 480 in **prod**, so it is unrelated to this change: 24
Manager Survey submissions falling outside every configured admin window.
Filed as #4974 with the analysis.

Refs #4721, #4638


cristinabaldor:
Verified all four findings against the data. Two were correct, one prescribed a
fix that would fail the build on day one. Details below.

## 1. Guard test didn't cover the column `DISTINCT` keys on — confirmed, fixed

Correct, and it is the real hole. `DISTINCT` keys on all three projected columns,
so two sheet rows agreeing on `rated_department_code` but differing in
`rated_department_name` both survive, and the `LEFT JOIN` fans out every response
row for that question.

Took the first option — one test over the pair, rather than two separate counts:

```sql
format('%T|%T', rated_department_code, rated_department_name)
  as department_mapping
...
count(distinct department_mapping) as distinct_department_mappings
```

`format` rather than `concat` because `concat` returns null when any argument is
null and would silently miscount violations while the columns are still sparsely
populated.

Not the second option (derive the name from a separate code-to-name step) — that
mapping needs its own dedupe and its own guard, so it moves the same problem one
hop downstream and adds a CTE.

Positive control against prod, injecting a per-row display name with a constant
code — precisely the scenario in the finding:

| test                    | violations |
| ----------------------- | ---------- |
| code only (as shipped)  | 0          |
| code and name pair (new)| 33         |

The old test passed the drift silently; the new one fails on all 33 affected
abbreviations. Also re-ran the no-fan-out proof unchanged — 4,121,562 rows both
sides.

The spec, the plan, and the model's inline comment now all describe the guard as
covering the pair, and Task 6 tells Ops to copy the pair down rather than retype
it.

## 2. `not_null` on `abbreviation` — drift is real, the prescribed fix is not

The spec-versus-code drift is real and I fixed it, but in the other direction:
`abbreviation` is nullable by design, so that test fails immediately rather than
guarding anything.

Measured on the prod staging table: **525 of 881 rows have a null
`abbreviation`.** Two legitimate sources —

- Section-header rows carry a `Title` and no abbreviation (e.g. a
  `Real Estate & Facilities` header row).
- The named range is row-unbounded, so the sheet's empty grid lands as ~478 fully
  null phantom rows. This is the documented Google Sheets phantom-row behaviour in
  `src/dbt/CLAUDE.md`.

`not_null` with `severity: error` would fail with 525 results on the first build.
The correct handling is the `where abbreviation is not null` filters already in
both `question_departments` and the guard test.

Fixed by correcting the spec — it now says explicitly that no `not_null` belongs
here and why, with the row counts. The plan's Global Constraints and Task 2 both
already said not to add it; the spec's Testing section was the stale copy, which
is what surfaced as drift.

On the underlying concern — an Ops edit blanking an existing abbreviation
silently drops that question's mapping — that is a real gap, and I did not close
it. Detecting it needs a baseline of which abbreviations should exist, which is
not something the current source carries. Not worth inventing here.

## 3. Pre-existing uniqueness gap — confirmed pre-existing, and the natural key isn't unique

Correct that the model has no uniqueness test and that this predates the PR. It
does not become addable opportunistically, though: `(form_id, item_id)` is not
unique in the data even after excluding phantom rows.

Among the 403 rows with a non-null `form_id`, there are **357 distinct
`(form_id, item_id)` pairs** — 46 duplicate rows — plus one row carrying a
`form_id` and no `item_id`. A `dbt_utils.unique_combination_of_columns` on that
key fails today regardless of how it is scoped.

So this is a source data-quality question, not a missing-test question, and
resolving it means deciding what those duplicate rows are. Left out of scope
rather than shipped broken.

## 4. Nits — both fixed

- Plan Task 3's snippet had the `lower()` expression before the plain columns,
  backwards from ST06. Corrected, with a note on why the order matters, since the
  plan is what a future implementer copies from.
- Tasks 2 through 4 are now checked off, and the plan carries a status line naming
  this PR and the two blockers (the sheet, and re-staging the external).

## CI

The red `dbt Cloud` check is the expected contract failure, confirmed from the run
error: both new columns report `missing in definition` because
`zz_stg_kipptaf_google_sheets.src_google_sheets__google_forms__form_items_extension`
has not been re-staged. It clears once the sheet gains the columns and
`stage_external_sources --target staging` runs. Every other check is green.

cristinabaldor:
## The red dbt Cloud check is unrelated to this PR

`rpt_clever__enrollments__student_id_resolves_to_students_feed` reports **1
orphan out of 57,832 enrollment rows**, in a model this PR does not touch.
Tracked in #4745. Summary for review:

### Why this PR pulls Clever into scope at all

`rpt_clever__students` is a descendant of the sheet staging model this PR edits,
via a chain that is not obvious: the item-name sheet feeds the Staff Information
Survey pivot, which feeds `int_people__staff_roster`, which
`base_powerschool__student_enrollments` refs. So `state:modified+` selects the
Clever students feed but not its sibling `rpt_clever__enrollments`, and CI
compares one freshly-built relation against one deferred copy.

### It is not caused by this change

`int_people__staff_roster` is identical between the PR build and production —
4,753 rows on both sides. The survey chain alters nothing downstream.

### Orphan counts tracked shared-state refreshes, not code

| State | Orphans |
| --- | --- |
| Initial | 55, then 67, then 68 |
| After refreshing `zz_stg_kippnewark` and `zz_stg_kipppaterson` | 1 |
| After also re-leveling `zz_stg_kipptaf` | 1 |

The 68 reconciled exactly against per-district staleness: Newark was 17 students
behind production and Paterson 47, summing to the 64-row gap in
`base_powerschool__student_enrollments`. Both districts were refreshed via
`dbt clone --target staging --full-refresh` and now match production exactly
(19,754 and 1,086).

The residual 1 is not stale state. Immediately before the final push,
`zz_stg_kipptaf_extracts.rpt_clever__enrollments` measured **0 orphans against
both the staging and the production students feed**, and production itself
measured 0. CI still returned 1, because it rebuilds the students side rather
than reading the copy that was verified. That is the structural issue in #4745.

## Verification of what this PR actually ships

- Both new tests pass against the populated sheet — `accepted_values` on the 14
  agreed department codes, and
  `stg_google_sheets__google_forms__form_items_extension__one_department_per_abbreviation`.
- **No fan-out, with codes that actually match.** Staged
  `rpt_tableau__survey_responses` holds 3,614,537 rows against production's
  3,614,537, with 123,628 rows carrying a department. An earlier check ran while
  the sheet was empty and therefore matched nothing; this one exercises the join.
- The sheet and the pinned list agree exactly — 14 codes in each, 67 coded rows,
  none missing a display name.
- Trunk, CodeQL, and all three CodeQL analyses pass.
- 19 test warnings, all pre-existing. The two largest sit on the survey chain and
  production returns identical counts for the same keys: 46,062 for
  `unique_fct_survey_responses_survey_response_key` and 2,559 for
  `unique_fct_survey_submissions_survey_submission_key`.

## Note for the reviewer

`claude-review`'s comment above predates the `accepted_values` test, the three
department merges, and the `development` code — it fires on open, not on push.
Roughly 40 lines of YAML in
`stg_google_sheets__google_forms__form_items_extension.yml` have not been
bot-reviewed.

One unrelated commit rode in on a `main` merge: `734df2bd3` changes
`.vscode/scripts/claude-install-plugins.sh` by one line, adding
`--scope project` to the plugin install. It is not on `origin/main`.

cristinabaldor:
@claude update your review

cristinabaldor:
Verified all four sub-claims of the spec-staleness finding against the file
before acting, and they hold. Fixed in `403c7809f`, plus a fifth the pass didn't
flag and a correction to one line of this review.

### Fixed

1. **The example calc used codes that never existed.**
   `('human_resources', 'cmo_hroperations')` at L193 — the shipped code is
   `human_resources_operations`. Replaced with a `CASE` over the code, one branch
   per department, keyed against the pinned 14.
1. **`accepted_values` no longer reads as blocked.** The Testing section said it
   was blocked on open question 1; it has been live since `ec1ead4ad`.
1. **Group coverage table replaced with a verified inventory.** Checked
   `stg_ldap__group` across all four naming schemes — `KNJ-SG-Tableau`, `TS-SG-`,
   `TS-DL-`, `Group Staff` — with member counts compared to ADP
   `home_department_name`. Ten of the fourteen codes already have a usable group;
   four need creating. The prior table's `leadership_development` →
   `Leadership Development` row was wrong: no such group exists under any scheme.
1. **Open question 1 recorded resolved**, with the three non-obvious merges
   spelled out. Questions 2 and 3 rewritten to what is actually still open.

### Also wrong, not in the findings

**The placement guidance would have broken every all-access group.** L170–176
said to keep `RLS - Department Gate` beside `Permissions - Support` on the filter
shelf, two filters both `TRUE`. That is an AND across all viewers, and the
all-access and cross-department-leadership tiers belong to no department group —
they would have dropped to zero rows. The gate has to be a disjunct *inside*
`Permissions - Support`, as a `Tier 4b` alongside the entity gate, so department
scoping is added for department members rather than subtracted from everyone
else. Rewritten accordingly.

Same class of trap as the stale calc, and worse: the calc fails closed for one
group, this fails closed for everyone who currently has broad access.

Two mechanics also added to the spec, both of which silently produce a
never-matching branch: `ISMEMBEROF()` returns NULL rather than FALSE when the
viewer is not signed in, and group names with non-alphanumeric characters need
HTML URL encoding — `KNJ-SG-Tableau All T&L` must be written
`KNJ-SG-Tableau All T%26L`.

### One correction to this review

The CI section reports dbt Cloud as green per the author's comment. It is not.
dbt Cloud is **failing** on
`rpt_clever__enrollments__student_id_resolves_to_students_feed` — 1 orphan of
57,832 enrollment rows, in a model this PR does not touch, now tracked in #4745.
The comment above says the check is red and explains why; the 3,614,537-row and
123,628-row figures quoted are from the local staging build, not from a green CI
run. Worth being precise about, since "CI green" is the one thing a reviewer
should not take on trust here.

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
```diff
@@ -168,11 +168,35 @@ with
...
+           (form_id, item_id) rows collapse to one row per shortname. Distinct
+           keys on the pair, so drift in either column would fan this out -- the
+           one_department_per_abbreviation singular test guards against both. */
+        select distinct
+            rated_department_code,
+            rated_department_name,
+
+            lower(abbreviation) as question_shortname,
+        from {{ ref("stg_google_sheets__google_forms__form_items_extension") }}
+        where abbreviation is not null
```
cbini, line 184: this should be its own sheet instead of forcing columns into a grain they don't fit

cbini (changes requested):

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
cristinabaldor: Accepted — moved to its own sheet in e7ace1699.

`src_google_sheets__google_forms__question_department_crosswalk`, a new
`Question Department Crosswalk` tab keyed on `abbreviation`, 309 rows. The join
is now a plain left join at grain.

What that let us delete:

- `select distinct` and the `question_departments` grain-projection CTE
- the `one_department_per_abbreviation` guard test, which existed only to prove
  the collapsed dimension was never real
- the `enriched_keyed` CTE and its `except (question_shortname_key)` — `lower()`
  in the join predicate does the same job

The `form_items_extension` source and staging contract are back to their
original 6 columns, and columns G and H are deleted from that tab. Both external
tables only ever declared 6 columns, so the delete needed no re-staging.

Two tests replace the deleted one. `unique_lowered_abbreviation` (error) asserts
uniqueness on `lower(abbreviation)` rather than the raw value: the join lowers
both sides, so `A1` and `a1` would both satisfy a column-level `unique` and then
collide. `covers_all_abbreviations` (warn) catches a question on a new form
arriving with no mapping — the one failure mode a separate sheet introduces. It
fails safe, since a null department routes to the most restricted audience.

Also dropped `not_null` on `abbreviation`: the named range is row-unbounded, so
phantom blank rows would fail it, which is the same reason `form_items_extension`
never had one. Every consumer filters instead.

Net 146 insertions, 95 deletions. Spec, plan, and PR body rewritten to match.

cristinabaldor:
@cbini ready for you again

cbini (approved):

PR #4734, opened 2026-08-05
fix(dagster): load dlt replace tables as parquet so 0-row extracts succeed

### LINKED ISSUE #4733: fix(dagster): zero-row Focus dlt table fails BigQuery load with Schema has no fields
## Problem

`kippmiami/dlt/focus/discipline_referrals` fails every daily run since 2026-08-05. Run `4e789338-8bea-485f-ad2a-bd3960615e35` (retry of `8c737efd-e201-473b-ae81-b2d844733dc0`); the other 76 Focus steps in the same job succeeded.

Bottom of the error chain:

```text
google.api_core.exceptions.BadRequest: 400 Schema has no fields.
Table: discipline_referrals_44973609_8769_4441_9dc0_2c7e6ad89cf0_source
reason: invalidQuery, location: query
```

## Root cause

The Focus source table returned 0 rows, and dlt cannot load a zero-row `replace` table into BigQuery while `autodetect_schema=True`.

1. `sql_database` with `backend="pyarrow"` yields nothing at all for a 0-row table (its arrow path iterates zero cursor partitions), so the table never lands in `tables_with_items`.
2. `dlt/extract/extract.py::_handle_empty_tables` then writes an empty file for it, so the `replace` root still gets truncated. That file comes from the object extractor, so it is `jsonl.gz` with an empty payload — not the parquet every successful run used.
3. `dlt/destinations/impl/bigquery/bigquery.py::_set_user_hints_with_schema_autodetection` sets `autodetect = True` because of `autodetect_schema=True`, so BigQuery infers the schema from that empty file, produces a temp source table with no fields, and rejects the job.
4. dlt treats `invalidQuery` as transient, so it burns 5 identical retries before failing the step.

The behavior is gated on `schema.data_tables(seen_data_only=True)`. That is why the same 0-row extract **succeeded** on 07-31 08:03 UTC (the table had never loaded data, so no empty file was written and normalize dropped the package) and **fails** now — the table loaded its first row at 07-31 22:32 UTC, so it qualifies for the truncate path.

Deterministic, not flaky: it fails on every run until Focus holds at least one referral row.

## Data impact

None. dlt truncates autodetect tables before the load job, so `dagster_kippmiami_dlt_focus.discipline_referrals` is correctly at 0 rows. The failure is a spurious alert plus a degraded asset.

## Blast radius

Any Focus or Illuminate table that loads data once and later goes empty. Both factories pair `autodetect_schema=True` with `write_disposition="replace"`:

- `src/teamster/libraries/dlt/focus/assets.py`
- `src/teamster/libraries/dlt/illuminate/assets.py`

Miami is mid-rollout with several sparse Focus tables, and school-year rollover is when tables get cleared, so recurrence is likely.

## Fix

Pass `loader_file_format="parquet"` to `dlt.run()` in both factories. An empty parquet still carries the schema's columns, so autodetect sees fields and the truncate load succeeds. Parquet is already the format every successful load used, so the only behavior that changes is the zero-row case.

Verified locally with a reproduction that runs extract plus normalize only (no warehouse credentials needed), on dlt 1.29.1:

| loader_file_format | pass-2 job file | fields |
| ------------------ | --------------- | ------ |
| destination default | `discipline_referrals.{id}.0.jsonl.gz`, empty payload | 0 |
| `parquet` | `discipline_referrals.{id}.0.parquet` | 3 |

Also correct the "Empty source tables" section of `src/teamster/libraries/dlt/focus/CLAUDE.md`: a 0-row extract is only harmless until the table has loaded data once.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop Focus and Illuminate dlt assets from failing the day their source table goes to 0 rows.

`kippmiami/dlt/focus/discipline_referrals` has failed every daily run since 2026-08-05 (run `4e789338-8bea-485f-ad2a-bd3960615e35`) with:

```text
google.api_core.exceptions.BadRequest: 400 Schema has no fields.
Table: discipline_referrals_44973609_8769_4441_9dc0_2c7e6ad89cf0_source
```

Root cause: once a `replace` table has loaded data, `Extract._handle_empty_tables` writes an EMPTY file for it on any later run that yields no items, so the table's root still gets truncated. That file comes from the object extractor, so it defaults to `jsonl.gz` — and BigQuery schema autodetection (`autodetect_schema=True`) cannot infer a schema from an empty jsonl file. dlt classes the resulting `invalidQuery` as transient, so it burns 5 identical retries before failing the step.

Passing `loader_file_format="parquet"` fixes it: an empty parquet carries the dlt schema's columns, so autodetect resolves and the truncate load succeeds. The pyarrow backend already loads data files as parquet, so nothing else about the load changes.

Both factories pair `autodetect_schema=True` with `write_disposition="replace"`, so both get the fix.

No data repair needed: dlt truncates autodetect tables before the load job, so `dagster_kippmiami_dlt_focus.discipline_referrals` is already correctly at 0 rows.

### Verification

Two local reproductions on dlt 1.29.1, neither needing Focus DB or warehouse access:

1. Extract plus normalize only, BigQuery destination — confirms which file dlt emits for the 0-row pass:

   | loader_file_format  | pass-2 job file                                        | fields |
   | ------------------- | ------------------------------------------------------ | ------ |
   | destination default | `discipline_referrals.{id}.0.jsonl.gz`, empty payload  | 0      |
   | `parquet`           | `discipline_referrals.{id}.0.parquet`                  | 3      |

1. Full `pipeline.run(write_disposition='replace', loader_file_format='parquet')` against duckdb — the empty-parquet job completes and the destination table ends at 0 rows, so the truncate still happens.

### AI Assistance

Claude Code performed the investigation, wrote both reproductions, and authored the code and docs changes. Human-directed: the choice of fix among three candidates (the alternatives were catching the load error, or dropping `autodetect_schema=True`).

### General

### Dagster

- [ ] `uv run dagster definitions validate` — **not runnable in the Codespace**: both `kippmiami` and `kipptaf` resolve their dlt credential specs eagerly at module load, so validate and `tests/test_dagster_definitions.py` fail on unset credentials regardless of this change (documented in `src/teamster/CLAUDE.md`). Verified instead with `py_compile` on both edited files plus the two dlt reproductions above; the Dagster Cloud branch deployment is the real check.
- [x] No new integration — this changes two existing library factories only.

### Docs

- [x] Corrected the "Empty source tables" section of `libraries/dlt/focus/CLAUDE.md`, which claimed a 0-row extract always succeeds. That holds only until the table has loaded data once.
- [x] Documented the mechanism under _`replace` write-disposition_ in `libraries/dlt/CLAUDE.md`, since it applies to both factories.
- [x] No schedule or sensor change, so no automations-catalog regeneration.

## CI checks

- [ ] **Trunk** — `trunk check --force` on all 4 changed files reports no issues.
- [ ] **Dagster Cloud** — `libraries/dlt/` is shared, so every consuming code location redeploys.

Refs #4733

### DIFF
--- a/src/teamster/libraries/dlt/CLAUDE.md
+++ b/src/teamster/libraries/dlt/CLAUDE.md
@@ -122,6 +122,22 @@ changes on existing tables.
   narrowed source to
   `DagsterDltResource.run(dlt_source=source.with_resources(*subset))` (its
   `is_subset` filter intersects with `selected_asset_keys`).
+- **`autodetect_schema=True` + `replace` requires
+  `loader_file_format="parquet"`** (passed to `dlt.run()` in the focus and
+  illuminate factories). `Extract._handle_empty_tables` writes an EMPTY file for
+  a `replace` table that received no items this run so its root still gets
+  truncated — but only once the table has **seen data** before
+  (`schema.data_tables(seen_data_only=True)`, restored from the destination). It
+  comes from the object extractor, so the default is empty `jsonl.gz`, and
+  BigQuery schema autodetection rejects that with
+  `400 Schema has no fields. Table: {table}_{uuid}_source` (dlt calls
+  `invalidQuery` transient, so it burns 5 retries first). An empty parquet
+  carries the dlt schema's columns, so autodetect resolves and the truncate load
+  succeeds. Forcing parquet changes nothing else — the pyarrow backend already
+  loads data files as parquet. Symptom to recognize: a table that loaded fine
+  for days starts failing the day its source goes to 0 rows; the destination
+  table is already correctly truncated (dlt truncates autodetect tables before
+  the load job), so there is no data to repair.
 - `replace` never changes an existing table's column **mode** (not just type):
   an all-`NULLABLE` load into a table whose column is `REQUIRED` fails with BQ
   400 "changed mode from REQUIRED to NULLABLE". Drop the table so the pipeline
--- a/src/teamster/libraries/dlt/focus/CLAUDE.md
+++ b/src/teamster/libraries/dlt/focus/CLAUDE.md
@@ -61,11 +61,22 @@ columns.
 
 ## Empty source tables
 
-The PyArrow backend writes NO BigQuery table for a 0-row source extract (the
-asset still SUCCEEDS — no `rows_loaded`, `jobs: []`). A configured Focus table
-absent from `dagster_<district>_dlt_focus` is therefore **empty in the source**
-(Focus is mid-rollout in Miami), not an extraction failure — confirm via the
-asset materialization metadata before investigating.
+A 0-row source extract behaves differently before and after the table's first
+successful load, because dlt's empty-table handling keys off
+`seen_data_only=True`:
+
+- **Never loaded data**: the PyArrow backend writes NO BigQuery table and no
+  load jobs (the asset SUCCEEDS — no `rows_loaded`, `jobs: []`). A configured
+  Focus table absent from `dagster_<district>_dlt_focus` is therefore **empty in
+  the source** (Focus is mid-rollout in Miami), not an extraction failure —
+  confirm via the asset materialization metadata before investigating.
+- **Has loaded data before**: dlt emits an empty file so the `replace` root is
+  truncated. That file must be parquet or BigQuery autodetection fails the load
+  — hence `loader_file_format="parquet"` in `assets.py`. See _`replace`
+  write-disposition_ in `../CLAUDE.md` (#4733).
+
+So a Focus asset going quiet is not evidence the table is new, and a table
+emptying out is not a data loss — the truncate is the intended outcome.
 
 ## Testing Constraints
 
--- a/src/teamster/libraries/dlt/focus/assets.py
+++ b/src/teamster/libraries/dlt/focus/assets.py
@@ -102,10 +102,9 @@ def build_focus_dlt_assets(
         op_tags=op_tags,
     )
     def _assets(context: AssetExecutionContext, dlt: DagsterDltResource) -> Iterator:
-        # loader_file_format="parquet": once a table has loaded data, a later 0-row
-        # extract still gets an empty file so `replace` truncates it. BigQuery
-        # schema autodetection rejects an empty jsonl file ("Schema has no
-        # fields"); an empty parquet carries the schema's columns, so it loads.
+        # loader_file_format="parquet": BigQuery schema autodetection rejects the
+        # empty jsonl file dlt writes to truncate a `replace` table whose source
+        # went to 0 rows. See `replace` write-disposition in ../CLAUDE.md (#4733).
         yield from dlt.run(
             context=context, write_disposition="replace", loader_file_format="parquet"
         )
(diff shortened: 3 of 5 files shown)


cbini:
## Review follow-ups

Verified each item against the installed packages, prod metadata, and the warehouse. Two code changes (#4, #5) in `ebfa372`; #1-#3 were already-verified facts the review job had no access to check.

### 1. `DagsterDltResource.run()` kwarg forwarding — verified, no change

Two links, both checked in the installed packages before the fix was written:

- `dagster_dlt/resource.py:298` — `load_info = dlt_pipeline.run(dlt_source, **kwargs)`. The wrapper's signature is `(self, context, dlt_source=None, dlt_pipeline=None, dagster_dlt_translator=None, **kwargs)`, so it is signature-agnostic and forwards.
- `inspect.signature(Pipeline.run)` declares `loader_file_format: Literal['jsonl', 'typed-jsonl', 'insert_values', 'parquet', 'csv', 'reference', 'model']`.

Both links are now asserted by `test_dlt_run_accepts_loader_file_format`, so a dagster-dlt upgrade that stops forwarding fails CI instead of silently making the fix inert.

`tests/assets/test_assets_dlt.py::test_dlt_illuminate_codes` is not runnable from a Codespace — Focus and Illuminate are both IP-allowlisted, which is why the branch deployment is the live check.

### 2. 'already parquet' — now shown, and the claim holds

Extended the reproduction to print the emitted file for every table in the package, both passes:

| table | destination default | `parquet` |
| ------------------------------- | ---------------------- | ---------------------- |
| `discipline_referrals` (2 rows) | `.0.parquet` | `.0.parquet` |
| `discipline_referrals` (0 rows) | `.0.jsonl.gz` | `.0.parquet` |
| `_dlt_pipeline_state` | `.jsonl.gz` | `.jsonl.gz` |

Prod corroborates the non-empty case independently: the 08-01 through 08-04 materialization metadata for this asset records `'file_format': 'parquet'` with `discipline_referrals.{id}.0.parquet`, under the destination default.

While checking this I chased the case you did not raise but which would have made 'changes nothing else' wrong — dlt's own `_dlt_*` tables. They are unaffected on two counts: `_dlt_pipeline_state` stays `jsonl.gz` in both runs above, and `dlt/destinations/impl/bigquery/bigquery.py:391` applies the autodetect hint only when `table_name not in self.schema.dlt_table_names()`, so system tables keep dlt-managed schemas. Their prod DDL confirms it — `load_id STRING(64) NOT NULL` is a shape autodetect never produces.

So the claim stands as written: the empty truncate file is the only file whose format changes.

### 3. Row count — confirmed

`select count(*)` on `dagster_kippmiami_dlt_focus.discipline_referrals` returns 0. dlt truncates autodetect tables in `initialize_storage` before the load job, so the truncate landed and the failing job was the redundant empty-file load. No ops action.

### 4. Regression test — added

`tests/libraries/test_dlt_replace_loader_file_format.py`. It invokes each factory's asset body with a recording stand-in for `DagsterDltResource` and asserts the run kwargs, following the offline factory-wiring pattern already in `test_dlt_focus_type_adapter.py`. No database, no Dagster instance. 12 passed together with the existing focus tests.

You were right that `explicit_args` cannot reach this — the kwarg is passed at `dlt.run()` call time, so the test drives `assets.op.compute_fn.decorated_fn` instead.

### 5. Comment duplication — trimmed

Both call sites now carry a 3-line pointer to the `replace` write-disposition section of `libraries/dlt/CLAUDE.md` rather than restating the mechanism.

### CI

All checks were green on `e0c249f2` before this push, including all 5 `dagster-cloud-deploy / deploy` runs. Worth noting for anyone reading the checks: `gh pr checks` reports one row per check name and showed `deploy: pass` while 4 of the 5 instances were still in progress — the per-instance `check-runs` API is the accurate surface.

PR #4738, opened 2026-08-05
refactor(extracts): move ParentSquare delivery to the kippnewark code location

### LINKED ISSUE #4735: refactor(extracts): move ParentSquare delivery from kipptaf to kippnewark
Follow-up to #4668 (Refs #4480). The ParentSquare feed is Newark-only, so the Newark Dagster instance should own delivery.

## Shape

The model logic cannot move. `rpt_parentsquare__staff` joins `stg_ldap__group`, `stg_ldap__user_person` and `int_people__staff_roster`; `rpt_parentsquare__sections` picks its section owner from the staff feed; `int_students__contacts` and `int_extracts__student_enrollments` are kipptaf-level cross-region models. So only delivery moves, following the `extracts/powerschool/` autocomm precedent:

- kipptaf keeps the 7 `rpt_parentsquare__*` views, their descriptions, and all 33 tests.
- The Newark region filter moves out of the kipptaf views and into the district wrappers. The views widen to the three PowerSchool NJ regions and exclude Miami, matching how `rpt_clever__*` scopes itself. This makes a future Camden or Paterson feed a copy of the wrappers plus config, with no kipptaf SQL change.
- kippnewark adds 7 contract-columns-only wrappers over `kipptaf_extracts`, plus the SFTP resource, job, and schedule.

## Work

- [ ] 7 wrappers in `src/dbt/kippnewark/models/extracts/parentsquare/` with contract-columns-only properties, no tests or descriptions
- [ ] 7 table entries under the existing `kipptaf_extracts` source in `src/dbt/kippnewark/models/extracts/sources.yml`, mapped to the kipptaf asset keys
- [ ] `parentsquare` exposure in `src/dbt/kippnewark/models/exposures/`
- [ ] Widen the 7 kipptaf views to NJ and expose the region column the wrappers filter on
- [ ] kippnewark `extracts/config/parentsquare.yaml`, assets, job, schedule (cron `0 18 * * *`, ships stopped)
- [ ] `SSH_RESOURCE_PARENTSQUARE` in kippnewark `resources.py` and `definitions.py`
- [ ] 3 ParentSquare SFTP variable mappings at 4 insertion points in kippnewark `dagster-cloud.yaml`
- [ ] Remove the kipptaf config, assets, job, schedule, SSH resource, and variable mappings so the feed does not double-deliver
- [ ] Regenerate `docs/reference/automations.md`
- [ ] Update the Delivery section of `docs/superpowers/specs/2026-07-29-parentsquare-newark-extracts-design.md`

## Notes

No 1Password or Kubernetes work. The `op-parentsquare-sftp` secret is already applied and shared across code locations, so moving the mappings in a single PR carries no ordering hazard.

The schedule still ships stopped. No SFTP round-trip has been attempted and the upload path is unconfirmed.

Widening the kipptaf views changes the population that the 33 error-severity tests run over. The synthetic per-grade `section_id` derivation is the specific thing to check for cross-region collisions.

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will move ParentSquare outbound delivery from the `kipptaf` code location to `kippnewark`. The feed is Newark-only, so the Newark Dagster instance owns delivery. Follow-up to #4668.

Refs #4735
Refs #4480

This follows the `extracts/powerschool/` autocomm precedent rather than the Clever one #4668 originally copied: the heavy `rpt_` views stay in kipptaf, each district adds a thin contract-columns-only wrapper over `kipptaf_extracts`, and the district owns the SFTP assets, job, and schedule.

### What moved, and what did not

| Layer | Before | After |
| --- | --- | --- |
| 7 `rpt_parentsquare__*` views, descriptions, tests | kipptaf | kipptaf (unchanged home) |
| Region filter | inside each kipptaf view | each district's thin wrapper |
| Extract config, assets, job, schedule | kipptaf | kippnewark |
| `SSH_RESOURCE_PARENTSQUARE` plus 3 variable mappings | kipptaf | kippnewark |
| Asset keys | `kipptaf/extracts/parentsquare/*` | `kippnewark/extracts/parentsquare/*` |

The model logic could not move. `rpt_parentsquare__staff` joins `stg_ldap__group`, `stg_ldap__user_person` and `int_people__staff_roster`; `rpt_parentsquare__sections` reads the staff feed; `int_students__contacts` and `int_extracts__student_enrollments` are kipptaf-level cross-region models.

Pushing the region filter down widened the views from Newark to the three PowerSchool NJ regions (Miami excluded, since it rosters from Focus rather than PowerSchool, the same carve-out `rpt_clever__*` makes) and exposed a `code_location` column each wrapper filters on, named to match `rpt_powerschool__autocomm_students`. A future Camden or Paterson feed is then a copy of the wrappers plus config, with no kipptaf SQL change.

### Widening needed two real changes, not just a filter swap

- **The staff school fan-out is now region-keyed.** A bare `cross join` would put every Newark leader at every NJ school. The join mirrors the 'all region' assignment branch of `rpt_clever__staff`.
- **The section owner is picked per region.** `min(staff_id)` over the whole staff feed would assign one leader network-wide, dangling in every other region's file, while the single-column `relationships` test on `staff_id` still passed, because it only proves the id exists somewhere in the feed. `section_owner` now groups by `code_location`; the join to it is a LEFT join so an emptied Ops group still trips the `not_null` guard instead of silently dropping that region's sections; and a new singular test `rpt_parentsquare__sections__staff_id_resolves_at_its_own_school` asserts the (`school_id`, `staff_id`) pair resolves. The Clever feeds shipped this same class of cross-region leak before.

### Validation

Widening changes the population that 33 error-severity tests run over, so it was checked against prod data before implementing:

- no school-number collisions across the three NJ regions
- `student_number` unique across them, 7894 of 7894
- every NJ grade level inside 0 to 12, so the `-4` to `12` `accepted_values` test still holds
- every NJ region carrying at least one Ops leader with no missing `google_email` (Newark 8, Camden 2, Paterson 1)

Then:

- `dbt build` of the 7 views, target dev with defer to prod: **48 pass, 0 warn, 0 error**
- **Newark row counts identical to the pre-change prod views, file for file**, measured in the same moment: schools 12, students 6790, parents 9287, emergency_contacts 15462, staff 96, sections 53, rosters 6790. NJ-wide totals are 19 / 9821 / 12982 / 20593 / 108 / 86 / 9821.
- `trunk check --force` over all 47 changed files: no issues
- `dagster definitions validate` passes for kippnewark; asset keys confirmed as `kippnewark/extracts/parentsquare/*`, kipptaf down to 42 extract assets and 9 schedules with no ParentSquare
- both `dagster-cloud.yaml` files re-parsed after editing; 3 ParentSquare mappings in each kippnewark block

The kippnewark wrappers cannot be built locally: their source carries no target-conditional schema (same as the autocomm wrappers), so they read prod `kipptaf_extracts`, which will not have `code_location` until the widened kipptaf views deploy. They compile clean, and post-merge ordering comes from the `sources.yml` `asset_key` mapping, which makes each kipptaf view an upstream of its wrapper in the Dagster graph.

### Not done here

`docs/reference/automations.md` is **not** regenerated. `kipptaf.definitions` cannot import in a codespace (dlt credentials) and the generator silently drops locations that fail, so running it here would delete kipptaf's whole section. It already omitted ParentSquare before this change, so this PR shrinks the existing gap rather than adding one. It needs one regeneration run in a full environment to pick up the new `kippnewark__extracts__parentsquare__asset_job_schedule` row.

The schedule still ships stopped. No SFTP round-trip has been attempted and the upload path is still unconfirmed.

### AI Assistance

AI-assisted throughout, human-directed on the two decisions that shaped it: choosing to move delivery to kippnewark (against an initial recommendation to keep it in kipptaf on the Clever precedent), and choosing to push the region filter down into the wrappers rather than leave it in kipptaf. Camden and Paterson ParentSquare remains unconfirmed with Ops, which is why the wrappers-plus-config expansion path was preferred.

### General

### Dagster

- [x] Run `uv run pytest tests/test_dagster_definitions.py` for any modified code location. 4 passed, 2 failed: `kippmiami` on the documented `FOCUS_DB` dlt credential and `test_definitions_all` on the same, both pre-existing in a codespace and unrelated to this change. `kipptaf` and `kippnewark`, the two locations touched here, both pass.
- [x] New integrations follow the Library plus Config pattern with the correct asset key format and IO manager

### dbt

- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application. Added `src/dbt/kippnewark/models/exposures/parentsquare.yml`, mirroring `powerschool.yml`. This reverses the 'no exposure' call in the design spec, which surveyed only the kipptaf-delivered feeds and missed `powerschool_autocomm`, the regionally-delivered one this now resembles.
- [x] **Breaking change?** Yes, but contained. `code_location` is an additive contract column on the 7 kipptaf views, and their row sets widen from Newark to NJ. The only consumers are the new kippnewark wrappers, which filter back to Newark, so Newark output is unchanged (verified above). Rollback is a revert; nothing outside this repo reads these views yet, since the schedule has never run.
- [x] If adding or modifying an external source, run `stage_external_sources` with `--target staging`. Not applicable, no external sources touched.

### Docs

- [ ] If adding or changing a schedule or sensor, regenerate the automations catalog: `uv run scripts/gen-automations-doc.py`. **Deliberately not done here**, see Not done here above.
- [x] If adding a new integration to a code location, update that location's `CLAUDE.md`. Updated `src/dbt/kipptaf/CLAUDE.md` with the `code_location` wrapper-filter convention and the widening pitfalls.

## CI checks

- [ ] **Trunk** passes
- [ ] **dbt Cloud** passes
- [ ] **Dagster Cloud** passes or not triggered

### DIFF
--- a/docs/superpowers/specs/2026-07-29-parentsquare-newark-extracts-design.md
+++ b/docs/superpowers/specs/2026-07-29-parentsquare-newark-extracts-design.md
@@ -275,6 +275,12 @@ All seven files sent. `Yes*` = one of email/phone required.
 
 ## Delivery
 
+- **Owned by the `kippnewark` code location**, not `kipptaf` — see _Delivery
+  moved to the kippnewark code location_ under _Implementation notes_. The seven
+  `rpt_parentsquare__*` views stay in `kipptaf` and now cover all three
+  PowerSchool NJ regions; `kippnewark` holds thin wrappers, the SFTP resource,
+  the job, and the schedule. Every reference to `kipptaf` owning the config,
+  resource, job, or schedule earlier in this document is superseded.
 - Server `sftp3.parentsquare.com`, **district-level single connection** (one set
   of files covering all Newark schools). Password or SSH key — created in the
   ParentSquare admin UI, which issues the username.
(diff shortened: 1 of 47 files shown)


cbini:
Finding 1 (missing `contains_pii` on the kippnewark wrappers): valid, fixed in 965b8cabd.

Verified the rule before applying it. `src/dbt/kipptaf/CLAUDE.md` states it under 'Exposing a package/district model as a kipptaf source', which is the opposite direction from this PR (district wrapper over a kipptaf source, not kipptaf wrapper over a package model) — but the rationale, that `config.meta.contains_pii` does not travel through `source()`, is direction-agnostic and holds here. It also does not conflict with the contract-columns-only wrapper convention in `src/dbt/CLAUDE.md`, which bars data tests and descriptions, not `config.meta`.

Scope was set by mirroring the kipptaf side rather than by scanning column names, and it lands on the same four models you named: `staff`, `students`, `parents`, `emergency_contacts` are the kipptaf views that tag `contains_pii`. `schools`, `sections`, and `rosters` tag nothing on the kipptaf side and stay untagged here — `schools` in particular carries `principal_email` and the principal's name, which kipptaf deliberately does not tag (institutional contact info), and `sections` / `rosters` carry `student_id`, which the repo's own note says the staging precedent omits. Expanding past kipptaf's tagging would have been a scope change, not a fix.

Confirmed in the parsed manifest: the four wrappers resolve `config.meta.contains_pii = True`, the other three `False`.

On the pre-existing identical gap in `rpt_powerschool__autocomm_students.yml` / `_teachers.yml` that you flagged as mitigating — agreed it predates this PR, and I have left it alone rather than widen this PR's scope. Worth its own issue if Ops governance wants the autocomm feeds tagged too.

Two notes on the rest of the review, neither needing a change:

- The '4 insertion points' gotcha is written for kipptaf's `dagster-cloud.yaml`, which splits each secret across a credentials group and a host/port group. kippnewark's file does not split that way, so the correct count there is 2 insertion points carrying all 3 variables each — which is what landed, and what you verified.
- Good call flagging the environment limits up front (no warehouse access, shallow clone defeating `git diff origin/main...HEAD`) and switching to per-file blob diffs. The data-dependent claims you took as reported are the ones in the PR description under Validation, all of which came from a local `dbt build` and prod queries.

PR #4746, opened 2026-08-06
feat(dbt): surface the walked school's canonical name on the ops EKG extract

cristinabaldor:
# Pull Request

## Summary and Motivation

...add `school_clean_name` to `rpt_tableau__operations_ekg`, so a consumer can
scope a row to the school that was **walked** rather than the school the
walkthrough's author works at.

The model already joins `int_people__location_crosswalk` on the walked school to
get `grade_band`. It just never selected the clean name from that join. This is
one additive line in the `final` CTE.

### Two columns on this model describe a school, and they are not the same school

- `location_clean_name` comes from the roster join on `respondent_email`, so it
  is where the walkthrough's **author** works.
- The school **walked** is the pivoted `school` column, which is raw form-dropdown
  text.

`school` cannot be grouped or gated on. 5 of its 23 values are retired or
abbreviated aliases, and two schools arrive under two labels each:

| Raw `school` label | rows | resolves to |
| --- | --- | --- |
| `KIPP Hatch Middle` | 146 | `KIPP Hatch Middle` |
| `KIPP Hatch Academy` | 9 | `KIPP Hatch Middle` |
| `KIPP Sumner Elementary` | 51 | `KIPP Sumner Elementary` |
| `KIPP Sumner Academy` | 5 | `KIPP Sumner Elementary` |
| `KIPP Cooper Norcross High School` | 103 | `KIPP Cooper Norcross High` |

So anything grouping on `school` splits Hatch and Sumner in half.
`school_clean_name` collapses all 23 labels to 21 canonical schools.

### Why now

This unblocks scoping the Operations Systems workbook's `operations_ekg` gate to
a school leader's own school. That gate is currently a flat list of group
memberships with no location check, because the workbook's embedded extract
predates the access contract and carries none of the columns a gate would need.
An audit of all 11 gated workbooks on 2026-08-05 found it; the remediation
sequence is in
`docs/superpowers/plans/2026-07-31-tableau-workbook-remediation.md`.

### Also in here

Corrected `location_clean_name`'s description, which claimed it "Drives the
Tableau location gate." It should not, and that is the mistake this PR exists to
prevent: a gate pointed at it would show school leaders the walkthroughs they
**performed**, not the ones **of their school**.

### Not a breaking change

Additive: one new nullable column, no renames, no removals. Reverting the commit
and rematerializing restores the prior column set.

## AI Assistance

Claude Code found the problem and wrote the change. Human-directed: the decision
that the broad Operations Systems groups stay cross-regional while school leaders
and DSOs get school-scoped, and the crosswalk sheet edit described below.

Claude's own first read of this was wrong in a way worth recording: it initially
proposed gating on `location_clean_name`, which would have scoped school leaders
to walkthroughs they authored rather than walkthroughs of their school. Reading
the model's roster join is what caught it.

### General

- [x] No Asana due-date change needed — this is a data-layer prerequisite, not a
      dated request

### dbt

- [x] Properties file updated — `description` and `data_type` for the new column,
      plus descriptions added to `school` and `grade_band`, which had none
- [x] Exposure — `rpt_tableau__operations_ekg` already has one; an additive
      column needs no exposure edit
- [x] **Breaking change?** No. One additive nullable column
- [x] No external source added or modified

### Verification done locally

- `dbt build --select rpt_tableau__operations_ekg --target dev --defer --favor-state`
  — contract enforced, view created, all 4 data tests pass.
- Against the dev view: `school_clean_name` non-null on all 3,577 rows, 23 raw
  labels collapsing to 21 canonical schools, both Hatch labels resolving to
  `KIPP Hatch Middle` and both Sumner labels to `KIPP Sumner Elementary`.

### A pre-existing defect this surfaced

`grade_band` on this model was null for **602 rows across 25 walkthroughs**. The
form labels `KIPP Paterson Prep ES` and `KIPP Paterson Prep MS` were absent from
the location crosswalk, so the model's existing `grade_band` join missed them.
The crosswalk sheet now carries both aliases, which cleared it — independent of
this change, and verified at 0 nulls afterwards. Worth knowing because the same
gap would have made `school_clean_name` null for those rows.

Refs #4638

### DIFF
--- a/docs/superpowers/plans/2026-07-31-tableau-workbook-remediation.md
+++ b/docs/superpowers/plans/2026-07-31-tableau-workbook-remediation.md
@@ -316,6 +316,13 @@ design doc's earlier estimate.
 - **Fields on `operations_ekg`:** `[region]` → `[home_business_unit_name]`,
   `[respondent_location]` → `[location_clean_name]`, `[respondent_job_title]` →
   `[job_title]`
+- **Location-gate variant on `operations_ekg`:** the Tier 5 location gate reads
+  `[school_clean_name]`, not `[location_clean_name]`. Those are two different
+  schools on this extract — `location_clean_name` comes from the roster join on
+  the respondent, so gating on it shows a school leader the walkthroughs they
+  **performed**, not the ones **of their school**. `school_clean_name` is the
+  walked school, added in PR #4746; it must be merged and materialized before
+  this workbook's Tier 5 edit.
 - **Role variant:** omit the AP branch entirely
 - **Note:** `operations_pm` keeps `[respondent_job_title]` and
   `[respondent_name]` — that model has two people per row and the respondent is
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__operations_ekg.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__operations_ekg.yml
@@ -39,6 +39,19 @@ models:
                   - KIPP Miami
                   - KIPP Paterson
                   - KIPP TEAM and Family Schools Inc.
+      - name: school_clean_name
+        data_type: string
+        description: >-
+          The walked school resolved to its canonical name via
+          int_people__location_crosswalk, so every alias of one school collapses
+          to a single value. This is the column a Tableau location gate scopes
+          on. Null when the form label is absent from the crosswalk, which is
+          also when grade_band is null.
+        data_tests:
+          - not_null:
+              config:
+                severity: warn
+                where: school is not null
       - name: respondent
         data_type: string
       - name: google_email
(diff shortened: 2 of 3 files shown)


cristinabaldor:
Addressed in 3f99fbb7d.

### Important #1 — stale runbook: confirmed, fixed

Verified `docs/superpowers/plans/2026-07-31-tableau-workbook-remediation.md:316-318` and the canonical Tier 5 location gate at line 139.

One refinement: the field-mapping line itself is correct *as a rename* — `respondent_location` really did become `location_clean_name` in #4656. The defect is the missing **Tier 5 location-gate variant** for this workbook. With no variant named, an executor pastes the canonical gate reading `[location_clean_name]` and nothing tells them otherwise. Added the variant bullet rather than editing the rename line.

### Minor #1 — drift guard: added, as a warn `not_null` rather than a crosswalk `relationships` test

`not_null` on `school_clean_name`, `config: severity: warn`, `where: school is not null`.

It measures the effect that matters (gate column null for a row that named a school) and also catches a case a `relationships` test on `school` would miss: `int_people__location_crosswalk` left-joins `stg_google_sheets__people__locations`, so a form label can match `location_name` and still yield a null `location_clean_name`. Moved the column to the top of the `columns:` list per the tested-columns-first convention in `src/dbt/CLAUDE.md`.

Local: `dbt build --select rpt_tableau__operations_ekg --target dev --defer --favor-state` — PASS=6, the new test included. `trunk check --force` clean on both files.

### Minor #2 — PII tag: not doing here

2 of the 114 properties files in `src/dbt/kipptaf/models/extracts/tableau/properties/` carry `config.meta.contains_pii`. The omission is the directory norm, not a gap specific to this model, and `school_clean_name` is a school name on a model that already exposes staff email addresses. Tagging that directory is its own change.

anthonygwalters (approved):

PR #4752, opened 2026-08-06
fix(extracts): send real relationship labels and emergency contacts to the NJ contact feeds

### LINKED ISSUE #4751: fix(extracts): restore contact relationship labels to the NJ DeansList and gsheet contact feeds
Zendesk 474545 (Newark DSOs) reports NJ families showing no parent contacts in DeansList, and a student's mother appearing as `Emergency 1` rather than by her relationship. Two defects in the extract layer, both ours.

## 1. `rpt_deanslist__family_contacts` discards the real relationship label

The `case` on `contact_slot` replaces `Relationship` with the literal strings `Emergency 1`..`Emergency 4` for every emergency-slot row. Finalsite supplies a real label on essentially all of them, so DeansList shows `Emergency 1` where it should show `Mother`.

Newark rows in `int_finalsite__student_contacts`:

| Slot | Rows | Carry a relationship label | Distinct labels |
| --- | --- | --- | --- |
| contact_1 | 10,747 | 10,747 | 7 |
| contact_2 | 3,485 | 3,485 | 4 |
| emergency_1 | 6,970 | 6,969 | 13 |
| emergency_2 | 6,798 | 6,796 | 13 |
| emergency_3 | 1,790 | 1,790 | 13 |
| emergency_4 | 671 | 671 | 13 |

Fix: pass `relationship` through unchanged. `ContactType` already carries `Emergency` for DeansList message routing, so the slot ordinal adds nothing.

## 2. `rpt_gsheets__student_contact_info` release fields are null by construction

`release_1`..`release_3` are built from `pickup_1`..`pickup_3`, and `release_4`/`release_5` are hardcoded null. No model emits a `pickup_*` contact slot after the Finalsite cutover — `int_finalsite__student_contacts` produces only `contact_1`, `contact_2`, and `emergency_1`..`emergency_4`, and the Focus branch produces the same set. So every release column is empty for every NJ student.

Fix: build `release_1`..`release_4` from `emergency_1`..`emergency_4`.

## Out of scope

512 active NJ students (Newark 292, Camden 154, Paterson 66) have no Parent 1 row because Finalsite has no `primary` relationship flag for them. Of Newark's 298, 265 have a `financial` relationship with no `primary`. That is a source data-entry gap plus a possible rule change, being worked with SRE and the Finalsite rep — not part of this change.

Also stale and worth correcting while in the file: the header comment on `int_students__contacts_pivot` claims `contact_2` has no source rows; Newark has 3,485.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make the NJ DeansList and Google Sheets
contact feeds carry the contact information they were already receiving from
Finalsite but discarding. Two defects, both in the extract layer, surfaced by
Zendesk 474545 (Newark DSOs, during Welcome Calls and BTSN RSVPs).

**1. `rpt_deanslist__family_contacts` discarded the real relationship label.**
The `case` on `contact_slot` replaced `Relationship` with the literal strings
`Emergency 1`..`Emergency 4` for every emergency-slot row. Finalsite populates a
real label on essentially all of them, so a student's mother sitting in an
emergency slot reached DeansList as `Emergency 1` and staff calling families
could not tell a parent from a neighbor. The fix passes `relationship` through
unchanged; `ContactType` already carries `Emergency` for DeansList message
routing, so the slot ordinal was redundant as well as harmful.

Newark rows in `int_finalsite__student_contacts`, showing the labels that were
being thrown away:

| Slot        | Rows   | Carry a relationship label | Distinct labels |
| ----------- | ------ | -------------------------- | --------------- |
| contact_1   | 10,747 | 10,747                     | 7               |
| contact_2   | 3,485  | 3,485                      | 4               |
| emergency_1 | 6,970  | 6,969                      | 13              |
| emergency_2 | 6,798  | 6,796                      | 13              |
| emergency_3 | 1,790  | 1,790                      | 13              |
| emergency_4 | 671    | 671                        | 13              |

Across NJ that recovers real labels on ~21,500 emergency rows, including 1,129
`Mother` and 3,369 `Father` — exactly the complaint in the ticket.

**2. `rpt_gsheets__student_contact_info` release columns were empty by
construction.** `release_1`..`release_3` were built from the `pickup_1`..
`pickup_3` slots and `release_4` was hardcoded null. No model has emitted a
`pickup_*` contact slot since the Finalsite cutover — `int_finalsite__student_contacts`
produces only `contact_1`, `contact_2`, and `emergency_1`..`emergency_4`, and the
Focus branch produces the same set — so every release column was empty for every
student in all four regions. Now built from `emergency_1`..`emergency_4`.

Verified against prod `int_extracts__student_enrollments`, current-vs-new:

| Region   | Students | release_1 before | release_1 after | release_2 | release_3 | release_4 |
| -------- | -------- | ---------------- | --------------- | --------- | --------- | --------- |
| Camden   | 2,148    | 0                | 1,585           | 1,441     | 164       | 53        |
| Miami    | 1,114    | 0                | 1,012           | 802       | 52        | 8         |
| Newark   | 6,795    | 0                | 6,641           | 6,489     | 1,731     | 647       |
| Paterson | 881      | 0                | 866             | 852       | 201       | 73        |

`array_to_string` rather than `concat`, because `concat` returns NULL if either
argument is NULL — that would drop the contact's name entirely for the ~6% of
emergency contacts with no mobile number on file. The phone falls back to
`phone_primary` for the same reason. `release_5` stays null; there is no fifth
emergency slot to source it from.

### Test change that needs a look

The DeansList model's uniqueness key was
`(StudentID, ContactType, Relationship)` at `severity: error`. That only held
because `Emergency N` doubled as a slot discriminator; with honest labels it
fails on 1,922 key groups. The source grain is `(student, contact_slot)`, but
the DeansList import template fixes the header set so the slot cannot be
projected. The key is now name plus relationship at `severity: warn` — two
emergency slots can legitimately hold the same person for one student (174 rows
NJ-wide), which is a Finalsite data-entry gap rather than an extract defect, and
papering over it with a dedupe would hide it from Ops.

Two other warns, for the record:

- `not_null` on `Relationship` now returns 6 rows. New, and intended: these are
  Finalsite emergency contacts with a name but no relationship recorded. Prod
  returns 0 today only because the literal `Emergency N` masked them.
- `not_null` on `ParentLastName` returns 9 rows. Pre-existing and unchanged —
  prod returns the same 9.

### Out of scope

512 active NJ students (Newark 292, Camden 154, Paterson 66) have no Parent 1
row because Finalsite carries no `primary` relationship flag for them. Of
Newark's 298, 265 have a `financial` relationship with no `primary`, 30 have
relationships that are neither, and 3 have no relationship rows at all. That is
a source data-entry gap plus a possible rule change (relaxing the Parent 1 pick
to `primary or financial` would fill 265 of the 298 today), being worked with
SRE and the Finalsite rep. Not touched here.

The header comment on `int_students__contacts_pivot` still claims `contact_2`
has no source rows, which is stale — Newark has 3,485. Left alone deliberately:
an inline SQL comment on that hub model marks it `state:modified` and fans a CI
rebuild across its whole descendant graph, for a comment fix.

`rpt_gsheets__student_contact_info.yml` carries no column descriptions. That is
pre-existing across all ~40 columns and backfilling them is a separate change.

## AI Assistance

Claude Code investigated the ticket, diagnosed both defects, quantified them
against prod, and wrote the changes and the test adjustment. Human-directed:
the decision to open this as a fix PR and the scoping of the Parent 1 population
problem out of it.

### General

- [x] Reviewed the diff

### dbt

- [x] Properties files updated for both changed models
- [x] No new or changed exposures — both models already have them
- [x] **Not a breaking change** — no columns renamed or removed. `release_4`
      changes type from `int64` (it was hardcoded null) to `string`, matching
      its sibling release columns; the Google Sheet consumer reads text.
- [x] No external source changes, so no `stage_external_sources` run needed

## Validation

Run from the worktree against the `kipptaf` project:

- `dbt build --select rpt_deanslist__family_contacts rpt_gsheets__student_contact_info --target dev --favor-state --defer` — PASS=8 WARN=3 ERROR=0. Contracts enforced on both models, including the `release_4` type change.
- `dbt test --select "test_type:unit,extracts.deanslist"` — 2 unit tests pass. `test_family_contacts_emergency_numbering` is renamed to `test_family_contacts_slot_contact_types` and now asserts relationships pass through instead of being renumbered.
- `trunk check --force` on all 4 changed files — no issues.

Refs #4751

### DIFF
--- a/src/dbt/kipptaf/models/extracts/deanslist/properties/rpt_deanslist__family_contacts.yml
+++ b/src/dbt/kipptaf/models/extracts/deanslist/properties/rpt_deanslist__family_contacts.yml
@@ -7,20 +7,26 @@ models:
       primary, falling back to financial; `contact_2`, the second parent from
       the student's household 1) plus emergency contacts (`emergency_1` through
       `emergency_4`). `ContactType` distinguishes `Parent1` / `Parent2` /
-      `Emergency` so DeansList can scope automated guardian messaging; emergency
-      contacts additionally carry a numbered `Emergency N` relationship, while
-      parents keep their real Finalsite relationship. Column names match the
-      DeansList import template headers exactly. Delivered by the Dagster asset
-      `kipptaf/extracts/deanslist/contacts_txt`.
+      `Emergency` so DeansList can scope automated guardian messaging, and every
+      contact carries its real Finalsite relationship to the student. Column
+      names match the DeansList import template headers exactly. Delivered by
+      the Dagster asset `kipptaf/extracts/deanslist/contacts_txt`.
     data_tests:
+      # The source grain is (student, contact_slot), but the DeansList template
+      # fixes the header set so the slot cannot be projected — name plus
+      # relationship is the closest key the output can carry. warn, not error:
+      # two emergency slots can legitimately hold the same person for one
+      # student, a Finalsite data-entry gap rather than an extract defect.
       - dbt_utils.unique_combination_of_columns:
           arguments:
             combination_of_columns:
               - StudentID
               - ContactType
+              - ParentFirstName
+              - ParentLastName
               - Relationship
           config:
-            severity: error
+            severity: warn
     columns:
       - name: StudentID
         data_type: int64
@@ -50,11 +56,13 @@ models:
       - name: Relationship
         data_type: string
         description:
-          Relationship of the contact to the student. For parents (`contact_1`
-          and `contact_2`) this is the Finalsite `rel_type` (`parent`,
-          `stepparent`, `guardian`, `grandparent`, etc.). For emergency contacts
-          this is `Emergency N`, numbered by the Finalsite emergency slot
-          (`emergency_1` through `emergency_4`).
+          Relationship of the contact to the student, for every slot. Parents
+          (`contact_1` and `contact_2`) carry the Finalsite `rel_type`
+          (`parent`, `stepparent`, `guardian`, `grandparent`, etc.); emergency
+          contacts carry the relationship recorded on their emergency
+          custom-field set (`Mother`, `Grandmother`, `Aunt`, `Neighbor`, etc.),
+          which is populated on nearly every emergency row. Null where Finalsite
+          holds an emergency contact with no relationship recorded.
         data_tests:
           - not_null
       - name: ContactType
@@ -104,13 +112,13 @@ models:
           send.
 
 unit_tests:
-  - name: test_family_contacts_emergency_numbering
+  - name: test_family_contacts_slot_contact_types
     description:
       One enrolled student with two parents plus all four emergency slots.
-      Asserts the parents keep their Finalsite relationship and get
-      Parent1/Parent2 contact types, and each emergency slot is relabeled
-      Emergency N by slot number with an Emergency contact type (exercising
-      every case branch).
+      Asserts contact_1 and contact_2 map to Parent1/Parent2 and every emergency
+      slot maps to Emergency (exercising each case branch), and that all six
+      rows pass their Finalsite relationship through unchanged rather than
+      having the emergency ones overwritten with a slot ordinal.
     model: rpt_deanslist__family_contacts
     given:
       - input: ref('int_finalsite__student_contacts')
@@ -200,7 +208,7 @@ unit_tests:
             WorkPhone: null,
             CellPhone: "13055550200",
             Email: null,
-            Relationship: Emergency 1,
+            Relationship: grandparent,
             ContactType: Emergency,
             Language: null,
           }
@@ -212,7 +220,7 @@ unit_tests:
             WorkPhone: null,
             CellPhone: null,
             Email: null,
-            Relationship: Emergency 2,
+            Relationship: aunt,
             ContactType: Emergency,
             Language: null,
           }
@@ -224,7 +232,7 @@ unit_tests:
             WorkPhone: null,
             CellPhone: null,
             Email: null,
-            Relationship: Emergency 3,
+            Relationship: uncle,
             ContactType: Emergency,
             Language: null,
           }
@@ -236,7 +244,7 @@ unit_tests:
             WorkPhone: null,
             CellPhone: null,
             Email: null,
-            Relationship: Emergency 4,
+            Relationship: neighbor,
             ContactType: Emergency,
             Language: null,
           }
--- a/src/dbt/kipptaf/models/extracts/deanslist/rpt_deanslist__family_contacts.sql
+++ b/src/dbt/kipptaf/models/extracts/deanslist/rpt_deanslist__family_contacts.sql
@@ -4,6 +4,7 @@ with
             sc.contact_first_name,
             sc.contact_last_name,
             sc.email,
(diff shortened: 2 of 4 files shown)


cbini:
Verified each item against the warehouse and the repo before responding.

## Issue 1 — `''` vs `NULL` on the release columns: correct, not changing it

The behavior claim holds exactly as described:

```sql
select
  array_to_string([cast(null as string), cast(null as string)], ' | ') is null,  -- false, returns ''
  concat(cast(null as string), ' | ', cast(null as string)) is null              -- true
```

So the new expression emits `''` where the old `concat` emitted `NULL`, and the cited precedent is real (`stg_focus__people.sql` and `stg_focus__address.sql` both wrap in `nullif(..., '')`).

Not applying the wrap, for two reasons.

**The nesting rule blocks the cheap version.** `nullif(array_to_string([name, coalesce(mobile, primary)], ' | '), '')` is three levels of function nesting. `src/dbt/CLAUDE.md` caps that at one level and gives `if(coalesce(x, y) > 0, 'a', 'b')` as the limit. The two `stg_focus__*` precedents fit because they wrap a bare array with no inner `coalesce` — two levels. Making this compliant means adding a CTE to derive the four phone fallbacks as named columns, which restructures a model that is currently one flat select.

**Nothing can observe the difference.** The only consumers are four Google Sheets exposures in `models/exposures/google-sheets.yml`. No SQL model reads this view, so there is no `IS NULL` check anywhere downstream to bite, and a sheet cell renders `''` and `NULL` identically. Restructuring the model to satisfy a nesting rule, for a value no consumer can distinguish, is not a trade worth making.

Dropping the phone fallback would let the `nullif` fit in two levels, but that costs a phone number on roughly 5.5% of emergency contacts (they have `phone_primary` with no `Cell`-typed number). For a feed whose purpose is Welcome Calls, the phone number is worth more than the null.

## Issue 2 — no unit test for the release derivation: declining

The derivation was validated against prod rather than mocked — the before/after population table in the PR body is a real query over `int_extracts__student_enrollments` across all four regions, and the contract build passed. A unit test here would need an `expect` fixture enumerating all ~40 output columns (every `expect` row must list the same columns, per `src/dbt/CLAUDE.md`) to cover logic whose only identified failure mode is the cosmetic empty string above. Reasonable ask if this model ever gets unit tests, but not proportionate on its own.

## Issue 3 — no uniqueness test on `rpt_gsheets__student_contact_info`

Agreed, and pre-existing. Out of scope here.

## Unrelated: the `Analyze (actions)` failure

Not this change. The job died in **Set up job**, before analyzing anything:

```text
Failed to resolve action download info. Error: Service Unavailable
Retrying in 29.963 seconds
Failed to resolve action download info. Error: Bad Gateway
##[error]Failed to resolve action download info.
```

A GitHub Actions infrastructure outage resolving action downloads. `Analyze (actions)` passed on each of the last four commits to `main`, and this PR touches only dbt SQL and YAML — no workflow files. A retry is rejected (`This workflow run cannot be retried`), so it needs a re-run from the Actions UI or it will clear on the next push to the branch.

PR #4757, opened 2026-08-06
fix(dagster): cursor powerschool users, schoolstaff, sectionteacher on transaction_date

### LINKED ISSUE #4754: fix(dagster): powerschool intraday sensor misses in-place updates that do not bump whenmodified
## Summary

The PowerSchool dlt intraday sensor detects change with a signature of `COUNT(*)` plus `MAX(cursor_column)`. An in-place UPDATE that changes row content without advancing the cursor is invisible to it. For tables configured `intraday: true` with `nightly: false` there is no full-refresh backstop, so the warehouse copy stays stale indefinitely.

Confirmed live against PowerSchool Paterson today.

## Evidence

Ops corrected `homeschoolid` for 10 Paterson staff. Those corrections are present in PowerSchool and absent from BigQuery 35+ hours later. Full-table comparison of all 219 live rows against the warehouse copy:

```text
rows live / warehouse        : 219 / 219   (no inserts, no deletes)
homeschoolid changed         : 10 rows
whenmodified changed         : 0 rows
transaction_date changed     : 105 rows

live   COUNT(*)              : 219                   stored baseline: 219
live   MAX(whenmodified)     : 2026-08-05T00:01:03   stored baseline: 2026-08-05T00:01:03
```

The probe signature is byte-identical to the stored baseline, so `_compute_changed` correctly reports no change. Each of the 10 rows still carries its original `whenmodified` value from 2023 through 2025. The sensor is behaving exactly as written; the cursor is the wrong signal.

Sensor tick history shows `no change across 37 probed tables` on every tick across the window.

## Root cause

`probe_signature` in `src/teamster/libraries/dlt/powerschool/assets.py` builds the signature from count plus max cursor. Three update classes are invisible:

1. An UPDATE that does not advance the cursor column. Confirmed live, above.
2. An UPDATE where the cursor is set to a value at or below the current max.
3. A balanced INSERT plus DELETE where the new row's cursor is at or below the current max.

The `_compute_changed` docstring already anticipates this for cursor-less tables: "in-place edits are caught by the nightly full refresh". Tables with a cursor and `nightly: false` have no such backstop.

## Scope

Tables configured `intraday: true` with a cursor and `nightly: false`:

- kipppaterson: 23
- kippnewark: 27
- kippcamden: 27

The set includes `users`, `students`, `attendance`, `storedgrades`, `cc`, `schoolstaff`, `sections`, and `studentcorefields`. Only `users` has been verified drifted; the others are exposed by configuration but unmeasured.

## Downstream impact

`rpt_powerschool__autocomm_teachers` joins the ADP staff roster to `stg_powerschool__users` on both `teachernumber` and `homeschoolid`. A stale `homeschoolid` fails that join, and the staff member also fails the new-user branch because their `teachernumber` does exist. They silently disappear from the autocomm extract and lose their PowerSchool login provisioning.

A count of active staff whose roster school does not match the warehouse `homeschoolid` returned 122 network-wide. That figure was measured against the possibly-stale warehouse copy, so it is an upper bound and some are likely already correct in PowerSchool. It needs re-measuring against live data once this is fixed.

## Candidate fixes, not yet decided

1. Set `nightly: true` for affected tables. Restores the documented backstop. Cost scales with table size, which is presumably why large tables are `nightly: false` today.
2. Switch the cursor to `transaction_date`. Verified it moved on all 10 drifted rows, and the live max advanced past the stored baseline, so the sensor would have fired. Caveat: it moved on 105 of 219 rows overnight, so it appears batch-maintained rather than per-edit and would trigger a near-daily reload.
3. Add a periodic reconciliation pass that compares a content checksum rather than a cursor, at a lower frequency than the 15 minute tick.

Option 1 and option 2 converge on roughly daily reloads for `users`. Neither obviously generalizes to `attendance` or `storedgrades`. Recommend brainstorming before any config change.

## Reproducing

A read-only probe over the SSH tunnel comparing live PowerSchool against the warehouse copy reproduces this. It was run as a throwaway test under `tests/assets/` and is not committed. Local runs need the `PS_` PowerSchool variables for the target code location present in the developer environment template; the Paterson set was missing and was added while investigating.

## Immediate remediation

A manual reload of the Paterson `users` asset loads unconditionally and will pull the 10 corrections into BigQuery, restoring those staff to the autocomm extract. That is a workaround, not a fix.

cbini:
# Pull Request

## Summary & Motivation

Change the intraday change-detection cursor from `whenmodified` to
`transaction_date` for `users`, `schoolstaff`, and `sectionteacher` in all three
PowerSchool districts.

The intraday sensor detects change with `COUNT(*)` plus `MAX(cursor_column)`, so
an in-place UPDATE that does not advance the cursor is invisible. These three
tables are `intraday: true` with `nightly: false`, so there is no full-refresh
backstop and the staleness persists indefinitely.

Confirmed live against PowerSchool Paterson: 10 staff rows had `homeschoolid`
corrected in PowerSchool, 0 of them advanced `whenmodified`, and the probe
signature was byte-identical to the stored baseline. `transaction_date` moved on
all 10. Across all three districts, every `users` row modified in 2026 carries a
`transaction_date` at or after its `whenmodified` — 284 of 284.

These three tables are the only exposed tables carrying both columns. The other
ten intraday tables in the same configs, including the four largest, already
cursor on `transaction_date`. Tables with only `whenmodified` are deliberately
untouched: no defect has been demonstrated for them.

This self-remediates. The stored baseline holds a `whenmodified` value, so the
first probe after deploy returns a differing signature and each table reloads
once, clearing the accumulated drift with no manual run.

Design:
`docs/superpowers/specs/2026-08-06-powerschool-dlt-cursor-transaction-date-design.md`

Closes #4754

## AI Assistance

Claude investigated the root cause, ran the live read-only comparison against
PowerSchool, and authored the spec, plan, and config change. Scope was
human-directed: limiting the fix to tables with a demonstrated defect and a
better cursor already available, and dropping a config parity test from scope.

### General

### Dagster

- [x] Config-only change consumed by `yaml.safe_load`; no Python modified
- [x] Verified the sensor builds `PowerSchoolTable` with the new cursor for all
      three tables

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes. Note this passes trivially: no dbt models are
      touched, so `state:modified+` selects nothing. Not validation.
- [ ] **Dagster Cloud** — passes for all three affected locations

### DIFF
--- a/docs/superpowers/plans/2026-08-06-powerschool-dlt-cursor-transaction-date.md
+++ b/docs/superpowers/plans/2026-08-06-powerschool-dlt-cursor-transaction-date.md
@@ -1,0 +1,585 @@
+# PowerSchool dlt Intraday Cursor Switch Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use
+> superpowers:subagent-driven-development (recommended) or
+> superpowers:executing-plans to implement this plan task-by-task. Steps use
+> checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Change the intraday change-detection cursor from `whenmodified` to
+`transaction_date` for the three PowerSchool tables that carry both columns, so
+in-place updates that do not bump `whenmodified` stop being invisible to the
+sensor.
+
+**Architecture:** Pure configuration change. The sensor already reads
+`cursor_column` per table from each district's `assets.yaml` and interpolates it
+into `SELECT COUNT(*), MAX({cursor_column}) FROM {table_name}`. Changing the
+configured value changes the probe with no code change. Ten sibling tables in
+the same files already use `transaction_date`.
+
+**Tech Stack:** YAML config consumed by `yaml.safe_load`, Dagster sensors, dlt,
+BigQuery, `uv` for Python execution.
+
+## Global Constraints
+
+- Spec:
+  `docs/superpowers/specs/2026-08-06-powerschool-dlt-cursor-transaction-date-design.md`
+- Issue: [#4754](https://github.com/TEAMSchools/teamster/issues/4754)
+- Worktree:
+  `/workspaces/teamster/.worktrees/cbini/fix/claude-powerschool-cursor-transaction-date`
+- Branch: `cbini/fix/claude-powerschool-cursor-transaction-date`
+- Every git command uses
+  `git -C /workspaces/teamster/.worktrees/cbini/fix/claude-powerschool-cursor-transaction-date`.
+  A bare `git` from the main repo commits to `main`.
+- Every Read, Edit, and Write targets the worktree path, not
+  `/workspaces/teamster/<path>`. Editing the main checkout silently leaves the
+  worktree unchanged.
+- Exactly three tables change: `users`, `schoolstaff`, `sectionteacher`. Do not
+  touch any other table entry.
+- Only `cursor_column` changes. Do not touch `intraday` or `nightly` on any
+  entry.
+- No Python changes. No new test files.
+- The `trunk` binary lives only in the main repo. Invoke it by absolute path
+  `/workspaces/teamster/.trunk/tools/trunk` with cwd set to the worktree. If
+  that path does not exist, use `~/.cache/trunk/launcher/trunk`.
+- `cursor_column: whenmodified` appears 13 times in the Paterson file and 17
+  times in each of the Newark and Camden files. Every edit must anchor on the
+  preceding `- table_name: <name>` line so it cannot match the wrong table.
+
+---
+
+## File Structure
+
+Three files are modified. No files are created.
+
+| File                                                                              | Change                                                         |
+| --------------------------------------------------------------------------------- | -------------------------------------------------------------- |
+| `src/teamster/code_locations/kipppaterson/powerschool/sis/dlt/config/assets.yaml` | 3 lines: `schoolstaff` L48, `sectionteacher` L56, `users` L88  |
+| `src/teamster/code_locations/kippnewark/powerschool/sis/dlt/config/assets.yaml`   | 3 lines: `schoolstaff` L48, `sectionteacher` L56, `users` L104 |
+| `src/teamster/code_locations/kippcamden/powerschool/sis/dlt/config/assets.yaml`   | 3 lines: `schoolstaff` L48, `sectionteacher` L56, `users` L104 |
+
+Line numbers are as of branch point `6d58d4607`. Anchor edits on the
+`table_name` line rather than trusting the number.
+
+---
+
+### Task 1: Switch the cursor for the three tables across all three districts
+
+**Files:**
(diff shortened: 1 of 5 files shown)


cbini:
Both findings verified. One fixed, one measured and answered.

## Issue 1 (Important): dcid literals in the plan doc — fixed in b91f7f0

Correct, and the internal contradiction is real: the same file said those values must stay local, then committed ten of them.

Worth noting the classification is narrower than stated. `dcid` is PowerSchool's internal surrogate key, not the employee number, and the root CLAUDE.md says PII-tagging precedent in powerschool staging "omits gender, race/ethnicity, and internal/student ids." So `dcid` is not `teachernumber`. But the same policy says to treat it as PII when unsure, ten rows map one-to-one to ten identifiable staff at a named school, and redaction costs nothing.

Rather than redact the list in place, I replaced the hardcoded row filter with a roster-join query that checks the invariant directly:

```sql
select count(*) as still_mismatched
from sr
inner join stg_powerschool__users as u
  on sr.powerschool_teacher_number = u.teachernumber
where u.homeschoolid != sr.adp_school
```

Better check, no identifiers. It verifies the property we actually care about instead of a frozen row list, and it stays correct if the affected set changes before deploy. Both docs are now free of individual identifiers; the specific rows live in git-ignored investigation artifacts.

## Issue 2 (Minor): steady-state reload cadence — measured, does not reproduce

Good catch on the mechanism, and the concern was the right shape: `replace` disposition behind a pool at limit 1 means frequent selection would be costly.

It does not hold, because `transaction_date` advances in discrete batches rather than continuously. Live `MAX(transaction_date)` on Paterson `users`:

| measured at | live MAX(transaction_date) |
| --- | --- |
| 16:12Z | 2026-08-06T07:17:33 |
| 17:11Z | 2026-08-06T07:17:33 |

Unchanged across about 10 hours, roughly 40 consecutive 15-minute ticks, every one of which would have compared equal and skipped. The 105 changed rows all moved inside a single batch window; between batches the signature is stable.

So steady state is closer to "reloads when a batch lands or a real edit happens" than "reloads most ticks." You flagged this at Medium confidence with no DB access, which was the right call from where you were sitting.

Keeping the suggestion anyway: Task 3 now has a step that pulls 100 ticks of sensor history post-deploy and counts how many selected these three tables, so the cadence is confirmed against production rather than inferred from a pre-deploy probe.

## Not changed

No functional change to the fix. Both findings were documentation and verification scope; the nine config lines are untouched since your review.

One note on CI state: `Analyze (python)` and `Trunk Check Runner` both show failure, and both are the ongoing GitHub Actions outage, not this diff. Each died at action-resolution before executing anything:

```text
Failed to resolve action download info.
Service Unavailable
```

githubstatus currently reports Actions at major_outage. Those two need a re-run once it clears.

PR #4759, opened 2026-08-06
docs(fresh): correct the goals live-read claim and map SRE goals workbook

GabyRangelB:
# Pull Request

## Summary & Motivation

Correct and substantially expand the FRESH goals documentation, based on a
column-by-column walkthrough of every tab in SRE's target workbook.

**The correctness fix.** Both the skill and the published data-model page stated
that the goals sheet is a live read, and the published page named
`stg_google_sheets__finalsite__goals` specifically as uncached. It is not: only
`src_google_sheets__finalsite__goals` reads the sheet live, while `stg_` and
`int_google_sheets__finalsite__goals_pivot` are native tables frozen at their
last build. The real hazard is therefore prod going **stale**, which is the
opposite of what the docs warned about. That inversion had two concrete costs:
it told a reader to dismiss a value mismatch as live-edit drift instead of
checking freshness, and it told the reconciliation loop that no rebuild was
needed after an analyst edit, which makes the loop never converge. Replaced with
a freshness check comparing the sheet's Drive `modifiedTime` against
`__TABLES__.last_modified_time`, plus an explicit rebuild step.

**What the walkthrough added.** None of this was written down:

- Which goals exist at which `goal_granularity`, so a reconciliation stops
  expecting values at levels a goal does not live at.
- That 17 of 39 goal combinations are scaffold rows that never carry values.
- Sourced vs derived, with matching a stated value taking precedence over
  computing a tidier one.
- Rounding is half-up: not ceiling, and not Python's `round()`, which is
  banker's rounding and differs at exactly `.5`.
- A tab-by-tab map of all five source tabs with column letters and cell ranges,
  and SRE's rule that only the main table on each tab is a source.
- Why Miami Tech's goals look broken and are correct: it opened to take KIPP's
  own 8th graders into grade 9, so it has returners and no application funnel.
  This also explains a previously unexplained exception in the skill.
- That tab-level reads are an MCP limitation rather than a permission gap, with
  the three connector paths tested and recorded so nobody re-tests them.

Docs only. No model, config, or code changes.

## AI Assistance

Claude drafted the documentation and ran the verification queries. The
walkthrough was human-directed throughout, and several of Claude's conclusions
were corrected in the process: the precedence of matching over deriving, the
main-tables-only rule, and the Miami Tech reading were all user corrections that
changed the outcome. Claude also removed goal values it had initially embedded,
after the user pointed out these docs should carry structure and process rather
than data.

## Self-review

### General

- [x] Reviewed. Docs-only change; no Asana due-date or assignee update needed.

Skipped: no Dagster changes.

Skipped: no dbt changes. No models, properties files, exposures, or sources
touched.

### Docs

- [x] No schedule, sensor, or integration changes, so the automations catalog
      does not need regenerating.
- [x] `docs/models/fresh-dashboard-data-model.md` is already in the `mkdocs.yml`
      nav; no nav entry needed.

## CI checks

- [x] **Trunk** — verified locally with
      `trunk check --force --no-fix` on both changed files, clean.
- [ ] **dbt Cloud** — expected to be a no-op; no dbt files changed.
- [ ] **Dagster Cloud** — expected not to trigger; no Dagster paths changed.

---
  - https://app.asana.com/0/0/1217251630729772


GabyRangelB:
@claude - can you review the md file and skill and compare it to the data model they support. please flag any discrepancies.

cbini (approved):

PR #4763, opened 2026-08-07
feat(launch): add the data tool catalog and intern runbook

### LINKED ISSUE #4761: feat(launch): migrate the data launch page off Google Sites to a git-sourced static page
## Problem

The data team launch page runs on a Google Site that is difficult to maintain. The tool catalog is hand-duplicated across five role pages and has already drifted between them. Nothing is version controlled, nothing is reviewable before it goes live, and link rot is silent.

## Approach

Author in git, serve from Zendesk Help Center, launch from Okta. One source of truth, multiple renderers.

Phase 1 covers the content migration plus the Zendesk publishing pipeline. Okta bookmark tiles and a Claude skill are follow-on work that reuse the same source files, and are not in this issue.

## Phase 1 scope

A new `src/launch/` directory:

- `links.yml` — the tool catalog, one entry per tool, tagged by audience
- `views.yml` — presentation metadata for the six catalog views
- `content/` — prose as markdown (Support, Our Team, Topline, Blog)
- `manifest.yml` — mapping from source path to Zendesk article id, CI maintained
- `publish/` — render and push

A GitHub Actions workflow on merge to `main`, path filtered, mirroring the existing MkDocs deploy pattern. It renders the catalog into six articles (All plus five role views), renders markdown into prose articles, and pushes both to the Zendesk Guide API.

Audience tags control relevance, never visibility. The All view always renders every entry, and every role view carries a persistent link to it so nobody is walled off from a tool by a mistagged entry.

Pipeline properties:

- **Idempotent** — bodies are compared before writing, so a no-op run does not churn `updated_at`
- **Never destructive** — a removed tool archives its article rather than deleting it
- **Adopts in place** — the existing Data Launch articles keep their URLs and search history

Validation runs on every PR and blocks the build on schema violations plus a content lint ruleset. A dry-run mode reports what would change in Zendesk before merge.

## Named open dependencies

1. A Zendesk Guide API credential with article create and update scope. Not yet verified — the available token is read-limited. This is the first task of implementation.
2. Help Center visibility. The Data categories are currently readable without signing in. Whether gating is applied per user segment or Help Center wide is undecided and involves other departments. The pipeline treats visibility as configuration, not an architectural assumption.
3. Whether the existing Data Launch articles map cleanly onto new sources.
4. Ownership of the content rewrite.

## Out of scope for phase 1

- Okta bookmark tiles
- Claude skill for tool and resource discovery
- Metrics glossary — becomes a catalog link to the existing Apps Script until specced separately
- Blog publishing — posts migrate into the repo, but going live stays gated on comms review

## Tracked follow-ups

A review of the current public site surfaced content remediation items covering published contact details, sign-in guidance that is both stale and inadvisable to publish, and a team page that needs rework. Specifics are held internally rather than restated here. These are tracked to be addressed as content moves, and the corresponding checks form part of the phase 1 lint ruleset.

anthonygwalters:
# Pull Request

## Summary & Motivation

...establish `src/launch/` as the source of truth for the data team's tool
catalog, plus the runbook for finishing it.

**Deliberately independent of where the catalog gets served.** The companion
design spec (#4762) proposes rendering this into Zendesk Help Center articles,
but that spec is waiting on several open questions. The catalog is not: it is
the same source whether it renders to Zendesk, to Okta bookmark tiles, to a
Claude discovery skill, or to a ticket-response agent. Splitting it out means it
can merge and be worked on now rather than inheriting the spec's blockers.

### What is in `links.yml`

**46 tools**, scraped from the five role pages of the existing Google Site and
merged — 35 Tableau, 7 Google Sheets, 3 AppSheet apps, 1 Zendesk. Duplicates
that appeared under different names with identical URLs were merged; navigation
chrome was dropped.

Every entry carries `status: needs-review`, with **61 `TODO` lines** marking
what a human has to resolve. This is a starting point, not a finished catalog.

The scrape surfaced real defects that the five-page duplication had been
hiding:

- A tool that resolves to **two different URLs** depending on which page you
  came from, with one of its two descriptions plainly belonging to a different
  tool
- **Two live dashboards with near-identical names**, indistinguishable to a
  teacher
- **Contradictory descriptions** for the same tool on different pages
- An entry that **points at a help article** rather than the dashboard it
  describes

Those four are exactly the failure mode this change exists to end.

### The four Student Contact Info Feeds

Added directly rather than scraped — they were never linked from the Google
Site, and they do not appear anywhere else in this repository either. Sharing
was confirmed with the data team: each is scoped to its region's group with CMO
staff holding access to all four, none link-shared.

### Google-hosted entries carry an extra rule

A Tableau URL is gated by SSO no matter who holds it. **A Drive URL is gated by
that file's sharing setting** — so if a file is shared "anyone with the link,"
the URL _is_ the access control, and publishing it in a public repository hands
it out. `README.md` makes that check mandatory for every `google-*` entry.

It also documents a trap: for a file you do not administer, the Drive API
returns only the owner, not the groups. An empty group list from a script is
inconclusive, not evidence of a problem. Check the Share dialog.

### The other three files

- `views.yml` — the six views and their intro copy. Presentation only;
  membership is derived from `audiences` tags, so the five-way duplication
  cannot come back.
- `README.md` — the reference: what each field means, what "verified" requires.
- `RUNBOOK.md` — the task sequence and the ownership split.

Refs #4761

## AI Assistance

Claude extracted and merged the catalog from the live site, wrote the runbook
and reference docs, and verified Drive sharing where the API permitted. Scope
and every judgment call were human-directed: which content is public, what
belongs in the catalog versus authored elsewhere, and the decision not to seed
this from the dbt Google Sheets exposures.

That last one is worth recording. Claude proposed scraping the 63 exposures in
`google-sheets.yml` as a catalog source and was overruled with two facts it did
not have: the exposures file does not contain the Student Contact Info Feeds at
all, and staff-facing tools sit in every category the names suggest — including
the ones that look most like machine plumbing. There is no heuristic that sorts
them. That triage is recorded in `RUNBOOK.md` as human work, explicitly not
delegated.

## Self-review

### General

N/A — no Dagster code. `src/launch/` contains data and documentation only.

N/A — no dbt changes.

N/A — no schedule, sensor, or integration changes.

## CI checks

- [ ] **Trunk** — passes. Verified locally with `trunk check --force` across all
      four files, since yamllint and markdownlint fire at push rather than at
      the commit hook.
- [ ] **dbt Cloud** — passes.
- [ ] **Dagster Cloud** — not expected to trigger; no watched paths change.

## Reviewer notes

**Worth a look before merge:** `src/launch/` has no `CODEOWNERS` entry, so it
currently matches only the catch-all `* @TEAMSchools/admins`. Every pull request
against this directory will require an admins approval, which will be a
bottleneck once someone is working through the catalog daily. Adding
`/src/launch/ @TEAMSchools/admins @TEAMSchools/data-team` would spread that
load. Not included here since it is a `.github/` change.

**Merging this unblocks the catalog work**, which is independent of every open
question in #4762.

### DIFF
--- a/src/launch/README.md
+++ b/src/launch/README.md
@@ -103,49 +103,11 @@ setting, _and_ it reads from a backing Google Sheet with its own separate
 sharing. Both have to require sign-in. There is no API shortcut here — open the
 app, then open the Sheet behind it, and check each.
 
-### Tools that are on the site but not on any role page
+### What order to do it in, and which entries need a decision
 
-Three AppSheet apps (`Leader PM App`, `Seat Tracker`, `Stipend App`) were linked
-only from the Support page, so there is no signal about who they are for. They
-are in `links.yml` with empty `audiences` and a TODO. Assigning those is a real
-decision, not a lookup.
-
-Related question worth raising rather than answering alone: Clever, DeansList,
-Grow, and PowerSchool are named on the Support page, but all four link to the
-same Okta dashboard URL rather than to the apps themselves. Should they be
-catalog entries at all, and if so what should they point at?
-
-### Suggested order
-
-Do the straightforward ones first to build context, then bring the judgment
-calls to a person rather than guessing.
-
-1. **Start with the tools you already know.** Confirm the pattern works before
-   scaling up.
-1. **Work through the rest alphabetically.** Most will be quick.
-1. **Save the `TODO(verify)` entries for last** and discuss them. There are four
-   and each needs a real decision, not a lookup:
-   - `Home Instruction Tracker` points at **two different URLs** depending on
-     which page of the old site you came from, and one of the two descriptions
-     is plainly the wrong tool's. Somebody has to decide which is canonical.
-   - `Literacy Tool (ELIT)` and `Early Literacy Tool (ELIT)` are **different
-     dashboards with nearly identical names**. Both appear live. They probably
-     need renaming so a teacher can tell them apart.
-   - `Testing Accommodations` has two contradictory descriptions on the old site
-     — results versus requests. These are not the same thing.
-
-### Things worth flagging as you go
-
-You are the first person to look at all of these at once, which makes you the
-most likely person to notice:
-
-- A tool that appears twice under different names
-- A tool nobody has used in years
-- A tool that is missing from the list entirely
-- A description that would confuse somebody who did not already know the tool
-
-Open an issue or leave a note. Finding these is genuinely part of the job, not a
-distraction from it.
+See [RUNBOOK.md](RUNBOOK.md). That file owns the sequence, the entries that need
+a judgment call rather than a lookup, and what to flag as you go. This file
+stays the field reference so the two cannot drift apart.
 
 ## Out of scope
 
--- a/src/launch/README.md
+++ b/src/launch/README.md
@@ -1,0 +1,156 @@
+# Data launch page — tool catalog
+
+This directory holds the source of truth for the data team's tool catalog. It
+gets rendered into role-specific articles in the Zendesk Help Center.
+
+Design: `docs/superpowers/specs/2026-08-06-launch-page-zendesk-design.md` (under
+review separately).
+
+| File         | What it is                                                      |
+| ------------ | --------------------------------------------------------------- |
+| `links.yml`  | Every tool: name, URL, description, who it is for               |
+| `views.yml`  | The six views and their intro copy. Presentation only           |
+| `RUNBOOK.md` | The task sequence — start there if you are picking up this work |
+| `README.md`  | This file: what the directory is and what "reviewed" means      |
+
+The renderer and publishing pipeline do not exist yet. This is the data they
+will consume.
+
+## Current state
+
+`links.yml` was **scraped from the existing Google Site** and merged. It is a
+starting point, not a finished catalog. Every entry carries
+`status: needs-review`.
+
+- **46 tools** — 35 Tableau, 7 Google Sheets, 3 AppSheet apps, 1 Zendesk
+- **61 `TODO` lines** flagging specific things a human needs to resolve
+- **9 tools** have a linked user guide; the rest may or may not have one
+- **4 tools** have no `audiences` at all and need roles assigned
+
+The scrape found several kinds of problem, which is a good sign — these are
+exactly the failures that come from maintaining the same list on five separate
+pages by hand.
+
+## The work
+
+Review every entry and change its `status` to `verified`. An entry is verified
+when all five of these are true:
+
+1. **The tool still exists** and the URL loads. You will need Tableau access for
+   most of them; ask if you do not have it.
+1. **The name matches what the tool actually calls itself** when you open it,
+   not what the old site called it.
+1. **The description is accurate and one sentence.** Say what the tool is for,
+   not how it works. Many scraped descriptions are decent; some are stale.
+1. **`audiences` is right.** Who actually needs this in their day-to-day? See
+   `views.yml` for what each role means. A tool can be in several. A tool in
+   none still appears in the All view, so an empty list is a real answer — but
+   it should be a decision, not an accident.
+1. **`system` is right** — `tableau`, `appsheet`, `zendesk`, `google-sheet`,
+   `google-slides`, `google-form`, `google-doc`, `apps-script`, or `other`.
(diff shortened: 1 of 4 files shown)


anthonygwalters:
All three findings verified and fixed in 39151654e, plus a related fix on #4762.

### 1. View count — valid, and it went deeper

Confirmed: `views.yml` defines five views (`all`, `teachers`, `leaders`, `ops`,
`region`), and the README said six.

It was not a stray word. The same miscount appears **five more times in the
design spec** on #4762, including one line that contradicts itself:

> Six catalog articles are generated: `All` plus five role views.

`All` plus five role views would be six, but only four role views exist. The
error came from counting the Google Site's five role _pages_ and then adding
`All` again — `All` was already one of the five.

Corrected to "five articles: `All` plus four role views" in both places
(a6bb61bcd on the spec branch).

### 2. Zendesk asserted as the destination — valid

Both file headers claimed a settled destination while the PR body argues the
catalog is independent of one, and they cited a spec path that is not in this
branch.

`README.md` now says the catalog renders into staff-facing articles, names
Zendesk as the current design candidate with a pointer to #4762, and states that
nothing in the directory depends on that choice being settled. `links.yml`'s
header got the same treatment, and `RUNBOOK.md`'s dangling path reference became
an issue reference.

### 3. `regions:` undocumented — valid

Added a field reference covering `guide`, `access`, and `regions`, with the
distinction spelled out explicitly, since the two read as the same thing:

- `regions:` is **geography** — which of `newark` / `camden` / `miami` /
  `paterson` a tool covers. Used by the four Student Contact Info Feeds.
- `audiences: [region]` is the **_Regional and CMO_ role view**.

A tool can carry `audiences: [region]` with no `regions:` at all, or be scoped
to one region and used by teachers.

### On the "confirmed clean" section

The counts, duplicate-id check, PII grep, `audiences`-against-`views.yml`
validation, and the YAML quoting check are all things worth having verified
independently rather than taken from the PR narrative. Noting that the review
distinguished what it verified from what it inferred, and flagged its own
confidence on finding 2 as Medium — that made triaging these faster.

anthonygwalters:
@claude review

anthonygwalters:
Second round: finding 1 valid and fixed in 92470b7b7, but at the cause rather
than the symptom. Finding 2 does not hold — details below.

### Finding 1 — valid, fixed differently than suggested

The count mismatch is real, but it is a symptom. `README.md` carried **three**
sections that duplicated `RUNBOOK.md`:

- `### Suggested order` — duplicated Step 1 and "Bring these to Anthony"
- `### Things worth flagging as you go` — duplicated Step 2
- `### Tools that are on the site but not on any role page` — duplicated the
  empty-`audiences` bullet and the Clever/DeansList/Grow/PowerSchool question

They were copied when `RUNBOOK.md` was split out and never deleted from
`README.md`. The divergence the review found is what that duplication produces.

Adding the missing fourth bullet, as the suggested fix proposed, would have kept
both copies and guaranteed the same drift again on the next edit. Instead all
three sections are removed; `README.md` points at `RUNBOOK.md` and stays the
field reference, which is what its own file table already claimed it was.
Verified `RUNBOOK.md` covers everything removed before deleting it.

### Finding 2 — does not hold

The claim:

> the last bullet (the four empty-`audiences` entries: `Early Literacy Tool`,
> `Leader PM App`, `Seat Tracker`, `Stipend App`) is tagged `TODO(audiences)` in
> `links.yml`, not `TODO(verify)`

Those entries carry **both** tags. Counting `TODO(verify)` lines per entry:

```text
home_instruction_tracker:      4
literacy_tool:                 3
early_literacy_tool:           1
testing_accommodations:        3
stipend_and_bonus_dashboard:   3
leader_pm_app:                 2
seat_tracker:                  2
stipend_app:                   2
```

Every entry referenced by all five RUNBOOK bullets has at least one
`TODO(verify)` line, the three AppSheet apps included — each has an
`# TODO(verify): AppSheet app. Confirm its access setting requires sign-in AND
check the sharing on the Sheet backing it.` pair alongside its
`TODO(audiences)` lines.

So `RUNBOOK.md:62` — "They are marked `TODO(verify)` in the file" — is accurate
for all five bullets. No change made. The review appears to have read the
`TODO(audiences)` lines at those offsets and concluded they were the only tags
present.

### On the round overall

The re-verification approach was the right one — checking whether the prior
fixes actually held rather than only diffing the fix commit is how finding 1
surfaced, and that one was worth having.

PR #4765, opened 2026-08-07
fix(level_data): retry transient 5xx responses in GrowResource

### LINKED ISSUE #4764: fix(level_data): retry transient 5xx responses in GrowResource
## Problem

The `kipptaf/schoolmint/grow/user_sync` asset check `zero_api_errors` failed and fired an alert on run `b25428e3-4095-4ddf-af49-6f2a230e0446`, even though the run itself succeeded.

The single recorded error was a Google frontend 502 ("The server encountered a temporary error and could not complete your request. Please try again in 30 seconds.") on a `PUT` to the Grow users endpoint for one staff user. Every other user and school update in that run succeeded.

## Root cause

`GrowResource` has no retry for HTTP errors:

```text
resources.py:76-84  _request() converts any non-2xx into GrowAPIError and raises immediately
resources.py:86-91  the only @retry is on get(), predicate = GrowIncompleteResponseError only
                    post() / put() / delete() have no retry at all
assets.py:170-180   grow_user_sync catches GrowAPIError, appends to errors, continues
```

So one transient upstream 5xx becomes a permanently failed check, and the WARN alert pages a human for something that self-heals: the sync is idempotent, and the next daily run re-applies the skipped update. The run after this one passed clean.

Frequency over the last 50 daily check executions: 2 failures. This 502, and one on 2026-06-30 that was a genuine data problem (a duplicate-email ValidationError on a `POST`) and is out of scope here.

## Proposed fix

Raise a `GrowServerError(GrowAPIError)` subclass for 5xx in `_request()` and retry it with the tenacity set already imported in that module (`stop_after_attempt`, `wait_exponential_jitter`, `retry_if_exception_type`), matching the pattern in `libraries/adp/workforce_now/api/resources.py`.

Scope the retry to `GET`, `PUT`, and `DELETE`. Leave `POST` failing fast: a 502 on a create may have landed server-side, so retrying it would produce exactly the duplicate-email ValidationError seen in the June failure.

4xx keeps its current behavior (raise `GrowAPIError`, recorded in the check) since those are deterministic.

## Verification

Unit test in `tests/resources/` that mocks a 5xx-then-200 sequence and asserts the `PUT` recovers, plus a test that a 4xx still raises immediately.

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will stop a single transient 5xx from the Grow API turning into a failed asset check and a WARN alert.

Run `b25428e3-4095-4ddf-af49-6f2a230e0446` succeeded, but its `zero_api_errors` check on `kipptaf/schoolmint/grow/user_sync` failed. The one recorded error was a Google frontend 502 on a `PUT` to the Grow users endpoint for a single staff user. `GrowResource._request` converted it to `GrowAPIError` and raised immediately: the only retry in that module was on `get()`, with a `GrowIncompleteResponseError` predicate, so `post` / `put` / `delete` had no retry at all.

This adds `GrowServerError(GrowAPIError)` for 5xx on idempotent requests and retries it in `_request` with the tenacity set already imported there, matching `libraries/adp/workforce_now/api/resources.py`.

- `POST` is excluded: a 5xx on a create may have landed server-side, so retrying risks a duplicate record. That is the exact duplicate-email ValidationError seen in the 2026-06-30 check failure.
- 4xx keeps its fail-fast behavior and still lands in the asset's error list.
- A retryable 5xx logs at WARNING instead of ERROR, so recovered flakes stop filing GCP Error Reporting groups.
- No change to `grow_user_sync`: `GrowServerError` subclasses `GrowAPIError`, which the asset already catches in all three places.

Frequency check over the last 50 daily executions: 2 failures, this 502 and the June ValidationError. Impact of the bug was a false-positive alert rather than stale data, since the sync is idempotent and the next run reapplied the skipped update.

## AI Assistance

Diagnosed and authored by Claude Code under the systematic-debugging skill, human-directed. Root cause was traced from the run logs and check metadata to the resource before any code changed. Fix scope, the POST carve-out, and the branch strategy were confirmed with the author.

### General

- [x] Review the Claude Code Review comment posted on this PR

### Dagster

- [x] `uv run pytest tests/resources/test_resource_grow_retry.py` passes, 4 tests, offline with a stubbed session
- [x] `trunk check --force` clean on both changed files
- [x] No new asset, schedule, sensor, or integration, so no docs regeneration needed

## CI checks

- [ ] Trunk
- [ ] dbt Cloud
- [ ] Dagster Cloud

Refs #4764

### DIFF
--- a/src/teamster/libraries/level_data/grow/resources.py
+++ b/src/teamster/libraries/level_data/grow/resources.py
@@ -1,11 +1,13 @@
 import copy
+from typing import NoReturn
 
 from dagster import ConfigurableResource, DagsterLogManager, InitResourceContext
 from dagster_shared import check
 from oauthlib.oauth2 import BackendApplicationClient
 from pydantic import PrivateAttr
 from requests import Response, Session
-from requests.exceptions import HTTPError
+from requests.exceptions import ConnectionError as RequestsConnectionError
+from requests.exceptions import HTTPError, Timeout
 from requests_oauthlib import OAuth2Session
 from tenacity import (
     retry,
@@ -32,11 +34,12 @@ class GrowAPIError(Exception):
 
 
 class GrowServerError(GrowAPIError):
-    """Raised when the Grow API returns a 5xx on an idempotent request.
+    """Raised when a request to the Grow API fails transiently.
 
-    An upstream gateway flake (502/504), not an application bug. Recoverable via
-    retry. POST is excluded: a 5xx on a create may have landed server-side, so
-    retrying it risks a duplicate record.
+    Covers a 5xx response and a connection-level failure (refused, reset, DNS,
+    timeout) on an idempotent request. An upstream flake, not an application
+    bug, so it is recoverable via retry. POST is excluded: the create may have
+    landed server-side, so retrying it risks a duplicate record.
     """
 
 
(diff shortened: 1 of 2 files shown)


cbini:
Verified each finding against the code before acting. Pushed `8354b09f0`.

**1. Connection-level errors — fixed.** Confirmed: `_request` caught only `HTTPError`, so a `ConnectionError` or `Timeout` from `self._session.request()` escaped unwrapped, and `grow_user_sync` catches only `(GrowAPIError, GrowIncompleteResponseError)`. That crashes the run rather than recording one row. Verified the cited sibling predicate in all three: `adp/workforce_now/api/resources.py:56,69`, `knowbe4/resources.py:43`, `coupa/resources.py:51,65`.

Fixed by catching `(RequestsConnectionError, Timeout)` and routing both failure classes through one `_raise_request_error` helper that makes the transient-vs-terminal call and picks the log severity. Same POST exclusion applies: a connection failure mid-`POST` can also mean the create landed.

One deliberate omission: no `kwargs.setdefault("timeout", (10, 60))` like ADP has. Without it the `Timeout` branch only fires when a caller passes `timeout=`, so including `Timeout` is coverage for future callers rather than a live path today. Adding a default read timeout would change behavior for the paginated `observations` and `assignments` pulls, which is a bigger change than this PR should carry.

**2. No `get()` coverage — fixed.** Added `test_get_retries_on_server_error` (single-record branch) and `test_get_paginated_retries_on_server_error` (paginated branch, with the outer `GrowIncompleteResponseError` retry's wait also stubbed). Plus two connection-error tests for finding 1. Suite is 8 tests, all passing.

**3. `_get_access_token` has no retry — not in this PR.** The claim is accurate and the convention citation is right. Deferring because it is a different failure mode with no observed occurrence: over the last 50 daily executions of this check, the two failures were this 502 and a data-level ValidationError, neither at token fetch. It also is not a pure copy of the ADP shape — `requests_oauthlib`'s `fetch_token` can raise oauthlib's own error types for a 5xx, so the predicate needs verifying rather than assuming. Better as its own change with its own test.

**4. Inverted exception convention — agreed, no change.** The method-aware POST exclusion cannot be expressed as a type predicate over a shared `HTTPError`, which is why the polarity differs from the siblings.

On the environment note: `uv run pytest` (8 passed) and `trunk check --force` (no issues) both ran locally against the worktree.

PR #4772, opened 2026-08-07
feat(finalsite): rank parent contacts densely and guard against student contacts

### LINKED ISSUE #4768: feat(finalsite): redefine parent selection in int_finalsite__student_contacts
## Problem

`int_finalsite__student_contacts` selects Parent 1 and Parent 2 with a household join that is not part of the package's documented rule, and caps output at two slots. An audit of the three cutover NJ regions found:

- A missing `primary` flag suppresses both slots, because `contact_2` is defined relative to Parent 1's household. 344 of 7,082 current Newark students got no contact at all under the pre-refresh snapshot.
- The Parent-1-household filter drops legitimate non-resident parents, and fails outright when a parent has a duplicate contact record that splits the household link.
- Nothing checks that a selected contact is not themselves a student, so a peer can be picked as a parent.
- A third qualifying parent is silently discarded by an arbitrary `relationship_id` tie-break. 32 students network-wide have three.

The package CLAUDE.md already states the intended rule: `primary` is Parent 1, an additional `financial`-without-`primary` relationship is Parent 2, and households carry only id plus address with no roles. The household join is an undocumented addition that diverges from it.

## Planned scope

Design doc to follow in `docs/superpowers/specs/`.

- Drop the Parent-1-household filter. Use household co-membership as a `contact_2` sort key instead of a gate.
- Guard both slots on the related contact being an adult, via `status = 'not_in_workflow'`.
- Emit `contact_3` rather than discarding extra qualifying parents.
- Add warn tests for students with 0 contacts and with 3 or more, plus a stale-enrolled test on `stg_finalsite__contacts`.
- Defensive fix in `dim_student_contact_persons` so a third parent cannot be misfiled as an emergency contact.

No student-scope filter is planned: the downstream receivers already exclude graduated students correctly via their PowerSchool enrollment joins, and filtering on the Finalsite side would wrongly drop 15 current retained seniors whose Finalsite record was never rolled forward.

Wide receivers (contacts pivot, ParentSquare, DeansList, bridge) keep their two-slot surface; extending them is deferred.

cbini:
# Pull Request

## Summary & Motivation

Redefine which people become a student's parent contacts in
`int_finalsite__student_contacts`, and stop four downstream models from
mishandling the result.

Previously `contact_1` was the relationship flagged `primary`, and `contact_2`
had to co-reside in a household with `contact_1`. That coupling meant a missing
`primary` flag suppressed **both** slots, and a third qualifying parent was
silently discarded by an arbitrary `relationship_id` tie-break. The household
join was also undocumented — the package CLAUDE.md already described the
intended rule as "`primary` = Parent 1, an additional `financial`-without-
`primary` relationship = Parent 2" and noted that "households carry only id plus
address — membership has no roles."

After this change:

- A **parent candidate** is any relationship flagged `primary` or `financial`
  whose related contact is an adult (Finalsite `status = 'not_in_workflow'`).
  That guard is what keeps a co-resident sibling out of a parent slot, and it
  works even though an adult sibling can legitimately be a guardian — which a
  `rel_type` filter could not.
- Candidates are ranked by the `primary` flag, then by sharing a household with
  the **student**, then by `relationship_id`, and numbered **densely** from
  `contact_1` with no ceiling. Household co-membership orders candidates instead
  of excluding them.
- Five warn-severity tests surface the Finalsite data gaps this exposes.

Measured across the three cutover NJ regions on the 2026-08-07 load, scoped to
currently enrolled students:

| Region   | Students | 0 slots | 1     | 2     | 3   |
| -------- | -------- | ------- | ----- | ----- | --- |
| Newark   | 6,786    | 0       | 3,682 | 3,079 | 25  |
| Camden   | 2,163    | 52      | 1,264 | 840   | 7   |
| Paterson | 879      | 43      | 539   | 297   | 0   |

Newark goes from 344 students with no contact at all to zero. Camden and
Paterson still have 52 and 43, which predate this change and are Finalsite
data-entry gaps, not rule failures.

### Downstream corrections

Dense numbering emits slots past `contact_2`, which four models mishandled. All
four assumed exactly two parent slots via a stale complementary predicate:

- `rpt_deanslist__family_contacts` labelled a third parent `Emergency` and
  exported it to DeansList. DeansList routes guardian-only automated messaging
  by `ContactType`, so those parents silently lost guardian messaging. 48 rows
  reach that extract today. **This was the highest-impact defect in the branch
  and no test caught it.**
- `dim_student_contact_persons` recorded a third parent as an emergency contact,
  keyed as if it had no person record.
- `bridge_student_contacts` kept the old key shape, which would have produced
  foreign-key orphans once the dimension stopped emitting those rows.
- Two `accepted_values` enumerations on `contact_slot` in kipptaf would have
  failed CI. Both are replaced by a pattern assertion, because dense ranking has
  no upper bound and an enumerated list keeps breaking.

`rpt_clever__students` is a deliberate exception: it is a per-contact feed
rather than a two-slot one, so a third parent flows through as a primary
contact. About 32 students gain a legitimate parent in the Clever sync.

Extending the wide receivers — the contacts pivot, ParentSquare, DeansList — to
carry a third parent is **out of scope** and tracked separately.

### What is deliberately not here

No student-scope filter. The model stays SIS-agnostic and emits a row for every
student record. Verified that receivers already scope correctly: of 303 Newark
records still marked `enrolled` for the prior school year, only 15 reach the
extracts, and PowerSchool confirms all 15 are actively enrolled retained
seniors. Filtering on the Finalsite side would have removed 15 current students'
contacts.

### Known follow-ups

- The DeansList fix has no regression test — its two existing unit tests have no
  third-parent fixture. This is the class of bug that regresses.
- The comment in `int_finalsite__student_contacts__no_extra_contacts.sql` names
  four of the six consumers. The spec is accurate and complete; the comment is a
  pointer.

## AI Assistance

Claude-authored under human direction throughout. The human made every design
call: replacing the slot enumeration with a pattern assertion, adding the
guard-exclusion test, keeping the Clever behaviour, and rejecting a
`rel_type`-based filter on the grounds that an adult sibling can be a legal
guardian. Claude ran the audit, wrote the spec and plan, and executed via
subagents with a review gate per task plus a whole-branch review.

Four defects in Claude's own spec surfaced during execution and are corrected in
the branch history — all from measuring against currently-enrolled students
while the model deliberately emits every student record.

### General

### dbt

      models; the five new singular tests get a `tests/properties.yml` carrying
      `meta.dagster.ref` with `package: finalsite`, so they report as asset
      checks rather than logging observations across all parents.
- [x] Include (or update) an
      — no new dashboard consumers.
- [x] **Breaking change?** No column renames or removals. `contact_slot` gains
      values past `contact_2`; every consumer is handled above.
- [x] External sources — none added or modified.

## CI checks

- [ ] **Trunk** — `trunk check --force` run locally over all 14 changed files:
      no issues.
- [ ] **dbt Cloud** — the kipptaf chain cannot be exercised locally, because in
      a dev target its finalsite sources resolve to prod schemas holding the
      pre-change two-slot output. CI is the first real end-to-end run.

Closes #4768

### DIFF
--- a/docs/superpowers/plans/2026-08-07-finalsite-parent-definition.md
+++ b/docs/superpowers/plans/2026-08-07-finalsite-parent-definition.md
@@ -597,9 +597,11 @@ Create
 -- stays SIS-agnostic and emits rows for every student record including
 -- prospects and applicants, who legitimately have no parent on file yet -- so
 -- the enrolled scope lives here, in the test, rather than in the model.
--- Every row is a Finalsite data-entry gap: a missing `primary`/`financial`
--- flag, or a parent whose own contact record is miskeyed with a student status
--- and therefore fails the adult guard.
+-- Every row is a Finalsite data-entry gap: no relationship on the record
+-- carries a `primary` or `financial` flag. A parent miskeyed with a student
+-- status does NOT surface here -- dense ranking backfills the slot from the
+-- next candidate -- so that case is reported by
+-- stg_finalsite__contact_relationships__caregiver_is_adult instead.
 select
     s.finalsite_enrollment_id,
     s.grade_name,
@@ -609,7 +611,7 @@ where
     s.status = 'enrolled'
     and s.school_year_start = {{ var("current_academic_year") }}
     and not exists (
-        select 1
+        select 1,
         from {{ ref("int_finalsite__student_contacts") }} as c
         where
             c.finalsite_enrollment_id = s.finalsite_enrollment_id
--- a/docs/superpowers/plans/2026-08-07-finalsite-parent-definition.md
+++ b/docs/superpowers/plans/2026-08-07-finalsite-parent-definition.md
@@ -52,16 +52,16 @@ prettier via trunk, dbt unit tests and singular tests.
 
 ## File Structure
 
-| Path                                                                                  | Responsibility                                     |
-| ------------------------------------------------------------------------------------- | -------------------------------------------------- |
-| `src/dbt/finalsite/dbt_project.yml`                                                   | Declare `current_academic_year` var (default `0`)  |
-| `src/dbt/finalsite/tests/stg_finalsite__contacts__no_stale_enrolled.sql`              | Warn: `enrolled` record stamped with a prior year  |
-| `src/dbt/finalsite/tests/stg_finalsite__contact_relationships__single_primary.sql`    | Warn: a student holds more than one `primary`      |
-| `src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql`       | Parent candidate set, dense ranking, adult guard   |
-| `.../properties/int_finalsite__student_contacts.yml`                                  | `accepted_values`, description, unit-test fixtures |
-| `src/dbt/finalsite/tests/int_finalsite__student_contacts__enrolled_has_contact_1.sql` | Warn: current enrolled student with no `contact_1` |
-| `src/dbt/finalsite/tests/int_finalsite__student_contacts__no_extra_contacts.sql`      | Warn: any slot past `contact_2`                    |
-| `src/dbt/kipptaf/models/marts/dimensions/dim_student_contact_persons.sql`             | Stop misfiling an unknown parent slot as emergency |
+| Path                                                                                  | Responsibility                                      |
+| ------------------------------------------------------------------------------------- | --------------------------------------------------- |
+| `src/dbt/finalsite/dbt_project.yml`                                                   | Declare `current_academic_year` var (default `0`)   |
+| `src/dbt/finalsite/tests/stg_finalsite__contacts__no_stale_enrolled.sql`              | Warn: `enrolled` record stamped with a prior year   |
+| `src/dbt/finalsite/tests/stg_finalsite__contact_relationships__single_primary.sql`    | Warn: a student holds more than one `primary`       |
+| `src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql`       | Parent candidate set, dense ranking, adult guard    |
+| `.../properties/int_finalsite__student_contacts.yml`                                  | Slot-pattern test, descriptions, unit-test fixtures |
+| `src/dbt/finalsite/tests/int_finalsite__student_contacts__enrolled_has_contact_1.sql` | Warn: current enrolled student with no `contact_1`  |
+| `src/dbt/finalsite/tests/int_finalsite__student_contacts__no_extra_contacts.sql`      | Warn: any slot past `contact_2`                     |
+| `src/dbt/kipptaf/models/marts/dimensions/dim_student_contact_persons.sql`             | Stop misfiling an unknown parent slot as emergency  |
 
 ---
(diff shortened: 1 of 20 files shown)


cbini:
Triaged all three findings. Two are fixed; one is dismissed with evidence.

## Important 2 — no regression test on the downstream consumers: **fixed** (`24aba0114`)

Correct, and broader than the PR body admitted. `dim_student_contact_persons.yml` and `bridge_student_contacts.yml` had no `unit_tests` block at all, so the fix for the highest-impact defect shipped with nothing locking it.

Added a unit test to each of the three receivers. Every one feeds a `contact_1`, `contact_2`, `contact_3` and `emergency_1` row for the same synthetic student, then asserts the `contact_3` row is absent from the output while the other three survive — the emergency assertion matters as much as the exclusion, since the filter must not take emergency contacts with it.

Each test was proven non-vacuous by removing that model's slot filter, confirming the test fails with the `contact_3` row leaking through, then restoring it and confirming a pass.

## Minor 1 — redundant `severity="warn"`: **fixed** (`78162ef5c`)

Correct. All four consuming districts set an unscoped `data_tests: +severity: warn`, and finalsite package tests only ever build inside a district, so the per-file config was dead. Dropped from all five singular tests; the `severity: error` overrides on the `expression_is_true` pattern tests are left alone, as they deliberately override that default.

Test outcomes are unchanged after the removal — 303 / PASS / 1 / PASS / 49 — which is the confirmation the district default does apply.

## Important 1 — untested `LIKE` escape: **dismissed**

The precedent half is right: `git grep` finds no other `\_` escape in `src/dbt`. The conclusion that it is "genuinely untested against real BigQuery" is not — it was verified three separate ways before this review ran:

1. Direct BigQuery execution of the predicate: `contact_1`, `contact_2` and `contact_15` match; `contactX1` and `emergency_1` do not.
2. A `countif` over the 21,843 `emergency_*` rows in the three cutover region tables — all 21,843 match `like 'emergency\_%'`. If the escape were broken this would have been 0.
3. This PR's own dbt Cloud run built the full chain. Had emergency rows silently stopped matching, the DeansList tests would have moved; instead its null and duplicate counts are identical to production — 9 null `ParentLastName`, 6 null `Relationship`, 174 duplicate groups.

The review states its own confidence here as Medium and notes it could not confirm against a live warehouse. It is confirmed. The silent-failure blast radius was the right thing to worry about; the worry is retired.

Keeping `LIKE`. Swapping four call sites to `regexp_contains` would touch three models again to replace a predicate now demonstrated correct, and `LIKE` is the cheaper operator. The consistency argument is fair, but not worth the churn on a proven expression.

## Note on the dbt Cloud warnings

All six warnings on this run are pre-existing and unrelated to the branch. The three on `rpt_deanslist__family_contacts` were checked against production, which still runs the old logic, and the counts match exactly. The 174 duplicate groups are the emergency-slot duplication documented as out of scope in the design spec — the same person emitted as a parent slot and again as two separate emergency contacts under a name-spelling variant. No tracking issue exists for it yet.

Issue #4587, opened 2026-07-28
fix: convert FK-carrying marts tables to views to stop BigQuery referenced-primary-key failures


cbini:
## Symptom

`dbt build` fails intermittently on marts models with a BigQuery `invalidQuery` error:

```text
Referenced primary key in table dataset_reference 	 {
  project_reference {
    project_id: "teamster-332318"
  }
  dataset_id: "kipptaf_marts"
}
table_id: "dim_course_sections"
 has been updated.
```

Most recent occurrence: Dagster run `d55363e8-76ce-43c9-b0cb-fcba98a1b9b7`, dbt node `dim_student_section_enrollments`, BigQuery job `a891227f-d486-4043-874b-4f69f8945b00`. The auto-retry `d326bbf5-01c1-4457-b237-232d8c5c7f8e` succeeded on identical table state.

## Root cause

Marts models declare BigQuery PK/FK constraints under an enforced contract. For `materialized: table`, dbt emits the FK into the CTAS DDL:

```sql
course_section_key string references `teamster-332318`.`kipptaf_marts`.`dim_course_sections` (course_section_key) not enforced,
```

BigQuery resolves the referenced table's primary key at query-analysis time and re-validates it at commit. A concurrent `create or replace table` on the parent produces a new table generation with a new PK identity, so the child's CTAS aborts. This is a catalog-metadata consistency failure, not a data-integrity violation — `not enforced` means no row is ever validated.

All four occurrences in the last 30 days overlap a parent rebuild:

| failed model | job window (UTC) | FK parent | parent CTAS window |
| --- | --- | --- | --- |
| `dim_student_section_enrollments` | 07-28 13:39:19 to 13:39:31 | `dim_course_sections` | 13:39:25 to 13:39:26 |
| `dim_student_section_enrollments` | 07-28 04:10:57 to 04:11:10 | `dim_student_enrollments` | 04:11:02 to 04:11:04 |
| `fct_assessment_scores_enrollment_scoped` | 07-23 16:17:33 to 16:17:45 | `dim_assessment_administrations` | 16:17:22 to 16:17:27 |
| `fct_assessment_scores_enrollment_scoped` | 07-21 01:48:47 to 01:48:57 | `dim_student_section_enrollments` | 01:48:37 to 01:48:48 |

[...]

PR #4773, opened 2026-08-07
perf(dbt): drop vacuous marts PK not_null tests and materialize a survey intermediate

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will remove 69 dbt tests that cannot fail,
materialize one survey intermediate as a table, and correct marts guidance that
currently points readers at a change known to break prod builds.

Three commits, in order:

**1. Remove 69 vacuous `not_null` tests from marts primary keys.** Every marts
PK is `dbt_utils.generate_surrogate_key` output, which never returns NULL —
`src/dbt/CLAUDE.md` already states the rule. Each of these tests still costs a
full BigQuery scan per CI run, and because most marts are views that scan
re-expands the whole upstream chain rather than reading one column. A single
`not_null` on `fct_assessment_scores_enrollment_scoped` dry-runs at 1.71 GiB
across 24 base tables.

Every deletion was verified against the model SQL, not just the YAML: each PK
traces to an unwrapped `generate_surrogate_key`, either directly or through a
non-nulling pass-through. Prod confirms zero nulls. Measured over 30 days, the
68 of them that ran cost 13,230 executions and 9.18 USD.

Two `not_null` tests were deliberately kept: `dim_dates.date_key` is a raw DATE,
and `fct_assessment_scores_enrollment_scoped.assessment_date_key` guards the
invariant the #4546 fix rests on.

**2. Correct the marts materialization guidance.** The exception paragraph in
`models/marts/CLAUDE.md` still described the eight-day window between #4464 and
#4587. It claimed the assessment-scores star and its FK-closure dims are
materialized as tables and warned against reverting them to views — but #4587
reverted the seven FK-carrying models to views on 2026-07-28, so the paragraph
named the current, correct setup as something to avoid. Replaced with the six
marts that are actually tables, why `dim_assessments` carries a nightly cron,
and the prohibition on materializing any FK-carrying mart.

**3. Materialize `int_surveys__manager_survey_details` as a table.** 175k rows
re-derived from 7 base tables on every read: 0.35 GiB per expansion against 0.03
GiB stored, the worst recompute-to-size ratio in the survey chain, and 32 survey
tests expand it. The survey marts cannot be materialized instead — they carry FK
constraints and #4587 covers exactly that — but this intermediate carries none.
Nightly rather than eager, because its only consumers are the survey marts (read
by Cube on a daily pre-aggregation refresh) and two Tableau exposures with no
Dagster-managed cron.

`int_surveys__survey_responses` was implemented and then backed out. Its hourly
consumers would force an hourly tick, and 24 rebuilds/day of 2.4 GiB plus
per-tick tests costs roughly 0.42 USD/day against roughly 0.9 USD/day of total
survey test spend measured over 30 days — net negative on quiet days. A nightly
tick would pay for itself but adds up to 24h of staleness to `survey_dashboard`
and `personalized_survey_links`.

### Not a breaking change

No model SQL, column, or contract changes. Test removals only drop checks that
are vacuous by construction, and the materialization change alters storage, not
output. One visible side effect: 69 dbt tests are 69 Dagster asset checks, so
that many checks disappear from the Dagster UI.

### Deploy note

A properties-yml-only materialization flip does not bump the dagster-dbt code
version (it is a SHA1 of raw SQL), so `int_surveys__manager_survey_details`
stays a view until the first midnight cron tick. Do not read BigQuery object
types right after merge and conclude it failed.

## AI Assistance

Claude Code did the investigation, measurement, and implementation; direction,
scope decisions, and the call to back out `int_surveys__survey_responses` were
human. The cost analysis that motivated this work is Claude's, and one of its
intermediate figures was wrong before correction — an early query mislabeled its
UTC window as America/New_York and inflated a day's CI cost by roughly 2.8x. All
figures quoted above are from corrected, calendar-day-aligned queries.

### General

### dbt

      models; edits are to existing properties files
- [x] Include (or update) an
      or changed consumers
- [x] **Breaking change?** No. No renamed or removed columns in any contracted
      model; see "Not a breaking change" above

### Docs

- [x] No schedule, sensor, or integration changes. The `models/marts/CLAUDE.md`
      edit is a correction to existing guidance, not a new integration

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

## Verification performed

- `dbt parse --no-partial-parse --target prod` clean; manifest `not_null` count
  drops 291 to 222, exactly 69
- `dbt build` is not run locally here — dbt Cloud CI builds
  `int_surveys__manager_survey_details` as a table on the PR schema, which is
  the real gate for the materialization change
- Diff shape checked: 69 files at exactly -1/+0, all 69 removed lines identical
- `trunk check --force` clean on all changed files

### DIFF
--- a/src/dbt/kipptaf/models/marts/CLAUDE.md
+++ b/src/dbt/kipptaf/models/marts/CLAUDE.md
@@ -251,11 +251,26 @@ Marts inherit `contract: enforced: true` and `materialized: view` from
 explicit uniqueness test on its PK (`unique` on a single column, or
 `dbt_utils.unique_combination_of_columns` for composite).
 
-Exception: the assessment-scores star (`fct_assessment_scores_enrollment_scoped`
-plus its FK-closure dims) and the three assessment intermediates are
-`materialized: table` for Cube query performance
-([#4464](https://github.com/TEAMSchools/teamster/issues/4464)) — don't revert to
-view without re-profiling Cube's BigQuery spend.
+Exception: `dim_assessments`, `dim_courses`, `dim_dates`, `dim_regions`,
+`dim_staff`, `dim_students`, and the four assessment intermediates are
+`materialized: table`. `dim_assessments` also carries a nightly
+`automation_condition.cron_schedule` — Cube is its only consumer and Cube's
+pre-aggregation refreshes daily, so intraday rebuilds are invisible
+([#4559](https://github.com/TEAMSchools/teamster/issues/4559)).
+
+**Never materialize an FK-carrying mart as a table.** A table child emits a
+`references` clause into its CTAS DDL, and BigQuery aborts the build when a
+concurrent `create or replace table` on the FK parent changes the parent's PK
+identity ("Referenced primary key in table ... has been updated"). The
+automation condition's `~any_deps_in_progress` guard is upward-facing only, so
+nothing serializes a parent against an in-flight child.
+[#4464](https://github.com/TEAMSchools/teamster/issues/4464) moved the whole
+assessment-scores star to tables for Cube query performance;
+[#4587](https://github.com/TEAMSchools/teamster/issues/4587) reverted the seven
+FK-carrying models to views eight days later, after four such build failures in
+30 days. Per-read recomputation was accepted deliberately to remove the failure
+class — restoring the performance means solving the collision, not re-flipping
+the config.
 
 Drop model-level `dbt_utils.unique_combination_of_columns` when its column set
 equals the surrogate-key hash inputs — `unique` on the PK detects the same
@@ -276,6 +291,9 @@ A `config.materialized: table` mart renders `constraints:` into CREATE TABLE DDL
 foreign_key both render into DDL and warn otherwise.
 
 ## Converting a view mart to a table
+
+Only a mart with NO outgoing `foreign_key` constraints is eligible — see the
+FK-carrying prohibition above.
 
 - BigQuery FK DDL requires every referenced relation to be a TABLE with a PK
   constraint. Convert the full FK closure (walk `to: ref(...)` edges) in one PR
--- a/src/dbt/kipptaf/models/marts/bridges/properties/bridge_assessment_expectations_enrollment_scoped.yml
+++ b/src/dbt/kipptaf/models/marts/bridges/properties/bridge_assessment_expectations_enrollment_scoped.yml
@@ -16,7 +16,6 @@ models:
             warn_unsupported: false
         data_tests:
           - unique
-          - not_null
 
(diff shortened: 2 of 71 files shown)


cbini:
Both findings verified against the code and the warehouse, both fixed in c28b91a5c.

**Cube consumer claim — correct, comment reworded.** No cube reads a survey mart. `src/cube/model/cubes/` has `conformed`, `courses`, `staff`, `student_assessments`, `student_attendance`, `students` and no `surveys`, and no `sql_table` in any cube points at a survey mart. The `cube_semantic_layer` exposure listing them is the every-mart-appears-here convention from `models/marts/CLAUDE.md`, not a query, exactly as you said. The comment now rests the nightly cadence on the two Tableau exposures alone.

One precision note on the supporting evidence: `grep -ri survey src/cube` returns 3 files, not zero — `staff/staff.yml` (column descriptions referencing the staff information survey as a source for gender and ethnicity), `staff/staff_reporting_relationships.yml` (a comment about future fact cubes), and `conformed/terms.yml` (a description listing period categories). All prose, none a model reference, so the conclusion holds and the fix is the same. Flagging it only because a future reader re-running that grep will get hits and may think the finding was wrong.

**Missing uniqueness test — correct, added.** Verified the grain in prod before adding so it cannot land red: `(survey_id, effective_survey_response_id, survey_question_id)` is unique at 175,670 of 175,670 rows. That grain is deliberate — it matches the hash inputs `fct_survey_responses` uses for its PK, so the test guards the fact's key at the source. Your reasoning about why it matters more now is the part that decided it: a data-change automation condition only re-materializes tables, so as a view this model had no recurring check.

Small correction to the framing — the model was not untested. It carries a pre-existing singular test, `test_int_surveys__manager_survey_details__terms_coverage`, plus a warn-level `not_null`. It was specifically the uniqueness test that was missing.

**On the unverifiable figures.** All of them came from `region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT` via the BigQuery MCP, and one was wrong before the PR body was written: an early query mislabeled its UTC window as America/New_York and inflated a day's CI cost about 2.8x. Everything quoted is from corrected, calendar-day-aligned queries.

Worth flagging that the query you suggested will undercount badly:

```sql
where query like '%int_surveys__manager_survey_details%'
```

Compiled test SQL names the mart it tests, not the intermediate underneath it, so that predicate catches only the models being built and misses every test reading through the marts. Across all principals it returns 0.71 USD over 30 days; the actual survey test spend is 25.96 USD. The reliable measurement is per-test cost keyed on the `node_id` in the dbt query comment, which is how the figures in the description were derived.

Issue #4771, opened 2026-08-07
feat(dbt): onboard Paterson to EdPlan so its IEP data matches the other NJ regions


cbini:
Paterson should resolve IEP status the same way Newark and Camden do, starting
AY2026. It cannot today: EdPlan carries no Paterson data at any layer.

## Current state

`dim_student_iep_status` has two legs. Newark and Camden take `nj_leg`, sourced
from `int_edplan__njsmart_powerschool_union`. Paterson takes `pm_leg`, sourced
from PowerSchool `studentcorefields.spedlep`, alongside Miami:

```sql
where e._dbt_source_project in ('kipppaterson', 'kippmiami')
```

EdPlan has no Paterson presence to switch to:

| Surface                                              | Newark | Camden | Paterson       |
| ---------------------------------------------------- | ------ | ------ | -------------- |
| Dagster EdPlan assets                                | yes    | yes    | none           |
| `kipppaterson` dbt EdPlan models                     | n/a    | n/a    | none           |
| `kipptaf` source declaration                         | yes    | yes    | not declared   |
| Rows in `int_edplan__njsmart_powerschool_union`      | 38,373 | 15,623 | 0              |

So making `pm_leg` Miami-only right now would drop Paterson IEP entirely rather
than move it. This is source availability, not a modeling choice.

## ELL is already correct

Worth stating so the two dimensions' asymmetry does not read as an oversight.
`stg_powerschool__s_nj_stu_x` carries Paterson (1,085 rows, 229 with
`lepbegindate`), so `dim_student_ell_status` already routes Paterson through
`nj_leg` and its `pm_leg` is already Miami-only. Only IEP is affected.

## Scope

1. Confirm Paterson has an EdPlan/NJSMART feed to ingest — SFTP account, file
   drop, and data. If it does not exist, that procurement is the real blocker
   and everything below waits on it.
1. Add EdPlan ingestion to the `kipppaterson` Dagster code location.
1. Add the `edplan` package to `kipppaterson` and build its
   `int_edplan__njsmart_powerschool_union`.
1. Declare `kipppaterson_edplan` in a `kipptaf` `sources-kipppaterson.yml` and
   add it to the `int_edplan__njsmart_powerschool_union` union list.
1. Change `dim_student_iep_status`'s `pm_leg` filter to Miami-only, with an
   AY2026 boundary so Paterson history stays on the PowerSchool path and only
   AY2026 forward comes from EdPlan.

## Validation

[...]

Issue #4729, opened 2026-08-04
fix(dbt): KIPP Miami absent from the kipptaf PowerSchool vertical for AY2026


cbini:
## Summary

Miami's SIS moved to Focus and the `kippmiami_powerschool` dataset is a frozen
archive (final pull 2026-07-01). Forty-nine kipptaf models still union that
archive as their Miami branch, so from AY2026 forward Miami contributes **zero
rows** to the entire kipptaf PowerSchool vertical — and to the 38 marts
downstream of it.

This is silent. The models build green; Miami simply stops appearing.

**Miami's first day of AY2026-27 is 2026-08-12.** Attendance-driven reporting
goes wrong for Miami from that date.

## Verified in prod

Already wrong (the enrollment dim is not date-bounded):

```text
model                                          Miami AY2025   Miami AY2026
int_powerschool__student_enrollment_union            1,942              0
dim_student_enrollments                              1,522              0
```

Wrong from 2026-08-12 (the daily attendance scaffold carries the full
AY2026-27 year for the three NJ regions and nothing for Miami):

```text
int_powerschool__ps_adaadm_daily_ctod              238,368              0
```

For scale, that scaffold holds 1,264,800 AY2026 rows for Newark spanning
2026-08-19 to 2027-06-17, and 426,472 for Camden from 2026-08-03. Miami has
none.

Focus holds the data — `kippmiami_focus.int_focus__student_enrollment` carries
1,585 Miami student-enrollment rows for AY2026 (and full back-history to
AY2018). Nothing is lost; it is simply not wired into the network layer.

## Reporting impact

Per `src/dbt/kipptaf/CLAUDE.md`, topline Total Enrollment derives from
`fct_student_attendance_daily`. That fact is date-bounded, so it is not
understating yet — AY2026-27 has not started for most regions. From
2026-08-12 it will understate network enrollment by roughly 1,585 students,
and Miami attendance, ADA, and chronic-absenteeism reporting will read empty.

`dim_student_enrollments` is already wrong today: 0 Miami rows for AY2026
against 10,117 for the NJ regions.

## Scope of the gap

- 49 kipptaf models union `kippmiami_powerschool`, across 90 declared source
  tables in `models/powerschool/sources-kippmiami.yml` (35 actually
  referenced). 47 of the 49 are same-named `union_relations` passthroughs.
- 38 of 79 marts models are downstream of the archive.
- Grouping those 38 marts by what drives their grain:
[...]

PR #4775, opened 2026-08-07
feat(dbt): restore KIPP Miami to the kipptaf identity spine for AY2026

### LINKED ISSUE #4731: feat(dbt): Focus branch into the kipptaf identity spine (restore Miami for AY2026)
Phase 1 of #4729. Restores KIPP Miami to the kipptaf network layer for AY2026 by
adding a Focus branch to the models that carry student, school, and enrollment
identity.

Miami's SIS moved to Focus, `kippmiami_powerschool` is a frozen archive, and 49
kipptaf models still union that archive as their Miami branch. Fixing the seven
identity models below restores Miami rows to **24 of the 38 affected marts** —
the highest-leverage slice of the gap, and the only one buildable today (Phase 2
and the grades half of Phase 3 wait on Focus producing attendance and gradebook
rows; see #4220).

## Scope

| Model | Focus source | Marts downstream |
| --- | --- | --- |
| `int_powerschool__teacher_grade_levels` | `int_focus__schedule` + `int_focus__users` + `stg_focus__co_teachers` | 17 |
| `int_powerschool__student_enrollment_union` | `int_focus__student_enrollment` | 14 |
| `stg_powerschool__students` | `int_focus__students` | 11 |
| `stg_powerschool__u_studentsuserfields` | Focus custom-field pivots | 11 |
| `stg_powerschool__schools` | `int_focus__schools` | 9 |
| `stg_powerschool__studentcorefields` | `int_focus__students` | 3 |
| `int_powerschool__terms` / `stg_powerschool__terms` | `stg_focus__marking_periods` | 5 |

Prerequisite inside this issue: declare and wrap the Focus models that #4725
shipped but never exposed at kipptaf level — `stg_focus__co_teachers`,
`stg_focus__students_join_users`, `int_focus__schedule`. Per
`kipptaf/CLAUDE.md`, every source added to a `sources-kipp*.yml` needs a
matching `union_relations` passthrough, and consumers read the wrapper.

## Approach

Miami's branch in `int_powerschool__student_enrollment_union` is Focus for real
enrollment stints plus the frozen archive for alumni graduate placeholders only,
filtered `enroll_status = 3 and entrydate is null`. The archive is never
dropped — it holds the 1,002 placeholder rows that `kipptaf/CLAUDE.md` requires
for KIPP Forward reporting, and Focus has no equivalent.

Focus covers AY2018 through AY2026 and matches the archive's real stints once
placeholders are excluded, so no history cutover date is needed.

`union_relations` builds a column superset and null-fills absent columns, so
PowerSchool-only columns (dcids, NJ state fields) resolve to null for Miami with
no special handling.

Value translations, all available on `int_focus__students`:

| Network column | Focus column |
| --- | --- |
| `lep_status` | `english_language_learner_pk_12` |
| `spedlep`, `special_education_code` | `ese_fefp_code` |
| `lunchstatus` | `free_reduced_meals_program` |
| race and ethnicity | `race_*`, `ethnicity_hispanic_or_latino`, `single_ethnicity` |
| `cohort` | derived from `year_entered_ninth_grade`, null for K-8 |

The governing rule for each translation is to reproduce the values Miami's
PowerSchool archive carried — every consumer was written against those, and
several already branch on `region = 'Miami'` to pass Miami's own domain through.

## Validation

1. **NJ parity** — the three NJ regions' output must be row-identical to current
   prod. `count(*)` plus `count(distinct format('%T|%T', ...))` on key columns.
2. **Miami historical reconciliation** — for AY2018 through AY2025, compare the
   Focus-conformed rows against the archive's real stints on student number,
   academic year, and entry date, then on each translated value. This is the only
   real test of the ESE, ELL, and meal translations, and it is possible only
   because Focus carries history.
3. **Consumer resolution** — `dbt build --empty` across the downstream graph.
4. Uniqueness test on each modified model's key.

## Out of scope

- SIS-neutral column vocabulary and the `base_` retirement — Phase 5, tracked in
  #3999 and #2541. This issue conforms Focus into the existing PowerSchool column
  names for zero consumer churn.
- Attendance (Phase 2) and gradebook grades (Phase 3) — blocked on Focus
  producing rows, see #4220.
- No mart output column changes, so no Cube updates.

Refs #4729
Refs #4220
Refs #3584

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request restores KIPP Miami to the kipptaf network layer
for AY2026 by adding a Focus branch to the seven models that carry student,
school, and enrollment identity.

Miami's SIS moved to Focus, `kippmiami_powerschool` is a frozen archive, and
kipptaf still unioned that archive as Miami's branch. Miami therefore contributed
**zero rows for AY2026** to the whole PowerSchool vertical. Miami's first day of
AY2026-27 is 2026-08-12.

Headline result, measured in a dev build:

| Model                                       | Miami AY2026 before | after   |
| ------------------------------------------- | ------------------- | ------- |
| `int_powerschool__student_enrollment_union` | 0                   | 1,632   |
| `stg_powerschool__terms`                    | 0                   | 49      |
| `int_powerschool__terms`                    | 0                   | 28      |
| `int_powerschool__teacher_grade_levels`     | 0                   | 136     |
| `dim_student_iep_status` (all years)        | 8,315               | 9,947   |
| `dim_student_meal_eligibility_status`       | 8,315               | 9,947   |
| `dim_student_ell_status`                    | 467                 | 568     |

Three Miami schools that opened after the PowerSchool freeze -- Miami Tech,
Legacy Elementary, Legacy Middle -- now resolve for the first time.

Design: `docs/superpowers/specs/2026-08-04-focus-identity-spine-design.md`
Plan: `docs/superpowers/plans/2026-08-06-focus-identity-spine.md`

### Decisions worth reviewing

**1. An AY2026 cutover, reversing the spec.** The spec concluded no cutover date
was needed because Focus covers AY2018 forward. That is true about coverage and
wrong about semantics. Focus dates a returning student's stint to the real first
day of school; PowerSchool used a July 1 administrative rollover. Reconciliation
found identical stint counts (8,315 both sides) with 954 AY2025 students differing
by exactly 42 days. July 1 is the network convention, not a Miami quirk -- Camden
dates 2,118 AY2025 stints there and Newark 6,472.

`entrydate` is an input to the `student_enrollment_key` hash, so adopting Focus
dates for closed years would recompose 954 historical keys and silently orphan the
**114,038 `fct_family_communications` rows** attached to Miami's historical
enrollments. The archive therefore keeps AY2025 and earlier; Focus supplies AY2026
forward. The spec has been corrected in place.

**2. `dcid` removed from the mart joins rather than minted for Focus.** Three
marts joined students to core and user fields on `students_dcid`, which Focus has
no equivalent for. Minting a synthetic one would have spread a
rubric-prohibited PowerSchool identifier (marts R2, R7, plumbing definition) into
a second SIS. Instead the new `int_students__*` intermediates resolve
`studentsdcid` to `student_number` at the layer where plumbing belongs, and the
marts join on `student_number`. The one remaining `dcid` join
(`stg_powerschool__s_nj_stu_x`) is NJ-only, which Miami correctly never matches.

**3. Three status fields carry forward and are null for new students.**
`spedlep`, `lunchstatus`, and `lep_status` have no usable Focus source -- Focus
`ese_fefp_code` covers 162 of 419 archive SPED students,
`free_reduced_meals_program` is a single school-wide CEP constant, and
`english_language_learner_pk_12` puts 98% of students at the not-applicable code.
Returning students carry the archive value; students new since the freeze get
`null`, never a fabricated default, each with a warn-severity test. Counts: 1,002
null `spedlep`, 502 null `lunchstatus`, 502 null `lep_status`.

**4. Archive-only records are kept, not dropped.** The archive holds 492 Miami
students and 1 school (the `999999` graduated-students sentinel, which the 1,002
alumni placeholders join to) that Focus never received. A blanket archive
exclusion would have removed them from `dim_students` silently, so both unions
anti-join instead.

### Unplanned improvement

Restoring Miami students to `dim_students` fixes a pre-existing FK gap:
`bridge_student_contacts` orphaned `student_key` values drop from **2,077 in prod
to 25**.

### Known-null and out-of-scope

- The three status dimensions carry null status for students new since the freeze
  until Focus ESE, meal, and ELL fields are populated. Tracked alongside #4220.
- 33 non-mart consumers keep reading `stg_powerschool__*` and stay Miami-less
  until Phase 5, so a Tableau extract and Cube will disagree for Miami during this
  window. Expected, not a defect. Full list in the spec appendix.
- One AY2026 Focus student id lacks the `8400` prefix and needs an Ops correction
  in Focus. A warn test flags it.
- `fct_grades_assignments` still selects and joins on `students_dcid`, the same
  rubric violation fixed here for the student dimensions. It sits in the Phase 3
  gradebook vertical and Miami has no gradebook data yet, so it is left alone.
- Paterson IEP should move to the NJ EdPlan path but cannot: EdPlan carries zero
  Paterson rows and no ingestion exists. Filed as #4771; it does not block this.

## AI Assistance

Claude Code authored the implementation and the design and plan documents, working
task-by-task against a written plan with human direction at each decision point.
The four decisions above were surfaced by Claude during implementation and
confirmed against prod before being applied; the reversal of the spec's no-cutover
conclusion in particular came out of reconciliation, not review.

### General

### dbt

- [x] Properties file for all new models -- all 11 new models have one, with a
      description on the model and every column, and a uniqueness test.
- [x] No exposure changes needed -- no mart output columns change, so no Cube or
      Tableau surface moves.
- [x] Not a breaking change -- no contracted column is renamed or removed. Column
      vocabulary stays PowerSchool's; renaming is Phase 5 (#3999, #2541).
- [ ] **Action required before CI can pass**: `zz_stg_kippmiami_focus` is missing
      all three models #4725 shipped, so kipptaf CI will fail deterministically
      until the staging copies are refreshed. This recreates shared tables and
      needs a human to run it:

      uv run dbt clone --target staging --state src/dbt/kippmiami/target/prod --project-dir src/dbt/kippmiami

## Validation

What was verified, and what could not be.

**Verified in a dev build:**

- Identity: all 3,453 returning Miami students keep their existing
  `dim_students.student_key` -- 0 keys moved. Zero ethnicity drift across the same
  3,453. `int_focus__students_conformed` is 1 row per student.
- The unprefix rule is independently corroborated by Focus's own
  `powerschool_id`, which equals the archive `student_number` for all 3,453.
- Alumni placeholders preserved at exactly 1,002.
- Teacher identity resolves: 194 of 195 Focus staff ids map to the network roster
  via `staff_number_identifier_local`.
- No NJ row is dropped by any new filter -- verified against the source rows each
  build actually read, per region, for every modified union.
- Trunk clean across all 47 changed files.

**Could not be verified locally, and why.** kipptaf `sources-kipp*` resolve to
personal `zz_<user>_*` copies under `--target dev`, and mine are stale for
Camden, Newark, and Paterson -- they predate AY2026. A dev build therefore
produces a `dim_student_enrollments` missing ~9,319 NJ AY2026 rows, which makes
every `student_enrollment_key` relationships test report orphans that are an
artifact of the environment, not the change. NJ parity was validated against what
the build actually read instead of against prod. **dbt Cloud CI, after the staging
clone above, is the real gate for FK integrity.**

Refs #4729
Refs #4220
Refs #4771
Closes #4731


cbini:
## CI status — one failing check, pre-existing and not from this branch

`dbt Cloud` is red on a single test:

```text
dbt_utils_unique_combination_of_columns_rpt_parentsquare__parents_student_id__first_name__last_name
Got 9 results, configured to fail if != 0
```

**This is not caused by this PR.** Evidence:

| Check                                              | Result                     |
| -------------------------------------------------- | -------------------------- |
| Duplicate groups in the PR schema                  | 9 (18 rows)                |
| Duplicate groups in **prod** `kipptaf_extracts`    | **9 (18 rows)** — identical |
| Region of every duplicate row                      | `kippnewark`, 0 Miami      |
| Files this branch changes under `parentsquare/`    | 0                          |

`rpt_parentsquare__parents` also excludes Miami explicitly in both of its source
CTEs (`_dbt_source_project != 'kippmiami'`, the same carve-out the
`rpt_clever__*` feeds make), so no Miami row this branch restores can reach it.

It surfaced here because merging `main` pulled in the recent
`int_finalsite__student_contacts` parent-selection work (#4768 and follow-ups),
which put this model back into CI's `state:modified+` selection. Prod carries the
same duplicates today; the test simply had not re-run against them.

No open issue tracks it. Flagged for a decision rather than fixed here, since the
defect is in the Newark contacts feed and is outside this PR's scope.

### The two failures that WERE this branch's, both now resolved

1. **Three `Not found: Table zz_stg_kippmiami_focus.*` errors** — the staging
   copies were missing every model #4725 shipped. Fixed by refreshing them.
   Worth recording: the kippmiami prod manifest was 7 days stale and did not
   contain those three models, and `dbt clone` silently skips any node absent
   from `--state`. Running the clone as-is would have reported success and fixed
   nothing. It needed `dbt deps` + `dbt parse --target prod` first.
2. **A finalsite `accepted_values` failure on `contact_slot`** (`contact_3`,
   `contact_4`) — already fixed on `main`, which this branch was 49 commits
   behind. Resolved by the merge.

All other checks pass: Trunk, CodeQL (actions / javascript / python),
`claude-review`.

Note that `claude-review` ran against the pre-merge commit and does not re-run on
`synchronize`, so its pass predates the `main` merge.

cbini:
The remaining `dbt Cloud` failure is now tracked as #4776 — root cause is the same person occupying both Finalsite parent slots for nine Newark students, which predates this branch and is present in prod today.

Picking it up next week. This PR stays open until then, since the failure is error severity and holds the merge gate.

Everything else on this PR is green and the two failures that *were* this branch's are resolved.

src/dbt/kipptaf/models/focus/intermediate/int_focus__schools_conformed.sql
cbini, line 1: this should be done in int_students__schools

src/dbt/kipptaf/models/focus/intermediate/int_focus__student_core_fields_conformed.sql
cbini, line 1: this should be done on int_students__student_core_fields

src/dbt/kipptaf/models/focus/intermediate/int_focus__student_user_fields_conformed.sql
cbini, line 1: this should be done on int_students__student_user_fields

src/dbt/kipptaf/models/focus/intermediate/int_focus__students_conformed.sql
```diff
@@ -0,0 +1,138 @@
...
+    --
+    -- ethnicity carries forward too, for a different reason: the archive is
+    -- internally inconsistent (Hispanic with no race flag maps to 'T' for 36
+    -- students and 'H' for 8), so no derivation can reproduce it. Carrying it
+    -- forward keeps dim_students.race byte-identical for returning students.
+    archive as (
+        select student_number, lunchstatus, ethnicity,
+        from {{ ref("stg_powerschool__students") }}
+        where _dbt_source_project = 'kippmiami'
+    )
```
cbini: move to int_students__students

src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__terms.sql
cbini, line 1: create an int_students__terms for this

src/dbt/kipptaf/models/focus/intermediate/int_focus__terms_conformed.sql
cbini, line 1: could this all be done in an int_students__terms?

src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__student_enrollment_union.sql
cbini, line 1: create int_students__student_enrollment_union for this

src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__terms.sql
cbini, line 1: create an int_students__terms for this

src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__teacher_grade_levels.sql
cbini, line 1: create an int_students__teacher_grade_levels for this

src/dbt/kipptaf/models/focus/intermediate/int_focus__student_enrollment_conformed.sql
```diff
@@ -0,0 +1,51 @@
...
+    -- The upstream student_number column is misnamed: it holds the PREFIXED
+    -- Focus id, not the network student number, so anyone joining on the column
+    -- name gets zero matches with no error. Unprefix it with the same rule
+    -- int_focus__students_conformed applies to the student spine.
+    --
+    -- _dbt_source_relation is deliberately NOT projected. The consuming union
+    -- generates its own, and carrying both would collide on the column name.
+    -- _dbt_source_project IS projected so the union can coalesce it through
+    -- rather than re-deriving 'kipptaf' from this model's own relation name.
+    identified as (
```
cbini: cte is superfluous

src/dbt/kipptaf/models/focus/intermediate/int_focus__schools_conformed.sql
cbini: Done. `int_focus__schools_conformed` is deleted and the conform logic now lives inline in `int_students__schools`.

src/dbt/kipptaf/models/focus/intermediate/int_focus__student_core_fields_conformed.sql
cbini: Done. `int_focus__student_core_fields_conformed` is deleted, logic inline in `int_students__student_core_fields`.

While relocating I found the `spedlep` / `lep_status` warn tests moved from a Miami-only model onto the network-wide union, which widened them from ~1,002 rows to 24,746 — they were counting every NJ student without an IEP. Scoped both to `_dbt_source_project = 'kippmiami'`, matching the `lunchstatus` test that already carried it.

src/dbt/kipptaf/models/focus/intermediate/int_focus__student_user_fields_conformed.sql
cbini: Done. `int_focus__student_user_fields_conformed` is deleted, logic inline in `int_students__student_user_fields`.

src/dbt/kipptaf/models/focus/intermediate/int_focus__students_conformed.sql
cbini: Done. The archive carry-forward CTE now lives in `int_students__students`, where both branches are already in scope. Verified zero ethnicity drift across all returning Miami students after the move.

src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__terms.sql
cbini: Done. Created `int_students__terms` and reverted `int_powerschool__terms` to `origin/main` byte-identical. Its six consumers now read the new model with a `term is not null` filter.

Flagging one thing, since my question to you understated it: the two terms models do not differ only by grain. `int_powerschool__terms` also resolves quarter dates through **termbins**, which disagree with the raw terms table for some schools. Merging both into one full-grain model therefore needed a full join plus a row_number pick for pre-2010 duplicate quarter records — 9 CTEs, not a union. Verified correct (Camden 268=268, Newark 480=480, Paterson 8=8, Miami +28 AY2026 quarters), but if you would have picked two models knowing that, say so and I will split it.

src/dbt/kipptaf/models/focus/intermediate/int_focus__terms_conformed.sql
cbini: Yes — it is all in `int_students__terms` now. `int_focus__terms_conformed` is deleted and its conform logic is inline there.

src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__terms.sql
cbini: Done. `stg_powerschool__terms` is reverted to `origin/main` byte-identical; its four consumers read `int_students__terms` at full grain.

src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__student_enrollment_union.sql
cbini: Done. Created `int_students__student_enrollment_union` and reverted `int_powerschool__student_enrollment_union` to `origin/main` byte-identical. All 13 mart consumers repointed, plus one singular test that referenced the old model directly.

src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__teacher_grade_levels.sql
cbini: Done. Created `int_students__teacher_grade_levels` and reverted `int_powerschool__teacher_grade_levels` to `origin/main` byte-identical. All 13 consumers repointed, including `int_people__staff_roster` (4,754 rows, stable across two builds).

src/dbt/kipptaf/models/focus/intermediate/int_focus__student_enrollment_conformed.sql
cbini: Right — it was a single CTE wrapping a plain select. That model is now deleted entirely; its logic lives in `int_students__student_enrollment_union` as CTEs that each do real work.

cbini:
## Review verdicts

Verified each finding against the code and against prod before acting. Four
confirmed, one partly. Details below, including the one where the diagnosis is
right but the proposed fix is wrong.

---

### 1. `district_student_identifier` null for Focus-migrated students — confirmed, fixed, but not the way suggested

The defect is real and I've fixed it. The proposed fix would have silently
published the wrong identifier.

`florida_student_number` is not the MDCPS student ID. Checked all 3,453
returning Miami students against the archive:

| Column | Shape | Rows equal to archive `state_studentnumber` |
| --- | --- | --- |
| `florida_student_number` | 10 digits | **0 of 3,305** |
| `florida_education_identifier` | 14 chars, `FL`-prefixed | 0 of 3,305 |
| `disis_id` | 6-7 digits | 3,304 of 3,305 (after zero-padding) |

These are three different Florida identifiers. `florida_education_identifier`
is the FLEID, already correctly wired to `state_student_identifier` via
`int_students__student_user_fields`. `florida_student_number` is a separate
10-digit FLDOE number that matches the MDCPS id on no student at all. Its zero
consumers are a fair observation — but the reason is that it is not the column
anything wants, not that a consumer is missing.

`disis_id` (Miami-Dade's DSIS) is the MDCPS id, off by a leading zero the
NUMERIC storage drops. But it is populated only for students the PowerSchool
migration carried over:

| Cohort | Students | `disis_id` populated |
| --- | --- | --- |
| Returning | 3,453 | 3,304 |
| Enrolled since the freeze | 506 | **0** |

So Focus has no live source for this id, and I carry it forward from the
archive instead, exactly like `lunchstatus` and `ethnicity` in the same CTE. I
built and then deleted a `disis_id` fallback once the coverage numbers came
back — it was dead code that added a padding rule with no rows to apply it to.

Verified in a local build: 3,436 of 4,451 Miami students now carry the id (the
Focus branch published zero before), and **0** returning students' values
changed against the archive.

New students get null, which joins the disclosed
`spedlep` / `lep_status` / `lunchstatus` gap. A confidently wrong district id
is worse than a null one.

### 2. Unprefix expression duplicated 5 times — confirmed, extracted

The count is right; my own first grep undercounted it at 2 because two sites
take `student_id` rather than `student_number`. Now one
`unprefix_focus_student_id()` macro in `kipptaf/macros/utils.sql`, called at
all 5 sites, with the `8400` rationale documented once instead of twice.

On the "cast early, once" half: the nesting is not removable. `student_id` is
INT64 on both sides and the prefix strip is textual, so the string round-trip
is inherent to the operation — a per-site CTE would relocate it, not remove it.
The macro is the reuse fix; I've noted the round-trip in the macro's own
comment.

### 3. Wrapper YAMLs duplicate tests and docs — confirmed, stripped

All four are pure `union_relations()` passthroughs, which is exactly the shape
`kipptaf/CLAUDE.md` says carries neither, and the precedent it points to
(`stg_focus__students.yml`, `int_focus__course_periods.yml`) sits in the same
directories. Confirmed each source YAML is fuller than the wrapper it feeds, so
nothing is lost: 174 vs 171 lines, 139 vs 75, 107 vs 95, 70 vs 46.

387 lines of duplicated docs and 4 redundant uniqueness tests removed.

### 4. Singular tests unregistered — partly confirmed, registered both

Registered, and the dagster-attachment argument for
`test_miami_students_spine_covers_archive` is sound — it refs two models, so
without the pin the result lands as an `AssetObservation` across both parents.
I also moved its `severity` out of an inline `config()` into the properties
entry, matching every other registered test.

One correction: "every other singular test has one — 29 of them" is not
accurate. There are 34 test files and 29 entries, and the 3 pre-existing
unregistered tests
(`fct_student_attendance_daily__enrollment_{,month_,week_}anchor_exists`) each
ref 2-3 models, so they are unregistered under the same conditions the finding
treats as disqualifying. The convention is real and worth following; it just
isn't currently at 100%, and the gap is not new to this PR.

### 5. Terms uniqueness covers only quarter rows — confirmed, second test added

Added a `(schoolid, yearid, abbreviation, _dbt_source_project)` uniqueness test
scoped `where term is null`, covering the year and semester rows the first test
skips. Verified against prod before adding it: 1,127 non-quarter keys across
all four districts, 0 duplicates, so it passes rather than papering over a
known failure.

Two tests rather than one because the halves key on different columns —
quarter rows key on `term` (which termbins supplies), year and semester rows on
`abbreviation` (straight off the raw row). A single test would have to key on a
column that is null for half the model.

---

### Separately: a dead `row_number()` in `int_students__terms`

Not from the review, and outside what it was asked to look at. The
`powerschool_raw_ranked` CTE dedupes what its comment calls "duplicate raw
quarter records ... in years before 2010", and I cannot reproduce that
condition. Across all four districts' `stg_powerschool__terms` in prod: 1,012
quarter keys and 1,127 non-quarter keys, every one of them singleton. With `rn`
always 1, the `and p.rn = 1` join predicate is a no-op.

The full-join carve-out two CTEs down does hold up — 86 termbins quarters have
no raw terms row, so a left join really would drop them.

I've left the `row_number()` in place rather than removing it in this PR, since
it is protective if the condition ever appears and the branch is otherwise
green. Happy to strip it in a follow-up if you'd rather not carry defensive
code the data doesn't justify.

src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
cbini, line 1: have historical student enrollments been imported into focus? if so, we can drop the filter on the focus cte and drop miami from the powerschool cte

src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
```diff
@@ -0,0 +1,100 @@
...
+    powerschool_with_project as (
+        -- trunk-ignore(sqlfluff/AM04): union_relations resolves columns at run time
+        select
+            *,
+
+            regexp_extract(_dbt_source_relation, r'(kipp\w+)_') as _dbt_source_project,
+
+            initcap(regexp_extract(_dbt_source_relation, r'kipp(\w+)_')) as region,
+        from powerschool_unioned
+    ),
```
cbini: this should be its own model

src/dbt/kipptaf/models/students/intermediate/int_students__schools.sql
```diff
@@ -0,0 +1,60 @@
...
+        select p.*,
+        from {{ ref("stg_powerschool__schools") }} as p
+        left join focus_schools as f on p.school_number = f.school_number
+        where p._dbt_source_project != 'kippmiami' or f.school_number is null
+    )
+
+select *,
+from powerschool_filtered
+
+full union all corresponding
```
cbini, line 44: what does this do differently from a bare "union all"?

src/dbt/kipptaf/models/students/intermediate/int_students__teacher_grade_levels.sql
```diff
@@ -0,0 +1,187 @@
...
+    ),
+
+    powerschool_with_project as (
+        -- trunk-ignore(sqlfluff/AM04): union_relations resolves columns at run time
+        select
+            *,
+
+            regexp_extract(_dbt_source_relation, r'(kipp\w+)_') as _dbt_source_project,
+        from powerschool_unioned
+    ),
```
cbini: this should be its own model

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
```diff
@@ -0,0 +1,209 @@
...
+            yearid,
+            term,
+            term_start_date,
+            term_end_date,
+            semester,
+            is_current_term,
+
+            regexp_extract(_dbt_source_relation, r'(kipp\w+)_') as _dbt_source_project,
+        from powerschool_quarters
+    ),
```
cbini: these should be their own models

src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
cbini: Yes, but not usably — I checked, and dropping either side would lose data.

Focus carries enrollment back to syear 2018. Two problems with treating it as the history of record:

**It is missing withdrawn students, and the gap widens toward the present.** Joined every Focus enrollment row to the archive on (student number, academic year):

| Academic year | Archive rows | In Focus | Archive-only | Focus-only |
| 2018 | 206 | 206 | 0 | 0 |
| 2021 | 1,659 | 1,659 | 0 | 0 |
| 2022 | 1,415 | 1,326 | 89 | 0 |
| 2023 | 1,517 | 1,324 | 193 | 0 |
| 2024 | 1,792 | 1,491 | 301 | 0 |
| 2025 | 1,956 | 1,536 | **420** | 0 |

Focus never has a row the archive lacks, so it adds nothing for closed years. Dropping Miami from the PowerSchool branch would delete 1,003 enrollment stints outright, most of them students who withdrew.

**The dates were not preserved.** For AY2025, only 563 of the 1,536 stints present in both carry the same `entrydate`. The older years survived the import mostly intact (198 of 206 in 2018), but the recent ones did not. `entrydate` feeds `student_enrollment_key`, so admitting those rows would recompose ~973 AY2025 keys and orphan every fact hanging off them.

So the AY2026 cutover stays. Worth revisiting if Ops backfills the missing withdrawals and reconciles the dates — at that point both halves of your suggestion become correct at once, which is the only way they work.

src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
cbini: It already is one, and I missed it — `int_powerschool__student_enrollment_union` at kipptaf level does exactly this union, including both the `_dbt_source_project` and `region` derivations. This branch had already reverted that model to `origin/main` byte-identical last round, and then I re-implemented its body inline here.

Fixed in 6333c4e65 — the CTE is now a `ref()` to it, and the two inline CTEs are gone. Same for the other two you flagged:

| Model | Now reads |
| --- | --- |
| `int_students__student_enrollment_union` | `int_powerschool__student_enrollment_union` |
| `int_students__teacher_grade_levels` | `int_powerschool__teacher_grade_levels` |
| `int_students__terms` | `stg_powerschool__terms` + `int_powerschool__terms` |

All four wrappers are materialized tables, so this also stops re-scanning the district relations once per consumer.

Verified behavior-preserving against the pre-refactor dev build: the PowerSchool branch lands on exactly the wrapper's 133,296 rows. (My earlier snapshot said 133,291 — the 5-row difference is upstream drift from rebuilding the wrapper, not the refactor.)

src/dbt/kipptaf/models/students/intermediate/int_students__schools.sql
cbini: It matches columns by **name** instead of by position, and tolerates each side having columns the other lacks.

A bare `union all` is positional: both sides need the same column count, in the same order, and it will happily line up `school_number` against `dcid` if the order drifts. `int_students__schools` outputs 53 columns, all from the PowerSchool side. The Focus branch projects 7 of them — `_dbt_source_relation`, `_dbt_source_project`, `name`, `location_key`, `abbreviation`, `school_number`, `school_level`. A bare `union all` would reject that outright.

The two modifiers do different jobs:

- `CORRESPONDING` — match by name, using only the columns **common** to both sides. Here that would silently narrow the model from 53 columns to 7.
- `FULL ... CORRESPONDING` — match by name over the **union** of both sides' columns, null-filling what a side omits. That is what lets the 7 Focus rows carry nulls for `dcid`, `portion`, and the other 46 PowerSchool-only columns while the archive rows keep their values.

`FULL` is the load-bearing word — plain `CORRESPONDING` compiles and builds clean, it just quietly drops 46 columns.

The practical benefit is that the conform CTEs only have to name the columns they can actually source. Without it every one of them would need a `cast(null as <type>)` line per unsourced column, which is where `dbt_utils.union_relations` spends most of its generated SQL.

src/dbt/kipptaf/models/students/intermediate/int_students__teacher_grade_levels.sql
cbini: It already exists — `int_powerschool__teacher_grade_levels` at kipptaf level does this exact union. Reverted to `origin/main` byte-identical last round, then re-implemented inline here.

Fixed in 6333c4e65: the two inline CTEs collapse to a `ref()`. It is a materialized table, so the union runs once rather than per consumer.

Row math checks out: 15,500 PowerSchool rows after the Miami cutover filter plus 135 Focus rows = 15,635 in the model.

One thing to note while you are in here — the model warns 16 rows on the `(teachernumber, academic_year, grade_level)` uniqueness test. All 16 come from the PowerSchool branch on its own, before any Focus rows are unioned in, so it is pre-existing rather than something the cutover introduced.

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
cbini: Both already exist — `stg_powerschool__terms` and `int_powerschool__terms` at kipptaf level, each doing the four-district union this was duplicating. Both were reverted to `origin/main` byte-identical last round, and I then re-implemented their bodies inline here.

Fixed in 6333c4e65: `powerschool_raw` / `powerschool_raw_with_project` collapse to a `ref("stg_powerschool__terms")`, and `powerschool_quarters` / `powerschool_quarters_with_project` to a `ref("int_powerschool__terms")` with the same 7-column projection. Both are materialized tables, so the union work now happens once and is shared.

Verified unchanged: 2,253 rows / 2,190 distinct keys, identical to the pre-refactor dev build.

Two things that fell out of this:

- Three `sqlfluff/AM04` ignores stopped suppressing anything once the `select *` read a `ref()` rather than an inline `union_relations` — sqlfluff can resolve columns through a ref. Trunk flagged them as `ignore-does-nothing`; removed.
- The `powerschool_raw_ranked` comment claimed duplicate raw quarter records "in years before 2010". That is not true — all 2,139 raw terms keys across the four districts are singletons in prod, so `p.rn = 1` never does anything. I corrected the comment to say it is defensive only rather than describing a condition that exists. Say the word and I will drop the `row_number()` too.

src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
cbini: Correcting myself — you were right to ask, and the coverage half of my answer above was wrong.

The "archive-only" rows I counted are almost entirely the alumni graduate placeholders: `enroll_status = 3`, null `entrydate` **and** `exitdate`, `schoolid = 999999`. Not withdrawals, not real stints.

| Academic year | Archive-only | Graduate placeholders | Real stints |
| --- | --- | --- | --- |
| 2022 | 89 | 88 | **1** |
| 2023 | 193 | 193 | 0 |
| 2024 | 301 | 301 | 0 |
| 2025 | 420 | 420 | 0 |

So it is 1 real stint, not 1,003. And the placeholders are already preserved by the third clause in `powerschool_conformed` (`or (enroll_status = 3 and entrydate is null)`), which is independent of the year filter — they would survive dropping Miami from the PowerSchool branch anyway.

Excluding placeholders, Focus's stint coverage matches the archive exactly:

| Academic year | Archive real stints | Focus stints | `entrydate` differs |
| --- | --- | --- | --- |
| 2018 | 206 | 206 | 8 |
| 2019 | 511 | 511 | 18 |
| 2020 | 723 | 723 | 16 |
| 2021 | 1,659 | 1,659 | **304** |
| 2022 | 1,327 | 1,326 | 54 |
| 2023 | 1,324 | 1,324 | 18 |
| 2024 | 1,491 | 1,491 | 30 |
| 2025 | 1,536 | 1,536 | **973** |

**The cutover still stands, but on one leg instead of two.** Coverage is not the problem — dates are. 1,421 of 8,776 historical stints carry a different `entrydate` in Focus, concentrated in AY2021 (18%) and AY2025 (63%). `entrydate` feeds `student_enrollment_key`, so admitting the Focus rows for closed years recomposes those 1,421 keys and orphans every fact row hanging off them across the 13 marts models that carry that key.

That is a narrower and more fixable objection than what I originally posted. If the AY2021 and AY2025 date drift gets reconciled, the remaining years are within ~3% and the cutover could move back to 2018 for everything except that one AY2022 stint. Worth a look at what happened to those two years specifically — the pattern (three clean years, one bad, three clean, one very bad) does not look like a systematic import rule.

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
```diff
@@ -0,0 +1,145 @@
...
+            disis_id,
+            florida_student_number,
+            florida_education_identifier,
+            race_white_label,
+            race_asian_label,
+            race_black_or_african_american_label,
+            race_american_indian_or_alaska_native_label,
+            race_native_hawaiian_or_other_pacific_islander_label,
+
+            {{ unprefix_focus_student_id("student_id") }} as student_number,
```
cbini: this should go on int_focus__students. unsure if the macro is warranted

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
```diff
@@ -0,0 +1,145 @@
...
+    ),
+
+    current_stint as (
+        {{
+            dbt_utils.deduplicate(
+                relation="enrollment_stints",
+                partition_by="student_number",
+                order_by="academic_year desc, startdate desc",
+            )
+        }}
```
cbini: add rn to int_focus__student_enrollments and filter on it in enrollment_stints

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
```diff
@@ -0,0 +1,145 @@
...
+            race_asian_label,
+            race_black_or_african_american_label,
+            race_american_indian_or_alaska_native_label,
+            race_native_hawaiian_or_other_pacific_islander_label,
+
+            {{ unprefix_focus_student_id("student_id") }} as student_number,
+
+            date(birthdate) as dob,
+
+            regexp_extract(sex_label, r'\[(\w+)\]') as gender,
```
cbini: move to the furthest upstream model available

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
```diff
@@ -0,0 +1,145 @@
...
+        select
+            *,
+
+            (
+                if(race_black_or_african_american_label = 'Yes', 1, 0)
+                + if(race_white_label = 'Yes', 1, 0)
+                + if(race_asian_label = 'Yes', 1, 0)
+                + if(race_american_indian_or_alaska_native_label = 'Yes', 1, 0)
+                + if(race_native_hawaiian_or_other_pacific_islander_label = 'Yes', 1, 0)
+            ) as race_count,
```
cbini: move to the furthest upstream model available

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
```diff
@@ -0,0 +1,145 @@
...
+            -- feed an economic-disadvantage proxy with no signal in it.
+            cast(null as string) as lunchstatus,
+
+            -- dim_students publishes this as district_student_identifier for
+            -- Miami. disis_id is the MDCPS id, stored as NUMERIC, which drops
+            -- the leading zero most of them carry -- pad back to the 7 digits
+            -- MDCPS issues. Populated only for students the PowerSchool
+            -- migration brought over, so it is null for anyone enrolled since,
+            -- until Ops fills it.
+            lpad(cast(r.disis_id as string), 7, '0') as state_studentnumber,
```
cbini: move to the furthest upstream model available

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
```diff
@@ -0,0 +1,145 @@
...
+                then 'B'
+                when r.race_white_label = 'Yes'
+                then 'W'
+                when r.race_asian_label = 'Yes'
+                then 'A'
+                when r.race_american_indian_or_alaska_native_label = 'Yes'
+                then 'I'
+                when r.race_native_hawaiian_or_other_pacific_islander_label = 'Yes'
+                then 'P'
+            end as ethnicity,
```
cbini: move to the furthest upstream model available

src/dbt/kipptaf/models/students/intermediate/int_students__student_core_fields.sql
```diff
@@ -0,0 +1,74 @@
...
+                then false
+                when 'LA'
+                then false
+                when 'LZ'
+                then false
+                when 'TZ'
+                then false
+                when 'ZZ'
+                then false
+            end as lep_status,
```
cbini: move to the furthest upstream model available

src/dbt/kipptaf/models/students/intermediate/int_students__schools.sql
```diff
@@ -0,0 +1,55 @@
...
+            -- closed/inactive schools.
+            case
+                s.school_level_label
+                when 'E - Elementary'
+                then 'ES'
+                when 'M - Middle'
+                then 'MS'
+                when 'H - High'
+                then 'HS'
+            end as school_level,
```
cbini: move to the furthest upstream model available

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
```diff
@@ -0,0 +1,177 @@
...
+            mp.start_date,
+            mp.end_date,
+            mp.syear as academic_year,
+
+            fs.schoolid,
+
+            if(mp.short_name in ('Q1', 'Q2'), 'S1', 'S2') as quarter_semester,
+
+            current_date('{{ var("local_timezone") }}')
+            between mp.start_date and mp.end_date as is_within_dates,
```
cbini: move to the furthest upstream model available

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
```diff
@@ -0,0 +1,177 @@
...
+            if(`type` = 'quarter', end_date, null) as term_end_date,
+            if(`type` = 'quarter', quarter_semester, null) as semester,
+            if(`type` = 'quarter', is_within_dates, null) as is_current_term,
+        from focus_marking_periods
+        -- Focus carries a full year/semester/quarter set for two schools in
+        -- every syear back to 1980, decades before any KIPP Miami school
+        -- existed -- roughly 760 template rows that would fabricate history.
+        -- Floor the branch at Miami's first real year, taken from the
+        -- enrollment history rather than hardcoded so it follows a backfill.
+        where academic_year >= (select ffy.academic_year, from focus_first_year as ffy)
```
cbini: move filter upstream. ok to hardcode 2018

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
```diff
@@ -0,0 +1,177 @@
...
+    -- (all 2,139 keys across the four districts are singletons in prod), so
+    -- this is defensive only.
+    powerschool_raw_ranked as (
+        select
+            *,
+
+            row_number() over (
+                partition by schoolid, yearid, abbreviation, _dbt_source_project
+                order by id
+            ) as rn,
```
cbini: move upstream

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
```diff
@@ -0,0 +1,177 @@
...
+            q.term_end_date,
+            q.semester,
+            q.is_current_term,
+
+            coalesce(p.schoolid, q.schoolid) as schoolid,
+            coalesce(p.yearid, q.yearid) as yearid,
+            coalesce(
+                p._dbt_source_project, q._dbt_source_project
+            ) as _dbt_source_project,
+            coalesce(p.academic_year, q.yearid + 1990) as academic_year,
```
cbini: refactor powerschool_quarters academic year calc upstream

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
cbini: Added `rn_all` to `int_focus__student_enrollment` — most recent stint per student across all years, ordered by academic year then exit then start date. `enrollment_stints` and the `dbt_utils.deduplicate` that followed it collapse into a single CTE filtering `rn_all = 1`.

Worth noting this was already duplicated: `rpt_google_directory__users_import` hand-rolls the same window under the same name. It can drop its copy now — I left that model alone since it is outside this PR, but it is a one-line change.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
cbini: Both — moved and deleted.

The unprefix now happens once in `int_focus__students`, which exposes `student_number` alongside the raw `student_id`. With that upstream, the macro had exactly one call site left, so it was not earning the indirection; I inlined the expression there and deleted `unprefix_focus_student_id`.

The one remaining site is `int_students__student_enrollment_union`, because `int_focus__student_enrollment.student_number` still holds the PREFIXED id despite its name. Fixing that properly means renaming it (`focus_student_id_prefixed` is the name the `rpt_focus__*` models already use) and repointing seven consumers — worth doing, but not something I wanted to fold into this pass. Say the word and it is a follow-up.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
cbini: Moved to `int_focus__students`, which now conforms every Focus field whose representation differs from the network's: `student_number`, `dob`, `gender`, `state_studentnumber`, `spedlep`, `gifted_and_talented`, `lep_status`, `race_count` and `ethnicity`. This model reads the conformed column.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
cbini: Moved to `int_focus__students`, which now conforms every Focus field whose representation differs from the network's: `student_number`, `dob`, `gender`, `state_studentnumber`, `spedlep`, `gifted_and_talented`, `lep_status`, `race_count` and `ethnicity`. This model reads the conformed column.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
cbini: Moved to `int_focus__students`, which now conforms every Focus field whose representation differs from the network's: `student_number`, `dob`, `gender`, `state_studentnumber`, `spedlep`, `gifted_and_talented`, `lep_status`, `race_count` and `ethnicity`. This model reads the conformed column.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__students.sql
cbini: Moved to `int_focus__students`, which now conforms every Focus field whose representation differs from the network's: `student_number`, `dob`, `gender`, `state_studentnumber`, `spedlep`, `gifted_and_talented`, `lep_status`, `race_count` and `ethnicity`. This model reads the conformed column.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__student_core_fields.sql
cbini: Moved to `int_focus__students`, which now conforms every Focus field whose representation differs from the network's: `student_number`, `dob`, `gender`, `state_studentnumber`, `spedlep`, `gifted_and_talented`, `lep_status`, `race_count` and `ethnicity`. This model reads the conformed column.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__schools.sql
cbini: Moved to `int_focus__schools` as `school_level`.

It collided with a column of the same name from `stg_focus__schools` — the raw Focus select-option id, which only the pivot consumes — so the intermediate now excludes the staging column and emits the decoded ES/MS/HS value under that name. That matches what `school_level` already means everywhere downstream.

This also removes a duplicate: `int_focus__student_enrollment` was deriving the identical CASE from `school_level_label`. It now reads the conformed column.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
cbini: Moved to a new `int_focus__marking_periods` in the focus package, exposed to kipptaf as a source plus `union_relations` wrapper.

It holds the progress-period drop, `quarter_semester`, `is_within_dates`, and the history floor. All that is left in this model is resolving Focus's internal school id to the network one, which needs the locations crosswalk and so cannot move upstream.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
cbini: Moved into `int_focus__marking_periods` and hardcoded at 2018, as you said.

That also drops the `focus_first_year` CTE and its scalar subquery — deriving the floor from `min(academic_year)` over the enrollment history was more machinery than the fact deserved.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
cbini: Moved to `stg_powerschool__terms`, which now emits `rn` alongside `_dbt_source_project`. `powerschool_raw_ranked` is gone; the full join reads the staging model directly.

Restating the evidence once since it now lives in a network-wide staging model: `rn` is never greater than 1. All 2,139 raw term keys across the four districts are singletons, in prod and in staging, so the guard has nothing to guard against today. I have kept it and labelled it defensive. If you would rather not carry it, deleting it is a smaller diff than the move was.

Done in 293dd4c42.

src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql
cbini: `int_powerschool__terms` already carries `academic_year` — the coalesce was recomputing `yearid + 1990` next to a column that already held it. `powerschool_quarters` now selects it and the coalesce reads `q.academic_year`.

I did briefly add the calc to the kipptaf wrapper before noticing the district model already exposes it; that is reverted, so the wrapper is untouched.

Done in 293dd4c42.

src/dbt/focus/models/intermediate/int_focus__marking_periods.sql
cbini, line 1: this model is unnecessary. move calcs and year filter upstream and type filter downstream

src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
```diff
@@ -0,0 +1,64 @@
...
+            state,
+
+            ps_schoolid as schoolid,
+            startdate as entrydate,
+            student_first_name as first_name,
+            student_last_name as last_name,
+
+            cast(
+                regexp_replace(cast(student_number as string), r'^8400', '') as int64
+            ) as student_number,
```
cbini: move upstream

src/dbt/focus/models/intermediate/int_focus__marking_periods.sql
cbini: Agreed, and deleted — the model, its kipptaf wrapper, and its source entry.

`quarter_semester` and `is_within_dates` are on `stg_focus__marking_periods` now, and the type filter is back down in `int_students__terms`.

**The year filter stayed downstream**, though, because moving it breaks a sibling. Flooring the staging model at 2018 orphans 321 `stg_focus__student_report_card_grades` rows that reference pre-2018 marking periods — that relationship has zero orphans in prod today, and the test caught it as soon as I built it. So those older periods are not purely template rows; grades hang off them.

Both filters therefore sit in `int_students__terms`, with 2018 hardcoded as you suggested. If you would rather the staging model carry the floor and the report-card-grades relationship get scoped instead, say so and I will flip it.

Done in f6f588ce8.

src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
cbini: Moved. `int_focus__student_enrollment` now derives `network_student_number`, and both `int_students__student_enrollment_union` and `int_students__students` read it.

I added a column rather than renaming `student_number` to the unprefixed form, which is what I would otherwise have done — that rename breaks account provisioning. `stg_people__student_logins` keys Miami students on the PREFIXED id: 3,955 of 3,955 match prefixed, 0 match unprefixed. `rpt_google_directory__users_import` joins Focus enrollments to it on `student_number`, and joins `finalsite_enrolled` on the same form (it casts `focus_student_id_prefixed` to get there). Renaming would have emptied both joins silently and dropped every Miami student out of Google account provisioning.

So both forms are exposed and documented on the model. The real cleanup is to move `stg_people__student_logins` and that extract onto the unprefixed id and then collapse the two columns — worth doing, but not two days before school starts.

Verified the unprefix is 1:1: 3,873 distinct ids in both forms, no nulls.

Done in f6f588ce8.

PR #4778, opened 2026-08-08
fix(dbt): exclude Miami from the Illuminate SFTP feeds

### LINKED ISSUE #4777: fix(dbt): exclude Miami from the Illuminate SFTP feeds
Miami is leaving Illuminate. The `kipptaf__extracts__illuminate__asset_job_schedule` schedule (cron `0 3 * * *` ET) is currently STOPPED and needs to be restarted for the 2026-27 year, but 6 of the 11 feed models still carry Miami rows and would push them into Illuminate on the first run.

## Current state, verified against prod

The 5 year-scoped student feeds are already Miami-free — the Focus cutover emptied Miami's 2026-27 PowerSchool rows. Paterson 2026-27 flows in with no code change (889 students, 3,934 non-dropped course enrollments, 140 sections, 4 sites, 135 courses).

The 6 feeds that still carry Miami:

| Feed model | Upstream | Why Miami survives | Miami rows |
| --- | --- | --- | --- |
| `rpt_illuminate__student_portal_accounts` | `int_extracts__student_enrollments` | filters `rn_all = 1 and enroll_status = 0`, no academic year | 1,114 students |
| `rpt_illuminate__users` | `int_people__staff_roster` + `int_people__temp_staff` | network-wide roster, no region filter | 185 active staff plus 5 temps |
| `rpt_illuminate__roles` | same | same | same population |
| `rpt_illuminate__terms` | `stg_powerschool__terms` | no academic year filter | 214 |
| `rpt_illuminate__courses` | `stg_powerschool__courses` | no academic year filter | 149 |
| `rpt_illuminate__sites` | `stg_powerschool__schools` | no academic year filter | 5 schools |

## Fix

Add a region exclusion to those 6 models only. The filter column differs per upstream:

- `stg_powerschool__{schools,terms,courses}` expose `_dbt_source_project`
- `int_extracts__student_enrollments` exposes `region`
- `int_people__staff_roster` exposes `home_work_location_dagster_code_location`
- `int_people__temp_staff` exposes `dagster_code_location`

The other 5 models get no filter — they are Miami-free by data, and a dead predicate on a year-scoped feed is noise.

## Sequencing

Merge and let prod rematerialize BEFORE restarting the schedule, or the 3am run pushes the Miami population above.

## Out of scope, worth tracking separately

Those 1,114 Miami students still read `enroll_status = 0` in PowerSchool roughly a year after the Focus cutover — nobody withdrew them. This filter stops them reaching Illuminate; it does not fix the stale PowerSchool records.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will drop every Miami student, staff member, site, term, and course from the 11-file Illuminate SFTP feed, so the feed can be restarted for 2026-27 without leaking a region that has left the platform.

`kipptaf__extracts__illuminate__asset_job_schedule` (cron `0 3 * * *` ET) is currently STOPPED. **Merge this and let prod rematerialize BEFORE restarting it** — otherwise the first 3am run pushes 1,114 Miami students and 477 Miami staff rows into Illuminate.

Six of the 11 feed models carried Miami rows: the five with no academic-year filter, plus the two staff feeds, which read the network-wide roster.

### Verified deltas

Prod view vs a dev build of this branch, both reading prod upstreams in the same query:

| Feed | prod | this branch | removed | reconciles to |
| `student_portal_accounts` | 10,963 | 9,849 | 1,114 | Miami students still reading `enroll_status = 0` |
| `users` | 4,780 | 4,303 | 477 | 469 Miami roster rows (all statuses) plus 5 Miami temps plus 3 NULL-location records |
| `roles` | 2,219 | 1,964 | 255 | 41 campus-based plus 129 school-based plus 5 temps plus 80 (40 CMO staff at 2 Miami sites) |
| `terms` | 2,139 | 1,925 | 214 | every Miami term row |
| `courses` | 1,537 | 1,443 | 94 | 94 of 149 Miami course tuples are Miami-only; 55 are shared with NJ and correctly survive `select distinct` |
| `sites` | 21 | 19 | 2 | the 2 Miami schools that pass `state_excludefromreporting = 0` |

Every number reconciles exactly against an independent count of the Miami population. `dbt build --target dev --defer --favor-state` passes on all 6 models, including both contract enforcements and both uniqueness tests.

### The NULL-location rows in `users` are excluded on purpose

Three roster rows carry a NULL `home_work_location_dagster_code_location`. A plain `!=` excludes them along with Miami, because `NULL != 'kippmiami'` evaluates to NULL rather than true — and that exclusion is **desired**: a staff row with no work location does not belong in the feed.

That behavior is a consequence of three-valued logic rather than something the SQL states outright, which makes it easy to misread as an oversight. An earlier revision of this PR did exactly that and "fixed" it to `is distinct from`, which wrongly re-admitted all three. The `users` model now carries a comment saying the exclusion is deliberate, so the next reader does not repeat it. Every predicate here uses the same `!=` idiom as the roughly 45 other region-exclusion sites across the project.

For the record, all three are ADP data-entry gaps rather than anything this feed should carry: probed against the ADP API, all three return HTTP 200 (not deleted), and all three show `homeOrganizationalUnits: []` with `homeWorkLocation: null` — including at `asOfDate` values mid-tenure while their assignments were still Active. One is a 3-day test record from 2021; the other two never had a location set at any point. Worth a People Operations follow-up on its own, not here.

In `roles`, the CMO branch filters the **site** side only. Those staff are network-wide, so they keep roles at every remaining site; filtering the roster side would have stripped Miami-seated central staff from all 19 sites.

### Models deliberately NOT changed

`roster`, `enrollment`, `studemo`, `programs`, and `mastschd` all filter `current_school_year` with no region restriction, and prod confirms zero Miami rows for 2026-27 — the Focus cutover emptied Miami's PowerSchool data for the year. Adding a dead predicate to a year-scoped feed is noise, so they are untouched.

The same absence of a region filter is why **Paterson 2026-27 needs no code at all**: 889 students, 3,934 non-dropped course enrollments, 140 sections, 4 sites, and 135 courses already flow through.

### Out of scope

The 1,114 Miami students still read `enroll_status = 0` in PowerSchool roughly a year after the Focus cutover — nobody withdrew them. This PR stops them reaching Illuminate; it does not fix the stale PowerSchool records. Worth a separate Ops ticket.

## AI Assistance

Claude Code investigated and authored this change end to end, human-directed. Two course corrections came from the human review during the session: the scoping decision to filter only the 6 feeds that actually carry Miami rather than all 11, and the correction that NULL-location exclusion is intended behavior rather than a bug to fix.

### General

### dbt

- [x] No new models, so no new properties files needed; no column changes, so no contract or exposure impact.
- [x] **Breaking change?** No. Filters only, no renames or removals — every contracted column set is unchanged.
- [x] No external sources added or modified.

## CI checks

- [x] **Trunk** — clean on all 6 files.
- [x] **dbt Cloud** — passed with zero warnings on the first commit; re-running for the second.

Refs #4777

### DIFF
--- a/src/dbt/kipptaf/models/extracts/illuminate/rpt_illuminate__courses.sql
+++ b/src/dbt/kipptaf/models/extracts/illuminate/rpt_illuminate__courses.sql
@@ -27,4 +27,4 @@ select distinct
 -- trunk-ignore-end(sqlfluff/RF05)
 from {{ ref("stg_powerschool__courses") }}
 -- Miami left Illuminate ahead of AY2026-27
-where _dbt_source_project is distinct from 'kippmiami'
+where _dbt_source_project != 'kippmiami'
--- a/src/dbt/kipptaf/models/extracts/illuminate/rpt_illuminate__courses.sql
+++ b/src/dbt/kipptaf/models/extracts/illuminate/rpt_illuminate__courses.sql
@@ -26,3 +26,5 @@ select distinct
     null as `24 Tech Prep`,
 -- trunk-ignore-end(sqlfluff/RF05)
 from {{ ref("stg_powerschool__courses") }}
+-- Miami left Illuminate ahead of AY2026-27
+where _dbt_source_project is distinct from 'kippmiami'
--- a/src/dbt/kipptaf/models/extracts/illuminate/rpt_illuminate__roles.sql
+++ b/src/dbt/kipptaf/models/extracts/illuminate/rpt_illuminate__roles.sql
@@ -17,7 +17,11 @@ select
 -- trunk-ignore-end(sqlfluff/RF05)
 from {{ ref("int_people__staff_roster") }} as sr
 inner join
-    {{ ref("stg_powerschool__schools") }} as sch on sch.state_excludefromreporting = 0
+    {{ ref("stg_powerschool__schools") }} as sch
+    on sch.state_excludefromreporting = 0
+    -- Miami left Illuminate ahead of AY2026-27. Filter the SITE side only: CMO
+    -- staff are network-wide, so they keep roles at every remaining site.
+    and sch._dbt_source_project is distinct from 'kippmiami'
 where
     sr.worker_status_code != 'Terminated'
     and sr.home_department_name in ('Teaching and Learning', 'Data', 'Executive')
@@ -51,6 +55,8 @@ where
     sr.worker_status_code != 'Terminated'
     and sr.home_department_name not in ('Teaching and Learning', 'Data', 'Executive')
     and sr.home_work_location_is_campus
+    -- Miami left Illuminate ahead of AY2026-27
+    and sr.home_work_location_dagster_code_location is distinct from 'kippmiami'
 
 union all
 
@@ -75,6 +81,8 @@ where
     worker_status_code != 'Terminated'
     and home_department_name not in ('Teaching and Learning', 'Data', 'Executive')
     and not home_work_location_is_campus
+    -- Miami left Illuminate ahead of AY2026-27
+    and home_work_location_dagster_code_location is distinct from 'kippmiami'
 
 union all
 
@@ -95,3 +103,5 @@ select
     1 as `05 Session Type ID`,
 -- trunk-ignore-end(sqlfluff/RF05)
 from {{ ref("int_people__temp_staff") }}
+-- Miami left Illuminate ahead of AY2026-27
+where dagster_code_location is distinct from 'kippmiami'
--- a/src/dbt/kipptaf/models/extracts/illuminate/rpt_illuminate__roles.sql
+++ b/src/dbt/kipptaf/models/extracts/illuminate/rpt_illuminate__roles.sql
@@ -21,7 +21,7 @@ inner join
     on sch.state_excludefromreporting = 0
     -- Miami left Illuminate ahead of AY2026-27. Filter the SITE side only: CMO
     -- staff are network-wide, so they keep roles at every remaining site.
-    and sch._dbt_source_project is distinct from 'kippmiami'
+    and sch._dbt_source_project != 'kippmiami'
 where
     sr.worker_status_code != 'Terminated'
     and sr.home_department_name in ('Teaching and Learning', 'Data', 'Executive')
@@ -56,7 +56,7 @@ where
     and sr.home_department_name not in ('Teaching and Learning', 'Data', 'Executive')
     and sr.home_work_location_is_campus
     -- Miami left Illuminate ahead of AY2026-27
-    and sr.home_work_location_dagster_code_location is distinct from 'kippmiami'
+    and sr.home_work_location_dagster_code_location != 'kippmiami'
 
 union all
 
@@ -82,7 +82,7 @@ where
     and home_department_name not in ('Teaching and Learning', 'Data', 'Executive')
     and not home_work_location_is_campus
     -- Miami left Illuminate ahead of AY2026-27
-    and home_work_location_dagster_code_location is distinct from 'kippmiami'
+    and home_work_location_dagster_code_location != 'kippmiami'
 
 union all
 
@@ -104,4 +104,4 @@ select
 -- trunk-ignore-end(sqlfluff/RF05)
 from {{ ref("int_people__temp_staff") }}
 -- Miami left Illuminate ahead of AY2026-27
-where dagster_code_location is distinct from 'kippmiami'
+where dagster_code_location != 'kippmiami'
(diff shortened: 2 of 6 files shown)


cbini:
Verified the review's findings against the code and the warehouse. Summary: the `rpt_parentsquare__staff.sql` citation is accurate, but the risk assessment is wrong in a way that matters — and the pattern is much wider than one file.

## The `!=` finding is real as a pattern, but drops zero rows today

`rpt_parentsquare__staff.sql:55` does use `!=` on `home_work_location_dagster_code_location`, confirmed. But it cannot drop anyone, now or under the current data shape, because the location fields co-vary. Every roster row with a NULL code location also has:

| field | value on all 3 NULL-location rows |
| --- | --- |
| `home_work_location_powerschool_school_id` | NULL |
| `home_department_name` | NULL |
| `worker_status_code` | Terminated |

Measured, not inferred — all four counts are 0:

- rows with NULL code location and non-NULL `home_work_location_powerschool_school_id`: **0**
- rows with NULL code location and non-NULL `home_department_name`: **0**
- rows with NULL code location and status other than Terminated: **0**
- rows matching the ops-leader shape (`home_work_location_powerschool_school_id = 0`, `home_department_name = 'Operations'`, not Terminated) with a NULL code location: **0**

So in `rpt_parentsquare__staff` the `!=` is masked twice over: the same WHERE also requires `home_department_name = 'Operations'` and `home_work_location_powerschool_school_id = 0`, and both are NULL on every NULL-location row. Those predicates exclude the rows before the region filter is reached. The suggested verification query returns zero rows.

## The scope is ~45 sites, not one

`rg "!= 'kippmiami'|!= 'Miami'"` across `src/dbt/` returns roughly 45 hits. Four are on this same nullable roster column:

- `rpt_parentsquare__staff.sql:55`
- `rpt_clever__teachers.sql:27`
- `rpt_clever__sections.sql:19`
- `rpt_clever__staff.sql:22`

All four are masked the same way: each co-filters on `worker_status_code != 'Terminated'` (or `assignment_status != 'Terminated'`), and every NULL-location row is Terminated. `rpt_clever__teachers.sql` additionally requires `home_work_location_powerschool_school_id is not null`. The remaining ~41 sites filter `_dbt_source_project` or `region`, both of which measure 0 NULLs.

That makes this a latent consistency issue rather than a live bug: correct today by coincidence of co-filters, and it would start dropping people the moment a NULL-location worker is Active. Worth a follow-up sweep, and I agree it does not belong in this PR.

## On the suggested regression test

Lukewarm, for a mechanical reason. These 6 models are views (the `extracts/` default), and per `src/dbt/CLAUDE.md` a test re-runs only when its host model materializes, while the data-change automation condition re-materializes table models, not views. A test anchored to these views would therefore fire on code changes and little else — it would not catch a Miami row reappearing because upstream data shifted, which is the failure mode worth guarding. Anchoring it to a table-materialized upstream would work, but that is a different and larger design question.

## Correction to the PR body I am carrying over

Two counts in the linked issue were wrong and are corrected in a comment there: `sites` removes 2 rows, not 5 (the model already filters `state_excludefromreporting = 0`), and `users` removes 474, not 190 (that feed does not filter terminated staff).

## Not verified by the review, verified here

The review noted it could not confirm the row-count reconciliation without warehouse access. Those numbers come from counting the prod views against a `--target dev --defer --favor-state` build of this branch in a single query, so both sides read the same prod upstreams at the same instant. Each delta was then decomposed against an independent count of the Miami population, which is how the original `!=` bug surfaced: the `users` delta came out at 477 against a Miami population of 474.

cbini:
Correcting my previous comment: the premise was wrong, and the recommendation in it is withdrawn.

Excluding roster rows with a NULL `home_work_location_dagster_code_location` is **desired behavior**, not a latent bug. A staff row with no work location does not belong in these feeds. The plain `!=` is doing the right thing at all four roster-column sites I listed, and at the ~41 other region-exclusion sites, so no sweep to `is distinct from` should happen — it would wrongly re-admit those rows.

This PR has been updated accordingly: all 8 predicates are back to `!=`, and `users` now removes 477 rows rather than 474 (469 Miami roster rows, 5 Miami temps, 3 NULL-location records). `roles` is unchanged at 255, since it already filters terminated staff and all 3 NULL-location rows are terminated.

What survives from the review's finding is narrower but still real: the NULL exclusion is implicit in three-valued logic rather than stated in the SQL, which is why both the review and I read it as a defect. `rpt_illuminate__users.sql` now carries a comment marking the exclusion deliberate. The sibling models (`rpt_parentsquare__staff`, `rpt_clever__teachers`/`__sections`/`__staff`) have no such comment and will likely attract the same false-positive finding again — a comment-only follow-up would be worth it, but that is documentation, not a correctness fix.

The measured counts in my previous comment stand and are what make the behavior safe to reason about: every NULL-location roster row is also NULL department, NULL `home_work_location_powerschool_school_id`, and Terminated.

Issue #4776, opened 2026-08-07
fix(dbt): rpt_parentsquare__parents sends nine Newark parents twice


cbini:
`rpt_parentsquare__parents` emits two rows for nine Newark students, failing its
uniqueness test at error severity:

```text
dbt_utils_unique_combination_of_columns_rpt_parentsquare__parents_student_id__first_name__last_name
Got 9 results, configured to fail if != 0
```

Present in prod today — `kipptaf_extracts.rpt_parentsquare__parents` carries the
same 9 groups / 18 rows. This is an outbound delivery path, so nine Newark
parents would reach ParentSquare twice, either creating duplicate parent accounts
or getting the file rejected depending on how the import keys rows.

## Root cause

The extract joins contacts to enrollments and emits one row per filled parent
slot:

```sql
from contacts as c
inner join students as s
    on c.student_number = s.student_number
    and c.code_location = s.code_location
```

`contacts` is filtered to `contact_slot in ('contact_1', 'contact_2')`, the two
Finalsite Household 1 parent slots.

Both sides checked:

- **Enrollment side is clean.** All nine students have exactly one enrollment row
  under the extract's filters (AY2026, `rn_year = 1`, in-district,
  `enroll_status in (0, -1)`). No fan-out there.
- **Contacts side is the cause.** For all nine students,
  `int_students__contacts` holds **the same person in both `contact_1` and
  `contact_2`** — nine rows in each slot, nine distinct students, matching names.
  Each student therefore emits that one adult twice.

## Two variants, and why a name-only dedupe is wrong

Of the 18 rows, only **10 are distinct across every column**:

| Variant                                                    | Students | Shape                                      |
| ---------------------------------------------------------- | -------- | ------------------------------------------ |
| Byte-identical pair                                        | 8        | same relationship, mobile, secondary, email |
| Same name, different contact details                       | 1        | differing phone and email between the slots |

The second variant is why `select distinct` or a dedupe on
`(student_id, first_name, last_name)` is not a safe fix — it would silently drop
a real phone number. The right fix likely belongs upstream, in whatever assigns a
person to a parent slot, not in the extract.

## Suspected origin, not yet confirmed

The uniqueness test is old; it shipped with the original Newark feed in
`294a7cbe6`, long before the current data. So either the duplicates are new, or
the test had not re-run against them.

Prime suspect is the recent parent-selection work — `979c5d982`
`feat(finalsite): rank parent contacts densely and guard against student
[...]

PR #2831, opened 2025-08-08
Add weighted ada

GabyRangelB:
# Pull Request

## Summary & Motivation

[//]: # "When merged, this pull request will..."

## Self-review

**If this is a same-day request, please flag that in our Slack channel!**

- [ ] Update **due date**, **assignee**, and **priority** on our
      [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
- [ ] <kbd>Format</kbd> has been run on all modified files

### SQL

- [ ] Ensure you are using the `union_dataset_join_clause()` macro for queries
      that employ any models using these datasets: `deanslist` `edplan` `iready`
      `overgrad` `pearson` `powerschool` `renlearn` `titan`
- [ ] All `distinct` usage must be accompanied by an comment explaining it's
      necessity
- [ ] If you are adding a new external source, run:

      dbt run-operation stage_external_sources --vars
      "ext_full_refresh: true"

## Troubleshooting

- [SqlFluff Rules Reference](https://docs.sqlfluff.com/en/stable/rules.html)
- [Trunk](https://teamschools.github.io/teamster/CONTRIBUTING/#trunk)
- [dbt](https://teamschools.github.io/teamster/CONTRIBUTING/#dbt-cloud_1)

PR #4785, opened 2026-08-10
fix(dbt): correct two weighted ADA calculation bugs in ada_term and its pivot

### LINKED ISSUE #4783: fix(dbt): correct two weighted ADA calculation bugs in int_powerschool__ada_term and its pivot
## Summary

Two independent bugs make weighted ADA wrong. Both originate in PR #2831 (commit `b4bcfc8ec`, 2025-08-08). Neither produces a visible symptom, which is why they survived a year: bug 1 shows up only as a threshold that never fires, and bug 2 returns a perfectly plausible rate that happens to be the wrong one.

Scope is narrow — see Impact. High-school ADA in Newark and Camden is affected; middle and elementary, and the Tableau attendance reports, are not.

## Bug 1 - `ada_weighted_term` is a raw sum, not a rate

`src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__ada_term.sql` line 79 omits the `safe_divide` that all three sibling weighted columns apply.

```sql
-- current
round(sum_attendance_value_weighted_term, 3) as ada_weighted_term,

-- fix
round(
    safe_divide(sum_attendance_value_weighted_term, count_attendance_value_term), 3
) as ada_weighted_term,
```

Evidence, `academic_year = 2025`, 42,731 rows: range 0 to 50, average 38.53, with 42,533 rows above 1. Applying the divide yields average 0.870 and zero out-of-range rows.

`count_attendance_value_term` is the correct denominator. The CTE filters `membershipvalue = 1`, so it equals the number of in-session days, and it is the same denominator every unweighted column already uses.

### Field audit of int_powerschool__ada_term

`ada_weighted_term` is the only broken field in this model.

| Field | Min | Max | Avg | Verdict |
| `ada_term` | 0 | 1 | 0.9196 | OK |
| `ada_semester` | 0 | 1 | 0.9194 | OK |
| `ada_year` | 0 | 1 | 0.9201 | OK |
| `ada_year_running` | 0 | 1 | 0.9290 | OK |
| `ada_weighted_term` | 0 | 50 | 38.53 | WRONG |
| `ada_weighted_semester` | 0 | 1 | 0.8696 | OK |
| `ada_weighted_year` | 0 | 1 | 0.8706 | OK |
| `ada_weighted_year_running` | 0 | 1 | 0.8817 | OK |

Weighted sits about 0.05 below unweighted at every grain, which is the expected direction. The underlying `is_present_weighted` is fine; this is purely a missing denominator.

## Bug 2 - the pivot rebuilds `ada_weighted_year` from unweighted columns

`src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__ada_term_pivot.sql` derives `ada_weighted_year` from the unweighted `ada_year_q*` columns.

```sql
-- current
round(coalesce(ada_year_q4, ada_year_q3, ada_year_q2, ada_year_q1), 3) as ada_weighted_year,

-- fix
round(
    coalesce(
        ada_weighted_year_q4,
        ada_weighted_year_q3,
        ada_weighted_year_q2,
        ada_weighted_year_q1
    ),
    3
) as ada_weighted_year,
```

`int_powerschool__ada_term` already computes `ada_weighted_year` correctly: 38,905 of 42,731 rows differ from unweighted, average 0.8706 versus 0.9201. The pivot discards that correct value and re-derives it, so in the pivot the two columns are identical across all 11,217 rows. Downstream consumers read the pivot, not the source.

## Impact

Measured on AY2025 (`rn_year = 1`, `enroll_status = 0`) by comparing a local dev build against production.

**Affected: high-school students in Newark and Camden only.** Middle and elementary read `ada_year`, which neither bug touched. Miami and Paterson have no high-school grades, so no rows qualify.

| Region | HS students | avg `ada` now | avg `ada` after | `ada` values changed | Chronic-absence flips |
| Newark | 995 | 0.8969 | 0.8380 | 975 | 257 |
| Camden | 380 | 0.8616 | 0.8252 | 343 | 42 |

The changed-value counts exclude rows that differ only by rounding. Both builds round ADA to 3 decimals, and a distributed float `sum()` has no guaranteed association order, so a rate sitting on a rounding midpoint can shift by 0.001 between evaluations independently of this fix — 6 high-school rows fall in that category. For the same reason the flip counts are accurate to roughly plus or minus 2 rather than exact.

### What changes

**The band-cased `ada` column, high school only.** 1,318 students get a different value, averaging about 5 points lower. That is the correct direction — partial-day presence now earns partial credit instead of silently falling back to the unweighted rate. Reports that display `ada` will show different high-school figures, including `rpt_tableau__hs_early_warning_dashboard`, `rpt_tableau__mtss_rti`, `rpt_gsheets__mtss_rti`, `rpt_tableau__gradebook_dashboard`, `rpt_tableau__gradebook_gpa`, `rpt_tableau__student_course_grades`, `rpt_tableau__gpa_cumulative_year`, `rpt_gsheets__grad_plan_tracking`, and the `int_tableau__gradebook_audit_*` chain.

**Chronic-absence classification, 299 students.** Two decisions threshold the band-cased column at 0.90:

- `rpt_tableau__attendance_interventions:22` — `if(co.ada <= .90, true, false) as is_chronic_absence`
- `rpt_deanslist__designations:28` — `if(co.ada <= 0.9, 'Chronic Absence', null)`

257 Newark and 42 Camden high schoolers change chronic-absence status. This is the change worth mentioning to stakeholders. The flip boundary matches the consumers' own test (`ada <= .90`), so a student landing exactly on 0.90 is not counted — 51 HS students sit exactly there and are chronically absent in both builds.

**Athletic eligibility.** `int_students__athletic_eligibility` gates high-school Q2 on `cy_weighted_term_q1` at 0.9. On the broken scale that gate passed 99.5%; corrected, 42.9% pass. That is a change to a sub-condition, not directly to final status — Q2 status also depends on GPA and credits, so the count of students whose status string changes is smaller and is not measured here. Qualitatively, `Probation - ADA` becomes reachable for high school in Q2, where it previously was not.

### What does not change

- **Both 0.80 flags are unaffected, 0 flips, verified against a dev build.** `ada_above_or_at_80`'s third CASE branch is unguarded by `school_level`, so a high-school student failing the weighted test at 0.80 falls through to `when ada.ada_year >= 0.80 then true`. `is_ada_above_or_at_80_cum_gpa_less_2` likewise ORs with `ada.ada_year >= 0.80`. Weighted never exceeds unweighted (0 rows), so the weighted branch cannot alter either result. Everything consuming `ada_above_or_at_80` is therefore untouched, including the gradebook-audit scaffold filter at `int_tableau__gradebook_audit_student_scaffold:164`.
- **The Tableau attendance reports are unaffected.** `rpt_tableau__attendance_comm_log`, `rpt_tableau__okrts_referrals` (including `is_chronically_absent`), `int_topline__attendance_contacts`, and `rpt_gsheets__kippmiami_payout_roster` all read `unweighted_ada`, not a weighted column. Note that a grep for `weighted_ada` matches `unweighted_ada` as a substring, which makes this easy to over-count.
- **`ada_weighted_term_q2`, `q3`, and `q4` have no consumers.** Only `q1` is read anywhere.

No columns are renamed or removed, so no exposure breaks and no coordination is needed to avoid a hard failure.

### Worth knowing

The band-casing at `int_extracts__student_enrollments:299` keys only on `school_level = 'HS' and academic_year >= 2025`, with no region test. Paterson currently tops out at grade 7 and adds a grade per year, so when it reaches grade 9 its high schoolers will begin receiving weighted ADA automatically rather than by decision. Miami is not a consideration — it has no high school and is no longer on PowerSchool.

## Preventive tests

No ADA column asserts its range today. Every one of them is a rate, so add to both models:

```yaml
data_tests:
  - dbt_utils.expression_is_true:
      arguments:
        expression: between 0 and 1
      config:
        severity: error
```

Bug 1 would have failed on the first build with this in place. Bug 2 needs a different guard, since a copied column is still in range. The durable protection there is for the pivot to stop re-deriving a value the source already computes correctly.

## Reproduce

Symptom, both bugs, aggregate only:

```sql
with
base as (
  select
    [
      struct('1. ada_term_q1 (unweighted, control)' as column_name, ada_term_q1 as value),
      struct('2. ada_weighted_term_q1 - BUG 1', ada_weighted_term_q1),
      struct('3. ada_semester_s1 (unweighted, control)', ada_semester_s1),
      struct('4. ada_weighted_semester_s1 (correct)', ada_weighted_semester_s1),
      struct('5. ada_year (unweighted, control)', ada_year),
      struct('6. ada_weighted_year - BUG 2, identical to 5', ada_weighted_year)
    ] as cols
  from `teamster-332318.kipptaf_powerschool.int_powerschool__ada_term_pivot`
  where academic_year = 2025
)

select
  c.column_name,
  count(c.value) as populated_rows,
  round(min(c.value), 3) as min_value,
  round(max(c.value), 3) as max_value,
  round(avg(c.value), 3) as avg_value,
  countif(c.value > 1) as rows_above_1_impossible_for_a_rate,
  countif(c.value >= 0.9) as rows_passing_the_0_9_gate
from base
cross join unnest(base.cols) as c
group by c.column_name
order by c.column_name
```

Bug 1 fix verification:

```sql
select
  academic_year,
  round(avg(ada_weighted_term), 3) as avg_as_shipped,
  round(
    avg(safe_divide(sum_attendance_value_weighted_term, count_attendance_value_term)), 3
  ) as avg_if_divided,
  countif(ada_weighted_term > 1) as impossible_rows_as_shipped,
  countif(
    safe_divide(sum_attendance_value_weighted_term, count_attendance_value_term) > 1
  ) as impossible_rows_if_divided
from `teamster-332318.kipptaf_powerschool.int_powerschool__ada_term`
where academic_year = 2025
group by academic_year
```

Bug 2, source correct versus pivot broken:

```sql
select
  'int_powerschool__ada_term (source)' as model,
  countif(ada_year != ada_weighted_year) as year_cols_differ,
  countif(ada_year = ada_weighted_year) as year_cols_identical
from `teamster-332318.kipptaf_powerschool.int_powerschool__ada_term`
where academic_year = 2025 and ada_year is not null and ada_weighted_year is not null

union all

select
  'int_powerschool__ada_term_pivot (consumed downstream)',
  countif(ada_year != ada_weighted_year),
  countif(ada_year = ada_weighted_year)
from `teamster-332318.kipptaf_powerschool.int_powerschool__ada_term_pivot`
where academic_year = 2025 and ada_year is not null and ada_weighted_year is not null
```

## Prior attempt

Bug 1 was diagnosed and fixed by @anthonygwalters in #3147, which was closed without merging and whose branch has since been deleted. Bug 2 was never addressed. PR #4785 supersedes that work.

## Scope

Two one-line corrections plus range tests. Explicitly excludes the athletic eligibility model refactor, which is tracked separately.

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will correct two weighted-ADA calculation bugs, both introduced in #2831 (2025-08-08), fix the non-reproducible rounding that made both of them awkward to measure, and add range tests so none of these classes of error can recur silently.

**Bug 1 — `ada_weighted_term` was a raw sum, not a rate.** [int_powerschool__ada_term.sql:79](https://github.com/TEAMSchools/teamster/blob/main/src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__ada_term.sql#L79) omitted the `safe_divide` that its three sibling weighted columns apply, so it emitted values on a 0-to-50 scale. 42,533 of 42,731 rows exceeded 1.

**Bug 2 — the pivot rebuilt `ada_weighted_year` from unweighted columns.** `int_powerschool__ada_term_pivot` coalesced `ada_year_q*` instead of `ada_weighted_year_q*`, discarding the correctly computed value from `int_powerschool__ada_term`. The two columns were identical in all 11,217 rows, which made the SY26 high-school weighted-ADA rule in `int_extracts__student_enrollments` a no-op from the day it was written.

**Bug 3 — ADA rounding was not reproducible between builds.** `int_powerschool__ada_term` summed `FLOAT64` attendance values, and a distributed float `sum()` has no guaranteed association order, so a rate landing on a 3-decimal rounding midpoint rounded up in one evaluation and down in the next. This is the artifact the impact measurements below had to work around, and it is now fixed rather than tolerated: the three daily values are cast to `NUMERIC` up front so every aggregate and window sum is exact and order-independent, then each rounded rate is cast back to `FLOAT64` on projection. Output types are unchanged, so the 15 contracted downstream models that declare these columns `FLOAT64` are untouched.

Measured across all years, **2,179 of 74,644 student-years sit exactly on a 3-decimal midpoint** for `ada_weighted_year` — the population whose rounded value was decided by float noise. In a single measurement, 213 of them disagree between the float and exact paths (16 for the unweighted `ada_year`), every difference exactly 0.001, and **all 213 move up** — which is the tell: at an exact midpoint, decimal rounding is half-away-from-zero and deterministic, while float noise had been landing some of them just below.

`count_attendance_value_term` is a safe denominator by construction, not by coincidence: `is_present_weighted` is `if(att_code like 'T%', 0.67, attendancevalue)`, so its only possible NULL source is `attendancevalue` — which this CTE already filters with `where attendancevalue is not null`. Numerator and denominator therefore cannot diverge on new data.

Bug 1 was previously diagnosed and fixed by @anthonygwalters in #3147, which was closed without merging and whose branch has since been deleted. This PR supersedes it. That fix used `avg(is_present_weighted)`, which is mathematically identical for the reason just given; `safe_divide` was chosen so all four weighted columns share one denominator expression. #3147 also proposed keeping the raw sum as a new `ada_weighted_present` column; that was left out as it has no consumers.

## AI Assistance

Claude Code did the investigation, root-cause analysis, and drafted the changes. Both bugs surfaced while scoping unrelated athletic-eligibility work, from a human-reported symptom that weighted and unweighted ADA looked identical. Scope decisions and the choice to fix upstream rather than work around it downstream were human-directed.

## Scope of impact

Measured on AY2025 (`rn_year = 1`, `enroll_status = 0`) by comparing a local dev build to production.

**Only high-school students in Newark and Camden are affected.** Middle and elementary read `ada_year`, which neither bug touched. Miami and Paterson have no high-school rows in this population.

| Region | HS students | avg `ada` now | avg `ada` after | `ada` values changed | Chronic-absence flag flips |
| Newark | 995 | 0.8969 | 0.8380 | 975 | 257 |
| Camden | 380 | 0.8616 | 0.8252 | 343 | 42 |

This table was measured before the bug-3 commit, against the float arithmetic, so its changed-value counts exclude 6 high-school rows that differed only by rounding and its flip counts are accurate to roughly plus or minus 2 rather than exact. The artifact responsible no longer exists — after this PR the same measurement would be exactly reproducible — but the table has not been re-measured, because bug 3 moves values by at most 0.001 and cannot change the shape of the impact.

### What actually changes

**The band-cased `ada` column, for HS only.** 1,318 students get a different value, averaging about 5 points lower — the correct direction, since partial-day presence now earns partial credit instead of silently falling back to the unweighted rate. Reports that display `ada` will show different high-school figures: `rpt_tableau__hs_early_warning_dashboard`, `rpt_tableau__mtss_rti`, `rpt_gsheets__mtss_rti`, `rpt_tableau__gradebook_dashboard`, `rpt_tableau__gradebook_gpa`, `rpt_tableau__student_course_grades`, `rpt_tableau__gpa_cumulative_year`, `rpt_tableau__gradebook_ms_hs_comments`, `rpt_gsheets__grad_plan_tracking`, and the `int_tableau__gradebook_audit_*` chain.

**Chronic-absence classification — 299 students flip.** Two threshold decisions read the band-cased column directly:

- `rpt_tableau__attendance_interventions:22` — `if(co.ada <= .90, true, false) as is_chronic_absence`
- `rpt_deanslist__designations:28` — `if(co.ada <= 0.9, 'Chronic Absence', null)`

This is the change most worth flagging to stakeholders: 257 Newark and 42 Camden high schoolers change chronic-absence status. Note the flip boundary matches the consumers' own test (`ada <= .90`), so a student landing exactly on 0.90 is not counted — 51 HS students sit exactly there and are chronically absent in both builds.

**Athletic eligibility.** `int_students__athletic_eligibility` gates HS Q2 on `cy_weighted_term_q1` at 0.9. On the broken scale that gate passed 99.5%; corrected, 42.9% pass. That is a change to a sub-condition, not directly to final status — Q2 status also depends on GPA and credits, so the count of students whose status string changes is smaller and is not measured here. Qualitatively, `Probation - ADA` becomes reachable for high school in Q2, where it previously was not. `rpt_gsheets__athletic_eligibility` also displays `py_y1_weighted_ada`, which was showing the unweighted value and will now be correct; that column is selected but never used in any eligibility CASE.

### What does NOT change, despite appearances

- **Both 0.80 flags are unaffected — 0 flips, verified.** `ada_above_or_at_80`'s third CASE branch is unguarded by `school_level`, so an HS student failing the weighted test at 0.80 falls through to `when ada.ada_year >= 0.80 then true`. `is_ada_above_or_at_80_cum_gpa_less_2` likewise ORs with `ada.ada_year >= 0.80`. Since weighted never exceeds unweighted (0 rows), the weighted branch cannot alter either result. Everything consuming `ada_above_or_at_80` is therefore untouched, including the gradebook-audit scaffold filter at `int_tableau__gradebook_audit_student_scaffold:164`.
- **The Tableau attendance reports are unaffected.** `rpt_tableau__attendance_comm_log`, `rpt_tableau__okrts_referrals` (including `is_chronically_absent`), `int_topline__attendance_contacts`, and `rpt_gsheets__kippmiami_payout_roster` all read `unweighted_ada`, not a weighted column. A grep for `weighted_ada` matches `unweighted_ada` as a substring, which is easy to over-count.
- **`ada_weighted_term_q2`, `q3`, and `q4` have no consumers at all.** Only `q1` is read anywhere.

**Not a schema change** — no columns renamed or removed, so no exposure breaks. Rollback is a straight revert of the two SQL hunks; both models rebuild on the next run.

### Latent issue, not fixed here

The band-casing at `int_extracts__student_enrollments:299` keys only on `school_level = 'HS' and academic_year >= 2025`, with no region exclusion. Weighted ADA is not KTAF practice for Miami, and this is inert today only because Miami has no high-school rows. If Miami opens a high school, it would silently pick up weighted ADA. Worth a follow-up issue.

## Test coverage

Range tests (`between 0 and 1`) on every ADA column in both models. These catch bug 1's class — an out-of-range value — and would have failed on its very first build.

They cannot catch bug 2's class, because a column copied from the wrong upstream is still a valid rate. So there is also a singular test, `int_powerschool__ada_term_pivot__weighted_year_matches_source`, asserting the pivot reproduces the `ada_weighted_year` the source model already computes. Verified as a real guard, not an assumed one: **66,843 failing student-years against production when measured**, where the bug is live, and **0 against this build**.

That test compares with **exact equality**. An earlier revision of it needed a `0.0015` tolerance to absorb the rounding artifact — exact equality failed 152 student-years on a build whose SQL was otherwise correct — and the bug-3 fix is what makes the tolerance unnecessary. Since `int_powerschool__ada_term` now aggregates in exact decimal, a student-year's rate is identical across its term rows and the pivot's coalesce reproduces it bit for bit, so nothing excuses a difference of any size. The test is a strict guard again rather than one with a hole in it three times the size of the artifact.

`int_powerschool__ada_term` also gains the uniqueness test the intermediate-layer convention requires and it was missing; grain `(_dbt_source_project, studentid, academic_year, term)` verified unique at 285,141 of 285,141 rows.

## Verification

- `dbt build --select int_powerschool__ada_term int_powerschool__ada_term_pivot --target dev --defer --favor-state` → **PASS=23, WARN=0, ERROR=0**, re-run after merging `main`
- The exact-equality guard test passes. That is the determinism evidence, not a formality: the pivot table is built in one BigQuery job and the source view is re-evaluated in a different one at test time, so the two sides agreeing bit for bit is a genuine cross-evaluation check — the same test shape failed 152 rows against the float arithmetic
- Range tests confirmed passing across **15 years and 285,141 rows** before being committed, so they will not fail CI on historical data
- Dev vs prod, all years: `ada_weighted_term` avg 37.44 → 0.8528, max 57 → 1, out-of-range rows 276,148 → 0; pivot `ada_weighted_year` now differs from `ada_year` in 65,845 of 74,644 rows where it previously differed in 0
- Bug 3 measured on identical source rows by computing both paths in one query, so no build-vintage confound: 2,179 student-years on an exact midpoint, 213 changed, max delta 0.001, all upward
- `trunk check --force` clean on all six changed files, run inside the worktree

One note on reading the commits. The first is a line-ending normalization, isolated deliberately: `int_powerschool__ada_term.sql` was the only model in its directory stored with CRLF, which made `git-diff-check` flag every added line and blocked the push. Content is unchanged, verified with `git diff --ignore-cr-at-eol`. The two bug fixes are in the second, the tests in the third, and the exact-decimal fix in the fourth (after a `main` merge whose only conflict was two branches appending a test entry to `tests/properties.yml` at the same position — both were kept).

Closes #4783

### General

      feedback; dismiss false positives with a brief reply explaining why. Both
      rounds addressed — the re-review's one remaining nit (a tracking-issue ref
      in a YAML description) is fixed.

### dbt

- [x] Properties files updated for both models — range tests plus corrected column descriptions (the previous auto-generated text described attendance-rate columns as student ID fields)
- [x] No exposure changes needed — no columns added, renamed, or removed
- [x] **Breaking change?** No schema break. Value change affecting high-school ADA in Newark and Camden; see Scope of impact
- [x] No external sources touched

## CI checks

- [x] **Trunk** — passes
- [ ] **dbt Cloud** — re-running after the `main` merge and the exact-decimal commit. On the previous head it failed for an unrelated, pre-existing reason: `dbt_utils_unique_combination_of_columns_rpt_parentsquare__parents_student_id__first_name__last_name` ("Got 3 results"). Production carried the same 3 violating groups, and it is already tracked in #4776. It surfaced on this PR only because ParentSquare sits downstream of `int_extracts__student_enrollments` and therefore falls inside `state:modified+`. Nothing in this PR touches it. Note that #4776 may have landed in the `main` now merged in — worth re-reading the new run rather than assuming the same failure.
- [x] **Dagster Cloud** — not triggered (no Dagster paths changed)


GabyRangelB:
Both Important items implemented in e7d3f5c7b. One declined, with reasoning. Details below.

## Important 1 — range tests can't catch bug 2's error class

Correct, and the PR body overclaimed. Added a singular test, `int_powerschool__ada_term_pivot__weighted_year_matches_source`, asserting the pivot's `ada_weighted_year` reproduces the value `int_powerschool__ada_term` already computes. Verified as a real guard rather than assumed: **66,843 failing student-years against current prod** (where the bug is live) and **0 against the fixed build**.

Two notes for anyone reusing the suggestion:

- The proposed `count(*) filter (where ...)` is Postgres syntax; BigQuery needs `countif()`. Went with a row-returning singular test instead of an aggregate check so failures identify the affected student-years.
- **Exact equality would have flaked.** The first version used `!=` and failed 152 rows on the *fixed* build. Every difference was exactly 0.001, across 10 academic years and 159 of ~74k student-years. Cause: both sides `round(..., 3)`, and a distributed float `sum()` has no guaranteed association order, so a rate sitting on a rounding midpoint rounds up in one evaluation and down in another. Float noise also put the diffs marginally *above* 0.001 (`0.001000000000000009`), so a `> 0.001` threshold didn't exclude them either. Settled on `> 0.0015`, which is 0 on the fixed build and 66,843 on prod. Rationale is documented in the test SQL and its description.

## Important 2 — no uniqueness test on int_powerschool__ada_term

Correct and added. Grain verified before adding: `(_dbt_source_project, studentid, academic_year, term)` is unique at **285,141 of 285,141 rows**. Agreed it's pre-existing; worth closing here since this PR already rewrites that file's test coverage.

## Minor 1 — tested columns not grouped at top of `columns:`

Declining, and the review already identifies why it's arguable — `fct_student_attendance_daily.yml` keeps `ada_tier`'s test in its logical position rather than hoisting it.

In the pivot, 10 of 25 columns now carry tests, and the columns are ordered by shape: four per-quarter blocks, then the semester/year rollups. Sorting by test-presence would interleave `ada_weighted_term_q1-q4` with `ada_semester_s1/s2` and `ada_weighted_year` while pushing `ada_term_q1-q4` and the grain keys below them, which reads worse than the current grouping. The convention's stated purpose is visibility, and with 10 tested columns spread across every logical block, hoisting doesn't deliver it. Happy to reorder if a maintainer prefers strict adherence.

## Recommendation — `round(safe_divide(a, b), 3)` repeated 8×

Agreed on both counts: 7 of the 8 predate this PR, and a macro isn't worth introducing in a bugfix. No change.

## Adopting the structural argument

The observation that `is_present_weighted` is `if(att_code like 'T%', 0.67, attendancevalue)` — so its only NULL source is `attendancevalue`, which the CTE already filters — is a stronger justification than the empirical row count in the PR body. It makes numerator/denominator consistency structural rather than a fact about current data. Updating the PR body to lead with that.

## On the unverifiable claims

Reasonable to flag given no warehouse access. All the figures have been re-verified locally, and the impact section was materially corrected before this review landed: an earlier draft wrongly listed `rpt_tableau__attendance_comm_log`, `rpt_tableau__okrts_referrals`, `int_topline__attendance_contacts`, and `rpt_gsheets__kippmiami_payout_roster` as affected. They read `unweighted_ada`, which neither bug touches — a grep for `weighted_ada` matches `unweighted_ada` as a substring.

## CI note

The `dbt Cloud` failure on this PR is unrelated and pre-existing: `dbt_utils_unique_combination_of_columns_rpt_parentsquare__parents_student_id__first_name__last_name`, "Got 3 results." Production has the same 3 violating groups, and it's already tracked in #4776. It surfaces here only because ParentSquare sits downstream of `int_extracts__student_enrollments` and so falls inside `state:modified+`.

GabyRangelB:
@claude re-review please — two commits since your pass (e7d3f5c7b).

**Implemented both Important items:**
- Singular test guarding bug 2 class: fails 66,843 student-years against prod, 0 against the fix.
- Added the missing uniqueness test on `int_powerschool__ada_term` (grain verified unique, 285,141/285,141).

**Heads up: exact equality flakes.** First version used `!=` and failed 152 rows on the *fixed* build, every diff exactly 0.001 — both sides round to 3 decimals and a distributed float `sum()` has no guaranteed association order. Using `> 0.0015` instead. Also your suggested `count(*) filter (where ...)` is Postgres, not BigQuery.

**Declined** the column-sort-to-top nit — reasoning in my earlier comment.

**Adopted** your `is_present_weighted` structural argument into the PR body.

**Impact re-verified** in this body and #4783: 1,318 changed values (rounding-only excluded), 299 chronic-absence flips, boundary now matches the consumers `ada <= .90` test.

**dbt Cloud red is pre-existing and unrelated** — `rpt_parentsquare__parents`, same 3 dupes in prod, tracked in #4776.

`dbt build` PASS=23 ERROR=0, trunk clean.

GabyRangelB:
Merged `main` (the branch had fallen well behind), fixed the one remaining nit, and fixed the rounding issue that both earlier rounds worked around rather than solved.

## Minor 1 — tracking-issue ref in a YAML description

Fixed. Verified the claim before acting on it: `src/dbt/CLAUDE.md` does state that descriptions carry no TODOs, history, or tracking-issue refs, and the description did contain the `#4783` reference. The cited line number had drifted (463 to 504 after the merge), but the finding itself was accurate. Dropped the ref, and while rewriting that description also dropped the row counts from it, since the same section of the guide asks for qualitative language.

## Bug 3 — the rounding artifact is now fixed, not tolerated

Both prior rounds treated the 0.001 flapping as a fact of the platform to absorb with a tolerance. It is a fixable defect. `int_powerschool__ada_term` summed FLOAT64 values, and a distributed float `sum()` has no guaranteed association order, so a rate landing on a 3-decimal midpoint rounded up in one evaluation and down in the next.

The fix casts the three daily values to NUMERIC in a new first CTE, so every aggregate and window sum is exact and order-independent, then casts each rounded rate back to FLOAT64 on projection. Decimal addition is exact, so association order stops mattering.

Casting back is the load-bearing half: 15 contracted models declare `ada`, `cy_weighted_term_q1`, or `py_y1_weighted_ada` as `float64`. Letting NUMERIC propagate would be the more literal reading of the cast-early-once convention, but it breaks all 15 plus the Tableau-facing types, so this intermediate is the containment boundary.

Measured by computing both paths over identical source rows in one query, so there is no build-vintage confound:

- 2,179 of 74,644 student-years sit exactly on a 3-decimal midpoint for `ada_weighted_year` — the population whose rounded value float noise was deciding
- 213 of them disagree between the float and exact paths in a single measurement (16 for the unweighted `ada_year`), every difference exactly 0.001
- all 213 move up, which is the tell: at an exact midpoint, decimal rounding is half-away-from-zero and deterministic, while float noise had been landing some of them just below

One correction to the earlier framing in this PR. I initially expected the flapping to show up as term rows disagreeing within a student-year. It does not — production has zero student-years whose term rows disagree, because BigQuery computes a window once per partition and broadcasts it. The instability is strictly across separate query evaluations, which is precisely why it surfaced as pivot-table-versus-source-view rather than as a range violation.

## The guard test is strict again

```sql
-- before
where abs_difference > 0.0015

-- after
where pivot_ada_weighted_year != source_ada_weighted_year
```

It passes, and that is evidence rather than a formality: the pivot table is built in one BigQuery job and the source view is re-evaluated in a different one at test time, so the two sides agreeing bit for bit is a genuine cross-evaluation determinism check. The same test shape failed 152 rows against the float arithmetic.

## Verification

- `dbt build` on both models after the merge — PASS=23, WARN=0, ERROR=0
- `ada_weighted_term`, all years, dev versus prod — avg 37.44 to 0.8528, max 57 to 1, out-of-range rows 276,148 to 0
- pivot `ada_weighted_year` now differs from `ada_year` in 65,845 of 74,644 rows, where it previously differed in 0
- `trunk check --force` clean on all six changed files, run inside the worktree
- the merge's only conflict was two branches appending a test entry to `tests/properties.yml` at the same position; both were kept

dbt Cloud is re-running. The previous head's failure was the pre-existing ParentSquare duplicate tracked in #4776, which may itself have landed in the `main` now merged in, so the new run is worth reading rather than assuming.

GabyRangelB:
@claude re-review please — two commits since your last pass (`e7d3f5c7b`), plus a `main` merge.

**New code to review** is `f12dacdb2`, which fixes the rounding non-determinism your last two passes saw only as a tolerance. `int_powerschool__ada_term` now casts the three daily attendance values to NUMERIC in a new first CTE, aggregates and windows in exact decimal, and casts each rounded rate back to FLOAT64 on projection. Worth checking specifically:

- whether the cast-back is complete — every column whose type would otherwise change should be cast, so the model's output schema is identical to today's
- whether the new `daily_attendance` CTE respects the cast-early-once and ST06 ordering conventions, given the cast column reuses the source column name
- whether `avg(attendancevalue)` on NUMERIC is safe to leave as-is next to seven `safe_divide` siblings

**The guard test is now exact equality** rather than `> 0.0015`. It passes. The determinism argument is that the pivot table is built in one BigQuery job while the source view is re-evaluated in another at test time, so agreement is a genuine cross-evaluation check.

**Correction to my earlier framing**, in case it shaped your read: the flapping does not appear across term rows within a student-year. Production has zero such disagreements, because BigQuery computes a window once per partition. It is strictly across separate evaluations.

**Also fixed** your remaining nit — the tracking-issue ref is out of the YAML description.

Measurements, all-years: 2,179 of 74,644 student-years sit exactly on a 3-decimal midpoint; 213 differ between the float and exact paths, every difference exactly 0.001, all upward.

CI is green on `f12dacdb2` — dbt Cloud, Trunk, and CodeQL all pass. The 11 dbt test warnings are pre-existing; the two in ADA consumers (`rpt_tableau__student_course_grades` at 12, `rpt_tableau__gradebook_gpa` at 3,791) reproduce exactly against production, so they are not from this change.

GabyRangelB:
No findings to address in the re-review. Closing the one claim it flagged as unverified, since that one is load-bearing rather than incidental.

## `AVG()` preserves NUMERIC — now verified, not assumed

The re-review reasoned that BigQuery's `AVG()` keeps NUMERIC rather than promoting to FLOAT64, but flagged it as high-confidence-and-unverified because the job has no warehouse access. It is worth closing: `ada_term` is the only column in the model still using `avg()`, and if `AVG()` promoted to FLOAT64, that column would have kept the exact non-determinism this PR is fixing — the cast to NUMERIC would have been silently undone for one of the eight rates.

```sql
select
  cast(avg(cast(v as numeric)) as string) as avg_of_numeric_input,
  cast(avg(v) as string) as avg_of_float_input
from unnest([0.1, 0.2]) as v
```

| avg_of_numeric_input | avg_of_float_input |
| --- | --- |
| 0.15 | 0.15000000000000002 |

Exact on the NUMERIC path, inexact on the FLOAT64 path, over inputs chosen so the two answers have to differ if the type is promoted. `sum()` behaves the same way (`0.3` versus `0.30000000000000004`). So `avg()` is safe to leave in place, and the reviewer's point 3 is confirmed by measurement rather than inference.

## The other two unverified items

- **CI on `f12dacdb2`** — verified green on both surfaces, and the status is confirmed to belong to that SHA rather than a stale one. dbt Cloud success, Trunk Check success, CodeQL success across all three analyses.
- **Row-level arithmetic** — measured as stated. Worth adding that the two dbt warnings landing in ADA consumers reproduce at exactly 12 (`rpt_tableau__student_course_grades`) and 3,791 (`rpt_tableau__gradebook_gpa`) against production, so neither is from this change. The ParentSquare failure that was red on the previous head is gone, having been fixed in the `main` merged in here.

One note on the environment limitation showing up in all three passes: the observation that this job cannot query the warehouse is accurate and has now shaped three reviews. The claims it cannot check are exactly the ones this PR turns on, so a reviewer reading only the bot's passes is taking the measurements on trust. That is an argument for a human spot-check of the numbers before merge, not for treating the review as complete on its own.

cbini (approved):

PR #4786, opened 2026-08-10
perf(dbt): remove 59 vacuous data tests that cannot fail

### LINKED ISSUE #4784: Remove 69 vacuous dbt tests that cannot fail by construction
## Summary

A sweep of all 16 dbt projects found 69 data tests that cannot fail by
construction. Each still costs a full BigQuery scan per CI run, and because most
of the host models are views, that scan re-expands the whole upstream chain
rather than reading one column.

[#4773](https://github.com/TEAMSchools/teamster/issues/4773) removed 69 of these
from marts **primary key** columns. This issue covers the rest: the **foreign
key** columns it did not touch, plus two other vacuous classes.

All findings are in `kipptaf` (66) and `finalsite` (3). Eleven of sixteen
projects are clean.

## Class A: `not_null` on an unwrapped surrogate key (55)

`dbt_utils.generate_surrogate_key` coalesces every input to a null sentinel
before hashing, so it never returns NULL. `src/dbt/CLAUDE.md` already states the
rule.

| Where | Count |
| --- | --- |
| `kipptaf` marts (bridges / dimensions / facts) | 52 |
| `kipptaf` intermediates (`int_schoolmint_grow__observations`, `int_schoolmint_grow__observation_types`) | 2 |
| `kipptaf` staging (`stg_google_sheets__people__locations.location_key`) | 1 |

Verified per model, not just per YAML: every definition site is an unwrapped
macro call, and none sits behind an outer join that could null it. Conditional
keys are excluded — a wrapped key such as

```sql
if(sr.employee_number is not null, {{ dbt_utils.generate_surrogate_key(["sr.employee_number"]) }}, cast(null as string)) as lead_teacher_staff_key
```

keeps its `not_null`, because that test is real.

One case the classifier missed and was confirmed by hand:
`bridge_student_contacts.student_contact_person_key` is an `if()` picking
between two surrogate keys, both non-null.

## Class B: `not_null` on a non-nullable expression (11)

| Model | Columns | Why it cannot fail |
| --- | --- | --- |
| `kipptaf/dim_staff_cube_access` | 7 scope columns | `coalesce(ovr.x, rp.x, 'none')` |
| `kipptaf/fct_survey_submissions`, `kipptaf/bridge_survey_expectations` | `respondent_type` | string literal in every union branch |
| `finalsite/int_finalsite__contact_address_of_record` | `candidate_count` | `coalesce(cc.candidate_count, 0)` |
| `finalsite/int_finalsite__student_address_of_record` | `resolution_status` | `coalesce(s.address_source, 'unresolved')` |

Open question for review: the seven `dim_staff_cube_access` scope columns feed
Cube row-level access policies. They cannot fail today, but the test doubles as
a guard if the `coalesce` is ever dropped. **Held out of the first pass** as
candidates for a deliberate keep, the way #4773 kept `assessment_date_key`.

## Class C: `relationships` on an inner-joined parent (3)

`src/dbt/CLAUDE.md` already documents this: the join has already dropped every
unmatched row.

- `dim_staff_cube_access.staff_key` to `dim_staff` — inner join on that column
- `fct_survey_responses.survey_submission_key` to `fct_survey_submissions` —
  inner join on that column
- `dim_student_enrollment_status.student_enrollment_key` to
  `dim_student_enrollments` — final select reads `from dim_student_enrollments`

## Checked and deliberately kept

- `dim_student_iep_status.is_current` and
  `dim_student_meal_eligibility_status.is_current` are **not** vacuous.
  `max(effective_end_date)` returns NULL when every value is NULL, and `least()`
  propagates that.
- `dim_student_ell_status.is_current` is vacuous, but only through a two-hop
  in-model chain (`coalesce` feeding `least`). Too fragile to justify removing.
- No duplicate generics anywhere in the repo. No single-column
  `dbt_utils.unique_combination_of_columns` shadowing a `unique`.

## Scope of the first pass

62 of the 69 are removed: all 55 in class A, all 3 in class C, and 4 of the 11
in class B. The 7 `dim_staff_cube_access` scope tests are held pending the
review question above.

Verified against the parsed manifests rather than the YAML alone: `kipptaf` goes
from 808 tests to 748 and `finalsite` from 39 to 37, and every removed node name
begins with `not_null_` or `relationships_`. Nothing was added.

## Not a breaking change

No model SQL, column, or contract changes. One visible side effect: 62 dbt tests
are 62 Dagster asset checks, so that many checks disappear from the Dagster UI.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will remove 59 dbt data tests that cannot fail by
construction. Each one still costs a full BigQuery scan per CI run, and because
most of the host models are views, that scan re-expands the whole upstream chain
rather than reading one column.

[#4773](https://github.com/TEAMSchools/teamster/issues/4773) removed 69 such
tests from marts **primary key** columns. This PR covers the rest of the sweep:
mostly the **foreign key** columns it did not touch, plus two other vacuous
classes. One correction to that framing, from review: `dim_terms.term_key` is a
single-column **primary key** that #4773's pass never reached — `dim_terms.yml`
was untouched by it. So Class A is not purely FK columns. Of the 59 removals, 9
sit on a PK-constrained column: 6 are composite-PK members in bridges (which are
FKs at the same time), and 3 are single-column PKs.

**Class A — 55 `not_null` tests on unwrapped surrogate keys.**
`dbt_utils.generate_surrogate_key` coalesces every input to a null sentinel
before hashing, so it never returns NULL. `src/dbt/CLAUDE.md` already states the
rule. 52 are marts columns, 2 are `int_schoolmint_grow__*` intermediates, and 1
is `stg_google_sheets__people__locations.location_key`.

Every removal was verified against the model SQL, not just the YAML: each
definition site is an unwrapped macro call, and none sits behind an outer join
that could null it. Conditional keys are excluded — a wrapped key such as
`if(sr.employee_number is not null, {{ generate_surrogate_key([...]) }}, cast(null as string)) as lead_teacher_staff_key`
keeps its `not_null`, because that test is real.

**Class B — 1 `not_null` test on a non-nullable expression.**
`int_finalsite__contact_address_of_record.candidate_count` is
`coalesce(cc.candidate_count, 0)`, and the column carries no `accepted_values`
test.

Three further Class B candidates were removed and then **restored** after
review. `src/dbt/CLAUDE.md` states that `accepted_values` compiles to
`where value not in (...)`, which NULL never satisfies, so it must be paired
with `not_null` on any enum column that must be non-null — explicitly including
one a `coalesce` makes non-null by construction. That is exactly the reasoning
those three were removed under, and
`int_finalsite__contact_address_of_record.resolution_status` sets the precedent
in the same project. Restored:
`bridge_survey_expectations.respondent_type`,
`fct_survey_submissions.respondent_type`, and
`int_finalsite__student_address_of_record.resolution_status`.

**Class C — 3 `relationships` tests on an inner-joined parent.** The join has
already dropped every unmatched row, which `src/dbt/CLAUDE.md` documents:
`dim_staff_cube_access.staff_key`, `fct_survey_responses.survey_submission_key`,
and `dim_student_enrollment_status.student_enrollment_key` — the last of which
selects directly `from dim_student_enrollments`.

### Held out for review

The 7 `dim_staff_cube_access` scope columns are `coalesce(ovr.x, rp.x, 'none')`
and so are equally vacuous, but they feed Cube's row-level access policies and
the test doubles as a tripwire if that `coalesce` is ever dropped. They are left
in place pending a call on this PR, the way #4773 kept `assessment_date_key`.

### Checked and deliberately kept

`dim_student_iep_status.is_current` and
`dim_student_meal_eligibility_status.is_current` are **not** vacuous —
`max(effective_end_date)` returns NULL when every value is NULL, and `least()`
propagates that. `dim_student_ell_status.is_current` is vacuous, but only
through a two-hop in-model chain, which is too fragile to justify removing.

The sweep also found no duplicate generics anywhere in the repo, and no
single-column `dbt_utils.unique_combination_of_columns` shadowing a `unique`.
Eleven of sixteen projects are clean.

### Verification

Checked against the parsed manifests rather than the YAML alone:

- `dbt parse --no-partial-parse` clean on `kipptaf` and `finalsite`
- `kipptaf` 808 tests to 750; `finalsite` 39 to 38 — delta exactly 59
- every removed node name begins with `not_null_` or `relationships_`; nothing
  added
- `trunk check --force` on all changed files: no issues

Two YAML files also had their tested-columns-first ordering restored
(`fct_survey_responses`, `int_schoolmint_grow__observation_types`). Removing a
column's only test had left an untested column above tested ones. Column order
in a properties file does not affect contract enforcement — `fct_survey_responses`
already differs from its select order and builds clean.

### Not a breaking change

No model SQL, column, or contract changes. One visible side effect: 59 dbt tests
are 59 Dagster asset checks, so that many checks disappear from the Dagster UI.

Refs #4784

## AI Assistance

Claude Code ran the sweep across all 16 dbt projects, verified each candidate
against the model SQL, applied the removals, and confirmed the result against
the parsed manifests. Human-directed: the decision to open an issue and work in
a worktree, and the scope call to hold the seven `dim_staff_cube_access` scope
tests pending review.

## Self-review

### General

### dbt

      models; existing properties files edited only
- [x] Include (or update) an
      model changes, so no exposure changes
- [x] **Breaking change?** No. No columns renamed or removed, no contract
      changes. Rollback is a straight revert of the single commit.
      no external source changes

## CI checks

### DIFF
--- a/src/dbt/CLAUDE.md
+++ b/src/dbt/CLAUDE.md
@@ -759,9 +759,25 @@ legitimately-superseded inactive rows that repeat the key.
 - Unscoped `+config` applies to tests from all installed packages, not just the
   current project
 - **`accepted_values` passes NULLs** — it compiles to
-  `where value not in (...)`, which NULL never satisfies. Pair it with
-  `not_null` on any enum column that must be non-null, including one a
-  `coalesce` makes non-null by construction.
+  `where value not in (...)`, which NULL never satisfies. Every enum column that
+  must be non-null carries `not_null` too, including one a `coalesce` makes
+  non-null by construction. **Never delete a `not_null` from a column that
+  carries `accepted_values`.** It is not vacuous, whatever the SQL looks like —
+  the pairing is the only thing making the enum test reject NULL.
+- **Never add `not_null` to a column that cannot be NULL by construction.** It
+  can never fail, and it still costs a full BigQuery scan per CI run — on a view
+  mart that scan re-expands the entire upstream chain. Non-nullable by
+  construction means every definition site is one of: an unwrapped
+  `generate_surrogate_key`, a `coalesce` / `ifnull` with a non-null default, a
+  literal in every UNION branch, or `count(...)`. The `accepted_values` pairing
+  above overrides this rule; nothing else does.
+
+### Verifying a test-removal PR
+
+Never report a count from the YAML diff — it does not say which dbt nodes
+actually disappeared. `dbt parse` on main and on the branch, then diff the
+`resource_type == 'test'` node names. That fixes the delta and proves nothing
+unintended was dropped.
 
 ### An FK check belongs on the pre-join model, as a column `relationships` test
 
@@ -892,8 +908,8 @@ if(
 Without this, relationship tests check the placeholder hash against the parent
 dimension and fail.
 
-Corollary: never add `not_null` tests on `generate_surrogate_key` output — it
-never returns NULL.
+**Never add a `not_null` test to `generate_surrogate_key` output** — it never
+returns NULL, so the test cannot fail. This holds for FK columns as much as PKs.
 
 #### Nullable PK inputs need a fallback, not a null-wrap
 
@@ -1152,8 +1168,11 @@ alias.
 - All new or modified models require `description:` on the model and every
   column. Profile staging data via BigQuery MCP; infer downstream from parents.
   Describe calculated fields by logic. Use qualitative language — no stats.
-- Columns with **per-column** `data_tests:` should be sorted to the top of the
-  `columns:` list for visibility. Model-level composite tests
+- Columns with **per-column** `data_tests:` must be sorted to the top of the
+  `columns:` list for visibility — including after a change that strips a
+  column's last test. Reorder freely under `contract: enforced`: BigQuery
+  matches contract columns by name, not position (`fct_survey_responses` already
+  differs from its `select` order and builds clean). Model-level composite tests
   (`dbt_utils.unique_combination_of_columns`, etc.) do not trigger this rule —
   they go in the model-level `data_tests:` block ABOVE `columns:`, and their
   referenced columns can stay in their natural / contract order.
--- a/src/dbt/finalsite/models/api/intermediate/properties/int_finalsite__contact_address_of_record.yml
+++ b/src/dbt/finalsite/models/api/intermediate/properties/int_finalsite__contact_address_of_record.yml
@@ -77,10 +77,6 @@ models:
           Number of distinct addresses on this contact's household linkage. One
           resolves outright; zero means no household carries a street; more than
           one means the emitted address was picked from among them.
-        data_tests:
-          - not_null:
-              config:
-                severity: error
       - name: address_1
         data_type: string
         description:
--- a/src/dbt/finalsite/models/api/intermediate/properties/int_finalsite__student_address_of_record.yml
+++ b/src/dbt/finalsite/models/api/intermediate/properties/int_finalsite__student_address_of_record.yml
@@ -57,6 +57,9 @@ models:
                   - unresolved
               config:
                 severity: error
+          - not_null:
+              config:
+                severity: error
       - name: address_source
         data_type: string
         description:
--- a/src/dbt/finalsite/models/api/intermediate/properties/int_finalsite__student_address_of_record.yml
+++ b/src/dbt/finalsite/models/api/intermediate/properties/int_finalsite__student_address_of_record.yml
@@ -57,9 +57,6 @@ models:
                   - unresolved
               config:
                 severity: error
-          - not_null:
-              config:
-                severity: error
       - name: address_source
         data_type: string
         description:
--- a/src/dbt/kipptaf/models/google/sheets/staging/properties/stg_google_sheets__people__locations.yml
+++ b/src/dbt/kipptaf/models/google/sheets/staging/properties/stg_google_sheets__people__locations.yml
@@ -16,9 +16,6 @@ models:
           - unique:
               config:
                 severity: error
-          - not_null:
-              config:
-                severity: error
       - name: location_name
         data_type: string
         description: |
--- a/src/dbt/kipptaf/models/marts/bridges/properties/bridge_course_section_teachers.yml
+++ b/src/dbt/kipptaf/models/marts/bridges/properties/bridge_course_section_teachers.yml
@@ -23,8 +23,6 @@ models:
             to_columns: [course_section_key]
             warn_unsupported: false
         data_tests:
-          - not_null
-
           - relationships:
               arguments:
                 to: ref('dim_course_sections')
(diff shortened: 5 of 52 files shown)


cbini:
Verified all four findings against the code before acting. Pushed as
`312315ae8`.

**1. `accepted_values` + `not_null` pairing — correct, fixed.**

The rule is at `src/dbt/CLAUDE.md:761-764` and it anticipates the exact
reasoning I removed these under ("including one a `coalesce` makes non-null by
construction"). The decisive check was the sibling:
`int_finalsite__contact_address_of_record.resolution_status` — untouched by this
PR, in the same project — already pairs `accepted_values` with a
`severity: error` `not_null`. That is the convention in practice, not just on
paper.

Restored all three, byte-identical to what was removed:

- `bridge_survey_expectations.respondent_type`
- `fct_survey_submissions.respondent_type`
- `int_finalsite__student_address_of_record.resolution_status`

Net removal is now 59, not 62. `candidate_count` has no `accepted_values` and
stays removed, as you noted.

**2. `dim_terms.term_key` is a PK — correct, PR narrative fixed.**

Confirmed: `dim_terms.yml` declares `type: primary_key` on `term_key` and is not
in #4773's diff. On your suggestion I checked the rest of the removals rather
than just this one — 9 of the 59 sit on a PK-constrained column: 6 are
composite-PK members in bridges (simultaneously FKs), and 3 are single-column
PKs. Only `dim_terms.term_key` is a `not_null` removal on a single-column PK
that #4773 missed. Updated the summary to say "mostly the foreign key columns"
and to state the split.

**3. `rpt_focus__demographics` single-column UCC — the instance is real, the
contradiction is not.**

Confirmed the file has a single-column `combination_of_columns: [stdt_id]`. But
the claim in the PR was specifically that no single-column
`unique_combination_of_columns` **shadows** a `unique` — that is, coexists with
one on the same column, making it redundant. That is what the sweep tested and
it still holds; this instance has no sibling `unique`, which you note yourself.

It is a genuine deviation from the "Uniqueness test examples" convention, just a
different defect from the one audited, in `kippmiami`, in a file this PR does
not touch. Leaving it out of scope rather than widening a deletion-only PR into
`kippmiami`. Happy to file it separately.

**4. Stale column ordering — correct, fixed.**

Both cases confirmed against `src/dbt/CLAUDE.md:1155`. Since I was already in
those files, I fixed rather than deferred:

- `fct_survey_responses.yml`: `survey_submission_key` moved below
  `survey_question_key`
- `int_schoolmint_grow__observation_types.yml`: `staff_observation_type_key`
  moved below `tag_id` and `abbreviation` (its description said "the unique test
  on tag_id below", now corrected to "above")

One thing worth recording for anyone who repeats this: reordering columns in a
**contracted** properties file looked risky, so I checked before doing it.
`fct_survey_responses` already has a YAML column order that differs from its
final `select` order and builds clean, so BigQuery contract enforcement matches
by name, not position. The reorder is safe.

**On the count you could not verify.** Your static count of 62 removed list
entries matched the pre-review diff exactly. Re-verified after the restore
against the parsed manifests: `kipptaf` 808 to 750, `finalsite` 39 to 38, total
59, every removed node named `not_null_*` or `relationships_*`, nothing added.

PR #4787, opened 2026-08-10
fix(google): validate groupKey before sending student group memberships

### LINKED ISSUE #4766: fix(google): validate groupKey exists before sending student group memberships
## Problem

`rpt_google_directory__users_import` builds `groupKey` by string concatenation with no check that the group actually exists in Google Workspace:

```sql
'group-students-' || region || '@teamstudents.org' as `groupKey`,
```

`orgUnitPath`, three lines up in the same select, **is** validated — it joins `stg_google_directory__orgunits` on the path, so a path missing from Google resolves to null and the student drops out of the extract. #4513 added that guard for org units and left the group name unguarded.

There is nothing to join against for groups: the `groups` asset is commented out in `src/teamster/code_locations/kipptaf/google/directory/assets.py`.

## How it surfaced

Dagster run `bc28dc20-ae50-4af7-b45a-b3a0bd05d440` (2026-08-07, `kipptaf__google__directory__user_sync_schedule`). The `zero_api_errors` check on `kipptaf/google/directory/user_create` failed with 430 errors:

- 429x `404 Resource Not Found: groupKey` for `group-students-paterson@teamstudents.org`
- 1x `409 Entity Already Exists` (unrelated, pre-existing account)

The 429 user accounts were created successfully. Only the membership adds failed, and they failed deterministically — the Paterson students group has never existed.

Paterson entered this extract on 2026-08-03 when 079381c63 (Refs #4513) removed it from the exclusion list. First provisioning wave ran 2026-08-04.

## Second problem: stranded memberships

Group membership is attempted only inside the `where is_create` branch of `google_directory_user_create`. Those students are now matched in the directory, so `is_create` is false and no future run retries them. Creating the group by hand does not fix them retroactively.

Google currently holds 883 Paterson student accounts (576 Paterson Prep Elementary, 307 Paterson Prep Middle), none in a students group. The backfill needs all 883, not just the 429 from this run.

## Scope

1. Create `group-students-paterson@teamstudents.org` in Google Workspace (Ops / admin console — not a code change).
2. Guard `groupKey` the way `orgUnitPath` is guarded. Two options:
   - Name the group explicitly in the `people__locations` sheet and resolve it through `int_people__location_crosswalk`, mirroring what #4513 did for org unit paths.
   - Re-enable the `groups` asset and join `stg_google_directory__groups` on the group email.
3. Backfill group membership for the 883 existing Paterson accounts. Decide whether this is a one-off or whether membership reconciliation belongs in the recurring sync — the current create-time-only design has no drift correction for any region.

## Related

- Refs #4513
- Same run also closed out a 4-day cascade on the same cohort: `412 Insufficient licenses` (08-04, 08-05) and `400 Invalid Password` (08-04 through 08-06). Both resolved on their own; noting them only so the alert history reads correctly.

cbini:
# Pull Request

## Summary & Motivation

...stop `rpt_google_directory__users_import` from sending group-membership adds
for a Google group that does not exist, without blocking student account
provisioning.

The extract built the student group address by string concatenation —
`'group-students-' || region || '@teamstudents.org'` — with nothing validating
that the group exists. `orgUnitPath`, three lines above in the same select, **is**
validated: it joins `stg_google_directory__orgunits`, so a missing path resolves
to null and the student drops out. #4513 added that guard for org units and left
the group address unguarded. When Paterson entered this extract on 2026-08-03,
its students group had never been created, so run
`bc28dc20-ae50-4af7-b45a-b3a0bd05d440` logged 429 identical
`404 Resource Not Found: groupKey` errors — one per created student.

`groupKey` now gets the same treatment, with one deliberate difference: **a null
`groupKey` does not drop the student.** The account is still created and only the
membership is skipped. Mirroring the `orgUnitPath` guard literally would mean no
Paterson student gets an account at all until Ops creates the group — a
provisioning regression traded for a cosmetic fix.

What changed, in four parts:

1. The dormant `groups` asset in the kipptaf Google Directory location is
   re-enabled and added to the existing 1:30am nonpartitioned schedule, so group
   data reaches GCS as Avro.
1. A `src_google_directory__groups` external source and a
   `stg_google_directory__groups` staging model expose it to dbt.
1. `rpt_google_directory__users_import` derives the constructed address as
   `group_key_target` in `with_google` and left joins the staged groups on
   `email` in `final`, projecting the resolved value as `groupKey`. The trailing
   `where` clause is unchanged.
1. `members_for_created_users` skips users whose `groupKey` is null and returns
   one aggregated error naming the count and the affected `orgUnitPath` values,
   which `user_create` folds into its `zero_api_errors` check. So the alert fires
   once per run instead of once per student.

Also fixes two things found along the way: `list_groups` was passing
`maxResults=500` to an endpoint documented at a 200 maximum (latent until this PR
made the method live — its discovery entry carries no `maximum` key, so
`googleapiclient` never validated it), and the extract's model description still
claimed Paterson was excluded, untrue since 079381c63.

### Prerequisite and post-merge actions

Neither is a code change, and both need an owner:

- [ ] **Ops:** create `group-students-paterson@teamstudents.org` in the admin
      console. Until it exists, Paterson accounts are created without membership
      and each run reports one warning.
- [ ] **On merge:** launch `kipptaf/google/directory/groups` in prod as soon as
      the deploy lands. The prod external table does not exist until a prod
      materialization drops a file, and `build_dbt_assets` runs
      `stage_external_sources` on every dbt asset run. External sources are
      excluded from the `any_deps_missing` gate, so the first post-deploy tick
      **will** request the new staging model — ahead of the 1:30am schedule.
      Losing that race costs one failed dbt run: `create or replace external
      table` leaves nothing dropped, the prod extract is a view that keeps its
      old definition, provisioning is unaffected, and it self-heals after the
      manual materialization. Accepted deliberately rather than splitting this
      into two PRs.

### Out of scope

Still open on #4766: backfilling group membership for the ~884 existing Paterson
accounts, and recurring membership reconciliation. Membership is attempted only
under `is_create`, so no region has drift correction today and a student whose
membership was skipped is never retried. That work needs the `members` asset,
which re-enabling `groups` now makes possible — `members` and `MEMBERS_SCHEMA`
are deliberately left commented out here.

Two smaller deferrals worth recording: `libraries/google/directory/ops.py` holds
an unguarded duplicate of the membership-payload builder (dead — its only
importer is an underscore-prefixed archived test pytest never collects; the right
fix is deleting it, not porting the guard), and `list_members` has the same
uncapped `max_results` that `list_groups` had, still dormant.

## AI Assistance

Claude Code authored the design, the plan, and the implementation, working from
issue #4766 and the existing `orgUnitPath` guard as the pattern. Human-directed:
the decision to validate against Google rather than the `people__locations`
sheet, the decision to keep the account and skip only the membership, and the
decision to ship as one PR and accept the post-deploy race. Design spec and
implementation plan are in `docs/superpowers/`.

## Self-review

### General

- [ ] Run `uv run pytest tests/test_dagster_definitions.py` for any modified code
      location

The two unchecked boxes cannot run in a Codespace: `kipptaf.definitions` fails at
module load on the Illuminate and Zendesk dlt credential resolvers, which are
unset locally and unrelated to this change. Verified instead by `py_compile` on
the three edited modules, a runtime import confirming `GROUPS_SCHEMA` generates
the expected 10 Avro fields, and the branch deployment. `uv run pytest tests/resources/test_resource_google_directory.py`
passes 7/7 on the `members_for_created_users` guard.

- [x] Include (or update) an
      summary above.
      with **`--target staging`** so the dbt Cloud CI job can find the table. See

No columns were renamed or removed. `groupKey` keeps its name and type and only
becomes nullable, which the contract does not constrain. The consumer is the
`user_create` / `user_update` asset pair, not a dashboard, so no exposure applies.

The staging run needs a location override, because CI's `target=staging` resolves
`cloud_storage_uri_base` to the prod bucket while the branch deployment writes to
the test one:

```text
uv run dbt run-operation stage_external_sources
  --args "select: google_directory.src_google_directory__groups"
  --vars '{cloud_storage_uri_base: gs://teamster-test/dagster/kipptaf, ext_full_refresh: true}'
  --target staging
  --project-dir src/dbt/kipptaf
```

The `groups` asset joins an existing schedule rather than adding one, so the
automations catalog is unchanged in substance. Regenerating it here would drop
every code location that fails to import in a Codespace, so it is left alone —
per `docs/CLAUDE.md`, that script must run in a full environment.

## CI checks

- [ ] **Trunk** — passes. If it fails, run `trunk check` and `trunk fmt` locally,
      fix any issues, and push again.

`trunk check --force --no-fix` was run on all 12 changed files from inside the
worktree: no issues.

## Open verification gate

The join compares the constructed address to `stg_google_directory__groups.email`
only, while the code it replaces passed that address to `members.insert`, which
also resolves aliases and unique ids. If `newark`, `camden`, or `miami`'s address
is an **alias** of a group whose primary address differs, this change would turn a
working region's `groupKey` null and silently stop its membership adds. Checked
once the staged table holds data:

```sql
select email
from `teamster-332318.zz_stg_kipptaf_google_directory.stg_google_directory__groups`
where email like 'group-students-%'
```

Expect `camden`, `miami`, `newark` — and not `paterson`. If one of the three is
missing, its address is an alias and the join needs
`or w.group_key_target in unnest(g.aliases)`; the staging model already carries
`aliases`.

Refs #4766

### DIFF
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -50,6 +50,14 @@
   filter matches the pushed delta (which includes the merge commit), NOT the net
   three-dot PR diff (where the merged-in files, now equal to main, don't
   appear).
+- **A markdown-only commit still re-triggered the `kipptaf` deploy** on
+  `pull_request`, despite `"!**/CLAUDE.md"` in BOTH trigger blocks of
+  `deploy-prod-kipptaf.yaml` (verified: the commit touched 3 files, all `.md`,
+  and a `kipptaf` run appeared for that headSha). This contradicts the
+  pushed-delta model above, which predicts no run — one of the two is
+  incomplete, and no mechanism is established here. Plan for the cost: a
+  docs-only push to a PR is NOT free, so batch doc changes into the code commit
+  rather than splitting them off to "avoid CI".
 - **Each `deploy-prod-<location>.yaml` push-`paths` must list every dbt package
   in that district's `src/dbt/<district>/packages.yml`** (`src/dbt/pearson/**`,
   etc.). Drift silently skips that district's prod deploy on a shared
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -275,7 +275,10 @@ file; domain specifics live in the nearest subdirectory CLAUDE.md.
   findings are advisory, and `git grep` settles it faster than complying.
   **Always invoke `superpowers:receiving-code-review` BEFORE processing
   `claude-review` findings** — verify each claim (including its file:line
-  citations) against the code before relaying or replying, not after.
+  citations) against the code before relaying or replying, not after. Fixing a
+  finding in code is not a reply — post a per-finding verdict as a PR comment
+  (declines included, with reasons). A silent fix reads to a human reviewer as
+  an unaddressed review.
 
 - **A dispatched code-review subagent's "confirmed non-issue" dismissals aren't
   authoritative** — one over-read the `unnest` scalar-aggregate carve-out to
@@ -331,6 +334,13 @@ file; domain specifics live in the nearest subdirectory CLAUDE.md.
 - **IDE selection arrives only via `<ide_selection>` tags**, not
   `<ide_opened_file>` (which only names the open path). When the user references
   "this" without an `<ide_selection>`, ask for the snippet — don't guess.
+
+- **Before claiming a harness artifact (rewritten output, phantom rendering,
+  truncated literal), verify with a DERIVED value** — line length, `grep -c`, a
+  checksum. A misread is far likelier than a rewriting pipeline, and a plausible
+  substitute string survives eyeballing where a length does not. This session an
+  asserted "tool output renders X as Y" artifact was a plain misread, and it
+  produced a false correction to the user before it was tested.
 
 - **Built-in tools over Bash**: Use dedicated tools for file I/O (Read, Grep,
   Glob, Edit, Write). Bash is only for commands with no dedicated tool (`git`,
--- a/docs/superpowers/plans/2026-08-10-google-directory-groupkey-guard.md
+++ b/docs/superpowers/plans/2026-08-10-google-directory-groupkey-guard.md
@@ -1,0 +1,1003 @@
+# Google Directory `groupKey` Guard Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use
+> superpowers:subagent-driven-development (recommended) or
+> superpowers:executing-plans to implement this plan task-by-task. Steps use
+> checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stop sending group-membership adds for a Google Workspace group that
+does not exist, without blocking student account provisioning, and surface the
+missing group once per run instead of once per student.
+
+**Architecture:** Re-enable the dormant `groups` asset in the kipptaf Google
+Directory code location so a `stg_google_directory__groups` model exists, then
+left join it in `rpt_google_directory__users_import` so an unresolvable group
+address becomes a null `groupKey` — the same shape `orgUnitPath` already uses.
+The membership-payload builder skips null-`groupKey` users and returns one
+aggregated error, which the `user_create` asset folds into its `zero_api_errors`
+check.
+
+**Tech Stack:** Dagster (`dagster`, `dagster-gcp`), Google Admin SDK Directory
+API via `googleapiclient`, Pydantic + `py_avro_schema`, dbt on BigQuery, pytest,
+uv.
+
+Spec:
+`docs/superpowers/specs/2026-08-10-google-directory-groupkey-guard-design.md`
+
+Refs [#4766](https://github.com/TEAMSchools/teamster/issues/4766)
+
+## Global Constraints
+
+- **Worktree:** every command and every file edit targets
+  `/workspaces/teamster/.worktrees/cbini/fix/claude-google-directory-groupkey-guard`.
+  Use `git -C <worktree>` for git and `uv --directory <worktree> run ...`
+  (prefixed with `VIRTUAL_ENV=`) for Python. Editing
+  `/workspaces/teamster/<path>` instead silently dirties `main`.
+- **Python:** always `uv run`, never bare `python` / `python3` / `pytest`.
+- **`requires-python = ">=3.13"`.** Built-in generics (`list[dict]`,
+  `tuple[list[dict], list[dict]]`), `X | None` for nullable.
+- **Docstrings:** Google Python Style Guide. Multi-line is expected here.
+- **Do not run `trunk fmt`.** The pre-commit hook formats. Run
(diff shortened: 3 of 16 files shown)


cbini:
All three findings verified against the code before acting. Two were fixed and one recommendation is declined with reasoning. Fixes are in `8a32bdced`.

### Important 1 — alias/case gap on the `groupKey` join

**Case half: real, and fixed.** The mechanism checks out, and it is measurable rather than hypothetical — 14 of 35,461 rows in the sibling `stg_google_directory__users` model carry a non-lowercase `primary_email`, so the Directory API preserves the case supplied at creation. Worse than a plain gap: the code this join replaced passed the address to `members.insert`, which resolves the key case-insensitively, so exact-match equality was a silent behavior regression of the same class as the alias risk.

Fixed in the **staging model**, not the `ON` clause. `src/dbt/CLAUDE.md` forbids one-sided calculations in join predicates, so `lower(g.email)` in the join was not available. `stg_google_directory__groups` now selects `lower(email) as email` with a comment recording why, and the column moves to the end of the select to keep ST06 ordering. An address is a case-insensitive identifier and Google will not permit two groups differing only by case, so normalizing cannot introduce a false match — and it strengthens the existing `unique` test, since case variants now collide and get caught.

**Alias half: verified clean, no code change needed.** Ran the gate against the staged table after materializing the asset in the branch deployment (243 groups):

| Region | Present | Primary or alias | Already lowercase |
| --- | --- | --- | --- |
| camden | yes | primary, `array_length(aliases) = 0` | yes |
| miami | yes | primary, `array_length(aliases) = 0` | yes |
| newark | yes | primary, `array_length(aliases) = 0` | yes |
| paterson | no | n/a | n/a |

All three working regions resolve as primary addresses with zero aliases, so the `unnest(g.aliases)` fallback is unnecessary and was not added — adding it speculatively would widen the match to addresses Google itself would reject. Paterson's absence is the bug this PR guards.

### Minor 2 — no default-parameter test for `list_groups`

Correct, and the sibling precedent made it cheap. Added `test_list_groups_uses_max_results_200`. Two deliberate differences from the siblings: the mocked response key is `groups` rather than `items`, and the test name omits `_and_items_key`, both because `list_groups` relies on the default `response_data_key` (the api_name) instead of overriding it.

### Minor 3 — `unresolved_group_errors` skipped the per-item log

Correct. Now loops with `context.log.error(msg=ue)` before appending, matching `create_errors` and `members_errors`.

### Recommendation declined — dbt unit tests for `rpt_google_directory__users_import`

Not taken here. The gap is real and pre-existing across the whole model, not introduced by this PR, and closing it means fixtures for three enrollment-source branches plus two Google directory joins — a larger change than the guard itself, in a model this PR touches in one place. Noted on #4766 alongside the other deferred items rather than bundled in.

### Note on review currency

This review passed before `8a32bdced` and `2539a1ac1`, so it has not seen the fixes it prompted. `claude-review` does not re-run on push. The changes were verified by the 10 passing unit tests, `dbt compile`, `trunk check --force` on the changed files, the gate query above, and a green dbt Cloud run with zero warnings.

Refs #4766

PR #4789, opened 2026-08-10
fix(marts,cube): resolve assessment reporting term by date, drop it from the section dim

### LINKED ISSUE #4484: fix(marts,cube): reporting-term resolution is wrong — arbitrary quarter on the section dim, wrong-grain path for assessments
Two related reporting-term-resolution defects, surfaced during the #4330
teacher-linkage refactor.

## 1. `dim_student_section_enrollments.term_key` is an arbitrary quarter

The model tags each section enrollment with a reporting quarter (`term_key`, FK
to `dim_terms`) by joining `stg_google_sheets__reporting__terms` on
`(cc_abs_termid = powerschool_term_id, sections_schoolid = school_id, region)`
where `type = 'RT'`.

`powerschool_term_id` is not quarter-specific. The reporting-terms grain is
`(name, school_id, region, powerschool_year_id)`, so one PowerSchool term id
maps to several reporting quarters. The join matches multiple RT rows per section
enrollment (140 keys, up to 12x), and the model collapses them to one row per PK
**arbitrarily** — for ~8k enrollments under those coarse term ids, `term_key`
points to a basically-random quarter.

Pre-existing; it was masked by a dedupe until the teacher-linkage refactor, which
preserves the status quo (a deterministic-but-arbitrary quarter).

### Deeper problem: grain mismatch

Section enrollments and reporting quarters are **different grains**. A section
enrollment — especially a year-long one — spans multiple quarters, so "the
reporting quarter for a section enrollment" is not well-defined. The coarse-key
join is a symptom; conflating the two grains is the root.

## 2. Assessment scores resolve their reporting term through the wrong path

`student_assessment_scores_view` reaches its term breakdown via
`student_assessment_scores.student_section_enrollments.terms` — i.e., through the
section enrollment's `term_key` (problem 1). Because of the grain mismatch above,
an assessment score's reporting term should **not** come from the section
enrollment. It should key off the assessment's **administration date** (or date
taken) — the quarter the assessment was actually given in — via the
administration/dates path, not the section.

(The `proficiency_rollup` already uses the administration date, not `date_taken`,
as its time anchor because `date_taken` is corrupt for a share of rows — so the
administration date is the reliable key.)

## Proposed fixes

- Resolve the assessment view's reporting term from the administration date (the
  reporting quarter covering it), not the section enrollment.
- For the section dim: decide whether a single `term_key` belongs on a
  section-enrollment grain at all; if kept for the roster, resolve by date (the
  quarter containing the entry date) rather than the coarse term id.

## Notes

- `dim_student_section_enrollments` is the only model joining reporting terms on
  `powerschool_term_id`.
- No dbt model consumes `dim_student_section_enrollments.term_key`; consumption
  is Cube-only (roster + assessment views, via the section cube's `terms` join).

Refs #4330

cbini:
# Pull Request

## Summary & Motivation

...fix reporting-term resolution for assessment scores, and remove the broken
term FK from the section-enrollment dim.

**What was actually wrong.** `dim_student_section_enrollments.term_key` joined
the reporting-terms sheet on `cc_abs_termid`, which for every current section is
the **full-year** PowerSchool term (`3500` for AY2025) while the sheet carries
quarter ids (`3503`-`3506`). Nothing matches. The column is 100% NULL for AY2024
and 0.5% populated for AY2025, so the Term folders in **both** Cube views
(`student_section_enrollments_view`, `student_assessment_scores_view`) return
NULL today.

| academic_year | rows | non-null `term_key` |
| ------------- | ------ | ------------------- |
| 2024          | 85,886 | 0                   |
| 2025          | 91,901 | 444 (0.5%)          |
| 2026          | 53,060 | 16                  |

The 8,109-row fan-out described in the issue is a separate, historical cause:
the join omits `academic_year`, and the sheet duplicates a term id across two
academic years for some AY2018-2023 rows (e.g. `2801` filed under both 2017 and
2018).

**Section dim.** A section enrollment spans several quarters, so a single term FK
does not belong at that grain. Dropped `term_key` from
`dim_student_section_enrollments` — and with it the reporting-terms join and the
`dbt_utils.deduplicate` that was masking the fan-out — plus the `terms` join on
the section cube and the Term folder on the roster view.
`bridge_course_section_terms` already models section-to-quarter correctly (by
date, ~3.9 quarters per section); it is the path to use if a Term-filtered roster
is ever wanted. Not wired into Cube here — nobody asked for it.

**Assessment scores.** The quarter is now resolved in the fact, from each row's
own `assessment_date_key`, at the resolved section's school and region. RT rows
never overlap within a `(school_id, region)` (verified), so `BETWEEN` matches at
most one.

Resolving it one layer up in `int_assessments__resolved_section_enrollments` was
tried and **rejected**: that model is one row per score _grain_, and scores
sharing a grain carry different dates, which put 925 rows on a quarter that did
not contain their date. The resolver now just carries `powerschool_school_id` and
`region` through. A new singular test asserts the resolved quarter contains the
score's date.

### Result

| metric                                            | before      | after       |
| ------------------------------------------------- | ----------- | ----------- |
| `term_key` populated on the assessment-scores fact | 0.4%        | 100% - 441 rows |
| rows whose quarter excludes their own date         | 925         | 0           |
| fact row count / distinct PK                       | 14,508,002  | 14,508,002  |
| section dim row count / distinct PK                | 837,722     | 837,722     |

The 441 remaining nulls are scores whose date falls outside any RT quarter for
their school (Paterson has no AY2024 RT rows). Row counts and PKs are identical
to prod on both models, so neither change fans out or drops rows.

## AI Assistance

Claude-authored, human-directed. The user pointed at the issue and chose to
delete the section dim's `term_key` rather than re-resolve it by entry date.
Claude did the diagnosis (the full-year-term-id finding, which reframed the fix),
the implementation, and the validation.

### General

### dbt

- [x] Properties files updated for every touched model, plus a properties entry
      for the new singular test.
- [x] Cube is the only consumer; its exposure (`cube.yml`) already covers both
      models, and the Cube YAML changes ship in this PR.
- [x] **Breaking change** — `dim_student_section_enrollments.term_key` is
      removed. Consumers: Cube only, updated here. The four Term members
      (`semester`, `term_name`, `term_code`, `term_type`) disappear from
      `student_section_enrollments_view` and keep working, unchanged in name, on
      `student_assessment_scores_view`. Rollback: revert the commit; nothing
      outside Cube reads the column.
- [ ] No external sources added or modified.

### Validation

Local `dbt build` against the branch, dev target, `--defer --favor-state`:
`PASS=29 WARN=0 ERROR=0` — including all 6 resolver unit tests, the fact PK
`unique`, every `relationships` test (the new `term_key` to `dim_terms` among
them), and the new
`fct_assessment_scores_enrollment_scoped__term_covers_assessment_date`.

`trunk check --force` clean after the commit-time `fmt`.

**Cube model changes are not validated by CI** — per `src/cube/CLAUDE.md`, branch
schema validation is manual, and the cubes read `kipptaf_marts`, which lacks
`term_key` until this merges. The added join mirrors the shape
`student_attendance` already uses (fact to `terms` on `term_key`, view
`join_path: <fact>.terms`), and the view's Term member names are unchanged.

`docs/reference/marts-data-models.md` was regenerated
(`uv run scripts/generate_marts_reference.py`). Five of its six added lines are
pre-existing drift the regeneration corrected — the four `dim_staff` FK edges and
the `assessment_date_key` FK row, all of whose constraints predate this PR. Only
`term_key` to `dim_terms` is new.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.
- [ ] **Dagster Cloud** — not triggered (`src/dbt/**` and `src/cube/**` are
      excluded from the deploy workflows' `pull_request` paths).

Closes #4484

### DIFF
--- a/docs/reference/marts-data-models.md
+++ b/docs/reference/marts-data-models.md
@@ -29,6 +29,7 @@ erDiagram
   fct_assessment_scores_enrollment_scoped }o--|| dim_assessment_administrations : "assessment_administration_key"
   fct_assessment_scores_enrollment_scoped }o--|| dim_student_section_enrollments : "student_section_enrollment_key"
   dim_assessment_administrations }o--|| dim_assessments : "assessment_key"
+  dim_student_section_enrollments }o--|| dim_staff : "lead_teacher_staff_key"
   dim_student_section_enrollments }o--|| dim_student_enrollments : "student_enrollment_key"
   dim_student_section_enrollments }o--|| dim_course_sections : "course_section_key"
   dim_student_enrollments }o--|| dim_students : "student_key"
@@ -40,7 +41,9 @@ erDiagram
 | FK column                        | References                        |
 | -------------------------------- | --------------------------------- |
 | `assessment_administration_key`  | `dim_assessment_administrations`  |
+| `assessment_date_key`            | `dim_dates`                       |
 | `student_section_enrollment_key` | `dim_student_section_enrollments` |
+| `term_key`                       | `dim_terms`                       |
 | `test_date_key`                  | `dim_dates`                       |
 
 ## fct_assessment_scores_student_scoped
@@ -120,6 +123,7 @@ erDiagram
 ```mermaid
 erDiagram
   fct_grades_assignments }o--|| dim_student_section_enrollments : "student_section_enrollment_key"
+  dim_student_section_enrollments }o--|| dim_staff : "lead_teacher_staff_key"
   dim_student_section_enrollments }o--|| dim_student_enrollments : "student_enrollment_key"
   dim_student_section_enrollments }o--|| dim_course_sections : "course_section_key"
   dim_student_enrollments }o--|| dim_students : "student_key"
@@ -139,6 +143,7 @@ erDiagram
 ```mermaid
 erDiagram
   fct_grades_category }o--|| dim_student_section_enrollments : "student_section_enrollment_key"
+  dim_student_section_enrollments }o--|| dim_staff : "lead_teacher_staff_key"
   dim_student_section_enrollments }o--|| dim_student_enrollments : "student_enrollment_key"
   dim_student_section_enrollments }o--|| dim_course_sections : "course_section_key"
   dim_student_enrollments }o--|| dim_students : "student_key"
@@ -172,6 +177,7 @@ erDiagram
 ```mermaid
 erDiagram
   fct_grades_term }o--|| dim_student_section_enrollments : "student_section_enrollment_key"
+  dim_student_section_enrollments }o--|| dim_staff : "lead_teacher_staff_key"
   dim_student_section_enrollments }o--|| dim_student_enrollments : "student_enrollment_key"
   dim_student_section_enrollments }o--|| dim_course_sections : "course_section_key"
   dim_student_enrollments }o--|| dim_students : "student_key"
--- a/src/cube/model/cubes/student_assessments/student_assessment_scores.yml
+++ b/src/cube/model/cubes/student_assessments/student_assessment_scores.yml
@@ -28,6 +28,14 @@ cubes:
           {CUBE}.student_section_enrollment_key
         relationship: many_to_one
 
+      # Reporting quarter of the score, keyed on the fact's own term_key (the
+      # quarter containing the administration/test date at the section's school).
+      # Reached only through this path: the section dim no longer carries a term
+      # FK, so there is no diamond back to terms (#4484).
+      - name: terms
+        sql: "{terms.term_key} = {CUBE}.term_key"
+        relationship: many_to_one
+
     dimensions:
       - name: assessment_score_key
         description: Surrogate key. Primary key for this fact.
@@ -48,6 +56,11 @@ cubes:
         sql: CAST(test_date_key AS TIMESTAMP)
         type: time
         public: true
+
+      - name: term_key
+        description: FK to terms — the reporting quarter containing the score.
+        sql: term_key
+        type: string
 
       - name: enrollment_resolution
         description: >-
--- a/src/cube/model/cubes/students/student_section_enrollments.yml
+++ b/src/cube/model/cubes/students/student_section_enrollments.yml
@@ -14,10 +14,6 @@ cubes:
         sql: "{course_sections.course_section_key} = {CUBE}.course_section_key"
         relationship: many_to_one
 
-      - name: terms
-        sql: "{terms.term_key} = {CUBE}.term_key"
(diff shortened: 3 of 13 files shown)


cbini:
Verified all four. Three applied in 519fc49f6, one answered below.

**1 — docs-regen miscount. Correct, fixed.** `assessment_date_key`'s
`foreign_key` block is on `origin/main` untouched by this diff, so five of the six
added lines are pre-existing drift, not four. Only `term_key` to `dim_terms` is
new. PR body corrected.

**2 — RT non-overlap on the full sheet. Confirmed on CI.** dbt Cloud run
`70403220343758` (target `staging`, `dbt build --select state:modified+
--full-refresh`) came back **success**, and `warning_only` on it returns exactly
one test warning: `unique_fct_grades_term_grades_term_key`, 13 results. That is
pre-existing — prod `kipptaf_marts.fct_grades_term` has the same 13 duplicate
`grades_term_key`s, and it is the known `base_powerschool__course_enrollments`
double-write (#3900, Ops cleanup #3915), untouched by this PR.
`unique_fct_assessment_scores_enrollment_scoped_assessment_score_key` ran clean in
that build, so the no-fan-out result holds against the full reporting-terms sheet,
not just the local `--defer` slice.

Worth adding to your reasoning: PK-uniqueness is a guardrail against fan-out, but
it is not the only one here. The overlap query itself was run — a self-join of RT
rows on `(school_id, region)` with `a.start_date <= b.end_date and b.start_date <=
a.end_date`, excluding same-row matches, returns **0 rows** across the entire
sheet, all years, all regions. And the new singular test
`fct_assessment_scores_enrollment_scoped__term_covers_assessment_date` catches the
weaker failure (a matched quarter that does not contain the row's date) that PK
uniqueness would not.

**3 — column-group blank line. Correct, fixed.** Blank line added between the `f.`
and `t.` groups, matching
`tests/fct_student_attendance_daily__enrollment_anchor_exists.sql`.

**4 — `term_key` description wording. Correct, fixed.** It was a leftover from an
earlier draft that did resolve the quarter in the intermediate. Reworded to say
the quarter is matched at the school of the section enrollment the intermediate
resolved, rather than implying the intermediate does the matching.

On the shared-macro note under "checked out clean": agreed it is a 5th
duplication, and deliberately left alone. Factoring the BETWEEN-join plus hash
into a macro would touch four models this PR has no reason to open, and the hash
input list is the thing that must not drift — a macro helps there but the
`relationships` test to `dim_terms` already fails loudly if it does.

PR #4797, opened 2026-08-10
refactor(focus): adopt probe-gated dlt sync from the powerschool template

### LINKED ISSUE #4447: refactor(focus): adopt probe-gated dlt sync from the powerschool template
## Context

Companion to #4446 (Illuminate). The PowerSchool dlt factory (`src/teamster/libraries/dlt/powerschool/assets.py`, landed via #4415; design spec `docs/superpowers/specs/2026-07-16-powerschool-dlt-probe-gated-sync-design.md`) established a probe-gated incremental sync style: one `@dlt_assets` multi-asset over all tables, per-table `cursor_column`, a `COUNT(*)` + `MAX(cursor)` change signature stored in dlt `resource_state`, full `replace` only on signature drift, parallel extract via the `table_rows` generator, and progress streamed to the Dagster event log. Unchanged tables are excluded from the run entirely, so an idle tick loads nothing.

Focus (`src/teamster/libraries/dlt/focus/assets.py` + `code_locations/kippmiami/dlt/focus/`) predates this: one asset factory call per table, each an unconditional full `replace` via the `sql_database` source, all reloaded by the daily 04:00 ET schedule regardless of upstream change.

## Goal

Adopt the probe-gated style for Focus: consolidate the per-table assets into one probe-gated multi-asset over the `public` schema, keep asset keys (`[kippmiami, dlt, focus, table]`) and the `dagster_kippmiami_dlt_focus` dataset unchanged, and share the probe/signature helpers with the powerschool library (extract to a shared module rather than copying). With gating in place, the sync could also run more frequently than daily at near-zero idle cost.

## Focus-specific considerations (not mechanical)

1. **Cursor column identification.** Derive a per-table `cursor_column` (or explicit none = always reload) from the landed schemas in `dagster_kippmiami_dlt_focus` and record it in `config/focus.yaml`. Focus is mid-rollout in Miami — new tables should declare a cursor as they're added.
2. **Empty source tables.** The PyArrow backend writes NO BigQuery table for a 0-row extract (see `libraries/dlt/focus/CLAUDE.md`). Confirm the probe/signature flow handles a 0-count table sanely (signature `{count: 0, max_cursor: null}` should gate it out after the first run rather than re-extracting nothing daily).
3. **Preserve `reflection_level=full_with_precision` + `remove_nullability_adapter`** — the `table_rows` generator accepts both, same as the powerschool factory.
4. **`autodetect_schema=True`** is set on the current bigquery destination; the powerschool pipeline omits it. Decide which applies — changing it against existing tables can surface schema-migration constraints.
5. **Testing constraint.** Focus uses an IP allowlist — the codespace cannot reach the database; verification requires a branch deployment (GKE static egress IP). Branch-deployment dlt runs write to the PROD dataset (no branch redirect), so plan validation accordingly.

## Acceptance

- An unchanged Focus table selected by the schedule is probed but not reloaded (asset not in the run).
- A changed table (signature drift) is fully replaced, identical output to today.
- Asset keys and the destination dataset are unchanged; downstream dbt staging models build without source edits.

## Related

- #4446 (Illuminate sibling — shared helper extraction should serve both)
- #4427 (PowerSchool dlt migration umbrella — the template this adopts)
- #4425 (shared `DagsterDltTranslator` base — natural to do alongside)

cbini:
# Pull Request

## Summary & Motivation

...adopt the PowerSchool dlt probe-gated sync style for Focus, so an unchanged
Focus table is probed but never reloaded.

Focus loaded all 77 tables from Postgres to BigQuery three times a day,
unconditionally. After this change:

- **One two-mode multi-asset** replaces the 77 per-table assets. With `probe` in
  run config (sensor mode) it loads exactly the run's selection using the passed
  signatures; without it (04:00 cron or a manual launch) it probes first, then
  loads everything unconditionally.
- **A new intraday sensor** (900s) probes every table each tick, compares
  `COUNT(*)` plus `MAX(updated_at)` against the baseline stored in dlt
  `resource_state`, and requests a run for only the drifted tables. An idle tick
  launches nothing.
- **The schedule drops to `0 4 * * *` and narrows to the count-only tables.** The
  12:00 and 14:00 crons are retired; the sensor makes the live-Focus snapshot
  fresher than two fixed times did. The remaining 04:00 tier reloads only the
  tables whose `cursor_column` is `null` (`co_teachers`, `login_history` — derived
  from config, not hardcoded), because a count-only probe cannot see an in-place
  edit that leaves the row count unchanged. The other 75 tables have a
  verified-reliable `updated_at`, so the sensor gates them alone and they get no
  unconditional reload.
- **Shared helpers move to `libraries/dlt/probe.py`** (`PowerSchoolTable` becomes
  `ProbeTable`), which Illuminate can reuse next.
- **Focus gains the `dlt_extract_workers` run tag** for per-run extract
  concurrency, matching PowerSchool's diagnostic knob. No factory parameter — no
  caller would pass one.

### Also fixes a latent PowerSchool bug (outside the Focus scope)

Adding that knob to Focus surfaced that PowerSchool's copy of it never worked.
Its op body ran `dlt.config["extract.workers"] = workers`, but the op's resource
parameter is named `dlt`, which shadows the `dlt` module inside the function —
so `.config` resolved to `DagsterDltResource`, which has no such attribute and
raises `AttributeError`. Verified against the exact initialized object the op
receives. It was dormant because nothing sets the tag and no code location
passes `max_extract_workers`, so the effect was that *using* the documented knob
crashed the run.

Fixed by writing through the module-level accessor (`from dlt import config as
dlt_config`), the form Focus already used and warned about in a comment. The
existing test only covered the pure `_resolve_extract_workers` helper — it
returned the right number and nothing exercised the line that applied it — so
both libraries now have op-body tests that assert the module config actually
received the value. Both were confirmed to fail with `AttributeError` against
the old line before it was fixed.

Asset keys, the `dagster_kippmiami_dlt_focus` dataset, and the destination schema
are all unchanged, so no dbt source edits are needed and rollback is a plain
revert.

Design spec and implementation plan are in the diff under `docs/superpowers/`.
Refs #4447.

### Cursor columns

`cursor_column: updated_at` on 75 tables; `null` (count-only) on `co_teachers`
and `login_history`, the only two Focus tables without that column. Verified
against the landed schemas in `dagster_kippmiami_dlt_focus`, which are 1:1 with
the 77 configured tables.

### One design question this answered

Whether dlt commits `resource_state` for a 0-row table that yields only
reflection hints plus `materialize_table_schema()` was the spec's one open risk.
**It does** — so an empty table gates out after its first load rather than
re-extracting nothing every 15 minutes. Pinned by
`tests/libraries/test_dlt_focus_signature_state.py::test_empty_table_persists_its_signature`,
so a dlt upgrade that changes the behavior fails CI instead of silently
regressing.

### Before merging, please look at

1. **Retiring the 12:00 and 14:00 crons.** Those times were chosen deliberately
   around the 12:45 delivery, and the reasoning is preserved in `schedules.py`.
   Continuous 15-minute gating should be strictly safer than two fixed times,
   but it does replace a clock-pinned contract with a dependency-based one.
2. **`dagster/max_runtime` of 900 on the 04:00 schedule.** New in this PR. The
   sensor's in-flight guard skips while a schedule run is non-terminal, so
   without a bound a hung 04:00 run would wedge intraday syncing silently. Since
   that tier now loads two small tables rather than 77, the bound is sized to
   roughly one sensor interval instead of to load time.
3. **The 75 cursor tables now have no unconditional reload at all.** That is the
   deliberate consequence of narrowing the tier, and it means a stopped sensor
   freezes them rather than degrading them to daily. The runbook in
   `kippmiami/CLAUDE.md` now names "is the sensor running?" as the first
   diagnostic for stale Focus data, and drift on a single table is fixed with a
   manual launch for that table.

## AI Assistance

Claude Code authored this end to end: the design spec, the implementation plan,
and the code, executed task-by-task with a review gate per task and a
whole-branch review at the end. Human-directed: the four design decisions
(sensor-based intraday tiering, consolidating to one asset, leaving
`autodetect_schema` alone, sharing only the pure helpers rather than building a
generic sensor factory), plus approval of the spec and plan before any code was
written.

The final review found one Important issue (the missing runtime bound above) and
four test gaps, all fixed and independently re-verified by mutation testing —
each new test was checked to actually fail when the behavior it covers is broken.

### General

### Dagster

- [x] `uv run pytest tests/libraries/ -k dlt` — 65 passed. 4 failures exist in
      `tests/libraries/` and every one was reproduced identically on
      `origin/main`: `test_dlt_powerschool_assets.py::test_config_matches_spec_membership_map`
      plus 3 under `tests/libraries/powerschool/sis/odbc/`. Pre-existing, not
      touched by this branch.
- [ ] `uv run dagster definitions validate` for `kippmiami` — **not runnable in
      the codespace**: the module resolves `FOCUS_DB` credentials eagerly at
      import and they are unset there. Substituted `py_compile` on every edited
      module plus a script that builds the real objects with fake credentials and
      asserts 77 asset specs, the op name `kippmiami__dlt__focus`, the sensor
      name, and schedule-name equality with what the sensor's in-flight guard
      matches on. The Dagster Cloud check on this PR is the real gate.
- [x] Follows the Library plus Config pattern; asset keys unchanged.

### Docs

- [x] `libraries/dlt/CLAUDE.md` and `libraries/dlt/focus/CLAUDE.md` updated for
      the new contract.
- [ ] `docs/reference/automations.md` **not** regenerated. Its generator
      silently skips code locations that fail to import, and `kippmiami` cannot
      import in the codespace, so regenerating here would drop locations from the
      catalog. Needs a run in a full environment.

## CI checks

- [ ] **Trunk** — `trunk check --force` over all 31 changed files was clean
      locally before push.
- [ ] **dbt Cloud** — no dbt changes in this PR.
- [ ] **Dagster Cloud** — this is the real validation gate for the Dagster
      definitions; see the checklist below.

## Validation before the sensor is enabled

Focus sits behind an IP allowlist, so the codespace cannot reach it — these need
the branch deployment. Note branch-deployment dlt runs write to the **prod**
dataset (dlt has no branch redirect), so the order below minimizes prod writes.

- [ ] Sensor left stopped. Evaluate one tick from the Dagster UI sensor test.
      Expect a `focus probe: N/77 changed` log line and either a skip or a run
      request; zero writes. **Also record the tick duration** — the sensor does
      77 sequential probes over the WAN and Dagster's default sensor evaluation
      timeout is 60s. A timing-out sensor never fires and never reports a
      failure.
- [ ] Launch the asset job manually for `districts`, `schools`, and one 0-row
      table. Confirm success, then re-evaluate the tick and confirm all three
      gate out. This is the end-to-end proof of the state round-trip through
      `_dlt_loads` that the unit test cannot cover.
- [ ] Confirm a changed table still fully replaces, with the same row count and a
      fresh `_dlt_load_id`.
- [ ] Run the one-off full 77-table load (this doubles as the baseline seeding
      step below). **Record its wall time and peak concurrent Postgres
      connections** — one pod now runs up to 5 parallel extracts with up to 20
      generators held open, a different shape from 77 pool-bounded pods. The
      04:00 tier no longer does a load this size, so this measurement is about
      the seeding load and any future manual full launch, not the nightly bound.
- [ ] Confirm the narrowed 04:00 tier: a scheduled run should touch only
      `co_teachers` and `login_history`, and finish well inside the 900s bound.
- [ ] Check the `dlt_focus_kippmiami` pool limit in deployment settings — with one
      op replacing 77, the pool no longer bounds table-level concurrency, and it
      is the only guard against a schedule run overlapping a sensor run.

## Cutover after merge

- [ ] Confirm the `kippmiami` location deployed.
- [ ] Seed every table's baseline with a **one-off manual launch of the Focus
      asset job over all 77 tables**. The `0 4 * * *` tier cannot do this — it
      now targets only the count-only tables. Doing it deliberately keeps the one
      unavoidable full load supervised.
- [ ] **Then** enable `kippmiami__dlt__focus__intraday_sensor`. It ships stopped
      on purpose: the sensor selects any table with no stored signature, so
      enabling it before a seeded baseline makes the first tick select all 77 at
      once — correct, but unsupervised.
- [ ] Watch the first three ticks — expect skips or small selections, never 77.
- [ ] Regenerate `docs/reference/automations.md` in a full environment.


cbini:
## Review response

Verified each finding against the code before acting. All three Important items were real; fixed in `1ce3a149d` (docs) and `93d3b47e3` (code).

### Fixed

**#1 — stale `kippmiami/CLAUDE.md` runbook.** Correct, and the most useful catch here: the Active Integrations row and the whole midday-import section were built on the 12:00/14:00 Focus pulls this PR removes. Rewritten to the 04:00-backstop-plus-sensor model. What I kept, because it is still true: the Finalsite 12:00 SFTP push and contacts pull (a different integration), the 12:45 delivery budget, the 2:00pm commitment, "Keep the 04:00 runs" and the FRESH 05:00 dependency, and the rule that safety is a dependency rather than a clock time — which the sensor makes easier to satisfy, not harder. Added one diagnostic the new model needs: **if midday Focus data looks stale, first check whether the intraday sensor is running**, since a stopped sensor silently reverts freshness to once a day. That replaces "did the 12:00 cron fire?" as the first thing to check.

**#2 — undocumented pool backstop.** Correct. Added the note mirroring `kippnewark/CLAUDE.md`: `dlt_focus_kippmiami` must stay at limit 1, because the 04:00 schedule is a plain cron with no in-flight guard of its own (only the sensor calls `in_flight_run`), so the pool limit is the only thing stopping a schedule run and a sensor run from overlapping and racing on the shared `_dlt_pipeline_state` row. The limit itself lives in Dagster+ deployment settings, not this repo — it is now an explicit item to confirm before the sensor is enabled.

**#3 — spec and CLAUDE.md disagreed on the 0-row risk.** Correct, and the distinction you drew is the right one. The unit test proves dlt's state-commit *mechanics* over sqlite and extract-only; the round-trip through the real BigQuery `_dlt_loads` / `_dlt_pipeline_state` tables is still unproven. Spec section retitled and reworded to say exactly that, with branch-deployment confirmation kept as the open item. The CLAUDE.md "Verified" claim stands but is now scoped to the mechanics.

**#4 and #7 — duplicated import-time work.** Correct on both counts, and they had one fix: `assets.py` now binds the parsed table list at module level, and `sensors.py` imports `sql_database_credentials` and `tables` from it instead of re-resolving the credentials and re-parsing the same YAML. One credential resolution and one parse per code-location import. `schedules.py` keeps its own read — it needs a different projection of the config for its `target` list.

### Not changing, with reasoning

**#5 — probe scale spot-check.** Already covered rather than dismissed: the PR's validation checklist has recording the tick's wall time as an explicit step, specifically because 77 sequential probes over the WAN run against Dagster's 60s default sensor evaluation timeout. Agreed it needs a real measurement; that measurement is scheduled.

**#6 — plan snippets predate the last two fixes.** Leaving as-is. `docs/superpowers/plans/` is a point-in-time planning artifact, not a living reference, and this repo's convention is explicit that a plan describing a superseded design is expected while a wrong *published* doc is a bug. The two fixes it predates came out of review, which is the process working; re-syncing the plan to match the code would misrepresent what was planned.

**#8 — `probe={}` would run sensor-mode with no baselines.** Leaving unguarded, agreeing with your own "not actionable" read. It is unreachable: the sensor never emits a `RunRequest` with an empty `changed`, and the schedule never sets `probe`. The failure mode if it were reached is an extra reload next tick, not data loss. A guard for an unreachable state is speculative.

### One note

The PR body was a placeholder when this review ran — it was patched in immediately after opening (the two-step avoids the entity re-encoding that round-tripping a body through the API causes). It now carries the `automations.md` follow-up, the unticked branch-deployment checklist, and the note that the sensor ships STOPPED pending baseline seeding.

`uv run pytest tests/libraries/ -k dlt` is green at 65 passed. Four failures exist in `tests/libraries/` and every one reproduces identically on `origin/main` — one in `test_dlt_powerschool_assets.py` and three under `tests/libraries/powerschool/sis/odbc/`.

PR #4800, opened 2026-08-10
feat(kipptaf): add rpt_gsheets__kfwd_student_salesforce_ids lookup

### LINKED ISSUE #4798: feat(kipptaf): expose Salesforce contact IDs to Google Sheets for Coupa aid tagging
## Problem

Coupa credit-card transactions for emergency postsecondary aid (Maher Fund) carry no student identifier. Disbursements are matched to Salesforce contacts by name at load time, and name alone is ambiguous for about 1 percent of students: 102 contacts across 50 colliding name groups.

## Proposal

Publish `rpt_gsheets__kfwd_student_salesforce_ids`, a Google Sheets extract exposing the Salesforce Contact ID alongside enough identifying context to pick the right student. The person entering a transaction in Coupa copies the ID into a Coupa field, so the ID rides along in the CSV export and no name matching is needed downstream.

### Model

- `src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kfwd_student_salesforce_ids.sql`
- View in `kipptaf_extracts`, contract enforced
- Sources `int_kippadb__roster`
- Filter: `contact_id is not null and ktc_status != 'MS8'`, giving 7,930 rows, one per Salesforce contact (uniqueness verified: 8,017 of 8,017)

### Columns (snake_case)

`student_name`, `last_name`, `first_name`, `birthdate`, `hs_class`, `region`, `salesforce_contact_id`, `student_number`, `record_type`, `ktc_status`

Disambiguation ladder, measured across the full student-contact set: name alone leaves 102 rows ambiguous; adding `hs_class` leaves 16; adding `region` leaves 8; `birthdate` closes the remainder.

### Population decisions

- `int_kippadb__roster` is the limiter. It omits roughly 2,113 Salesforce student contacts, including 977 grade-8 withdrawals whose `kipp_ms_graduate` flag is unset in Salesforce.
- `ktc_status != 'MS8'` drops 87 currently-enrolled 8th graders.
- Validated against award history: every Maher award ever made goes to a student this population keeps. The residual is 6 non-Maher awards across 4 students.
- Salesforce `record_type` is deliberately NOT a filter. It lags reality in both directions: 1,093 `TAF` students and 445 current 9th graders are still tagged `MS Student`, and 14 `HSG` are still tagged `HS Student`. It is carried as context only.
- The filter is an exclusion rather than the inclusion list used by `rpt_tableau__kfwd_dashboard`. The two are identical today and diverge only if a new status value is introduced. Exclusion fails safe: a student missing from the sheet causes the mis-tagging this model exists to prevent, while an extra middle schooler is harmless noise.

### Exposure

Sheet `1AAiEQ46OeMbgLjRawoFayfW-4WavtkNPwCvxMOkdSLo`, named range matching the model name, `kinds: [googlesheets]`. No cron: the model is a view and is current at read time, so cadence is the Google-side connected-sheet refresh.

### PII

Model-level `contains_pii: true` plus column-level tags on the six direct identifiers (`student_name`, `last_name`, `first_name`, `student_number`, `salesforce_contact_id`, `birthdate`). Access is restricted Google-side to CMO staff covered by data sharing agreements.

## Convention

`kfwd_` is the going-forward prefix for KIPP Forward Google Sheets extracts, superseding the mixed `kfwd_` and `kippfwd_` usage in the repo today.

## Out of scope

- Coupa transaction ingestion. Exports stay manual CSV.
- ID validation at load time, which is handled by a separate process.
- Pointing the `build_maher_upload_file.py` name resolver at this model. See #4277 and #4278.

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will publish `rpt_gsheets__kfwd_student_salesforce_ids` to a Google Sheet, exposing the Salesforce contact id for every KIPP Forward student so the person entering an emergency postsecondary aid transaction in Coupa can attach the id to it.

Today Coupa transactions carry no student identifier, so disbursements get matched to Salesforce contacts by name when the `KIPP_Aid__c` records are loaded. Across all 10,132 Salesforce student contacts — the set that name-matching runs against today — name alone is ambiguous for 102 contacts in 50 colliding name groups. Capturing the id at Coupa entry time means it rides along in the CSV export and no name matching is needed downstream.

One row per Salesforce contact, sourced from `int_kippadb__roster` filtered to `contact_id is not null and ktc_status != 'MS8'`, giving 7,930 rows.

### Population reasoning

- `int_kippadb__roster` is the limiter, so Salesforce contacts without a KIPP secondary enrollment do not appear.
- `ktc_status != 'MS8'` drops currently enrolled 8th graders. Validated against the full award history: **every Maher award ever made goes to a student this population keeps**; the residual is 6 non-Maher awards across 4 students.
- Salesforce `record_type` is deliberately **not** a filter. It lags real transitions in both directions — 1,093 `TAF` students and 445 currently enrolled 9th graders are still tagged `MS Student`, and 14 `HSG` are still tagged `HS Student`. It is carried as context only, and its column description says so.
- The filter is an exclusion rather than the inclusion list `rpt_tableau__kfwd_dashboard` uses. The two are identical today and diverge only if a new `ktc_status` value is introduced. Exclusion fails safe: a student missing from the lookup causes the mis-attribution this feed exists to prevent, while an extra middle schooler is inert.

### Disambiguation

Measured **on the 7,930 rows this model actually publishes**: name alone leaves 65 rows ambiguous, adding `cohort` leaves 8, adding `region` leaves 6, and `birthdate` closes the remainder. All three are fully populated, and each now carries a `not_null` check so that stays true.

`cohort` is the 9th-grade entry year — a distinct concept from graduation year, which reflects retentions and skipped grades. The two differ for 324 of these rows. They are equally effective as disambiguators (either one takes 65 down to 8), so the column carries cohort on the grounds that it is the stable identifier, not because it discriminates better.

### Convention

`kfwd_` is now recorded in `src/dbt/kipptaf/CLAUDE.md` as the going-forward prefix for KIPP Forward Google Sheets extracts. Existing `kippfwd_` models are not being renamed.

`CONTEXT.md` is new at the repo root — a glossary of the terms pinned down while designing this (student, Salesforce contact ID, KTC status vs record type, Maher Fund, cohort vs graduation year, region). It is a working vocabulary file for developers and agents, deliberately outside the MkDocs nav.

## AI Assistance

Claude Code co-authored this PR.

**Human-directed:** every design decision — population scope, that the roster is the limiter, filtering on `ktc_status` rather than record type, the column set, snake_case headers, the `kfwd_` naming convention, PII tagging scope, the decision to accept the excluded grade-8-withdrawal population without a tracking issue, and the cohort-versus-graduation-year correction.

**AI-assisted:** fact-finding against the warehouse (name-collision counts, record-type lag, null profiling, award-history validation of the population choice), then writing the model, properties YAML, exposure, and `CONTEXT.md`, and running the local build and lint.

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Two findings
      implemented, two answered with reasoning in a reply below.

### dbt

- [x] Include (or update) an
- [x] **Breaking change?** No. Additive only — no existing model, column, or
      exposure is modified, so reverting these commits is a complete rollback.
- [x] External sources — not applicable, none added or modified.

## Local validation

- `dbt build --select rpt_gsheets__kfwd_student_salesforce_ids --target dev --defer --favor-state` — view created, all 4 data tests PASS (`unique` on the contact id, `not_null` on `birthdate` / `cohort` / `region`).
- Contract confirmed independently via `INFORMATION_SCHEMA.COLUMNS`: all 10 columns present with the declared types.
- Dev view returns 7,930 rows, and its stored SQL resolves upstream to prod `kipptaf_kippadb.int_kippadb__roster` — so the count is not a stale-dev-copy artifact.
- `trunk check --force` over all changed files — no issues.

## Post-merge, owner action

The destination sheet needs a named range `rpt_gsheets__kfwd_student_salesforce_ids` pointed at `kipptaf_extracts.rpt_gsheets__kfwd_student_salesforce_ids`, plus CMO-only sharing. This cannot be done before merge, because the prod view does not exist until Dagster materializes the new asset.

## CI checks

- [x] **Trunk** — passed on `ea1246bbe`.
- [x] **dbt Cloud** — passed on `ea1246bbe`.
- [x] **Dagster Cloud** — not triggered, no Dagster paths changed.

Refs #4798

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -47,15 +47,16 @@ urgent, unplanned need of a single student.
 
 _Avoid_: emergency aid, hardship grant, micro-grant
 
-**Disbursement**: One payment of Maher Fund money made to or on behalf of one
-student.
+**Cohort**: The year-group a student enters 9th grade with. Fixed at entry, so a
+retention or a skipped grade does not change it.
 
-_Avoid_: award, grant, aid payment
+_Avoid_: class, graduating class, graduation year
 
-**Cohort**: The year-group a student is associated with for KIPP high school
-graduation.
+**Graduation year**: The calendar year a student graduates high school, or is
+projected to. Reflects retentions and skipped grades, so it can differ from the
+student's cohort.
 
-_Avoid_: class, graduating class, grad year
+_Avoid_: cohort, grad year
 
 **Region**: One of the four cities whose schools the network operates — Newark,
 Camden, Miami, and Paterson.
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,63 @@
+# KIPP TEAM and Family data platform
+
+The shared vocabulary for the network's data models. Terms are added as they get
+pinned down in design work, so this file grows in clusters rather than arriving
+complete.
+
+## Language
+
+### KIPP Forward and postsecondary aid
+
+**KIPP Forward**: The network's postsecondary program, supporting students from
+middle school through college and career. Its legacy name survives inside
+identifiers such as `ktc_status`.
+
+_Avoid_: KTC, KIPP Through College
+
+**Student**: A person holding a Salesforce contact record whose record type is
+one of the student lifecycle stages. Parents, staff, and other contacts on the
+same object are not students.
+
+_Avoid_: alum, alumni, contact
+
+**Salesforce contact ID**: The eighteen-character Salesforce identifier for a
+contact record, and the network's canonical identifier for a student outside the
+student information system. The SIS holds no identifier of its own for a
+Salesforce record.
+
+_Avoid_: SF ID, contact ID, `salesforce_id`
+
+**KTC status**: Where a student sits relative to KIPP secondary enrollment —
+enrolled at a given grade level, graduated from a KIPP high school, or a KIPP
+middle school graduate who has not graduated from a KIPP high school.
+
+_Avoid_: postsecondary status, record type, lifecycle stage
+
+**Record type**: The lifecycle stage a Salesforce contact record was last set
+to. It describes the state of the record, not the state of the student, and lags
+real transitions in both directions.
+
+_Avoid_: status, KTC status
+
+**Maher Fund**: The fund from which the network pays discretionary awards to
+KIPP Forward students, divided into emergency and enrichment purposes.
+
+**Emergency funding request**: A request to pay Maher Fund money toward an
+urgent, unplanned need of a single student.
+
+_Avoid_: emergency aid, hardship grant, micro-grant
+
+**Disbursement**: One payment of Maher Fund money made to or on behalf of one
+student.
+
+_Avoid_: award, grant, aid payment
+
+**Cohort**: The year-group a student is associated with for KIPP high school
+graduation.
+
+_Avoid_: class, graduating class, grad year
+
+**Region**: One of the four cities whose schools the network operates — Newark,
+Camden, Miami, and Paterson.
+
+_Avoid_: district, market, area
--- a/src/dbt/kipptaf/CLAUDE.md
+++ b/src/dbt/kipptaf/CLAUDE.md
@@ -504,3 +504,7 @@ snapshot — before removing it.
   `models/marts/`. Actively being developed; see
   `src/dbt/kipptaf/models/marts/CLAUDE.md` for column-naming rubric, hash-change
   discipline, and strict-chain rules.
+
+New KIPP Forward Google Sheets extracts take the `rpt_gsheets__kfwd_` prefix.
+Existing models use both `kfwd_` and `kippfwd_`; `kfwd_` is the going-forward
+choice, and the older ones are not being renamed.
--- a/src/dbt/kipptaf/models/exposures/google-sheets.yml
+++ b/src/dbt/kipptaf/models/exposures/google-sheets.yml
@@ -77,6 +77,19 @@ exposures:
         dagster:
           kinds:
             - googlesheets
+  - name: kfwd_student_salesforce_ids
+    label: KIPP Forward Student Salesforce IDs
+    type: application
+    owner:
+      name: Data Team
+    url: https://docs.google.com/spreadsheets/d/1AAiEQ46OeMbgLjRawoFayfW-4WavtkNPwCvxMOkdSLo
+    depends_on:
+      - ref("rpt_gsheets__kfwd_student_salesforce_ids")
+    config:
+      meta:
+        dagster:
+          kinds:
+            - googlesheets
   - name: camden_sar_accommodations_errors20241031
     label: "[Camden] SAR Accommodations Errors - 20241031"
     type: application
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__kfwd_student_salesforce_ids.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__kfwd_student_salesforce_ids.yml
@@ -9,8 +9,9 @@ models:
       int_kippadb__roster because no aid request originates from one; the roster
       itself only covers students with a KIPP secondary enrollment, so a
       Salesforce contact without one does not appear here. The remaining columns
-      exist to tell same-name students apart — graduation cohort separates most
-      pairs, region and birth date separate the rest.
+      exist to tell same-name students apart — cohort separates most pairs,
+      region and birth date separate the rest, which is why all three carry a
+      null check.
     config:
       meta:
         contains_pii: true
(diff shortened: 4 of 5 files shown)


CONTEXT.md
```diff
@@ -0,0 +1,63 @@
...
+
+**Disbursement**: One payment of Maher Fund money made to or on behalf of one
+student.
+
+_Avoid_: award, grant, aid payment
+
+**Cohort**: The year-group a student is associated with for KIPP high school
+graduation.
+
+_Avoid_: class, graduating class, grad year
```
anthonygwalters: we use cohort and grad year as separate things. cohort is based on entry year into 9th grade whereas graduation year is the calendar year when they either graduate or are projected to, taking into account retentions, skipped grades, etc.

CONTEXT.md
```diff
@@ -0,0 +1,63 @@
...
+
+**Emergency funding request**: A request to pay Maher Fund money toward an
+urgent, unplanned need of a single student.
+
+_Avoid_: emergency aid, hardship grant, micro-grant
+
+**Disbursement**: One payment of Maher Fund money made to or on behalf of one
+student.
+
+_Avoid_: award, grant, aid payment
```
anthonygwalters: remove - not enough use cases

CONTEXT.md
anthonygwalters: Corrected in `04ad0cf8e`. Cohort and graduation year are now separate entries — cohort as the 9th-grade entry year, fixed at entry; graduation year as the actual or projected calendar year, reflecting retentions and skipped grades. Each lists the other under `_Avoid_` so the conflation is explicitly ruled out rather than merely unmentioned.

This also caught a real defect in the model, not just the glossary. The column was named `hs_class` and described as "graduation cohort" — but it is sourced from `ktc_cohort`, i.e. cohort. Renamed to `cohort` and the description rewritten to say it is fixed at 9th-grade entry and can differ from the projected graduation year. On the 7,930 rows this model publishes, the two values differ for **324** of them, so the distinction was load-bearing rather than pedantic.

One thing worth your call. As disambiguators the two are equally effective: on this population, name alone leaves 65 rows ambiguous, and adding either cohort or graduation year leaves 8. So the choice is purely about which one a finance reader recognizes on an incoming request. I kept cohort on the grounds that it is the stable identifier, but `contact_graduation_year` is already on `int_kippadb__roster` and equally well populated if you would rather show that instead, or both. Say which and it is a one-line change — worth deciding before merge, since the sheet is not wired yet and the column names are free to move until it is.

CONTEXT.md
anthonygwalters: Removed in `04ad0cf8e`.

Verified before removing, and the case is stronger than "not enough use cases": `disbursement` appears **nowhere** in the repo, while `award` — which I had listed under `_Avoid_` — appears 88+ times as an identifier (`award_date`, `award_letter`, `award_count`, `subsequent_financial_aid_award`, and `aid_name` on the Maher report itself). The entry canonized a term nobody uses and flagged the one the codebase actually standardized on. Deleting it rather than inverting it, since "award" is already unambiguous in context and does not need a glossary slot to defend it.

anthonygwalters:
## Responses to the automated review

**1 — `not_null` on the disambiguation columns. Implemented in `04ad0cf8e`.**

Correct, and it draws the right distinction: `salesforce_contact_id` *is* non-null by construction (the model's own `where contact_id is not null`), so a `not_null` there would be vacuous — but `birthdate`, `cohort`, and `region` are nullable at the SQL level and the disambiguation design leans on them. Ran the verification the finding asked for before adding anything: 0 nulls across all three over 7,930 rows. Tests added and passing locally.

Left deliberately at the project's `warn` default rather than `error`. If a Salesforce birth date goes blank, a warning is the right outcome and a failed build is not — an offline lookup pushes people back to name matching, which causes more mis-attribution than a partially degraded lookup does. Warn-level `not_null` to flag a null-column condition is also the documented repo idiom (`src/dbt/CLAUDE.md` → Test config defaults).

**2 — `unique` severity. Not changing, and the stated rationale does not hold.**

The finding describes this as "the only test guarding the exact ambiguity problem the feed exists to eliminate." It isn't. `unique` on `salesforce_contact_id` guards **grain**, not name ambiguity. A duplicate contact id means the same id appears on two rows, so whoever copies it still gets the right id. The actual risk is two *different* students sharing a name — which `unique` cannot detect by construction, and which is exactly what `cohort`, `region`, and `birthdate` are carried for.

Grain is separately guarded upstream at error severity: `int_kippadb__roster` carries `unique` on `student_number` with `config: severity: error`. Leaving this at the documented `rpt_` warn allowance.

**3 — PII tags on `cohort` and `region`. Not changing; deliberate.**

Model-level `contains_pii: true` plus column-level tags on the six direct identifiers, matching the narrower powerschool-staging precedent rather than tagging indirect identifiers. Root `CLAUDE.md` → PII reference notes that the staging precedent is narrower than the full FERPA list and says to confirm scope with the owner for a PII-heavy new model; that was done before implementation. Worth stating for future readers that this tag is a code-level marker for maintainers and does not gate access — access is Google-side, restricted to CMO staff covered by data sharing agreements.

**4 — `CONTEXT.md` at the repo root. Confirmed intentional.**

Root is where the `domain-modeling` skill expects a single-context glossary; `CONTEXT-MAP.md` is used only when a repo carries multiple contexts. It is a working vocabulary file for developers and agents rather than a published reference page, so it sits outside the `mkdocs.yml` nav by design.

---

For the record, the human review caught a defect this pass did not: the `hs_class` column conflated cohort with graduation year, which are distinct concepts here and differ for 324 of the 7,930 rows. Renamed to `cohort` with a corrected description — details in the inline thread. The PR body's disambiguation figures are also corrected; they previously described the full 10,132-contact set rather than the rows this model publishes.

Issue #4796, opened 2026-08-10
feat(kipptaf): choose replacement metrics for Miami middle school GPA on the KIPP Forward roster


CGibson17:
## Problem

`gpa_cumulative` and `gpa_y1` in `rpt_gsheets__kippfwd_miami_roster` are cast null as of #4782.

This is **not** a temporary modeling gap waiting on a Focus GPA model. Miami middle school will no longer produce a GPA at all, so these two columns have no future source and cannot simply be rebuilt against Focus.

## What is needed

A decision from KIPP Forward on what replaces GPA as the academic-standing signal for Miami grades 7 and 8 on this roster. Candidates worth evaluating against what Focus actually holds:

- Course-grade based measures from `report_card_grades`, if the grain and coverage support it
- FAST achievement levels, already on the roster as `ela_pm1` through `math_pm3`
- Attendance-based standing, once Focus attendance ingestion exists
- Something else entirely, if the program has already moved off a GPA-shaped metric

Once the replacement is chosen, decide whether to repurpose the two existing columns or retire them and add new ones. Either path changes the sheet's header row, so it needs KIPP Forward sign-off before implementation.

## Interim state

Both columns remain in the contract as typed nulls so the Google Sheet's column layout does not shift for consumers. Each carries a `TODO` comment in `rpt_gsheets__kippfwd_miami_roster` pointing here.

## Context

Split out of #4782, which re-sourced the roster from Focus. Note that Focus `report_card_grades` currently holds roughly 1,300 rows network-wide, so it is thin regardless of which direction is chosen.

cbini:
Folding into #4795.

Both issues split out of #4782, both cover a column cast null on `rpt_gsheets__kippfwd_miami_roster`, and both are blocked on a decision outside engineering rather than on modeling work. They also change the same thing: the sheet's header row, which needs KIPP Forward sign-off once either is resolved.

Doing them as one conversation and one PR that removes both `TODO` comments.

Carrying forward to #4795: `gpa_cumulative` and `gpa_y1` need a replacement academic-standing signal for Miami grades 7-8, not a Focus GPA model — Miami middle school will not produce a GPA at all. Candidates listed there. Note `report_card_grades` holds roughly 1,300 rows network-wide, so it is thin regardless of direction.

Closing as duplicate of #4795.

Issue #4795, opened 2026-08-10
feat(kipptaf): source an advisory assignment for Miami grades 7-8


CGibson17:
## Problem

`advisor_lastfirst` in `rpt_gsheets__kippfwd_miami_roster` is cast null as of #4782, because Focus has no advisory structure for grades 7 and 8.

## What was checked

- Homeroom courses exist only for K-5 — `Homeroom - Kindergarten` through `Homeroom - 5thGrade`. There is no grade 7-8 equivalent.
- `student_enrollment.team_id` is null for **all** 365 Miami grade 7-8 students, in both SY2025 and SY2026.
- No course in the grade 7-8 schedule functions as an advisory. The largest by enrollment is `M/J FDN SKL MATH 6-8` at 262 students, which is an academic course.
- The `SISStudent` custom fields `Advisor Name` and `Advisor Email` exist in the Focus catalog but are type `computed`, so Focus never stores a value in them.

## What is needed

Either an advisory or homeroom course section for grades 7-8 in Focus that can be joined through `schedule` to `course_periods` to `users`, or a populated stored custom field carrying the advisor.

This is likely an Ops or school-operations configuration question before it is a modeling question — the data does not exist to be modeled today.

Once a source exists, remove the null cast and its `TODO` comment from `rpt_gsheets__kippfwd_miami_roster`.

## Context

Split out of #4782, which re-sourced the roster from Focus.

Issue #4794, opened 2026-08-10
fix(focus): enroll_status in int_focus__student_enrollment misreads closed years as transferred out


CGibson17:
## Problem

`src/dbt/focus/models/intermediate/int_focus__student_enrollment.sql` derives `enroll_status` from the presence of a drop code:

```sql
case
    when dc.grad_type = 'graduated'
    then 3
    when e.drop_code is not null
    then 2
    else 0
end as enroll_status
```

Focus stamps a withdrawal code on nearly every enrollment span when a school year closes out. For Miami grades 7 and 8 in SY2025, 227 spans carry `W01` (In School Transfer) and 133 carry `W02` (In District Transfer). Neither is a withdrawal — both are year-end rollover codes.

So any **closed** school year reads as a near-total exodus.

## Evidence

`kipptaf_focus.int_focus__student_enrollments`, grades 7 and 8, `rn_year = 1`, against the PowerSchool equivalent for the same students and year:

| Year | Focus `enroll_status` | PowerSchool actual |
| --- | --- | --- |
| AY2025 | 0 = 2, 2 = **361**, 3 = 2 | 0 = 222, 2 = 143 |
| AY2026 | 0 = 353, 2 = 22 | no PowerSchool data |

AY2026 looks healthy only because that year has not closed yet, so no rollover codes have been stamped. It will flip the same way at year end.

## Suggested fix

Derive from whether the student holds an open enrollment in the latest school year Focus holds, rather than from drop-code presence:

| Output | Condition |
| --- | --- |
| `-1` | `start_date` is in the future |
| `3` | drop code `grad_type` is graduated |
| `0` | an enrollment row exists in the latest `syear` with `drop_code is null` |
| `2` | otherwise |

Validated against PowerSchool on the same population: this reproduces `enroll_status` for 341 of 365 students. The 24 disagreements are explainable — PowerSchool froze at the Focus cutover, so Focus is the more current value in both directions.

Anchor on the max `syear` present in Focus, **not** `var("current_academic_year")` — that var is currently 2025 and has not rolled for July 2026, so using it would mark nearly every student withdrawn.

This mirrors how PowerSchool `enroll_status` behaves: a current-status value copied onto every historical row, not a per-stint status.

## Scope

`int_focus__student_enrollment` is a source-package model consumed by `kippmiami`, which kipptaf reads via `sources-kippmiami.yml`. A fix needs the district-then-kipptaf two-PR sequence, and affects every consumer of the column.

## Context

[...]

PR #4806, opened 2026-08-10
feat(kipptaf): source kippfwd_miami_roster from Focus instead of PowerSchool

### LINKED ISSUE #4782: feat(kipptaf): source kippfwd_miami_roster from Focus instead of PowerSchool
## Problem

`rpt_gsheets__kippfwd_miami_roster` sources every student column from `base_powerschool__student_enrollments` via `int_extracts__student_enrollments`. Miami PowerSchool has no data past `academic_year = 2025`. Focus already holds `academic_year = 2026` — 194 students in grade 7 and 179 in grade 8.

At the July `current_academic_year` rollover to 2026, this extract goes empty for the current year.

## Approach

kipptaf already exposes a complete Focus layer, built by the `focus` package inside `kippmiami` and read through `sources-kippmiami.yml` (source `kippmiami_focus`). Three of those models cover nearly the whole extract, so this is a rewrite of one reporting view — **no new models, no new sources**.

| Model | Supplies |
| --- | --- |
| `int_focus__student_enrollments` | `academic_year`, `student_name`, `grade_level`, `fteid`, `rn_year`, `exitcode` |
| `int_focus__students` | `powerschool_id`, `disis_id`, `sex_label`, `ese_fefp_code` |
| `int_focus__student_contacts` | `contact_name`, `email`, `phone_home`, `phone_mobile`, `sort_order` |

The contract shape stays identical — all 30 columns keep their names, types, and select order, so the Google Sheet and the exposure are unchanged.

### Two naming traps

`int_focus__student_enrollments.student_number` is the **Focus** `student_id`, not the PowerSchool `student_number` — it is the join key to the other two models. The roster's `ps_id` must come from `int_focus__students.powerschool_id`.

`int_focus__student_contacts.sort_order` is a STRING.

## Landmine — enroll_status must be derived locally

`int_focus__student_enrollments.enroll_status` is wrong for any closed school year. It derives from drop-code presence, and Focus stamps a rollover withdrawal code on nearly every span at year end (W01 In School Transfer, W02 In District Transfer):

| Year | Upstream value | PowerSchool actual |
| --- | --- | --- |
| AY2025 | 0 = 2, 2 = **361**, 3 = 2 | 0 = 222, 2 = 143 |
| AY2026 | 0 = 353, 2 = 22 | no data |

AY2026 reads correctly only because that year has not closed yet. Filed separately as #4794. This extract derives the value locally until that lands, anchoring on the max `academic_year` in Focus rather than `current_academic_year`, which reads 2025 and has not rolled for July 2026.

## Columns with no Focus source

| Column | Treatment | Tracked |
| --- | --- | --- |
| `previous_year_ada` | Keep the PowerSchool join on `powerschool_id`. Focus `attendance_completed` holds 1 row and `fl_days_present` is entirely null | breaks at the SY2027 rollover, not SY2026 |
| `advisor_lastfirst` | Cast null. Homeroom courses exist only for K-5, `team_id` is null for all 365 students, and no grade 7-8 course functions as an advisory | #4795 |
| `gpa_cumulative`, `gpa_y1` | Cast null. Miami middle school will no longer produce a GPA at all, so these need a replacement metric rather than a rebuild | #4796 |

## Other known limitations

- **`ps_id` is null for roughly 80 students** across both years — post-cutover enrollees with no `powerschool_id` yet. They also get no `previous_year_ada`, which is correct, and drop out of the FAST join, which needs a FLEID.
- **`iep_status` has 79 percent recall.** Focus ESE FEFP Code finds 42 of the 53 IEP students PowerSchool identifies for SY2025 grades 7 and 8. Accepted, documented in the column description.
- **`contact_1_phone_home` is sparse** — 55 of 338 primary contacts. Focus mostly stores mobile numbers.

## Grain

One row per student per `academic_year`, via the upstream `rn_year = 1` filter — the same mechanism the current model uses. The two `sort_order` contact CTEs must each join once or the roster fans out.

## Design and plan

- Spec: `docs/superpowers/specs/2026-08-10-focus-miami-kippfwd-roster-design.md`
- Plan: `docs/superpowers/plans/2026-08-10-focus-miami-kippfwd-roster.md`

An earlier draft of this issue proposed reading the raw `dagster_kippmiami_dlt_focus` landing tables directly at the kipptaf level and adding two new intermediates. That was written against a stale checkout and missed kipptaf's existing Focus layer; commit `4bad76cad` on `main` had already closed that pattern. The spec records the correction.

CGibson17:
# Pull Request

## Summary and Motivation

When merged, this pull request will re-source `rpt_gsheets__kippfwd_miami_roster` from kipptaf's existing Focus layer, so the extract keeps producing data now that Miami PowerSchool has gone dark.

`current_academic_year` rolled to 2026 on 2026-07-17 (`7cfd878bf`). Miami PowerSchool holds nothing past AY2025, so prod currently emits **zero AY2026 rows**. This is a live gap, not a preventive change.

No new models and no new sources — this is a rewrite of one reporting view over `int_focus__student_enrollments`, `int_focus__students`, and `int_focus__student_contacts`. The 29-column contract is unchanged in name, type, and select order, so the Google Sheet and the exposure are untouched.

Closes #4782. Related: #4794, #4795, #4796.

### Verification, both years

| Check | AY2025 | AY2026 |
| --- | --- | --- |
| Rows | 365 (gr 7 = 184, gr 8 = 181) | 375 (gr 7 = 195, gr 8 = 180) |
| Grain, rows vs distinct keys | 365 / 365 | 375 / 375 |
| `previous_year_ada` populated | 269 — identical to prod | 295 |
| `advisor_lastfirst` / `gpa_cumulative` / `gpa_y1` | 362 / 338 / 303 — identical to prod | 0 (PowerSchool has no AY2026) |
| `enroll_status` (-1 / 0 / 2 / 3) | 0 / 219 / 144 / 2 | 374 / 0 / 1 / 0 |
| `iep_status` domain | `Has IEP`, `No IEP` | same |
| `fleid` null | 5 | 81 |

### Bug fixed in passing

The current prod view **fans out** — 1,274 rows for 365 AY2025 students. Its `gpa_term` CTE was dead code, so the join went to `int_powerschool__gpa_term` directly with no `is_current` filter, multiplying every row 4x. This PR joins the filtered CTE (verified 1:1 — 4,128 rows over 4,128 distinct keys) and emits 365.

### Hybrid sourcing

`advisor_lastfirst`, `gpa_cumulative`, `gpa_y1`, and `iep_status` read from PowerSchool for the years PowerSchool covers, and from Focus onward. There is no explicit year gate — PowerSchool holds no Miami AY2026 row, so the same-year left join naturally yields null and the hybrid ages out on its own.

This preserves real values for the prior-year half of the sheet, and keeps PowerSchool's authoritative IEP flag for AY2025 rather than substituting a 79%-recall proxy on a year PowerSchool covers correctly.

### enroll_status is derived locally, deliberately

`int_focus__student_enrollments.enroll_status` is wrong for any closed school year — it derives from drop-code presence, and Focus stamps a rollover withdrawal code on nearly every span at year end. For AY2025 it reads 361 of 365 students as transferred out against PowerSchool's 143. Filed as #4794; this view derives the value locally until that lands, anchored on `current_academic_year`.

**Disclosure — the current-year roster reads `-1` until the first day of school.** Miami's first day is roughly Aug 17, so 374 of 375 AY2026 rows are `-1` (pre-registered) and **no student reads `0`**. This is PowerSchool-faithful and self-heals once school starts, but any consumer filtering `enroll_status = 0` sees an empty current-year roster until then. KIPP Forward should expect this every August.

### Known limitations

- **`fleid` is null for 81 of 375 AY2026 rows (22%)**, against 5 of 365 in AY2025. It is the only grain-test key and both FAST join keys, so the grain test covers 78% of current-year rows and those 81 students receive no FAST scores. `mdcps_id` is null for exactly the same 81 and `ps_id` for 80 — those rows carry no usable identifier except `lastfirst`. Not recoverable by coalescing across years. A second grain key would require adding Focus's `student_id` to the contract, which is a KIPP Forward decision.
- `previous_year_ada` is bounded by `powerschool_id` coverage, and empties for good at the SY2027 rollover when the prior year is itself a Focus year.
- `iep_status` for AY2026 onward uses Focus ESE FEFP Code, which identifies 42 of the 53 students PowerSchool flagged for AY2025 — 79% recall.

### One change KIPP Forward will see in the sheet

`iep_status` changes value domain from PowerSchool's raw `spedlep` to `Has IEP` / `No IEP`. Anything filtering on `SPED` needs updating. This was a deliberate choice for a clearer value over a stable domain.

Also: 104 of 365 AY2025 `lastfirst` values differ from the PowerSchool spellings (68 punctuation or spacing, 36 substantive). Inherent to re-sourcing.

### Rollback plan

No contract change, so revert is a straight `git revert` of this PR. Note that reverting also **restores the 3.5x fan-out** described above, so a revert should be paired with fixing the `gpa_term` join on the restored version.

## AI Assistance

Authored by Claude Code, human-directed throughout. The engineer chose the sourcing approach, the hybrid, the `iep_status` domain, and the decision to disclose the pre-first-day `-1` rather than force `0`.

Two rounds of review found real defects, both of which came from the same root cause — analysis run against a stale checkout on an older branch. That produced a first design targeting the raw dlt landing tables (reverted), a false claim that Focus contacts were unpopulated, and the false claim that `current_academic_year` had not rolled. The spec records the corrections.

The second review also caught that all parity verification had run against AY2025 — the year PowerSchool already covers — which is how a broken AY2026 shipped. Every figure in this PR body is now reported per year.

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Triaged in the reviewer's top-level comment; all valid items addressed or documented

### dbt

- [x] Properties file updated — descriptions for all 29 columns, PII tags, a scoped uniqueness test, `accepted_values` on `enroll_status` and `iep_status`, and a singular test asserting the current academic year is non-empty
- [x] Exposure unchanged — `ref()` name and contract are both stable
- [x] **Breaking change?** No columns renamed or removed. The `iep_status` domain change is called out above

### Docs

- [x] No schedule, sensor, or integration changes. Design spec and implementation plan included under `docs/superpowers/`

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes


anthonygwalters:
## Adversarial review — spec, plan, and shipped model

Reviewed the issue, the design spec, the implementation plan, and the model, with every data claim re-derived against prod BigQuery. Summary up front: **the re-sourcing design is sound and the AY2025 parity work is real, but every parity check in the plan ran against AY2025 — the year PowerSchool already covers correctly.** On AY2026, the only reason this PR exists, three columns are empty or wrong, and the spec documents two of them as working.

One framing correction: `current_academic_year` rolled to 2026 in `7cfd878bf` on 2026-07-17, three weeks before this PR. Prod holds 1,274 rows for AY2025 and **zero** for AY2026. This is a live incident, not a preventive change.

---

### Blocking

#### 1. `previous_year_ada` is null for 100% of AY2026 rows

The join keys `e.academic_year = ada.academic_year` and reads `ada_unweighted_year_prev` off the *same* year's PowerSchool row. PowerSchool has no Miami AY2026 row, so it misses entirely.

| Year | rows | `previous_year_ada` populated |
| --- | --- | --- |
| 2025 | 365 | 269 |
| 2026 | 375 | **0** |

The spec's *Known limitations* asserts the opposite: "`previous_year_ada` empties at the SY2027 rollover, not SY2026. For SY2026 rows the prior year is SY2025, which PowerSchool holds." The data is there — the model just doesn't reach it.

The fix is exact, not approximate. I confirmed `ada_unweighted_year_prev` on year N equals `unweighted_ada` on year N-1 for 269/269 rows, so re-keying to `e.academic_year - 1 = ada.academic_year` and reading `ada.unweighted_ada` is **byte-identical for AY2025** and recovers ~295 AY2026 rows (the ones with a `powerschool_id`).

#### 2. `enroll_status` is `-1` for 374 of 375 AY2026 rows

Reconstructed the CASE against prod:

| Year | -1 | 0 | 2 | 3 |
| 2025 | 0 | 219 | 144 | 2 |
| 2026 | **374** | **0** | 1 | 0 |

AY2025 matches the PR body exactly. AY2026 does not appear in the PR body or the plan. Miami's first day is roughly Aug 17, so `e.startdate > current_date()` fires for every current-year span.

It is PowerSchool-faithful and it self-heals once school starts, but any consumer filtering `enroll_status = 0` gets **zero** current-year students — the sheet is still functionally empty, just with rows in it. `src/dbt/kipptaf/CLAUDE.md` is explicit: "`-1` is pre-registered (not yet enrolled) — never report against either." At minimum this needs to be disclosed to KIPP Forward; better, decide whether a pre-first-day roster should read `0`.

The plan called this "the single most important check in the plan" and ran it on AY2025 only.

Related, undocumented: the derived column mixes grains. `0` and `2` are student-level (correct — it mirrors the PowerSchool behavior `kipptaf/CLAUDE.md` documents), while `-1` and `3` are row-level.

#### 3. `fleid` is null for 81 of 375 AY2026 rows (22%), undisclosed

AY2025 has 5 nulls, matching prod. But `fleid` is now load-bearing three ways: it is the **only** grain-test key, the FAST join key, and the prior-year FAST join key. So:

- `dbt_utils.unique_combination_of_columns` covers 78% of current-year rows.
- Those 81 students get no FAST scores in either the current or prior-year columns.

Not recoverable by coalescing across years — none of the 81 has a prior-year Focus row with a non-null `fteid`. It is a Focus data-entry lag, not a re-sourcing defect, but it belongs in the PR body next to the `ps_id` coverage number and it argues for a second, non-null grain key.

---

### Design decisions worth revisiting

#### 4. Nulling `advisor_lastfirst`, `gpa_cumulative`, and `gpa_y1` discards AY2025 values PowerSchool still has

Populated in prod today: advisor 1,271 rows, `gpa_cumulative` 1,247, `gpa_y1` 1,212. The sheet shows current + prior year, and the prior year is PowerSchool-covered.

A hybrid — Focus for AY2026 and later, PowerSchool for AY2025 and earlier — would preserve advisor, both GPAs, and an authoritative `iep_status` for half the roster. The spec frames these as unsourceable, which is true only for the current year. This may still be the right call for simplicity, but it is a KIPP Forward decision and the spec never puts the option on the table.

#### 5. The `iep_status` domain break is avoidable

Prod's domain is already exactly `{'SPED', 'No IEP'}`. Emitting `'SPED'` instead of `'Has IEP'` keeps the domain 100% stable and removes the consumer breakage the PR flags as a known cost.

Separately: for AY2025 rows this swaps PowerSchool's authoritative flag for a 79%-recall proxy. I verified that recall precisely — 53 students flagged `SPED` in PowerSchool, 42 caught by Focus ESE FEFP Code, 11 missed, 1 Focus-only — so 11 AY2025 students flip from correct to `No IEP`.

---

### Spec and plan integrity

Two load-bearing factual claims appear in all three artifacts (issue #4782, spec, plan). Both are wrong.

#### 6. "`int_focus__student_contacts.sort_order` is a STRING"

It is `NUMERIC`. The plan's prescribed `where sort_order = '1'` would have **failed the build** — BigQuery has no matching signature for `NUMERIC = STRING`. The shipped code correctly uses `= 1`, but the divergence from the plan is silent and the spec's "two naming traps" section is still wrong on disk.

#### 7. "`current_academic_year` reads 2025 and has not rolled for July 2026"

It has read 2026 since 2026-07-17. This voids the stated rationale for the `latest_year` / `open_enrollment` pair, and it made the plan's acceptance criterion actively misleading:

> "Zero rows means the `academic_year >= current_academic_year - 1` filter excluded them because the var still reads 2025 — correct behavior for today."

An implementer following that instruction would have accepted an empty current year as success.

#### 8. The `latest_year` anchor is less stable than the var it replaced, and defers the failure by one year

Focus holds grade 7/8 enrollments back to AY2020 (102-301 rows per year), and exit codes are stamped on essentially every closed span — **AY2018 through AY2024 have zero null `exitcode`s**. Consequences:

- **No minimum-population guard.** One stray AY2027 row flips `max(academic_year)`, collapses `open_enrollment` to that single student, and flips roughly 739 of 740 rows to "withdrawn."
- **The failure recurs.** When AY2026 closes, the all-withdrawn condition #4794 describes returns unless AY2027 spans land first. The anchor only works while `max(academic_year)` is an *open* year, and nothing asserts that.

Using the var for both the status derivation and the row filter is the more stable option, not less.

#### 9. Nothing durable guards any of this

The plan's most important check is a one-time manual query. Missing:

- `accepted_values` on `enroll_status` (`-1, 0, 2, 3`) and `iep_status`.
- Anything asserting the current academic year is non-empty — which is the entire premise of this PR. The next silent regression will look exactly like the one being fixed.

#### 10. Smaller items

- Issue and plan say "30 columns"; there are 29.
- The rollback plan omits that `git revert` restores the 3.5x fan-out.
- The plan mandated a `trunk-ignore(sqlfluff/ST06)`; the shipped file has none. Fine today (Trunk CI is green on `eef68fc`), but per `src/dbt/focus/CLAUDE.md` ST06 firing is expression-shape-dependent, so a later edit will trip it and invite a column reorder that silently reshuffles the sheet.
- Two unreported AY2025 diffs: `mdcps_id` differs for 4 students, `gender` for 2.

---

### On the automated review comment above

For triage, with everything re-checked against prod:

| Item | Verdict |
| --- | --- |
| 1 — ST06 violation, likely trunk failure | **False positive.** Trunk Check is `success` on `eef68fc`. The job noted it could not run trunk. |
| 2 — contact `sort_order` fan-out | **Valid, latent.** Zero collisions in current data; the grain test catches it only for non-null `fleid`. |
| 3 — var reused for the row filter, `ref()` 3x | **Valid, under-read.** Its open question resolves to yes: Focus has AY2020-2024 grade 7/8 rows. But its proposed fix (drive the filter off `latest_year`) makes both sites data-driven; see item 8 — the var is the stable side. |
| 4 — student-level `enroll_status` undocumented | Valid, minor. |
| 5 — spec `sort_order` claim wrong | Valid, and confirmed rather than low-confidence. |

---

### What holds up

- **The fan-out catch is real and correctly diagnosed** — 1,274 rows for 365 distinct students, dead `gpa_term` CTE, no `is_current` filter. Verified.
- **Grain is clean.** Rebuilt the full join chain including both contact slots: 740 rows, 740 distinct `(academic_year, student_number)`.
- Every AY2025 figure in the PR body is accurate: the 219/144/2 split, 104 name diffs, ADA identical, 42/53 IEP recall, contact coverage.
- Dropping `_dbt_source_project` from the FAST pivot is safe — 19,006 pivot cells, zero duplicate or conflicting.
- Contact slots as separate single-join CTEs, `ON` vs `WHERE` placement, PII tagging scope, and the dbt 1.11+ `arguments:` nesting all follow the documented conventions.
- The spec's "Corrections to the first draft" section is good practice and should stay.

### Suggested sequencing

1. Fix `previous_year_ada` (item 1) — small, verified, byte-identical on AY2025.
2. Decide `enroll_status` behavior before first day (item 2), then disclose whichever way it lands.
3. Add the `fleid` coverage number to the PR body and add `accepted_values` tests (items 3, 9).
4. Correct the two false claims in the spec and issue (items 6, 7) so the record is accurate.
5. Items 4, 5, and 8 are judgement calls for the author and KIPP Forward.

---

Produced with Claude Code as an adversarial review pass. Every count, type, and parity figure above was re-derived against prod BigQuery rather than taken from the PR body; anything I could not verify is not asserted.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
```diff
@@ -30,30 +26,83 @@ with
...
+        select max(academic_year) as academic_year,
+        from {{ ref("int_focus__student_enrollments") }}
+    ),
+
+    open_enrollment as (
+        select distinct e.student_number,
+        from {{ ref("int_focus__student_enrollments") }} as e
+        inner join latest_year as ly on e.academic_year = ly.academic_year
+        where e.exitcode is null
+    ),
```
anthonygwalters, line 38: Two problems with this anchor.

**1. The premise is wrong.** The comment says `var('current_academic_year')` 'lags the July rollover'. It rolled to 2026 in `7cfd878bf` on 2026-07-17, three weeks before this PR. The same claim appears in the spec and in #4782, and it is the entire stated justification for these two CTEs. Line 147 already trusts the var for row scoping, so the model is internally inconsistent about whether the var is reliable.

**2. No minimum-population guard, and the failure recurs.** Focus stamps exit codes on essentially every closed span — AY2018 through AY2024 have **zero** null `exitcode`s. Two consequences:

- One stray AY2027 row flips `max(academic_year)`, collapses `open_enrollment` to that single student, and flips roughly 739 of 740 rows to withdrawn.
- When AY2026 closes, the all-withdrawn condition #4794 describes returns unless AY2027 spans land first. This anchor only works while `max(academic_year)` is an *open* year, and nothing asserts that.

Anchoring both sites on the var is the more stable option, not less.

Whichever you pick, an `accepted_values` test on `enroll_status` plus something asserting the current academic year is non-empty would catch the regression. Right now nothing does — and a non-empty current year is the entire premise of the PR.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
```diff
@@ -30,30 +26,83 @@ with
...
+
+    contact_1 as (
+        select student_id, contact_name, email, phone_home, phone_mobile,
+        from {{ ref("int_focus__student_contacts") }}
+        where sort_order = 1
```
anthonygwalters, line 43: Correct here, wrong in the docs: `sort_order` is `NUMERIC`, not STRING.

The spec, the plan, and #4782 all flag it as a 'naming trap' that is a STRING, and the plan's prescribed `where sort_order = '1'` would have **failed the build** — BigQuery has no matching signature for `NUMERIC = STRING`. The shipped code diverged from the plan correctly, but silently. Worth correcting the spec so the record is accurate.

Adjacent: nothing upstream guarantees one row per `(student_id, sort_order)` — `stg_focus__students_join_people` is unique on `id` only, and `int_focus__student_contacts` is explicitly 'unslotted and uncapped by design'. Zero collisions in current data, and the grain test would catch a fan-out for non-null `fleid`, but that leaves 22% of current-year rows uncovered (see the `fleid` comment below).

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
```diff
@@ -30,30 +26,83 @@ with
...
+    /* TODO(#4795): Focus has no advisory structure for grades 7-8 */
+    cast(null as string) as advisor_lastfirst,
+
+    cast(s.powerschool_id as int64) as ps_id,
+
+    lpad(s.mdcps_id_raw, 7, '0') as mdcps_id,
+
+    regexp_extract(s.sex_label, r'\[(\w)\]') as gender,
+
+    if(s.ese_fefp_code is not null, 'Has IEP', 'No IEP') as iep_status,
```
anthonygwalters: The value-domain break flagged for KIPP Forward is avoidable. Prod's `iep_status` domain is already exactly `{'SPED', 'No IEP'}` — emitting `'SPED'` instead of `'Has IEP'` keeps it 100% stable and nothing downstream needs updating.

Separately, for AY2025 rows this swaps PowerSchool's authoritative flag for a 79%-recall proxy on a year PowerSchool still covers correctly. Verified the recall exactly: 53 students flagged `SPED` in PowerSchool for SY2025 grades 7-8, 42 caught by Focus ESE FEFP Code, 11 missed, 1 Focus-only. So 11 AY2025 students flip from correct to `No IEP`.

Worth considering whether the prior-year half of the sheet should keep the PowerSchool value while it exists.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
```diff
@@ -30,30 +26,83 @@ with
...
+
+    case
+        when e.startdate > current_date('{{ var("local_timezone") }}')
+        then -1
+        when e.enroll_status = 3
+        then 3
+        when oe.student_number is not null
+        then 0
+        else 2
+    end as enroll_status,
```
anthonygwalters, line 124: **Blocking, or at minimum needs disclosing.** The `startdate` branch fires for every AY2026 span because Miami's first day is roughly Aug 17. Reconstructed this CASE against prod:

| Year | -1 | 0 | 2 | 3 |
| 2025 | 0 | 219 | 144 | 2 |
| 2026 | **374** | **0** | 1 | 0 |

AY2025 matches the PR body exactly. AY2026 appears nowhere in the PR body or the plan — the plan called this 'the single most important check in the plan' and ran it on AY2025 only, the year PowerSchool already covers.

It is PowerSchool-faithful and it self-heals once school starts, but any consumer filtering `enroll_status = 0` gets **zero** current-year students, so the sheet is still functionally empty for its main purpose. `src/dbt/kipptaf/CLAUDE.md` is explicit: '-1 is pre-registered (not yet enrolled) — never report against either.' Either decide that a pre-first-day roster should read 0, or tell KIPP Forward what to expect for the first two weeks of every year.

Also worth an inline note: `0` and `2` are student-level here (deliberate, and it correctly mirrors the PowerSchool behavior kipptaf's CLAUDE.md documents), while `-1` and `3` are row-level. A future maintainer could plausibly 'fix' the `open_enrollment` join by scoping it to `academic_year` and reintroduce the bug this PR removes.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
```diff
@@ -62,33 +111,37 @@ select
...
-    co.cumulative_y1_gpa_unweighted as gpa_cumulative,
-    co.ada_unweighted_year_prev as previous_year_ada,
-    co.state_studentnumber as fleid,
+    /* TODO(#4796): Miami middle school no longer produces a GPA; awaiting a
+       replacement academic-standing metric from KIPP Forward */
+    cast(null as float64) as gpa_cumulative,
+
+    ada.ada_unweighted_year_prev as previous_year_ada,
+
+    e.fteid as fleid,
```
anthonygwalters, line 139: `fteid` is null for **81 of 375 AY2026 rows (22%)**. AY2025 has 5, matching prod.

Not a re-sourcing defect — none of the 81 has a prior-year Focus row with a non-null `fteid`, so it is a Focus data-entry lag and not recoverable by coalescing across years. But it is undisclosed, and `fleid` is now load-bearing three ways:

- the **only** key in the `unique_combination_of_columns` grain test
- the FAST join key (line 133)
- the prior-year FAST join key (line 136)

Net effect: the grain test covers 78% of current-year rows, and those 81 students get no FAST scores in either the current or prior-year columns.

Please add the figure to the PR body alongside the `ps_id` coverage number, and consider a second grain key that is non-null (`mdcps_id` is populated where `fleid` is not, worth checking).

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
```diff
@@ -62,33 +111,37 @@ select
...
-    {{ ref("int_powerschool__gpa_term") }} as gpa
-    on co.studentid = gpa.studentid
-    and co.yearid = gpa.yearid
-    and co.schoolid = gpa.schoolid
-    and co._dbt_source_project = gpa._dbt_source_project
+    {{ ref("int_extracts__student_enrollments") }} as ada
+    on s.powerschool_id = ada.student_number
+    and e.academic_year = ada.academic_year
+    and ada.region = 'Miami'
+    and ada.rn_year = 1
```
anthonygwalters: **Blocking.** This reads `ada_unweighted_year_prev` off the *same* academic year's PowerSchool row, and PowerSchool has no Miami AY2026 row — so `previous_year_ada` is null for all 375 current-year rows.

| Year | rows | `previous_year_ada` populated |
| --- | --- | --- |
| 2025 | 365 | 269 |
| 2026 | 375 | **0** |

The spec's Known limitations says the opposite: '`previous_year_ada` empties at the SY2027 rollover, not SY2026. For SY2026 rows the prior year is SY2025, which PowerSchool holds.' The reasoning is right, but the join points at the wrong year, so the value is never reached. Only AY2025 was verified, which is why this passed.

Verified fix, byte-identical on AY2025: `ada_unweighted_year_prev` on year N equals `unweighted_ada` on year N-1 for 269/269 rows. So re-key to the prior year and read that year's own value:

```sql
left join
    {{ ref("int_extracts__student_enrollments") }} as ada
    on s.powerschool_id = ada.student_number
    and e.academic_year - 1 = ada.academic_year
    and ada.region = 'Miami'
    and ada.rn_year = 1
```

with `ada.unweighted_ada as previous_year_ada` at line 118. That leaves AY2025 unchanged and recovers roughly 295 AY2026 rows — the ones carrying a `powerschool_id`.

anthonygwalters (commented):
Six inline notes. The full write-up — including the spec and plan findings, the triage of the automated review above, and what holds up — is in the separate top-level comment on this PR.

Headline: the re-sourcing design is sound and the AY2025 parity work is real, but every parity check in the plan ran against AY2025, the year PowerSchool already covers. On AY2026 three columns are empty or wrong, and the spec documents two of them as working. `previous_year_ada` (null for all 375 rows) and `enroll_status` (-1 for 374 of 375) are the two I would fix before merge.

Every count and type above was re-derived against prod BigQuery rather than taken from the PR body.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
anthonygwalters: Correction to my own parenthetical: I suggested `mdcps_id` as a candidate second grain key. It is not — I checked, and it is null for exactly the same 81 students.

That makes the finding worse rather than better. For those 81 AY2026 rows, `fleid`, `mdcps_id`, and `ps_id` are **all** null, so no column in the 29-column contract identifies them except `lastfirst`:

| AY2026 grade 7-8 rows | 375 |
| --- | --- |
| null `fleid` | 81 |
| null `mdcps_id` | 81 |
| null `fleid` **and** null `mdcps_id` | 81 |
| null `ps_id` | 80 |

So 22% of the current-year roster ships with no usable identifier for KIPP Forward to join or dedupe on, and no grain test can cover those rows without adding a column to the contract. Focus's own `student_id` (`int_focus__student_enrollments.student_number`) is populated for all of them and is the natural candidate, but adding it is a contract change — which is presumably why it was not considered.

Not necessarily a blocker for this PR, but it should be a known limitation in the spec and something KIPP Forward hears about, since the practical question 'who are these 81 students' has no answer in the sheet today.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
CGibson17: Both points confirmed and fixed in `f5afba7b6`.

**The premise was wrong, as you said.** `current_academic_year` reads 2026 — rolled 2026-07-17 in `7cfd878bf`. I had read it out of a stale checkout that sits on an older branch, which is also where the "30 columns" figure and the `sort_order` STRING claim came from. The `latest_year` CTE is deleted; `open_enrollment` now anchors on `var('current_academic_year')`, so both sites trust the same value and the internal inconsistency is gone.

**The fragility point I confirmed independently** before acting on it: AY2018 through AY2024 have zero null `exitcode`s, so the max-year anchor only ever held while the max year happened to be open.

Also added the durable guards you asked for in item 9 — `accepted_values` on `enroll_status` (`-1, 0, 2, 3`) and on `iep_status`, plus a singular test asserting the current academic year is non-empty.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
CGibson17: Corrected in the spec and the plan in `f5afba7b6` — `sort_order` is NUMERIC, filters are unquoted. You are right that the plan's `= '1'` would have failed the build, and that the divergence was silent.

The latent fan-out risk stands and is not fixed. Current data has zero collisions on `(student_id, sort_order)`, and the grain test covers only non-null `fleid` — 78% of current-year rows. It is recorded in the spec's known limitations rather than fixed, because closing it properly means adding an identifier column to the contract. See my reply on your `fleid` follow-up.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
CGibson17: Split decision after checking with the author.

**Domain:** keeping `Has IEP` / `No IEP` rather than reverting to `SPED`. The author weighed the stable-domain argument and chose the clearer value, accepting the one-time consumer break. It is disclosed in the PR body.

**Authority:** your second point is addressed. AY2025 now reads `iep_status` from PowerSchool via the hybrid added in `f5afba7b6`, so the 11 students Focus misses keep their authoritative flag for the year PowerSchool still covers. Focus supplies AY2026 onward only. Verified: `iep_status` distinct values are exactly `Has IEP` and `No IEP` across both years.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
CGibson17: Confirmed exactly — AY2026 is 374 at `-1`, 1 at `2`, and **zero** students at `0`.

The author's decision is to keep the PowerSchool-faithful `-1` and disclose it, rather than force `0` before the first day. It is now in the spec's known limitations, the column description, and the PR body, so KIPP Forward knows the current-year roster reads pre-registered until roughly Aug 17 and self-heals.

Also added the grain note you asked for: `0` and `2` are student-level, mirroring PowerSchool, while `-1` and `3` are row-level — with an inline warning against "fixing" the `open_enrollment` join by scoping it to `academic_year`, which would reintroduce the bug this PR removes.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
CGibson17: Confirmed: 81 of 375 AY2026 rows, against 5 of 365 in AY2025. Added to the spec's known limitations and to the PR body next to the `ps_id` figure.

Not fixed. A second grain key needs a non-null identifier, and per your own follow-up `mdcps_id` is null for exactly the same 81 — so it would mean adding Focus's `student_id` to the contract. That is a contract change and a KIPP Forward conversation, not something to slip into this PR.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
CGibson17: Fixed in `f5afba7b6`, using your suggested join.

I verified the equivalence claim independently before applying it: `ada_unweighted_year_prev` on year N equals `unweighted_ada` on year N-1 for **1,126 of 1,126** rows.

Result, per year:

| Year | before | after |
| --- | --- | --- |
| 2025 | 269 | 269 — byte-identical, as you predicted |
| 2026 | **0** | **295** |

Bounded by `powerschool_id` coverage, exactly as you called it.

This one was entirely my miss: my own plan named the AY2025 check "the single most important check in the plan" and then ran it on the year PowerSchool already covers.

src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
CGibson17: Confirmed — the same 81 students for `fleid` and `mdcps_id`, 80 for `ps_id`. Those rows carry no usable identifier except `lastfirst`.

Recorded in the spec's known limitations, including your point that Focus's own `student_id` is populated for all of them and is the natural second grain key but would be a contract change. Flagged for KIPP Forward rather than resolved here.

Issue #4808, opened 2026-08-10
fix(dbt): ddi dashboard iready metrics are keyed to the assessment subject_area and undercount


anthonygwalters:
## Accepted limitation, and the part that is not

It is understood and accepted by stakeholders that `rpt_tableau__ddi_dashboard` emits a row only for weeks in which an assessment was given, so iReady lesson data does not appear for other weeks. Granular week-to-week lesson data lives on a separate dashboard. That is not what this issue is about.

The additional, separate loss: even within weeks that **did** have an assessment, roughly half the iReady data still cannot be attributed, because the row's `subject_area` comes from the assessment rather than from iReady.

## Mechanism

Every iReady measure on the DDI Suite resolves through an LOD of this shape:

```text
{FIXED [student_number], [week_start_monday], [subject_area] :
  MAX(CASE [subject_area]
      WHEN 'Text Study' THEN [is_passed_iready_2plus_reading_int]
      WHEN 'Mathematics' THEN [is_passed_iready_2plus_math_int] END)}
```

`subject_area` on a dashboard row is the subject of the assessment that produced the row. The iReady columns, by contrast, are joined per student-week irrespective of subject. So a week's iReady **reading** figure is only attributable if that student happened to sit a `Text Study` assessment in the same week, and the **math** figure only if they sat a `Mathematics` one. A student-week with an assessment in any other subject carries the iReady values on the row but resolves the CASE to NULL.

## Size of the gap

AY2025, grades K-8, at 2 or more lessons passed:

| | Reading | Math |
| --- | --- | --- |
| True iReady student-weeks | 154,755 | 111,654 |
| Present on the dashboard at all | 79,582 (51%) | 57,735 (52%) |
| Attributable by the LOD | 38,715 (25%) | 34,850 (31%)  |

The middle row is the accepted limitation. The drop from the middle row to the bottom row is this issue.

## Why the obvious fix does not work

Moving `r.module_type != 'WPP'` into the `ON` clause to restore the LEFT JOIN does not resolve this. The recovered rows carry a null `subject_area`, the Tableau CASE returns NULL for them, and the LOD still cannot attribute the values. Any real fix has to stop keying iReady off the assessment subject — either emit iReady on its own subject key in the model, or rework the workbook measures — so it is a joint model and workbook change, not a predicate move.

## Affected worksheets

[...]

PR #4809, opened 2026-08-10
fix(dbt): gate ddi dashboard assessments on enrollment dates

### LINKED ISSUE #4807: fix(dbt): gate ddi dashboard assessments on enrollment dates instead of enroll_status
## Problem

`rpt_tableau__ddi_dashboard` decides whether an assessment appears by filtering `enroll_status = 0`. That column is student-level and current-only — it reflects a student's status today, not their status on the date an assessment was administered. Any student who has since withdrawn, transferred, or graduated loses their entire assessment history retroactively.

The loss is concentrated on transition grades. Newark AY2025, share of each grade's enrolled students retained by the current filter:

| Grade | Enrolled during SY25-26 | Kept by filter | Retained |
| --- | --- | --- | --- |
| K-7 | 515-669 each | - | 85-91% |
| 8 | 618 | 348 | 56% |
| 9-11 | 321-419 each | - | 86-91% |
| 12 | 310 | 14 | 4.5% |

95.5% of the Newark senior class is absent from the dashboard for the year they were seniors.

This is the documented `enroll_status` trap in `src/dbt/kipptaf/CLAUDE.md` — a current-only status cannot answer a question about the past.

## Change

Drop `co.enroll_status = 0` from the `identifiers` CTE. Eligibility then rests on `is_enrolled_week_end`, which is already a point-in-time enrollment test. An inline SQL comment records the rule at the derivation site, and the properties yml gains a model-level `description:` stating it.

Retained deliberately:

- `not co.is_out_of_district` — out-of-district placements are legitimately off a DDI roster
- `is_enrolled_week_end` as the enrollment gate
- The academic year window and the WPP predicate, untouched

No contract change. `enroll_status` stays projected; it simply stops being constant.

## Validation

CRQ2 / Mathematics / Newark / AY2025, measured against `rpt_tableau__assessment_dashboard`:

- Students: 4,357 current, 5,091 after the change, 5,110 on the assessment dashboard — 734 of the 753-student gap closed (97.5%)
- Mastery rate: 58.43% to 58.45%, against 58.39% on the assessment dashboard

The change is a completeness fix, not a metric shift.

The 19 residual students break down as 6 flagged out-of-district (deliberate divergence), 8 who withdrew mid-week but were enrolled on the administered-at date, and 5 who were enrolled neither at week end nor on the administered-at date. The last group is correctly excluded here, and appears on the assessment dashboard only because that model applies no date gate at all.

## Out of scope

- Gating on the administered-at date rather than week end. Worth 8 students in the validated slice; deferred.
- Backfilling `description:` across the yml's 72 columns, none of which currently carry one, and adding the missing model uniqueness test. Both are pre-existing convention gaps. A 72-column docs change would bury a one-predicate fix, and the uniqueness test needs its own look at the existing fan-out.

## Notes

Retroactive to AY2025, cleared with the stakeholder. SY25-26 figures on the DDI Suite will move, most visibly for grades 8 and 12.

The DDI Suite workbook references `enroll_status` in no worksheet, so the change surfaces on the dashboard rather than being filtered out downstream.

Primary antecedent: https://teamschools.zendesk.com/agent/tickets/434923

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will make an assessment appear on the DDI Suite dashboard based on whether the student was enrolled when it was administered, rather than on whether they are enrolled today.

`rpt_tableau__ddi_dashboard` pinned `co.enroll_status = 0` in its `identifiers` CTE. That column is student-level and current-only — it reflects a student's status now, not their status on the administered-at date — so every student who has since withdrawn, transferred, or graduated lost their entire assessment history retroactively. The loss lands on transition grades:

| Grade (Newark, AY2025) | Enrolled during SY25-26 | Kept by the old filter | Retained |
| --- | --- | --- | --- |
| K-7 | 515-669 each | - | 85-91% |
| 8 | 618 | 348 | 56% |
| 9-11 | 321-419 each | - | 86-91% |
| 12 | 310 | 14 | 4.5% |

95.5% of the Newark senior class was absent from the dashboard for the year they were seniors.

Eligibility now rests on `is_enrolled_week_end`, which is already a point-in-time enrollment test, bounded to valid statuses `(0, 2, 3)` so inactive (1) and pre-registered (-1) students stay out per the `enroll_status` rule in `src/dbt/kipptaf/CLAUDE.md`. `not co.is_out_of_district` is retained deliberately. No contract change — `enroll_status` is still projected in every UNION branch and no column type moved.

**Validation.** CRQ2 / Mathematics / Newark / AY2025, measured against `rpt_tableau__assessment_dashboard`:

- Students 4,357 to 5,091, against 5,110 on the assessment dashboard — 734 of the 753-student gap closed (97.5%)
- Mastery rate 58.43% to 58.45%, against 58.39% on the assessment dashboard

So this is a completeness fix, not a metric shift. Of the 19 residual students, 6 are deliberately excluded as out-of-district, 8 withdrew mid-week after sitting the assessment (deferred), and 5 were enrolled neither at week end nor on the administered-at date — correctly excluded here, and present on the assessment dashboard only because that model applies no date gate at all.

**Retroactive to AY2025, cleared with the stakeholder.** SY25-26 figures on the DDI Suite will move, most visibly for grades 8 and 12. Rollback is a one-line revert of the predicate.

The DDI Suite workbook references `enroll_status` in no worksheet, so the change surfaces on the dashboard rather than being filtered out downstream.

**Deliberately out of scope**, each examined and consciously left alone:

- `and r.module_type != 'WPP'` sits in the WHERE against a left-joined table, so the LEFT JOIN behaves as an INNER JOIN and zero-assessment student-weeks drop out. Known and stakeholder-accepted.
- The iReady measures are keyed to the assessment row's `subject_area` and undercount as a result. Filed separately as #4808.
- The properties yml's 72 columns carry no `description:`, and the model has no uniqueness test. Pre-existing convention gaps; a 72-column docs backfill would bury a one-predicate fix, and the uniqueness test needs its own look at the existing fan-out.
- Gating on the administered-at date instead of week end, worth 8 students in the validated slice.

Primary antecedent: https://teamschools.zendesk.com/agent/tickets/434923

Closes #4807

## AI Assistance

AI-assisted throughout, human-directed on every decision. Claude Code did the investigation (reading both models, quantifying the gap against prod, unpacking the packaged Tableau workbook to confirm no worksheet filters `enroll_status`), wrote the change, and validated it. The scope calls — Option B over the alternatives, keeping `not is_out_of_district`, deferring the administered-at gate, retroactive application, and keeping the PR minimal rather than backfilling docs — were all made by the requester. The `enroll_status in (0, 2, 3)` bound was raised by Claude after querying the value domain and accepted.

### General

### dbt

- [x] Properties file present — `properties/rpt_tableau__ddi_dashboard.yml` exists and gains a model-level `description:` stating the new eligibility rule.
- [x] Exposure present — the `ddi_suite` exposure in `models/exposures/tableau.yml` already covers this model. No change needed.
- [x] Not a breaking change — no column renamed or removed, no type changed. `enroll_status` keeps its position and type; it simply stops being constant `0`.
- [x] No external source added or modified, so no `stage_external_sources` run is required.

## CI checks

- [ ] **Trunk** — passes. Locally, `trunk check --force` on both changed files returned no issues after the commit-time `fmt`.
- [ ] **dbt Cloud** — passes. This is a kipptaf-only change, so `state:modified+` genuinely selects and builds the model.
- [ ] **Dagster Cloud** — not triggered; `src/dbt/**` is excluded from the deploy workflow `pull_request` paths.

## Local verification

- `uv run dbt compile --select rpt_tableau__ddi_dashboard --target prod` succeeds; the compiled artifact contains zero occurrences of `enroll_status = 0` and one of the new bound, with `co.enroll_status` still projected in all three UNION branches.
- No `dbt build` was run. The model is a view, and a view build does not evaluate data, so validation is the prod-query comparison above.

### DIFF
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__ddi_dashboard.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__ddi_dashboard.yml
@@ -6,9 +6,10 @@ models:
       assessment response appears for a student when that student was enrolled
       as of the end of the week the assessment was administered, determined from
       enrollment dates rather than from whether the student is enrolled today,
-      so students who have since withdrawn or graduated keep their history.
-      Students with an inactive or pre-registered enrollment status, and
-      students in an out-of-district placement, are excluded.
+      so students who have since withdrawn or graduated keep their history for
+      as long as the extract's two-year window still covers it. Rows are
+      restricted to students holding a valid enrollment status, and students in
+      an out-of-district placement are excluded.
     columns:
       - name: student_number
         data_type: int64
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__ddi_dashboard.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__ddi_dashboard.yml
@@ -1,5 +1,14 @@
 models:
   - name: rpt_tableau__ddi_dashboard
+    description: >-
+      Weekly extract behind the DDI Suite dashboard, unioning ES/MS assessment
+      responses, HS assessment responses, and staff walkthrough observations. An
+      assessment response appears for a student when that student was enrolled
+      as of the end of the week the assessment was administered, determined from
+      enrollment dates rather than from whether the student is enrolled today,
+      so students who have since withdrawn or graduated keep their history.
+      Students with an inactive or pre-registered enrollment status, and
+      students in an out-of-district placement, are excluded.
     columns:
       - name: student_number
         data_type: int64
--- a/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__ddi_dashboard.sql
+++ b/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__ddi_dashboard.sql
@@ -153,8 +153,14 @@ with
             and cc.courses_credittype = sf.powerschool_credittype
             and sf.rn_year = 1
         where
-            co.enroll_status = 0
-            and co.is_enrolled_week_end
+            /* Eligibility is enrollment as of the week end, not current
+               enroll_status: that column is student-level and current-only, so
+               pinning it to 0 retroactively erased the assessment history of every
+               student who had since withdrawn, transferred, or graduated (#4807).
+               Withdrawn (2) and graduated (3) are in scope; inactive (1) and
+               pre-registered (-1) stay out -- never report against either. */
+            co.is_enrolled_week_end
+            and co.enroll_status in (0, 2, 3)
             and not co.is_out_of_district
             and co.academic_year >= {{ var("current_academic_year") - 1 }}
             {# TODO: Remove SY26 #}


anthonygwalters:
## Review responses

Verdicts on both automated findings, plus one material finding from a local adversarial review dispatched alongside this one. Every claim below was re-verified against the code or against prod before writing.

### Finding 1 — duplicated `enroll_status in (0, 2, 3)`, no shared macro. Declined for this PR, agreed as a follow-up.

Verified the call-site count. Five pre-existing sites plus this one:

```text
students/intermediate/int_students__dibels_participation_roster.sql:25
extracts/tableau/rpt_tableau__student_course_grades.sql:350
extracts/tableau/rpt_tableau__dibels_dashboard.sql:173
extracts/tableau/rpt_tableau__dibels_dashboard.sql:369
extracts/tableau/rpt_tableau__gpa_cumulative_year.sql:123
```

The precedent is closer than the finding suggests, which is worth recording. `rpt_tableau__gpa_cumulative_year.sql:118-124` and `rpt_tableau__student_course_grades.sql:346-354` both carry the same four-part construct this PR now uses — `not is_out_of_district`, `enroll_status in (0, 2, 3)`, a date-based enrollment flag, and the `academic_year >= current_academic_year - 1` window — with a comment explaining the status guard drops pre-registered (-1) and invalid (1) rows. So this is the established idiom rather than a new one, which is also why the guard is motivated rather than merely defensive.

Not extracting a macro here, per this PR's scope. Noted for a follow-up.

### Finding 2 — model description does not fully enumerate excluded statuses. Fixed in 92b10a695, by removing the enumeration rather than completing it.

The finding is correct that `(0, 2, 3)` also excludes 4 (Imported as Historical) and -2, neither mentioned. A local review independently flagged the same sentence from the opposite direction — that naming pre-registered as excluded describes a filter effect that never fires, since the observed `enroll_status` domain in this model is `{0, 1, 2, 3}` with no -1 and no -100 phantoms.

Completing the enumeration would satisfy one reading and worsen the other, so the description now states the rule positively — restricted to a valid enrollment status — and drops the value list entirely. The inline SQL comment keeps the -1 and 1 wording, matching the sibling models cited above.

The same commit adds one clause the local review caught, unrelated to statuses. The old text said withdrawn and graduated students "keep their history" full stop, but that history is bounded by the two-year `academic_year` window at line 165. The description now says so.

### Material finding from the local review, not a blocker, not fixed here

`is_enrolled_week_end` can read false for a student who **was** enrolled at week end. `int_extracts__student_enrollments_weeks.sql:31-37` reduces a student-week to one row with `dbt_utils.deduplicate(partition_by="_dbt_source_project, student_number, week_start_monday", order_by="is_enrolled_week desc")`. The tiebreak ranks on `is_enrolled_week`, the week-**start** flag, while every consumer gate here reads `is_enrolled_week_end`. The surviving row's week-end flag is therefore collateral.

Re-measured against prod for AY2025 onward under this PR's exact predicate:

| Case | Partitions | Involving withdrawn or graduated students |
| --- | --- | --- |
| A week-end-true row exists, but every week-start-true row is week-end-false, so the dedup deterministically keeps a false row | 17 | 8 |
| No row is week-start-true, so all candidates tie and `array_agg(... limit 1)` picks arbitrarily | 42 | 14 |

Failure scenario — a student with two same-district stints inside one week, exiting School A on Tuesday and enrolling at School B on Thursday. The School A row is week-start-true and week-end-false and wins the dedup; the School B row covering Sunday is discarded. An assessment administered Friday at School B never appears.

Scale is 59 partitions out of 825,542 student-weeks passing the predicate, about 0.007%. The weakness is pre-existing and lives in a shared intermediate with other consumers, so fixing it here would be both out of scope and wider-reaching than this PR. This PR newly exposes 22 of the 59 by admitting statuses 2 and 3.

Worth noting alongside this: the automated review traced the same `deduplicate` call and cleared it, correctly, on a different question — whether admitting `enroll_status = 3` reintroduces the null-date graduate placeholders. It does not. The flag mismatch above is a separate property of that same call.

### One related caveat for a human reviewer

The old `enroll_status = 0` pin was an incidental cross-district guard, since a student is active in at most one district. The new predicate leans instead on cross-district enrollment date ranges not overlapping, which is a property of the data rather than an enforced invariant, and the upstream uniqueness test partitions by `_dbt_source_project` so it would not catch a violation. Currently zero students appear in more than one code location, and there are zero duplicate `(student_number, week_start_monday)` keys across the 825,542 rows the new predicate admits.

### Not re-verified

The automated review notes it has no warehouse access and took the PR description's numbers as reported. Those numbers were produced by BigQuery queries against prod and re-checked after the `enroll_status in (0, 2, 3)` bound was added, which left them unchanged at 5,091 students and 30,036 rows on the validated slice.

Issue #4811, opened 2026-08-10
fix(extracts): Miami rows in the student contact info feed are frozen at AY2025


anthonygwalters:
## Problem

`rpt_gsheets__student_contact_info` carries 1,114 Miami students (619 ES, 495 MS) at `academic_year = 2025` with `enroll_status = 0`. Every other region reports `academic_year = 2026`. Miami's rows will never advance.

## Cause

The feed reads `int_extracts__student_enrollments`, which reads `base_powerschool__student_enrollments`, which unions only the four PowerSchool sources. `kippmiami_powerschool` is a frozen pre-Focus archive — final ODBC pull 2026-07-01, never rebuilt (see the source description in `src/dbt/kipptaf/models/powerschool/sources-kippmiami.yml`). Miami's live enrollment moved to Focus, and nothing in this lineage reads `int_focus__student_enrollments`.

Net effect: every column on Miami's rows in this feed is a year stale, not just enrollment. Anyone consuming the Miami tabs of the linked sheets is reading SY25-26 data as if it were current.

## Scope of the blast radius

`int_extracts__student_enrollments` is consumed by roughly 150 models. Any Miami row sourced through it has the same staleness, so this is not limited to the contact feed — that model is just where it was noticed.

## Affected sheets

The four exposures depending on `rpt_gsheets__student_contact_info` in `models/exposures/google-sheets.yml`, including `gsheets_student_contact_info` and `gsheets_student_logins`.

## Not yet decided

Whether the fix is to union the Focus enrollment vertical into the PowerSchool base model, add a Focus branch at the `int_extracts` layer, or drop Miami from PowerSchool-sourced extracts entirely and build a parallel Focus path. Needs design.

## Context

Found while adding `es_attended` / `ms_attended` to the contact feed. Miami was intentionally shipped there with no special-casing, since those two columns are no more stale than the rest of the row. Refs #4810.

cbini:
Folding into #4729 — same root cause, same fix, and keeping them apart means designing the Focus-branch insertion twice.

#4729 covers the frozen `kippmiami_powerschool` archive still being the Miami branch of the kipptaf PowerSchool vertical. This issue is that same staleness observed at `int_extracts__student_enrollments`, which reads `base_powerschool__student_enrollments` five times. That model is inside #4729's 49.

[...]

PR #4812, opened 2026-08-10
feat(extracts): add prior-level es_attended and ms_attended to the contact feed

### LINKED ISSUE #4810: feat(extracts): add es_attended and ms_attended to the student contact info feed
## Ask

Add two columns to `rpt_gsheets__student_contact_info` (all four regions) showing the most recent KIPP elementary school and most recent KIPP middle school a student previously attended.

Requested behavior by the student's current school level:

| Current level | `es_attended` | `ms_attended` |
| ------------- | ------------- | ------------- |
| ES            | null          | null          |
| MS            | most recent prior KIPP ES | null |
| HS            | most recent prior KIPP ES | most recent prior KIPP MS |

## Approach

`int_extracts__student_enrollments` already computes `es_attended` / `ms_attended` (the `esms_attend` CTE, most recent enrollment per school level ordered by `exitdate desc`). Those columns include the student's *current* school, so a current MS student's `ms_attended` is their own school.

Rather than change the hub — it feeds roughly 150 models and any edit fans `state:modified+` across the whole graph — the prior-level rule is applied in the extract model only:

```sql
if(school_level = 'ES', null, es_attended) as es_attended,
if(school_level in ('ES', 'MS'), null, ms_attended) as ms_attended,
```

Nulling every level at or below the student's current level (rather than only the matching one) also suppresses a stray MS enrollment record carried by one Camden ES student.

Columns are appended to the END of the select list, per standing practice for sheet extracts — the consuming Google Sheets pull these views through BigQuery Connected Sheets, so appending keeps existing column letters and formulas intact.

## Notes

- The new columns intentionally reuse the `es_attended` / `ms_attended` names for consistency with the Tableau extracts, but carry different semantics than the identically-named upstream columns. The divergence is documented in each column `description:`.
- Sumner grades 5 and 6 (106 Camden students, coded `school_level = 'ES'` with `school_level_alt = 'MS'`) show blank for both. Only 4 of them ever attended a different KIPP elementary school, so no carve-out is warranted.
- Miami is included with no special-casing. Its rows are computed off the frozen pre-Focus PowerSchool archive, consistent with every other column in those rows. Tracked separately.

## Expected result against current production data

| Current level | Students | `es_attended` populated | `ms_attended` populated |
| ------------- | -------- | ----------------------- | ----------------------- |
| ES | 4,975 | 0 | 0 |
| MS | 4,021 | 2,479 | 0 |
| HS | 1,986 | 777 | 1,386 |

anthonygwalters:
# Pull Request

## Summary and Motivation

When merged, this pull request will add two columns to `rpt_gsheets__student_contact_info`, for all four regions:

- `es_attended` — the most recent KIPP elementary school the student attended **before** their current school
- `ms_attended` — the same for middle school

Behavior by the student's current school level:

| Current level | `es_attended`             | `ms_attended`             |
| ------------- | ------------------------- | ------------------------- |
| ES            | null                      | null                      |
| MS            | most recent prior KIPP ES | null                      |
| HS            | most recent prior KIPP ES | most recent prior KIPP MS |

### Why the rule lives here and not upstream

`int_extracts__student_enrollments` already computes `es_attended` / `ms_attended` (the `esms_attend` CTE, most recent enrollment per school level ordered by `exitdate desc`). Those picks include the student's _current_ school, so a current MS student's `ms_attended` is their own school.

That model feeds roughly 150 models. Redefining its columns would fan `state:modified+` across the whole graph and silently change behavior for the Tableau extracts already reading them. So the prior-level rule is applied in the extract instead:

```sql
if(school_level = 'ES', null, es_attended) as es_attended,
if(school_level in ('ES', 'MS'), null, ms_attended) as ms_attended,
```

Blanking every level at or below the current level, rather than only the matching one, also suppresses a stray MS enrollment record carried by one Camden ES student.

Columns are appended at the **end** of the select list. The consuming sheets read this view through BigQuery Connected Sheets, so appending keeps existing column letters and formulas intact.

### Naming

The columns deliberately reuse the upstream names for consistency with the Tableau extracts that already carry them, but they hold narrower values. Each column `description:` records the difference so it is discoverable from dbt docs rather than only from this PR.

### Miami

Included with no special-casing. Miami's rows compute off the frozen pre-Focus PowerSchool archive, consistent with every other column in those rows. The underlying problem — Miami rows in this feed are frozen at AY2025 while every other region reads AY2026 — predates this change and is tracked separately in #4811.

### Verification

A view build does not evaluate data, so the built model was queried rather than trusted on a green build:

| Current level | Rows  | `es_attended` populated | `ms_attended` populated | Mismatches vs spec |
| ------------- | ----- | ----------------------- | ----------------------- | ------------------ |
| ES            | 4,975 | 0                       | 0                       | 0                  |
| MS            | 4,021 | 2,479                   | 0                       | 0                  |
| HS            | 1,986 | 777                     | 1,386                   | 0                  |

Feed grain unchanged at 10,982 rows, 1:1 against the upstream population, no fan-out. Confirmed the dev view resolved to prod `kipptaf_extracts` rather than a stale personal copy.

**Rollback plan:** not a breaking change. Two additive columns, nothing renamed or removed, so a revert restores the prior column set with no downstream coordination.

## AI Assistance

Claude Code authored this change.

**Human-directed:** the requirement itself, the null-by-current-level rule, the decision to reuse the upstream column names rather than more explicit ones, the call to ship Miami without a carve-out, and the choice to skip adding a uniqueness test.

**AI-assisted:** discovering that the two columns already existed upstream with different semantics, quantifying the Sumner and cross-region edge cases against production data, the implementation, and the verification above.

### General

### dbt

- [x] Properties file updated with both new columns, `data_type`, and descriptions
- [x] Exposures — no change needed. The four existing exposures on this model still describe every consumer; no new consumer is introduced.
- [x] **Breaking change?** No. Two additive columns appended at the end of a contracted model; nothing renamed or removed.
- [x] External sources — none added or modified, so no `stage_external_sources` run is required.

### Dagster

Skipped — no Dagster changes.

### Docs

## CI checks

- [ ] **Trunk** — `trunk check --force` was clean locally on both changed files
- [ ] **dbt Cloud**
- [ ] **Dagster Cloud** — expected not to trigger; `src/dbt/**` is excluded from the branch-deployment paths

Closes #4810

### DIFF
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__student_contact_info.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__student_contact_info.yml
@@ -100,16 +100,22 @@ models:
       - name: es_attended
         data_type: string
         description: >-
-          Most recent KIPP elementary school the student attended before their
-          current school. Blank while the student is enrolled at an elementary
-          school. Narrower than the identically named column on
-          int_extracts__student_enrollments, which also picks up the student's
-          current school.
+          Most recent KIPP school the student attended while in grades K through
+          4. Blank while the student is currently in grades K through 4, so the
+          column reads as prior history rather than the current placement.
+          Banded by the grade the student was enrolled in rather than by the
+          school's designated level, so a school serving grades outside its
+          designation is attributed to the right band. Distinct from the
+          identically named column on int_extracts__student_enrollments, which
+          bands by school level and includes the student's current school.
       - name: ms_attended
         data_type: string
         description: >-
-          Most recent KIPP middle school the student attended before their
-          current school. Blank while the student is enrolled at an elementary
-          or middle school. Narrower than the identically named column on
-          int_extracts__student_enrollments, which also picks up the student's
-          current school.
+          Most recent KIPP school the student attended while in grades 5 through
+          8. Blank while the student is currently in grade 8 or below, so the
+          column reads as prior history rather than the current placement.
+          Banded by the grade the student was enrolled in rather than by the
+          school's designated level, so a school serving grades outside its
+          designation is attributed to the right band. Distinct from the
+          identically named column on int_extracts__student_enrollments, which
+          bands by school level and includes the student's current school.
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__student_contact_info.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__student_contact_info.yml
@@ -97,3 +97,19 @@ models:
         data_type: string
       - name: home_language
         data_type: string
+      - name: es_attended
+        data_type: string
+        description: >-
+          Most recent KIPP elementary school the student attended before their
+          current school. Blank while the student is enrolled at an elementary
+          school. Narrower than the identically named column on
+          int_extracts__student_enrollments, which also picks up the student's
+          current school.
+      - name: ms_attended
+        data_type: string
+        description: >-
+          Most recent KIPP middle school the student attended before their
+          current school. Blank while the student is enrolled at an elementary
+          or middle school. Narrower than the identically named column on
+          int_extracts__student_enrollments, which also picks up the student's
+          current school.
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__student_contact_info.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__student_contact_info.yml
@@ -1,5 +1,21 @@
 models:
   - name: rpt_gsheets__student_contact_info
+    data_tests:
+      # Pin the prior-level banding so it stays a standing invariant rather than
+      # a point-in-time check. Bands are grades K-4 elementary, 5-8 middle, 9-12
+      # high; a student is blank on every band at or below the one they are
+      # currently in. Both tests are scoped by grade, so a grade_level outside
+      # the K-12 domain is skipped rather than flagged.
+      - dbt_utils.expression_is_true:
+          arguments:
+            expression: es_attended is null
+          config:
+            where: grade_level in ('K', '1', '2', '3', '4')
+      - dbt_utils.expression_is_true:
+          arguments:
+            expression: ms_attended is null
+          config:
+            where: grade_level in ('K', '1', '2', '3', '4', '5', '6', '7', '8')
     columns:
       - name: student_number
         data_type: int64
--- a/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__student_contact_info.sql
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__student_contact_info.sql
@@ -101,5 +101,15 @@ select
     gifted_and_talented,
     salesforce_id as salesforce_contact_id,
     home_language,
+
+    -- Prior-level history only. The upstream columns of the same name pick the
+    -- most recent enrollment at each level INCLUDING the current one, so a
+    -- current MS student's ms_attended is their own school. Blanking every
+    -- level at or below the student's current level makes these read as "KIPP
+    -- schools attended before this one". At-or-below rather than the matching
+    -- level alone also suppresses a stray MS enrollment record carried by one
+    -- Camden ES student.
+    if(school_level = 'ES', null, es_attended) as es_attended,
+    if(school_level in ('ES', 'MS'), null, ms_attended) as ms_attended,
 from {{ ref("int_extracts__student_enrollments") }}
 where enroll_status in (0, -1) and rn_all = 1
--- a/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__student_contact_info.sql
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__student_contact_info.sql
@@ -1,3 +1,69 @@
+with
+    -- Band every enrollment by the GRADE the student was in, not by the school's
+    -- school_level. An increasing number of schools are off-model (e.g. an
+    -- ES-coded school serving grades 5-8), so school_level misfiles their
+    -- enrollments -- it would report such a school as a student's elementary
+    -- school while they attend it for middle school.
+    grade_band_enrollments as (
+        select
+            _dbt_source_project,
+            student_number,
+            school_abbreviation,
+            exitdate,
+
+            case
+                when grade_level between 0 and 4
+                then 'ES'
+                when grade_level between 5 and 8
+                then 'MS'
+                when grade_level between 9 and 12
+                then 'HS'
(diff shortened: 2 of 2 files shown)


anthonygwalters:
## Review response

Each finding verified against the code before responding. Note the SQL changed substantially after this review posted — the two-line `school_level` version it examined is gone, replaced by grade-level banding (commit `31759714c`), so finding 1 is answered by that rework rather than by a targeted patch. Re-firing the review on the current head.

### Finding 1 — `school_level` values outside the ES/MS/HS set fall through to the populate branch — **accepted, resolved**

Correct diagnosis, and the underlying reason turned out to be bigger than the NULL case: `school_level` describes the **building**, not the student. An increasing number of schools are off-model, and the school-keyed pick misfiled them in both directions. Measured against production:

- An ES-coded school serving grades 5-8 (Sumner) was reported as the elementary school of 106 students who were currently attending it for middle school.
- MS-coded schools that served K-4 grades (Hatch 368 enrollment rows, Rise 53, Purpose 31) could never be anyone's elementary school.
- ES-coded schools serving grades 5-8 (Sumner 159 rows, PPES 55) could never be anyone's middle school.

Rather than guard the `school_level` branch, both columns are now derived in this model from `base_powerschool__student_enrollments`, banding every historical enrollment by the grade the student was **in** — K-4 elementary, 5-8 middle, 9-12 high — and taking the most recent school per band. The blanking is keyed on the current `grade_level` for the same reason.

On the `'OD'` sub-point: agreed it was a non-issue, though for a reason not visible without warehouse access — those 142 rows still carry real KIPP school abbreviations (BOLD, Hatch, KHS), so `'OD'` is a placement flag rather than a distinct school. Grade-banding them is correct.

Verified for the transferred concern, since `grade_level` now occupies the same structural position: across all 10,982 feed rows there are zero NULLs and zero values outside 0-12. **Residual, disclosed:** a NULL or out-of-band `grade_level` would still take the populate branch, and the tests added below are grade-scoped, so they would skip such a row rather than flag it. Not guarded; recorded in the properties file comment.

### Finding 2 — no test enforces the blanking invariant — **accepted, implemented**

Added in `9724a7529`, rewritten for grade bands rather than school levels:

```yaml
- dbt_utils.expression_is_true:
    arguments:
      expression: es_attended is null
    config:
      where: grade_level in ('K', '1', '2', '3', '4')
- dbt_utils.expression_is_true:
    arguments:
      expression: ms_attended is null
    config:
      where: grade_level in ('K', '1', '2', '3', '4', '5', '6', '7', '8')
```

Warn severity via the project default, per the suggestion. Both pass, and both scopes are non-vacuous — 4,869 and 8,996 of 10,982 rows respectively, so they are not passing trivially on an empty scope.

### Finding 3 — model is not tagged `contains_pii` — **confirmed accurate, declined for this PR**

Verified: the properties file has no `config` block at all, while 12 sibling YAMLs under `extracts/` carry `contains_pii: true`, including `rpt_parentsquare__students.yml` and `rpt_deanslist__family_contacts.yml`. The model also projects `student_web_password` and `family_web_password` alongside `dob`, `address`, `home_phone`, `guardianemail`, and names — credentials next to direct identifiers.

Declined here as the author's call. It is pre-existing, untouched by this diff, and tagging a model carrying credentials is a governance decision that deserves its own review rather than riding along on a two-column change.

### Verification available to this review that was not available to the bot

The review correctly labeled its data-dependent claims as unverified. For the record, run locally against production:

- Feed grain unchanged at 10,982 rows, 1:1, no fan-out from the two added joins.
- Grades K-4: both columns blank. Grades 5-8: `ms_attended` blank, 2,606 carry an elementary school. Grades 9-12: 818 and 1,391.
- Coverage improved over the school-keyed version, which returned 2,479 / 777 / 1,386.
- Every case where a column equals the student's current school was inspected and is correct rather than a leak — 159 students in grades 5-8 at K-8 schools hold a real K-4 enrollment record there, and 6 in grades 9-12 attend KHS or NLH, whose minimum served grade is 8.

anthonygwalters:
## Review response — second pass

No blocking findings. Verdicts below, including the one item this review could not check.

### New finding — `esms_attend` ranks without region scoping — **verified, empirically empty**

The convention point is correct: `src/dbt/kipptaf/CLAUDE.md` pairs `student_number` with `_dbt_source_project` precisely because student numbers are district-assigned. Ran the check (note the model lives in `kipptaf_powerschool`, not `kipptaf_extracts` as the suggested query had it):

```sql
with collisions as (
  select student_number
  from `teamster-332318.kipptaf_powerschool.base_powerschool__student_enrollments`
  group by student_number
  having count(distinct _dbt_source_project) > 1
)
select (select count(*) from collisions) as colliding_student_numbers
```

Result: **0 colliding student numbers out of 28,059 distinct values**, and **0 affected rows** in the current feed population. So the upstream gap is real as a convention matter but has no live failure mode. Not filing an issue for a zero-impact pre-existing pattern in a model this PR deliberately does not touch; recording the measurement here so the next person who reads that CTE does not have to re-derive it.

Worth noting the described failure mode was right in shape too: the downstream join still requires `_dbt_source_project` to match, so a collision would surface as a silently null `es_attended`/`ms_attended`, not cross-region contamination.

### Subtlety — bare `grade_level` in the same select list that aliases a differently-typed `grade_level` — **analysis confirmed, no change**

Confirmed: BigQuery has no lateral column aliases, so the bare references resolve against `enrollments.grade_level` (int64 from the driving CTE), not the string alias defined earlier in the same list. Behaves as intended.

Acknowledging the related shape you did not raise: the band boundaries now live in two places — the `CASE` in `grade_band_enrollments` and the two `if()` gates in the final select — plus a third string-valued expression of the same boundaries in the test `where` clauses (the tests necessarily run against the model's *output* `grade_level`, which is the string form, so `0` becomes `'K'`). That duplication is a real maintenance edge if the band definitions ever move. Left as-is deliberately rather than churning verified logic at the end of a review cycle; flagged here so it is on the record.

### Efficiency note — extra full-history scan of `base_powerschool__student_enrollments` — **accepted tradeoff, acknowledged**

Correct, and it is the direct cost of the choice documented in the PR body: deriving here rather than fixing `esms_attend` upstream avoids fanning `state:modified+` across roughly 150 consumers. If this view's Connected Sheets refresh ever becomes a complaint, the fix is to move the banding upstream and pay the fanout once, not to optimize this view in place.

### Prior findings

- **Finding 1** — agreed resolved. Also confirming the specific loose end raised here: the `-2` preschool sentinel does fall through the `CASE` to a null band and is filtered by `where grade_band is not null`, and would take the populate branch in the final gates. Measured across all 10,982 feed rows: zero nulls and zero values outside 0-12. Real edge, currently empty, disclosed in the properties comment rather than silently swallowed.
- **Finding 2** — agreed resolved. Adding one measurement the review could not take: both test scopes are non-vacuous, covering 4,869 and 8,996 of 10,982 rows, so they are not passing trivially on an empty scope.
- **Finding 3** — unchanged. Confirmed accurate, deliberately out of scope, author's call.

Issue #4813, opened 2026-08-10
fix(dbt): is_homeless and the new homeless_status column disagree on historical rows


anthonygwalters:
## Context

`rpt_tableau__okrts_behavior` and `rpt_tableau__okrts_referrals` are gaining a decoded `homeless_status` column. It resolves its source by academic year:

- current academic year: `stg_powerschool__studentcorefields.homeless_code`
- prior academic years: `stg_powerschool__s_nj_ren_x.homeless_code`

The re-enrollment extension carries the value that was true for that enrollment stint. The student-level `STUDENTCOREFIELDS` value is overwritten in place, so it reflects only the student's current status.

The pre-existing boolean `is_homeless` is derived once, at package staging:

```sql
if(homeless_code in ('Y1', 'Y2'), true, false) as is_homeless
```

It reads `STUDENTCOREFIELDS` only, and is joined onto every enrollment year by `students_dcid` — so a student's current status is stamped across their entire history.

## The divergence

On any row where `academic_year` is earlier than `current_academic_year`, the two columns read different sources and can disagree. Non-Miami enrollment rows:

| Academic year | Rows | Disagree | SCF homeless, ren_x not | ren_x homeless, SCF not |
| ------------- | ------ | -------- | ----------------------- | ----------------------- |
| 2022 | 10,691 | 269 | 192 | 77 |
| 2023 | 11,838 | 315 | 168 | 147 |
| 2024 | 12,279 | 204 | 171 | 33 |
| 2025 | 13,050 | 115 | 113 | 2 |

903 rows total. The `s_nj_ren_x` value is the more accurate one for a closed year.

## Why this is not fixed alongside the OKRTS change

Re-deriving `is_homeless` from the year-resolved code would silently shift counts on every current consumer:

- `rpt_tableau__dibels_dashboard`
- `rpt_gsheets__grad_plan_tracking`
- `rpt_gsheets__student_contact_info`
- `int_extracts__student_enrollments_subjects_weeks`

Those are live dashboards. Changing them deserves its own reconciliation and its own communication to the teams that read them — not a side effect of adding two OKRTS columns.

## Proposed resolution

Pick one:

1. Re-derive `is_homeless` from the same year-resolved code so both columns agree, and reconcile the affected dashboards.
2. Keep `is_homeless` as a deliberate current-status flag and rename it to say so, e.g. `is_homeless_current`.
3. Leave both as-is and rely on the column descriptions shipped with the OKRTS change to document the difference.

[...]

Issue #4441, opened 2026-07-17
chore(kippmiami): remove legacy PowerSchool ODBC stack (SIS on Focus now)


cbini:
Sub-issue of #4427 — kippmiami **cleanup**, not migration. Miami's SIS moved to Focus (`code_locations/kippmiami/dlt/focus/`); its PowerSchool ODBC stack is still on main.

- [ ] Delete `code_locations/kippmiami/powerschool/` (assets, schedules, sensors, config) + its definitions wiring + the PS SSH/DB config in `code_locations/kippmiami/dagster-cloud.yaml`.
- [ ] Disable/remove Miami's PowerSchool dbt (`staging/odbc` variant for kippmiami); confirm no downstream mart still reads Miami PowerSchool (`code_locations/kippmiami/extracts/config/powerschool.yaml`, `src/dbt/kippmiami`).

Independent of the Newark/Camden migrations; can land any time.

PR #4815, opened 2026-08-11
feat(dbt): add decoded homeless fields to OKRTS extracts

### LINKED ISSUE #4814: feat(dbt): add decoded homeless status and nighttime residence to the OKRTS extracts
## Goal

Surface PowerSchool's `[STUDENTCOREFIELDS]Homeless_Code` and `[S_NJ_STU_X]HomelessPrimaryNighttimeRes` on the OKRTS dashboard, as decoded labels rather than raw codes.

Two new string columns on `rpt_tableau__okrts_behavior` and `rpt_tableau__okrts_referrals`:

| Column | Decodes | Values |
| ------ | ------- | ------ |
| `homeless_status` | `Homeless_Code` | `Not Homeless` / `Homeless - With Guardian` / `Homeless - Unaccompanied` |
| `homeless_primary_nighttime_residence` | `HomelessPrimaryNighttimeRes` | `Shelters, transitional housing` / `Doubled-up` / `Unsheltered` / `Hotels or Motels` |

Labels come from the PowerSchool dropdown definitions. The two `Y` labels are shortened; the residence labels are used as they appear. Residence code `0` maps to null — the dropdown's first entry is blank, so `0` reads as no selection. Unmapped or dirty codes fall to null rather than being guessed at.

## Source resolution

Both fields are read per academic year:

- current academic year: `stg_powerschool__studentcorefields` and `stg_powerschool__s_nj_stu_x`
- prior academic years: `stg_powerschool__s_nj_ren_x`

The re-enrollment extension holds the value that was true for that enrollment stint, but is not written until well after year end — it is empty for the current academic year. The student-level tables are maintained live but are overwritten in place, so they only reflect current status. Splitting by year gives an accurate current year and accurate history.

## Implementation

1. `base_powerschool__student_enrollments` (kipptaf) — add a join to `stg_powerschool__studentcorefields` (1:1 on `studentsdcid` plus code location; the `s_nj_stu_x` and `s_nj_ren_x` joins already exist). Emit the two year-resolved raw code columns with a single `if()` on `academic_year`.
2. `int_extracts__student_enrollments` — decode each raw code to its label with one flat `CASE`.
3. `int_extracts__student_enrollments_weeks` — carried through by the existing star projection; properties YAML entries only.
4. Both `rpt_tableau__okrts_*` models — project the two label columns and declare them in the contract YAML.

Tests: `accepted_values` at warn severity on the two raw code columns, so a new PowerSchool dropdown value surfaces instead of silently decoding to null. No `not_null` — nulls are legitimate.

No `contains_pii` tag: the repo reserves it for direct identifiers, and `is_homeless`, `lunch_status`, and `iep_status` are all untagged today.

## Known gaps, out of scope

- **Miami is blank.** `Homeless_Code` is null for all 3,943 Miami rows, and Miami has no `s_nj_stu_x` or `s_nj_ren_x` source. Both columns will be null for every Miami row in every year. Miami's homeless flag lives in Focus as `homeless_student_pk_12`.
- **`is_homeless` will disagree with `homeless_status` on historical rows** — 903 rows across AY2022 to AY2025. Tracked separately in #4813.
- **AY2026 is thinly populated** in both sources (361 and 269 rows network-wide, non-Miami). A data-entry gap for Ops, not something dbt can resolve.

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will add two decoded columns —
`homeless_status` and `homeless_primary_nighttime_residence` — to
`rpt_tableau__okrts_behavior` and `rpt_tableau__okrts_referrals`, surfacing
PowerSchool's `Homeless_Code` and `HomelessPrimaryNighttimeRes` on the OKRTS
dashboard as readable labels rather than raw codes.

Refs #4814
Refs #4813

### Decode

Labels come from the PowerSchool dropdown definitions. The two `Y` labels are
shortened; the residence labels are used as they appear.

| Column                                 | Values                                                                            |
| -------------------------------------- | --------------------------------------------------------------------------------- |
| `homeless_status`                      | Not Homeless / Homeless - With Guardian / Homeless - Unaccompanied                |
| `homeless_primary_nighttime_residence` | Shelters, transitional housing / Doubled-up / Unsheltered / Hotels or Motels      |

Residence code `0` maps to null — the dropdown's first entry is blank, so `0`
reads as no selection. Codes outside the mapped set decode to null rather than
being guessed at; the `accepted_values` tests on the raw codes surface anything
new instead of letting it silently disappear.

### Source resolution

Both fields are read per academic year:

- current academic year: `stg_powerschool__studentcorefields` and
  `stg_powerschool__s_nj_stu_x`
- prior academic years: `stg_powerschool__s_nj_ren_x`

The NJ re-enrollment extension records the value that was true for each
enrollment stint, but is not written until well after year end and is empty for
the current year. The student-level tables are maintained live yet overwritten
in place, so they describe only current status. Splitting by year gives an
accurate current year and accurate history.

Populated non-Miami enrollment rows, by source, showing why neither source
works alone:

| Academic year | `homeless_code` (ren_x) | `homeless_code` (SCF) | residence (ren_x) | residence (stu_x) |
| ------------- | ----------------------- | --------------------- | ----------------- | ----------------- |
| 2022          | 7,208                   | 5,196                 | 260               | 288               |
| 2023          | 7,611                   | 4,985                 | 310               | 377               |
| 2024          | 7,581                   | 4,027                 | 636               | 403               |
| 2025          | 8,157                   | 2,914                 | 726               | 365               |
| 2026          | 0                       | 361                   | 0                 | 269               |

### Implementation

1. `base_powerschool__student_enrollments` — adds a 1:1 join to
   `stg_powerschool__studentcorefields` (the `s_nj_stu_x` and `s_nj_ren_x`
   joins already existed) and emits the two year-resolved raw code columns.
   Resolving the source here keeps the decode downstream to a single flat
   `CASE`.
1. `int_extracts__student_enrollments` — decodes each raw code to its label.
   The raw codes are added to the `except` list so only labels flow downstream.
1. `int_extracts__student_enrollments_weeks` — carried through by the existing
   star projection; properties YAML entries only.
1. Both `rpt_tableau__okrts_*` models project the two labels and declare them
   in their contract YAML.

Not a breaking change: both models gain columns, and nothing is renamed or
removed.

### Known gaps, out of scope

- **Miami is blank.** `Homeless_Code` is null for all 3,943 Miami rows, and
  Miami has no `s_nj_stu_x` or `s_nj_ren_x` source, so both columns are null
  for every Miami row in every year. Miami's homeless flag lives in Focus as
  `homeless_student_pk_12`.
- **`is_homeless` disagrees with `homeless_status` on historical rows** — 903
  rows across AY2022 to AY2025. Re-deriving it would silently shift counts on
  the DIBELS dashboard, grad plan tracking, student contact info, and the
  subjects-weeks extract, so it deserves its own reconciliation. Tracked in
  #4813.
- **AY2026 is thinly populated** in both sources (361 and 269 rows
  network-wide, non-Miami). A data-entry gap for Ops.
- **The two PowerSchool fields disagree with each other at source** — 74 Newark
  students in AY2025 are marked not homeless yet carry a doubled-up residence
  code, with smaller counts across the other residence values. Both fields are
  reported faithfully; reconciling them is an Ops question.

### Local validation

`dbt build` against `dev` with `--defer --favor-state`, selecting every changed
model: `PASS=11 WARN=1 ERROR=0 SKIP=0`. The single warning is the pre-existing
`test_incorrect_student_number_pearson` (WARN 8), unrelated to these columns.

Verified in the built relations rather than from the build alone, since a view
build creates the relation without evaluating any rows: all seven labels render
as specified, Miami is null in both years, AY2025 shows 8,077 of 11,251
students populated (reading the re-enrollment extension) and AY2026 shows 346
(reading the live field).

Note for reviewers: three personal `zz_` district source copies had to be
re-cloned from prod before the build succeeded. They predated the powerschool
package dropping its contact columns, which made `base` 209 columns instead of
prod's 135 and collided with `int_extracts`. `source()` does not honor
`--favor-state`, so this is environment staleness, not a change in this PR.

## AI Assistance

Claude Code co-authored this PR. Human-directed: the decision to surface these
two fields, the choice of decoded labels over raw codes, the authoritative
PowerSchool dropdown definitions, the instruction to shorten the `Y` labels and
keep the residence labels as-is, and the current-year-versus-history source
split. AI-assisted: lineage tracing, the source coverage and divergence
profiling, model and YAML edits, local validation, and this description.

### General

### dbt

      models; every new column is documented in the existing properties files.
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — the
      `okrts_dashboard` exposure already depends on both models; no change
      needed.
- [x] **Breaking change?** No — additive columns only, nothing renamed or
      removed.
- [x] External sources — none added or modified.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

### DIFF
--- a/src/dbt/kipptaf/models/powerschool/base/base_powerschool__student_enrollments.sql
+++ b/src/dbt/kipptaf/models/powerschool/base/base_powerschool__student_enrollments.sql
@@ -43,7 +43,7 @@ with
     )
 
 select
-    ar.* except (lep_status, lunchstatus, spedlep, prevstudentid),
+    ar.* except (lep_status, lunchstatus, spedlep, prevstudentid, homeless_code),
 
     -- same value as _dbt_source_project, named for the Dagster code location;
     -- projected here rather than re-derived from _dbt_source_relation (#3142)
@@ -148,7 +148,7 @@ select
     both the live value and the history accurate (#4814). */
     if(
         ar.academic_year = {{ var("current_academic_year") }},
-        scf.homeless_code,
+        ar.homeless_code,
         njr.homeless_code
     ) as homeless_code,
 
@@ -245,10 +245,6 @@ left join
     {{ ref("stg_powerschool__s_nj_ren_x") }} as njr
     on ar.reenrollments_dcid = njr.reenrollmentsdcid
     and ar._dbt_source_project = njr._dbt_source_project
-left join
-    {{ ref("stg_powerschool__studentcorefields") }} as scf
-    on ar.students_dcid = scf.studentsdcid
-    and ar._dbt_source_project = scf._dbt_source_project
 left join
     {{ ref("stg_people__student_logins") }} as sl
     on ar.student_number = sl.student_number
--- a/src/dbt/kipptaf/models/powerschool/base/base_powerschool__student_enrollments.sql
+++ b/src/dbt/kipptaf/models/powerschool/base/base_powerschool__student_enrollments.sql
@@ -140,6 +140,24 @@ select
 
     if(adb.latest_fafsa_date is null, 'No', 'Yes') as salesforce_contact_df_has_fafsa,
 
+    /* The NJ re-enrollment extension records homelessness per enrollment stint,
+    but is not written until well after year end -- it is empty for the current
+    year. The student-level tables are maintained live yet overwritten in place,
+    so they describe only current status. Reading the current year from the
+    student-level tables and closed years from the re-enrollment extension keeps
+    both the live value and the history accurate (#4814). */
+    if(
+        ar.academic_year = {{ var("current_academic_year") }},
+        scf.homeless_code,
+        njr.homeless_code
+    ) as homeless_code,
+
+    if(
+        ar.academic_year = {{ var("current_academic_year") }},
+        njs.homelessprimarynighttimeres,
+        njr.homelessprimarynighttimeres
+    ) as homeless_primary_nighttime_residence_code,
+
     coalesce(
         if(ar.region in ('Miami', 'Paterson'), ar.spedlep, sped.spedlep), 'No IEP'
     ) as spedlep,
@@ -227,6 +245,10 @@ left join
     {{ ref("stg_powerschool__s_nj_ren_x") }} as njr
     on ar.reenrollments_dcid = njr.reenrollmentsdcid
     and ar._dbt_source_project = njr._dbt_source_project
+left join
+    {{ ref("stg_powerschool__studentcorefields") }} as scf
+    on ar.students_dcid = scf.studentsdcid
+    and ar._dbt_source_project = scf._dbt_source_project
 left join
     {{ ref("stg_people__student_logins") }} as sl
     on ar.student_number = sl.student_number
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__okrts_behavior.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__okrts_behavior.yml
@@ -59,6 +59,16 @@ models:
         data_type: string
       - name: self_contained_status
         data_type: string
+      - name: homeless_status
+        data_type: string
+        description:
+          Decoded homelessness status for the enrollment year. Null for Miami,
+          which does not populate the underlying PowerSchool field.
+      - name: homeless_primary_nighttime_residence
+        data_type: string
+        description:
+          Decoded primary nighttime residence for the enrollment year. Collected
+          in New Jersey only, so null elsewhere.
       - name: behavior_month
         data_type: int64
       - name: school_enrollment_by_week
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__okrts_referrals.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__okrts_referrals.yml
@@ -113,6 +113,16 @@ models:
         data_type: string
       - name: self_contained_status
         data_type: string
+      - name: homeless_status
+        data_type: string
+        description:
+          Decoded homelessness status for the enrollment year. Null for Miami,
+          which does not populate the underlying PowerSchool field.
+      - name: homeless_primary_nighttime_residence
+        data_type: string
+        description:
+          Decoded primary nighttime residence for the enrollment year. Collected
+          in New Jersey only, so null elsewhere.
       - name: entry_staff
         data_type: string
       - name: last_update_staff
--- a/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__okrts_behavior.sql
+++ b/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__okrts_behavior.sql
@@ -139,6 +139,8 @@ select
     co.ml_status,
     co.status_504,
     co.self_contained_status,
+    co.homeless_status,
+    co.homeless_primary_nighttime_residence,
     co.quarter as term,
     co.week_start_monday,
     co.week_end_sunday,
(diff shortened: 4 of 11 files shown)


anthonygwalters:
## Review verdicts

**1. Inconsistent YAML list style within the same file (Low) — declined, no change.**

Verified both citations before deciding: `int_students__graduation_path_codes.yml`
does use block style, and `stg_google_sheets__people__cube_access_role.yml` does
use flow style, so neither is an enforced convention and the finding is correctly
labelled cosmetic. Keeping the mixed styles deliberately — the string list gives
`N` its own line, where a bare `N` inside a flow sequence invites a
double-take about YAML boolean coercion (PyYAML does not coerce it, confirmed,
but it still reads ambiguously), while the five-element integer list is short
enough that flow style stays legible.

Separately, I spot-checked the review's load-bearing dismissal rather than
taking it on trust, since a fan-out on the new join would be the one thing here
that could quietly corrupt row counts. `stg_powerschool__studentcorefields.studentsdcid`
does carry `unique` at `config.severity: error`. That agrees with the grain I
measured directly before adding the join — 30,701 rows, 30,701 distinct
`(studentsdcid, code location)` keys — so the LEFT JOIN is 1:1 by both
enforcement and measurement.

One caveat on the review's scope note for other readers: it states the author's
local `dbt build` covers execution. That build did pass at PASS=11 WARN=1
ERROR=0, but three of the five models involved are views, and a view build
creates the relation without evaluating a row. The value-level evidence in the
PR description comes from querying the built relations afterwards, not from the
build result.

src/dbt/kipptaf/models/powerschool/base/base_powerschool__student_enrollments.sql
```diff
@@ -140,6 +140,24 @@ select
...
 
+    /* The NJ re-enrollment extension records homelessness per enrollment stint,
+    but is not written until well after year end -- it is empty for the current
+    year. The student-level tables are maintained live yet overwritten in place,
+    so they describe only current status. Reading the current year from the
+    student-level tables and closed years from the re-enrollment extension keeps
+    both the live value and the history accurate (#4814). */
+    if(
+        ar.academic_year = {{ var("current_academic_year") }},
+        scf.homeless_code,
```
cbini: surface this field on the powerschool package

cbini (changes requested):

src/dbt/kipptaf/models/powerschool/base/base_powerschool__student_enrollments.sql
anthonygwalters: will do - did it this way to avoid 2 PRs, but it makes sense

cbini: you can do it in one, just need to build staging locally

anthonygwalters: Done in `967df0bb7` (committed, not yet pushed — see below). `homeless_code`
now comes off `scf` in the package base right next to the `is_homeless` flag
derived from it, and the kipptaf-level `studentcorefields` join is deleted.
kipptaf reads the code through the union and excepts the passthrough so it does
not collide with the year-resolved output column of the same name.

Verified equivalent rather than assumed: joined the new union path against the
old direct-join source over current-year rows and got zero value mismatches
across 10,250 rows, 383 populated on both sides. Built the package model into
dev for kippnewark, kippcamden and kipppaterson first, then rebuilt the kipptaf
chain — PASS=11 WARN=1 ERROR=0, the warning being the pre-existing
`test_incorrect_student_number_pearson`.

Miami null-fills through the union superset as expected: it left powerschool
for Focus (#4441) and has no `homeless_code` either way, so the column is
present and null for its rows rather than missing.

Holding the push until the `zz_stg` staging build runs. Pushing first would
give a knowingly-red dbt Cloud check — CI resolves the union superset from
`zz_stg_kipp{newark,camden,paterson}_powerschool`, which will not carry
`homeless_code` until the package model is built there. Deliberate hold, not a
blocker.

anthonygwalters:
@claude review the refactor

PR #4767, opened 2026-08-07
Edited descirption for attendance dashboard

ChideraOkaforNJIT:
# Pull Request

## Summary & Motivation

> "When merged, this pull request will..."

## AI Assistance

> If Claude Code authored or co-authored this PR, briefly note what was
> AI-assisted vs. human-directed in the summary above

## Self-review

> Complete only the sections relevant to your changes.

### General

> If this is a same-day request, please flag that in the #data-team Slack

- [ ] Update **due date** and **assignee** on the
      [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
- [ ] Review the **Claude Code Review** comment posted on this PR. Address valid
      feedback; dismiss false positives with a brief reply explaining why.
      (Claude is advisory — use your judgement, but don't ignore it.)

### Dagster _(skip if no Dagster changes)_

- [ ] Run `uv run dagster definitions validate` for any modified code location
- [ ] Run `uv run pytest tests/test_dagster_definitions.py` for any modified
      code location
- [ ] New integrations follow the
      [Library + Config pattern](https://teamschools.github.io/teamster/reference/adding-an-integration/)
      with the correct asset key format and IO manager

### dbt _(skip if no dbt changes)_

- [ ] Include a `[model_name].yml` properties file for all new models — see
      [dbt Conventions](https://teamschools.github.io/teamster/reference/dbt-conventions/#model-properties-file)
- [ ] Include (or update) an
      [exposure](https://teamschools.github.io/teamster/reference/dbt-conventions/#exposures)
      for all models consumed by a dashboard, analysis, or application
- [ ] **Breaking change?** Renaming or removing columns in a contracted model
      (`stg_`, `rpt_`, mart) will break downstream exposures. Coordinate with
      affected teams before merging and document your rollback plan in the
      summary above.
- [ ] If adding or modifying an external source, run `stage_external_sources`
      with **`--target staging`** so the dbt Cloud CI job can find the table.
      See
      [Staging external sources](https://teamschools.github.io/teamster/guides/dbt-development/#staging-external-sources)

### Docs _(skip if no schedule, sensor, or integration changes)_

- [ ] If adding or changing a schedule or sensor, regenerate the automations
      catalog: `uv run scripts/gen-automations-doc.py`
- [ ] If adding a new integration to a code location, update that location's
      `CLAUDE.md` by running `/claude-md-management:revise-claude-md`

## CI checks

[...]

PR #4819, opened 2026-08-11
docs(launch): design spec for the catalog build and CI gate

### LINKED ISSUE #4818: feat(launch): validation gate and pytest job for the tool catalog build
Productionize the prototype that renders `src/launch/links.yml` into the staff launch page, and build the CI gate that has to exist before it can be wired into publishing.

## Why now

Zero of 46 catalog entries are `verified` today, so the generated page is empty. That makes this the safest possible moment to stand up the pipeline — it cannot publish anything wrong yet.

When #4767 merges, nine entries flip to verified at once. That merge is the first live publish. Because the MkDocs deploy workflow also fires on `docs/**`, a bad catalog on `main` freezes all documentation publishing, not just the launch page. The gate needs to exist first.

## Scope

- `src/launch/build.py` — load / select / validate / render, reading the working tree rather than `git show origin/main:`, and returning an HTML string rather than writing a file
- Two-tier validation: structural rules over every entry, publish-readiness rules only over `status: verified` entries
- A minimum-verified threshold, configured in `groups.yml`, below which no page is generated at all
- Generation via an MkDocs `on_files` hook using `File.generated`, so the page never touches `docs_dir`
- `tests/launch/` — rule-level unit tests plus one test that runs the real catalog through tier 1
- A new pytest PR workflow, scoped to `tests/launch`
- Preview artifact: the gate uploads the rendered page for review before merge
- `src/launch/**` added to the deploy workflow path filter, which does not watch it today

## Explicitly out of scope

- Wiring the build into the deploy job beyond the path filter
- The per-entry `group` field and grouped rendering — sequenced after #4767 to avoid conflicting with in-flight verification work
- `tests/cube` remains unprotected by CI
- The 80 pre-existing `docs/superpowers` link warnings

## Prerequisite outside this work

The new check has no force until an admin adds it to ruleset `816683`, which currently requires only dbt Cloud and Trunk. `strict_required_status_checks_policy` is `false`, so two PRs green in isolation can still break `main`.

Refs #4761

anthonygwalters:
# Pull Request

## Summary & Motivation

...add the design spec for the launch page build pipeline and the CI gate that
has to exist before it can publish, plus a project overview for the team.
**Docs-only — two files, no code.**

`src/launch/PROJECT.md` is the orientation doc: the problem, the content split,
live state of every issue and PR in this project, who owns what, the decisions
that are settled, what still needs a human, and the sequence. It sits alongside
`README.md` (field reference) and `RUNBOOK.md` (task sequence) so the three read
as a set — why, what, how. Start there if you are picking this up cold.

### Why this is urgent rather than routine

Zero of 46 catalog entries are `verified` on `main` today, so the generated page
is empty. That makes now the safest possible moment to stand up the pipeline: it
cannot publish anything wrong, because it has nothing to publish.

When #4767 merges, nine entries flip to verified at once and that merge becomes
the first live publish. Because the MkDocs deploy workflow also fires on
`docs/**`, a catalog that fails validation on `main` freezes **all**
documentation publishing, not just the launch page. The gate needs to exist
before that merge.

### The design

Four functions in `src/launch/build.py` — load, select, validate, render —
called from an MkDocs `on_files` hook. One code path serves `mkdocs serve`
locally, the PR gate, and the deploy. `render()` returns a string; nothing in
the module writes to disk.

**Two-tier validation.** Structural rules apply to every entry: unique `id`,
known `system`, HTTPS `url`, family members that resolve. Publish-readiness
rules apply only to `status: verified` entries: non-empty description, at least
one audience, a region if the entry is in a family. A half-finished entry never
blocks a commit; nothing reaches staff malformed.

**A minimum-verified threshold**, configured in `groups.yml` and seeded at 25.
Below it the hook generates no page at all and the URL 404s.

### Three findings from adversarial review changed the design

Each reproduced before being applied.

- **`mkdocs build --strict` cannot be the gate.** It fails on `main` right now —
  exit 1, exactly 80 warnings, every one an unresolvable relative link inside
  `docs/superpowers/**`. The live deploy passes only because it does not pass
  `--strict`. Dropped; `CatalogError` fails the build on its own.
- **Writing into `docs_dir` from `on_pre_build` loops `mkdocs serve`.** Measured
  on a minimal project: one edit produced 61 builds in 30 seconds. `serve.py`
  watches `docs_dir` recursively and livereload clears its rebuild flag before
  running the build, so the hook's own write is re-detected mid-build. Moved to
  `on_files` with `File.generated`, which never touches `docs_dir` — and which
  also removes the gitignore entry and the stale-artifact problem.
- **Selection has to run before validation.** Otherwise the two tiers collapse
  into one, and the collapsed version fails on `main` today.

### Two things the review found that are not mine to fix

The **deploy workflow does not watch `src/launch/` at all**. It fires on
`docs/**` and `mkdocs.yml` only, so merging a catalog change currently triggers
nothing. The earlier launch page spec claims otherwise; that claim is false
until this lands. One line, included in the implementation.

**Required checks do not cover this, and are not up-to-date-gated.** Ruleset
`816683` requires only dbt Cloud and Trunk, and
`strict_required_status_checks_policy` is `false`. Until an admin adds the new
check the gate is advisory, and without strict mode two PRs green in isolation
can still break `main`.

Refs #4818, #4761

## AI Assistance

Claude drafted the spec. Every architectural decision was human-directed: the
preview-artifact approach over committing the rendered page, generation via a
hook rather than a workflow step, scoping the pytest job to `tests/launch` only,
two-tier validation, the threshold gate over an empty-state design, and
deferring the per-entry `group` field so it does not conflict with the 37
entries in flight on #4767.

An adversarial reviewer was dispatched against the architecture before the spec
was written. It returned three blockers with reproductions, all of which held
under independent verification and all of which changed the design. A spec
self-review then caught four more gaps — an undefined rendering behaviour during
the deferral window, a validation rule asserted in prose but absent from the
rule table, an unstated disposition for the prototype's `assignments` block, and
an unhandled artifact upload below the threshold.

## Self-review

### General

N/A — docs only.

N/A — docs only.

N/A — no schedule, sensor, or integration changes.

## CI checks

- [ ] **Trunk** — verified locally with `trunk check --force` before pushing; no
      issues.
- [ ] **Dagster Cloud** — not triggered; no watched paths change.

## Reviewer notes

Two decisions in the spec are open and belong to the catalog owner rather than
to this design:

1. **The threshold number**, seeded at 25 of 46. One line in `groups.yml`.
1. **Whether to enable `strict_required_status_checks_policy`** on ruleset
   `816683`, or add a merge queue for `src/launch/**`.

A third is a loose end from the earlier design: `views.yml` still exists on
`main` with per-view titles and intro copy, and the launch page spec says that
copy moves into the template — but the prototype template contains none of it.

**Note on merging.** This branch is linked to #4818, and a linked branch closes
its issue on PR merge regardless of body keywords. #4818 covers the
implementation, not just the spec, so it will need reopening after this merges —
the same thing happened to #4761 when #4763 landed.

### DIFF
--- a/docs/superpowers/specs/2026-08-11-launch-page-build-gate-design.md
+++ b/docs/superpowers/specs/2026-08-11-launch-page-build-gate-design.md
@@ -1,0 +1,313 @@
+# Launch page: build, validation, and CI gate
+
+Design for productionizing the launch page generator and building the CI gate
+that must exist before it can publish. Companion to
+[the launch page design](2026-08-06-launch-page-design.md), which settles what
+the page is; this settles how it gets built and what stops it shipping broken.
+
+## Context
+
+`src/launch/links.yml` holds 46 tool entries scraped from the Google Site and
+merged in #4763. Verification is in progress on #4767. Only `status: verified`
+entries publish, and **zero of 46 are verified on `main` today**.
+
+That makes now the safest moment to stand up the pipeline: it cannot publish
+anything wrong, because it has nothing to publish. When #4767 merges, nine
+entries flip to verified at once, and that merge becomes the first live publish.
+The MkDocs deploy workflow also fires on `docs/**`, so a catalog that fails
+validation on `main` freezes **all** documentation publishing. The gate has to
+exist before that merge, not after.
+
+## What this delivers
+
+- `src/launch/build.py` — a four-function module, unit-testable without git,
+  temp files, or subprocesses
+- Two-tier validation, so in-progress entries never block a commit but nothing
+  reaches staff malformed
+- A minimum-verified threshold below which no page is generated at all
+- Generation through an MkDocs `on_files` hook
+- `tests/launch/` and a pytest PR workflow scoped to it
+- A downloadable preview of the rendered page on every catalog PR
+
+## Architecture
+
+```text
+  src/launch/links.yml      catalog, 46 entries, each with status:
+  src/launch/groups.yml     group defs, families, promos, publish threshold
+  src/launch/template.html  page shell, __CATALOG__ placeholder
+            |
+            v
+  src/launch/build.py
+      load()      read the three files from the working tree
+      select()    keep only status: verified
+      validate()  tier 1 over all entries, tier 2 over selected
+      render()    inject JSON, RETURN an HTML string
+            |
+            +--> errors        -> raise CatalogError, listing every problem
+            +--> below minimum -> return None, log the count
+            |
+            v
+  docs/hooks.py :: on_files
+      File.generated(config, "launch/index.html", content=html)
+            |
+            v
+  site/launch/index.html -> gh-pages -> /teamster/launch/
+```
+
+One entry point, three callers: `mkdocs serve` locally, `mkdocs build` in the PR
+gate, `mkdocs gh-deploy` on merge. There is no second code path to drift.
+
+### Why `on_files` and not `on_pre_build`
+
+The obvious design — write `docs/launch/index.html` during `on_pre_build` and
+let MkDocs collect it — **causes an infinite rebuild loop under
+`mkdocs serve`**. Measured on a minimal project: one edit to a source file
+produced 61 builds in 30 seconds, roughly two per second, indefinitely.
+
+`serve.py` watches `docs_dir` recursively, and livereload clears its
+`_want_rebuild` flag _before_ running the build, so the hook's own write is
+re-detected as a user change mid-build. Identical content does not help; writing
+the file bumps its mtime.
+
+`File.generated(config, src_uri, content=...)` — present in mkdocs 1.6.1, which
+is what resolves here — registers a virtual file that never touches `docs_dir`.
+Verified against a minimal project: the page lands at `site/launch/index.html`,
+`docs_dir` stays clean, the file is absent from `sitemap.xml`, and a single edit
+produces two builds rather than 61.
+
+This also removes the need for a `.gitignore` entry and eliminates a class of
+bug where a stale generated file lingers in a local checkout.
+
+### `render()` returns a string
+
+Nothing in `build.py` writes to the filesystem. The hook decides where output
+goes. This is what makes every test run on in-memory data, and it is a
+structural guarantee that the serve loop cannot return.
+
+## Validation rules
+
+Selection runs **before** validation, so tier 2 knows which entries it is
+judging. Getting this backwards would force every rule into tier 1.
+
+### Tier 1 — every entry, regardless of status
+
+| Rule                                                             | Why                                                            |
+| ---------------------------------------------------------------- | -------------------------------------------------------------- |
+| `links.yml` parses as a list of mappings                         | Everything else depends on it                                  |
+| `id` present, unique, matching `^[a-z0-9_]+$`                    | Duplicates surface at the worst possible moment                |
+| `name` present and non-empty                                     | Nothing can reference an unnamed entry                         |
+| `url` present, scheme is `https`                                 | A staff-facing link must not be plaintext                      |
+| `system` in the nine-value enum documented in `README.md`        | The prototype hard-codes four; the other five degrade silently |
+| `status` in `{needs-review, verified}`                           | A typo would silently unpublish an entry                       |
+| `access`, if present, is exactly `limited`                       | Prototype treats any truthy value as limited                   |
+| `regions` values all in `{newark, camden, miami, paterson, all}` | Unknown regions render blank                                   |
+| Every family member in `groups.yml` names a real entry           | A rename silently drops a tool                                 |
+| Every family names a real group id                               | An unknown group throws before anything renders                |
+
+### Tier 2 — `status: verified` entries only
+
+| Rule                                                  | Why                                                 |
+| ----------------------------------------------------- | --------------------------------------------------- |
+| `description` present and non-empty                   | An unlabelled tile is useless                       |
+| `audiences` non-empty, all values known               | Four entries have none today                        |
+| A family member carries exactly one region, not `all` | Three GPA Rosters render blank buttons without this |
+| `guide`, if present, is `https`                       | Same reason as `url`                                |
+
+A half-finished `needs-review` entry never fails CI. Flipping one to `verified`
+while it is still missing a description does, and the error names the entry and
+the field.
+
+### Deferred
+
+`group` present on every entry and resolving to a known group id. The
+[launch page design](2026-08-06-launch-page-design.md) records `group` as a
+required per-entry field, and that is the right shape — a side table in
+`groups.yml` has to stay exhaustive, which is precisely the coupling that lets
+two individually-green PRs break `main` together.
+
+Adding it means touching all 46 entries while 37 are in flight on #4767.
+Sequenced into a follow-up PR after that merges. Nothing is lost by waiting:
+with zero verified entries there is no grouped rendering to do yet.
+
+## Publication threshold
+
+Below a configured number of verified entries the hook generates **no page at
+all**, logs the count, and lets the docs build succeed. `/teamster/launch/`
+returns 404 until the catalog is genuinely ready.
+
+This exists because the zero-verified page is not merely empty — it is broken.
+With no tools the template falls through to its search-miss branch and renders
+"No tool matches that", above a masthead reading `0 tools` and a footer
+disclaiming `0 of 0 entries are still being verified`. Four of the five promo
+cards in `groups.yml` point at `#`. The first live version would be five cards,
+four of them dead, under an error message.
+
+Gating on a count is cheaper than designing an empty state that should never be
+seen, and it makes a broken intermediate page impossible rather than merely
+unlikely.
+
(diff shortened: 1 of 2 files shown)


anthonygwalters:
@claude execute an adversarial review of this spec

anthonygwalters:
Verified all five against repo state before applying. Four accepted, one accepted with a correction to the reasoning. Fixed in `3d0707b02`.

### 1. No import path for `tests/launch` — accepted, and it is worse than described

Confirmed: `pyproject.toml` packages only `src/teamster`, there is no `[tool.pytest.ini_options] pythonpath`, no `pytest.ini`, `tests/conftest.py` does no path manipulation, and `tests/` has no root `__init__.py`.

The observation that `tests/cube/test_cube_schema.py` sidesteps the problem is right — it only reads YAML through `pathlib.rglob`, never imports `src/cube`. But there **is** a precedent in the same directory that the review missed, and it is evidence for the finding rather than against it: `tests/cube/test_mcp_server.py:15-28` reaches `src/cube/mcp/server.py` through `importlib.util.spec_from_file_location`, which is what you write precisely because no import path exists.

Specced a `tests/launch/conftest.py` that inserts `src` on `sys.path`, anchored on `Path(__file__).resolve().parents[2]`. Verified empirically on a scratch tree — `src/launch` imports as a PEP 420 namespace package with no `__init__.py`, and a test importing `launch.build` passes.

Picked over the three alternatives for stated reasons, now in the spec: `[tool.pytest.ini_options] pythonpath` is global config on a tree that already fails to collect; adding `src/launch` to the hatch `packages` list would ship it inside the `teamster` wheel, which it is not part of; `spec_from_file_location` works but is a heavy idiom for a module we own. The shim also matches what `docs/hooks.py` does, so there is one mechanism rather than two.

The "out of scope" pythonpath bullet now says explicitly that it refers to the pre-existing collection failure in `tests/sensors/sftp/test_sensors_sftp_renlearn.py`, and that the shim neither touches it nor depends on it.

### 2. Companion doc not on `main` — accepted

Correct, and it is the same defect class the spec complains about elsewhere. All references are now by PR number (#4762) rather than relative link, with a note at the top saying where the file lives and why it is cited that way.

### 3. Threshold versus tier-1 ordering — accepted, this was a real contradiction

The diagram put the short-circuit in `select()` and the CI section claimed validation still gates below the threshold. Both cannot be true, and the diagram would have been the one an implementer followed.

Resolved in favour of the CI section, which was describing the behaviour actually wanted. `select()` now only partitions and never exits; the threshold is checked in `render()` after every rule has run. The pre-threshold window is long — today, and for a stretch after this PR lands nine of 46 — so it is exactly the period that most needs a structural gate on the build path.

The failure table and the `test_select.py` / `test_render.py` responsibilities moved with it, and the decision is in the log.

### 4. Prototype claims unverifiable — accepted

Fair. The spec now says the prototype is gitignored and its behavioural claims are unverifiable until the implementation PR turns them into a diff.

Two of the data claims do not depend on it and are checkable against `main` today, so they are called out separately: `links.yml` has three GPA Roster entries and exactly four `regions:` keys in the whole file, all four on the Contact Info Feeds. The promo-card count is prototype-only, since `groups.yml` does not exist in the repo yet.

### 5. Path-filter ordering — accepted, with a correction

The conclusion is right and the reasoning needed fixing in both directions. `src/launch/**` and `docs/superpowers/**` are disjoint prefixes, so no path can match both and position in the list is irrelevant. My original text cited GitHub's last-match-wins rule, which is a real rule that simply does not bear on this pair. Trimmed to say the patterns are disjoint.

### On the unverified item

The 80-warning figure was reproduced independently before the spec was written:

```text
$ uv run --group docs mkdocs build --strict
Aborted with 80 warnings in strict mode!
EXIT=1
```

All 80 are `validation.links.not_found` under `docs/superpowers/**`. Noting it since the review flagged it as unverified-by-me rather than disputed.

anthonygwalters:
@claude re-run the adversarial review against the revised spec at `3d0707b02`.

The previous pass produced five findings, all now addressed. Do not re-review the spec from scratch — attack the **revisions**, and treat the earlier findings as claims to be re-tested rather than as settled.

Specifically:

1. **The import shim.** The spec now specs a `tests/launch/conftest.py` that inserts `src` on `sys.path` via `Path(__file__).resolve().parents[2]`, and asserts `src/launch` will import as a PEP 420 namespace package with no `__init__.py`. Test that claim. Does it survive pytest's rootdir handling, `-p no:cacheprovider`, running from a different cwd, and running `pytest tests/launch/test_render.py` directly rather than the directory? Does it collide with the existing `tests/cube/__init__.py`, or with any other conftest in the tree?

2. **The threshold moved from `select()` to `render()`.** The claim is that validation now always runs in full below the publication threshold. Check the diagram, the failure table, the Testing section and the decisions log actually agree with each other now, and that no other passage still implies the old ordering.

3. **`docs/hooks.py` uses the same `sys.path` mechanism as the conftest.** Two importers, one module. Is that actually one mechanism or two that merely look alike? What happens if mkdocs and pytest run in the same process, or if `launch` is already in `sys.modules` from a prior build?

4. **The promo-card and `groups.yml` tier-1 rules** were added after the first review. `groups.yml` does not exist in the repo yet, so those rules are specced against a file only the gitignored prototype has. Is anything specified there unimplementable or self-contradictory?

5. **Anything the first pass missed** because it was looking at the pre-revision text.

Verify against repo state rather than the prose. Where you cannot execute something, say so explicitly rather than inferring. If a revision fully resolves a finding, do not restate it — silence means it held.

anthonygwalters:
All six addressed in `5b7756142`. One correction to my trigger comment accepted, and one finding is a round-one miss of mine rather than a new defect.

### 1. Import shim — no action

Holds. The `tests/cube/__init__.py` analysis is a useful addition and is now recorded in the spec: a regular package with a loader always beats a same-named namespace portion, and `src/cube/` has no `__init__.py`, so `import cube` keeps resolving to `tests/cube` after `src` joins `sys.path`. The `uv.lock` check for a `launch` collision is worth having too.

### 2. `test_select.py` row — fixed

Correct. `select()` has no threshold awareness under this design, so a unit test scoped to it has nothing to assert about validation continuing. That guarantee lives in the `validate()`/`render()` call order. The row now reads "Status filtering only", and `test_catalog.py` picked up "at any entry count", which is where the property is actually observable.

### 3. `docs/hooks.py` — accepted, and this one is mine from round one

Two distinct problems, both real.

**The production-side import was never specified.** The spec asserted the conftest shim "matches what `docs/hooks.py` does" while `docs/hooks.py` does nothing of the kind — it is four lines with no imports. Worse, the spec never said how the hook reaches `launch.build` at all, which left the more important of the two importers undefined. Now specified: the same `sys.path.insert`, anchored on `Path(__file__).resolve().parent.parent / "src"` rather than cwd, because `mkdocs build -f` from an arbitrary directory is supported. The text now also says plainly that `hooks.py` does not do this today and that the shim arrives with this work.

**The `watch:` and reload gaps are a round-one finding I failed to carry over.** The first reviewer raised both as a MAJOR — that `mkdocs serve` neither watches `src/launch` nor picks up an edited `build.py` across rebuilds. I fixed that round's three blockers and did not carry this into the spec. Confirmed: `mkdocs.yml` has no `watch:` key, and `src/launch/` is outside `docs_dir`, so a catalog edit during `mkdocs serve` triggers no rebuild at all.

Both now specced: `watch: [src/launch]` in `mkdocs.yml`, and an `importlib.reload()` when the module is already in `sys.modules`. The framing is worth keeping — it is the round-one loop bug's mirror image, not watched rather than over-watched, and it breaks precisely the local-preview caller that "one entry point, three callers" was meant to protect.

### 4. Promo cards — correction accepted, contradiction fixed, task added

**The correction is right.** Those rows were added in `8633a80de` at 14:29:49Z; the first review comment posted at 14:36:11Z. They predate it. My trigger comment said they were added after round one, which was wrong.

**The contradiction was real.** The threshold section asserted four dead promo cards as present-tense fact about the repo, while the prototype section — rewritten last round to walk back exactly that class of claim — said `groups.yml` does not exist here. Both now attribute the count to the prototype explicitly, and the threshold section cross-references the walk-back.

**The unstated dependency is the most useful part of this finding.** The tier-1 rule rejects the prototype's own seed data, so `groups.yml` cannot be lifted across unchanged, and nothing assigned that as work. Added to the "Fixed on the way" list with a default: drop the four cards rather than invent URLs, since a promo pointing at a page that does not exist is worse than no promo. Also in the decisions log.

The "Same" rationale on the `groups.yml` parse row was a copy-paste that did not track its own claim. Now reads "A missing key raises deep in `render()`, not at validation."

### 5. Tier 1 heading — fixed

Right that "every entry, regardless of status" is narrower than the table beneath it. Retitled to "structural, regardless of status", with a line noting that the last four rows check `groups.yml` structures, which have no `status` of their own.

### On the execution limitation

Noted, and it is the same constraint the first pass hit. The two claims that most needed executing were checked outside the sandbox: `mkdocs build --strict` aborting with 80 warnings, and the conftest shim importing `src/launch` as a namespace package with no `__init__.py`. Everything else in this round was resolvable by inspection, which is what you did.

PR #4820, opened 2026-08-11
refactor(finalsite): make the contacts pull incremental and land it before its consumers

### LINKED ISSUE #4715: refactor(finalsite): make the contacts API pull incremental instead of a full nightly snapshot
## Context

`build_finalsite_asset` (`src/teamster/libraries/finalsite/api/assets.py`) does one thing: `finalsite.list(path=asset_name)`, which walks the cursor to the end and returns every record, then writes the whole result as one non-partitioned GCS Avro object. Every tick is a full snapshot replace. There is no cursor persistence and no change filter — all four code locations pass only `includes=contacts.relationships`.

Cost of that today:

- Pagination is sequential and cursor-only, and the API caps `count` at 25 records/page, so per-record request cost is fixed. Roughly 1 request/second. kippnewark (~24k contacts) takes about 20 minutes.
- All four districts share the `finalsite_api` pool at limit 1, because the gateway throttles by shared egress IP and returns a bare `403` under concurrency (#4408). So the districts run one after another — the nightly `0 4 * * *` window is the sum, not the max.
- kippnewark carries `MAX_RUNTIME_SECONDS_TAG: 3600` purely to survive its own pull length.
- Freshness is capped at once daily, which is the real constraint: the Focus enrollment and contacts feeds (#4073, #4346, #4652) can only ever be a day stale, and running more often is not affordable at full-snapshot cost.

The API spec is vendored at `docs/superpowers/specs/references/finalsite-api-spec.yml`.

## What the API supports (confirmed against the vendored spec)

`GET /api/external/contacts` takes two relevant parameters we do not use:

- **`since`** — "Filter results to contacts updated on or after the given date."
- **`since_includes_expanded`** — "By default `since` only includes contacts that have been updated since the date. If true, contacts will be included if any records being expanded have been updated on or after the since date as well." We expand `relationships`, so this must be `true` or a relationship-only change (a new guardian, a flipped `primary`) is silently missed.

Both go straight through `_request`'s `params`, so the request side of this is nearly free. Two constraints shape the design:

- **`since` is date-grained, not timestamp-grained** (`type: date`). The finest possible increment is one day. An intraday cadence means repeatedly asking for "since yesterday" and accepting overlap, not a tight watermark.
- **The `Contact` response carries no change timestamp.** No `updated_at`, no `created_at` — only `birth_date`, the three `*_submit_date` fields, and billing dates (`billing_overdue_at`, `next_billing_date`). Nested `User.updated_at`, `Grade.created_at` and `SchoolYear.created_at` exist but are not the contact's. So the watermark cannot be derived from the landed data; it has to be stored by us (Dagster asset metadata, a sensor cursor, or dlt `resource_state`) and advanced only on a successful pull.

Overlap plus no per-record timestamp means the destination needs **merge or dedupe semantics keyed on `id`** — a `replace` of a partial result would truncate the other 23k contacts. That, not the request change, is the bulk of the work.

## Goal

Pull only contacts changed since the last successful sync on a normal tick, keep a periodic full refresh for correctness, and leave asset keys (`[{code_location}, finalsite, contacts]`) and the downstream dbt source working without model edits.

Shape to copy: #4453 / #4446 / #4447 — gated incremental on the frequent tick, unconditional full refresh nightly. The gating mechanism differs (an API `since` parameter, not a SQL `COUNT(*)` plus `MAX(cursor)` probe), but the two-tier schedule split is the same.

## Design decisions to settle

1. **Where the watermark lives, and its failure behavior.** It must advance only after the Avro/destination write succeeds, and a failed or partial run must not skip a day. Subtract a safety day (`since = last_success_date - 1`) so a run that straddles midnight or a vendor clock skew cannot drop records. Also decide the cold-start path: no watermark means a full pull.
2. **Storage shape — the main decision.** Today the source is one Avro object that BigQuery globs (`src/dbt/finalsite/models/sources-external.yml`, `metadata_cache_mode: MANUAL`, `max_staleness: INTERVAL 7 DAY`). Two options:
   - Accumulate Hive-partitioned Avro (partition by pull date), then pick latest-per-`id` in staging. Keeps the existing Avro factory and its `check_avro_schema_valid` coverage; adds a dedupe layer, a growing object count, and inherits the external-metadata refresh race in #4151.
   - Move `contacts` to dlt with a BigQuery merge destination keyed on `id`. Cleaner semantics and no dedupe layer, but abandons the Avro asset factory and its schema-validation check, and changes the source definition for all four districts.

   Option 2 is a materially larger change than #4446 / #4447 were. Pick deliberately.
3. **Deletes and scope exits.** A `since` pull cannot see a hard-deleted contact or one that dropped out of scope; the full snapshot handles both implicitly. The nightly full refresh is what covers this — set its cadence explicitly rather than inheriting the current one.
4. **Downstream singleton invariants.** `int_finalsite__contacts__households`, `int_finalsite__student_address_of_record`, and the `is_primary` assertions (#4616, #4617, #4680) assume one row per contact from a clean snapshot. Any accumulate-plus-dedupe shape must guarantee that before those models see it, or the address-of-record picks break silently.
5. **Confirm `since` semantics against the live API before building on them.** Specifically: whether it is inclusive, whether it accepts a timestamp despite the date-typed spec, and whether `since_includes_expanded=true` genuinely catches a relationship-only edit. A one-off probe against one district settles all three; the spec is vendored and may lag the deployed API.
6. **Scope is the API `contacts` asset only.** `status_report` arrives by SFTP and is already school-year partitioned — leave it alone.

## Acceptance

- A normal tick requests only contacts changed since the stored watermark and issues far fewer requests than a full pull.
- A relationship-only change (new guardian, flipped `primary`) is picked up, proving `since_includes_expanded`.
- A changed or new contact lands with the same column values it has today.
- A deleted or scope-exited contact stops appearing within one full-refresh cycle.
- A failed run does not advance the watermark, and the next run recovers the missed window.
- Asset keys and dbt source names are unchanged; `stg_finalsite__contacts` and everything downstream builds with no model edits, and one row per `finalsite_enrollment_id` still holds.
- kippnewark's `MAX_RUNTIME_SECONDS_TAG` override can be dropped or materially reduced.

## Related

- #4453 — PowerSchool probe sensor plus full-refresh nightly (the pattern)
- #4446, #4447 — same adoption for Illuminate and Focus
- #4408 — the shared-IP `403` and the concurrency pool that forces serialization
- #4494 — bounded retry in `_request` (same file)
- #4286 — Finalsite enrollment is current-state only; a partitioned accumulate shape interacts with that
- #4151 — external-source metadata-cache race, relevant to option 2a

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make the Finalsite `contacts` pull
incremental and move it ahead of the consumers that read it.

The asset becomes daily-partitioned and derives the vendor API's `since`
parameter from its own partition key (`partition_date - 1`), so the partition key
IS the watermark — no stored cursor, and a failed run writes no partition and
advances nothing. Each pull writes its own Hive partition instead of replacing
one GCS object, `stg_finalsite__contacts` dedupes to the newest partition per
contact `id`, and `stg_finalsite__contact_relationships` reads that deduped model
instead of the raw source. The schedule moves from 04:00 (kippmiami: 04:00 +
12:00) to **00:15 + 12:00** for all four districts, and
`MAX_RUNTIME_SECONDS_TAG` drops from 3600 to 900.

Two things came out of investigating this that changed the shape of the work:

**1. The freshness problem was partly a scheduling bug.** Three of the four NJ
consumers of `stg_finalsite__contacts` were running *before* the 04:00 pull that
fed them, so they shipped day-old contacts:

| NJ destination                      | model                                  | runs at (ET) |
| ----------------------------------- | -------------------------------------- | ------------ |
| Google Directory user create/update | `rpt_google_directory__users_import`   | 01:00        |
| DeansList family contacts           | `rpt_deanslist__family_contacts`       | 01:25        |
| ParentSquare emergency contacts     | `rpt_parentsquare__emergency_contacts` | 18:00        |
| FRESH Dashboard (Tableau)            | `rpt_tableau__fresh_*`                 | 05:00        |

A student entered in Finalsite Monday morning was pulled Tuesday 04:00 and got a
Google account Wednesday 01:00 — about 45 hours, of which ~21 were pure
mis-ordering. Moving the pull earlier was unsafe at full-snapshot cost (four
districts serialized through the limit-1 `finalsite_api` pool is up to ~46 min,
so a 00:30 start would have missed the 01:00 sync). The incremental change is
what makes the correct ordering fit.

**2. `since` is far more over-broad than the issue assumed.** Probed live against
kippmiami and kippnewark: it is server-enforced and date-grained (a timestamp is
accepted but truncated to the day; a malformed value 400s), `count` is hard-capped
at 25, and `since_includes_expanded` is a measured no-op so it is not adopted.
Volume, measured for `since=today`:

|                                   | kippmiami  | kippnewark   |
| --------------------------------- | ---------- | ------------ |
| total contacts                    | 7,688      | 25,052       |
| returned by `since=today`         | 990 (13%)  | 2,013 (8%)   |
| pages: incremental vs full        | 40 vs ~308 | 81 vs ~1,002 |
| identical to landed snapshot      | 984 (99%)  | 1,889 (94%)  |
| genuinely changed                 | 6          | 122          |

So the request win is roughly **8-12x**, not the order of magnitude #4715
implies. Still enough to make the intraday cadence affordable and to retire
kippnewark's runtime override, but worth having on record.

Design and rollout documents are in this PR:

- `docs/superpowers/specs/2026-08-10-finalsite-incremental-contacts-design.md`
- `docs/superpowers/plans/2026-08-10-finalsite-incremental-contacts.md`
- `docs/superpowers/plans/2026-08-10-finalsite-incremental-contacts-cutover.md`

### Deliberate limitations

- **No full refresh, so hard deletes leak.** Latest-per-`id` dedupe never forgets
  a contact whose record vanishes from Finalsite. Scope exits (withdrawn,
  graduated) are unaffected — they are status updates. Observed rate is a few a
  day. Accepted; revisit if it becomes a problem.
- **A future `CONTACTS_SCHEMA` field addition needs the partition history
  re-encoded** (`scripts/reencode_avro_partitions.py`), because staging now scans
  full partition history and a mixed-schema scan drops the new column.
- **A day where BOTH ticks fail does not self-heal** — partition `D+1` asks for
  `since = D` and never re-requests the `D-1` window. Nothing alerts on a missing
  partition today; the cutover runbook adds an ongoing watch step.

### dbt Cloud CI is red — triaged, and NOT caused by this PR

Corrected after triaging the actual run (70403220458583). An earlier version of
this section predicted CI would fail on `stg_finalsite__contacts` with
`Unrecognized name: _dagster_partition_date`. **That prediction was wrong.**

Two tests fail, both pre-existing and data-driven:

| Failing test                                                | Evidence it is pre-existing                                                                                                                                                                              |
| ----------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `unique_fct_student_attendance_streaks_student_attendance_streak_key` | The PROD table has the same 5 duplicate keys today. Attendance streaks are unrelated to Finalsite.                                                                                               |
| `dbt_utils_unique_combination_of_columns_rpt_parentsquare__parents_student_id__first_name__last_name` | Prod's own `int_students__contacts` has the same 3 duplicate groups today. Prod's `rpt_parentsquare__parents` reads 0 only because that materialization predates the drift — it will fail on its next prod rebuild. |

That second one is worth someone's attention on its own merits, independent of
this PR: it is a latent failure sitting in prod data.

**This CI job also does not build the `finalsite` package models at all.** The
PR's CI datasets contain no `*_finalsite` schema, so `stg_finalsite__contacts`
and `stg_finalsite__contact_relationships` were never built here. dbt Cloud CI
therefore does not validate this PR's dbt changes in either direction — a green
run would not have proven them correct.

What remains true, and is the real constraint on merging:

1. Production's `finalsite.contacts` external table has zero
   `_dagster_partition_*` columns until the cutover re-stages it (verified via
   `INFORMATION_SCHEMA`: `status_report` has 1, `contacts` has 0). After merge,
   PROD's `stg_finalsite__contacts` fails to build with
   `Unrecognized name: _dagster_partition_date`, blocking everything downstream
   until cutover step 6.
1. **Schedule the merge and the cutover together** — do not merge and walk away.
1. The dbt half's real validation is cutover steps 6 and 7, not CI.

Rollback: revert this PR. Nothing is destroyed by the code alone; the only
irreversible step is the legacy-object deletion in cutover step 5, which happens
after merge and is gated on verifying the seed landed.

## AI Assistance

Claude Code authored the implementation, the spec, the plan and the runbook, and
ran the live-API probe. Human-directed: the storage-shape decision (option 2a,
Hive-partitioned Avro accumulate rather than dlt merge), keeping Avro at all, the
no-full-refresh call, putting the dedupe in `stg_finalsite__contacts` and
carrying the `relationships` array to satisfy it, the 00:15 + 12:00 cadence with
04:00 dropped, shipping as one PR with red CI accepted, and keeping the one-day
safety window with a docs-only fix for the both-ticks-fail gap.

Executed via `superpowers:subagent-driven-development`: a fresh implementer per
task with a spec-plus-quality review between each, then a whole-branch review on
Opus and one scoped re-review of its fix wave. That final review caught a
Critical defect — `DailyPartitionsDefinition` defaults to `end_offset=0`, so
today's partition did not exist and every tick in all four districts would have
raised `DagsterUnknownPartitionError`, writing no partition at all. Fixed, and
`tests/libraries/test_finalsite_contacts_partitions.py` now reads the four real
per-district assets so a regression fails a test (verified: removing
`end_offset` from kippnewark fails with `assert 0 == 1` for that district alone).

### General

### Dagster

- [ ] `uv run dagster definitions validate` — **not run.** It needs the dbt
      manifest and full credentials; kippmiami's `definitions` additionally fails
      to load in a codespace on the unrelated Focus dlt credential spec. Verified
      instead: each of the four `finalsite.assets` submodules imports, each
      `contacts` asset resolves to a `DailyPartitionsDefinition` with
      `end_offset=1`, and each asset key is unchanged at
      `{code_location}/finalsite/contacts`. Please run this in a full
      environment before merging.
- [ ] `uv run pytest tests/test_dagster_definitions.py` — **not run**, same
      reason.
- [x] Follows the Library + Config pattern: the new schedule factory lives in
      `src/teamster/libraries/finalsite/api/schedules.py` and each code location
      calls it. Asset keys and the Avro IO manager are unchanged.

### dbt

- [x] No new models, so no new properties file is required. The two changed
      staging models keep their existing properties YAML, and the new
      `relationships` column is declared in the enforced contract.
- [x] No new exposure — no new dashboard, analysis or application consumer.
- [x] **Not a breaking change.** `stg_finalsite__contacts` gains a column
      (`relationships`); nothing is renamed, removed or retyped. Its output
      column set is otherwise identical, and
      `stg_finalsite__contact_relationships` keeps its exact column set, so
      downstream `int_` models and every feed need no edits.
- [ ] `stage_external_sources --target staging` — **not run, deliberately.** The
      GCS data is not Hive-partitioned until the cutover seeds it, so staging the
      source now would point the CI table at a prefix holding only the legacy
      root object. This is the reason for the red-CI callout above; the re-stage
      happens in cutover step 6.

### Docs

- [ ] `uv run scripts/gen-automations-doc.py` — **not run, and must not be run
      from a codespace.** The script silently skips code locations that fail to
      import, which would drop them from the catalog. `docs/reference/automations.md`
      is stale for the four contacts schedules (still shows 04:00) and needs
      regenerating in a full environment.
- [x] Updated the affected `CLAUDE.md` files: `libraries/finalsite/` (the asset
      is now partitioned, schedules fire at 00:15 and 12:00, incremental vs full
      pull timings) and all four code locations' integration tables. kippmiami's
      midday-Focus-chain constraints were rewritten to separate the dlt Focus
      04:00 snapshot, which stays, from the contacts pull, which moved.

## CI checks

- [ ] **Trunk** — `trunk check --force` was clean on every changed file locally,
      and the pre-push hook passed.
- [x] **dbt Cloud** — FAILS on two pre-existing, data-driven test failures
      unrelated to this PR, triaged with evidence in the section above. CI does
      not build the finalsite package models, so it neither validates nor
      invalidates this PR's dbt changes.
- [x] **Dagster Cloud** — kippnewark and kippcamden deployed green. kippmiami's
      first attempt failed with
      `DagsterUserCodeUnreachableError: ... Connection refused` on the code
      server — a transient GKE pod-startup failure, not a definitions error (no
      import traceback, and the identical change deployed fine on the other
      locations). Re-run as attempt 2.

## Verification performed

```text
16/16 pytest   tests/libraries/test_finalsite_contacts_{partitions,schedule,since}.py
 2/2  dbt unit stg_finalsite__contacts (dedupe newest-partition-wins, phone normalization)
 1/1  dbt unit stg_finalsite__contact_relationships (no partition fan-out)
      trunk check --force clean on all changed files
```

Not yet verified, and only verifiable at cutover: that the external table reads
the partitioned Avro at runtime, and that the real downstream consumers
(`int_finalsite__student_contacts`, `int_finalsite__student_address_of_record`,
and the two `stg_finalsite__contact_relationships` singular tests) build against
it. Cutover steps 6 and 7 cover both, and step 7 gates resuming the schedules on
them passing.

Refs #4715

### DIFF
--- a/docs/superpowers/plans/2026-08-10-finalsite-incremental-contacts-cutover.md
+++ b/docs/superpowers/plans/2026-08-10-finalsite-incremental-contacts-cutover.md
@@ -1,0 +1,148 @@
+# Finalsite incremental contacts: cutover runbook
+
+Refs #4715. Run these in order. Steps marked **manual** need a human — they
+touch production GCS or the Dagster+ UI.
+
+The external glob `contacts/*` matches recursively, so while both the legacy
+root object and partition files exist the table unions them and every `id`
+appears twice at the storage layer. That window is steps 3 to 5.
+
+## Before you merge
+
+dbt Cloud CI will be **red** on this PR, by design:
+
+- CI runs `dbt build` only — it never runs `stage_external_sources` — so it
+  reads the pre-cutover production `finalsite.contacts` external table, which
+  has no `_dagster_partition_*` columns. `stg_finalsite__contacts` therefore
+  fails in CI and cannot pass until the cutover in step 6 runs.
+- Merging requires overriding that required check.
+- Every prod dbt model downstream of contacts is broken from merge until step 6
+  completes. Schedule the merge and the cutover together — do not merge and walk
+  away.
+
+## 1. Record the pre-cutover baseline (manual)
+
+```sql
+select count(*) as contacts, count(distinct id) as ids
+from `teamster-332318.kippnewark_finalsite.contacts`
+```
+
+Repeat for `kippcamden`, `kippmiami`, `kipppaterson`. Expected around 25,052
+(kippnewark) and 7,688 (kippmiami). Keep these numbers for step 6.
+
+## 2. Pause the schedules (manual, Dagster+ UI)
+
+Stop all four `<location>__finalsite__contacts__daily_asset_job_schedule`
+schedules. Do not rename them.
+
+## 3. Merge the PR and let the deploy land
+
+Confirm every `dagster-cloud-deploy / deploy` check-run reaches a terminal
+conclusion — a shared-library change redeploys every consuming location, so
+expect one same-named check per location.
+
+## 4. Seed one full partition per district (manual, Dagster+ UI)
+
+For each location, materialize `<location>/finalsite/contacts` for partition
+`2026-08-11` with run config:
+
+```yaml
+ops:
+  <location>__finalsite__contacts:
+    config:
+      full_pull: true
+```
+
+Override the run's `dagster/max_runtime` tag to `3600` for these four runs only
+— a full pull is ~20 minutes for kippnewark, well past the new 900s default.
+They serialize through the `finalsite_api` pool, so budget ~35 minutes total.
+
+Verify each run's `record_count` metadata matches step 1's baseline, and that
+its `since` metadata reads `FULL PULL`.
+
+## 5. Delete the legacy root object (manual, destructive)
+
+One per bucket. Check before deleting:
+
+```bash
+gsutil ls -l gs://teamster-kippnewark/dagster/kippnewark/finalsite/contacts/data
+gsutil rm gs://teamster-kippnewark/dagster/kippnewark/finalsite/contacts/data
+```
+
+Repeat for `teamster-kippcamden`, `teamster-kippmiami`, `teamster-kipppaterson`.
+
+Confirm only partition directories remain:
+
+```bash
+gsutil ls gs://teamster-kippnewark/dagster/kippnewark/finalsite/contacts/
+```
+
+Expected: only `_dagster_partition_fiscal_year=.../` entries, no bare `data`.
+
+## 6. Re-stage the external source and rebuild staging
+
+```bash
+uv run dbt run-operation stage_external_sources \
+  --project-dir src/dbt/kippnewark --args "select: finalsite.contacts"
+uv run dbt build --project-dir src/dbt/kippnewark --select stg_finalsite__contacts
+```
+
+Repeat the `stage_external_sources` call for `kippcamden`, `kippmiami`, and
+`kipppaterson` before their `dbt build`.
+
+From the moment this PR merges (step 3) until this step finishes, expect
+`stg_finalsite__contacts` to fail to **build**, not to fail a test. The
+production `finalsite.contacts` external table currently has zero
+`_dagster_partition_*` columns (confirmed via `INFORMATION_SCHEMA`:
+`status_report` has one, `contacts` has none), and `stg_finalsite__contacts` now
+orders by `_dagster_partition_date`. Until `stage_external_sources` recreates
+the table over the Hive-partitioned files, any query against
+`stg_finalsite__contacts` — and everything downstream of it — errors with
+`Unrecognized name: _dagster_partition_date`. The window to minimize is merge →
+re-stage, not seed → delete; that's why "Before you merge" above says to
+schedule the merge and this step together.
+
+Then confirm the grain and the count, with `--nouse_cache` so the results cache
+cannot mask a stale read:
+
+```sql
+select count(*) as rows, count(distinct finalsite_enrollment_id) as ids
+from `teamster-332318.kippnewark_finalsite.stg_finalsite__contacts`
+```
+
+`rows` must equal `ids`, and both must match step 1's baseline within a day of
+churn. The `unique` test on `finalsite_enrollment_id` is the automated form of
+this check.
+
+## 7. Verify downstream consumers
+
+`stg_finalsite__contacts` dedupes to one row per contact, but nothing that
+assumes that grain could be build-verified before now — the partition column
+didn't exist until step 6. Build and test the real consumers:
+
+```bash
+uv run dbt build --project-dir src/dbt/kippnewark --select stg_finalsite__contacts+
+```
+
+Repeat with `--project-dir src/dbt/<location>` for `kippcamden`, `kippmiami`,
+and `kipppaterson`. A pass proves the deduped one-row-per-contact grain holds
+through every model built on it, specifically:
+
+- `int_finalsite__student_contacts`
+- `int_finalsite__student_address_of_record`
+- the singular tests `stg_finalsite__contact_relationships__caregiver_is_adult`
+  and `stg_finalsite__contact_relationships__single_primary`
+
+If any of these fail, the cutover has not actually restored one row per contact
+— do not resume schedules until they pass.
+
+## 8. Resume the schedules (manual, Dagster+ UI)
+
+Start all four schedules. The next tick is 00:15 or 12:00 ET.
+
+## 9. Confirm the first incremental tick
+
+After the first 00:15 run, check its metadata: `since` should read the previous
+day's date, and `record_count` should be in the low thousands, not the tens of
+thousands. Then confirm the 01:00 Google Directory sync and 01:25 DeansList ship
+read same-day contacts.
--- a/docs/superpowers/plans/2026-08-10-finalsite-incremental-contacts-cutover.md
+++ b/docs/superpowers/plans/2026-08-10-finalsite-incremental-contacts-cutover.md
@@ -77,14 +77,27 @@ would then ship empty files to DeansList, ParentSquare and Focus instead of
 failing loudly. That is the reason for the step 4 → step 5 order; do not reorder
 these under time pressure.
 
-One per bucket. Check before deleting:
+One per bucket, and every delete is preceded by its own inspection — the paths
+are written out rather than left as "repeat" because this is an irreversible
+delete across four production buckets:
 
 ```bash
 gsutil ls -l gs://teamster-kippnewark/dagster/kippnewark/finalsite/contacts/data
-gsutil rm gs://teamster-kippnewark/dagster/kippnewark/finalsite/contacts/data
+gsutil rm    gs://teamster-kippnewark/dagster/kippnewark/finalsite/contacts/data
+
+gsutil ls -l gs://teamster-kippcamden/dagster/kippcamden/finalsite/contacts/data
+gsutil rm    gs://teamster-kippcamden/dagster/kippcamden/finalsite/contacts/data
+
+gsutil ls -l gs://teamster-kippmiami/dagster/kippmiami/finalsite/contacts/data
+gsutil rm    gs://teamster-kippmiami/dagster/kippmiami/finalsite/contacts/data
+
+gsutil ls -l gs://teamster-kipppaterson/dagster/kipppaterson/finalsite/contacts/data
+gsutil rm    gs://teamster-kipppaterson/dagster/kipppaterson/finalsite/contacts/data
 ```
 
-Repeat for `teamster-kippcamden`, `teamster-kippmiami`, `teamster-kipppaterson`.
+If an `ls` reports the object missing, STOP rather than moving on — either the
+delete already ran or the path is wrong, and the next `rm` would be operating
+blind.
 
 Confirm only partition directories remain:
 
@@ -161,3 +174,15 @@ After the first 00:15 run, check its metadata: `since` should read the previous
 day's date, and `record_count` should be in the low thousands, not the tens of
 thousands. Then confirm the 01:00 Google Directory sync and 01:25 DeansList ship
 read same-day contacts.
+
+## Ongoing: watch for missing partitions
+
+`since` is derived from the partition key, so partition `D+1` asks for
+`since = D` and never re-requests changes made on `D-1`. One failed tick is
+harmless — the other tick that day covers the same window. **A day where both
+ticks fail does not self-heal**, and nothing currently alerts on it.
+
+Check the `contacts` asset for gaps in the Dagster UI's partition view. Backfill
+any missing partition before the next tick; a backfilled partition requests its
+own `D-1` window, which closes the gap. The longer a gap sits, the more contacts
+hold stale values with no error raised anywhere.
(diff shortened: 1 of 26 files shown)


cbini:
Verified each finding against the repo before acting. Four applied in
`209aee72b`, one deliberately not — verdicts below, declines included.

### 1. No test coverage for the `full_pull=True` seed path — FIXED

Valid, and the most useful of the five: it guards the manual, once-per-district
step 4 of the cutover.

Materializing `_asset` locally isn't viable — the Avro IO manager uploads to GCS
even with `test=True`, and the codespace service account has no object
permissions on the `teamster-*` buckets. So I took the other option you offered
and extracted `build_contacts_request_params(params, partition_key, full_pull)`.
Four new tests assert:

- `full_pull=True` omits `since` entirely
- an incremental run sends `since = partition_date - 1`
- an unpartitioned asset (no partition key) pulls in full
- the captured `params` dict is never mutated

That last one turns your "Strengths" observation about `{**(params or {})}` into
an actual regression test — `params` is captured once at asset-definition time,
so writing `since` into it would leak the first run's watermark into every later
run.

Scope note: the tests assert the request-params contract, not the
`"FULL PULL"` metadata string, which is `request_params.get("since", "FULL PULL")`
and only observable through a materialization. The load-bearing behavior (no
`since` reaches the API) is covered; the metadata label is not.

### 2. Unnecessary pass-through `source` CTE — FIXED

Valid, and the citation checks out: `src/dbt/CLAUDE.md:1074` states the rule, and
`stg_deanslist__behavior.sql` is the precedent — it passes
`relation=source(...)` straight into `dbt_utils.deduplicate`. Inlined it, which
removed both the CTE and its `# trunk-ignore(sqlfluff/ST03)` line.

The CTE came from the implementation plan rather than from the codebase, which is
why it slipped past — worth noting for anyone reading the plan doc in this PR.

### 3. Column ordering — FIXED, with a correction to the reasoning

The convention is real and documented at `src/dbt/CLAUDE.md:1175` ("SQL column
ordering in SELECT clauses"), and `relationships` was indeed sitting after the
`clean_phone(...)` and `nullif(...)` columns. Moved it up beside `households`,
where the other raw passthrough arrays live.

One correction, since you flagged medium confidence and asked for a local check:
`trunk check --force` on this file is clean **both before and after** the move.
sqlfluff does not flag this case, so this is a convention/readability fix, not a
lint fix — the PR's earlier "clean" claim was accurate.

### 4. `relationships` carries `rel_name`, untagged for PII — NOT ACTIONED HERE

You were right to flag it as informational and defer, and I'm declining it in
this PR for the reason you half-identified: it is consistent with the model's
existing pattern, and tagging only the new column would introduce a real
inconsistency rather than remove one.

Concretely, the model tags its derived scalar PII (`address_1`, `address_2`,
`city`, `zip`) but no raw passthrough array:

| passthrough array                    | carries                          | tagged |
| ------------------------------------ | -------------------------------- | ------ |
| `households`                         | raw address, city, state, zip    | no     |
| `custom_attributes` / `id_attributes` | student ids, free-text values    | no     |
| `relationships` (new here)           | `rel_name`, a guardian's name    | no     |

Tagging `relationships` while `households` keeps carrying untagged addresses
would be the odd outcome. Tagging all four is defensible under this repo's
FERPA direct-identifier list, but it is a PII-scope change to a contracted model
and `CLAUDE.md` says to confirm that scope with a human first — so it should not
ride along on an ingestion refactor. Raised with the PR author for a decision on
its own merits.

### 5. `partitions_def` has no type annotation — FIXED

Valid. Your read of the mixed precedent matches what I found
(`libraries/deanslist/assets.py:30` types it, `libraries/sftp/assets.py` doesn't).
Now `partitions_def: PartitionsDefinition | None = None`, consistent with the
already-typed `params` beside it.

### Verification after the fixes

```text
21/21 pytest  tests/libraries/test_finalsite_contacts_{since,schedule,partitions}.py
 2/2  dbt unit stg_finalsite__contacts
 1/1  dbt unit stg_finalsite__contact_relationships
      trunk check --force clean on all three changed files
```

### On your environment note

Your caveat about being unable to verify data-dependent claims was well placed,
and it is worth recording what the CI triage turned up, since it changes how this
PR's red dbt Cloud check should be read: both failing tests are pre-existing and
data-driven, not caused by this PR. Details are in the PR body. One of them —
`rpt_parentsquare__parents` — is a latent failure sitting in prod data right now
and deserves its own issue.

PR #4822, opened 2026-08-11
perf(dbt): materialize the assessment star by recording its FKs as metadata

### LINKED ISSUE #4821: perf(dbt): materialize the assessment star marts onto a shared 5x-per-day cron
## Problem

BigQuery spend spiked on 2026-08-05 (4.60 TiB) and 2026-08-10 (5.02 TiB) against a 1.8-2.7 TiB baseline for a normal work day. Cost triage traced almost all of it to one root cause: the assessment star marts are `materialized: view`, so every reader recomputes a 21-28 base-table chain. The readers are dbt CI tests, Cube queries, and Cube pre-aggregation partition builds.

## Evidence

All figures from `INFORMATION_SCHEMA.JOBS_BY_PROJECT`, 2026-08-01 through 2026-08-11, on-demand billing at 6.25 USD per TiB.

dbt Cloud spend is 85 percent tests, not model builds:

| node kind | jobs | TiB billed | pct |
| --- | --- | --- | --- |
| test | 66,510 | 7.79 | 84.7 |
| model | 109,176 | 1.34 | 14.6 |

100 percent of dbt Cloud spend ran against the `staging` target, so all of it is PR CI. Prod builds run through Dagster on a separate service account.

Test cost tracks materialization of the test target and nothing else:

| test target | distinct tests | GiB per run | TiB billed |
| --- | --- | --- | --- |
| 1-2 base tables (target is a table) | 247 | 0.016 | 0.25 |
| 3-10 base tables | 468 | 0.085 | 2.98 |
| 11-25 base tables (view chain) | 157 | 0.245 | 2.64 |
| 26 or more base tables (deep view chain) | 39 | 0.486 | 1.92 |

A test against a materialized table costs 0.016 GiB. Against a deep view chain it costs 0.486 GiB, a 30x difference. 96.8 percent of test spend goes to tests whose target is not materialized.

`fct_assessment_scores_enrollment_scoped` expands to 27 base tables and carries 10 tests, so CI recomputes the whole fact 10 times per build to check 10 assertions on one result set.

Cube pays the same tax. Its service account billed 1,450 GiB in the window, and 1,416 GiB of that (98 percent) is assessment-mart queries averaging 21.4 base tables each.

Attributable spend for the window:

| source | TiB | est. USD |
| --- | --- | --- |
| dbt CI tests against view-chain targets | 4.56 | 28.50 |
| Cube queries plus pre-aggregation builds on assessment marts | 1.38 | 8.65 |

## Change: materialize the assessment star

Flip these six to `materialized: table` with `automation_condition.cron_schedule: 0 0,10,13,15,17 * * *`:

- `fct_assessment_scores_enrollment_scoped`
- `fct_assessment_scores_student_scoped`
- `dim_assessment_administrations`
- `bridge_assessment_expectations_enrollment_scoped`
- `bridge_assessment_expectations_student_scoped`
- `bridge_assessment_administration_members`

Plus a seventh: raise `dim_assessments` from its existing nightly cron to the same tick. See "Why dim_assessments moves too" below.

The tick matches `int_assessments__response_rollup`, the heaviest upstream, which was moved off eager for this same reason in #4559. Sharing the tick lets the `any_deps_in_progress` guard serialize the pass so `response_rollup` builds first, per the view-to-table flip rules in `src/dbt/CLAUDE.md`. Eager is not an option here: every other upstream is eager, and eager on these marts would reproduce the 125-plus rebuilds per day that #4559 fixed.

### Consumer cadence check

Only one exposure depends on these marts: `cube_semantic_layer`. No Tableau workbook, Sheets extract, or AppSheet app reads them. Checked against all five files in `models/exposures/`.

Cube's own freshness contract on the fact is already daily. The `proficiency_rollup` pre-aggregation in `src/cube/model/cubes/student_assessments/student_assessment_scores.yml` sets `refresh_key: every: 1 day`. A 5-times-daily table is fresher than the rollup Cube serves on its hot path.

### Why dim_assessments moves too

`dim_assessments` is already a table on a nightly cron from #4559, with a comment reading "nightly instead of eager: Cube-only consumer, daily refresh_key -- see fct_assessment_scores_enrollment_scoped". That cross-reference is dangling: the fact carries no such comment and was never flipped. This change finishes what #4559 started, and updates the comment to point at a rationale that now exists.

Leaving it nightly while the rest of the star runs 5 times a day would create an intraday FK skew. `dim_assessment_administrations` has a `relationships` test to `dim_assessments`, so a new Illuminate assessment could land in administrations at 10:00 and not appear in `dim_assessments` until midnight -- a fact row pointing at a missing dim row, warning until the next night. All seven models share one tick to avoid that.

### Accepted trade-off

Cube queries that miss the rollup read the fact view live today and will become 5-times-daily fresh. Given the rollup is already 1-day fresh and the heaviest upstream is already on this tick, the mart views buy no freshness Cube actually consumes. Decision recorded here rather than left implicit.

## Considered and rejected: deleting the dim_dates FK tests

The first draft of this issue proposed deleting 37 `relationships` tests pointing at `dim_dates`, on the theory that they cannot fail because `dim_dates` is a complete 2000-9999 calendar spine. That theory was wrong, and the correction is worth recording.

`dim_dates` covers every date from 2000-01-01 to 9999-12-31, and BigQuery's DATE ceiling is 9999-12-31, so the entire failure surface of such a test is a non-null value before 2000-01-01. Those do occur: PowerSchool and Illuminate emit sentinel dates.

Auditing all 37 columns in prod:

| category | count |
| --- | --- |
| Structurally cannot fail (a `where` clause excludes the only failing rows) | 1 |
| Zero rows in the failure domain, would fire on an upstream regression | 36 |
| Currently catching a defect | 0 |

The single structurally-vacuous one is `dim_assessment_administrations.administered_date_key`, which holds 2 rows dated before 2000 (min 1982-02-27) and already carries `config.where: administered_date_key >= '2000-01-01'` to skip them. Its `where` clause and comment document a real data quirk, so it stays.

The other 36 are dormant rather than vacuous. The pre-2000 junk they would catch is already handled upstream -- sanitized to NULL by #3809 for calendars, and guarded by an `expression_is_true` on `int_assessments__response_rollup.date_taken`. They are a second line of defense and would fire if that upstream sanitization regressed.

Materialization removes the cost argument for touching them: each drops to roughly 0.016 GiB per run. All 37 stay.

## Out of scope

- Project-wide `data_tests: +severity: warn` means none of these tests can fail CI. Worth a separate conversation about whether that spend buys a gate or a monitoring signal.
- The intermediate views still in the chain (`int_assessments__assessments_canonical`, `int_amplify__all_assessments`, `int_pearson__all_assessments`, `int_fldoe__all_assessments`) will expand once per mart build after this change instead of once per test. Flipping those too is a separate evaluation.
- `fct_assessment_scores_enrollment_scoped.test_date_key` has a minimum of exactly 2000-01-01, which looks like a clamped sentinel rather than a real assessment date. Surfaced by the audit above; not investigated here.

## Verification

- `dbt parse` clean, and the manifest confirms all 7 models resolve to `materialized: table` with the shared cron and `contract.enforced: true`.
- `trunk check --force` on the 7 changed yml files.
- After merge, confirm the 7 relations are tables in BigQuery. A properties-yml-only flip does not fire `code_version_changed` at deploy, so each relation stays a view until its first cron tick.
- Post-merge, re-run the `JOBS_BY_PROJECT` test-cost query and confirm the assessment tests move into the 1-2 base-table bucket.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will materialize the assessment-star marts as tables on a shared 5x-per-day cron, by moving their `foreign_key` constraints into `config.meta.foreign_key` so no FK DDL is rendered. It then applies the same move to **every** remaining mart (146 constraints, 56 files) so the failure class cannot recur anywhere, and disables one bridge model nothing consumes.

**This re-attempts [#4464](https://github.com/TEAMSchools/teamster/issues/4464), which [#4587](https://github.com/TEAMSchools/teamster/issues/4587) reverted.** That is deliberate, and the mechanism that broke it is addressed rather than re-risked. Please read the "Why this is not #4464 again" section before reviewing the rest.

### The cost problem

BigQuery spend hit 5.02 TiB on 2026-08-10 against a 1.8-2.7 TiB baseline. Measured over 2026-08-01 to 2026-08-11:

- **85% of dbt Cloud's BigQuery spend is tests**, not model builds (7.79 TiB vs 1.34 TiB). All of it is PR CI -- 100% ran against the `staging` target.
- Test cost tracks materialization of the test target and nothing else: **0.016 GiB per run against a table vs 0.486 GiB against a deep view chain, a 30x spread.** 96.8% of test spend goes to tests whose target is not materialized.
- `fct_assessment_scores_enrollment_scoped` expands to 27 base tables and carries 10 tests, so CI recomputes the entire fact 10 times per build to check 10 assertions on one result set.
- Cube pays the same tax: 1,416 of its 1,450 GiB (98%) is assessment-mart queries averaging 21.4 base tables each.

Attributable: **28.50 USD** in CI test scans plus **8.65 USD** in Cube, per 11 days, at on-demand 6.25 USD per TiB.

### Why this is not #4464 again

#4587 reverted the materialization after four `Referenced primary key in table ... has been updated` failures in 30 days. The recorded explanation blamed concurrent rebuilds under the eager automation condition, which implies a cron would fix it. Checking the actual history and graph shows something sharper:

- **#4464 set only `materialized: table`** across 11 models, with no `automation_condition` at all. Eager was the only configuration ever tried, so the cron hypothesis was never tested.
- **But a cron would not have helped.** No FK edge under `marts/` is also a `ref()` edge -- all 26 closure edges are declared-only, because these models re-hash the parent's surrogate key rather than joining the parent. dbt's topological sort therefore has no edge to order parent before child; they land in the same dependency tier and build concurrently across `threads: 40`. A cron controls when a run starts, not ordering inside it, and `any_deps_in_progress` guards `ref()` deps only.

So the failure class is not "eager" and not "cron" -- it is **rendering FK constraints into CTAS DDL at all**. This PR removes that, which removes the failure class:

- The 17 `foreign_key` constraints on the six marts move to `columns[].config.meta.foreign_key`, same `to:` / `to_columns:` shape. `primary_key` constraints stay.
- A view mart carries **no** constraints in BigQuery, so a table with a PK and no FK is strictly better than the view it replaces on every axis.
- With no FK DDL rendered, the parents can stay views -- no need to convert the 11-model FK closure, which would have pulled in core dims (`dim_student_enrollments`, `dim_locations`, `dim_course_sections`) well outside this PR.

The invariant every existing table mart already satisfies is preserved and now stated explicitly in `marts/CLAUDE.md`: **a table mart carries no outgoing FK constraints.** All 6 pre-existing table marts have zero; this change takes the set from 6 to 12.

### The reference diagram is unaffected

`generate_marts_reference.py` derives edges only from literal `foreign_key` constraints, so moving them would have silently emptied the diagram. It now reads the `meta` form alongside constraints.

Verified rather than assumed: regenerating the page and normalizing away padding leaves **all 80 FK rows identical**. The one remaining line difference is a Mermaid reorder that reproduces on unmodified `main` -- pre-existing drift the regeneration absorbs. A new fixture and test case cover the meta form; all 13 generator tests pass.

### Why the 5x-per-day tick

Matches `int_assessments__response_rollup`, the heaviest upstream, moved off eager for the same reason in [#4559](https://github.com/TEAMSchools/teamster/issues/4559). `dim_assessments` is raised from nightly to the same tick so the whole star shares one cadence -- otherwise a new Illuminate assessment could land in `dim_assessment_administrations` at 10:00 and not appear in `dim_assessments` until midnight.

### Consumer cadence check

`cube_semantic_layer` is the only exposure touching these marts -- no Tableau workbook, Sheets extract, or AppSheet app (checked all five files in `models/exposures/`). It lists 6 of the 7; `bridge_assessment_administration_members` has no exposure anywhere, a pre-existing gap noted in the self-review. Cube's `proficiency_rollup` pre-aggregation declares `refresh_key: every: 1 day`, so a 5x-per-day table is **fresher** than what Cube serves on its hot path.

**Accepted trade-off:** Cube queries that miss the rollup read the fact view live today and become 5x-per-day fresh.

### Every other mart follows the same pattern

`0` `foreign_key` constraints remain anywhere under `marts/`: 146 more move to `config.meta.foreign_key` across 56 files. A view mart's constraints are inert in BigQuery, so this is behavior-neutral today. What it buys is that materializing **any** mart is now safe by default, and a future view-to-table flip no longer drags its FK closure along.

Six columns already carried a `config` block (`meta.source_system` / `source_model` / `source_column`). Those are merged into, not replaced -- a duplicate `config` key would have silently dropped the existing mapping. All six verified to retain their `source_*` keys alongside the new `foreign_key`.

Sizing what this unlocks: 169 tests on other marts' deep view chains cost **16.37 USD per 11 days**, and every one of those models is now a candidate for the same treatment.

### Disabled: bridge_assessment_administration_members

Nothing reads it -- no `ref()` under `src/`, no exposure, no reference under `src/cube/model/`. 180 days of BigQuery history shows only its own build and its own three tests: **171 GiB (~1.04 USD) per 11 days** for a relation with no consumer. `dim_assessments` names it in a comment but does not select from it.

Each test carries its own `enabled: false`, because **dbt does not disable a model's tests when the model is disabled in properties yml** -- verified with `dbt parse --no-partial-parse`, where the model moved to `disabled` while all three tests stayed in `nodes`. Left implicit they would keep scanning the stale prod relation, which is the cost the disable exists to remove.

### A latent defect the refactor surfaced

Touching every mart pulled models into CI's `state:modified+` selection that CI had **never built before**, and one of them failed at `severity: error`: `fct_student_attendance_streaks` had 5 duplicate primary keys. Prod carries the identical 5.

Root cause is source data. Two kippcamden students each hold two active AY2026 enrollment stints sharing `entrydate: 2026-07-01` -- a 2-3 day stint and a full-year one. The fact joins the enrollment union by half-open date range, so both stints match one streak and it fans out. Because `entrydate` is the only enrollment input to `student_enrollment_key`, both matches hash identically and the duplicate rows are byte-identical.

Filed as #4835 (`ops-tracked`, no student identifiers -- those are PII). Interim mitigation here collapses the fan-out with `dbt_utils.deduplicate` and a `TODO(#4835)`. This is not duplicate-masking: both matches yield the same key, so the collapse is information-preserving, and a **genuine** overlap (two stints with different entrydates covering one streak) still fans out and still fails the PK test.

Verified by running the compiled SQL against prod: `duplicated_keys` 0 (was 5), total rows 6,164,285 equal to distinct keys exactly -- removing precisely those 5 rows and nothing else.

The PK test is left at `severity: error`. #3923 deliberately restored it to error after an earlier warn period; downgrading it would quietly reverse that decision.

### Two data-quality fixes riding along

- **Dropped one unfailable test.** `relationships` on `dim_assessment_administrations.administered_date_key` carried `where: administered_date_key >= '2000-01-01'`, which excluded the only 2 rows in the codebase that could fail it. Auditing all 37 `relationships`-to-`dim_dates` tests found 1 structurally vacuous, 36 dormant (zero rows in their failure domain, but they would fire on an upstream regression), 0 currently catching anything. Only the vacuous one is removed; its sentinel-row note moves into the column description.
- **Closed a floor-sentinel gap.** `int_assessments__response_rollup` nulled `date_taken < date '2000-01-01'` -- strictly less-than, so the sentinel that IS the floor passed through, surfacing as a minimum `test_date_key` of exactly `2000-01-01` (41 rows, 6 students). Widened to `<=`, and the paired `expression_is_true` guard tightened to `> '2000-01-01'`, still `severity: warn`. Verified no fan-out: all 32 affected fact rows keep a real `administered_date`, so `assessment_date_key` and the #4546 `not_null` invariant are untouched, and nulling a date can only remove LEFT JOIN matches.

This does **not** address the broader `date_taken` corruption (1,327,443 rows, 5.66%, predating any real administration), which is already documented in `student_assessment_scores.yml` and worked around by anchoring calendar rollups on `assessment_date_key`.

## AI Assistance

Claude Code authored this PR. The cost triage, exposure/cadence analysis, the pre-2000 audit, the FK-vs-ref graph analysis, and all edits were AI-executed; scope, the one-PR decision, the 5x-per-day tick, and the decision to drop FK constraints rather than convert the closure were human-directed. Claude initially proposed deleting 37 tests on a wrong premise and dropped it after the audit contradicted it, and initially missed that #4587 had already reverted this change.

### General

      -- not done, needs a human

### Dagster

- [x] `dagster definitions validate` -- `kipptaf` passes, which also proves the cron
      expression parses (cron typos surface only at Dagster load, per
      `automation-conditions.md`)
- [x] `uv run pytest tests/test_dagster_definitions.py` -- 4 passed, 2 failed.
      Both failures are `kippmiami` on an unresolvable local `FOCUS_DB`
      credential, reproduced identically on `main`; pre-existing and unrelated
- [x] New integrations follow the Library + Config pattern -- N/A

### dbt

- [x] Properties file for all new models -- N/A, no new models
- [x] Exposure for models consumed by an external tool -- N/A, `cube_semantic_layer`
      covers 6 of the 7 and is unchanged. `bridge_assessment_administration_members`
      is absent from it, but Cube references that model nowhere in
      `src/cube/model/`, so adding a `depends_on` edge would declare a dependency
      that does not exist. Pre-existing gap, flagged for follow-up
- [x] **Breaking change?** No column or contract changes. The FK constraints are
      removed from the DDL but preserved as metadata and in the diagram, and the
      `relationships` tests that actually detect orphans are untouched.
      **Rollback:** revert; the models return to views on the next deploy. Note
      the reverse conversion needs a drop -- per `src/dbt/CLAUDE.md`,
      `create or replace view` does not replace an existing table, so a revert
      needs `DROP TABLE IF EXISTS` on the six relations or one
      `dbt build --select <model> --full-refresh` each
- [x] `stage_external_sources` -- N/A

### Docs

- [x] `marts/CLAUDE.md` updated: the FK-carrying prohibition now states the
      FK-vs-`ref()` mechanism, records that cron does not fix it, and documents
      the `config.meta.foreign_key` form as the sanctioned path
- [x] `docs/reference/marts-data-models.md` regenerated
- [x] Automations catalog -- **not needed.** It covers schedules and sensors;
      `gen-automations-doc.py` filters out `__automation_condition_sensor`, and
      `docs/CLAUDE.md` warns it silently drops code locations when run in a
      codespace

## CI checks

- [ ] **Trunk** -- `trunk check --force` clean locally on all 11 changed files
      (2 markdown files needed prettier padding, applied by the pre-commit hook)
- [ ] **dbt Cloud** -- the previous run failed on exactly the FK DDL this commit
      removes; awaiting the re-run
- [ ] **Dagster Cloud** -- awaiting CI

## Post-merge verification

A properties-yml-only flip does not fire `code_version_changed`, so each relation stays a view until its first cron tick. After that tick:

1. Confirm the 5 remaining flipped relations report `type = 1` in `kipptaf_marts.__TABLES__`
1. Drop the now-orphaned `kipptaf_marts.bridge_assessment_administration_members` view (dbt no longer manages it; needs a human, BQ MCP is SELECT-only)
2. Confirm none carries a FOREIGN KEY in `INFORMATION_SCHEMA.TABLE_CONSTRAINTS`, and each carries its PRIMARY KEY
3. Re-run the `JOBS_BY_PROJECT` test-cost query and confirm the assessment tests move into the 1-2 base-table bucket

Refs #4821

### DIFF
--- a/docs/reference/marts-data-models.md
+++ b/docs/reference/marts-data-models.md
@@ -386,8 +386,8 @@ erDiagram
 
 ```mermaid
 erDiagram
-  fct_survey_responses }o--|| fct_survey_submissions : "survey_submission_key"
   fct_survey_responses }o--|| dim_survey_questions : "survey_question_key"
+  fct_survey_responses }o--|| fct_survey_submissions : "survey_submission_key"
   fct_survey_submissions }o--|| dim_survey_administrations : "survey_administration_key"
   fct_survey_submissions }o--|| dim_staff : "staff_key"
   fct_survey_submissions }o--|| dim_student_enrollments : "student_enrollment_key"
--- a/scripts/generate_marts_reference.py
+++ b/scripts/generate_marts_reference.py
@@ -101,10 +101,14 @@ def parse_fk_edges(yaml_path: Path) -> list[FkEdge]:
     model-level edges (in constraint order, then column order within each
     constraint).  Non-foreign-key constraints are ignored in both locations.
 
-    FK edges come only from literal ``foreign_key`` constraints — never inferred
-    from ``relationships`` data tests.  Every mart (views and table-materialized
-    alike) declares its FKs as constraints, so the constraint blocks are the
-    single source of truth for the diagram.
+    Table-materialized marts cannot carry ``foreign_key`` constraints (they
+    render into the CTAS DDL — see #4587), so they declare the same edge under
+    ``columns[].config.meta.foreign_key`` in the same ``to:`` / ``to_columns:``
+    shape.  Those are read too, immediately after each column's constraints.
+
+    FK edges come only from literal ``foreign_key`` constraints and their
+    ``config.meta.foreign_key`` equivalent — never inferred from
+    ``relationships`` data tests.
     """
     doc = yaml.safe_load(yaml_path.read_text(encoding="utf-8")) or {}
     edges: list[FkEdge] = []
@@ -121,6 +125,22 @@ def parse_fk_edges(yaml_path: Path) -> list[FkEdge]:
                     _warn_if_legacy_expression(constraint, source)
                     continue
                 edges.append(FkEdge(source, column["name"], target))
+
+            # Column-level FK metadata, for table-materialized marts.
+            #
+            # A table mart cannot declare a foreign_key CONSTRAINT: it renders
+            # into the CTAS DDL, and no FK edge under marts/ is also a ref()
+            # edge, so nothing serializes the parent build ahead of the child
+            # (#4587). Those marts record the edge under config.meta.foreign_key
+            # instead, in the same to: / to_columns: shape, so it still reaches
+            # the diagram.
+            meta_fk = (column.get("config") or {}).get("meta", {}).get("foreign_key")
+            if isinstance(meta_fk, Mapping):
+                target = _constraint_target(meta_fk)
+                if target is None:
+                    _warn_if_legacy_expression(meta_fk, source)
+                else:
+                    edges.append(FkEdge(source, column["name"], target))
 
         # Model-level FK constraints.
         for constraint in model.get("constraints", []):
--- a/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
+++ b/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
@@ -56,47 +56,49 @@ with
     ),
 
     -- Illuminate `date_taken` is unreliable for a small share of rows, in BOTH
-    -- directions: the raw column spans 0001-01-01 to 4024-12-04. Two guards,
-    -- because neither covers the other's cases.
+    -- directions: the raw column spans 0001-01-01 to 4024-12-04. Every row is
+    -- judged against a date it should sit near, never against a calendar
+    -- constant — an absolute floor is arbitrary, and being one-sided it can
+    -- never reject a future date.
     --
-    -- 1. Relative window. Judge a sitting against its OWN administration, which
-    -- catches a 2001 date attached to a 2020 administration and any absurd
-    -- future date. +/-365 days keeps 99.78% of dated rows; 7,508 fail it.
-    -- An absolute floor alone cannot do this — it is arbitrary and one-sided,
-    -- and can never reject a future date.
+    -- Preferred anchor is the row's own administration (+/-365 days). ~296k
+    -- rows have none, so those fall back to their `academic_year`, which is
+    -- always populated; the +/-1 year window is wide enough to absorb the
+    -- academic-year tagging drift of #3801. Together they null 11,455 rows
+    -- (0.34% of dated rows) and bring the range to 2013-06-03 .. 2026-12-27.
     --
-    -- 2. Absolute floor, for rows with NO administration date to anchor to
-    -- (~296k rows). The window cannot judge them, so without a floor that
-    -- path is unbounded — it is how 1969-12-31 survived an earlier version of
-    -- this guard. Only 1 row needs it today; it is here for the class, not
-    -- the row.
+    -- The fallback is what earns its keep: an earlier version of this guard
+    -- used a 2000-01-01 floor for unanchored rows, which caught exactly 1 row
+    -- and let a ~1,300-row cluster of 2001-01-01 sittings through, because
+    -- they cleared the constant while sitting a decade from their academic
+    -- year.
     --
-    -- Anchored rows inside the window pass through untouched, so a legitimately
-    -- future-dated sitting within its administration window is preserved.
+    -- Rows inside their window pass through untouched, so a legitimately
+    -- future-dated sitting within its administration window is preserved, and
+    -- `assessment_date_key` (`coalesce(administration, date_taken)`) keeps a
+    -- real administration date where one exists — the invariant #4546 rests on.
     --
     -- This matters more than the row counts suggest: the final select takes
     -- `min(date_taken)` per canonical group, so one junk row poisons its whole
     -- group's date. Nulling before the aggregate is what actually fixes it.
-    --
-    -- Unanchored rows that clear the floor are left as-is rather than guessed
-    -- at, which also keeps `assessment_date_key`
-    -- (`coalesce(administration, date_taken)`) non-null downstream — the
-    -- invariant #4546 rests on.
     sanitized_responses as (
         select
             * except (date_taken),
 
-            if(
-                date_taken < date '2000-01-01'
-                or (
+            case
+                when date_taken is null
+                then null
+                when
                     canonical_administered_at is not null
                     and date_diff(date_taken, canonical_administered_at, day)
-                    not between -365
-                    and 365
-                ),
-                null,
-                date_taken
-            ) as date_taken,
+                    between -365 and 365
+                then date_taken
+                when
+                    canonical_administered_at is null
+                    and extract(year from date_taken)
+                    between academic_year - 1 and academic_year + 1
+                then date_taken
+            end as date_taken,
         from scaffold_responses
     ),
 
--- a/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
+++ b/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
@@ -40,7 +40,9 @@ with
                 s.is_internal_assessment, c.grade_level_id, s.grade_level_id
             ) as canonical_grade_level_id,
 
-            if(s.date_taken < date '2000-01-01', null, s.date_taken) as date_taken,
+            -- <= not <: the sentinel IS the floor date (41 rows, 6 students),
+            -- so a strict < left it indistinguishable from a real sitting date.
+            if(s.date_taken <= date '2000-01-01', null, s.date_taken) as date_taken,
         from {{ ref("int_assessments__scaffold") }} as s
         left join
             {{ ref("int_illuminate__agg_student_responses") }} as asr
--- a/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
+++ b/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__response_rollup.sql
@@ -20,6 +20,7 @@ with
             s.is_replacement,
             s.student_assessment_id,
             s.canonical_assessment_id,
+            s.date_taken,
 
             asr.response_type,
             asr.response_type_id,
@@ -40,9 +41,6 @@ with
                 s.is_internal_assessment, c.grade_level_id, s.grade_level_id
             ) as canonical_grade_level_id,
 
-            -- <= not <: the sentinel IS the floor date (41 rows, 6 students),
-            -- so a strict < left it indistinguishable from a real sitting date.
-            if(s.date_taken <= date '2000-01-01', null, s.date_taken) as date_taken,
         from {{ ref("int_assessments__scaffold") }} as s
         left join
             {{ ref("int_illuminate__agg_student_responses") }} as asr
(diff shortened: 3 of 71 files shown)


cbini:
## Review response

Verified every finding against the code before replying. All five were correct. Four are fixed; one is accepted as a documentation error but its implied code fix is declined, with reasoning below.

Head is now `c765972d5`; **dbt Cloud CI passes** as of `bae9382fb`.

---

### 🔴 Critical 1 — reintroduces the #4587 FK-carrying-table pattern — ACCEPTED, fixed in `bae9382fb`

Correct, and I had missed it entirely when opening the PR. The six marts no longer declare any `foreign_key` constraints; each edge moved to `columns[].config.meta.foreign_key`, and `generate_marts_reference.py` reads that form so the diagram keeps all 80 FK rows. Manifest confirms 0 FK constraints and `materialized: table` on all six.

Two things the investigation added on top of your finding:

1. **Your `dim_students` sub-point was right and I had not identified it.** It is `materialized: table` with no `automation_condition`, so it rebuilds eagerly on its own schedule. My "shared tick" reasoning covered `dim_assessments` but not `dim_students` — a real gap in the original design, now moot since no FK DDL references it.
2. **The rule's stated reason is incomplete, and a cron would not have fixed it.** No FK edge under `marts/` is also a `ref()` edge — these models re-hash the parent's surrogate key rather than joining the parent, so all 26 closure edges are invisible to dbt's topological sort. Parent and child land in the same dependency tier and build concurrently at `threads: 40`. A cron controls when a run starts, not ordering within it. For the record, `#4464` set only `materialized: table` with no automation condition at all, so eager was the only configuration ever tried.

### 🔴 Critical 2 — three models FK to still-view relations; should hard-fail — CONFIRMED, it did, fixed in `bae9382fb`

It hard-failed exactly as predicted. The dbt Cloud run on `cd858d891` errored with `Table teamster-332318.zz_stg_kipptaf_marts.dim_student_section_enrollments does not have Primary Key constraints` on the models you named, plus two more that resolved their FK target to the deferred `zz_stg` copy.

Your suggested `dbt compile` pre-check is superseded by that actual failure, but worth recording for the next person: under `--defer` dbt resolves FK constraint refs to the DEFERRED relation even for models rebuilt in the same run, so the PR-schema build being correct does not save it.

The fix removes the need to convert the FK closure at all — with no FK DDL rendered, the parents stay views. The closure would have been 11 models, pulling in `dim_student_enrollments`, `dim_locations`, and `dim_course_sections`.

### 🟡 Medium — missing `warn_unenforced: false` — ACCEPTED, fixed in `c765972d5`

Half of this was superseded (the FK constraints are gone), but the `primary_key` half survived and was real: the passing CI run logged exactly one warning, `The constraint type primary_key is not enforced by bigquery ... Set 'warn_unenforced: false'`.

Fixed on the three models that still carry a PK — `fct_assessment_scores_student_scoped`, `bridge_assessment_expectations_enrollment_scoped`, `bridge_assessment_expectations_student_scoped`. Your count of 4 was based on the pre-fix diff; `bridge_assessment_administration_members` declares no PK, so it is 3. Audited all 12 table marts: every constraint now sets the flag.

### 🟢 Low — `marts/CLAUDE.md` inventory stale — ACCEPTED, fixed in `bae9382fb`

Updated the exception list and the "`dim_assessments` is nightly" line. Went further and corrected the rule itself: it now states the FK-vs-`ref()` mechanism, records explicitly that a cron does not fix it, and documents `config.meta.foreign_key` as the sanctioned path for materializing an FK-carrying mart.

### 🟢 Low — self-review overstates Cube exposure coverage — ACCEPTED as an error, implied fix DECLINED

You are right that the claim was inaccurate; `bridge_assessment_administration_members` is not in `cube_semantic_layer.depends_on`. PR body corrected.

Declining to add it, though. `grep -rn 'bridge_assessment_administration_members' src/cube/model/` returns nothing — Cube does not query this model anywhere. Adding it to `depends_on` would declare a dependency that does not exist, which is worse than the doc gap: exposures are how dbt learns who breaks on a column change, and a false edge makes that signal wrong.

The real observation is that this model has no exposure and no Cube reference at all, which raises whether anything consumes it. That is pre-existing and out of scope here — happy to file it as a follow-up if you want it tracked.

---

Nothing to add on the "What checked out fine" section; those verifications match what I found independently.

Issue #3728, opened 2026-04-23
refactor(kipptaf): key downstream crosswalks off location_clean_name instead of raw ADP name


cbini:
## Problem

Multiple downstream crosswalks are keyed on the raw ADP `home_work_location_name` value rather than the canonical `location_clean_name` from `int_people__location_crosswalk`. When ADP changes a physical location name, every one of these sheets must be updated in lockstep — easy to miss, and the failure mode is silent NULL joins or dropped rows.

## Crosswalks to migrate

Each should key off `int_people__location_crosswalk.location_clean_name` (or be sourced via `int_people__staff_roster` which already carries the enriched clean name).

- [ ] `stg_google_sheets__coupa__address_name_crosswalk` → consumed by [rpt_coupa__users.sql](src/dbt/kipptaf/models/extracts/coupa/rpt_coupa__users.sql)
- [ ] `stg_google_sheets__egencia__traveler_groups` → consumed by [rpt_egencia__users.sql](src/dbt/kipptaf/models/extracts/egencia/rpt_egencia__users.sql)
- [ ] zendesk onboarding_locations → consumed by [rpt_zendesk__users.sql](src/dbt/kipptaf/models/extracts/zendesk/rpt_zendesk__users.sql)
- [ ] `stg_google_sheets__people__campus_crosswalk` (`ccw.name`) → consumed by:
  - [rpt_clever__staff.sql](src/dbt/kipptaf/models/extracts/clever/rpt_clever__staff.sql)
  - [rpt_clever__sections.sql](src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql)
  - [rpt_illuminate__roles.sql](src/dbt/kipptaf/models/extracts/illuminate/rpt_illuminate__roles.sql)
  - [rpt_tableau__ddi_dashboard.sql](src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__ddi_dashboard.sql)
  - [rpt_gsheets__pm_assignment_roster.sql](src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__pm_assignment_roster.sql)
- [ ] Hardcoded string literals (`'Room 11'`, `like '%Room%'`) in [rpt_appsheet__seat_tracker_roster.sql](src/dbt/kipptaf/models/extracts/google/appsheet/rpt_appsheet__seat_tracker_roster.sql) — convert to a proper crosswalk or canonical clean-name matching

## Approach

[...]

PR #4823, opened 2026-08-11
refactor(dbt): repoint location crosswalk consumers to int_people__location_crosswalk

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will narrow the `src_people__location_crosswalk` external to the two columns it is actually for — `name` and `clean_name` — and route every attribute lookup through `int_people__location_crosswalk`.

`int_people__location_crosswalk` already resolves every attribute (region, abbreviation, school ids, dagster code location, etc.) from `stg_google_sheets__people__locations`, joining the crosswalk sheet only for its alias-to-clean-name mapping. The crosswalk sheet's own attribute columns were a parallel, hand-maintained copy of that data with four consumers left.

### Part 1 — repoint the remaining consumers

- `stg_amplify__mclass__sftp__pm_student_summary_aimline` — `region`, `abbreviation`, `powerschool_school_id`, `dagster_code_location`, plus the join key
- `stg_powerschool__schools` — `powerschool_school_id` relationships test
- `int_deanslist__comm_log` — `deanslist_school_id` relationships test
- `int_deanslist__incidents__penalties` — `deanslist_school_id` relationships test

Value-neutral, verified against prod. Across all 251 crosswalk aliases the intermediate's values match the sheet's exactly, with three exceptions where the sheet is blank and the intermediate resolves a value: `KIPP Paterson Prep Middle`, `KIPP Paterson Prep Elementary`, and `Poinciana Campus`. No alias loses a value.

The aimline model's own output columns are unchanged, so no contracted column is renamed or removed and no downstream consumer is affected.

### Part 2 — narrow the source to the v2 named range

Ops added a `src_people__location_crosswalk_v2` named range covering `'Location Crosswalk'!A1:B1193` — the same 251 alias rows, columns A and B only.

- `sources-external.yml` — `sheet_range` moves to `src_people__location_crosswalk_v2`, plus an explicit `columns:` block pinning `Name` and `Clean_Name` (see the autodetect note below). The source `name:` and the Dagster asset key are unchanged, matching the existing `src_google_sheets__people__locations_v3` precedent of versioning only the range.
- `stg_google_sheets__people__location_crosswalk.yml` — drop the 10 attribute columns from the contract, keeping `Name` (with its uniqueness test) and `Clean_Name`.
- `docs/guides/adp-location-renames.md` — Scenario 1 told Ops to copy region, school level, and the rest onto each new alias row. It now says to leave those columns alone and explains where the attributes actually resolve from.

Nothing is deleted from the spreadsheet, so the Seat Tracker AppSheet app — which depends on `stg_google_sheets__people__location_crosswalk` directly, separate from `rpt_appsheet__seat_tracker_roster` — keeps whatever columns it reads off the sheet.

### Why the source needs an explicit `columns:` block

BigQuery's Google Sheets header autodetect only infers column names when the header row differs in type from the data below it. The old A:L range contained int64 columns, so row 1 was recognized as a header and the external came back with real names. The v2 range is all-string, so autodetect could not distinguish the header, and the first re-stage produced `string_field_0` / `string_field_1` — which would have failed the contract. Declaring `columns:` at the source level pins the names. `skip_leading_rows: 1` still drops the header row, so the row set is unchanged.

## Pre-merge state

The staging external has already been re-staged against the v2 range, so dbt Cloud CI should build against the correct 2-column schema:

```sh
uv run dbt run-operation stage_external_sources \
  --args "select: google_sheets.src_google_sheets__people__location_crosswalk" \
  --vars '{ext_full_refresh: true}' \
  --target staging --project-dir src/dbt/kipptaf
```

`zz_stg_kipptaf_google_sheets.src_google_sheets__people__location_crosswalk` now reports exactly `Name` and `Clean_Name`, both STRING.

**After merge**, re-stage the prod external with the same command and `--target prod`. Doing it before merge would break the deployed 12-column contract in prod.

## AI Assistance

Claude Code authored the change. Human-directed: the decision to repoint everything to `int_people__location_crosswalk`, to narrow the source in the same PR, the v2 named range itself, and authorizing the staging re-stage. AI-assisted: locating the consumers, the edits, diagnosing the autodetect failure, and the verification below.

### General

### dbt

- [x] No new models, so no new properties file is needed.
- [x] No exposure change. The `seat_tracker_app` exposure still depends on `stg_google_sheets__people__location_crosswalk`, which still exists with two columns.
- [x] **Breaking change to a contracted model.** `stg_google_sheets__people__location_crosswalk` loses 10 columns. Nothing in dbt reads them after Part 1, and the spreadsheet itself is untouched, so AppSheet is unaffected. Rollback is to revert `sheet_range` to `src_people__location_crosswalk`, drop the `columns:` block, restore the yml columns, and re-stage.
- [x] **The external source is modified**, so `stage_external_sources --target staging` was run with `ext_full_refresh: true`. See above.

Local verification from the feature branch:

- `dbt parse --target prod` — clean on every commit, no ref cycles from staging and intermediate models referencing an intermediate.
- `dbt build --select stg_amplify__mclass__sftp__pm_student_summary_aimline --target dev --defer` — PASS=4 (model plus its 3 tests, including the repointed relationships test).
- `dbt test --select stg_powerschool__schools int_deanslist__comm_log int_deanslist__incidents__penalties --target dev --defer` — the `stg_powerschool__schools` test passes. **The two deanslist tests passed vacuously and should not be read as verification**: the run resolved both child models to stale copies in a personal dev schema instead of falling through to prod. CI reports the real numbers, 8,135 and 552 orphan rows, and they are pre-existing rather than caused by the repoint — holding the child side fixed, the old parent (sheet `Deanslist_School_ID`) and the new parent (`location_deanslist_school_id`) both yield exactly 8,135 and 552, which follows from their value sets being identical. Both tests are warn-severity, so CI stays green. Detail in the review reply comment.
- `dbt build --select stg_google_sheets__people__location_crosswalk+1 --target dev --defer` against a personally-staged v2 external — PASS=25, contract satisfied. The single WARN is the known free-text `adp_location` orphan issue tracked by #3728, and it is not caused by this PR: holding the seats side at prod gives 7 orphans against either the old or the new crosswalk, so swapping the crosswalk changes nothing. The 7-versus-5 difference comes from a stale dev copy of `stg_google_appsheet__seat_tracker__seats`.
- Row-level parity of the v2 range against the current prod table: 251 rows on both sides, zero `Name` values present in one and not the other, zero `Clean_Name` mismatches.
- Both named ranges read directly through the Sheets API: the old resolves to `'Location Crosswalk'!A1:L1193`, the new to `'Location Crosswalk'!A1:B1193`, both 252 rows including the header.
- `trunk check --force` over all 10 changed files, run from inside the worktree — no issues.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes. 25 warn-severity test warnings, all pre-existing; the 3 crosswalk-related ones are analyzed in the review reply comment.

### DIFF
--- a/docs/guides/adp-location-renames.md
+++ b/docs/guides/adp-location-renames.md
@@ -11,9 +11,11 @@ ADP occasionally renames a location in its system without any real-world change
 to the school or office. When that happens:
 
 1. Open the `src_people__location_crosswalk` Google Sheet.
-2. Add a **new alias row**: set `name` to the new ADP value, set `clean_name` to
-   the unchanged existing clean name, and copy all other columns (region, school
-   level, etc.) from the existing row for that location.
+2. Add a **new alias row**: set `name` to the new ADP value and `clean_name` to
+   the unchanged existing clean name. Those are the only two columns on the
+   sheet — region, school level, school ids, and every other attribute resolve
+   from the `src_google_sheets__people__locations_v3` tab in the same
+   spreadsheet, matched on `clean_name`.
 3. No downstream action is needed — `int_people__location_crosswalk` absorbs the
    rename via alias resolution. All models that join on `clean_name` are
    unaffected.
--- a/docs/guides/adp-location-renames.md
+++ b/docs/guides/adp-location-renames.md
@@ -12,8 +12,9 @@ to the school or office. When that happens:
 
 1. Open the `src_people__location_crosswalk` Google Sheet.
 2. Add a **new alias row**: set `name` to the new ADP value and `clean_name` to
-   the unchanged existing clean name. Those are the only two columns on the
-   sheet — region, school level, school ids, and every other attribute resolve
+   the unchanged existing clean name. Leave the remaining columns alone — dbt
+   reads only those two, via the `src_people__location_crosswalk_v2` named
+   range. Region, school level, school ids, and every other attribute resolve
    from the `src_google_sheets__people__locations_v3` tab in the same
    spreadsheet, matched on `clean_name`.
 3. No downstream action is needed — `int_people__location_crosswalk` absorbs the
--- a/src/dbt/kipptaf/models/amplify/mclass/sftp/staging/properties/stg_amplify__mclass__sftp__pm_student_summary_aimline.yml
+++ b/src/dbt/kipptaf/models/amplify/mclass/sftp/staging/properties/stg_amplify__mclass__sftp__pm_student_summary_aimline.yml
@@ -20,8 +20,8 @@ models:
         data_tests:
           - relationships:
               arguments:
-                to: ref("stg_google_sheets__people__location_crosswalk")
-                field: name
+                to: ref("int_people__location_crosswalk")
+                field: location_name
       - name: _dbt_source_relation
         data_type: string
         description: >-
--- a/src/dbt/kipptaf/models/amplify/mclass/sftp/staging/stg_amplify__mclass__sftp__pm_student_summary_aimline.sql
+++ b/src/dbt/kipptaf/models/amplify/mclass/sftp/staging/stg_amplify__mclass__sftp__pm_student_summary_aimline.sql
@@ -26,18 +26,20 @@ with
         select
             t.* except (_dbt_source_relation),
 
-            lc.region,
-            lc.abbreviation as school,
-            lc.powerschool_school_id as schoolid,
+            lc.location_region as region,
+            lc.location_abbreviation as school,
+            lc.location_powerschool_school_id as schoolid,
 
             regexp_replace(
-                t._dbt_source_relation, r'kipp[a-z]+_', lc.dagster_code_location || '_'
+                t._dbt_source_relation,
+                r'kipp[a-z]+_',
+                lc.location_dagster_code_location || '_'
             ) as _dbt_source_relation,
 
         from transformations as t
         left join
-            {{ ref("stg_google_sheets__people__location_crosswalk") }} as lc
-            on t.school_name = lc.name
+            {{ ref("int_people__location_crosswalk") }} as lc
+            on t.school_name = lc.location_name
     )
 
 select
--- a/src/dbt/kipptaf/models/deanslist/api/intermediate/properties/int_deanslist__comm_log.yml
+++ b/src/dbt/kipptaf/models/deanslist/api/intermediate/properties/int_deanslist__comm_log.yml
@@ -7,5 +7,5 @@ models:
         data_tests:
           - relationships:
               arguments:
-                to: ref("stg_google_sheets__people__location_crosswalk")
-                field: deanslist_school_id
+                to: ref("int_people__location_crosswalk")
+                field: location_deanslist_school_id
--- a/src/dbt/kipptaf/models/deanslist/api/intermediate/properties/int_deanslist__incidents__penalties.yml
+++ b/src/dbt/kipptaf/models/deanslist/api/intermediate/properties/int_deanslist__incidents__penalties.yml
@@ -21,5 +21,5 @@ models:
         data_tests:
           - relationships:
               arguments:
-                to: ref("stg_google_sheets__people__location_crosswalk")
-                field: deanslist_school_id
+                to: ref("int_people__location_crosswalk")
+                field: location_deanslist_school_id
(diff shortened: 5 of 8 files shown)


cbini:
Verdicts on both findings, plus a correction to my own PR description and an analysis of the CI warnings.

Note this review ran against `a9ebc5ab9` only — three commits landed after it, covering the source range swap, the contract narrowing, and an explicit `columns:` block. Summary at the end.

## Finding 1 — missing `config: severity: error` on the `school_name` relationships test

**Declined for this PR**, which matches your own recommendation.

Verified the mechanism: `dbt_project.yml:22-23` sets `data_tests: +severity: warn` project-wide, and the `amplify:` block (line 77) sets only `+schema` and the `dds` disable — no severity override. So the test does degrade to a warning, as described.

One correction to the supporting stat: the convention is applied in **42 of 267** staging `properties.yml` files, not ~9. Still a minority at 16%, so the conclusion is unchanged — the rule is documented but not consistently enforced.

Declining because promoting the test to `error` is a CI-gating behavior change beyond this PR's scope, and fixing only the test this diff touches while leaving the sibling `unique` and `not_null` tests in the same file at `warn` would be arbitrary. The test currently returns 0 rows, so the promotion would be safe whenever someone wants to do it as its own change.

## Finding 2 — the data-parity claim could not be re-verified from the job

**Verified.** Your query is the right one; here is what it returns.

Across all 251 crosswalk aliases, comparing the sheet's attribute columns to the intermediate's `location_*` columns: 0 differences on `deanslist_school_id`, and 3 differences each on `region`, `powerschool_school_id`, and `dagster_code_location`, 2 on `abbreviation`. Every one of those differences is the sheet being blank where the intermediate resolves a value — `KIPP Paterson Prep Middle`, `KIPP Paterson Prep Elementary`, and `Poinciana Campus`. No alias loses a value in either direction.

Row parity of the v2 named range against the current prod table was checked separately: 251 rows on both sides, zero `Name` values present in one and not the other, zero `Clean_Name` mismatches.

## Correction to my PR description

My description claimed the three repointed relationships tests passed locally. That was wrong for two of them. The local `dbt test --defer` run resolved `int_deanslist__comm_log` and `int_deanslist__incidents__penalties` to stale copies in my personal dev schema rather than falling through to prod, so both passed vacuously. CI surfaced the real numbers: 8,135 and 552 orphan rows.

Those orphans are pre-existing and are **not** caused by the repoint. Direct comparison, holding the child side fixed at the same relation CI used:

| Test | vs old parent (sheet `Deanslist_School_ID`) | vs new parent (`location_deanslist_school_id`) |
| --- | --- | --- |
| `int_deanslist__comm_log` | 8,135 | 8,135 |
| `int_deanslist__incidents__penalties` | 552 | 552 |

Identical, which follows from the 0-difference parity on `deanslist_school_id` above — the parent's non-null value set is unchanged, so the orphan set cannot move. The same counts hold against the prod child relations, so this is a real standing data condition rather than `zz_stg` staleness, and it predates this PR. Both tests are intermediate-layer and warn-severity, so CI stays green.

The third crosswalk-related warning, `relationships_stg_iready__personalized_instruction_summary_school__name__...` at 490 rows, targets `field: name` — column A, untouched by this PR. It reports 490 against both the `zz_stg` and prod children, so it is likewise pre-existing; it only appears in this run because modifying the crosswalk model pulls its dependent tests into `state:modified+`.

The `seat_tracker__seats` warning at 7 rows is the known free-text `adp_location` issue tracked by #3728.

## What changed after this review

- `sources-external.yml` — `sheet_range` moved to the new `src_people__location_crosswalk_v2` named range, which covers columns A and B only, so nothing is deleted from the spreadsheet.
- An explicit `columns:` block pinning `Name` and `Clean_Name`. BigQuery's Sheets header autodetect only infers names when the header row differs in type from the data; the old A:L range had int64 columns so row 1 was detected as a header, but the all-string A:B range is not, and the first re-stage produced `string_field_0` / `string_field_1`.
- `stg_google_sheets__people__location_crosswalk.yml` — the 10 attribute columns dropped from the contract.
- `docs/guides/adp-location-renames.md` — Scenario 1 no longer tells Ops to copy the attribute columns onto each new alias row.
- The staging external was re-staged with `ext_full_refresh: true`; `zz_stg_kipptaf_google_sheets.src_google_sheets__people__location_crosswalk` now reports exactly `Name` and `Clean_Name`, both STRING.

PR #4850, opened 2026-08-12
fix(dbt): strip grade-type token from focus marking_period_id cast

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will unblock the `kippmiami_dbt_assets` step,
which has failed on every automation-condition tick since Miami posted its first
live report-card grade this morning
([failing run](https://kipptaf.dagster.cloud/prod/runs/3d770372-c72a-4b36-b7f1-aea66598f85e)).

`stg_focus__student_report_card_grades` cast `marking_period_id` straight to
int64. Focus stores that FK as text on this table alone — the only one of the 8
Focus tables carrying the column where it is not INT64 — because it prefixes
live grade postings with a grade-type token. The first such row, `DT7181`, was
created 2026-08-12 16:33 UTC and failed the build with
`Bad int64 value: DT7181`. All 14,786 pre-existing rows are imported course
history with bare-digit ids, which is why the cast held until today. Today is
also the first day of Q1 2026, so more live postings arrive daily.

Fix:

```sql
safe_cast(marking_period_id as int) as marking_period_id,
```

A prefixed value goes null rather than being decoded. The token *could* be
stripped — `7181` is a real marking period (Quarter 1, school 58, syear 2026,
matching the grade row) — but stripping cannot be done safely without knowing
Focus's full token vocabulary: marking period ids run 3 to 4 digits (212 to
7198), so the trailing digit run of some future token could concatenate into a
*valid* id and pass silently. Nulling is the honest failure mode. The consequence
is that a live-posted grade carries no marking period, and the
`marking_period_` columns on `int_focus__report_card_grades` are null for those
rows, until the vocabulary is known and the token decoded deliberately.

`not_null` is dropped from the column, which is now nullable by design;
`relationships` stays and skips nulls.

This also refreshes descriptions the change invalidates. The model description
still asserted that every row was imported course history and that the FKs a live
report card uses were all empty; the `course_history` column description said "Y
on every Miami row today" in both the staging and intermediate properties files.
All stopped being true this morning.

## AI Assistance

Claude ran the diagnosis (Dagster run logs, then BigQuery profiling of the source
column, then a local reproduction), wrote the fix, and updated the descriptions.
Human-directed throughout, including the decision to null rather than decode.

### Verification

Built in a worktree against real prod dlt data, since a district-only dbt PR gets
no branch deployment and dbt Cloud CI builds only `kipptaf` — a local build is
the only pre-merge validation here.

- Before the fix: reproduced `Bad int64 value: DT7181` exactly.
- After: `stg_focus__student_report_card_grades+` builds PASS=9, ERROR=0.
- Result table: 14,787 rows, exactly 1 null `marking_period_id`, and it is the
  live-posted row — 0 nulls among the 14,786 imported-history rows.
- `trunk check --force` on all three changed files: no issues.

### General

- [x] Reviewed the **Claude Code Review** comment. All three findings verified and
      addressed; verdicts posted per finding, including the residual it surfaced,
      which is what drove the switch to `safe_cast`.

### dbt

- [x] Properties files updated alongside the models (no new models).
- [x] **Behavior change, not a contract change** — `marking_period_id` keeps its
      name and `int` type. It becomes nullable for live-posted grades, and its
      `not_null` test is dropped. No exposure or external-source changes; the only
      consumer is `int_focus__report_card_grades`, which has no consumers of its
      own.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes (a trivial no-op: this PR touches only the `focus`
      package and selects zero `kipptaf` models).
- [ ] **Dagster Cloud** — not triggered; `deploy-prod-*.yaml` excludes
      `src/dbt/**` from `pull_request` paths.

### DIFF
--- a/src/dbt/focus/models/intermediate/properties/int_focus__report_card_grades.yml
+++ b/src/dbt/focus/models/intermediate/properties/int_focus__report_card_grades.yml
@@ -72,7 +72,7 @@ models:
       - name: course_history
         description: >-
           Whether the row is imported course history rather than a grade posted
-          in Focus. Y on every Miami row today.
+          in Focus. Y on imported rows, empty on grades posted live in Focus.
       - name: grade_level
         description: >-
           Zero-padded grade level the student was in when the grade was earned
--- a/src/dbt/focus/models/intermediate/properties/int_focus__report_card_grades.yml
+++ b/src/dbt/focus/models/intermediate/properties/int_focus__report_card_grades.yml
@@ -37,7 +37,11 @@ models:
       - name: student_id
         description: Focus student id the grade belongs to.
       - name: marking_period_id
-        description: Focus marking-period id the grade was earned in.
+        description: >-
+          Focus marking-period id the grade was earned in. Null on grades posted
+          live in Focus, which store a token-prefixed id that staging cannot
+          safely decode — the marking_period_ columns below are null on those
+          rows too.
       - name: grad_subject_id
         description: >-
           Focus graduation-subject id the course counts toward, populated only
--- a/src/dbt/focus/models/staging/properties/stg_focus__student_report_card_grades.yml
+++ b/src/dbt/focus/models/staging/properties/stg_focus__student_report_card_grades.yml
@@ -2,15 +2,16 @@ models:
   - name: stg_focus__student_report_card_grades
     description: >-
       Focus student report-card grades — one row per student per course per
-      marking period. Every row Miami carries today is imported course history
+      marking period. Nearly every row is imported course history
       (course_history is Y) spanning school years well before the Focus rollout,
-      so the letter grade, GPA points and credits are denormalized onto the row
-      and the foreign keys a live report card would use — report_card_grade_id,
-      grade_scale_id, course_period_id, course_id — are all empty and are not
-      projected. Add them once Focus starts posting live report-card grades in
-      Miami. Distinct from stg_focus__report_card_grades, which is the grade
-      definition table (the available letter grades and their GPA values), not
-      student grades.
+      so the letter grade, GPA points and credits are denormalized onto the row.
+      Miami began posting live grades in Focus in August 2026; those rows leave
+      course_history empty and do populate the foreign keys a live report card
+      uses — report_card_grade_id, grade_scale_id, course_period_id, course_id —
+      which are still not projected here. Add them once live posting volume
+      warrants it. Distinct from stg_focus__report_card_grades, which is the
+      grade definition table (the available letter grades and their GPA values),
+      not student grades.
     columns:
       - name: id
         description: Primary key — Focus student_report_card_grades id.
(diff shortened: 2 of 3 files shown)


cbini:
Verdicts on the review, all addressed in `fcc85d460`. I verified each claim against the code before acting; all three were accurate.

**Important 1 — stale `course_history` description: accepted, fixed.** Confirmed the contradiction — `stg_focus__student_report_card_grades.yml:118` said "Y on every Miami row today" while the model description four lines up, edited in this PR, says live rows leave it empty. Both now read "Y on imported rows, empty on grades posted live in Focus." I also fixed the duplicated sentence at `int_focus__report_card_grades.yml:75` rather than leaving a follow-up: it is the same rotted claim from the same reality change, and a one-line doc fix is not worth its own PR.

**Minor 1 — the fail-safe claim is an assumption: accepted, and fixed in code rather than just noted.** You are right that `[0-9]+$` grabs the trailing digit run wherever it starts, so my "any new prefix fails loudly" claim was stronger than the pattern earned. Rather than caveat it, the pattern is now anchored:

```sql
cast(regexp_extract(marking_period_id, r'^[A-Z]*([0-9]+)$') as int) as marking_period_id
```

Anything that is not letters-then-digits extracts null and trips the `not_null` test at `severity: error`. Verified against BigQuery: `DT7181` and `7181` both yield `7181`; `Q2-7181` and `dt7181` both yield null.

One residual I should state rather than paper over: marking period ids are 3 to 4 digits (212 to 7198 across 1,327 rows), so a hypothetical future token ending in a digit could still concatenate into a *valid* 4-digit id and pass silently — e.g. a 1-digit-suffixed token plus a 3-digit id. Anchoring does not close that; only knowing Focus's token vocabulary would. Live postings use current-year ids in the 7000s, where any extra leading digit produces a 5-digit value that fails `relationships` loudly, so the exposure is limited to historical-id rows.

**Minor 2 — `TODO:` should lead its own line: accepted, fixed.** Confirmed the package precedent at `int_focus__student_enrollment.sql:90`. The TODO now opens its own comment line.

Re-validated after the change: `stg_focus__student_report_card_grades+` builds PASS=10, ERROR=0 against prod dlt data; 14,787 rows, 0 null `marking_period_id`, 1 row at `7181`. `trunk check --force` on all three files: no issues.

Noting for the record that `claude-review` does not re-run on `synchronize`, so this comment is the audit trail for the follow-up commit rather than a second review pass.

cbini:
Approach changed in `f144a8ba7`, reversing what I argued for in my previous comment. The residual I flagged there — that a future token ending in a digit could concatenate into a valid id and pass silently — was judged not worth carrying, so the extraction is gone entirely:

```sql
safe_cast(marking_period_id as int) as marking_period_id,
```

A prefixed value now goes null instead of being decoded. Nothing about the collision risk changed; the call is that a silent wrong FK is worse than a missing one, and anchoring the pattern narrowed that risk without eliminating it. Marking period ids run 3 to 4 digits (212 to 7198 across 1,327 rows), which is what leaves the door open.

Consequences, stated plainly:

- A live-posted grade carries no marking period. The `marking_period_` columns on `int_focus__report_card_grades` (title, short name, type, start date, end date) are null for those rows.
- `not_null` is dropped from `marking_period_id` — the column is nullable by design now. `relationships` stays, and skips nulls.
- Recovering the marking period for live postings needs Focus's token vocabulary. That is now a `TODO` at the derivation site rather than a guess in the model.

Re-validated: `stg_focus__student_report_card_grades+` builds PASS=9, ERROR=0 against prod dlt data. 14,787 rows, exactly 1 null `marking_period_id`, and it is the live-posted row — 0 nulls among the 14,786 imported-history rows. `trunk check --force` clean on all three files. PR body updated to describe this approach rather than the previous one.

PR #4851, opened 2026-08-12
fix(kipptaf): re-point the DeansList missing-assignments feed onto the gradebook audit lineage

### LINKED ISSUE #4849: fix(kipptaf): re-point the DeansList missing-assignments feed onto the gradebook audit lineage
## Problem

`rpt_deanslist__missing_assignments` still reads `rpt_tableau__gradebook_assignments`, a pre-AY2627 view the gradebook audit revamp never touched (last changed 2023, no exposure, not mentioned in the reference doc). The audit's assignment tracking now lives in `int_powerschool__gradebook_assignments_scores`. Measured against AY2025 prod data, the feed diverges from the audit in these ways:

- **Paterson is structurally excluded.** `int_powerschool__section_grade_config` unions Newark, Camden and Miami only, and Paterson's copy is deliberately disabled because `gradesectionconfig` is empty in Paterson's PowerSchool. 3,267 AY2025 rows never shipped, and this cannot be fixed in place.
- **Miami is still included** (14,307 AY2025 rows) though Miami is off DeansList and its gradebook moved to Focus.
- **No `is_expected` gate**: 10,980 exempt rows and 2,029 not-counted-in-final-grade rows reported as missing.
- **No enrollment-window guard**: 14,020 rows for assignments due before a student enrolled in the section or after they left.
- **No dropped-section filter**: 25,760 rows.
- **Fan-out from the `section_grade_config` join**: 260,945 rows against 251,317 distinct student-assignment pairs, shipped as duplicate records because no assignment id is projected.

It is also an `rpt_` reading an `rpt_`, the layering violation the July 2026 teacher/student split removed from the audit reports.

## Decisions

- Missing means `is_expected_missing` (excludes exempt and not-counted-in-final-grade).
- Exclude Miami.
- Exclude ES. ES does not use PowerSchool for assignments; it gets EOQ comments only, via `rpt_tableau__gradebook_es_comments`.
- Keep single-quarter-term sections. The audit's `section_quarter_count` minimum scopes which gradebooks are audited against per-quarter assignment-count expectations; it has no bearing on whether a student turned work in. 3,624 AY2025 rows.
- Project `assignmentsectionid` as a seventh feed column so the uniqueness test has a key. Approved by the DeansList owner.
- Disable `rpt_tableau__gradebook_assignments` rather than delete it, per the repo rule that retiring a model is always a disable. The SQL and the prod view stay in place, so any unknown consumer degrades to frozen data rather than vanishing. It is safe to retire: no exposure, no Tableau extract config, and no remaining references once the feed is re-pointed.

## Scope

1. `int_powerschool__gradebook_assignments_scores` — project `student_number`, `course_name` and `teacher_name` from the `base_powerschool__course_enrollments` join it already carries. Both live consumers project explicit columns, so adding columns is safe, and the join was verified 1:1.
2. `rpt_deanslist__missing_assignments` — read `_scores`; filter on academic year, `is_expected_missing`, not Miami, not ES. Project `assignmentsectionid` alongside the existing six columns and add a uniqueness test on `student_number` plus `assignmentsectionid`, which is exactly unique on AY2025 data.
3. `rpt_tableau__gradebook_assignments` — set `enabled: false` in its properties yml.

Expected AY2025 shape: 209,311 rows. Newark 139,704, Camden 66,340, Paterson 3,267.

## Design

Spec: `docs/superpowers/specs/2026-08-12-deanslist-missing-assignments-lineage-design.md`

## Follow-up

The published reference doc states the audit's single-quarter section filter has no recoverable business rationale; the data team says there is one. Worth correcting that passage separately.

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will re-point the nightly DeansList
missing-assignments feed off `rpt_tableau__gradebook_assignments` — a 2023-era
view the AY2026-2027 gradebook audit revamp never touched — and onto the audit's
own assignment model, `int_powerschool__gradebook_assignments_scores`. DeansList
and the gradebook audit dashboard then agree on what a missing assignment is.

Closes #4849. Design spec and implementation plan are included in the diff under
`docs/superpowers/`.

Measured against AY2025 prod data, the old feed diverged from the audit six ways:
Paterson was structurally excluded (3,267 rows never shipped, and unfixable in
place because `gradesectionconfig` is empty in Paterson's PowerSchool); Miami was
still included (14,307 rows) though it is off DeansList; exempt (10,980) and
not-counted-in-final-grade (2,029) assignments were reported missing; 14,020 rows
were assignments due outside the student's enrollment window; 25,760 were
dropped-section enrollments; and 9,628 rows shipped as duplicate records from a
join fan-out.

Three changes:

1. `int_powerschool__gradebook_assignments_scores` projects `student_number`,
   `course_name` and `teacher_name` from the `base_powerschool__course_enrollments`
   join it already carried. Verified 1:1 — prod and dev row counts for AY2025 are
   both 3,975,970.
1. `rpt_deanslist__missing_assignments` reads that model, filtering on
   `is_expected_missing`, excluding Miami and ES, and projecting
   `assignmentsectionid` so a uniqueness test has a key. AY2025 output is 209,311
   rows: Camden 66,340, Newark 139,704, Paterson 3,267, zero Miami, zero null
   labels.
1. `rpt_tableau__gradebook_assignments` and the now-stranded kipptaf
   `int_powerschool__section_grade_config` wrapper are disabled, not deleted. Their
   prod relations are left standing; no `drop view` is issued.

### Behavior change worth reviewer attention

`grade_category` changes values, not just provenance. In the old feed it was the
single constant string `Total_Points` on all 246,810 AY2025 rows — a
grading-formula type rather than a category, because the legacy view's
Total_Points branch joined assignments with no category predicate. It now carries
real assignment categories: Homework 83,544, Work Habits 71,471, Formative
Mastery 48,885, Summative Mastery 5,403, Classwork 8. The column becomes
meaningful for the first time, but anyone reading or filtering that field on the
DeansList side will see different values. The DeansList owner approved the added
column and is being told about the label change separately.

The feed also grows from six columns to seven (`assignmentsectionid`), approved by
the DeansList owner. Column names are otherwise unchanged, so the Dagster extract
asset is unaffected.

### Rollback plan

Revert the merge commit. The two disabled models keep their SQL and their prod
relations, so re-enabling is a properties-yml change with no rebuild required.

## AI Assistance

Claude Code did the investigation, the spec, the plan and the implementation,
executed task-by-task with a review after each task and a whole-branch review at
the end. Human-directed throughout: the scope decisions are all the requester's —
the definition of missing (`is_expected_missing`), excluding Miami and ES,
deliberately NOT inheriting the audit's single-quarter-section exclusion,
disabling rather than deleting retired models, downgrading the new uniqueness test
to `warn`, and disabling the stranded config wrapper in this branch rather than a
follow-up.

### General

### dbt

      models; all four touched models have updated properties files, and every
      projected column on the contracted feed carries a `data_type`.
- [ ] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — not done.
      No DeansList extract has an exposure today; the outbound feeds are driven by
      Dagster config at
      `src/teamster/code_locations/kipptaf/extracts/config/deanslist-annual.yaml`.
      Pre-existing gap for the whole `extracts/deanslist/` directory rather than
      something this change introduces.
- [x] **Breaking change?** No contracted column was renamed or removed. One column
      was added to `rpt_deanslist__missing_assignments`, and `grade_category`
      changes values as described above. Rollback plan is in the summary.
- [ ] If adding or modifying an external source, run `stage_external_sources` with
      **`--target staging`** — not applicable, no external sources touched.

### Local verification

- Full chain built with `--target dev --defer`: 5 models PASS, 8 tests PASS, 0
  warn, 0 error.
- `dbt parse --no-partial-parse` clean after both model retirements, so nothing
  references them.
- `trunk check --force` clean on all five changed non-doc files.
- Feed validated by overriding the year var to 2025, since the current academic
  year holds no assignments yet and an empty build proves nothing.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

## Post-merge note

The feed will ship zero rows until teachers begin entering AY2026 assignments;
the extract runs nightly at 1:25am Eastern regardless. Nobody has confirmed how
DeansList treats an empty file. That behavior is unchanged by this PR, but worth
knowing before the first post-merge run.

---
  - https://app.asana.com/0/0/1217444272233461


GabyRangelB:
## CI note: one new dbt Cloud warning, expected

dbt Cloud passed. Pulling its warnings with `warning_only=true` surfaces exactly one, and it is caused by this PR:

```text
[WARNING]: Breaking change to contracted, unversioned model rpt_tableau__gradebook_assignments
While comparing to previous project state, dbt detected a breaking change to an unversioned model.
  - Contracted model 'model.kipptaf.rpt_tableau__gradebook_assignments' was disabled.
```

This is the expected consequence of retiring that view, not a problem to fix. dbt flags disabling any contracted model as breaking because it cannot know whether something outside the project reads the relation. In this case the retirement was checked first: no dbt exposure, no Tableau extract config, and no remaining `ref()` from an enabled model. The prod BigQuery view is deliberately left standing rather than dropped, so anything that does still read it keeps working against frozen data instead of breaking.

No equivalent warning appears for `int_powerschool__section_grade_config`, which confirms that model is not contract-enforced.

Full CI state on `c40f4ce`: dbt Cloud pass, Trunk pass, CodeQL pass (4 checks). No test warnings, no freshness warnings.

GabyRangelB:
## Review response

Verdict per finding. All three verified against the code or the warehouse before replying.

### 1. Docs show `severity: error`, shipped is `severity: warn` — CONFIRMED

Real drift, and on an intentional decision, which is the worst kind to leave. Verified: the spec's sample block at line 163 and the plan's at line 310 both still read `severity: error`, while `rpt_deanslist__missing_assignments.yml:24` ships `severity: warn` with a `TODO(#3915)` comment explaining why. The fix-wave commit updated those docs for the `grade_category` value-domain note but did not touch the sample YAML. Doc-only, no effect on the shipped models. Queued for a follow-up touch-up pending the author's call on whether to push another commit.

### 2. Plan's file count reads seven, branch is eight — CONFIRMED

Also real, and self-inflicted in an amusing way: the same commit that corrected six to seven added `int_powerschool__section_grade_config.yml`, making it eight. `git diff --stat origin/main...HEAD` confirms 8 files. Doc-only, at a checkpoint whose execution is already complete. Same follow-up as finding 1.

### 3. NULL handling on the two exclusion filters — DECLINED, with data

The concern is sound in principle: `school_level_alt != 'ES'` and `_dbt_source_project != 'kippmiami'` both evaluate to NULL rather than TRUE when the column is NULL, so such a row would be silently dropped. You flagged that a live query was needed to know whether it ever fires. It does not — neither column is ever NULL on this population:

| `school_level_alt` | region | AY2025 rows with `is_expected_missing` |
| --- | --- | --- |
| HS | kippnewark | 82,367 |
| MS | kippnewark | 57,337 |
| HS | kippcamden | 47,371 |
| MS | kippcamden | 18,969 |
| MS | kippmiami | 11,026 |
| MS | kipppaterson | 3,267 |
| ES | kippmiami | 228 |
| ES | kipppaterson | 11 |

220,576 rows, zero with either column NULL. Both columns are populated by construction upstream: `_dbt_source_project` is derived at the union view and passed through, and `school_level_alt` comes from an `if()` over `stg_powerschool__schools.school_level` that this lineage reaches through an inner join to a scoped enrollment. Given that plus the roughly 30 existing call sites of the same idiom, no change is warranted.

As a bonus, this table independently re-derives the feed total from a different direction: 139,704 Newark plus 66,340 Camden plus 3,267 Paterson MS equals the 209,311 in the PR description, with Paterson's 11 ES rows and all 11,254 Miami rows correctly excluded.

### Not a finding, but worth recording

Also checked the dbt Cloud warning surface, reported separately above: one warning, the expected breaking-change notice from disabling a contracted model.

PR #4852, opened 2026-08-13
feat(scripts): de-identified demographics extract for the Quill pilot research partner

### LINKED ISSUE #4848: feat: de-identified demographics extract for the Quill pilot research partner
## Context

A signed Data Sharing Agreement (2026-02-13) covers the SY25-26 Quill.org pilot with an external research partner. The partner needs student demographics for the pilot roster so they can join their own platform telemetry to demographic covariates. Due Friday 2026-08-14.

Requested fields: race, gender, free or reduced meal status, IEP status, MLL status.

## What the request assumed versus what the data supports

The roster spreadsheet column labeled `Studetn ID` holds a Quill platform user ID, not a KIPP `student_number`. Zero of 248 values match `student_number` in the warehouse. The only usable join key in the roster is student email.

This works in our favor. A file keyed only on the Quill ID carries no KIPP identifier at all, which is what the agreement actually asks for: a persistent pseudonymous identifier, with the supplier retaining the re-identification key.

## Match rates against `int_extracts__student_enrollments`

| Measure | Count |
| --- | --- |
| Roster rows / unique students | 269 / 248 |
| Resolved to SY25-26 grade 8 (Rise, Purpose) | 244 |
| Resolved but last enrolled SY24-25 | 1 |
| Never in the warehouse under the roster email | 3 |
| Nulls across the five fields for the 244 | 0 |

Value domains are clean: race `B/T/H/I/P`, gender `F/M`, meal `F/R/P`, plus the `iep_status` and `ml_status` labels.

## Scope

- `scripts/extract_quill_seta_ed_demographics.py` — reads the roster from a local path (no roster data in git), queries `int_extracts__student_enrollments` at `academic_year = 2025`, `rn_year = 1`, and writes a two-sheet workbook plus a local-only re-identification key.
- `docs/reference/quill-seta-ed-demographics-extract.md` — data dictionary, variable-level lineage, inclusion and exclusion rules, rerun steps. Added to the Reference nav. The agreement requires this documentation alongside any derived dataset.
- Design spec under `docs/superpowers/specs/`.

Deliverable grain is one row per student-classroom pair (269 rows; 21 students sit in two sections of their own teacher). Classroom and teacher are pseudonymized to `C01`-`C09` and `T1`-`T2` so the partner can model classroom nesting without receiving any staff name. Label maps reuse the existing convention in `rpt_gsheets__csgf_enrollment`.

Grade level and school are deliberately excluded: every student is grade 8, and teacher code already separates the two schools 1:1, so both would add identifiability without adding information.

## Open items for the requester

- Confirm the five fields are enumerated in Exhibit A of the agreement. If they are not, this is a scope change requiring written approval from both signers rather than just a file.
- Race has singleton cells at n=244 (one AI/AN, one NH/OPI). Default is to ship real categories uncollapsed, since the agreement assigns small-cell suppression to the researcher at publication rather than to the supplier at transfer.
- Sending the file, and any pre-send review by the agreement owner, stay with the requester.

anthonygwalters:
# Pull Request

## Summary & Motivation

Add a one-off script and its documentation for producing a de-identified student
demographics workbook for the SY25-26 Quill.org pilot's external research
partner, under the Data Sharing Agreement signed 2026-02-13. Refs #4848.

**The finding that shaped the design:** the roster export's `Studetn ID` column
holds a Quill platform user ID, not a KIPP `student_number` — zero of 248
distinct values matched `student_number` in the warehouse. The join therefore
runs on student email, and the delivered workbook carries no KIPP identifier at
all. Had that column really held `student_number`, honoring the request
literally would have transmitted a FERPA direct identifier.

What merges:

- `scripts/extract_quill_seta_ed_demographics.py` — pure transforms (label maps,
  code assignment, row building, whitelist validation) wrapped by thin I/O edges
  (roster read, warehouse fetch, workbook write, retained key). Roster, output,
  and key paths are all arguments, so no student data enters git. PEP 723
  dependency header, matching `scripts/extract_ceds_schema.py`.
- `tests/test_extract_quill_seta_ed_demographics.py` — 73 tests over the pure
  core and the I/O edges.
- `docs/reference/quill-seta-ed-demographics-extract.md` — data dictionary,
  variable-level lineage, inclusion and exclusion rules, rerun steps. The
  agreement requires documentation sufficient for a qualified third party to
  reproduce the dataset; this page is that obligation.
- Design spec and implementation plan under `docs/superpowers/` (not in the
  docs nav), plus one `mkdocs.yml` nav entry and one `scripts/CLAUDE.md`
  catalog row.

**Deliverable already produced and verified locally** (not in this repo): 269
rows over 248 distinct Quill IDs, two sheets, 245 of 248 emails resolved, 1
prior-year fallback, 3 unmatched. Zero cells containing an at-sign or
name-shaped text. The retained re-identification key holds 269 records over 248
distinct IDs and is never committed or transmitted.

### Reviewer notes

Three things worth knowing before reading the diff:

1. **`validate_rows` is a whitelist, not a blacklist.** Every delivered cell
   must be in a known vocabulary. A name blacklist would false-fire on surnames
   colliding with labels (`Free`, `Male`) and miss anything unenumerated.
2. **Validation runs twice** — once in memory to fail fast, once against the
   written file re-read from disk, because a serialization fault is invisible to
   the first. The workbook is written to a temp path and only moved onto the
   delivery filename after both pass.
3. **Classroom and teacher codes are assigned by order of first appearance in
   the roster, not alphabetically.** This script is public; an alphabetical rule
   would be derivable by anyone who knows which teachers taught the sections.

### AI Assistance

Claude Code authored the script, tests, and documentation under human direction.
The design was settled interactively before any code was written. Execution ran
five task-scoped implementer/reviewer pairs plus a whole-branch review, with
nine coordinator rulings recorded; the substantive defects those reviews caught
were in the plan rather than the implementation, including two silent-wrong-data
paths in the roster reader and a validation gate that admitted a blank
demographic on a matched row. All warehouse access is a single read-only
`SELECT` with bound query parameters.

### Open items — not blocked by this PR, but blocking the send

Refs rather than Closes #4848, because these remain open on that issue:

- **Exhibit A of the agreement is unconfirmed.** If the five requested fields
  are not enumerated there, sending the file is a scope change requiring both
  signers. Merging this PR sends nothing.
- **Small-cell exposure is broader than race.** Alongside two singleton race
  categories, the delivered file has `mll_status = ML` at 9, `meal_status =
  Reduced` at 22, and a single `SY24-25 (fallback)` row, which is itself a
  quasi-identifier. Crossed with `classroom_code`, several narrow to one
  student. The agreement assigns suppression to the researcher at publication,
  so the default here ships real categories uncollapsed.
- **Label parity with `rpt_gsheets__csgf_enrollment` is partial and now
  documented as such.** That model emits `BL/AA`, `AI/AN`, `NH/OPI` with
  slashes; this extract uses hyphens, so the strings will not join in a
  KTAF-internal comparison. Gender follows
  `int_extracts__student_enrollments.aligned_gender`, and the Free/Reduced/Paid
  trichotomy is introduced here. Switching to exact parity is three dict values
  and a deterministic rerun, and is the agreement owner's call.

## Self-review

### General

No Dagster changes.

No dbt changes. The extract reads `int_extracts__student_enrollments` but adds
no model, so no properties file or exposure applies.

No schedule, sensor, or integration changes. The new reference page is added to
the `mkdocs.yml` nav.

## CI checks

- [ ] **Trunk** — passes. `trunk check --force` reports no issues across all
      seven changed files locally.
- [ ] **dbt Cloud** — passes.

## Troubleshooting


anthonygwalters:
Verified each finding against the code and the warehouse before acting. Three fixed, three declined with reasoning, one fixed plus test coverage. Fixes are in `77ca60c72`.

### Important 1 — retained key written after the workbook reaches its sendable filename

**Fixed.** Confirmed by reading the statement order: `os.replace(...)` ran, then `write_key_file(...)`.

This one also overturned a prior reviewer on this branch, which had argued the opposite ordering was correct on the grounds that the key should not record a delivery that did not ship. The failure modes are asymmetric and the argument does not survive that:

- Key after rename, `write_key_file` fails: a correctly-named, fully-formed workbook sits on disk with no re-identification key. That key is the one artifact the agreement obliges the supplier to retain, and the file looks ready to send.
- Key before rename, `os.replace` fails: the key holds records for a delivery that did not ship. Nothing consumes it as a delivery ledger, it is cumulative, and a rerun rewrites it.

`write_key_file` now completes before the rename, with a comment at the call site explaining the ordering so it does not get swapped back.

Added `test_key_file_is_written_before_workbook_is_renamed` to guard it. I required evidence the guard actually bites rather than passing vacuously: with the reordering reverted it fails (`assert not True` — the workbook existed at its final name despite the raised exception); with the fix restored, 75/75 pass.

### Important 2 — no deterministic tiebreak in the demographics window function

**Fixed**, though the collision is not reachable in current data. I ran the verification query you suggested, and the broader one:

- Lowercased emails shared by more than one `student_number` within an `academic_year`, network-wide: **0** across 123,832 non-null emails.
- Same, restricted to this pilot's 248 roster emails: **0**.

Added `student_number desc` as the secondary key anyway. It costs one identifier in an `order by`, and it makes reproducibility a property of the query rather than a property of today's data — which matters because the agreement imposes the reproducibility requirement. Confirmed a no-op on current data: the post-fix rerun reproduced every figure exactly (269 roster rows / 248 students, 245 matched, 265 / 1 / 3 on `demographics_source`, two singleton race cells).

### Minor 4 — JSON reads and writes omit an explicit encoding

**Fixed.** All three call sites now pass `encoding="utf-8"`. Classroom and teacher names can carry non-ASCII, and the retained key is the artifact least tolerant of an environment-dependent round-trip.

### Minor 7 — `main()` has no test

**Fixed.** Added two: a happy-path run with the warehouse fetch and client construction stubbed out (no network, no credentials), asserting the workbook lands at its final name, the key file exists, and no temp file is left behind; plus the ordering guard described under Important 1. You were right that this was the one place with no coverage and that it was exactly where the Important bug lived.

### Minor 3 — `gender` re-derives `aligned_gender`

**Declined for this PR**, recorded as a follow-up. The duplication is real and the reuse argument is correct. It is a refactor touching the query, the label map, and its tests, and it has now been raised and deferred three times across this branch's reviews. Widening a PR that is otherwise verified end to end, against a same-week deliverable, is the wrong trade. Filing it as follow-up work rather than pretending it does not exist.

### Minor 5 — SELECT column ordering does not follow ST06

**Declined.** The convention is genuine — `src/dbt/CLAUDE.md` documents it under "SQL column ordering in SELECT clauses (enforced by ST06)" — but it is scoped to `.sql` files that sqlfluff actually lints. This is an ad-hoc query embedded in a Python string that no linter touches, and the current order leads with `lower(student_email)`, which is the join and partition key for the whole query. Reordering it to satisfy a rule that is not enforced here would make the query read less clearly for its purpose.

### Minor 6 — `trim()` asymmetry between roster and SQL normalization

**Declined.** Ran the verification you specified: warehouse `student_email` values differing from their trimmed form: **0** of 123,832 non-null rows. The roster side strips because a spreadsheet export is an untrusted input; the warehouse side is a modeled column with no observed whitespace. Adding `trim()` would be defensive against a condition that does not occur in the source.

---

Note that `claude-review` triggers only on `opened` and `ready_for_review`, so this push will not re-fire it. All eight checks were terminal and green before the fix commit; Trunk will re-run on the new head.

PR #4854, opened 2026-08-13
fix(dbt): restore crosswalk schemas as rpt_gsheets views for seat tracker

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will add two extract views that restore the
pre-2026-08-11 schemas of the people location and campus crosswalks, so the
Seat Tracker AppSheet can be repointed off the staging models it was reading
directly.

**What broke.** The people spreadsheet was refactored: the `Location Crosswalk`
and `Campus Crosswalk` tabs were trimmed to two columns each and every
attribute moved to the `Locations` tab. Both `stg_google_sheets__people__*_crosswalk`
models are `select *`, so they went from 12 and 8 columns to 2. The Seat
Tracker AppSheet connected straight to those staging tables (against
convention) and lost 10 and 6 columns respectively.

**The fix.**

- `rpt_gsheets__people__location_crosswalk` renames
  `int_people__location_crosswalk` back to the old column names. That
  intermediate already performs the alias to `Locations` join.
- `rpt_gsheets__people__campus_crosswalk` joins the trimmed campus tab to
  `stg_google_sheets__people__locations` on `Location_Name`.
- `seat_tracker_app` exposure now depends on the two `rpt_` models instead of
  the `stg_` ones.

Both views are 1:1 with the old tables on column names, order, and types, so
the AppSheet keeps its existing column config (keys, labels, types) and only
needs its data sources repointed to `kipptaf_extracts`.

**Verification.** Diffed both views against the old tables recovered via
BigQuery time travel. Campus matches on all 17 rows. Location matches on 248 of
251; the 3 that differ (Paterson Prep Elementary, Paterson Prep Middle,
Poinciana Campus) are rows where the old table carried nulls and the new view
resolves the attributes correctly from `Locations`.

The attribute columns on the campus crosswalk describe the child
`Location_Name`, not the parent campus `Name` — confirmed against the old data,
where campus rows carry blank abbreviation and grade band and `0` school IDs.

## AI Assistance

Claude Code authored the models, properties files, and exposure update, and ran
the verification queries. Human-directed: the requirement to restore the old
schemas as `rpt_gsheets__*` views rather than revert the spreadsheet refactor,
and the decision to keep the views exactly 1:1 rather than add extra columns
from the `Locations` tab.

### General

### dbt

      analysis, or application
- [x] **Breaking change?** No. Both models are new; no existing contracted
      column is renamed or removed. Rollback is deleting the two models and
      reverting the exposure, which leaves the AppSheet exactly as it is today.
      is needed.

## CI checks

- [ ] **Trunk** — `trunk check --force` clean locally on all 5 changed files.
- [ ] **dbt Cloud** — both views built and passed contract enforcement locally
      against the deferred prod state.
- [ ] **Dagster Cloud** — no Dagster paths changed.

### DIFF
--- a/src/dbt/kipptaf/models/exposures/google-appsheet.yml
+++ b/src/dbt/kipptaf/models/exposures/google-appsheet.yml
@@ -6,8 +6,8 @@ exposures:
       name: Data Team
     depends_on:
       - ref("rpt_appsheet__seat_tracker_roster")
-      - ref("stg_google_sheets__people__campus_crosswalk")
-      - ref("stg_google_sheets__people__location_crosswalk")
+      - ref("rpt_gsheets__people__campus_crosswalk")
+      - ref("rpt_gsheets__people__location_crosswalk")
     config:
       meta:
         dagster:
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__people__campus_crosswalk.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__people__campus_crosswalk.yml
@@ -6,24 +6,50 @@ models:
       directly until the source sheet was refactored down to
       `Name` + `Location_Name` and every attribute moved to the `Locations`
       tab. Column names, order, and types are 1:1 with that old table so the
-      app only has to repoint its data source. The attribute columns describe
-      the child `Location_Name`, not the parent campus `Name`.
+      app only has to repoint its data source. Every attribute column
+      describes the child `Location_Name`, not the parent campus `Name` —
+      campus rows in the `Locations` tab carry blank abbreviation and grade
+      band and zero school IDs.
     columns:
       - name: Location_Name
-        description: Canonical location name of the child school.
+        description: |
+          Canonical location name of the child school. Join key to
+          `int_people__staff_roster.home_work_location_reporting_name` and to
+          `int_people__location_crosswalk.location_clean_name`.
         data_type: string
+        data_tests:
+          - unique:
+              config:
+                severity: error
       - name: Name
-        description: Parent campus the location rolls up to.
+        description: |
+          Parent campus the location rolls up to. Repeats across sibling
+          schools sharing a campus, and equals `Location_Name` for a location
+          that is itself a campus.
         data_type: string
       - name: Region
+        description: |
+          Legal-entity name the child school rolls up to (e.g. TEAM Academy
+          Charter School, KIPP Miami).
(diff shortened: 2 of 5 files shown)


cbini:
## Review verdicts

Verified each citation against `src/dbt/CLAUDE.md` before acting. Fixed in
`8ab1fa275`.

### Important 1 — missing per-column descriptions: **accepted, fixed**

Citation checks out verbatim at `src/dbt/CLAUDE.md:1246-1247`. All 12 location
columns and all 8 campus columns now carry a `description:`, sourced from
`stg_google_sheets__people__locations.yml` where prose already existed.

### Important 2 — no uniqueness test: **accepted, fixed**

Citation checks out verbatim at `src/dbt/CLAUDE.md:759-762`, and the
thin-wrapper exception at `:764-768` does not apply — it covers district
wrappers *sourcing* `kipptaf_extracts`, not views that own the logic. Added
`unique` with `severity: error` on `Name` (location) and `Location_Name`
(campus). Both pass locally.

One correction to the calibration note — actual practice in this directory is
16 of 59 properties files, not 15 of 61. Same conclusion either way.

I had originally skipped these as redundant with the upstream `stg_` tests. The
convention is stated without a redundancy carve-out, and each test is one line,
so redundancy was the wrong reason to skip.

### Minor 1 — `Head_of_Schools_Employee_Number` untagged for PII: **declined**

`employee_number` is a direct identifier under the repo's PII reference, so the
underlying observation is right. Declining the change because no
`employee_number` column anywhere in this repo carries `contains_pii` —
including `stg_google_sheets__people__locations.head_of_schools_employee_number`,
the exact upstream this column restores. Tagging one leaf column would make the
lineage inconsistent without changing who can read the data, since nothing in
`src/cube/` or `src/teamster/` reads the flag for extract models.

Worth deciding network-wide rather than in a restore PR. Raising it with the
data team separately.

### Recommendation — `relationships` test from campus `Location_Name` to
location `Clean_Name`: **declined**

No stated convention requires it, and `Clean_Name` is deliberately non-unique
in the location view (one row per alias), so the test would document less than
the two `unique` tests already added. Leaving it out as YAGNI.

### Data-dependent item

Already done pre-merge, via BigQuery time travel against the pre-refactor
tables rather than post-merge row counts: campus matches on all 17 rows,
location on 248 of 251. The 3 that differ are the Paterson Prep ES/MS and
Poinciana Campus rows, where the old table carried nulls and the new view
resolves the attributes correctly from the `Locations` tab. Details in the PR
body.

PR #4856, opened 2026-08-13
fix(finalsite): route untyped phone numbers into the DeansList and ParentSquare feeds

### LINKED ISSUE #4855: fix(finalsite): untyped phone numbers are dropped from the DeansList and ParentSquare feeds
## Problem

Finalsite stores each phone as a number plus a `phone_type` label (`Cell` / `Home` / `Work`). When the label is left unset it arrives as the empty string, not null.

`int_finalsite__student_contacts` derives `phone_mobile` / `phone_home` / `phone_work` by exact-matching those three labels:

```sql
coalesce(
    if(cp.phone_1_type = 'Cell', cp.phone_1_number, null),
    if(cp.phone_2_type = 'Cell', cp.phone_2_number, null),
    if(cp.phone_3_type = 'Cell', cp.phone_3_number, null)
) as phone_mobile
```

An untyped number matches none of the three, so all three columns come out null and the number is dropped. `phone_primary` retains `phone_1_number` unconditionally, but nothing retains an untyped `phone_2` or `phone_3`.

Downstream, `rpt_deanslist__family_contacts` projects only the three typed columns into `HomePhone` / `WorkPhone` / `CellPhone`, so the family lands in DeansList with a name, a relationship, an email, and no callable number. `rpt_parentsquare__parents` reads `phone_mobile` only and has the same gap.

Reported by school staff at KIPP Cooper Norcross High (Camden) during attendance calls: numbers visible in Finalsite, blank in DeansList.

## Scope, measured against current prod

Phone entries carrying a number but no type, across Newark, Camden and Paterson:

| Position | Untyped entries with a number |
| --- | --- |
| `phone_1` | 2,941 |
| `phone_2` | 2,432 |
| `phone_3` | 2,093 |

Only the `phone_1` group is reachable today via `phone_primary`, so a fix confined to the extract recovers 39 percent of them.

Impact on the current DeansList feed (enrolled students, NJ regions):

| Region | Contact rows losing a number | Students with no phone at all |
| --- | --- | --- |
| Newark | 598 | 34 |
| Camden | 161 | 10 |
| Paterson | 53 | 2 |

Every one of those students has a number on file in Finalsite.

The type domain in the data is exactly four values: `Cell` (17,438 numbers), `Home` (714), `Work` (340), and empty string (7,466). There is no `Other` / `Mobile` / `Fax` variant, so "missing type" is the single case to handle.

Among phones that DO carry a type, 95 percent Newark / 96 percent Camden / 85 percent Paterson are `Cell`.

## Fix

Add a `phone_untyped` column to `int_finalsite__student_contacts` — the first number across positions 1-3 whose type is blank — in the parent branch and in each of the four `emrg_N` blocks, mirroring the existing coalesce shape. Leave `phone_mobile` / `phone_home` / `phone_work` meaning exactly what they mean now.

Then have each consumer decide:

- `rpt_deanslist__family_contacts`: `coalesce(phone_mobile, phone_untyped)` into `CellPhone`
- `rpt_parentsquare__parents`: same recovery for the mobile candidate

Unit tests to cover the untyped path in all three models.

## Follow-up, not in scope here

Requiring a type at data entry in Finalsite so the backlog stops growing. That is an Ops change and does not help the families already affected.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop Finalsite phone numbers from being
dropped on their way into DeansList and ParentSquare.

Finalsite lets a phone be saved without selecting a Cell/Home/Work label.
`int_finalsite__student_contacts` derived `phone_mobile` / `phone_home` /
`phone_work` by exact-matching those three labels, so an untyped number reached
none of them and was dropped. `rpt_deanslist__family_contacts` projects only
those three columns, so the family arrived in DeansList with a name, a
relationship, an email, and no callable number. Reported by KIPP Cooper Norcross
High while making attendance calls.

Three changes:

1. `stg_finalsite__contacts` normalizes a blank phone type to NULL. A contact
   record emits an empty string where the `emrg_N` custom-field sets emit NULL,
   and both mean the same thing. `rpt_focus__contacts` already asserts NULL for
   the unset case in its own unit tests, so this brings prod in line with what
   the tests already assume.
2. `int_finalsite__student_contacts` gains `phone_untyped` — the first number
   across all three phone positions carrying no type label, derived in the
   parent branch and in each of the four emergency blocks. `phone_mobile`,
   `phone_home` and `phone_work` keep meaning exactly what they meant before, so
   no existing consumer changes behavior.
3. The two outbound feeds opt in. `rpt_deanslist__family_contacts` coalesces it
   into `CellPhone`, which is where it has to ride in because the DeansList
   template has no header for an untyped number; roughly 95 percent of NJ
   numbers that do carry a type are Cell, and a Finalsite-typed Cell still wins
   outright. `rpt_parentsquare__parents` takes the same fallback for its mobile
   candidate, where a parent carrying neither a mobile nor an email was being
   dropped as undeliverable — which is how these families went missing from
   ParentSquare entirely.

`int_students__contacts` threads the column through so ParentSquare can reach
it. Its Focus branch is a NULL literal: Focus types every number it stores.

### Impact, measured against current prod

Recovers 812 contact rows across the NJ regions — Newark 598, Camden 161,
Paterson 53 — including 46 students who currently have no phone in DeansList at
all despite having one on file in Finalsite.

### AI Assistance

Claude Code did the root-cause investigation, wrote the change, and ran the
verification below. Human-directed: the decision to normalize blank to NULL at
staging rather than work around it downstream, and the scope call to fix both
feeds in one change.

### Rollback plan

Additive. `phone_untyped` is a new column and the two feeds fall back to it only
when all three typed columns are NULL, so reverting restores the prior output
exactly. The staging normalization changes `phone_1_type` / `phone_2_type` /
`phone_3_type` from empty string to NULL, which also reaches
`rpt_focus__contacts` (Miami, Focus CONTACTS import) — that feed passes the
label straight through, and its unit tests already expect NULL for the unset
case.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — column
      docs added for `phone_untyped` on both intermediate models, and the
      `phone_1_type` / `phone_2_type` / `phone_3_type` docs updated to state
      that null means no type was selected.
- [x] **Breaking change?** No columns renamed or removed. One column added to
      two intermediate models, one value normalized in a staging model.

## Verification

Run locally against the worktree:

- `dbt parse --no-partial-parse` — clean on both `kippcamden` and `kipptaf`.
- `dbt build --select stg_finalsite__contacts int_finalsite__student_contacts`
  against `kippcamden` dev — `PASS=11 WARN=3 ERROR=0`. The 3 warns are
  pre-existing contact-slotting and staleness tests that reference no phone
  column.
- Unit tests — 7 pass, covering `int_finalsite__student_contacts` (2),
  `rpt_deanslist__family_contacts` (4) and `int_students__contacts` (1).
  `test_family_contacts_untyped_phone_fills_cellphone` is new and asserts both
  that an untyped number surfaces as `CellPhone` and that a typed Cell still
  wins when both are present.
- `trunk check --force` over all 9 changed files — no issues.

## CI note

This adds a column to a package model that `kipptaf` reads through a
`union_relations` wrapper, so the per-district `zz_stg_*` copies need seeding
before dbt Cloud CI can see `phone_untyped`. Expect the wrapper rebuild to fail
on `Name phone_untyped not found` until that runs.

Refs #4855

### DIFF
--- a/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
+++ b/src/dbt/finalsite/models/api/intermediate/int_finalsite__student_contacts.sql
@@ -101,6 +101,11 @@ with
                 if(cp.phone_2_type = 'Work', cp.phone_2_number, null),
                 if(cp.phone_3_type = 'Work', cp.phone_3_number, null)
             ) as phone_work,
+            coalesce(
+                if(cp.phone_1_type is null, cp.phone_1_number, null),
+                if(cp.phone_2_type is null, cp.phone_2_number, null),
+                if(cp.phone_3_type is null, cp.phone_3_number, null)
+            ) as phone_untyped,
             nullif(
                 array_to_string(
                     [cp.address_1, cp.address_2, cp.city, cp.state, cp.zip], ', '
@@ -122,6 +127,7 @@ with
             phone_mobile,
             phone_home,
             phone_work,
+            phone_untyped,
             home_address,
             first_name as contact_first_name,
             last_name as contact_last_name,
@@ -171,6 +177,11 @@ with
                 if(emrg_1_phone_2_type = 'Work', emrg_1_phone_2_number, null),
                 if(emrg_1_phone_3_type = 'Work', emrg_1_phone_3_number, null)
             ) as phone_work,
+            coalesce(
+                if(emrg_1_phone_1_type is null, emrg_1_phone_1_number, null),
+                if(emrg_1_phone_2_type is null, emrg_1_phone_2_number, null),
+                if(emrg_1_phone_3_type is null, emrg_1_phone_3_number, null)
+            ) as phone_untyped,
         from {{ ref("int_finalsite__contact_custom_attributes") }}
         where emrg_1_name_first_name is not null and emrg_1_name_first_name != ''
 
@@ -208,6 +219,11 @@ with
                 if(emrg_2_phone_2_type = 'Work', emrg_2_phone_2_number, null),
                 if(emrg_2_phone_3_type = 'Work', emrg_2_phone_3_number, null)
             ) as phone_work,
+            coalesce(
+                if(emrg_2_phone_1_type is null, emrg_2_phone_1_number, null),
+                if(emrg_2_phone_2_type is null, emrg_2_phone_2_number, null),
+                if(emrg_2_phone_3_type is null, emrg_2_phone_3_number, null)
+            ) as phone_untyped,
         from {{ ref("int_finalsite__contact_custom_attributes") }}
         where emrg_2_name_first_name is not null and emrg_2_name_first_name != ''
 
@@ -245,6 +261,11 @@ with
                 if(emrg_3_phone_2_type = 'Work', emrg_3_phone_2_number, null),
                 if(emrg_3_phone_3_type = 'Work', emrg_3_phone_3_number, null)
             ) as phone_work,
+            coalesce(
+                if(emrg_3_phone_1_type is null, emrg_3_phone_1_number, null),
+                if(emrg_3_phone_2_type is null, emrg_3_phone_2_number, null),
+                if(emrg_3_phone_3_type is null, emrg_3_phone_3_number, null)
+            ) as phone_untyped,
         from {{ ref("int_finalsite__contact_custom_attributes") }}
         where emrg_3_name_first_name is not null and emrg_3_name_first_name != ''
 
@@ -282,6 +303,11 @@ with
                 if(emrg_4_phone_2_type = 'Work', emrg_4_phone_2_number, null),
                 if(emrg_4_phone_3_type = 'Work', emrg_4_phone_3_number, null)
             ) as phone_work,
+            coalesce(
+                if(emrg_4_phone_1_type is null, emrg_4_phone_1_number, null),
+                if(emrg_4_phone_2_type is null, emrg_4_phone_2_number, null),
+                if(emrg_4_phone_3_type is null, emrg_4_phone_3_number, null)
+            ) as phone_untyped,
         from {{ ref("int_finalsite__contact_custom_attributes") }}
         where emrg_4_name_first_name is not null and emrg_4_name_first_name != ''
     ),
@@ -301,6 +327,7 @@ with
             phone_mobile,
             phone_home,
             phone_work,
+            phone_untyped,
             phone_primary,
             is_pickup,
             is_custodial,
@@ -327,6 +354,7 @@ with
             phone_mobile,
             phone_home,
             phone_work,
+            phone_untyped,
             phone_daytime,
             phone_primary,
             home_address,
@@ -350,6 +378,7 @@ with
             phone_mobile,
             phone_home,
             phone_work,
+            phone_untyped,
             phone_daytime,
             phone_primary,
             home_address,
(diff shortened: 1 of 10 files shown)


cbini:
Verdicts on each finding. Fixed in 9f43fc69d.

**1. Staging normalization has no test that exercises it — accepted, fixed.**

Verified before acting: `test_contacts_phone_normalization` fed `phone_3` a
`cast(null as string) as phone_type`, so the empty string Finalsite actually
emits never reached the new `nullif`. The test passed identically with and
without the change — the central transformation of this PR was unexercised.

`phone_3` now carries `struct('' as phone_type, ...)`. The expect block already
asserted `cast(null as string) as phone_3_type`, so that existing assertion is
now the proof the normalization fired — nothing else in the model turns `''`
into null. Passes.

**2. ParentSquare fallback has no unit test — accepted, fixed.**

Confirmed the file had no `unit_tests:` block at all. Added
`test_parents_untyped_phone_fills_mobile` with the two-row shape you described:
a parent carrying only an untyped number surfaces it as `mobile`, and a parent
carrying both a typed Cell and an untyped number keeps the typed Cell. Passes.
Your point about a flipped `coalesce` argument order is the reason it is worth
having — that failure mode is silent, and its symptom is the parent being
dropped as undeliverable rather than an error.

**3. `mobile` doc not updated — accepted, fixed.**

Reworded to describe the fallback and why it matters (a parent with neither a
mobile nor an email is dropped as undeliverable), for parity with the
`CellPhone` doc.

**4. Verification only built one of three NJ regions — already addressed, after
the PR body was written.**

All three NJ districts have since been built with `--target staging` to seed the
`zz_stg_*` copies for CI: `kippnewark`, `kippcamden`, `kipppaterson`, each
`ERROR=0`. Those are real builds of this SQL against each region's own data, so
the region-specific data-shape concern is covered — the `kippcamden` dev build
quoted in the PR body was just the first of the four. The kipptaf-level union
wrapper was also rebuilt into `zz_stg_kipptaf_finalsite`, which is what the
CI note anticipated; dbt Cloud is green as of that seeding.

**5. Focus-branch test does not assert `phone_untyped` is null — declined.**

That test is `test_focus_slotting_contact_1_and_emergency_ranks`, and its job is
slot assignment; its `expect` rows deliberately carry only the slotting columns.
Adding a phone assertion widens what the test claims to cover without adding
signal, since the value is a `cast(null as string)` literal sitting in the SELECT
list with no expression that could drift. If the Focus branch ever derives a real
untyped number, the assertion belongs in a test written for that behavior rather
than grafted onto this one.

**6. Macro to collapse the repeated coalesce/if phone-picking blocks — declined,
agreeing with your own read.**

The duplication predates this PR across the mobile/home/work columns and five
blocks. Collapsing it is a refactor of the emergency-block structure, not part of
a targeted bug fix, and it would put a large unrelated diff in front of a change
that is currently reviewable line by line.

One correction to the record, unprompted: the PR body argued the extract-level
alternative would recover only 39% of untyped numbers. That figure is true of
untyped phone *entries* (7,466 total, 2,941 in position 1) but not of *feed
rows* — measured against prod, both approaches recover the same 812 rows, because
a contact whose position-1 phone is typed already had a number in the feed and was
never blank. The design still stands on semantics (`phone_mobile` keeps meaning
"Finalsite says Cell", and every consumer gets the fix rather than one extract),
but the coverage gap it claimed does not exist in today's data.

Issue #4736, opened 2026-08-05
feat(dagster): midday cadence for Finalsite to Focus import, ready in Focus by 2pm ET


cbini:
## Goal

Enrollment leadership wants to tell stakeholders that **new students entered in Finalsite by 12:00pm ET are usable in Focus by 2:00pm ET**. Focus imports are run by hand today, so the pipeline has to deliver the four CSVs to the Focus SFTP `incoming/` folder early enough for ops to run the imports before 2pm.

Secondary ask: sync the warehouse from Focus **after** 2pm so the warehouse reflects what was just imported.

## Current chain (America/New_York)

| Step | When | Owner |
| --- | --- | --- |
| Finalsite SFTP `status_report` drop lands | ~2:50am observed | Finalsite-side export |
| Finalsite API `contacts` pull | 4:00am cron | us |
| Focus dlt pull | 4:00am cron | us |
| dbt rebuild | dependency-driven, ~3m 34s after last ingest | us (no cron) |
| 4 CSVs to Focus SFTP `incoming/` | 5:00am cron, schedule **STOPPED** | us |
| Focus import | manual | ops |

## Measured latencies

- dbt: `status_report` ingested 11:07:04 ET, `int_finalsite__enrollment_lifecycle` rebuilt 11:10:39 ET, so **3m 34s** end to end. The `rpt_focus__*` models are views, evaluated at query time.
- Focus dlt pull: 3-6 min.
- Finalsite `contacts` pull: 4-7 min of work, but up to 46 min of queue wait at 4am because all four districts share the `finalsite_api` pool at limit 1. A midday run has the pool to itself.

## Proposed schedule

| Time | Step |
| --- | --- |
| 12:00 | Finalsite cutoff; ops manually pushes the SFTP export |
| 12:00-12:10 | Couchdrop sensor ingests `status_report` (10 min poll, no change) |
| 12:10 | Finalsite API `contacts` pull (**added** midday cron) |
| 12:25 | Focus dlt pull (**added** midday cron) |
| ~12:30 | dbt rebuilds automatically |
| 12:45 | 4 CSVs delivered to Focus SFTP `incoming/` (**retimed**, schedule **started**) |
| 12:45-2:00 | Ops runs the four Focus imports |
| ~2:45 | Focus dlt pull #2 (**added**) so the warehouse reflects the imports |

## Add midday runs, do not move the overnight ones

[...]

PR #4861, opened 2026-08-14
feat(focus): SFTP feed changes for withdrawals, contact rules, phones, and schedules

### LINKED ISSUE #4769: feat(focus): SFTP feed changes for withdrawals, presence filter, contact rules, phones, and schedule
## Context

Eight changes to the KIPP Miami Focus SFTP import feed.

The feed is built in two layers. kipptaf `rpt_focus__*` builds desired state from Finalsite; kippmiami `rpt_focus__*` reconciles that against live Focus and is what actually ships. Dagster delivers 4 CSVs to the Focus `incoming/` folder per `src/teamster/code_locations/kippmiami/extracts/config/focus.yaml`.

## Sources

- Asana: https://app.asana.com/1/913513768672/project/1215209393570629/task/1217278230672585
- Spec doc: [Focus Import v2 Specs](https://docs.google.com/document/d/1tMWa_-vRXk7LWMtbacs9kOEGuMd-9a--6JdDBclQJpA) — a five-goal list plus Laszlo's working BigQuery queries for demo, enrollments, addresses, and contacts.

Read the doc's SQL as the target output shape Laszlo assembles by hand today, not as a drop-in replacement for our models. Its demo aliases (`custom_53`, `custom_200000004`, `custom_100000105`) are Focus custom field IDs, and its demo query selects 8 columns where `rpt_focus__demographics` ships roughly 40.

The doc covers four exports — Demo, Enrollments, Addresses, Contacts. It says nothing about `LINKED_STUDENTS`.

## Resolved decisions

| # | Question | Resolution |
| --- | --- | --- |
| A | Is presence in Focus absolute, or scoped to the academic year? | Per-academic-year for **enrollment only**. Demographics keeps the absolute presence check. |
| B | Does per-feed backfill end? | No. Addresses and contacts import for a student already in Focus when the target is empty. The #4320 blank-name and street-line guards stay alive. |
| C | Gender split on the relation crosswalk, or flat? | Keep the gender split. The doc's objection is to name-based guessing; this reads a real `gender` column, populated 100% for three of the four rollup buckets. |
| D | Seven phone slots or one? | Seven — clean each phone in place, in the slot Finalsite put it in. No ranking, no packing, no dropping contacts. |
| E | Normalize emergency-contact phones too? | **Already done.** `int_finalsite__contact_custom_attributes` lines 239-250 already apply `clean_phone` to all 12 `emrg_N_phone_N_number` columns. No work required. |
| F | Where does that normalization go? | Moot — see E. It is already at the custom-attributes pivot, the earliest point each emergency phone is a scalar column. |
| G | `emergency`, `pick_up`, and `callout` rules? | Follow the doc: blanket `Y` on all three. |
| H | Which side does the contacts presence anti-join live on? | kippmiami. |
| I | Are withdrawals part of the import? | No — not the exit rows, not the columns, not the students. See subtask 1. |
| J | Blank or unrecognized phone type? | Follow the doc: default to Cell Phone, with an inline comment explaining why. |
| K | Dashed `###-###-####` or E.164? | E.164. No format conversion. |
| L | Missing `contact7_callout` — intentional or a bug? | Treat as a truncation bug. Add the column. |
| M | Adding `contactN_sms` columns? | Proceed. 7 new columns under decision D. |
| N | Does `priority` replace `sort_order`? | Neither — `priority` is redundant under decision D. Keep `sort_order`, drop `priority`. |
| O | Is `Emergency` a valid Focus relation type? | **No.** Verified against live Focus data: 17 distinct `student_relation` values over 3,645 rows, no `Emergency`. Keep the domain at 13 values. |
| P | `rpt_focus__linked_students` — wire up or delete? | Neither. Disable it. See subtask 8. |
| Q | Where does repeated-digit junk rejection go? | In `rpt_focus__contacts.sql`, not the `clean_phone` macro. The macro documents "never returns NULL for a non-null input" and is shared by four extracts; a guard inside its `CASE` would pass junk through de-garbled rather than rejecting it, and changing that contract would alter values for ParentSquare, DeansList, and `int_students__contacts`. The requirement is Focus-specific, so scope it to the Focus feed. |

No open decisions remain.

### 1. Exclude withdrawals from student enrollment entirely

Per decision I, no withdrawal reaches the import — not as an exit row, not as a closed entry row, not as a column. In `src/dbt/kippmiami/models/extracts/focus/rpt_focus__student_enrollment.sql`:

1. Delete the `exits` CTE and its `UNION ALL` branch. 9 of 161 current rows are exits.
1. Filter the `entries` branch to `where d.end_date is null`, so a withdrawn student never imports at all.
1. Drop `end_date` and `drop_code` from the output column list.
1. Delete the `stg_focus__student_enrollment_codes` decode join in the `desired` CTE — dead once `drop_code` is gone.
1. Delete the `focus_open` CTE — it only fed `exits`.

Step 2 is the load-bearing one and goes beyond what the doc's commented-out columns imply. The `entries` branch today sends the full row "with a same-run exit filled in if the desired row already carries one", so a student who enrolled and withdrew and is not yet in Focus currently imports already-closed. Dropping only the columns would import that same student as an **active** enrollment with no end date — a phantom active student. The filter prevents that.

Consequence to accept knowingly: Focus will hold no record that those students were ever enrolled. Confirm that is acceptable for Florida state reporting before shipping.

This also settles Laszlo's commented-out `end_date` and `drop_code` — they are correctly excluded.

### 2. Presence filter across imports

| Feed | Presence check today | Target rule |
| --- | --- | --- |
| `rpt_focus__demographics` | `stg_focus__students` | unchanged — absolute presence |
| `rpt_focus__addresses` | `stg_focus__students_join_address` | in the demo feed, or in Focus with no address |
| `rpt_focus__contacts` | `stg_focus__students_join_people` | in the demo feed, or in Focus with no contacts |
| `rpt_focus__student_enrollment` | `student_id` plus `syear` match only | no enrollment row for the student in the current academic year |

Per decisions A and B this is a per-feed rule, not one global anti-join. The two backfill feeds gate on the target being empty for that student, not on the student being absent from Focus.

Per decision H the contacts anti-join lives on the **kippmiami** side, in the reconciliation layer that actually ships — not on the kipptaf `int_focus__student_contacts` union the doc's query uses.

### 3. Validate contact relationship type

Crosswalk `student_relation` in `src/dbt/kipptaf/models/extracts/focus/rpt_focus__contacts.sql` to the Focus vocabulary, plus an `accepted_values` test on the crosswalked output. No such test exists today.

**This is already a live data-quality problem, not a hypothetical.** Querying `stg_focus__students_join_people` shows four of our own un-crosswalked feed values have already imported into Focus: `parent` (9 rows), `relative` (1), `aunt/uncle` (1), `stepparent` (1). Twelve rows of lowercase leakage sitting in production Focus today, alongside the 13 legitimate values.

Focus does not enforce this field. `int_focus__custom_field_options` has no person or contact `source_class`, and neither `focus-api-spec.md` nor `focus-db-erd.md` carries an enum. Our own test is the only gate this field will have.

Domain, confirmed by query against live Focus (decision O) — 13 values:

`Mother` (1,310), `Father` (633), `Grandmother` (418), `None` (412), `Aunt` (318), `Parent` (286), `Uncle` (104), `Grandfather` (85), `Guardian` (27), `Stepfather` (22), `Stepparent` (9), `Stepmother` (7), `Surrogate` (2)

Do **not** add the doc's `Emergency`. It does not appear in Focus at all, and the emergency branch does not emit that literal — it sets `student_relation` from `coalesce(a.emrg_1_relationship_ss, a.emrg_1_relationship_txt)`, so those rows carry real relationship labels that run through the same crosswalk as everything else. Anything unmatched already lands on `None`.

Decided (decision C): roll up to the nearest Focus value, splitting by contact gender from `stg_finalsite__contacts.gender` wherever gender is present, else fall back to the non-gendered value or `None`.

| Feed value | Rows | Maps to | Basis |
| --- | --- | --- | --- |
| `Grandmother`, `Aunt`, `Uncle`, `Grandfather` | 686 | pass through | already valid |
| `parent` | 2,261 | `Mother` 1,089 / `Father` 759 / `Parent` 413 | gender, 82% populated |
| `grandparent` | 178 | `Grandmother` 37 / `Grandfather` 141 | gender, 100% populated |
| `aunt/uncle` | 70 | `Aunt` 61 / `Uncle` 9 | gender, 100% populated |
| `stepparent` | 35 | `Stepmother` 7 / `Stepfather` 28 | gender, 100% populated |
| `guardian` | 26 | `Guardian` | direct, not gendered |
| `Great Uncle`, `Great Aunt` | 7 | `Uncle`, `Aunt` | label |
| `relative` | 86 | `None` | no nearest role |
| `Other`, `Sister`, `Brother`, `Cousin`, `Neighbor`, `Step-Sister`, null | 321 | `None` | no Focus equivalent |

`parent` is the only bucket where gender is not fully populated, so it is the only one whose fallback fires. Its 413 gender-null rows land on `Parent`, matching how Focus already stores 286 rows today. The other three gendered rollups resolve completely.

Rejected: the doc's flat alternative, which sends `grandparent`, `aunt/uncle`, and `relative` all to `None` — 334 rows losing role information that gender resolves cleanly for two of those three buckets.

Worth deciding separately from this work: whether the 12 already-leaked rows get corrected in Focus, since the import-once design means they will never be overwritten by a corrected feed.

### 4. Contact custody, resides-with, emergency, and pickup rules

`rpt_focus__contacts.sql` hardcodes all four to null for guardians. Only the emergency-contact branch sets them, from the Finalsite `emrg_N_*_yn` custom fields. Every parent and guardian therefore imports with blank custody, resides-with, and pickup.

Per decision G, follow the doc:

- `emergency`: `Y` for every row.
- `pick_up`: `Y` for every row.
- `resides_with_stud` and `custody`: identical logic. The first contact per student by `sort_order` is `Y`. A later contact is `Y` only when it shares a household with the first contact, `N` when household membership is unknown on either side.

Use household membership, not the doc's address string equality. The doc compares a `concat` of address, address2, city, state, and zipcode, lowercased and trimmed — so `123 Main St` versus `123 Main Street`, or a unit number that sits in `address` on one row and `address2` on the other, both read as `N`. Household ID is already available three ways:

- `stg_finalsite__contacts.household_ids` — array of every household id for a contact
- `stg_finalsite__contact_relationships.household_1_id` — first household only, so not sufficient alone for a multi-household contact
- `int_finalsite__contacts__households` — one row per contact-household, the right grain for a set intersection

Sharing a household with the student is already an established ranking signal here; `src/dbt/finalsite/CLAUDE.md` documents the contact_1 through contact_N pick using exactly that.

Per decision N, do not add the doc's `priority` column. It is `row_number()` recomputed after the doc drops phoneless contacts, purely to re-close the resulting gaps. Decision D drops no rows, so it would equal the existing `sort_order` on every row.

### 5. Callout and SMS flags

Per decision D these apply per slot across all 7 slots, not slot 1 only. The doc states them for slot 1 because its design collapsed to a single phone; that collapse is rejected.

Rules:

- `callout`: `Y` on every slot (decision G).
- `sms`: `Y`, except `N` where the slot type is Work Phone or Workplace.
- `blocked` and `unlisted` stay null.

Column work:

- Add `contact7_callout` (decision L). The select currently ends at `contact7_unlisted`, leaving 20 hardcoded-null flag columns rather than 21. Treated as a truncation bug; worth a glance at the Focus CONTACTS column contract when implementing.
- Add `contactN_sms` for all 7 slots (decision M). No such column exists today, so this widens the CSV contract by 7 columns.

Known consequence of blanket `callout`: `int_finalsite__contact_custom_attributes` carries `emrg_N_phone_N_opt_in`, a real per-phone consent boolean on the emergency branch. A blanket `Y` overwrites it. Guardian phones have no equivalent field — the Finalsite phone object is only `phone_type` and `number` — so honoring consent was only ever possible on one of the two branches. Accepted per decision G.

### 6. Reschedule the SFTP extract

`src/teamster/code_locations/kippmiami/extracts/schedules.py` runs one `ScheduleDefinition` at `45 12 * * *`. Replace with two on the same job, at `15 13 * * *` and `45 3 * * *`.

The 25-line comment justifying 12:45 argues entirely from the noon Finalsite push and the 2pm stakeholder commitment. It says nothing about an overnight run and needs rewriting, not just retiming.

### 7. Phone handling in the contacts feed

Much smaller than first scoped. Normalization is already done everywhere:

- **Guardian phones** — `stg_finalsite__contacts` wraps `phone_1`, `phone_2`, `phone_3` in `clean_phone` (lines 59-61). Those are the only three phone objects the Finalsite contact source has.
- **Emergency phones** — `int_finalsite__contact_custom_attributes` wraps all 12 `emrg_N_phone_N_number` columns in `clean_phone` (lines 239-250), plus `additional_phone_number_1`, `med_doctor`, and `med_hospital`. 15 call sites total.

So decisions E and F need no code. Per decision K there is no dashed-format conversion either. And per decision D the doc's `phone_rank`, `phone_choice`, and `phone_picked` layers are all dropped.

That leaves two items, both in `rpt_focus__contacts.sql`:

1. **Repeated-digit junk rejection** (decision Q). `5555555555` is NANP-valid and survives `clean_phone` as `+15555555555`. Null it in the Focus extract. Do NOT add the guard to the macro: `clean_phone` documents "never returns NULL for a non-null input", so a guard inside its `CASE` would fall through to `coalesce(..., degarbled)` and emit raw `5555555555` — un-normalizing junk instead of rejecting it — and changing that contract would alter values for `rpt_parentsquare__parents`, `rpt_deanslist__family_contacts`, and `int_students__contacts`.
1. **Phone type mapping** (decision J). Default a blank or unrecognized type to Cell Phone rather than dropping the contact, with an inline comment recording that it is deliberate and that it makes an untyped number an SMS target under subtask 5's rule.

On the type vocabulary: `stg_focus__people_join_contacts` already regex-maps the free-typed Focus contact-detail title to `mobile`, `home`, `work`, `daytime`, and `src/dbt/focus/tests/focus_unmapped_phone_contact_titles.sql` warns on unmapped and ambiguous titles. The doc's hand-rolled CASE invents a second, title-cased vocabulary with no `daytime` and no email-title exclusion. If Focus needs title-cased strings on import, add a single display map on top of the existing `phone_type` rather than a parallel CASE.

Out of scope, noted so nobody fixes it here: `rpt_deanslist__family_contacts` strips phones to `[^0-9x]` with no leading-`1` strip and no length check, so every E.164 value — guardian and emergency alike — reaches DeansList as an 11-digit `1XXXXXXXXXX`. Fully pre-existing and unaffected by this work, since both phone sources were already normalized before it started.

### 8. Disable rpt_focus__linked_students

Per decision P: disable, do not delete. The model is not in `focus.yaml`, has never shipped, and the spec doc does not mention the `LINKED_STUDENTS` export at all.

Four files, both projects:

- `src/dbt/kipptaf/models/extracts/focus/rpt_focus__linked_students.sql` and its properties yml
- `src/dbt/kippmiami/models/extracts/focus/rpt_focus__linked_students.sql` (a thin wrapper selecting from the `kipptaf_extracts` source) and its properties yml

Disabling a model does NOT disable its tests — they stay in `nodes` and keep scanning the stale prod relation. The kipptaf properties yml carries a model-level `dbt_utils.unique_combination_of_columns`, two column-level `not_null` tests, and a `unit_tests` block. Every one needs `config: enabled: false` alongside the model config.

Also handle the source side: `src/dbt/kippmiami/models/extracts/sources.yml` declares `rpt_focus__linked_students` with a `meta.dagster.asset_key` pointing at the kipptaf asset. Disabling the kipptaf model removes that asset, so the source entry needs disabling too.

## Deferred

**Expand demographics and compliance fields.** A fifth goal in the doc with no detail attached. Out of scope here and deliberately not tracked as a separate issue; noted so the goal is not lost.

`rpt_focus__demographics` already emits roughly 40 columns, most of them `cast(null as string)` placeholders: `residence_county`, `contry_birth`, `resident_st`, `birth_loc`, `bdate_verif`, `immun_st`, `grde_enter_dist`, `msix_id`, `pmrn`, `internt_perm`, `act_perm`, `direct_perm`, `screen_perm`, `survey_perm`, `mckay_sch_attend`, `fhsaa_el3_ind`, `fhsaa_el3ch_ind`, `dt_home_lang_survey`, `casas_track`, `lcp_cont_stdt`. The doc's demo query maps only 8 columns to Focus custom field IDs. Picking this up needs a named field list first: which placeholders to populate, the Finalsite source for each, and the Focus custom field ID each maps to.

## Sequencing

All eight subtasks are ready to spec. No open decisions.

Subtasks 3, 4, 5, and 7 all rewrite `rpt_focus__contacts.sql` and the same output contract — do them together, not as independent changes. Subtasks 1, 2, 6, and 8 are independent of that file and of each other.

No shared upstream model changes in this work: `clean_phone`, `stg_finalsite__contacts`, and `int_finalsite__contact_custom_attributes` are all left alone, so no other extract's output moves.

Note that dbt Cloud CI does not build district projects, so a local `dbt build` against `src/dbt/kippmiami` is the only pre-merge validation for the kippmiami side (subtasks 1, 2, and the kippmiami half of 8).

cbini:
# Pull Request

## Summary & Motivation

...implement all seven changes to the KIPP Miami Focus SFTP import feed specified in #4769.

| # | Change | Where |
| --- | --- | --- |
| 1 | Exclude withdrawals from student enrollment entirely — not the exit rows, not the `end_date`/`drop_code` columns, not the withdrawn students | kippmiami |
| 2 | Add an unstaffed 03:45 delivery alongside the staffed one, retimed 12:45 to 13:15 — two crons on ONE schedule | Dagster |
| 3 | Disable the never-shipped `rpt_focus__linked_students` model, its tests, and its source entry | both |
| 4 | Crosswalk `student_relation` to the 13-value Focus vocabulary, split by contact gender, with an `accepted_values` gate | kipptaf |
| 5 | Populate `resides_with_stud`, `custody`, `emergency`, `pickup` — household-derived rather than address-string-derived | kipptaf |
| 6 | Reject repeated-digit junk phone numbers and map phone types to the Focus display vocabulary | kipptaf |
| 7 | Add callout and SMS flags across all seven contact slots, widening the CSV contract from 50 to 58 columns | both + Dagster |

Subtask 2 of the issue (presence filters) turned out to need **no code** — all four anti-joins already implemented exactly what decisions A, B, and H landed on. Verified rather than assumed; recorded in the issue.

Three changes went beyond the issue's literal text, each for a reason worth reviewing:

- **Withdrawn students are excluded outright, not just their exit columns.** Stripping `end_date` while keeping the row would have imported a withdrawn student as an *active* enrollment with no end date. The entry branch now filters `where d.end_date is null`.
- **`custody` and `resides_with_stud` use household membership, not the spec doc's concatenated-address comparison.** `123 Main St` versus `123 Main Street`, or a unit number in `address` on one row and `address2` on the other, would both have read as a false `N`.
- **Callout and SMS are null on empty slots, and honor the recorded phone opt-out.** 210 of 530 in-scope emergency contacts have `emrg_N_phone_1_opt_in` explicitly `false`. A blanket `Y` would have overwritten every one of them permanently, because this feed never overwrites a contact twice.

### Before you merge — three operational items

1. **Deploy kipptaf before kippmiami.** The kippmiami wrapper reads `kipptaf_extracts`, a source with no target-conditional schema, so it always resolves to real production. Production does not have the 8 new columns until the kipptaf side deploys and materializes. Until then `dbt build` on `kippmiami.rpt_focus__contacts` fails with `Unrecognized name: contact1_sms` — expected, and the only reason the local kippmiami build is not green.
2. **Start `kippmiami__extracts__focus__asset_job_schedule` by hand in the Dagster+ UI if it is not already running.** `ScheduleDefinition` defaults to `STOPPED`, and `kippmiami/CLAUDE.md` records that this schedule had never run in prod as of #4736. Both deliveries (13:15 and 03:45) are crons on that ONE schedule, so enabling it enables both — there is no second schedule object to start. The name is pinned to its existing job-derived default deliberately: renaming it would mint a new stopped schedule and abandon this one's status and tick history.
3. **124 students who withdrew before ever reaching Focus will never be imported.** Focus will hold no record they were enrolled. Worth a Florida state-reporting check before merge. The enrollment feed currently emits 0 rows, which is correct: all 1,535 currently-enrolled 2026 students are already in Focus, so those 124 were the only remaining candidates.

### Verification

- **kipptaf** `dbt build --select path:models/extracts/focus`: **PASS=34, WARN=1, ERROR=0**. The warn is a pre-existing `not_null_rpt_focus__contacts_last_name` on 4 rows; confirmed pre-existing, this diff never touches `last_name` derivation.
- **kippmiami** `dbt build --select path:models/extracts/focus`: **PASS=12, ERROR=1** — the expected `contact1_sms` failure from item 1 above.
- **Trunk** `check --force --no-fix` over all 17 changed files: **no new issues** (2 pre-existing, non-blocking).
- **Column contract**: 58 columns agree across the kipptaf model, kipptaf YAML, kippmiami passthrough, kippmiami YAML, and `focus.yaml` `header_replacements`.
- **Not run**: `dagster definitions validate` for kippmiami. It needs the Focus DB connection credential, which is not present in a Codespace — a documented limitation, not a regression. Evidence it is environmental: the same test fails identically for kippcamden, kippnewark, and kipppaterson, three locations this branch does not touch. The schedule change was instead verified by asserting both schedule names and cron strings via direct import, and by checking the pinned name string against the job-name derivation.

Data checks against the dev relation, all confirming intent: no phone type on a slot with no number (0 rows, all slots), no callout or SMS flag on a slot with no number (0 rows, all 7 slots), `student_relation` closed to the 13-value domain with no lowercase value surviving, and `sms = 'N'` on 461 rows — the 444 opt-outs plus 17 Work Phone rows, which are disjoint sets.

## AI Assistance

Human-directed: every design decision. The issue records 17 lettered decisions (A through Q) plus three more taken during execution — the overnight cron placement, honoring the phone opt-out, and dropping an inert test. AI-assisted: the implementation plan, the code, and the review passes.

Worth knowing for reviewers: the plan document itself was the main source of defects, not the implementations. Six plan defects were caught and corrected during execution — an inert data test, two wrong figures in a schedule comment, a phone-type default that fired on empty slots, unconditional flags on empty slots, and two verification steps that read production instead of the dev build. The final review then caught two Critical defects in the ten-line Dagster change that three earlier review passes had missed.

## Self-review

### General

### Dagster

- [ ] `uv run dagster definitions validate` — **could not run**, missing Focus DB credential in Codespace. See Verification above for what was checked instead.
- [ ] `uv run pytest tests/test_dagster_definitions.py` — **fails for the same reason**, including for three code locations this branch does not touch.
- [x] No new integration; existing Library + Config pattern unchanged.

### dbt

- [x] Properties files updated for every changed model, including the 8 new contract columns in both projects.
- [x] No exposure changes — this feed is an SFTP extract, not a dashboard.
- [x] **Breaking change?** Yes, deliberately, and coordinated in #4769: `end_date` and `drop_code` are removed from `rpt_focus__student_enrollment`, and 8 columns are added to `rpt_focus__contacts`. Both models are consumed only by the Focus SFTP extract in this repo. Rollback plan: revert this PR; the feed is import-once, so already-delivered rows are unaffected, and no already-imported record is overwritten by either the change or its revert.
- [x] No external source added or modified, so no `stage_external_sources` run needed.

### Docs

- [ ] Regenerate the automations catalog (`uv run scripts/gen-automations-doc.py`) — **not run**: the script silently skips code locations that fail to import, so running it in a Codespace would drop locations from the catalog. Needs a full environment.
- [x] Updated the published ops-facing reference `docs/reference/finalsite-focus-import.md` for the retime, the withdrawal change, and the new contact-flag behavior. It had an instruction telling enrollment staff to record a last-attended date on withdrawal, which now causes the opposite of the intended result.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes. Note it builds `kipptaf` only; the kippmiami half of this change gets no CI coverage.

Closes #4769


src/teamster/code_locations/kippmiami/extracts/schedules.py
cbini, line 1: make this a single schedule

src/dbt/kipptaf/macros/focus_phone_type.sql
cbini, line 1: is extracting this as a macro justified?

src/teamster/code_locations/kippmiami/extracts/schedules.py
cbini: Done in 40db1c6f9 — one `ScheduleDefinition` with `cron_schedule=["15 13 * * *", "45 3 * * *"]`.

This fixed more than the duplication. Two definitions on the same job meant a second Dagster+ schedule object carrying its own status and tick history, which would have needed enabling by hand after deploy. Collapsing keeps the existing `kippmiami__extracts__focus__asset_job_schedule` identity, so there is nothing new to start — one less post-deploy step.

`finalsite/schedules.py` already used a cron list on a single definition. I should have followed that idiom rather than inventing a second object.

src/dbt/kipptaf/macros/focus_phone_type.sql
cbini: Checked the precedent rather than argue for my own choice, and it holds — but narrowly, so your call.

`src/dbt/kipptaf/macros/parse_html.sql` is a single macro in its own file, a single-column transform, with 4 call sites in one model. `focus_phone_type` is structurally identical at 3 call sites. So a single-purpose macro file consumed by one model is already the local pattern, not a new one.

The alternative is three inline copies of the same 8-line `CASE` in the final select, which the review rubric this repo uses treats as verbatim duplication of a logic block.

Two things that would change my answer, neither of which applies here: if it had only one call site (then it is pure indirection), or if the vocabulary belonged upstream (it does not — `stg_focus__people_join_contacts` maps Focus-side free-typed titles, not Finalsite phone types, so there is nothing to reuse).

If you would rather single-model logic stay in the model regardless of call count, say so and I will inline it and drop the file.

cbini: Inlined and the macro file is deleted, in b50715660. Your call stands over my precedent argument.

It came out better than a wash. The macro forced a `CASE` nested inside the null-gate `CASE`, because the gate had to wrap the macro call. Each slot is now one flat `CASE` with the null check as its first branch — fewer levels, and the whole mapping is readable where it is used.

Verified behavior-unchanged rather than assumed: all 34 focus tests pass, including the unit tests that assert exact phone-type outputs on fixed fixtures (immune to the source drift that makes raw distributions wander), and the type/value null-pairing invariant is still 0 orphans on every slot.

One leftover I deliberately did not touch: `docs/superpowers/plans/2026-08-13-focus-import-v2.md` still describes creating the macro. That is a working plan document excluded from the mkdocs nav, and a superseded plan is expected to read as a snapshot of what was planned. Say the word if you would rather I amend it.

cbini:
Verdict per finding. Each claim was checked against the code before acting; two are already resolved by @cbini's inline review, which landed while this review was running.

## Important

**1. No PII tagging in the Focus contacts feed — CONFIRMED, deferred to @cbini.**

Verified both halves: `contains_pii` appears zero times under `src/dbt/kipptaf/models/extracts/focus/`, and `rpt_parentsquare__emergency_contacts.yml` does tag four columns (lines 61, 69, 81, 89) over the same Finalsite contact/phone/email data. The precedent and the gap are both real.

Not fixing this unilaterally. Root `CLAUDE.md` says for a PII-heavy model to "confirm scope (direct-only vs direct+indirect) with the user before tagging," and the choice is consequential: direct-only would tag names, email, address, and the phone values; direct+indirect would also pull in `student_relation`, which the same reference lists as linkable. That is a policy call, not a code call. Flagged for @cbini with both options.

Note the two new flag columns are not PII on any reading — `contactN_callout` and `contactN_sms` are `Y`/`N` derived values carrying no identifier.

**2. Overnight schedule ships STOPPED and undocumented — ALREADY RESOLVED, by a different route.**

@cbini's inline review asked for a single schedule, so there is no separate overnight schedule object anymore. `40db1c6f9` collapsed both deliveries onto `kippmiami__extracts__focus__asset_job_schedule` as two crons. That schedule's start-by-hand requirement was already documented in both the code comment and `kippmiami/CLAUDE.md`, and I added an explicit line that enabling that one schedule enables both crons. The finding was correct against the commit under review; the fix removed the object it described.

## Minor

**1. PR body says 04:30, code says 03:45 — CONFIRMED, FIXED.** PR description updated. Your history trace is exactly right: `25ef6658c` moved it to 04:30 to trail the old 04:00 dlt pull, then the `origin/main` merge brought in the probe-gated refactor that made that rationale obsolete, and it reverted to 03:45. Good catch that the description had drifted from the code.

**2. `CASE` nested inside `CASE` via the macro — ALREADY RESOLVED, same reasoning.** `b50715660` inlined the mapping and deleted `focus_phone_type.sql`, after @cbini questioned whether the macro was justified. Each slot is now one flat `CASE` with the null check as its first branch. You and @cbini landed on the same objection independently.

**3. Stale `drop_code` in 3 kippmiami fixtures — CONFIRMED, FIXED** in `c1fd1fe8d`. Checked before removing: the kipptaf model's only surviving `drop_code` is inside a comment, and its contract YAML has none, so the fixtures were genuinely mocking a column the upstream stopped emitting. All 5 kippmiami enrollment tests still pass.

**4. `+18888888888` is a valid toll-free NPA — DECLINED.** Correct that 888 is an assigned toll-free NPA, so the number is technically reachable. Keeping it in the reject list anyway: the list exists to catch repeated-digit placeholder data in a population of personal guardian contacts, where a toll-free vanity line is implausible. The trade is one contact losing one phone slot in a case I have not observed, against shipping placeholder digits that Focus will actually dial. If a real one ever appears the fix is one literal.

## Needs local verification

**Gender-split coverage — RESOLVED with data.** I have warehouse access, so I ran your suggested query scoped to the six guardian `rel_type` values. Only three values exist:

| `gender` | guardian contacts |
| --- | --- |
| `(null)` | 2,305 |
| `F` | 1,784 |
| `M` | 1,303 |

No lowercase, no alternate casings, no `Female`/`Male` spellings. Nothing silently falls through to a non-gendered label from an unmatched casing; the 2,305 nulls fall through by design. The `Female`/`Male` branches are dead in current data and retained deliberately to match the pattern `rpt_focus__demographics` already uses on the same column.

**The 124 pre-Focus withdrawals — still open**, correctly deferred. It stays in the PR's before-you-merge list as a Florida state-reporting question for a human.

cbini:
Update on Important 1 (PII tagging) — now closed in `37789eea6`.

@cbini chose direct-identifiers-only scope. Tagged 16 columns in both the kipptaf model and the kippmiami passthrough: `first_name`, `middle_name`, `last_name`, `email`, the address block (`address`, `address2`, `city`, `state`, `zipcode`), and all seven `contactN_value` phone slots.

Deliberately NOT tagged: `student_id` (matching the `rpt_parentsquare__emergency_contacts` precedent, which leaves it untagged, and the root `CLAUDE.md` note that this repo tags more narrowly than the full FERPA list), `student_relation`, and the derived `contactN_type` / `contactN_callout` / `contactN_sms` columns.

Slots 4 through 7 are tagged even though no branch populates them today. That is a schema statement rather than a data one — they are phone-value columns by contract, and leaving them untagged would be a latent gap the moment anything fills them.

Declared on both projects because `config.meta.contains_pii` does not travel through `source()` — `src/dbt/kipptaf/CLAUDE.md` requires a wrapper to re-declare it. Verified before committing that `contains_pii` is consumed only by three documentation and propagation scripts, not by any runtime masking or access path, so this is metadata with no behavior change.

kipptaf builds clean (16 pass, 1 pre-existing `last_name` warn), kippmiami parses, trunk clean on both files.

PR #4378, opened 2026-07-10
feat(deanslist): add deanslist to paterson and union into kipptaf

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make Paterson DeansList data queryable
network-wide by wiring the `deanslist` dbt package into `kipppaterson` and
unioning Paterson into the kipptaf cross-district DeansList models. This is the
dbt half of #4367 (the Dagster ingestion half landed in #4372, #4374, and
#4376).

**kipppaterson district project**

- `packages.yml`: import the local `deanslist` package.
- `dbt_project.yml`: materialize `deanslist` as tables; disable the
  `stg_deanslist__dff_stats`, `stg_deanslist__followups`,
  `stg_deanslist__homework`, and `stg_deanslist__lists` staging models and their
  `src_deanslist__*` sources — Paterson does not pull those endpoints, so no
  Avro exists for them (leaving them enabled would fail `stage_external_sources`
  and the staging build).
- Paterson pulls `behavior`, `comm_log`, `incidents`, `roster_assignments`,
  `rosters`, `students`, `terms`, and `users`.

**kipptaf cross-district unions**

- New `sources-kipppaterson.yml` (mirrors the other three districts) with the
  `dev` / `staging` (`zz_stg_`) / prod schema branch so the single-PR
  cross-project CI can read the staged Paterson tables.
- Added `source("kipppaterson_deanslist", ...)` to all 9 kipptaf `deanslist/api`
  union models: `stg_deanslist__{users,behavior,terms}` and
  `int_deanslist__{roster_assignments, incidents__attachments,
  students__custom_fields__pivot, incidents__penalties, incidents, comm_log}`.

`behavior` is included in the union because Paterson `behavior` ingestion was
added in #4376 (it was missing from the other districts' parity — a paginated
asset the other three ingest). This PR depends on #4376 being deployed and
Paterson `behavior` materialized.

> Note: after #4376 deployed and backfilled, Paterson's `behavior` endpoint
> returns **0 records** (the Avro files exist but are empty). The ingestion +
> staging pipeline is correct and verified end-to-end; Paterson simply has no
> DeansList behavior/points data logged yet. `behavior` is kept in the union for
> structural parity and forward-compatibility — it will populate automatically
> once data exists. Whether Paterson should have behavior data is a
> DeansList-side question, not a code issue.

## AI Assistance

AI-assisted (Claude Code), human-directed. Claude executed the plan at
[...]

PR #4864, opened 2026-08-14
feat(kipppaterson): add titan person_data asset, sensor, and dbt wiring

cbini:
# Pull Request

## Summary & Motivation

Give Paterson the Titan (school meals) ingest that Newark and Camden already
have, and make Titan authoritative for Paterson meal eligibility.

Dagster:

- New `kipppaterson/titan/person_data` SFTP asset, static-partitioned by fiscal
  year from `2025` (Paterson's first year), matching the Newark/Camden factory
  call
- New `kipppaterson__titan__sftp_asset_sensor` (10-min interval)
- `ssh_titan` resource wired in `definitions.py`; `TITAN_SFTP_HOST` /
  `TITAN_SFTP_USERNAME` / `TITAN_SFTP_PASSWORD` added to both the
  `container_context` and `run_k8s_config` blocks of `dagster-cloud.yaml`,
  sourced from a new `op-titan-sftp-kipppaterson` secret

dbt:

- `kipppaterson` picks up the local `titan` package (`stg_titan__person_data`
  enabled, `stg_titan__income_form_data` disabled, same as the other NJ
  regions)
- New `sources-kipppaterson.yml` in kipptaf, and `kipppaterson_titan` joins the
  `stg_titan__person_data` union
- `dim_student_meal_eligibility_status`: Paterson moves from the PowerSchool
  `lunchstatus` leg to the Titan leg. That leg is now Miami only, so `pm_leg`
  is renamed `miami_leg`. Without this, Paterson students would land in both
  legs and produce duplicate dim rows.

`dim_student_enrollment_status` already treats Paterson as NJ for meal
selection (Oct-15 span logic), so no change was needed there — that logic
starts working as intended once Titan supplies real spans.

Two prerequisites before this can build, both outside the code:

1. The `op-titan-sftp-kipppaterson` 1Password item and its k8s secret must
   exist, with `host` / `username` / `password` keys.
2. `src_titan__person_data` for `kipppaterson` is a brand-new external source.
   It cannot be staged until the asset has materialized at least once, so per
   the repo's external-source flow: materialize
   `kipppaterson/titan/person_data` in this PR's branch deployment, then run
   `stage_external_sources --target staging` with the
   `gs://teamster-test/dagster/kipppaterson` URI override. Until that happens
   the kipptaf union model reads a table that does not exist yet.

## AI Assistance

Authored by Claude Code (Opus 5), human-directed. The meal-dim source decision
(Titan authoritative for Paterson vs. leaving the mart alone) was made by the
requester, not inferred.

### Dagster

- [x] `kipppaterson.definitions` imports clean in the worktree; the asset graph
      resolves `kipppaterson/titan/person_data` plus the dbt-derived
      `kipppaterson/titan/stg_titan__person_data`, and the sensor registers as
      `kipppaterson__titan__sftp_asset_sensor`
- [x] New integration follows the Library + Config pattern with the standard
      asset key format and Avro IO manager
- [ ] `pytest tests/test_dagster_definitions.py` not run (needs all five dbt
      manifests plus credentials this environment lacks)

### dbt

- [x] `dbt parse` clean for `kipppaterson` and `kipptaf`
- [x] `dbt compile --select stg_titan__person_data` succeeds
- [x] No new models without a properties file — `stg_titan__person_data` is an
      existing package model; only its source list changed
- [ ] `stage_external_sources --target staging` — required, see prerequisite 2
      above
- [x] No contracted column renames or removals

### Docs

- [x] `docs/reference/automations.md` updated with the new sensor. Added by
      hand rather than by `scripts/gen-automations-doc.py`: the generator
      silently drops any code location it cannot load, and `kipptaf` and
      `kippmiami` fail to load in a codespace, so a full regeneration would
      have deleted both sections.
- [x] `kipppaterson` CLAUDE.md files updated (Dagster and dbt)

## CI checks

- [x] **Trunk** — `trunk check --force` clean on all changed files
- [ ] **dbt Cloud** — will fail until prerequisite 2 is done
- [ ] **Dagster Cloud** — pending


cbini:
Verdicts on the review, one per finding. Citations checked against the code.

### Important 1 — shared-model blast radius reaches Newark and Camden

**Accepted, no code change.** Verified: `stg_titan__person_data` is one model
unioning all three regions, so it is stale-or-failing for all three until
`kipppaterson_titan.stg_titan__person_data` exists in prod. The review is right
that this is wider than the PR body said.

It is not fixable pre-merge. CI reads district staging tables from prod, unprefixed
(`src/dbt/CLAUDE.md` "Source Schema Resolution"), and only Dagster builds the
district projects — so the Paterson table cannot exist until after this merges.
This is how the same shape of change has shipped before: PR #4378 (deanslist for
Paterson, unioned into kipptaf) merged 2026-07-10 with dbt Cloud red for exactly
this reason.

Mitigation is sequencing, and the window is minutes, not days: materialize
`kipppaterson/titan/person_data` in prod immediately after merge, let the
`kipppaterson/titan/stg_titan__person_data` dbt asset stage its external source
and build, then confirm the kipptaf union rebuilds green for all three regions
before calling the rollout done.

If a zero-risk window is preferred over a short one, the alternative is to split
this into two PRs — Dagster plus the Paterson dbt package first, then the kipptaf
union and meal-dim change once Paterson's table is live in prod. Say the word and
I will restructure it.

### Minor 1 — missing FreshnessPolicy parity

**Fixed** in b259f38. The finding is correct and load-bearing: `freshness.py` in
both kippnewark and kippcamden exists *only* for Titan (2 asset keys each, 16
lines), so Paterson's "no freshness checks" was a consequence of having no Titan,
not a deliberate choice. Added `kipppaterson/freshness.py` mirroring the sibling
locations, wrapped `load_assets_from_modules` in `apply_freshness_policies`, and
updated the code location's CLAUDE.md.

Verified by resolving the asset graph: both `kipppaterson/titan/person_data` and
`kipppaterson/titan/stg_titan__person_data` now carry
`CronFreshnessPolicy(deadline_cron='0 1 * * *', lower_bound_delta=1h,
timezone='America/New_York')`.

### Informational 1 — wall-clock sensor cursor

**Acknowledged, out of scope.** Correct that
`libraries/titan/sensors.py` sets the cursor to `now_timestamp` rather than max
matched-file mtime, and correct that it is pre-existing and shared with
Newark/Camden. This PR adds a third caller of the unmodified factory. Not
changing it here — a fix belongs in its own PR so the behavior change is
reviewable against all three regions at once.

src/teamster/code_locations/kipppaterson/titan/assets.py
```diff
@@ -0,0 +1,15 @@
+from teamster.code_locations.kipppaterson import CODE_LOCATION, CURRENT_FISCAL_YEAR
+from teamster.code_locations.kipppaterson.titan.schema import PERSON_DATA_SCHEMA
+from teamster.libraries.titan.assets import build_titan_sftp_asset
+
+person_data = build_titan_sftp_asset(
+    asset_key=[CODE_LOCATION, "titan", "person_data"],
+    remote_file_regex=r"[Pp]erson\s?[Dd]ata(?P<fiscal_year>\d{4})\.csv",
+    schema=PERSON_DATA_SCHEMA,
+    partition_start_date="2025-07-01",
```
cbini: this should be 2026-07-01

src/teamster/code_locations/kipppaterson/titan/sensors.py
```diff
@@ -0,0 +1,15 @@
+from teamster.code_locations.kipppaterson import CODE_LOCATION, LOCAL_TIMEZONE
+from teamster.code_locations.kipppaterson.titan.assets import person_data
+from teamster.libraries.titan.sensors import build_titan_sftp_sensor
+
+titan_sftp_sensor = build_titan_sftp_sensor(
+    code_location=CODE_LOCATION,
+    asset_selection=[person_data],
+    timezone=LOCAL_TIMEZONE,
+    minimum_interval_seconds=(60 * 10),
+    exclude_dirs=["Scipt", "Script"],
```
cbini: "Scipt" is a one-off typo on another instance

src/teamster/code_locations/kipppaterson/titan/assets.py
cbini: Changed to `2026-07-01` in 06a53ab. Partition keys now resolve to `['2026']` — one key, FY2027, matching `PersonData2026.csv`. (I had inferred 2025 from the DeansList start date, which was wrong for Titan.)

src/teamster/code_locations/kipppaterson/titan/sensors.py
cbini: Dropped, sensor now excludes `Script` only.

Note the asset side still carries both spellings: `build_titan_sftp_asset` in `libraries/titan/assets.py` hardcodes `exclude_dirs=["Script", "Scipt"]` for every region. Left alone since it is shared with Newark and Camden — worth a separate cleanup if the typo directory only ever existed on the one instance.

PR #4870, opened 2026-08-14
refactor(focus): drop unused login_history table from dlt sync

### LINKED ISSUE #4869: Remove unused Focus login_history table from the dlt sync and staging
`kippmiami/dlt/focus/login_history` earns no place in the warehouse:

- `stg_focus__login_history` is a leaf — no `ref()`s in dbt, nothing in Cube, nothing in the published docs.
- The Focus ERD names it once in a bare "Related:" list under Users, with no PK/FK/join path documented.
- It is an authentication audit log, not an analytics entity: `username`, `ip_address`, `user_agent`, and `object_id` are all tagged `contains_pii`. Focus's own UI is the system of record.
- It has `cursor_column: null`, so it can never sync incrementally, and its row count moves on nearly every 15-minute probe — it wakes the intraday sensor and pulls a full-table `replace` more often than any other Focus table (6,252 rows / 1.59 MB, so the cost is run frequency, not bytes).

Scope: drop the `config/focus.yaml` entry (77 to 76 tables), the `sources-bigquery.yml` source, and the two `stg_focus__login_history` files; sweep the count-only-tier docs and comments so `co_teachers` is the only count-only table; rename the test fixture. Dropping the existing BigQuery table and staging relation is manual DDL, handed off separately.

cbini:
## Summary and Motivation

When merged, this pull request will stop syncing `kippmiami/dlt/focus/login_history` into the warehouse and delete its staging model.

The table earns no place here:

- `stg_focus__login_history` was a leaf. No `ref()`s in dbt, nothing in Cube, nothing in the published docs.
- The Focus ERD names it once in a bare Related list under Users, with no PK/FK/join path documented.
- It is a sign-in audit log, not an analytics entity. `username`, `ip_address`, `user_agent`, and `object_id` are all tagged `contains_pii`, and Focus's own UI is the system of record for that data.
- It cost the most to keep. `cursor_column: null` means it can never sync incrementally, and its row count moves on nearly every 15-minute probe, so it woke the intraday sensor and pulled a full-table `replace` more often than any other Focus table. 6,252 rows / 1.59 MB, so the cost was run frequency, not bytes.

Changes:

- drops the `config/focus.yaml` entry (77 to 76 tables), the `sources-bigquery.yml` source, and `stg_focus__login_history.sql` plus its properties file
- sweeps the count-only-tier docs and comments so `co_teachers` is the only count-only table. The 04:00 schedule keeps its job.
- makes `test_schedule_targets_only_count_only_tables` handle a single-key target: Dagster returns a bare `KeysAssetSelection` for one asset key and a union with `.operands` for several, so the test broke on the tier shrinking to one table
- renames the `test_dlt_focus_sensors.py` count-only fixture to `co_teachers`, which still exists

Follow-up handed off separately: dropping the two live BigQuery relations, `dagster_kippmiami_dlt_focus.login_history` and `kippmiami_focus.stg_focus__login_history`. That is manual DDL.

## AI Assistance

Claude Code did the investigation, the removal, and the doc sweep. Human-directed: the question of whether the table earns its place, and the decision to remove it.

### General

- [x] Review the Claude Code Review comment posted on this PR

### Dagster

- [ ] `uv run dagster definitions validate -m teamster.code_locations.kippmiami.definitions` cannot run in the Codespace. It fails identically on `main` with `ConfigFieldMissingException: Missing 1 field(s) in ConnectionStringCredentials: drivername` because the Focus connection string is not available locally. Relying on the Dagster Cloud deploy check instead.
- [x] `uv run pytest tests/libraries -k focus` passes: 36 passed. `tests/test_dagster_definitions.py` fails for the same pre-existing reason as above, plus missing dbt manifests for the untouched locations.
- [x] No new integration. Asset key format unchanged.

### dbt

- [x] No new models. Two files deleted, and `dbt parse` is clean for both `kippmiami` and `kipptaf`.
- [x] No exposure referenced this model, which is why it is being removed.
- [x] Breaking change check: `stg_focus__login_history` had zero downstream consumers, so removing it breaks nothing. Rollback is a revert of this PR plus one manual Focus asset job launch to reload the table.
- [x] Not an external source.

### Docs

- [x] No schedule or sensor was added, renamed, or re-croned. The 04:00 schedule's target set shrinks by one asset, and the Focus dlt schedule and intraday sensor are not listed in `docs/reference/automations.md`, so the catalog does not drift.

## CI checks

- [x] Trunk passes locally: `trunk check --force` on all 8 changed files reports no issues, and the commit hook formatted the three markdown files.
- [ ] dbt Cloud
- [ ] Dagster Cloud

Refs #4869


cbini:
Review response: no issues were raised, so there is nothing to fix. One item to close out.

**Recommendation (informational): the selection-shape claim in the new test comment was unverified in the review environment.** Confirmed here against the installed Dagster, by constructing a `ScheduleDefinition` with one target and with two and inspecting `target.resolvable_to_job.selection`:

```text
1 key : ('KeysAssetSelection', has_operands=False, has_selected_keys=True)
2 keys: ('OrAssetSelection',   has_operands=True,  has_selected_keys=False)
```

So the comment is accurate as written, and the two branches are genuinely disjoint: a multi-key selection carries `.operands` but no `.selected_keys`, and a single-key selection is the reverse. That is also why the old code failed rather than degraded when the tier shrank to one table. The `getattr(selection, "operands", (selection,))` fallback covers both, and still raises `AttributeError` if a future Dagster version returns a third shape.

Issue #4871, opened 2026-08-15
Clean up duplicate kippadb score records and resolve the strategy_case fan-out


GabyRangelB:
Two data-quality items found while reworking the CARAT lineage in #4658, both deliberately left alone there because neither is a modelling defect and both need decisions outside the data team. Recorded together because both distort assessment attempt counts.

Figures measured 2026-08-15 against production. No student identifiers here — record lists stay local.

## 1. Duplicate score records in kippadb

Counting distinct test dates rather than rows exposed real duplicates in Salesforce, not a modelling artifact. Row counts are identical at `stg_kippadb__standardized_test`, `int_kippadb__standardized_test` and the unpivot, and every duplicate carries its own Salesforce record id, so nothing in dbt is fanning them out.

Three populations, and they do **not** share a remedy:

| population       | redundant records | groups where scores are identical | groups where scores disagree |
| ---------------- | ----------------: | --------------------------------: | ---------------------------: |
| AP, 2011 to 2026 |             2,221 |                               669 |                      **835** |
| PSAT, 2024       |             1,434 |                             1,434 |                            0 |
| SAT, 2026        |               258 |                               258 |                            0 |

Grouped on student, score type and test date.

**PSAT 2024 and SAT 2026 are pure redundancy** — every duplicate pair carries an identical score, so they are a double import rather than two different results. Safe to delete the redundant record in each pair without a judgment call. The SAT set is one cohort on one administration: Camden class of 2027, the April 2026 school-day SAT.

**The AP set is not.** 835 of its 1,504 duplicate groups hold *different* scores for the same student, exam and date, so someone has to decide which record is correct. A bulk delete would silently pick one. This is the larger half of the problem and probably belongs to whoever owns the AP ingest, not to the same cleanup pass as the other two.

### Why it matters

Attempt counts are overstated wherever a consumer counts rows. Camden 2027 SAT attempts are currently high by 86, which inflates that cohort's 2-or-more-attempts rate against an 0.80 goal.

Scores themselves are unaffected in the two pure-redundancy populations — identical values, so maxes, superscores and benchmark attainment are all correct. `attempt_lifetime` is also immune, because #4658 moved it to counting distinct test dates rather than rows. Anything still counting rows is not.

### Asks

- [ ] Ops: delete the redundant record in each PSAT 2024 and SAT 2026 pair in Salesforce.
[...]

PR #4876, opened 2026-08-17
feat(dbt): pivot college assessments long subjects wide by administration

### LINKED ISSUE #4875: feat(dbt): pivot rpt_gsheets__college_assessments_long subjects wide by administration
KIPP Forward wants the College Assessments Long sheet to stay long by student and administration, but read wide by subject — one row per sitting with a score column per subject, instead of three rows per sitting.

Figures measured 2026-08-17 against production. No student identifiers here.

## Current shape

`rpt_gsheets__college_assessments_long` is 7,434 rows: 2,425 distinct student × `test_type` × `test_date` sittings, times 3 subjects, plus 159 duplicate pairs (see _Incidental findings_). 33 columns, 1,306 students.

`test_date` is the administration key. `administration_round` exists on the upstream hub but is a coarser `Apr 24`-style label spanning multiple dates, so it cannot key a row.

## Target shape

One row per student × `test_type` × `test_date`. 2,425 rows, 40 columns, same 1,306 students.

Three subject columns, not five. The existing trailing `score_type` exclusion filter already removes `act_english`, `act_science`, `psat10_reading`, `psat10_math_test`, `sat_reading_test_score` and `sat_math_test_score`, so exactly three subjects survive per test type — and `int_assessments__college_assessment.aligned_subject` already collapses them to `Total`, `EBRW/Reading`, `Math` across all five scopes. No new alignment logic is needed here.

Removed (5): `test_subject`, `scale_score`, `highest_score_by_test`, `hs_grad_ready`, `college_ready`.

Added (12), grouped by subject so the sheet reads in blocks:

| Total | EBRW/Reading | Math |
| --- | --- | --- |
| `total_scale_score` | `ebrw_reading_scale_score` | `math_scale_score` |
| `total_highest_score_by_test` | `ebrw_reading_highest_score_by_test` | `math_highest_score_by_test` |
| `total_hs_grad_ready` | `ebrw_reading_hs_grad_ready` | `math_hs_grad_ready` |
| `total_college_ready` | `ebrw_reading_college_ready` | `math_college_ready` |

The 26 roster and context columns through `sat_math_highest` are unchanged, as are `test_type` and `test_date`.

## Approach

Conditional aggregation, matching the pattern the `_wide` sibling already uses in the same directory — not the `PIVOT` operator. The readiness flags need a two-dimensional pivot (subject × metric name), because `goals_distinct` returns two rows per score type, and stacking `PIVOT` clauses for that is worse than being explicit:

```sql
max(if(aligned_subject = 'Total', scale_score, null)) as total_scale_score,
coalesce(
    max(if(aligned_subject = 'Total' and expected_metric_name = 'HS-Ready', met_minimum, null)),
    'NA'
) as total_hs_grad_ready,
```

The six-item `score_type` exclusion list is replaced by `where aligned_subject in ('Total', 'EBRW/Reading', 'Math')`. This is **exactly equivalent** on the current domain — verified across all 21 `(scope, score_type, aligned_subject)` combinations: every excluded score type maps to `Reading`, `Math Test`, `Reading Test`, `English` or `Science`, and every retained one maps to one of the three. It also makes the filter and the pivot use the same dimension.

One deliberate consequence: a new subject appearing upstream is silently excluded rather than surfacing. For a fixed-column extract that is the safe failure, but adding a subject later becomes a column change rather than a data change.

If ACT data returns, `act_reading` lands in the `ebrw_reading` column alongside SAT EBRW. ACT contributes zero rows today — kippadb ACT scores stop at 2024-10-26, so no currently enrolled HS student in scope has one.

## Incidental findings

**`hs_grad_ready` has never carried a value.** All 7,434 rows read `NA`. The pivot looks for `HS-Grad Ready`; `stg_google_sheets__kippfwd__goals` spells it `HS-Ready`. Fixed here, since the pivot is being rewritten anyway. The goals sheet has HS-Ready thresholds for all nine retained score types, so the column populates broadly — a visible data change for sheet users, called out in the PR body so nobody reads it as a regression.

**The 159 duplicate pairs are the kippadb duplicates already tracked in #4871.** They differ only on `highest_score_by_test`, carry identical scores, and are entirely Camden class of 2027 on the April 2026 school-day SAT — the exact population that issue describes. Nothing is fixed here; the pivot collapses them incidentally, and `max()` over `Yes`/`No` picks `Yes`, which is correct since the sitting is the student's highest if any duplicate row says so. This extract becomes immune to the source duplication; the source problem remains #4871's.

**Uniqueness test.** #4871 item 3 lists this model among three `rpt_` views with no uniqueness test, and proposed landing all three together once the `strategy_case` pick is settled. That reasoning applies to `rpt_tableau__college_assessment_dashboard_over_time`, which carries `strategy_case` in its grain — it does not apply here, since this model never references `strategy_case`. Adding `dbt_utils.unique_combination_of_columns` on `(student_number, test_type, test_date)` now closes this model's bullet and leaves the other two for #4871.

Left at the project default `warn` rather than `error`: it is a new invariant on a sheet extract, and the `ccr_course` join could in principle fan out if a student ever enrolled in two College and Career courses in one year. Zero cases today.

## Out of scope

**The 625 students with no qualifying score.** The current `score_type` filter sits after the LEFT JOIN, so a score-less student's null `score_type` evaluates to null and is filtered — the LEFT JOIN is an inner join by accident, excluding 625 of the 1,931 roster students. This change preserves that behaviour, just legibly. Surfacing those students as blank-score rows so advisors can see who has not tested is a separate ask; open a follow-up if KIPP Forward wants it.

The other two views in #4871 item 3, and the `strategy_case` decision.

## Files

- `src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__college_assessments_long.sql`
- `src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__college_assessments_long.yml`

The model name and the `google-sheets.yml` exposure are unchanged — the view is still long by administration, so no new asset key, exposure entry or sheet is needed. No published docs reference this model.

## Verification

`extracts/` materializes as a view, so a green `dbt build` proves nothing about values. Build into dev with `--defer --state`, then:

- [ ] Row count 2,425 and student count 1,306 in the built view
- [ ] Per-subject `scale_score` parity against the production long view for all 2,425 keys
- [ ] `hs_grad_ready` no longer uniformly `NA`
- [ ] Uniqueness test passes
- [ ] `trunk check --force` on both changed files

## After merge

The sheet owner has to refresh the tab at the exposure URL: the column set goes from 33 to 40 and the row count falls by roughly two thirds.

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will reshape `rpt_gsheets__college_assessments_long` so the College Assessments Long sheet stays long by student and administration but reads **wide by subject** — one row per sitting with a score and readiness column per subject, instead of three rows per sitting.

**7,434 rows to 2,425. 33 columns to 40. Same 1,306 students.**

Three subject columns, not five. The trailing `score_type` exclusion filter already removed `act_english`, `act_science`, `psat10_reading`, `psat10_math_test`, `sat_reading_test_score` and `sat_math_test_score`, so exactly three subjects survive per test type — and `int_assessments__college_assessment.aligned_subject` already collapses them to `Total` / `EBRW/Reading` / `Math` across all five scopes. No new alignment logic was needed. If ACT data returns, `act_reading` lands in the `ebrw_reading` column alongside SAT EBRW; ACT contributes zero rows today, since kippadb ACT scores stop at 2024-10-26.

Removed (5): `test_subject`, `scale_score`, `highest_score_by_test`, `hs_grad_ready`, `college_ready`.

Added (12), grouped by subject so the sheet reads in blocks:

| Total | EBRW/Reading | Math |
| --- | --- | --- |
| `total_scale_score` | `ebrw_reading_scale_score` | `math_scale_score` |
| `total_highest_score_by_test` | `ebrw_reading_highest_score_by_test` | `math_highest_score_by_test` |
| `total_hs_grad_ready` | `ebrw_reading_hs_grad_ready` | `math_hs_grad_ready` |
| `total_college_ready` | `ebrw_reading_college_ready` | `math_college_ready` |

The 26 roster and context columns through `sat_math_highest`, plus `test_type` and `test_date`, are unchanged.

### Two behaviour changes reviewers should see

**`hs_grad_ready` starts carrying values.** It read `NA` on all 7,434 rows because the pivot looked for `HS-Grad Ready` while `stg_google_sheets__kippfwd__goals` spells it `HS-Ready`. Now 741 Yes / 1,684 No. This is a fix, not a regression — but it will look like new data appearing in the sheet.

**159 duplicate rows collapse.** They differed only on `highest_score_by_test`, carried identical scores, and are entirely Camden class of 2027 on the April 2026 school-day SAT — the population already documented in #4871. Nothing is fixed here; the pivot collapses them incidentally, and `max()` over Yes/No picks Yes, which is correct since the sitting is the student's highest if any duplicate row says so. Per-subject reconciliation: SAT Combined 853, EBRW 854, Math 855 all become 801, i.e. 52 + 53 + 54 = 159.

### Filter change

The six-item `score_type` exclusion list becomes `where aligned_subject in ('Total', 'EBRW/Reading', 'Math')`. Equivalent across all 21 `(scope, score_type, aligned_subject)` combinations — every excluded score type maps to `Reading`, `Math Test`, `Reading Test`, `English` or `Science`. Being a positive filter it also excludes score-less students explicitly, which the prior list did implicitly, since a null `score_type` never satisfied `not in`. That preserves the current population exactly: 1,306 of 1,931 roster students. Surfacing the other 625 as blank-score rows so advisors can see who has not tested is a separate ask, noted in the issue.

### Rollback plan

Revert the commit. The model name, the exposure and the Dagster asset key are unchanged, so a revert restores the prior column set with no coordination beyond re-refreshing the sheet tab.

## AI Assistance

Claude Code authored the SQL, the properties file and the verification queries. Human-directed: the request to pivot subjects wide, and the four design decisions — three aligned subject columns rather than a five-column superset, per-subject rather than total-only flags, reshaping in place rather than renaming the model, and fixing the `hs_grad_ready` mismatch in this PR rather than deferring it.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — properties file updated with all 40 contract columns, descriptions on the 12 new ones, and a `dbt_utils.unique_combination_of_columns` test on `(student_number, test_type, test_date)`. This view has never carried a uniqueness test; the collapse makes one possible. Left at the project default `warn` rather than `error` — new invariant on a sheet extract, and the `ccr_course` join could in principle fan out if a student enrolled in two College and Career courses in one year (zero cases today). Closes the `rpt_gsheets__college_assessments_long` bullet of #4871 item 3; the other two views there still depend on the `strategy_case` decision.
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — the existing `google-sheets.yml` exposure needs no change (same model name, same sheet URL).
- [x] **Breaking change?** Yes — five columns removed from a contracted `rpt_` model. The only consumer is the Google Sheet named in the exposure, owned by the Data Team. **The sheet owner must refresh or re-point that tab after merge**: the column set goes from 33 to 40 and the row count falls by roughly two thirds. Rollback plan above.
      with **`--target staging`** — not applicable, no external source changes.

## Verification

`extracts/` materializes as a view, so a green `dbt build` proves only column resolution. Built into a dev schema with `--defer --favor-state` against the production manifest, then compared to production:

- Contract satisfied, build clean, **uniqueness test passes** against real data — 2,425 rows, 2,425 distinct keys
- 40 columns confirmed via `INFORMATION_SCHEMA.COLUMNS`
- 2,425 rows / 1,306 students, matching a pre-implementation prototype run against production
- **Zero key differences and zero value mismatches** on all three subject scores, the highest-score flags and the readiness flags, via a full join to the production view pivoted the same way
- `hs_grad_ready` no longer uniformly `NA` — 741 Yes / 1,684 No, zero NA
- Re-verified after the pre-commit `sqlfmt` pass rewrote the SQL, so the formatted version is what was validated
- `trunk check --force` on both files — no rule violations; the two formatting findings were fixed by the commit hook, which then re-checked clean

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.
- [ ] **Dagster Cloud** — passes or not triggered (no Dagster paths changed).

Closes #4875
Refs #4871

### DIFF
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__college_assessments_long.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__college_assessments_long.yml
@@ -1,5 +1,16 @@
 models:
   - name: rpt_gsheets__college_assessments_long
+    description:
+      One row per student per test administration, with scores and readiness
+      flags pivoted wide across the three aligned subjects (Total, EBRW/Reading,
+      Math).
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          arguments:
+            combination_of_columns:
+              - student_number
+              - test_type
+              - test_date
     columns:
       - name: region
         data_type: string
@@ -55,15 +66,66 @@ models:
         data_type: numeric
       - name: test_type
         data_type: string
+        description:
+          Assessment scope -- SAT, ACT, PSAT 8/9, PSAT10 or PSAT NMSQT.
       - name: test_date
         data_type: date
-      - name: test_subject
-        data_type: string
-      - name: scale_score
+        description: Date of the administration. One row per student per date.
+      - name: total_scale_score
         data_type: numeric
-      - name: highest_score_by_test
+        description:
+          Composite or combined scale score for this administration. Null when
+          the administration produced no overall score.
+      - name: total_highest_score_by_test
         data_type: string
-      - name: hs_grad_ready
+        description:
+          Whether this administration is the student's highest total score for
+          this test type.
+      - name: total_hs_grad_ready
         data_type: string
-      - name: college_ready
+        description:
+          Whether the total score meets the HS-Ready benchmark. NA when no
+          benchmark exists for this score type.
+      - name: total_college_ready
         data_type: string
+        description:
+          Whether the total score meets the College-Ready benchmark. NA when no
+          benchmark exists for this score type.
+      - name: ebrw_reading_scale_score
+        data_type: numeric
+        description:
+          EBRW scale score, or ACT Reading where the test type reports Reading
+          instead.
+      - name: ebrw_reading_highest_score_by_test
+        data_type: string
+        description:
+          Whether this administration is the student's highest EBRW or Reading
+          score for this test type.
+      - name: ebrw_reading_hs_grad_ready
+        data_type: string
+        description:
+          Whether the EBRW or Reading score meets the HS-Ready benchmark. NA
+          when no benchmark exists for this score type.
+      - name: ebrw_reading_college_ready
+        data_type: string
+        description:
+          Whether the EBRW or Reading score meets the College-Ready benchmark.
+          NA when no benchmark exists for this score type.
+      - name: math_scale_score
+        data_type: numeric
+        description: Math section scale score for this administration.
+      - name: math_highest_score_by_test
+        data_type: string
+        description:
+          Whether this administration is the student's highest math score for
+          this test type.
+      - name: math_hs_grad_ready
+        data_type: string
+        description:
+          Whether the math score meets the HS-Ready benchmark. NA when no
+          benchmark exists for this score type.
+      - name: math_college_ready
+        data_type: string
+        description:
+          Whether the math score meets the College-Ready benchmark. NA when no
+          benchmark exists for this score type.
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__college_assessments_long.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__college_assessments_long.yml
@@ -14,63 +14,114 @@ models:
     columns:
       - name: region
         data_type: string
+        description: Region the student's school belongs to.
       - name: schoolid
         data_type: int64
+        description: The School_Number of the associated Schools record.
       - name: school
         data_type: string
+        description: Name of the school the student attends.
       - name: student_number
         data_type: int64
+        description: Student Number assigned by the school. Indexed.
       - name: salesforce_id
         data_type: string
+        description:
+          Identifier for the student's matching kippadb contact record.
       - name: student_name
         data_type: string
+        description: Last, First, Mi. Indexed.
       - name: student_first_name
         data_type: string
+        description: Students' first name.
       - name: student_last_name
         data_type: string
+        description: Students last name.
       - name: grade_level
         data_type: int64
+        description:
+          The grade the student is in. Since this is an integer, 0=Kindergarten
+          1, -2=Preschool. Indexed.
       - name: student_email
         data_type: string
+        description: The student's school-issued email address.
       - name: enroll_status
         data_type: string
+        description:
+          The enrollment status of the student. Categorical mapping of
+          enroll_status.
       - name: ktc_cohort
         data_type: int64
+        description: KIPP Through College cohort year.
       - name: graduation_year
         data_type: int64
+        description: Student graduation year.
       - name: year_in_network
         data_type: int64
+        description: Years enrolled in the network as of that year.
       - name: iep_status
         data_type: string
+        description:
+          Database extension field associated to the Students table. Conditional
+          on spedlep.
       - name: grad_iep_exempt_status_overall
         data_type: string
+        description:
+          The enrollment status of the student. Conditional on enroll_status,
+          grad_iep_exempt_overall.
       - name: cumulative_y1_gpa
         data_type: float64
+        description: Cumulative weighted Y1 GPA through this academic year.
       - name: cumulative_y1_gpa_projected
         data_type: float64
+        description:
+          Weighted projected cumulative Y1 GPA, incorporating current
+          in-progress grades not yet stored.
       - name: college_match_gpa
         data_type: numeric
+        description: Cumulative GPA used for college match categorization.
       - name: college_match_gpa_bands
         data_type: string
+        description: GPA band grouping for college match purposes.
       - name: ccr_course
         data_type: string
+        description:
+          Name of the College and Career course the student is enrolled in this
+          academic year. No Data when the student has no such enrollment.
       - name: ccr_teacher_name
         data_type: string
+        description:
+          Teacher of record for that College and Career section. No Data when
+          the student has no such enrollment.
       - name: ccr_section
         data_type: string
+        description:
+          Section identifier for that College and Career course. No Data when
+          the student has no such enrollment.
       - name: sat_total_superscore
         data_type: numeric
+        description:
+          The student's lifetime SAT superscore — highest section scores summed
+          across administrations, not a single sitting.
       - name: sat_ebrw_highest
         data_type: numeric
+        description:
+          The student's highest SAT EBRW section score across all
+          administrations.
       - name: sat_math_highest
         data_type: numeric
+        description:
+          The student's highest SAT Math section score across all
+          administrations.
       - name: test_type
         data_type: string
         description:
-          Assessment scope -- SAT, ACT, PSAT 8/9, PSAT10 or PSAT NMSQT.
+          Assessment scope — SAT, ACT, PSAT 8/9, PSAT10 or PSAT NMSQT.
       - name: test_date
         data_type: date
-        description: Date of the administration. One row per student per date.
+        description:
+          Date of the administration. One row per student per test type per
+          date, since a student may sit two different assessments on one day.
       - name: total_scale_score
         data_type: numeric
         description:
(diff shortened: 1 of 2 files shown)


anthonygwalters:
Verdicts on all five findings, addressed in 75c849d16.

## 1. CCR attribute mismatch — **accepted, fixed**

Correct, and the observation that the uniqueness test cannot catch it is the important half: a blended row still collapses to one row on the key, so it would pass silently.

Ran the suggested check against AY2026 — zero students hold more than one CCR row, so this was latent rather than live.

Fixed by **grouping** the CCR triple rather than deduping or switching to `first_value`. Two reasons for that shape:

- `src/dbt/CLAUDE.md` explicitly forbids defensive dedupes on `base_powerschool__course_enrollments` (the #3900 / #3915 double-write corpus), so `dbt_utils.deduplicate` was not available here.
- `first_value(... order by ...)` is a window function and this is a `GROUP BY` context, so the cited guidance could not be applied literally.

Grouping keeps the triple intact by construction and converts a future fan-out into a duplicate row the uniqueness test flags, rather than a silently blended one. Applied the same reasoning to `sat_total_superscore` / `sat_ebrw_highest` / `sat_math_highest`, which have the same independent-tiebreak exposure through three separate `rn_highest` joins — that one was not in the finding, but it is the same hazard.

## 2. ST06 column ordering — **claim declined, observation accepted, restructured anyway**

The specific claim, that this "would fail the Trunk CI check," is incorrect:

- `trunk check --force --no-fix` on the committed blob with a clean working tree returned `Checked 1 file / No issues`
- **Trunk Check passed in CI on that exact SHA** (666d09d1e)

sqlfluff does not flag the ordering, so the PR's stated verification stands. Flagging this because the finding presents it as contradicting a verification claim, and that framing would lead a reader to distrust a check that actually ran.

The underlying observation about interleaving against the documented rubric is fair, though. Rather than reorder the columns and lose the by-subject grouping the sheet reads in, moved the six `coalesce(..., 'NA')` wrappers into the `pivoted` CTE. The final `select` is now entirely plain column references, so there is no interleaving at all and the sheet column order is unchanged. Re-linted clean.

## 3. Undocumented legacy columns — **accepted, fixed**

All 26 now documented. Every description is copied from an existing definition elsewhere in the repo rather than authored here, since several of these are KIPP-specific fields whose semantics should not be guessed into a contract file:

| Columns | Sourced from |
| --- | --- |
| roster fields | `int_extracts__student_enrollments` |
| `year_in_network` | `dim_student_enrollments` |
| `cumulative_y1_gpa` | `int_powerschool__gpa_cumulative_year` |
| `cumulative_y1_gpa_projected` | `rpt_gsheets__gpa_roster` |
| `college_match_gpa`, `college_match_gpa_bands` | `rpt_tableau__college_assessment_dashboard_benchmark_calcs` |

The six `ccr_*` / `sat_*` columns are derived in this model, so those are described from its own joins.

## 4. `test_date` description contradicts the grain — **accepted, fixed**

Right, and it was my error: the description said one row per student per date while the key is `(student_number, test_type, test_date)`. Now reads "One row per student per test type per date, since a student may sit two different assessments on one day."

## 5. Em dash — **accepted, fixed**

Confirmed the convention: zero instances of a literal double hyphen in descriptions across this folder, nine files using an em dash. Changed.

## Note on the `_long` name

Correct that it was deliberate. Recorded in #4875: the view is still long by administration, and keeping the name avoids a new Dagster asset key, an exposure edit and a new sheet. The naming collision with `rpt_gsheets__college_assessments_wide` is real and worth a human opinion, but renaming has a coordination cost this change does not otherwise incur.

## Verification of the fixes

The restructure is behavior-preserving — the rebuilt dev view matches the CI-built relation from the previous commit exactly: zero key differences and zero value differences across scores, readiness flags, CCR columns and SAT highs.

One thing worth recording since it looks like a regression and is not: row count reads 2,424 rather than the 2,425 in the PR body. Production drifted during review (7,434 to 7,431 rows, 1,306 to 1,305 students, 2,425 to 2,424 distinct keys) — one student left the qualifying population. Both the old and new code return 2,424 against current data, so the delta is upstream movement, not this change.

PR #4877, opened 2026-08-17
feat(dbt): add Focus id, status, and name-id hash to kippfwd Miami roster

### LINKED ISSUE #4873: feat(dbt): add Focus student id, enroll status, and name-id hash to rpt_gsheets__kippfwd_miami_roster
Append three columns to `rpt_gsheets__kippfwd_miami_roster` for the KIPP Forward Miami team. All three go at the end so existing sheet column positions are unchanged.

- `focus_student_id` — the 8400-prefixed Focus student id, which is Focus `students.student_id` and the value Focus itself displays. Sourced from `int_focus__student_enrollments.student_number`.
- `focus_enroll_status` — the upstream Focus `enroll_status`, surfaced alongside the model's locally-derived `enroll_status`. Carries the `TODO(#4794)` caveat: Focus stamps W01/W02 rollover drop codes at year end, so it reads 361 of 365 AY2025 grade 7-8 students as withdrawn. The derived `enroll_status` remains the column to filter on.
- `student_name_id_hash` — `lastfirst` joined to `focus_student_id` with a hyphen, matching the `rpt_gsheets__kfwd_rem_roster` convention.

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will append four columns to
`rpt_gsheets__kippfwd_miami_roster` for the KIPP Forward Miami team. All four
land at the end of the select, so existing sheet column positions are unchanged.

| Column                    | Type   | Value                                                                                                     |
| ------------------------- | ------ | --------------------------------------------------------------------------------------------------------- |
| `focus_student_id`        | int64  | The 8400-prefixed Focus student id, which is Focus `students.student_id` and the value Focus itself displays |
| `focus_exit_code`         | string | Raw Focus withdrawal code closing that academic year's span (`W01`, `W02`, `DNE`, and so on), null while the span is still open |
| `student_name_id_hash`    | string | `Last, First - 8400000000`, matching the `rpt_gsheets__kfwd_rem_roster` format                             |
| `focus_enrollment_status` | string | Plain in-or-out flag, `Enrolled` or `Not Enrolled`                                                          |

### Why the enrollment flag does not read the exit code

Focus rolls students forward by closing the prior year's span with an in-school
(`W01`) or in-district (`W02`) transfer code. Those two codes cover 359 of 365
AY2025 grade 7-8 rows, while only 4 students actually left mid-year. A flag
derived from the exit code would mark hundreds of currently enrolled students as
gone.

`focus_enrollment_status` instead reuses the model's existing `open_enrollment`
CTE, the same input the derived `enroll_status` already uses. That keeps the two
in lockstep without duplicating the CASE branches, and it handles the
pre-first-day case correctly, since a student who has not started yet still has
an open current-year span.

`focus_exit_code` is kept alongside it as the reason a student shows as Not
Enrolled, but note it answers how that year's span closed, not whether the
student left.

### Validation

- `dbt build` plus all 6 data tests pass against the committed state.
- Row counts unchanged at 365 for AY2025 and 376 for AY2026, so nothing fans
  out.
- No nulls in any of the four new columns across all 741 rows.
- `student_name_id_hash` is well formed on every row, and the id it embeds
  matches `focus_student_id` on every row.
- `focus_enrollment_status` agrees with `enroll_status = 0` on all 741 rows,
  confirming it restates current standing rather than introducing a second
  source of truth.
- `trunk check --force` clean on both files.

Column ordering inside the new block follows sqlfluff ST06 (plain references,
then functions, then logicals), which is why the hash sits before the flag.

## AI Assistance

Authored by Claude Code. Human-directed: which columns to add, using the
prefixed Focus id, and the call to reduce an exit-code decode down to a plain
in-or-out flag. AI-assisted: finding the naming precedent, identifying the
rollover-code trap, writing the SQL and properties, and running the validation
above.

### General

### dbt

- [x] Properties file updated with all four new contract columns, each with a
      description. Both id columns and the hash carry `contains_pii`.
- [x] Exposure already covers this model. The `access_denied` exposure in
      `models/exposures/google-sheets.yml` depends on the model rather than on
      its column list, so no exposure change is needed.
- [x] Not a breaking change. Columns are added, none renamed or removed, and
      they append at the end so existing sheet positions hold. Rollback is a
      straight revert.

### Docs

Skipped, no schedule, sensor, or integration changes.

## CI checks

- [ ] **Trunk**
- [ ] **dbt Cloud**
- [ ] **Dagster Cloud** — expected not to trigger, no Dagster paths changed.

Refs #4873

### DIFF
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__kippfwd_miami_roster.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__kippfwd_miami_roster.yml
@@ -210,3 +210,42 @@ models:
           Prior academic year's FLDOE FAST math progress monitoring 3
           achievement level and scale score.
         data_type: string
+      - name: focus_student_id
+        description: >-
+          Focus student id - the network student number prefixed with 8400,
+          Miami-Dade's FLDOE district number. This is the value Focus displays
+          and the id to search on there. One known student carries an id without
+          the prefix, which passes through unchanged.
+        data_type: int64
+        config:
+          meta:
+            contains_pii: true
+      - name: focus_exit_code
+        description: >-
+          Raw Focus withdrawal code closing this academic year's enrollment
+          span, for example W01 in-school transfer, W02 in-district transfer,
+          W06 graduated, DNE did not enter. Null while the span is still open.
+          Focus rolls students forward by closing the prior span with a transfer
+          code, so a completed academic year carries a code even for students
+          who never left - read focus_enrollment_status, not this column, to
+          tell who is still here.
+        data_type: string
+      - name: focus_enrollment_status
+        description: >-
+          Whether the student is currently enrolled at a KIPP Miami school,
+          Enrolled or Not Enrolled. Resolved from the student's open
+          current-year Focus enrollment rather than from this row's exit code,
+          so a prior-year row reflects where the student is now. A student whose
+          enrollment starts later in the current year reads Enrolled.
+          Student-level, so every row for the same student carries the same
+          value.
+        data_type: string
+      - name: student_name_id_hash
+        description: >-
+          Student name and Focus student id joined by a hyphen, formatted as
+          Last, First - 8400000000. Matches the identifier format the other KIPP
+          Forward sheet rosters use for lookups.
+        data_type: string
+        config:
+          meta:
+            contains_pii: true
--- a/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__kippfwd_miami_roster.sql
@@ -56,6 +56,7 @@ with
             sex_label,
             ese_fefp_code,
             cast(disis_id as string) as mdcps_id_raw,
+            cast(student_id as string) as student_id_string,
         from {{ ref("int_focus__students") }}
     ),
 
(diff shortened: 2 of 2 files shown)


anthonygwalters:
Verified each finding against the code before replying, line citations included — all four cite accurately. Declining all four, reasoning per item. No code changes made.

### 1. New-block column order inverts join order — declined

Citation accurate: `s.student_id` is at `:147` and `e.exitcode` at `:149`, so the second-joined table does precede the base table.

Declining because the convention is not enforced here and the change is not free. `trunk check --force` passes on both files, so sqlfluff ST06 does not flag it, and this model already interleaves `e.` refs across three non-contiguous places, as you noted. Swapping the two lines would also flip the new columns' order in the published Sheet for no reader benefit. It is a two-line change if a human reviewer prefers it.

### 2. `concat()` vs `||` inconsistency — declined

Citation accurate, and `rpt_gsheets__kfwd_rem_roster.sql:167` does use `||`.

Declining because `concat()` is the dominant idiom in this directory by a wide margin: 50 `concat(` calls against 6 lines using `||` for string concatenation. The PR description claims the column produces the same output format as that sibling, `Last, First - id`, which it does. It does not claim to copy the operator. Switching to the rarer operator to match one cited model would move away from the directory norm.

### 3. Duplicated `oe.student_number is not null` predicate — declined

Citation accurate, `:122` and `:162`.

Declining on cost. The predicate is evaluated after the `open_enrollment` join, so it cannot be hoisted into a CTE the way a single-table expression could. Sharing it would mean wrapping the entire final SELECT in a CTE and re-enumerating all 33 output columns, since `src/dbt/CLAUDE.md` forbids a bare star in the final select of an `rpt_` model. That is a large diff to de-duplicate one null check.

### 4. `mdcps_id_raw` vs `student_id_string` naming asymmetry — declined

The suffixes encode different roles rather than drifting. `mdcps_id_raw` is raw because it is not the final value: it is consumed by `lpad(s.mdcps_id_raw, 7, '0') as mdcps_id` at `:88`. `student_id_string` is a pure type conversion with no further processing. Naming it `_raw` would imply a padding or cleanup step that does not exist.

### Data-dependent items you flagged as unverifiable

All confirmed locally against a `--defer --favor-state` build. Compiled SQL was checked to confirm every upstream resolved to a prod relation, so no stale dev copy shadowed the results.

- Row counts unchanged at 365 for AY2025 and 376 for AY2026, 741 total. No fan-out from the new column.
- No nulls in any of the four new columns across all 741 rows.
- The `students` LEFT JOIN you correctly flagged as theoretically nullable: 0 unmatched rows in both years.
- `student_name_id_hash` is well formed on all 741 rows, and the id it embeds equals `focus_student_id` on every row, 0 mismatches.
- `focus_enrollment_status` equals `enroll_status = 0` on all 741 rows, 0 disagreements.
- dbt Cloud CI passed with 0 warnings, checked explicitly rather than inferred from the success status.

Confirming your read on the `rn_year = 1` grain as well: an open span takes `coalesce(end_date, date(syear + 1, 6, 30))`, so it sorts last under `exitdate desc`, and `e.exitcode` is therefore genuinely the code that closes that academic year's span.

### Unrelated red check

`Analyze (actions)` is failing from the GitHub incident during this PR's first CI pass, not from anything in the diff. CodeQL never ran; it failed downloading its own action from `codeload.github.com` with two 429s and a 503. This PR changes two dbt files and no workflow files, so that job had nothing in the diff to scan. It needs a re-run from the Actions tab.

PR #4879, opened 2026-08-17
feat: use Paterson real U_EXPECTATIONS data and drop the Newark spoof

### LINKED ISSUE #4878: feat: use Paterson real U_EXPECTATIONS data and drop the Newark spoof
## Background

Paterson MS gradebook-audit expectations are currently **spoofed from Newark**. The KIPP NJ Gradebook Audit plugin could not be installed on Paterson's PowerSchool instance, so `U_EXPECTATIONS` was never populated there and `u_expectations` was pruned from Paterson's dlt asset config after a `NoSuchTableError` (see the Paterson dlt plan, Task 5 Step 2).

The workaround landed in two commits:

- `7e2e834fa` — an override model inside the `kipppaterson` project reading Newark's source
- `af1e53b67` — moved to a `paterson_spoof` CTE in `kipptaf`'s `stg_powerschool__u_expectations`, per review

Today the union view returns Paterson MS rows that are byte-identical to Newark MS (41 rows, all four quarters).

The plugin has since been installed on Paterson's instance. The table now exists in Oracle, though academics has not entered AY 2026-2027 expectations yet.

## Scope

One PR, commits ordered so the ingestion is proven before the dbt cut-over lands.

### Ingestion

- Add `u_expectations` to `src/teamster/code_locations/kipppaterson/powerschool/sis/dlt/config/assets.yaml`, matching Newark's config (`cursor_column: whenmodified`, `intraday: true`, `nightly: false`)
- Bump the three hardcoded Paterson asset counts from 48 to 49:
  - `tests/assets/test_powerschool_dlt_paterson_defs.py`
  - `tests/libraries/test_dlt_powerschool_assets.py` (two assertions)
- Materialize the asset and confirm `dagster_kipppaterson_dlt_powerschool.u_expectations` lands in BigQuery, even at zero rows

### dbt cut-over

- Remove the `+enabled: false` for `stg_powerschool__u_expectations` in `src/dbt/kipppaterson/dbt_project.yml`
- Re-add the `stg_powerschool__u_expectations` source entry to `src/dbt/kipptaf/models/powerschool/sources-kipppaterson.yml`
- In `src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__u_expectations.sql`: add the Paterson relation to `union_relations` and delete the `paterson_spoof` CTE plus its trailing `union all`

### Docs

- `docs/models/gradebook-audit-data-model.md` — correct the AY 2026-2027 coverage table, the Paterson GradeBook plugin note, and the open-questions item that says the plugin cannot be installed

### Cleanup

- Drop the orphaned relation left behind when the override model was deleted. It is a real table with 41 stale rows, last modified on the spoof PR date, and reads as a decoy:

  ```sql
  drop table `teamster-332318.kipppaterson_powerschool.stg_powerschool__u_expectations`;
  ```

## Risks

- **Empty source table.** If dlt does not create a BigQuery table for a zero-row source, the newly enabled staging model fails on a missing relation, and because it feeds the kipptaf union that failure cascades into the whole gradebook audit rather than just Paterson. Confirm the table lands before pushing the cut-over commits.
- **Coverage gap.** Once the spoof is removed, Paterson MS produces no `assignment_teacher` rows until academics enters expectations in PowerSchool. The summer toggle is currently applied, so this is also visible in any summer dashboard work.

Refs #3908

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will make Paterson's gradebook-audit expectations
use Paterson's own `U_EXPECTATIONS` data instead of values spoofed from Newark,
and adds the Google Sheet T&L uses to push expectations back into PowerSchool.

Paterson MS expectations were spoofed from Newark because the KIPP NJ Gradebook
Audit plugin could not be installed on Paterson's PowerSchool instance —
`U_EXPECTATIONS` was never populated there, and `u_expectations` had been pruned
from Paterson's dlt asset config after a `NoSuchTableError`. The workaround was a
`paterson_spoof` CTE reading Newark's relation filtered to `school_level = 'MS'`
and relabelled `kipppaterson`, which returned 41 rows byte-identical to Newark
MS.

The plugin is now installed, so the spoof comes out.

## Action required by the merger

One warehouse `DROP` needs running, ideally **before** the post-merge prod
rebuild. The PR author does not have DDL rights:

```sql
drop table `teamster-332318.kipppaterson_powerschool.stg_powerschool__u_expectations`;
```

This is an orphan left behind when the original spoof override model was deleted
in `af1e53b67` — dbt dropped the model but not the relation, so it survives as a
real table holding 41 stale Newark-derived rows.

Order matters. Enabling the staging model makes dbt `create or replace` that same
relation, so a **successful** build overwrites it harmlessly. But if the build
ever fails, the orphan stays and the kipptaf union reads those 41 stale rows as
though they were real Paterson data. Dropping it first converts that silent
wrong-data mode into a loud missing-relation error.

### AI Assistance

Claude Code authored the commits. Human-directed: the decision to un-spoof, the
sequencing (prove ingestion before cutting over), the single-PR structure, the
template model's design, and accepting sparse Paterson data. AI-assisted:
tracing the spoof history, auditing warehouse state, diagnosing the ingestion
failure, and writing the config, test, SQL, and docs changes.

## Commits

| Commit | What |
| --- | --- |
| `0176d15ef` | Add `u_expectations` to Paterson's dlt config; bump three asset counts; repair a stale test spec map |
| `df1ff0ab4` | Library fix — tolerate a dlt resource that produced no load job |
| `53ce59042` | Derive the Paterson table count from the spec sets (review finding) |
| `f62ee6f28` | Enable the staging model, add the source and union relation, delete the spoof |
| `ef1a26d77` | Add `rpt_gsheets__gradebook_audit_template` plus docs and skill updates |

## Verification

The dlt asset was materialized against the branch deployment before the cut-over
commits landed. First run **failed**: Paterson's `U_EXPECTATIONS` existed in
Oracle but held zero rows, so dlt produced no load job and created no BigQuery
table. After one row was entered in PowerSchool, the re-run succeeded and
`dagster_kipppaterson_dlt_powerschool.u_expectations` now exists.

That failure is what commit `df1ff0ab4` fixes. `dagster_dlt`'s
`fetch_row_count()` raises when a materialization carries no `jobs` metadata,
which is what dlt emits for the first load of a table that extracts zero rows —
`_handle_empty_tables` only writes the empty truncate file once a table has
`seen_data`. Because all PowerSchool tables share one multi-asset op, that raise
aborts the whole step. Combined with the intraday sensor treating a missing
baseline as drift, Paterson would have re-selected the table every 15 minutes and
taken its other PowerSchool tables down with it. Established tables were never
at risk.

`rpt_gsheets__gradebook_audit_template` builds clean with its uniqueness test
passing, and returns the intended coverage: Newark MS/HS, Camden MS/HS, Paterson
MS; no ES, no Miami.

## Known trade-offs and follow-ups

- **Paterson data is sparse.** Academics has entered one AY 2026-2027
  expectations row, so Paterson goes from 41 spoofed rows to that single real
  one. Category flags will be nearly empty until the full set is loaded.
  Accepted deliberately.
- **dbt Cloud CI cannot validate the cut-over.**
  `zz_stg_kipppaterson_powerschool.stg_powerschool__u_expectations` still holds
  the 41 spoofed rows, so the union wrapper compiles green against stale data
  either way. Refreshing it needs a `dbt clone --target staging --full-refresh`.
- **An orphaned relation needs a manual drop** — left behind when the original
  override model was deleted; a real table with 41 stale rows, last modified on
  the spoof PR date. Dropping it BEFORE the prod rebuild converts a
  silent-wrong-data failure mode into a loud missing-relation one:

  ```sql
  drop table `teamster-332318.kipppaterson_powerschool.stg_powerschool__u_expectations`;
  ```

- **`docs/reference/automations.md` has no `kipppaterson` PowerSchool dlt rows
  at all.** Flagged by review and confirmed — only the retired couchdrop SFTP
  sensor's asset list remains under that section. Pre-existing staleness, not
  caused here, and out of scope: per `docs/CLAUDE.md` the catalog can only be
  regenerated where every code location imports, which a codespace cannot do.

### General

- [x] Reviewed the **Claude Code Review** comment; per-finding verdicts posted as
      a PR comment.

### Dagster

- [x] `uv run pytest tests/libraries/test_dlt_powerschool_assets.py tests/assets/test_powerschool_dlt_paterson_defs.py` — 10 passed
- [x] dlt asset materialized successfully against the branch deployment

### dbt

- [x] Properties file included for the new model, with contract column types and
      a uniqueness test
- [x] Exposure added for the new Google Sheet
- [x] No contracted column renames or removals — the union model returns to its
      pre-spoof shape
- [x] No external source added or modified, so no `stage_external_sources` run
      is needed

### Docs

- [x] No schedule or sensor change — dlt trigger membership derives from
      `assets.yaml`, so the new table is covered by the existing sensor without
      a config edit. Note this does **not** mean the automations catalog is
      current for this integration; see the follow-up above.
- [x] Updated `src/teamster/code_locations/kipppaterson/CLAUDE.md`, the
      gradebook-audit reference doc, and the `gradebook-audit` skill.

Closes #4878
Refs #3908

### DIFF
--- a/.claude/skills/gradebook-audit/SKILL.md
+++ b/.claude/skills/gradebook-audit/SKILL.md
@@ -129,10 +129,15 @@ Current `depends_on` list (update if the exposure changes):
 There is also a disabled exposure `gradebook_audit_teacher_report` — mention it
 only if the user asks about disabled or archived workbooks.
 
-The companion student-flags Google Sheet has its own separate exposure in
-`src/dbt/kipptaf/models/exposures/google-sheets.yml`, named
-`rpt_gsheets__gradebook_audit_student_flags` — check there if asked about the
-gsheets side rather than the Tableau side.
+Two companion Google Sheets have their own exposures in
+`src/dbt/kipptaf/models/exposures/google-sheets.yml` — check there if asked
+about the gsheets side rather than the Tableau side:
+
+- `rpt_gsheets__gradebook_audit_student_flags` — the flagged-student review
+  sheet (read side, carries student PII)
+- `rpt_gsheets__gradebook_audit_template` — the expectations upload template
+  (write side; T&L exports it as the CSV they load into `U_EXPECTATIONS` via the
+  PowerSchool plugin)
 
 ---
 
@@ -246,9 +251,27 @@ correctly.
    region's PowerSchool instance and the `U_EXPECTATIONS` table is populated.
    Plugin source and update instructions:
    [TEAMSchools/ps-plugins](https://github.com/TEAMSchools/ps-plugins)
-2. Verify `int_powerschool__u_expectations_qtd_unpivot` returns rows for the new
+2. Wire the ingestion and the union — four files, in this order (Paterson's
+   rollout in #4879 is the worked example):
+   - add `u_expectations` to
+     `src/teamster/code_locations/<district>/powerschool/sis/dlt/config/assets.yaml`
+     (`cursor_column: whenmodified`, `intraday: true`, `nightly: false`) and
+     bump that district's hardcoded asset counts in `tests/`
+   - drop the `stg_powerschool__u_expectations: +enabled: false` entry from
+     `src/dbt/<district>/dbt_project.yml`
+   - add the `stg_powerschool__u_expectations` source entry to
+     `src/dbt/kipptaf/models/powerschool/sources-<district>.yml`
+   - add the relation to `union_relations` in
+     `src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__u_expectations.sql`
+3. **Materialize the dlt asset BEFORE enabling the dbt staging model, and
+   confirm the BigQuery table exists.** dlt creates no table at all for the
+   first load of a source table that is empty, and the newly enabled staging
+   model then fails on a missing relation — which cascades into the kipptaf
+   union and takes the whole gradebook audit down, not just the new region. The
+   plugin being installed is not sufficient; the table needs at least one row.
+4. Verify `int_powerschool__u_expectations_qtd_unpivot` returns rows for the new
    region.
-3. No flag sheet changes needed — the flag columns in
+5. No flag sheet changes needed — the flag columns in
    `rpt_tableau__gradebook_audit` apply to all regions. The only exclusions,
    applied in `category_join`'s `WHERE` clause (and matched in
    `int_extracts__gradebook_audit_student_flags`'s own filters, which both
@@ -297,8 +320,9 @@ reads; the gsheets report itself no longer carries any toggle):
 - `src/dbt/kipptaf/models/students/intermediate/int_extracts__gradebook_audit_student_flags.sql`
 - `src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__u_expectations_qtd_unpivot.sql`
 - `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_es_comments.sql`
+- `src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__gradebook_audit_template.sql`
 
-**Four changes to make:**
+**Five changes to make:**
 
 1. In `rpt_tableau__gradebook_audit` — change the year filter in
    `category_join`'s `WHERE` clause (1 occurrence, marked
@@ -398,7 +422,25 @@ reads; the gsheets report itself no longer carries any toggle):
    fallback data source available; escalate instead of shipping a change that
    silently reports every comment as missing.
 
-Build and verify after all four changes:
+5. In `rpt_gsheets__gradebook_audit_template` — the expectations upload template
+   T&L uses to build the PowerSchool CSV. Change both occurrences (one filters
+   `int_powerschool__calendar_week` in the `term_weeks` CTE, marked
+   `-- summer toggle: see skill`; one stamps the output `academic_year` column,
+   marked `/* summer toggle: see skill */`):
+
+   ```sql
+   -- change this (appears 2 times):
+   {{ var("current_academic_year") }}
+   -- to this:
+   {{ var("current_academic_year") - 1 }}
+   ```
+
+   **While toggled, this model shows the PRIOR year's week grid.** It is the
+   sheet T&L exports to upload the NEW year's expectations, so do not hand it
+   over as the new-year grid until the toggle is reverted — they would be
+   editing last year's weeks.
+
+Build and verify after all five changes:
 
 ```bash
 uv run dbt build \
@@ -406,6 +448,7 @@ uv run dbt build \
     int_extracts__gradebook_audit_student_flags \
     rpt_gsheets__gradebook_audit_student_flags rpt_tableau__gradebook_audit \
     rpt_tableau__gradebook_es_comments \
+    rpt_gsheets__gradebook_audit_template \
   --project-dir src/dbt/kipptaf \
   --defer \
   --state target/prod
@@ -417,7 +460,7 @@ would otherwise read the un-toggled prod copy.
 
 **When to revert:** once the new school year starts and teachers begin entering
 grades in PowerSchool (typically Q1), revert all changes:
-`current_academic_year - 1` → `current_academic_year` in all four files, and
+`current_academic_year - 1` → `current_academic_year` in all five files, and
 `'last_year'` → `'current_year'` in
 `int_extracts__gradebook_audit_student_flags`.
 
--- a/docs/models/gradebook-audit-data-model.md
+++ b/docs/models/gradebook-audit-data-model.md
@@ -52,35 +52,27 @@ The audit operates at several grains:
 
 ### Dashboard coverage (AY 2026-2027)
 
-| Region   | School level | Coverage depth                              | Notes                                                          |
-| -------- | ------------ | ------------------------------------------- | -------------------------------------------------------------- |
-| Camden   | MS, HS       | Full audit (all applicable flags)           |                                                                |
-| Newark   | MS, HS       | Full audit (all applicable flags)           |                                                                |
-| Paterson | MS           | Full audit (all applicable flags)           | Added AY 2026-2027; expectations spoofed from Newark's MS data |
-| Camden   | ES           | EOQ comments only (`qt_es_comment_missing`) | ES schools do not enter assignments in PS gradebook            |
-| Newark   | ES           | EOQ comments only (`qt_es_comment_missing`) | ES schools do not enter assignments in PS gradebook            |
-| Paterson | ES           | EOQ comments only (`qt_es_comment_missing`) | Added AY 2026-2027                                             |
-| Miami    | n/a          | Removed                                     | Moved to Focus gradebook; excluded at scaffold level           |
+| Region   | School level | Coverage depth                              | Notes                                                |
+| -------- | ------------ | ------------------------------------------- | ---------------------------------------------------- |
+| Camden   | MS, HS       | Full audit (all applicable flags)           |                                                      |
+| Newark   | MS, HS       | Full audit (all applicable flags)           |                                                      |
+| Paterson | MS           | Full audit (all applicable flags)           | Added AY 2026-2027                                   |
+| Camden   | ES           | EOQ comments only (`qt_es_comment_missing`) | ES schools do not enter assignments in PS gradebook  |
+| Newark   | ES           | EOQ comments only (`qt_es_comment_missing`) | ES schools do not enter assignments in PS gradebook  |
+| Paterson | ES           | EOQ comments only (`qt_es_comment_missing`) | Added AY 2026-2027                                   |
+| Miami    | n/a          | Removed                                     | Moved to Focus gradebook; excluded at scaffold level |
 
-!!! note "Paterson GradeBook plugin" The **KIPP NJ Gradebook Audit** plugin
-cannot be installed on Paterson's PowerSchool instance, so `U_EXPECTATIONS` is
-never populated there natively. `kipptaf`'s own
-`stg_powerschool__u_expectations` model handles this directly: a
-`paterson_spoof` CTE reads Newark's real U_EXPECTATIONS data via
-`source("kippnewark_powerschool", "stg_powerschool__u_expectations")`, filtered
-to `school_level = 'MS'`, with a hardcoded
-`'kipppaterson' as _dbt_source_project` (bypassing the usual
-`extract_source_project()` regex-on-schema-name derivation, since the physical
-relation really is Newark's — deriving from it would mislabel these rows as
-`kippnewark`). This is `UNION ALL`-ed alongside the real Camden/Newark union
-(`src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__u_expectations.sql`).
-There is no per-district override model in `kipppaterson` anymore — that
-approach (a `kipppaterson`-project model reading Newark's source under a
-misleadingly-named `kippnewark_powerschool` source declaration) was replaced
-after review; see PR #4132 review discussion.
-`int_powerschool__u_expectations_qtd_unpivot`'s kipptaf-level union picks this
-up as a third region alongside Camden and Newark — no different from a real
-region's data at that point. Tracked:
+!!! note "Paterson GradeBook plugin" The **KIPP NJ Gradebook Audit** plugin is
+now installed on Paterson's PowerSchool instance, so `U_EXPECTATIONS` is
+populated there natively and Paterson is an ordinary third region in the
+kipptaf-level union
+(`src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__u_expectations.sql`),
+alongside Camden and Newark. The raw table is ingested by the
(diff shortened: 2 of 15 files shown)


GabyRangelB:
## Review response

Verdicts on all three findings, plus the outcome of the gate this PR was
blocking on.

### 1. Three independently-maintained magic-number table counts — **fixed** (`53ce59042`)

Agreed, and the framing was right: it is the same drift shape the PR was already
repairing elsewhere in that file. `test_config_matches_spec_membership_map` now
derives the count instead of hardcoding it:

```python
assert len(entries) == len(
    INTRADAY_TRANSACTION_DATE
    | INTRADAY_WHENMODIFIED
    | NIGHTLY_WHENMODIFIED
    | NIGHTLY_NO_CURSOR
)
```

A table added to the YAML but not to a spec set now fails here rather than
drifting silently. Took the suggestion to leave the other two literal counts
alone — the spec sets are not in scope in those tests, so they stay a second
independent signal.

Also confirming the set arithmetic independently: 13 + 11 + 11 + 14 = 49, and
the four sets are disjoint by construction, so the derived form is equivalent to
the literal it replaced.

### 2. Two logically separate changes bundled in one commit — **declined for that commit, adopted going forward**

Fair observation, but not worth acting on retroactively: `0176d15ef` was already
pushed and under review, so isolating the drive-by fix would mean rewriting
published history for a marginal gain in blame precision.

Adopted for everything after it — the four subsequent commits each do one thing
(`df1ff0ab4` library fix, `53ce59042` this test fix, `f62ee6f28` the cut-over,
`ef1a26d77` the upload template and docs).

### 3. `docs/reference/automations.md` has no `kipppaterson` PowerSchool dlt rows — **confirmed, PR body corrected**

Verified: the only `kipppaterson` entry in that file is the retired couchdrop
SFTP sensor's asset list, which still enumerates `powerschool/sis/sftp/*` assets
from the pre-dlt era. There are no dlt rows for this integration at all.

The point about the self-review checkbox was the useful part — the claim was
true, but for the wrong reason. The PR body now states plainly that trigger
membership derives from `assets.yaml` (so no config edit is needed) **and** that
this does not mean the catalog is current for this integration, with the
staleness listed as a follow-up. Regeneration stays out of scope: per
`docs/CLAUDE.md` the script silently skips code locations that fail to import,
so it can only be run in an environment where all of them load — not a codespace.

### On "what I could not verify"

The sandbox limits called out were the right ones to flag. Filling them in:

- `uv run pytest` on both affected files: **10 passed**.
- The gate this PR was structured around — materializing the dlt asset and
  confirming the BigQuery table lands — has now run. The **first attempt failed**:
  Paterson's `U_EXPECTATIONS` existed in Oracle but held zero rows, dlt produced
  no load job, and no table was created.

That failure turned out to matter well beyond this PR, and produced commit
`df1ff0ab4`. `dagster_dlt`'s `fetch_row_count()` raises when a materialization
carries no `jobs` metadata, which is exactly what dlt emits for the first load of
an empty table (`_handle_empty_tables` only writes the empty truncate file once a
table has `seen_data`). Since all PowerSchool tables share one multi-asset op,
that raise aborts the entire step — and because the intraday sensor treats a
missing baseline as drift, the table would have re-selected every 15 minutes and
taken Paterson's other PowerSchool tables down with it indefinitely. Established
tables were never exposed to this.

After one expectations row was entered in PowerSchool, the re-run succeeded and
the table now exists, so the cut-over commits were safe to land.

GabyRangelB:
@claude — verdicts on all six findings. Both Important ones were checkable
locally, so I ran them rather than reasoning about them. One confirmed, one
partly right for a reason that changes the fix, and your Informational #6 is
refuted by data. Fixes in `4b7d54149`.

### 1. `_FakeRunResult` not iterable — **CONFIRMED, fixed**

You were right, and the trace was exact. Running it:

```text
>               for event in dlt.run(
E               TypeError: '_FakeRunResult' object is not iterable
src/teamster/libraries/dlt/powerschool/assets.py:323: TypeError

FAILED tests/libraries/test_powerschool_dlt_extract_workers.py::test_op_sets_extract_workers_from_tag
FAILED tests/libraries/test_powerschool_dlt_extract_workers.py::test_op_leaves_extract_workers_unset_without_tag
2 failed, 4 passed
```

A real regression I shipped in `df1ff0ab4`, and the diagnosis of why I missed it
is correct too: I ran only the two files that commit edited, not the sibling
file testing the same factory.

Fixed by taking your first option — the stub now implements `__iter__` yielding
nothing, which matches the interface the op actually uses:

```python
class _FakeRunResult:
    """Stand-in for dlt's `LoadInfo`; the op iterates it for materializations."""

    def __iter__(self) -> Iterator[Any]:
        return iter(())
```

Also swept for any other caller or stub still assuming the chained form —
`grep -rn "fetch_row_count" src/ tests/` returns only the doc from finding #4.
Nothing else was affected.

Now **18 passed** across `test_powerschool_dlt_extract_workers`,
`test_dlt_powerschool_assets`, `test_dlt_powerschool_sensors`, and
`test_powerschool_dlt_paterson_defs`.

### 2. Inverted empty-metadata guard — **right that it was wrong, but it was dead code; removed**

Correct that the guard made no sense, but the impact analysis lands differently
once you can read the installed package, which the sandbox blocked. The first
statement of `fetch_row_count_metadata` is:

```python
def fetch_row_count_metadata(materialization, context, dlt_pipeline) -> TableMetadataSet:
    if not materialization.metadata:
        raise Exception("Missing required metadata to retrieve row count.")
```

So a falsy `event.metadata` raises *before* returning, and control lands in the
`except` branch. Reaching `if not event.metadata:` at all guaranteed metadata was
truthy — the branch was unreachable. It could never have discarded a fetched
`row_count_metadata`, so the "silently dropped" impact doesn't occur.

That makes removal the honest fix rather than inverting it, so the guard is gone:

```python
return event._replace(metadata={**row_count_metadata, **(event.metadata or {})})
```

The `or {}` is only there to satisfy the type checker; it restores no branch.

For the record, the guard was inherited — `DltEventIterator.fetch_row_count`
does `if event.metadata:` with no `else`, which **drops** the event entirely.
Keeping the event was the deliberate part; the condition was cargo.

### 3. Broad `except Exception` — **declined, no narrower type exists**

Checked the installed source. `dagster_dlt` raises a bare
`Exception("Missing jobs metadata to retrieve row count.")` — no custom class. The
only way to narrow is matching on the message, which this repo's retry/exception
convention in `src/teamster/CLAUDE.md` explicitly forbids ("Match on a specific
exception class ... never on error message"). A broad catch that logs a warning
is the lesser evil here. Happy to revisit if upstream ever types it.

### 4. Stale `libraries/dlt/CLAUDE.md` note — **valid, fixed**

Correct, and it was doc staleness I introduced. That bullet documented
`.fetch_row_count()` as the pattern to copy while the code had moved away from
it. It now records why the chained call is unsafe here, points at
`_with_row_count()` as the shape to copy, and adds the thing that would have
caught finding #1: a stub for the `run()` return value must be **iterable**, not
expose `.fetch_row_count()`.

### 5. Duplicated category decode — **acknowledged, not acted on**

Agreed on both the duplication and the conclusion. The decode block is already
repeated across the gradebook-expectations models, no macro exists, and this PR
adding a second consumer isn't the right moment to extract one. Left as-is,
matching your read.

### 6. `week_end_sunday` possible non-determinism — **refuted by data**

Reasonable inference, but it doesn't hold. The distinction is that
`school_week_end_date` is the last *in-session* day (so it moves when only some
schools lose a day), whereas `week_end_sunday` is a pure calendar offset from
`week_start_monday` — the same grouping key — so it cannot vary within the
partition:

```sql
select count(*) as split_groups
from (
  select _dbt_source_project, region, school_level, `quarter`, week_number_quarter
  from `teamster-332318.kipptaf_powerschool.int_powerschool__calendar_week`
  where academic_year = 2025
  group by 1, 2, 3, 4, 5
  having count(distinct week_end_sunday) > 1
)
-- split_groups = 0
```

Zero split groups, so the sibling model's "functionally determined by the
partition key" annotation is accurate and its `dbt_utils.deduplicate` tiebreaker
has nothing to disambiguate. No follow-up needed.

### On CodeQL

Unrelated to the code: `Analyze (javascript)` and `Analyze (python)` failed
because the CodeQL `init` step couldn't reach GitHub's API during the outage
earlier — `HttpError: No server is currently available to service your request`,
then `Debugging artifacts are unavailable since the 'init' Action failed`.
Nothing was scanned. Needs a re-run, not a fix.

GabyRangelB:
Correcting the record on my own finding #3 response, since a wrong convention
citation is the kind of thing that gets copied forward.

I wrote that declining to narrow the `except Exception` was backed by "this
repo's retry/exception convention in `src/teamster/CLAUDE.md`". That
overstates it. The rule reads:

```text
**Retry pattern**: Use `tenacity` — standard import set:
`retry, retry_if_exception_type, stop_after_attempt, wait_exponential_jitter`.
Match on a specific exception class (define one if the upstream code raises bare
`Exception`), never on error message or bare `Exception`.
```

It sits under **Retry pattern** and governs `tenacity` retry predicates — not
`except` blocks generally. `_with_row_count`'s catch is a catch-log-continue, not
a retry, so that rule never applied to it.

The decision stands, on better grounds: `except Exception` used as
catch-log-continue is established precedent in this library tree
(`email/resources.py`, `ssh/resources.py`, `dlt/focus/sensors.py`,
`dlt/powerschool/sensors.py`, `powerschool/sis/odbc/utils.py`,
`amplify/mclass/api/resources.py`), and `dagster_dlt` raises a bare `Exception`
with no subclass to catch, so the only available narrowing is message-matching —
which is genuinely worse. No code change.

No other verdict in that comment is affected.

GabyRangelB:
@claude — tagging you on the correction above
(https://github.com/TEAMSchools/teamster/pull/4879#issuecomment-5319358492),
which I posted untagged by mistake.

Short version: your finding #3 point was right and I've conceded it. I claimed
`src/teamster/CLAUDE.md`'s "never on error message or bare `Exception`" rule
backed my decline. It does not — that rule sits under **Retry pattern** and
governs `tenacity` retry predicates, not `except` blocks generally.
`_with_row_count`'s catch is catch-log-continue, so the rule never applied.

The decline stands on the grounds you supplied instead: catch-log-continue
`except Exception` is established precedent across this library tree, and
`dagster_dlt` raises a bare `Exception` with no subclass, so the only available
narrowing is message-matching. No code change.

Two other things from your verification comment, for the record:

- Nice catch deriving the **18 passed** count from the test-function totals
  (6 + 7 + 3 + 2) rather than taking my word for it. That is the right instinct
  given the sandbox can't execute.
- On #6, agreed the data claim isn't provable from your sandbox. Your code-level
  trace is the stronger evidence anyway: `week_end_sunday` resolving to
  `date_trunc(date_value, week) + 6 days` with no `schoolid` dependency is what
  makes the zero-split result a structural property rather than a lucky sample.

Nothing outstanding for you unless you see something in `4b7d54149` you want to
revisit.

GabyRangelB:
@cbini — convention question, **not** a review request. This PR isn't ready for
approval yet (CodeQL needs a re-run after the outage). Flagging this now only
because it came up while working here.

## The observation

To get `claude-review` to look at commits pushed after its first pass, I followed
the root `CLAUDE.md`, which says:

> To get `claude-review` onto code pushed after its pass, toggle draft state —
> that re-fires `ready_for_review`. REST `gh api -X PATCH .../pulls/<n> -f
> draft=true` silently no-ops (returns `draft: false`, no error); use GraphQL
> `convertPullRequestToDraft` then `markPullRequestReadyForReview`.

That works, but it has a cost the note doesn't mention: `ready_for_review` also
re-fires `dagster-cloud-deploy`. On this branch that meant a **full rebuild and
redeploy of all five code locations** — not just `kipppaterson` — because the
branch touches `src/teamster/libraries/dlt/powerschool/assets.py`, and
`src/teamster/libraries/dlt/**` is in the `pull_request` paths of every
`deploy-prod-*.yaml`:

```text
deploy-prod-kippcamden.yaml
deploy-prod-kippmiami.yaml
deploy-prod-kippnewark.yaml
deploy-prod-kipppaterson.yaml
deploy-prod-kipptaf.yaml
```

Five multi-arch Docker builds plus five deploys, to obtain a code review.

## The alternative

Posting a comment containing `@claude` triggers `claude.yaml`
(`issue_comment: types: [created]`), which reviews the current head with no
deploy side effect at all. I used that for the second round on this PR and it
behaved identically from a review standpoint.

One caveat if this does get written down: `claude.yaml` listens for `created`
only, so **editing** a comment to add the mention does nothing — it has to be a
new comment.

## Why I'm not just changing it

Two reasons this is your call rather than mine. The draft-toggle note may be
deliberate — `ready_for_review` re-running the branch deployment could be the
point on a PR where you want a fresh deployment anyway, and the tag method
skips that. And the two aren't strictly equivalent: the toggle re-fires the
`claude-code-review` workflow, while the tag fires `claude.yaml`, which are
separate workflows with separate configs.

If you want it changed, the line is `CLAUDE.md:332`. Happy to send a one-line PR,
or leave it — no action needed from me either way.

cbini (approved):

PR #4893, opened 2026-08-17
fix(dbt): sort DeansList family contacts extract parents before emergency

cbini:
## Summary & Motivation

...add an explicit `ORDER BY` to `rpt_deanslist__family_contacts` so
Parent1/Parent2 contacts sort ahead of Emergency contacts, per student.

DeansList support confirmed (via the linked Zendesk thread) that their
importer displays a student's contacts in the row order they arrive in the
file, with no independent sort of their own. The extract had no `ORDER BY` at
all, so row order was whatever BigQuery happened to return — currently
Emergency-first, the reverse of last year and the reverse of what schools
expect (parents/guardians at the top).

## AI Assistance

Investigated and drafted entirely by Claude Code from a DeansList support
email thread, including locating the model, confirming the terminal-extract
`ORDER BY` exception has precedent elsewhere in `extracts/`, adding the sort,
running the model's dbt unit tests, and `trunk check --force`. Human-directed
review before merge.

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Address
      why.

### dbt

- [x] No new model — properties file already existed, description updated to
      document the new row order.
- [ ] No exposure change needed (Dagster asset `deanslist/contacts_txt`
      already documented in the model description).
- [ ] Not a breaking change — no columns added, renamed, or removed.
- [ ] No external source changes.

## CI checks

- [ ] **Trunk** — passes locally (`trunk check --force` on both changed
      files, clean).
- [ ] **dbt Cloud** — this PR touches `kipptaf` directly, so CI will actually
      build/test this model (not a no-op run). Model's 4 unit tests + 7 data
      tests pass locally against `--target dev --defer <prod>`.
- [ ] **Dagster Cloud** — not triggered; no Dagster code changed, dbt-only
      change.

### DIFF
--- a/src/dbt/kipptaf/models/extracts/deanslist/properties/rpt_deanslist__family_contacts.yml
+++ b/src/dbt/kipptaf/models/extracts/deanslist/properties/rpt_deanslist__family_contacts.yml
@@ -9,8 +9,10 @@ models:
       `emergency_4`). `ContactType` distinguishes `Parent1` / `Parent2` /
       `Emergency` so DeansList can scope automated guardian messaging, and every
       contact carries its real Finalsite relationship to the student. Column
-      names match the DeansList import template headers exactly. Delivered by
-      the Dagster asset `kipptaf/extracts/deanslist/contacts_txt`.
+      names match the DeansList import template headers exactly. Rows are
+      ordered per student with parents before emergency contacts, since
+      DeansList displays contacts in the file's row order. Delivered by the
+      Dagster asset `kipptaf/extracts/deanslist/contacts_txt`.
     data_tests:
       # The source grain is (student, contact_slot), but the DeansList template
       # fixes the header set so the slot cannot be projected — name plus
--- a/src/dbt/kipptaf/models/extracts/deanslist/rpt_deanslist__family_contacts.sql
+++ b/src/dbt/kipptaf/models/extracts/deanslist/rpt_deanslist__family_contacts.sql
@@ -74,3 +74,7 @@ inner join
     on c.student_number = s.student_number
     and c._dbt_source_project = s._dbt_source_project
     and s.enroll_status = 0
+-- DeansList's importer displays contacts in the order this file lists them per
+-- student (no independent sort on their end), so Parent1/Parent2 must precede
+-- Emergency here or families see emergency contacts ranked above parents.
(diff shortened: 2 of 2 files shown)


cbini:
Verdicts on the claude-review findings, verified locally before acting on any of them:

**Important 1 — does the ORDER BY survive the Dagster asset's bare `SELECT *`?** CONFIRMED as a real general concern, but empirically verified safe here. Read `libraries/extracts/assets.py`'s `construct_query()` and confirmed it issues a literal `SELECT * FROM {schema}.{table}` with no `.order_by()` for `type: schema` configs. Built the model into a dev-schema view and confirmed the region-source `source()` resolution left it empty under `--target dev` (as documented in this repo's own dbt CLAUDE.md), so instead I wrapped the compiled model SQL (refs resolved to prod) in a bare `select * from (...) where StudentID = X` — structurally identical to the asset's query — against a real student with 2 parents + 4 emergency contacts. Result: rows came back Parent1, Parent2, then 4x Emergency, in that order. No code change needed; this is genuinely working as intended, and the same architecture is already shipping successfully for 3 other `extracts/` models.

**Important 2 — no test locks the ordering.** CONFIRMED. Stripped the `ORDER BY` locally and reran the model's 4 unit tests — all 4 still passed, proving dbt's unit-test `expect:` comparison doesn't check row order at all. I then tried adding a singular test computing `row_number() over (partition by StudentID order by <same sort key>)` to lock it — but that test is tautological: it recomputes its own order from `ContactType` values, so it passes even against the *current, unfixed* prod view (verified: ran it deferred against prod, which still lacks this fix, and it passed anyway). dbt has no mechanism to assert a query's actual delivered/physical row order, so there's no test I can write here that would be more than decorative. Declined — no test added. Manual verification (Important 1) is the closest thing to real coverage this claim can get without a live post-merge check of the delivered `contacts.txt`.

**Minor 1 — simplify `if(...)` to a boolean sort key.** Applied as-is; confirmed the `(<bool expr>) asc` idiom is already established at `rpt_tableau__student_course_grades.sql:809,817`. Also had to add explicit `asc` to every column in the `ORDER BY` (sqlfluff AM03 — direction on all columns or none) — `trunk check --force` is clean and all 11 model tests still pass.

PR #4894, opened 2026-08-17
fix(dbt): dedupe double-scheduled Miami Tech Focus enrollments in rpt_littlesis__enrollments

### LINKED ISSUE #4891: fix(dbt): rpt_littlesis__enrollments duplicates every double-scheduled Miami Tech enrollment
`rpt_littlesis__enrollments` fails its
`dbt_utils.unique_combination_of_columns` on `(section_id, student_id)` with
**299 duplicate groups / 598 rows**. Every one of them is `school_id = 14`,
KIPP Miami Tech.

The same 299 groups are present in prod
(`kipptaf_extracts.rpt_littlesis__enrollments`) right now, so this is a live
defect rather than a CI artifact. It has simply never been caught: the model
only entered a `state:modified+` selection for the first time on #4882, whose
wide sweep pulled it into CI.

## Root cause

`int_focus__schedule` schedules Miami Tech students into the same course period
twice — once as a full-year row and once term-specific. For AY2026,
`schoolid = 14`:

```text
rows                                1,599
distinct (student_id, course_period_id)  1,300
difference                            299
distinct marking_period_id              2
rows with null marking_period_id     1,248
```

The 299 difference is exactly the duplicate-group count. Focus uses
`marking_period_id = 0` as a full-year sentinel, which `int_focus__schedule`
normalizes to null, so a student carrying both a full-year row and a
term-specific row for one course period yields two rows that
`rpt_littlesis__enrollments` does not collapse.

The PowerSchool branch is unaffected — it carries no Miami rows at all.

## Introduced by

The Focus branch added in #4799. Miami Tech is a new school for AY2026, so this
could not have surfaced before this year regardless.

## Consequence

Little SIS receives two enrollment records for each affected
student-and-section, across 26 sections. Worth confirming with whoever owns the
Little SIS sync whether duplicate records are rejected, deduplicated, or
silently doubled downstream.

## Blocking

The `dbt Cloud` check on #4882 is red solely because of this test. That PR
changes nothing this model reads — it reaches
`base_powerschool__course_enrollments`, not the enrollment vertical #4882
touches — and the decision there was to leave the test at `severity: error`
rather than downgrade it, so this needs resolving or explicitly accepting
before #4882 can merge on a green check.

## Fix options

1. Collapse full-year and term-specific rows in the Little SIS Focus branch,
   preferring the full-year row where both exist.
2. Fix the scheduling in Focus, if the double entry is a data-entry problem
   rather than an intentional pattern. Needs an Ops conversation.
3. Downgrade the test to `severity: warn` with a `TODO` pointing here, matching
   the precedent set for `base_powerschool__course_enrollments` under #3915.

Option 1 is the smallest correct change if the double scheduling is legitimate.

Refs #4799
Refs #4882

cbini:
# Pull Request

## Summary & Motivation

...fix `rpt_littlesis__enrollments` failing its `dbt_utils.unique_combination_of_columns` test on `(section_id, student_id)` for every Miami Tech (`school_id = 14`) student scheduled into a course period both as a full-year row and a term-specific row. Focus's own schedule table carries both rows legitimately; this collapses them to the full-year row in the Little SIS Focus branch so each student gets one enrollment per section, matching the PowerSchool branch's grain.

Verified against prod data in a worktree: school 14 goes from 1,599 rows / 1,300 distinct `(section_id, student_id)` pairs down to 1,300 rows, and the uniqueness test passes with `--favor-state` forcing upstream resolution to prod.

## AI Assistance

Claude Code diagnosed the root cause (already scoped in #4891), chose the fix option (collapse in the Little SIS Focus branch, per user direction), implemented the `dbt_utils.deduplicate` CTE, and verified against prod data. No human-written code in this PR.

### General

### dbt

- [x] No new model — no properties file needed
- [x] No exposure change needed
- [x] Not a breaking change — output grain narrows (dedup), no column changes

## CI checks

### DIFF
--- a/src/dbt/kipptaf/models/extracts/littlesis/rpt_littlesis__enrollments.sql
+++ b/src/dbt/kipptaf/models/extracts/littlesis/rpt_littlesis__enrollments.sql
@@ -72,10 +72,11 @@ with
         where academic_year = {{ var("current_academic_year") }}
     ),
 
-    -- Focus schedules some students into the same course period twice: once
-    -- as a full-year row (marking_period_id null, the FY sentinel already
-    -- normalized upstream) and once term-specific. Keep the full-year row
-    -- where both exist so Little SIS gets one enrollment per section.
+    -- Focus schedules some students into the same course period twice --
+    -- either two full-year rows or two rows in the same term, never one of
+    -- each. Keep one row per (student_id, course_period_id), preferring a
+    -- full-year row (marking_period_id null) if a term-specific row is ever
+    -- present too, so Little SIS gets one enrollment per section.
     focus_schedule as (
         {{
             dbt_utils.deduplicate(
--- a/src/dbt/kipptaf/models/extracts/littlesis/rpt_littlesis__enrollments.sql
+++ b/src/dbt/kipptaf/models/extracts/littlesis/rpt_littlesis__enrollments.sql
@@ -54,7 +54,8 @@ with
        per course period) and are keyed on the prefixed student id that
        stg_people__student_logins already assigns Google accounts to.
     */
-    focus_schedule as (
+    -- trunk-ignore(sqlfluff/ST03): referenced via dbt_utils.deduplicate below
+    focus_schedule_raw as (
         select
             academic_year,
             schoolid,
@@ -69,6 +70,20 @@ with
             cast(course_id as string) as course_id,
         from {{ ref("int_focus__schedule") }}
         where academic_year = {{ var("current_academic_year") }}
+    ),
+
+    -- Focus schedules some students into the same course period twice: once
+    -- as a full-year row (marking_period_id null, the FY sentinel already
+    -- normalized upstream) and once term-specific. Keep the full-year row
+    -- where both exist so Little SIS gets one enrollment per section.
+    focus_schedule as (
+        {{
+            dbt_utils.deduplicate(
+                relation="focus_schedule_raw",
+                partition_by="student_id, course_period_id",
+                order_by="(marking_period_id is not null) asc",
+            )
+        }}
     ),
 
     /*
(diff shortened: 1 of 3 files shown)


cbini:
Verified each finding against the data before responding.

**Important #1 (order_by has no tiebreaker for same-side ties)** — Confirmed, and it wasn't an edge case: all 299 duplicate pairs are same-side ties (225 pairs both null `marking_period_id`, 74 pairs both non-null on the same term). The original `order_by` never actually broke a real tie. Fixed by keying the dedup on `end_date` instead of a `student_schedule_id` tiebreaker — every pair turned out to be one same-day-superseded stint (`start_date = end_date`) alongside the current open one (`end_date` null), so `end_date` is the real, 100%-deterministic signal, not an arbitrary stable pick.

**Important #2 (`int_students__teacher_grade_levels` double-counts on the same duplication)** — Confirmed, and materially worse than "plausible": verified against prod that every one of the 299 duplicate pairs has non-null `teacher_id` on both rows (it's a course-period-level attribute, identical across the pair), so all 299 were double-counted in `student_count` today. Impact wasn't marginal either — several grade-9 teachers were off by 60-190 students, traced to ~15-20 large sections where a schoolwide schedule correction duplicated up to 31 students at once in a single section.

Fixed by moving the dedup from `rpt_littlesis__enrollments` up into the kipptaf-level `int_focus__schedule` wrapper — its only two consumers (verified via `grep -rln 'ref("int_focus__schedule")'`) — rather than patching each caller. Added the uniqueness test this intermediate model was missing per `src/dbt/CLAUDE.md`'s per-layer requirements, now that the grain is actually enforced. `rpt_littlesis__enrollments` no longer carries its own dedup CTE.

**Minor (undocumented dual-row behavior)** — Addressed via the properties.yml description update alongside the new test, rather than a standalone note, since the model now enforces the grain instead of just describing it.

All four commits are pushed; CI is green through the latest.

PR #4895, opened 2026-08-17
feat(dbt): add last_ccr_date to rpt_tableau__kfwd_career_launch_survey

### LINKED ISSUE #4892: feat(dbt): add last_ccr_date to rpt_tableau__kfwd_career_launch_survey
Add a `last_ccr_date` column to `rpt_tableau__kfwd_career_launch_survey`: the contact note date of the alumnus's most recent Career Conversation record in Salesforce.

Definition: `max(date)` over `stg_kippadb__contact_note` rows whose `subject` matches `^CC\d` — the same CC1 through CC5 classifier that `int_kippadb__contact_note_rollup` uses for the `cc1_count` through `cc5_count` columns already in this model. CCDM, CCC, and the 2020 CC CTE subjects are out of scope.

The existing `career_conversations` CTE cannot supply this. `int_kippadb__contact_note_rollup` pivots `min(contact_date)` per contact / academic year / CC step, so its latest CC date is earlier than the true most recent note for 83 of 461 alumni with CC records.

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will add a `last_ccr_date` column to
`rpt_tableau__kfwd_career_launch_survey` — the contact note date of each
alumnus's most recent Career Conversation record in Salesforce.

Definition: `max(date)` over `stg_kippadb__contact_note` rows whose `subject`
matches `^CC\d` — the same CC1 through CC5 classifier that
`int_kippadb__contact_note_rollup` already uses for this model's `cc1_count`
through `cc5_count` columns. The new date and those counts therefore always
agree, including subjects like `CC1 attempt` and `CC1 follow-up`, which that
regex folds into CC1 upstream. CCDM, CCC, and the 2020 CC CTE subjects are out
of scope.

The `career_conversations` CTE already in this model cannot supply the value.
`int_kippadb__contact_note_rollup` pivots `min(contact_date)` per contact /
academic year / CC step, so its latest CC date is earlier than the true latest
note for 83 of 461 alumni with CC records. Hence the read straight from
`stg_kippadb__contact_note`, which this model already does for four other
`stg_kippadb__*` models.

Additive only — no renames or removals. The model has no downstream dbt
consumers; it feeds the KIPP Forward Tableau workbook via the existing exposure.

### Validation

- `dbt build` on the dev target with `--favor-state` against prod upstreams: view
  created, contract enforced, `PASS=1 WARN=1 ERROR=0`.
- The WARN is pre-existing, not introduced here: 127 duplicate
  `(contact_id, response_id)` groups in prod and 127 in the dev build, with
  identical row counts (4,641 both). The new left join is grouped on its join
  key, so it cannot fan out.
- Values: 455 of 4,321 contacts carry a date, and all 455 equal the max CC note
  date in staging — 0 spurious, 0 missing, and 0 cases of `cc_total_count > 0`
  with a null date. Range 2020-03-08 through 2026-08-07.
- `trunk check --force` clean on both changed files.

## AI Assistance

Claude Code wrote the SQL and column documentation and ran the validation above.
Human-directed: the request, the column name, and the decision to scope CC
records to the numbered CC1 through CC5 series rather than including CCDM or CCC
notes.

One open naming question for the reviewer: `CCR` elsewhere in this project means
College and Career Readiness (`ebrw_ccr_benchmark`, `ccr_course`), so
`last_career_conversation_date` may read less ambiguously in the Tableau field
list. Shipped as requested; happy to rename.

## Self-review

Dagster and Docs sections omitted — dbt-only change.

### General

### dbt

- [x] Properties file updated — `last_ccr_date` declared in
      `rpt_tableau__kfwd_career_launch_survey.yml` with `data_type: date` and a
      description.
- [x] Exposure — the existing `models/exposures/tableau.yml` KIPP Forward
      exposure already depends on this model; no new exposure needed.
- [x] **Breaking change?** No. Additive column only, no renames or removals, and
      no downstream dbt models reference this one.
- [x] External sources — none added or modified, so no
      `stage_external_sources` run is required.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Closes #4892


anthonygwalters:
## Review responses

Verdict per finding. Fix pushed in c84c32d4.

### 1. Important — `^CC\d` classifies more subjects than the counts cover → **accepted, fixed**

Correct, and the failure mode is slightly wider than described: besides a step outside 1-5, a year-prefixed subject like `CC2026 kickoff` also matches `^CC\d` and extracts `CC2026`, which the rollup's pivot drops. The filter now mirrors `int_kippadb__contact_note_rollup` exactly — same `regexp_extract(subject, r'^CC\d+')`, same `CC1`-`CC5` domain as its pivot `IN (...)` list — so the date and the `cc*_count` columns are wired to the same note set by construction rather than by coincidence.

Ran the verification query (against `stg_kippadb__contact_note` — `kipptaf_kippadb.contact_note` isn't a relation; the raw Salesforce object is the `kippadb` source): **0 rows**, so no divergence in current data and this is a no-op on today's values. Rebuilt and re-validated after the change: same 455 contacts with a date, all 455 still equal the max CC note date in staging, and the reverse direction you correctly noted was untested now checks out — **0** contacts with a non-null `last_ccr_date` and `cc_total_count = 0`.

### 2. Minor — `CCR` collides with College and Career Readiness → **acknowledged, author's call**

Your grep matches mine (`ccr_course` / `ccr_teacher_name` / `ccr_section`, `ebrw_ccr_benchmark`). The column name was specified in the original request, and I raised this ambiguity in the PR description before review; it stays open for the requester rather than being renamed unilaterally. It's a one-line change in the SQL and the properties yml if they want `last_career_conversation_date`.

### 3. Nice to have — collapse the two CTEs → **declined**

The proposed snippet puts `regexp_contains(subject, ...)` in `WHERE`, which `src/dbt/CLAUDE.md` prohibits: "No row-level calculations in `WHERE`. No functions applied to table columns — precompute as a named column." The two-CTE split exists for that rule, not for style. After the fix the filter is a plain column comparison (`where cc_step in (...)`), which complies; collapsing it would reintroduce the violation.

### 4. Informational — model lacks `config.meta.contains_pii` → **confirmed, out of scope here**

Verified: the properties yml has no `meta:` block at all, and sibling extracts `rpt_tableau__fresh_dashboard_qc` and `rpt_tableau__gradebook_es_comments` do carry the tag, so the convention exists in this directory. Agreed it isn't this PR's job — and per `src/dbt/CLAUDE.md`, tagging scope for a PII-heavy model (direct identifiers only vs direct plus indirect) needs confirming with the data team first, so it needs its own change. Happy to open a tracking issue.

---

Note for anyone waiting on a re-review: `claude-review` triggers on `opened` / `ready_for_review` only, so it will not re-run against c84c32d4 without a draft toggle.

PR #4900, opened 2026-08-18
revert(extracts): drop es_attended and ms_attended from the contact feed

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will remove the `es_attended` and `ms_attended`
columns from `rpt_gsheets__student_contact_info`, backing out #4812 in full.
Those two prior-level banding columns pushed the Student Contact Info Google
Sheet over its file size limit, so the feed returns to its pre-#4812 shape.

Both files are restored to their merge-base blobs, which makes this diff the
exact inverse of #4812 (verified line-for-line, +2/-115):

- removes the `grade_band_enrollments`, `grade_band_ranked`,
  `most_recent_school_by_grade_band`, and `enrollments` CTEs, plus the
  `base_powerschool__student_enrollments` ref
- removes the `es_attended` and `ms_attended` columns and their descriptions
- removes the two `expression_is_true` tests that pinned the banding invariant

The identically named columns on `int_extracts__student_enrollments`, and the
Tableau reports that read them, are untouched. Those band by `school_level` and
are a separate lineage from the grade-banded columns removed here.

Worth a follow-up look, not blocking this revert: two string columns over
~11.4k rows is roughly 23k cells, so if a spreadsheet is up against a hard
ceiling, this buys modest headroom. Since four distinct spreadsheets read this
view, the first step on that follow-up is confirming which one actually hit the
limit, then auditing what else is wide in it.

Local verification:

- `dbt build --select rpt_gsheets__student_contact_info` passes, which is what
  runs contract enforcement — 48 columns, down from 50
- `dbt parse --no-partial-parse` on both sides drops exactly the two intended
  test nodes and nothing else (804 to 802 tests); model node count unchanged
- the reverted view returns 11,439 rows, identical to prod, so the revert is
  row-neutral, with zero leftover band columns in `INFORMATION_SCHEMA`
- `trunk check --force` clean on both files

## AI Assistance

AI-assisted. Claude Code identified the originating PR, performed the revert as
a merge-base blob restore, and ran the verification above. Human-directed: the
decision to back the columns out, and the branch and issue strategy.

### General

Skipped — no Dagster changes.

### dbt

- [x] Properties file updated alongside the model — the two column blocks and
      the two model-level tests are removed from
      `rpt_gsheets__student_contact_info.yml`
- [x] Exposure — `gsheets_student_contact_info` already exists and still points
      at this model; no exposure change is needed for a column removal
- [x] **Breaking change?** Yes — this removes two columns from a contracted
      `rpt_` model, and that removal is the entire point of the change. Correcting
      my earlier wording in this box: this view feeds **four** Sheets exposures,
      each a distinct spreadsheet, not one — `gsheets_student_contact_info`,
      `miami_demographic_update_sy25`, `miami_demographic_update_sy24`, and
      `gsheets_student_logins`. All four lose the two columns. All four also only
      gained them one week ago in #4812, so no dashboard or analysis has had time
      to build on them, and exposures don't declare columns so none needs editing.
      **Rollback plan:** revert this PR, which restores #4812 verbatim.
- [ ] External sources — not applicable, no external source added or modified.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

### DIFF
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__student_contact_info.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__student_contact_info.yml
@@ -1,21 +1,5 @@
 models:
   - name: rpt_gsheets__student_contact_info
-    data_tests:
-      # Pin the prior-level banding so it stays a standing invariant rather than
-      # a point-in-time check. Bands are grades K-4 elementary, 5-8 middle, 9-12
-      # high; a student is blank on every band at or below the one they are
-      # currently in. Both tests are scoped by grade, so a grade_level outside
-      # the K-12 domain is skipped rather than flagged.
-      - dbt_utils.expression_is_true:
-          arguments:
-            expression: es_attended is null
-          config:
-            where: grade_level in ('K', '1', '2', '3', '4')
-      - dbt_utils.expression_is_true:
-          arguments:
-            expression: ms_attended is null
-          config:
-            where: grade_level in ('K', '1', '2', '3', '4', '5', '6', '7', '8')
     columns:
       - name: student_number
         data_type: int64
@@ -113,25 +97,3 @@ models:
         data_type: string
       - name: home_language
         data_type: string
-      - name: es_attended
-        data_type: string
-        description: >-
-          Most recent KIPP school the student attended while in grades K through
-          4. Blank while the student is currently in grades K through 4, so the
-          column reads as prior history rather than the current placement.
-          Banded by the grade the student was enrolled in rather than by the
-          school's designated level, so a school serving grades outside its
-          designation is attributed to the right band. Distinct from the
-          identically named column on int_extracts__student_enrollments, which
-          bands by school level and includes the student's current school.
-      - name: ms_attended
-        data_type: string
-        description: >-
-          Most recent KIPP school the student attended while in grades 5 through
-          8. Blank while the student is currently in grade 8 or below, so the
-          column reads as prior history rather than the current placement.
-          Banded by the grade the student was enrolled in rather than by the
-          school's designated level, so a school serving grades outside its
-          designation is attributed to the right band. Distinct from the
-          identically named column on int_extracts__student_enrollments, which
-          bands by school level and includes the student's current school.
--- a/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__student_contact_info.sql
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__student_contact_info.sql
@@ -1,69 +1,3 @@
-with
-    -- Band every enrollment by the GRADE the student was in, not by the school's
-    -- school_level. An increasing number of schools are off-model (e.g. an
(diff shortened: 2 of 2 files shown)


anthonygwalters:
Verdicts on the three notes, verified against the code rather than taken at face value.

**1. Missing `contains_pii` tag — confirmed, declining in this PR.**

Verified: `grep -c contains_pii` on the properties file returns 0, and it also
returns 0 on the merge-base blob (`d858aaa93^1`), so this predates #4812 and is
untouched by the revert. The finding is accurate.

Declining to fix here for two reasons. First, scope — this PR is a byte-exact
inverse of #4812, and adding a tag would make it something other than a revert.
Second, the root `CLAUDE.md` PII reference says tagging scope for a PII-heavy
model (direct identifiers only, versus direct plus indirect) should be confirmed
with the user before applying, so it is not a drive-by fix. Context for whoever
picks it up: only 4 of the 61 properties files in
`models/extracts/google/sheets/properties/` carry the tag today, so this is a
directory-wide gap, not a one-model oversight — which matches the root
`CLAUDE.md` note that `contains_pii` is authoritative but incomplete.

**2. Four exposures, not one — confirmed, and I corrected the PR body.**

Verified, and it is worth more than an informational note: the four exposures
point at four **distinct** spreadsheet URLs, not four references to one sheet.

| Exposure | Line |
| --- | --- |
| `gsheets_student_contact_info` | 152 |
| `miami_demographic_update_sy25` | 218 |
| `miami_demographic_update_sy24` | 231 |
| `gsheets_student_logins` | 297 |

My breaking-change box said "the sheet is the only consumer," which was wrong. I
have rewritten that box and the follow-up paragraph to name all four and to say
the first step on the size follow-up is confirming which spreadsheet actually hit
the ceiling. The revert itself is unaffected — exposures don't declare columns,
and all four sheets only gained these columns a week ago in #4812.

**3. Local test-count and row-count claims — already run, restating the evidence.**

These were run before the PR was opened; the reviewer had no warehouse access, so
it is a fair thing to ask for. Following the "Verifying a test-removal PR"
convention exactly, I parsed both sides with `--no-partial-parse` (main's blobs
temporarily checked into the same worktree so only these two files differed) and
diffed the `resource_type == 'test'` node names:

```text
--- test: main=804 revert=802 delta=-2
  REMOVED: test.kipptaf.dbt_utils_expression_is_true_rpt_gsheets__student_contact_info_es_attended_is_null.1315347dcb
  REMOVED: test.kipptaf.dbt_utils_expression_is_true_rpt_gsheets__student_contact_info_ms_attended_is_null.c12647c18d
--- model: main=793 revert=793 delta=0
parse-main:   refs=['base_powerschool__student_enrollments', 'int_extracts__student_enrollments'] columns=50
parse-revert: refs=['int_extracts__student_enrollments'] columns=48
```

So exactly the two intended test nodes disappear and nothing else does. Row and
column parity came from the built dev view against prod: 11,439 rows on both
sides, 48 columns, and 0 rows in `INFORMATION_SCHEMA.COLUMNS` matching
`es_attended`/`ms_attended`.

Nothing in this review changes the code. The only edit it produced is the PR body
correction in note 2.

Issue #3992, opened 2026-05-21
fix(dbt): unexpected fafsa_status_mismatch_category value in int_extracts__student_enrollments (1 row)


cbini:
## Context

`accepted_values_int_extracts__student_enrollments_fafsa_status_mismatch_category__No_issues` test warns on 1 row in CI: a `fafsa_status_mismatch_category` value falls outside the accepted-values list (currently only `'No issues'`).

Scope filter: `where grade_level != 99 and academic_year = 2025`.

## Acceptance

- Identify the offending row + value.
- Decide: expand accepted-values list to include the new category, or fix upstream so the value is normalized to `'No issues'`.

## Surfaced in

PR #3985 CI run.

PR #4902, opened 2026-08-18
feat(dbt): DIBELS benchmark completion tracking for Camden

### LINKED ISSUE #4896: feat(dbt): DIBELS benchmark completion tracking for Camden
## What and why

KIPP Sumner Academy and Lanning Square Primary track DIBELS benchmark testing
completion in a hand-maintained
[Google Sheet](https://docs.google.com/spreadsheets/d/1yc-fihaeVw2t2ggyeeOcXcqlLbZiLmSgNIn6d43IJqA).
One tab per window, rows are grade and advisory, one column per morning holding
percent complete, plus a hand-typed list of students still outstanding and which
measure each one is missing. Someone rebuilds it every school morning.

Replace it with modeled data. Two things are needed:

1. **A daily completion rate** per advisory for each day of a benchmark window.
2. **A list of students who have not completed their measures**, naming the
   measures they still owe.

Scope: Camden, grades K-8, Benchmark windows. Progress monitoring is out of
scope as a deliverable, though the models carry PM rows where the expectations
sheet defines them.

## Status

Built and pushed to PR #4902. Three models:

| Model | Grain |
| --- | --- |
| `int_tableau__dibels_benchmark_completion_daily` | student x expected measure x calendar day, all regions |
| `rpt_gsheets__dibels_benchmark_completion_daily` | advisory x day, Camden |
| `rpt_gsheets__dibels_benchmark_measures_student` | outstanding measure x student x day, Camden |

An earlier attempt was removed in favor of this rewrite. The design notes that
previously filled this issue described that abandoned approach and were deleted
deliberately.

## Decisions made during the build

**The denominator comes from enrollment, not from assessment records.** This was
the single biggest correction. Building the population from
`int_amplify__all_assessments` meant a student only existed once Amplify
reported a row for them, so early in a window the population was weighted
toward students who had already finished. Measured on Camden BOY 2026: 965 of
1,703 enrolled students were invisible and the reported completion rate read
**77.9% when the true figure was 13.2%**. The population now comes from
`int_extracts__student_enrollments_subjects`, gated so a student is only held to
a window their enrollment overlaps.

**Reporting is complete versus not complete. There is no "in progress."** A
measure either carries a `client_date` or it does not, so at measure grain there
is no intermediate state to detect. A partly-finished student could in principle
be inferred from a mix of finished and outstanding measures, but across the
entire completed BOY 2025 Camden window only **two** students ever showed that
mix. This needs managing with the stakeholder, since the sheet being replaced
implies three states.

**Every calendar day in a window is reported, in session or not.** An earlier
version joined the PowerSchool calendar to keep only in-session days. Decided
against — the curve simply runs flat across weekends and closures.

**No student names.** The worklist carries `student_number`, grade, school and
advisory but no names, so no `contains_pii` tagging is required. Names are
available on `int_extracts__student_enrollments_subjects` if the stakeholder
needs them, which would change the PII posture of the destination tab.

## Open

- Self-contained and out-of-district students are **not** excluded. The school's
  own tracker omits the self-contained advisories and
  `rpt_tableau__dibels_dashboard` excludes both at the reporting layer, so the
  two reports here probably should too.
- **Sheet volume ceiling.** Projected off the completed AY2025 season, a full
  year of Camden reaches roughly 9.5M cells against Google Sheets' 10M cap. The
  aggregate is small; the student-level worklist is what grows. Likely needs
  scoping to the current window rather than the full year.
- The worklist has no exposure entry yet — undecided whether it joins the
  existing DIBELS source sheet or gets its own.

## Destination

Per the `rpt_gsheets__` convention: a sheet in the IMPORTRANGE Sources folder
connected directly to BigQuery, then a user-friendly IMPORTRANGE copy in the
Reports folder. The exposure carries the source sheet only. Source sheet for the
daily rate already exists as
[DIBELS Daily Completion](https://docs.google.com/spreadsheets/d/1Peu2YG3kDKV9qCT6h9ez-StDqsrwNUZ5focjBg_xStM).

<details>
<summary>For Claude</summary>

Facts measured against production. Kept because they are expensive to
rediscover and constrain any future change here.

### Why the owed-measure signal is row absence

- **An unscored row does not mean an outstanding measure.** In
  `int_amplify__all_assessments`, a row with a null `measure_standard_score`
  carries a real `measure_standard_level` and belongs exclusively to students
  Amplify marks Complete. These rows cluster by grade and season into whole
  measure sub-batteries outside the grade's standard band — the same measures
  for every affected student. They read as not-applicable, not as owed. This
  killed the original design.
- **A measure a student genuinely still owes has no row at all.** All 57 Camden
  AY2025 Partial student-seasons are short rows against the expectations sheet,
  152 rows missing in total; every Complete student-season has a full row set.
  This is why the population has to be spined from enrollment — no filter over
  assessment rows can manufacture a student who has none.
- **`completion_status` is admin-grain**, one value per student per window, so
  it cannot resolve individual measures. Unused by the current models.

### Amplify feed behavior

- **Amplify stamps a date only on completed benchmarks.** Every Partial student
  has no `client_date`. This is the mechanical reason "in progress" is not
  reportable.
- **`client_date` is constant per student per window** — exactly one distinct
  value, never more.
- **`completion_status` is null for AY2023 and AY2024** because those years come
  from the frozen API archive, which lacks the column; AY2025 onward comes from
  the SFTP feed. The feed changed, not the vendor. Re-enabling the api layer for
  a region would introduce rows with a null status.

### Grain and join hazards

- `assessment_grade_int` must be in any join key against Amplify — one Camden
  student carries a spurious off-grade summary row, so the pick is otherwise
  non-deterministic.
- A student can hold more than one enrollment stint overlapping one window, and
  the stints can differ in school and advisory (14 Camden students AY2025+, 5
  differing in both). `int_extracts__student_enrollments_subjects` filtered to
  `discipline = 'ELA'` is one row per student, verified 1,703 rows for 1,703
  students, so the current models need no dedupe — but that invariant is load
  bearing.
- **`advisory` is the ELA section identifier.** `cc_section_number` is the grade
  prefixed onto the advisory, so no course-enrollment join is needed.
- The expectations sheet's include flags are `BOOL` holding only NULL or false,
  never true. Filter with `is null`, matching all six existing call sites.
- `count(*) over (...)` for the expected measure count must be evaluated
  **before** the date spine is generated, or it counts measure-days and inflates
  by the window length (180 instead of 6 for grade 0).

### Reconciliation figures

Camden BOY 2025, at window close: 1,533 students complete, 107 with nothing ever
recorded, 2 partial. Camden BOY 2026 as of 2026-08-21: 8,570 measures expected,
1,130 finished (13.2%), 226 of 1,647 students complete, 7,440 outstanding
measure rows — which reconciles exactly against 8,570 minus 1,130.

</details>

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will replace a hand-maintained DIBELS completion
spreadsheet with modeled data: a shared intermediate plus two Camden reports —
a daily completion rate per advisory, and a worklist of the measures each
student still owes.

Someone currently rebuilds that sheet every school morning, typing in percent
complete per advisory and the list of students still outstanding.

The important thing in this PR is where the denominator comes from. An earlier
approach built the student population out of assessment records, which meant a
student only existed in the report once Amplify had reported a row for them.
Early in a testing window that leaves the population weighted toward students
who have already finished, and the completion rate reads far too high — on
Camden BOY 2026 it reported **77.9% complete when the real figure was 13.2%**,
with 965 of 1,703 enrolled students missing from the report entirely. The
population now comes from enrollment, so a student who has not been assessed
appears from the first day of the window as outstanding.

## Reviewer Notes

**There is no "in progress" state, deliberately.** A measure either has a
`client_date` or it does not. A partly-finished student could in theory be
inferred from a mix, but across the whole completed BOY 2025 Camden window only
two students ever showed one. The sheet being replaced implies three states, so
this needs managing with the stakeholder rather than modeling.

**Every calendar day is reported, in session or not.** An earlier version joined
the PowerSchool calendar to keep in-session days only; decided against, so the
curve runs flat across weekends and closures.

**The intermediate is network-wide; the reports filter to Camden.** This follows
the existing DIBELS dashboard pattern rather than pinning the shared model to
one region.

**Two things are knowingly not done.** Self-contained and out-of-district
students are not excluded, though the school's own tracker and
`rpt_tableau__dibels_dashboard` both omit them. And a full year of the
student-level worklist projects to roughly 9.5M cells against Google Sheets' 10M
cap, so it likely needs scoping to the current window before it is wired up.

### General

### dbt

- [x] Properties file included for all three new models, with descriptions on
      every column and `data_type` for contract enforcement
- [x] Uniqueness test on every model's grain, verified against production before
      the test was written rather than after
- [x] Exposure already present for `rpt_gsheets__dibels_benchmark_completion_daily`
      and still resolves. The worklist has no exposure entry yet — see Reviewer
      Notes
- [x] **Not a breaking change.** The one renamed model
      (`rpt_gsheets__dibels_benchmark_completion_daily` to
      `int_tableau__dibels_benchmark_completion_daily`) has never been deployed;
      confirmed no production relation exists under either name
      is needed
- [x] `dbt build` green locally including all data tests
- [x] `trunk check --force` clean on all six files

## CI checks

- [ ] **Trunk**
- [ ] **dbt Cloud**
- [ ] **Dagster Cloud** — not expected to trigger, no Dagster paths changed

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The author drove every design decision; Claude
implemented, and its own first attempt contained the assessment-record
denominator bug described above, which measurement caught rather than review.

### Verification performed

`dbt build --select int_tableau__dibels_benchmark_completion_daily+ --target dev
--favor-state --defer` — 3 view models, 9 data tests, `PASS=12 ERROR=0`.

Grain verified before writing the uniqueness tests, not after. The 2,626,673 /
85,548 row zero-duplicate check was run against the earlier single-model design,
whose key included `grade`; the intermediate's key is the eight columns
`(academic_year, region, student_number, assessment_type, period, round_number,
expected_measure_standard, calendar_day)` and is verified by its passing
uniqueness test, since the enrollment-spined population changed the row set.

Figures reconcile exactly: 8,570 measures expected minus 1,130 finished equals
the 7,440 outstanding rows in the worklist; 226 complete plus 1,421 outstanding
equals 1,647 students. `students_expected` is 1,647 rather than the full 1,703
because the enrollment-overlap gate correctly drops 56 students.

### Bugs found and fixed during the build, worth not reintroducing

`count(*) over (partition by ...)` for the expected measure count must be
evaluated **before** the date spine is generated. Evaluated in the same SELECT
as the `cross join unnest`, its partition holds one row per measure-day, so
grade 0 read 180 instead of 6 and the round-completion flag could never fire.
This is why `expectations` and `spine` are two CTEs rather than one.

The date comparison belongs in the flag, not the join. With `calendar_day =
client_date` as a join predicate, `NULL` never matches, so every outstanding
measure was silently excluded — which is precisely the population the worklist
exists to show. `finished_this_day` preserves that same-day semantics as a
column instead, verified to reproduce the join's daily counts exactly.

`row_number() ... order by client_date asc` puts NULLs first in BigQuery, ranking
unfinished measures ahead of finished ones (37 rows across 22 students). That
window function was removed entirely rather than patched, since counting after
the join makes it redundant.

The expectations sheet's `assessment_include` is `BOOL` holding only NULL or
false, never true, so `is null` is the correct filter — matching all six
existing call sites, including `int_amplify__all_assessments` itself.

### Deliberate non-issues

`select *,` inside the `rolled` CTE follows the precedent in
`rpt_tableau__seat_tracker_snapshot`, which does the same inside an `rpt_` CTE.
The final SELECT enumerates columns per the no-`select *`-in-final rule.

Omitting `assessment_type` from the expected-measure-count partition is safe:
zero groups span both Benchmark and PM in any academic year, verified.

`finished_round` compares against the grade's full expected count, so a student
missing assessment records can never read as finished. That is intentional and
makes round-level completion immune to the row-absence problem, unlike the
measure-level rate.

</details>


GabyRangelB:
Verdicts on all four findings, verified against production before acting. Fixes in d2db63f05.

## 1. Non-deterministic row pick — fixed

Confirmed, and live today rather than theoretical. Ran the verification you suggested against the exact filtered population:

| | Count |
| --- | --- |
| Camden students with more than one enrollment row, AY2025 and later | 14 |
| ...whose rows differ in both `schoolid` and `advisory` | **5** |
| ...whose rows share an `entrydate` | 0 |
| Max rows for one student | 3 |

So for 5 students the school and advisory attribution could flip between runs. Fixed with `order by s.entrydate desc, s.exitdate desc`, which is fully deterministic on this data (no duplicate shares an entrydate) and semantically lands on where the student ended the window rather than an arbitrary stint.

Also checked one thing your query did not cover: whether `discipline = 'ELA'` fans out across `iready_subject`. It does not — zero of the 14 have more than one subject, so the duplicates are genuinely multiple stints and nothing else.

Re-verified after the fix: season status counts are **unchanged** (BOY 1,520 / 18 / 33 of 1,571; MOY 1,537 / 25 / 33 of 1,595; EOY 1,528 / 9 / 37 of 1,574). Only the attribution is now stable, which is the correct outcome — the dedup always picked one row per student, so totals were never at risk.

## 2. Missing grain-projection annotation — fixed

Convention and precedent both confirmed (20 files carry it). Added the standard two-line form above `school_windows`.

## 3. Incomplete date-range overlap join — declined here, with evidence

Your reading of the predicate is correct: it covers 2 of the 4 interval-overlap cases, and a stint contained entirely within a window would drop. But it does not fire on this data. I tested both the narrow and the precise formulation:

```text
academic_year  season  stint_fully_inside_window  overlaps_but_currently_dropped
2025           BOY     0                          0
2025           MOY     0                          0
2025           EOY     0                          0
2026           BOY     0                          0
```

The second column is the strict test — rows where a true overlap (`entrydate <= end_date and exitdate >= start_date`) is satisfied but the current two-condition OR is not. Zero across every year and season.

Declining for this PR for two reasons beyond the zero impact. It is copied verbatim from `int_students__dibels_participation_roster` and `rpt_tableau__dibels_dashboard`, so fixing it only here makes three models disagree on the same join for no observable gain. And changing the predicate changes the population, which would invalidate the reconciliation figures this PR is verified against — I would rather not swap verified numbers for unverified ones to close a gap that currently affects nobody.

Worth doing as its own change across all three models, where the shared predicate can be fixed once and all three re-reconciled together. Happy to open that if you want it tracked.

## 4. Reuse note — acknowledged, no action

Agreed on the overlap, and agreed it does not warrant a rewrite. If a third consumer of the windows-by-grade join appears, extracting it becomes worthwhile; at two it would be indirection for its own sake.

## Unrelated: the dbt Cloud red on the first run

For the record, since it is visible in the history. It failed on `dbt_utils_unique_combination_of_columns` on `int_extracts__student_enrollments`, a model this branch never touches. Prod carries the same 2 duplicate groups right now, so it is pre-existing; it only entered `state:modified+` because the branch was 18 commits behind main. Merging main cleared it and the run is green with zero warnings.

That prod duplicate is real and untracked, though — the test is `severity: error`, and the only nearby issue (#3992) covers a different test on that model. Flagging rather than filing, since it is outside this PR.

GabyRangelB:
Heads-up for reviewers: `08ae58dcc` carries two gradebook-audit exposure fixes that are **not** DIBELS work. They ride along here deliberately rather than in a separate PR, since they were found while reviewing this PR's own exposure and are two lines in the same file.

| Exposure | Fix |
| --- | --- |
| `rpt_gsheets__gradebook_audit_student_flags` | URL pointed at the Reports IMPORTRANGE copy; repointed to the source sheet the model actually loads into |
| `rpt_gsheets__gradebook_audit_template` | Added the missing `ref("stg_powerschool__u_expectations")` — the sheet reads current expectations alongside the template rows |

Found by checking where each exposure URL actually lives in Drive. Every source sheet in the IMPORTRANGE Sources folder is named after its model, which made the mismatch visible: the student-flags exposure pointed at a file titled "Gradebook - Student Flags" sitting in Reports, while `rpt_gsheets__gradebook_audit_student_flags` exists in IMPORTRANGE Sources.

`dbt parse --no-partial-parse` clean, so the added ref resolves.

GabyRangelB:

It was substantially restructured since it was last reviewed — the previous
design was removed entirely and rebuilt, so please review the current three
models rather than any earlier state.

Areas where a second opinion is most useful:

1. **The population spine.** `int_tableau__dibels_benchmark_completion_daily`
   builds its expected population from
   `int_extracts__student_enrollments_subjects` filtered to `discipline = 'ELA'`,
   `enroll_status in (0, 2, 3)`, `grade_level <= 8`, then gates a student to a
   window with `start_date between entrydate and exitdate or end_date between
   entrydate and exitdate`. That gate is copied from
   `int_students__dibels_participation_roster`. Is it right, and does the
   one-row-per-student invariant it relies on actually hold outside Camden?

2. **Duplicated logic versus a new dependency.** Those filters and that date gate
   now exist in both this intermediate and
   `int_students__dibels_participation_roster`. Refing the roster instead was
   considered and rejected because its grain is student x season x round with no
   measures and no days. Is that the right call, or should the shared population
   be factored out?

3. **The two window functions.** `expected_measure_count` is a `count(*) over`
   evaluated in the `expectations` CTE specifically so it runs before the
   `cross join unnest` date spine — inside the spine it would count measure-days
   and inflate by the window length. `measures_finished_to_date` is a
   `sum(if(...)) over` evaluated after the join, in a separate `rolled` CTE
   because BigQuery has no lateral column aliases. Are both partitions correct?

4. **Model placement.** The intermediate sits in
   `models/extracts/tableau/intermediate/` as `int_tableau__`, but its only
   current consumers are two `rpt_gsheets__` reports. Flag if that is the wrong
   home.

5. **Volume.** A full year of `rpt_gsheets__dibels_benchmark_measures_student`
   projects to roughly 9.5M cells for Camden against Google Sheets' 10M cap.
   Worth a view on whether it should be scoped to the current window in the
   model rather than in the Sheet.

Known and intentional, so no need to flag: there is no in-progress state (a
measure either carries a `client_date` or it does not); every calendar day is
included whether or not school was in session; the intermediate is network-wide
while the reports filter to Camden; and self-contained / out-of-district
students are not yet excluded.

GabyRangelB:
## Verdicts

Verified each finding against the code and against production before replying.
Two are fixed, one is corrected, three are confirmed-but-open, and two are
deferred to the author with reasons. Nothing dismissed.

### PM probe-eligibility gate — CONFIRMED, real and latent, not yet fixed

This is a correct catch and the reasoning about why the reconciliation numbers
could not have caught it is exactly right. Verified both halves:

PM rows do survive the `assessment_include is null and pm_goal_include is null`
filter — AY2025 carries 258 `BOY->MOY` and 211 `MOY->EOY` rows through it. And
AY2026 currently holds only `Benchmark`/`BOY` (192 rows), which is why every
figure quoted in the PR body is Benchmark-only.

Sized against AY2025 Camden: 1,082 of 1,768 ELA-enrolled students (61.2%) are
BOY-probe-eligible by the roster's own derivation
(`dibels_boy_composite in ('Below Benchmark', 'Well Below Benchmark')`). So once
a PM window opens, the model would treat all 1,768 as expected — a 686-student
overstatement of the PM population, and 686 students listed in the worklist for
measures they were never meant to sit.

Left for the author to decide rather than patched, because it is a design call:
replicate the roster's two-branch eligibility derivation here, or gate PM out of
this model entirely and let the roster own PM. Flagged as the top open item.

### Interval-overlap gate — CONFIRMED logically, zero current impact

The logic gap is real: with
`window.start between entrydate and exitdate or window.end between entrydate and exitdate`,
an enrollment span falling entirely inside a window matches neither branch.
`entrydate <= end_date and exitdate >= start_date` is the correct form.

Measured impact today: **zero** Camden AY2025 students have an enrollment span
entirely inside the BOY window. So it is a latent correctness issue, not a
current data defect. Not yet fixed — and worth noting the fix would make this
model diverge from `int_students__dibels_participation_roster`, which has the
same gap, so it is really an argument for fixing both.

### One-row-per-student not guaranteed by construction — CONFIRMED, no code change

Accurate. `int_extracts__student_enrollments_subjects` filtered to
`discipline = 'ELA'` happens to be one row per student (1,703 rows for 1,703
students, Camden AY2026) but nothing enforces it. That invariant is already
recorded as load bearing in issue #4896, and the `unique_combination_of_columns`
test is the guard — a regression fails the build rather than silently
double-counting. No change made.

### Duplicated population logic — fair, author's call

Agreed that the filter set plus the date gate now exist in two places and that a
fix to either has to land in both. Worth noting the PM finding above is already
an instance of exactly that risk. Factoring the population out is a reasonable
suggestion; leaving it to the author since it changes the dependency graph.

### Window functions — agreed, no action

Independent confirmation of both partitions is useful, particularly the point
that `measures_finished_to_date` does not need to be an ordered running sum
because the cumulation already lives in `finished_measure`.

### Model placement — factually right, deferred

The factual claim checks out: `int_students__dibels_participation_roster` does
live in `models/students/intermediate/`, not under `extracts/tableau/`, so the
precedent does cut against the current location, and `+schema: tableau` does
describe a model with no Tableau consumers today.

Deferring rather than moving it: the `int_tableau__` placement was an explicit
instruction from the author, who may intend Tableau consumption. Relayed to them
as a decision rather than actioned unilaterally.

### Volume scoping — partially fixed, and the remaining gap is worse than stated

Agreed the scoping belongs in the model, not the Sheet. Both reporting views now
carry `calendar_day <= current_date(...)`. On the worklist that removed
**150,108 of 220,376 rows** — 68% of the model was future-dated, asserting a
measure was outstanding on a date that had not happened.

But this does not solve the volume problem, and the numbers should be on record:
the cap scales with elapsed days, so by the close of BOY the worklist returns to
roughly 220K rows / 4.2M cells, and three windows across a full year land near
12.5M cells — past the 10M Sheets cap. A history limit (latest day only, or a
rolling window) is still needed before this is wired to a sheet. Tracked as open.

### Minor: "9-column key" — CONFIRMED, my error, corrected

Right, the intermediate's key is 8 columns. The PR body has been corrected, and
while fixing it a second inaccuracy in the same sentence was also corrected: the
2,626,673 / 85,548 zero-duplicate figures were measured against the earlier
single-model design whose key included `grade`, not against the current
enrollment-spined intermediate. That distinction was being blurred.

Issue #3834, opened 2026-05-07
refactor(dbt): align new diverging assessments strategies for DIBELS PM


GabyRangelB:
## Background

DIBELS 8 (Dynamic Indicators of Basic Early Literacy Skills) is a literacy
assessment created by the University of Oregon and delivered by Amplify mCLASS.
KIPP TAF uses it to assess literacy knowledge and growth for students in grades
K-8 via two assessment types:

**Benchmark (BM)** — three snapshots per year (BOY, MOY, EOY) that track growth
across administrations and years.

**Progress Monitoring (PM)** — administered between BOY and MOY, and between MOY
and EOY. Targets students whose composite score on the preceding benchmark was
Below Benchmark or Well Below Benchmark.

## Scope decisions

Three decisions set the shape of this work:

1. **PM reporting covers SY26-27 forward only.** T&L does not need historical PM
   data carried into the new model. No retrofit, no backfill.
1. **BM is untouched.** Benchmark keeps full history plus current year, and is
   out of scope except where PM and BM share a model.
1. **K-2 and 3-8 are separate calculation paths, but should land in one
   reporting table.** ~~K-2 keeps the in-house PM goals process; 3-8 moves to
   Amplify aimline.~~ **Superseded — see _Update: both methods run K-8_ at the
   end.** Academics asked for both methods across K-8, so neither replaces the
   other and the split is by method, not by grade band. The calculations do not merge. Reporting should, though:
   the target is a single table where K-2's internally-calculated goal is
   surfaced in the same goal column as the aimline goal, with a rollup column
   alongside the grade-band split — the same shape Finalsite uses for
   returner / new / all. Keeping one table makes board and Foundation questions
   aggregate cleanly. Ideal, not a blocker if it proves impossible. This also
   voids most of the original deprecation list — see *Models to keep* below.

## What is changing: the Amplify aimline file

The PM model for grades 3-8 moves to
`stg_amplify__mclass__sftp__pm_student_summary_aimline`.

- Already staged and flowing. The Dagster asset is defined only in kippnewark,
  but the file itself is network-wide; the kipptaf union model rewrites
  `_dbt_source_relation` per school through the location crosswalk, so all four
  regions resolve from the one drop.
- Probe-level detail: one row per student, measure and probe attempt within a PM
  period.
- Carries individualized student goals (`goal`), expected aimline score by date
  (`aimline_value_by_date`), and goal status (`aimline_status`, either `Below`
[...]

Issue #3552, opened 2026-03-30
fix(dbt): pm_student_summary surrogate key missing probe_number


cbini:
The surrogate key in `stg_amplify__mclass__sftp__pm_student_summary` uses `student_primary_id_studentnumber, school_year, pm_period, measure` but the data has multiple probes per combination. `probe_number` needs to be added to the surrogate key, matching the fix applied to `pm_student_summary_aimline` in #3463.

Additionally, both PM models have ~31 duplicate rows where the same student/period/measure/probe has multiple rows with different `device_date` and `measure_standard_score` values — likely re-assessments or duplicate exports. Need to determine if deduplication (e.g., latest `device_date` per probe) is appropriate.

PR #4903, opened 2026-08-18
fix(dbt): tolerate Amplify PM measure label drift that emptied SY2526 DIBELS progress monitoring

### LINKED ISSUE #4901: fix(dbt): Amplify PM measure label drift nulls measure and empties SY2526 DIBELS progress monitoring
## Summary

All SY2526 (AY2025) DIBELS progress-monitoring data is missing from the DIBELS dashboard. The rows survive ingestion, but `measure` is NULL on every one of them, so the INNER JOIN that assigns each probe to an expected assessment drops all of them.

## Root cause

Amplify changed the `measure` value format in the `dibels8_PM_*.csv` SFTP export, from short codes to long labels:

```text
ORF       ->  Reading Fluency (ORF)
NWF-CLS   ->  Letter Sounds (NWF-CLS)
Maze      ->  Reading Comprehension (Maze)
```

`stg_amplify__mclass__sftp__pm_student_summary` translates short codes **to** those long labels with a `CASE` that has no `ELSE`, so every arm now misses and the column nulls out:

```sql
case
    measure
    when 'Maze' then 'Reading Comprehension (Maze)'
    ...
    when 'PSF' then 'Phonemic Awareness (PSF)'
end as measure,
```

The break wiped the whole year retroactively rather than degrading from a start date. The PM export is a single full-year file written to one school-year partition and overwritten on every sync (confirmed: exactly one GCS object under `_dagster_partition_key=2025-2026`), so the first new-format pull replaced every short-code row at once.

Nothing in this repository caused the format change. This is vendor drift.

## Impact, measured in prod

| Layer | AY2025 | AY2024 |
| --- | --- | --- |
| `stg_amplify__mclass__sftp__pm_student_summary` (Newark + Paterson) | 67,984 rows, 100% `measure` NULL | populated |
| `int_amplify__all_assessments`, PM branch | 0 rows | 19,469 |
| `int_amplify__pm_met_criteria` | 0 rows | 13,943 |
| `rpt_tableau__dibels_dashboard` PM rows carrying a score | 0 of 44,860 | 13,943 of 16,959 |

The drop happens at `int_amplify__all_assessments`:

```sql
and p.measure = e.expected_measure_standard
```

The dashboard still renders 44,860 PM scaffold rows for AY2025 because that scaffold is built from Benchmark composites, which are unaffected. Every score and met-goal column on those rows is blank.

## Secondary casualty

`measure_name_code` for Maze is now `Maz` rather than `Comprehension` (7,811 rows). The `when 'Maze'` arm misses and the expression falls through to the `substr(measure, strpos(measure, '(') + 1, 3)` branch. That branch is why `measure_name_code` otherwise survived the change: it parses the parenthetical and works under both formats.

## Fix, as shipped

Tracked in #4903. The shape changed several times during review — this reflects
the branch tip, not the original proposal or any intermediate state.

Both models normalize `measure` in a first CTE and derive everything downstream
from the normalized value, so vendor format tolerance lives in exactly one place
per model:

- **The short-code mapping is retained, with an `else measure` passthrough.** It
  is inert on today's PM feed, which arrives pre-converted to long labels, but it
  makes a vendor reversion a no-op rather than an outage. An intermediate
  revision removed it on the theory that `accepted_values` would make a
  reversion fail loudly; that was wrong, because no staging test blocks
  downstream propagation — see Guard.
- **`measure_name_code` is derived from the NORMALIZED measure**, not the raw
  source column. Previously it read the raw value, which is why it needed both a
  short-code and a long-label arm for the same measure, and why deleting either
  regenerated the `Maz` defect on 7,811 Maze rows. That trap is now structurally
  absent rather than documented.
- **No fallback on `measure_name_code`.** The former
  `substr(measure, strpos(measure, '(') + 1, 3)` silently truncated an
  unrecognized code to three characters, which could collide with a real one and
  pass every guard — a hypothetical `ORFX` became `ORF` and merged into an
  existing measure group unnoticed. An unrecognized measure now yields NULL.
- **The aimline model runs the identical process.** All three CASE blocks are
  byte-identical between the two models. This also fixed an `else measure` on
  aimline's `measure_name_code` that leaked whole labels into a code field, and a
  `measure_name` that mapped Maze to `Reading Comprehension` where the PM model
  and the sheet both say `Comprehension`.
- **The intervention measure is mapped explicitly in both models** — aimline
  normalizes the bare `(DEC-IW)` to the long label the PM feed sends.
- **`assessment_edition` added to the PM surrogate key.** The key also now
  hashes the normalized `measure`, so the same logical row keys identically
  under either vendor format instead of rehashing on a format flip.

The two exports send different formats for the same measures, which is the fact
the fix reconciles:

| code | `pm_student_summary` sends | `pm_student_summary_aimline` sends |
| --- | --- | --- |
| ORF | `Reading Fluency (ORF)` | `ORF` |
| Maze | `Reading Comprehension (Maze)` | `Maze` |
| DEC-IW | `Irregular Words (DEC-IW)` | `(DEC-IW)` |

All 8 measures now emit identical values across both models on all three fields.

Verified in dev across both consuming districts: **39,662 Newark PM rows rejoin**
the expected-assessments spine, versus 0 in production. An earlier revision of
this issue cited 34,820 — that was an undercount, because the verification query
used raw `device_date` while all 7,382 Maze rows carry it NULL and the kipptaf
union coalesces it to `sync_date`.

## Guard

`not_null` plus `accepted_values` on **both** `measure` and `measure_name_code`,
in **both** models, at error severity, and **unconditionally** — no `where`
scope on any of the four.

An intermediate revision scoped the `measure` guards
`where assessment_edition = 'DIBELS 8th Edition'`. That was removed: it made the
guard depend on a second vendor-controlled string that nothing tests, so a
rename or null there would match zero rows, report PASS, and silently disable the
protection — the same failure shape this issue exists to fix.
`Irregular Words (DEC-IW)` is in the accepted domain as a result, so every row
must carry a known measure regardless of edition.

**These tests alert; they do not block.** `dbt_table_automation_condition()`
never gates on asset-check outcome, so no staging test in this repository stops
cross-code-location propagation. A bad value still reaches kipptaf. That is why
the mapping is retained rather than relying on the guard alone, and why NULL is
the chosen failure value for an unrecognized measure — it is the most inert, since
a NULL partition key can only group with other NULLs rather than merging into a
real measure group.

## Notes

- Only `pm_student_summary` flipped format. The aimline feed still carries short
  codes, which is why its mapping is load-bearing while the PM one was not.
- The Literacy Dashboard extract refresh schedule is RUNNING on `0 5 * * *`, so
  the published dashboard is not masked by a frozen extract.
- No re-pull is needed. The current file holds the entire year; only the labels
  changed.
- Independent of the aimline migration in #3834, which targets AY2627 and does
  not restore SY2526.
- A filter excluding mCLASS Intervention rows from `int_amplify__all_assessments`
  was added and then reverted as dead code — the PM branch already inner-joins
  `measure` to the sheet, and no intervention measure exists there. Measured
  across all years: 1 intervention row, 0 surviving the join.
- Two pre-existing gaps surfaced during verification, neither fixed here: the
  AY2024 sheet has no PM entry for Maze (1,404 rows drop at the join), and the
  sheet carries no PM rows at all before AY2024. Both are Ops decisions.

GabyRangelB:
# Pull Request

## Summary & Motivation

...restore SY2526 DIBELS progress-monitoring data, which is currently missing
network-wide.

Amplify changed the `measure` value format in the `dibels8_PM_*.csv` SFTP
export, from short codes to long labels:

```text
ORF       ->  Reading Fluency (ORF)
NWF-CLS   ->  Letter Sounds (NWF-CLS)
Maze      ->  Reading Comprehension (Maze)
```

`stg_amplify__mclass__sftp__pm_student_summary` translated short codes **to**
those long labels with a `CASE` that had no `ELSE`, so every arm missed and the
column nulled out. `int_amplify__all_assessments` inner-joins that column
against `expected_measure_standard`, and NULL matches nothing — so every SY2526
PM row was dropped.

The break wiped the year retroactively rather than degrading from a start date.
The PM export is a single full-year file written to one school-year partition
and overwritten on every sync (confirmed: exactly one GCS object under
`_dagster_partition_key=2025-2026`), so the first new-format pull replaced every
short-code row at once. Nothing in this repo caused the change; it is vendor
drift.

### Impact, measured in prod before the fix

| Layer | AY2025 | AY2024 |
| --- | --- | --- |
| `stg_amplify__mclass__sftp__pm_student_summary` (Newark + Paterson) | 67,984 rows, 100% `measure` NULL | populated |
| `int_amplify__all_assessments`, PM branch | 0 rows | 19,469 |
| `int_amplify__pm_met_criteria` | 0 rows | 13,943 |
| `rpt_tableau__dibels_dashboard` PM rows carrying a score | 0 of 44,860 | 13,943 of 16,959 |

The dashboard still renders 44,860 PM scaffold rows for AY2025 because that
scaffold is built from Benchmark composites, which are unaffected. Every score
and met-goal column on those rows is blank.

### What changed

The two Amplify PM exports send **different value formats for the same
measures**, which is what this PR reconciles:

| code | `pm_student_summary` sends | `pm_student_summary_aimline` sends |
| --- | --- | --- |
| ORF | `Reading Fluency (ORF)` | `ORF` |
| Maze | `Reading Comprehension (Maze)` | `Maze` |
| DEC-IW | `Irregular Words (DEC-IW)` | `(DEC-IW)` |

- **`measure` on the PM model now passes through.** Amplify sends that feed
  pre-converted to long labels, so the short-code mapping matched nothing and was
  removed as dead code. The `accepted_values` guard below carries the alarm
  instead: if the vendor reverts to short codes, the build fails loudly rather
  than silently converting and hiding the change.
- **Maze long-form arm** on `measure_name_code`. The `when 'Maze'` arm missed
  and the expression fell through to `substr(measure, strpos(measure, '(') + 1, 3)`,
  yielding `Maz` instead of `Comprehension` on 7,811 rows. Every other long
  label resolves correctly through that `substr` branch.
- **The aimline model now runs the PM model's process** for `measure`,
  `measure_name_code` and `measure_name` — the three CASE blocks are byte
  identical between the two files. Its `measure_name_code` previously ended in
  `else measure`, which leaked whole labels into a code field, and its
  `measure_name` mapped Maze to `Reading Comprehension` where the PM model and
  the expected-assessments sheet both say `Comprehension`. Deriving
  `measure_name` from `measure_name_code` required restructuring it to the PM
  model's CTE-plus-outer-select shape, since BigQuery has no lateral column
  aliases; most of that file's diff is re-indentation.
- **The intervention measure is mapped explicitly in both models.** Aimline
  normalizes the bare `(DEC-IW)` to the long label PM sends, and both derive
  `measure_name` of `Irregular Words` from the `DEC` code.
- **`assessment_edition` added to the PM surrogate key**, so an intervention row
  cannot collide with a DIBELS row at the same student, period and measure.
  Grain is unchanged today — 30,906 distinct keys before and after — and no key
  overlaps prod, confirming the input landed. Its only consumer uses the key as a
  partition key, so new hash values do not shift behavior.
- **Maze long-form arm** on `measure_name_code`. The `when 'Maze'` arm missed
  and the expression fell through to `substr(measure, strpos(measure, '(') + 1, 3)`,
  yielding `Maz` instead of `Comprehension` on 7,811 rows. Every other long
  label resolves correctly through that `substr` branch, so Maze is the only one
  needing an arm.
- **Guard tests on `measure` in BOTH models** at error severity — `not_null`
  plus `accepted_values` over the known label domain. `accepted_values` is the
  one that fires: it flags an unrecognized or renamed vendor value. Note it does
  NOT flag a format flip on aimline, by design — the mapping converts short codes
  to the same output, so nothing broke and nothing should alarm. On the PM model,
  which no longer maps, a format revert IS an unrecognized value and does fail.

  `accepted_values` is scoped `where assessment_edition = 'DIBELS 8th Edition'`.
  Investigating the one unmapped measure revealed that a second Amplify product
  shares this export: 1 of the 67,984 SY2526 rows carries
  `assessment_edition = 'mCLASS Intervention'` with a measure of
  `Irregular Words (DEC-IW)`, which is not a DIBELS 8 measure. Scoping keeps the
  guard asserting the real invariant — every DIBELS PM row carries a known
  DIBELS measure — instead of blessing a foreign product's measure as valid, and
  stops a future intervention measure failing an error-severity test over rows
  that never reach a dashboard. Full write-up in #4901.

  An earlier revision of this PR also filtered the intervention rows out of
  `int_amplify__all_assessments`. That was reverted as dead code: the PM branch
  already inner-joins `measure` to the expected-assessments sheet, and no
  intervention measure exists there. Measured across all years and both
  editions — 1 intervention row, **0** surviving the measure join. The sheet is
  KTAF-controlled, so a label collision is not a live risk.

### Verification

Built in dev against **both** consuming districts, with the source external
re-staged into the personal dev schema first:

- Newark 61,387 rows, Paterson 6,597 rows, `measure` fully populated, both
  guard tests PASS in each district.
- Maze restored to `measure_name_code` / `measure_name` of
  `Comprehension` / `Comprehension`, matching the AY2024 shape.
- **39,662 Newark PM rows rejoin** the expected-assessments spine across all 7
  measures, versus **0** in production today.
  - An earlier revision of this PR cited 34,820. That was an undercount: all
    7,382 Maze rows carry a NULL `device_date` (only `sync_date` is populated)
    and the kipptaf staging union coalesces the two, but the original
    verification query used raw `device_date` and so excluded every Maze row.
    The corrected figure includes 4,842 Maze rows.
- Field-alignment check across every year — `measure`, `measure_name_code` and
  `measure_name` match the expected-assessments sheet exactly for all 7 SY2526
  measures, and the mapping is stable across AY2022-AY2025.
- **All 8 measures now emit identical values across the two models** on all
  three fields, despite arriving in different formats.
- The aimline surrogate-key uniqueness warning is pre-existing duplication
  (#3552), byte-identical to prod at 32 duplicate keys — not introduced here.
- The `not_null` guard would fail with 61,387 violations against production's
  current data, confirming it catches this exact regression.
- `trunk check --force` clean on all changed files.

### Notes for reviewers

- **The aimline feed did not change format.** It still carries short codes.
  Only `pm_student_summary` flipped. This corrects an earlier assumption that
  the drift was global.
- **The Literacy Dashboard extract refresh schedule is RUNNING** on `0 5 * * *`,
  so the published dashboard is not being masked by a frozen extract.
- **No re-pull is needed.** The current file holds the entire year; only the
  labels changed.
- This is independent of the aimline migration in #3834, which targets AY2627
  and does not restore SY2526.
- Two adjacent gaps were left alone deliberately, as neither causes this bug and
  both resolve correctly under the current vendor format via pass-through: there
  is no `LNF` arm in the `CASE` despite `Letter Names (LNF)` appearing in
  `expected_measure_standard`, and no `DEC-IW` arm for the newly-observed
  `Irregular Words (DEC-IW)` measure. Both would only bite if Amplify reverted
  to short codes.
- This model has no uniqueness test, contrary to the staging-layer convention.
  Pre-existing and out of scope here — its surrogate key omits `probe_number`
  (see #3552), so adding one would likely fail.

## AI Assistance

Human-directed: the report that SY2526 PM data was missing, and the decisions to
branch via worktree, anchor to an issue, and include a regression guard.

AI-assisted (Claude Opus 5): root-cause investigation and the prod queries
behind every number above, the fix, the guard tests, dev verification in both
districts, and this write-up.

## Self-review

### General

No Dagster changes.

- [x] Include (or update) an
      for all models consumed by a dashboard — no new consumer; the DIBELS
      dashboard exposure is unchanged.
- [x] **Breaking change?** No. Value-only change plus two added tests. The
      column set is unchanged, so kipptaf consumers compile as-is and corrected
      values land on the next prod rebuild.
      with **`--target staging`** — not applicable; `sources-external.yml` is
      untouched. The external was re-staged only into a personal dev schema for
      verification.

## CI checks

- [ ] **Trunk** — passes. Clean locally via `trunk check --force` on both
      changed files.
- [ ] **dbt Cloud** — note that this PR touches only the `amplify` source
      package. dbt Cloud CI builds `kipptaf` alone and the column set is
      unchanged, so it selects zero models and goes green **trivially, not as
      validation**. The local two-district `dbt build` above is the real
      pre-merge validation.
- [ ] **Dagster Cloud** — no branch deployment is expected; the `pull_request`
      paths exclude `src/dbt/**`. On merge, the `push` paths of both
      `deploy-prod-kippnewark.yaml` and `deploy-prod-kipppaterson.yaml` list
      `src/dbt/amplify/**` (verified), so both consuming districts will deploy.

### Post-merge verification

`code_version` changes on the staging model, so the automation-condition sensor
should rematerialize it in both districts and cascade to the kipptaf table
`int_amplify__mclass__pm_student_summary`. Confirm AY2025 PM rows return to
`int_amplify__all_assessments` and that scores repopulate on
`rpt_tableau__dibels_dashboard` before closing #4901.

## Troubleshooting

Closes #4901

---
  - https://app.asana.com/0/0/1217590144627830


GabyRangelB:
## Responses to the review findings

Verdicts on each item from the earlier review, plus a summary of what has landed
since — that pass saw commit `8b4919863`, and the branch has moved substantially.

### 1. Duplicated short-code to long-label CASE across two files — accepted, deferred

Valid, and **more apt now than when raised**. The two models' `measure`,
`measure_name_code` and `measure_name` CASE blocks are now byte-identical
(verified programmatically, whitespace-normalized), so the duplication the
finding described is exact rather than approximate.

Deferring anyway, deliberately. This PR is a production restoration for a live
outage, and a shared macro changes the call sites in both models plus anything
that later reuses it. Worth doing as a fast-follow rather than widening a fix
that is already four files. Recorded so it is not lost.

### 2. `accepted_values` domain completeness against historical data — verified, non-issue

You flagged this as unverifiable without warehouse access. Checked:

- The external holds **only school year 2025-2026**. The PM export is a single
  full-year file written to one partition and overwritten on every sync, so
  there is no older-schema history for the test to trip on.
- **8 distinct values**, no nulls, across both districts.
- Seven are in the accepted list. The eighth, `Irregular Words (DEC-IW)`, is an
  mCLASS Intervention row and is excluded by the test's
  `where assessment_edition = 'DIBELS 8th Edition'` scope.

No false trip is possible on current data. Both guard tests pass in both
districts.

### 3. Downstream `expected_measure_standard` alignment — verified, non-issue

Also checked directly, for every academic year and all three fields:

- All 7 SY2526 measures match the sheet exactly on `measure`,
  `measure_name_code` and `measure_name`.
- The mapping is stable across AY2022 through AY2025 — every measure resolves to
  the same code and name in each year it appears.

The check did surface two **pre-existing** gaps, neither caused by this PR and
neither fixed here: the AY2024 sheet has no PM entry for Maze (1,404 rows drop at
the join), and the sheet carries no PM rows at all before AY2024. Both are
recorded on #4901 as Ops decisions.

### 4. Info items — acknowledged

The missing uniqueness test on `stg_amplify__mclass__sftp__pm_student_summary`
is still missing, and that is worth stating plainly rather than leaving implied.
Its surrogate key is the reporting grain, not a uniqueness key — roughly half the
rows share one by design, because the dedupe happens downstream in
`int_amplify__all_assessments`. That was settled in #3552 and closed as "we
dedupe downstream", so a `unique` test here would fail by construction. The
sibling aimline model does carry one, at warning severity, over pre-existing
duplication from the same issue.

Your read on the column-description convention matches ours — this project does
not apply it anywhere today, so it was not applied to untouched columns.

## What changed since your pass

- The `int_amplify__all_assessments` filter you saw was **reverted as dead code**
  — the PM branch already inner-joins `measure` to the sheet and no intervention
  measure exists there. Measured: 1 intervention row, 0 surviving the join.
- The **aimline model now runs this model's process** for the three measure
  fields, fixing an `else measure` that leaked whole labels into a code field and
  a Maze `measure_name` that disagreed with both the PM model and the sheet.
- The **short-code mapping was removed from the PM model** as inert — that feed
  arrives pre-converted — with `accepted_values` carrying the alarm instead, so a
  vendor revert fails loudly rather than converting silently.
- **`assessment_edition` added to the PM surrogate key.**
- The guard tests are now on **both** models.

@claude please re-review the branch as it now stands. The areas most worth a
fresh look are the aimline restructure into the CTE-plus-outer-select shape, and
whether removing the short-code mapping from the PM model while relying on
`accepted_values` to flag a vendor revert is the right trade between availability
and visibility.

GabyRangelB:
@claude - we've addressed your concerns. please check

GabyRangelB:
## Follow-up — an independent pass found gaps both reviews missed

Your last verdict was ready-to-merge, and I agree with the reasoning. But before
merging I ran a fresh full-diff review with no memory of the prior passes, and it
surfaced three real gaps. Two are now fixed in `bcdc751f1`; the rest are recorded
below.

### Fixed — `measure_name_code` was unguarded, and it is the column that broke

The defect this PR exists to fix was `measure_name_code` reading `Maz` on 7,811
Maze rows. Every guard added so far sits on `measure` — the column that failed
**loudly**, as NULL. The column that failed **silently** had no test at all.

That matters because `int_amplify__pm_met_criteria` partitions
`avg(met_measure_standard_goal) over (partition by ..., measure_name_code, ...)`,
so a wrong three-character code regroups every met-goal calculation with nothing
surfacing.

Added `not_null` plus `accepted_values` over the known code domain
(`Composite`, `Comprehension`, `DEC`, `LNF`, `NWF`, `ORF`, `PSF`, `WRF`) to both
models, at error severity. Deliberately **unscoped** by `assessment_edition`,
unlike the `measure` guards — intervention rows produce a real code and should be
covered, and it means this particular guard does not depend on a second
vendor-controlled string staying stable.

### Correction — the truncation hazard is only PARTLY mitigated, not fixed

I originally posted this as fixed. That was wrong, and the correction matters
because it leans on a mechanism I flag as weak further down.

The `substr(measure, strpos(measure, '(') + 1, 3)` expression is **unchanged** —
it still truncates. Harmonizing aimline onto it replaced that model's previous
`else measure`, so for a feed sending short codes with no parenthesis the
fallback becomes `substr(measure, 1, 3)`. The new `measure_name_code` guard only
catches truncations landing OUTSIDE the code domain:

| hypothetical new code | resulting code | `measure_name_code` guard | `measure` guard |
| --- | --- | --- | --- |
| `DAZE` | `DAZ` | caught | caught |
| `ORFX` | `ORF` | **passes** | caught |
| `NWFZ` | `NWF` | **passes** | caught |
| `Reading Something (ORFX)` | `ORF` | **passes** | caught |

The collision case — a new code sharing a three-character prefix with an existing
one, which is the hazard by definition — produces a value that IS in the domain
and therefore passes. Those rows are caught by the `measure` guard instead, which
is `where assessment_edition = 'DIBELS 8th Edition'` — the edition-scoped guard
listed below as able to self-disable. So the coverage is real but conditional.

A genuine fix would replace the `substr` fallback with an explicit code per known
measure and no fallback, letting an unrecognized value go NULL so `not_null`
catches it unconditionally, independent of the edition scope. That is a larger
change than this PR should absorb; recording it with the fast-follow items rather
than claiming it is done.

### Fixed — the raw-read behavior you flagged is now documented at the column

Your observation that the surrogate key hashes the **raw** vendor `measure`
rather than the mapped output is correct, and I confirmed it directly rather than
by inference:

```sql
select * replace ('LONG' as m), m as bare_ref from (select 'SHORT' as m)
-- star expansion -> LONG, bare reference -> SHORT
```

The same applies to `measure_name_code`, which is the more consequential half:
it is the *only* reason both the short-code and long-label Maze arms are
required. A future editor removing the "redundant" short-code arms — on the
reasonable assumption that `measure` was normalized upstream — regenerates `Maz`
silently. That is now written on the column description in both models rather
than left as tribal knowledge.

### Not fixed — recorded, and worth your read

- **The `measure` guards can self-disable.** Both are scoped
  `where assessment_edition = 'DIBELS 8th Edition'`, a second vendor-controlled
  free-text string that nothing tests. If Amplify renames or nulls it, the scope
  matches zero rows and the test reports PASS — the same silent-drift class this
  PR exists to catch. The new `measure_name_code` guard is unscoped and therefore
  immune, which partially covers this, but the `measure` guards remain exposed.
- **The accepted list and the CASE arms can drift apart.** Neither `measure` CASE
  has an `LNF` arm, so a short `LNF` would pass through verbatim and is not in the
  accepted list — an error-severity build failure on a legitimate row. Low
  probability today (LNF is not a PM measure in our expected-assessments sheet)
  but structurally real, because the list and the arms are maintained separately.
- **A third sibling was left out.** `stg_amplify__mclass__api__pm_student_summary`
  unions into the same intermediate and has no short-code arms, a four-input
  surrogate key against this model's five, and no guards on `measure` at all. It
  serves AY2022-2024 while the SFTP feed serves AY2025, so the feeds are
  year-disjoint and nothing is currently wrong — but it is a third variant of the
  same mapping.

The last two strengthen the case for the shared-macro fast-follow you have twice
declined to block on: three feeds, three hand-maintained copies, and a value
domain maintained separately from the arms that produce it.

### One claim from that pass I disproved

It flagged the added `assessment_edition` surrogate-key input as breaking
cross-feed dedup between the SFTP and API models, emitting duplicate PM rows. The
premise is false — the feeds are year-disjoint (SFTP holds AY2025 only, API holds
AY2022-2024 only), so no row ever collapsed across them and none can now
duplicate. Recording it because a future reader may reach for the same theory.

@claude one more look please, focused on the two `measure_name_code` guards and
whether the unscoped-versus-scoped split between them and the `measure` guards is
the right call.

GabyRangelB:
## All open findings now closed

Every item left hanging in my previous comment is resolved. Retracting one claim
from that comment along the way.

### Retraction — I marked two things "fixed" that were not

The previous comment claimed the `measure_name_code` guard closed the truncation
hazard, and that the raw-read behavior was fixed. Neither was true: the `substr`
expression was unchanged, and the raw-read was documented rather than removed.
Both are now actually fixed, below.

### Normalize once — the raw-read trap is structurally gone

Both models now normalize `measure` in a first CTE and derive
`measure_name_code` from the normalized value in a second. Previously
`measure_name_code` read the RAW source column, because BigQuery has no lateral
column aliases — which is why it needed a short-code AND a long-label arm for the
same measure, and why deleting either regenerated the original `Maz` defect.

Format tolerance now lives in exactly one place, so there is no longer a
redundant-looking arm for a future editor to remove.

### The truncating fallback is gone

`else substr(measure, strpos(measure, '(') + 1, 3)` is replaced by an explicit
code per measure with no `else`. The old fallback silently truncated an
unrecognized code to three characters, which could collide with a real one and
pass every guard. Measured, using the cases that previously slipped through:

| hypothetical vendor code | before | now |
| --- | --- | --- |
| `ORFX` | resolved to `ORF`, passed every guard | NULL, caught |
| `NWFZ` | resolved to `NWF`, passed | NULL, caught |
| `Reading Something (ORFX)` | resolved to `ORF`, passed | NULL, caught |
| `DAZE` | resolved to `DAZ`, caught | NULL, caught |

An unrecognized measure now yields NULL and `not_null` catches it
unconditionally.

### The guards no longer depend on a second vendor string

Both `accepted_values` tests were scoped
`where assessment_edition = 'DIBELS 8th Edition'` — a second vendor-controlled
free-text field that nothing tests. A rename or null there would match zero rows,
report PASS, and silently disable the protection. The scope is dropped and the
intervention measure added to the accepted domain, so every row must carry a
known measure regardless of edition. All four guards across both models are now
unconditional.

### The surrogate key no longer churns on a format flip

Because the key now hashes the normalized `measure`, the same logical row keys
identically under either vendor format. This resolves the tradeoff flagged as
"known and named" in the previous pass.

### The third-variant concern is moot, not deferred

`stg_amplify__mclass__api__pm_student_summary` is disabled in both consuming
districts (`amplify: mclass: api: +enabled: false` in kippnewark and
kipppaterson), so it builds nowhere — the relation does not exist in
`kippnewark_amplify`. kipptaf reads a frozen BQ-native archive instead,
`kipptaf_amplify.stg_amplify__pm_student_summary`, a static table last modified
2025-09-16 whose 63,996 rows match the AY2022-2024 totals exactly. There is no
live third mapping to harmonize and the archive cannot drift.

### Verification

Output is unchanged on current data across both districts: all 8 measures still
emit identical `measure`, `measure_name_code` and `measure_name` across the two
models, and the three CASE blocks remain byte-identical between them. Newark
PASS=10, Paterson PASS=5. Trunk clean.

The only remaining warning is the pre-existing aimline surrogate-key duplication
from #3552, unchanged from prod.

### Still deferred, deliberately

The shared-macro extraction. With the mapping now in one CTE per model rather
than duplicated across CASE blocks, the duplication is smaller than when first
raised, but it is still two copies.

@claude final look please — particularly whether dropping the `else` from
`measure_name_code` in favor of NULL-plus-`not_null` is the right failure mode
for a staging model, given a NULL there propagates to kipptaf in a separate run
before the test result is known.

cbini (approved):

PR #4909, opened 2026-08-18
refactor(dlt): use dlt rows_loaded instead of a row_count helper

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will delete the `_with_row_count()` helper added
in #4879 and let the PowerSchool dlt op yield dlt's materializations straight
through, relying on the `rows_loaded` metadata `dagster_dlt` already attaches to
every event. It also removes test assertions and doc figures that go stale
whenever a table is added or moved.

### Why the helper can go

`_with_row_count()` existed to stop `fetch_row_count_metadata()` from aborting
the whole multi-asset op when a table's first load extracts zero rows. That
failure is real (it happened on Paterson's `u_expectations` in #4879), and the
cause is that the missing-`jobs` check sits *outside* that function's own
`try`/`except`:

```python
jobs_metadata = materialization.metadata.get("jobs")
if not jobs_metadata or not isinstance(jobs_metadata, list):
    raise Exception("Missing jobs metadata to retrieve row count.")
try:
    return TableMetadataSet(row_count=_fetch_row_count(dlt_pipeline, table_name))
```

But `dagster_dlt`'s own `resource.py` already puts dlt's per-table count on every
materialization for free:

```python
rows_loaded = dlt_pipeline.last_trace.last_normalize_info.row_counts.get(
    normalized_table_name
)
if rows_loaded:
    base_metadata["rows_loaded"] = MetadataValue.int(rows_loaded)
```

For a full `replace` load that is the same number the helper was querying for,
it cannot raise, and it costs no round-trip — the old path billed a `count(*)`
per table per run (49 extra BigQuery queries on Paterson alone). So the op just
`yield from`s `run()`.

**Trade-off, deliberate:** `rows_loaded` is plain metadata, not the canonical
`dagster/row_count` key, so the asset catalog's row-count column stays empty.
The number is still on every materialization. Re-adding the chained call to win
that column back would restore the crash, so the `libraries/dlt/CLAUDE.md` note
now says so explicitly.

Verified against the installed `dagster-dlt` 0.29.18 source and confirmed
byte-identical to dagster `master` — there is no newer version to bump to that
fixes this, and no config flag or translator hook that attaches row counts
tolerantly.

### Stale figures and vacuous tests removed

Each of these breaks on an unrelated config change and asserts nothing about
behavior:

| Removed | Why |
| --- | --- |
| 49-name spec-membership map in `test_dlt_powerschool_assets.py` | Restated `config/assets.yaml` in Python; a stale copy had to be repaired in #4879 |
| `len(config["assets"]) == 49`, `len(assets[0].keys) == 49`, `len(nightly) == 25` | Hardcoded counts, no behavioral claim |
| `tests/assets/test_powerschool_dlt_paterson_defs.py` | Its trigger test duplicated the unit suite; its asset-key and coverage tests move there |
| "49 tables" / "57 tables" in 3 code-location `CLAUDE.md` files | Same staleness in docs |
| `(dlt 1.28.1)` in a `libraries/dlt/CLAUDE.md` heading | Already wrong; installed is 1.30.0 |

The real invariant the map was standing in for is kept, derived from the YAML
rather than restated: every configured table belongs to at least one trigger,
and the intraday/nightly overlap is exactly the no-cursor set. It moves from
`tests/assets/` into `tests/libraries/`.

**Corrected after review:** an earlier draft of this description claimed the
move puts the test "in CI". That is wrong — `claude-review` checked and I
confirmed it: no workflow in `.github/` invokes `pytest` at all, and the only
mentions of it are manual checklist items in the PR template. What the move
actually buys is that the test leaves a directory `tests/CLAUDE.md` documents
as needing environment variables and external connections (so it gets skipped
in practice) and joins the plain unit suite, which anyone can run with a bare
`uv run pytest`. More likely to be run, not automatically run.

**Mutation-tested, and the result corrects the framing above.** Two mutations
against the config, each reverted afterwards:

1. Orphan a table from both tiers (`u_expectations` to `intraday: false`,
   `nightly: false`). Caught — but by *production* code, not the test:
   `_nightly_targets()` raises `ValueError: table(s) in neither tier (would
   never materialize)` at `schedules.py:17`. So this invariant was never
   protected by the deleted 49-name map either; it fails any definitions load.
1. Break the tier overlap without orphaning anything (`gen`, a no-cursor table,
   to `intraday: false` while staying `nightly: true`). Caught by the test's own
   second assertion, `intraday & nightly == no_cursor`. Production code stays
   quiet here.

So the surviving test earns its place on the second case, and the first case is
guarded more strongly than any test could manage. Neither depended on the map.

Net: 88 insertions, 265 deletions.

### Review round 2 — `claude-review` findings applied

Full per-finding verdicts are in a separate PR comment. Two code changes landed:

- **Reuse (finding 1):** the test no longer re-declares the `assets.yaml` path. It
  imports `config_file` from the code location, which derives from `__file__` and
  so resolves regardless of pytest's cwd — the literal it replaced only worked
  when pytest ran from the repo root. Imported inside the helper, not at module
  scope, because a module-scope code-location import needs the dbt manifest at
  collection time.
- **Coverage (finding 2):** the finding was right that the old stub yielded zero
  items, so nothing ever flowed through the `yield from` and my "exercises the
  changed line" claim was overstated. `_FakeRunResult` now accepts events, and
  `test_op_yields_dlt_events_unmodified` asserts the op yields dlt's event
  through by **identity**. Mutation-tested: re-adding a `_replace()` wrapper
  fails it. Identity matters — the re-wrapped event compares *equal* to the
  original, so an equality assertion would have passed the mutation silently.

## Self-review

### General

### Dagster

- [x] `uv run pytest tests/libraries/test_dlt_powerschool_assets.py tests/libraries/test_powerschool_dlt_extract_workers.py tests/libraries/test_dlt_powerschool_sensors.py` — 15 passed
- [ ] `uv run dagster definitions validate` — **not run.** Per
      `src/teamster/CLAUDE.md` this needs a prepared dbt manifest per location,
      and `kipptaf` / `kippmiami` fail at module load in a codespace on
      credential specs regardless. The op body itself IS executed by
      `test_powerschool_dlt_extract_workers.py`, which drives
      `assets_def.op.compute_fn.decorated_fn(...)` and iterates it — that
      exercises the changed `yield from` line. Branch deploy covers the rest.
- [x] No new integration; no asset key, IO manager, or config-pattern change

### Docs

- [x] No schedule or sensor change, so no automations-catalog regeneration
- [x] Updated `libraries/dlt/CLAUDE.md` and the `kipppaterson` / `kippnewark` /
      `kippcamden` location `CLAUDE.md` files

## CI checks

- [x] **Trunk** — `trunk check --force` clean on all changed files (the two
      markdown re-wraps were applied by the pre-commit `fmt` hook, which then
      re-checked clean)
- [ ] **dbt Cloud** — no dbt changes; expected to pass or skip
- [ ] **Dagster Cloud** — this touches `src/teamster/libraries/dlt/**`, a shared
      path, so the branch deploy fans out to every consuming location, not just
      Paterson

<details>
<summary>For Claude</summary>

Claude Code authored the commit. Human-directed: the decision to drop the helper
in favour of `rows_loaded`, and the instruction to strip stale figures and
vacuous tests. AI-assisted: verifying the `dagster-dlt` behaviour against
installed source and dagster `master`, finding the `rows_loaded` attachment
point, and writing the code, test, and doc changes.

The judgement call worth a look is the trade-off above — losing the
`dagster/row_count` catalog column in exchange for deleting the helper. If that
column is something anyone actually reads, the alternative is re-keying the
event's own `rows_loaded` into `TableMetadataSet(row_count=...)`, which keeps the
column with no SQL query and no raise, at the cost of ~6 lines back.

Also intentionally out of scope: `docs/superpowers/plans/` and `specs/` still
reference `.fetch_row_count()` and the deleted test file. Those are working
documents excluded from the mkdocs nav and are expected to describe superseded
designs.

Refs #4879

</details>


cbini:
## Review response — per-finding verdicts

All four verified against the code before responding. Two accepted and fixed,
one accepted as a documentation correction, one already-verified with the
evidence below.

### 1. Reuse — hardcoded config path duplicates `config_file` — ACCEPTED, fixed

Confirmed: `config_file` exists at
`src/teamster/code_locations/kipppaterson/powerschool/sis/dlt/assets.py:9`, and
the deleted test imported it rather than re-declaring the path.

The finding undersells itself. It frames this as future-drift risk, but the two
literals are not actually equivalent: `config_file` is
`pathlib.Path(__file__).parent / "config" / "assets.yaml"`, an absolute path
derived from the module, while my version was a **repo-root-relative** literal
that only resolved when pytest ran from the repo root. So this was a latent
cwd-dependency, not just duplication.

Fixed, with one deviation from the suggestion: the import goes **inside**
`_config_entries()`, not at module scope. A module-scope code-location import
needs the dbt manifest at collection time (see `src/teamster/CLAUDE.md`), which a
fresh worktree lacks — that is why every code-location import in this file is
already function-local.

### 2. Test coverage — `yield from` never sees a real event — ACCEPTED, fixed

Correct, and it catches an overstatement of mine. `_FakeRunResult.__iter__`
returned `iter(())`, so the `yield from` line executed but nothing flowed through
it. My self-review's "exercises the changed `yield from` line" was true only in
the trivial sense.

Fixed as suggested: `_FakeRunResult` now takes events, and
`test_op_yields_dlt_events_unmodified` asserts the op passes dlt's event through.

One deliberate choice worth flagging, because it is the difference between this
test working and not working: the assertion is **identity** (`yielded is event`),
not equality. I mutation-tested it by re-adding a `_replace()` wrapper to the op,
and the re-wrapped `MaterializeResult` compares **equal** to the original —
identical repr — so an equality assertion would have passed the mutation
silently. Identity fails it, which is the regression this is meant to catch.

Suite is 15 passed, up from 14.

### 3. "run in CI" claim not backed by workflow config — ACCEPTED, you were right

Checked it myself: no workflow in `.github/` invokes `pytest`. The only mentions
are manual checklist items in `pull_request_template.md` (lines 24 and 82).
Nothing runs the test suite automatically in this repo.

So the claim was wrong, and I have corrected the PR description rather than
softened it. What the move actually buys: the test leaves a directory that
`tests/CLAUDE.md` documents as requiring environment variables and external
connections — so it gets skipped in practice — and joins the plain unit suite
that runs with a bare `uv run pytest`. More likely to be run, not automatically
run.

Your medium confidence and the "possible out-of-repo branch-protection check"
hedge were both appropriate; there is no such mechanism, so the simpler reading
was the right one.

### 4. `rows_loaded` claim not reproducible in your environment — verified here

Reasonable to flag rather than assert. I verified it in a checkout with the venv
present, reading the installed source directly. `dagster_dlt/resource.py:134-138`:

```python
rows_loaded = dlt_pipeline.last_trace.last_normalize_info.row_counts.get(
    normalized_table_name
)
if rows_loaded:
    base_metadata["rows_loaded"] = MetadataValue.int(rows_loaded)
```

Two details worth recording, since they are the whole basis for the change:

- The `if rows_loaded:` guard means the key is **omitted** at zero rows rather
  than set to 0. That is fine here — the point is that it cannot raise — but it
  does mean a zero-row table carries no count at all.
- The raise this PR removes is at `dlt_event_iterator.py:48-49`, and it sits
  **outside** that function's own `try`/`except` (which starts at line 51 and
  only guards the SQL query). So the missing-`jobs` case was never covered by
  dagster-dlt's own error handling. I also diffed the installed 0.29.18 file
  against dagster `master` — byte-identical, so there is no newer release that
  fixes this.

### Not addressed

Nothing. No findings declined.

### Unrelated CI note

The single failing check is `dagster-cloud-deploy / deploy` for **kippcamden**:
`DagsterUserCodeUnreachableError ... Connection refused` reaching the code-server
pod on gRPC port 4000. Not this PR — `kippnewark` imports the same edited factory
(`build_powerschool_dlt_assets`; Camden, Newark and Paterson are its three
consumers) and deployed successfully, as did the other three locations. Same code,
different outcome, and the traceback carries no `teamster` frames. Needs a rerun.

Issue #4907, opened 2026-08-18
fix(k8s): code server reload failures from deploymentStartupTimeout default plus scheduling over-constraint


cbini:
## What's happening

Code locations intermittently fail to reload, and the Dagster+ hybrid agent logs
`DagsterUserCodeUnreachableError: Could not reach user code server` at roughly
100 occurrences per week.

Expected: a code-server redeploy swaps pods with a brief gRPC gap of a few
seconds.

Actual: the replacement pod cannot be placed, sits `Pending`, and the agent
abandons it after 300 seconds with `Timed out waiting for deployment`, which
surfaces as a code location reload failure.

The reported "uptick over the past couple weeks" is **not** in the data. The peak
week was 2026-07-28 to 08-04 and the two most recent weeks are both lower. This
is a chronic condition, not a new regression. Counts are for code-server pods
only (`involvedObject.name` containing `-prod-`) in the `dagster-cloud`
namespace:

| Week | Agent gRPC errors | Killing | ScaleDown | Preempted | Evicted | FailedScheduling |
| --- | --- | --- | --- | --- | --- | --- |
| Jul 20-27 | 85 | 122 | 42 | 21 | 0 | 472 |
| Jul 28-Aug 4 | 134 | 197 | 59 | 37 | 0 | 454 |
| Aug 4-11 | 106 | 118 | 48 | 22 | 0 | 213 |
| Aug 11-18 | 100 | 142 | 38 | 32 | 0 | 438 |

Agent errors track code-server kills at a stable ratio (0.70, 0.68, 0.90, 0.70),
which is what ties the errors to pod deaths rather than to anything in this
repository. Cloud Logging retention is 30 days, so an inflection before
2026-07-19 cannot be ruled out.

Impact is low and should be weighed before spending much here. All 17 terminal
sensor tick failures across the five code locations launched **zero** runs, and
every sensor stayed RUNNING and recovered on its next tick. No materialization
was lost. The cost is reload failures in the UI, tick latency, and alert noise.

## Steps to reproduce

1. Fetch reload history for any location and look for `loadStatus: ERROR`:
   `mcp__dagster__get_location_load_history` with `location_name: kipptaf`,
   `limit: 8`. The error message carries the pod's `FailedScheduling` events.
1. Count agent gRPC failures over a window, in Cloud Logging:

   ```text
   resource.type="k8s_container"
   resource.labels.namespace_name="dagster-cloud"
[...]

PR #4913, opened 2026-08-18
fix(kipptaf): chunk bigquery table_modified sensor on a tick time budget

### LINKED ISSUE #4908: fix(kipptaf): chunk bigquery table_modified sensor across ticks; worst case exceeds the tick timeout
## What's happening

`kipptaf__google__bigquery__table_modified_sensor` polls all of its assets
serially inside a single tick. Its bounded worst case is larger than the tick
timeout, so the tick fails whenever enough calls are slow.

Expected: the sensor checks table modification times and returns well within the
600 second tick budget.

Actual: the tick occasionally runs to the timeout and fails with:

```text
dagster._core.errors.DagsterUserCodeUnreachableError: The sensor tick timed out
due to taking longer than 600 seconds to execute the sensor function. One way to
avoid this error is to break up the sensor work into chunks, using cursors to let
subsequent sensor calls pick up where the previous call left off.
```

The worst case is guaranteed to exceed the budget. `asset_selection` holds 32
assets and each one gets its own serial `bq.get_table()` call:

| Source | Count | Derived from |
| --- | --- | --- |
| `google_appsheet_asset_specs` | 10 | dbt manifest sources where `source_name` is `google_appsheet` |
| `airbyte_asset_specs` | 22 | `airbyte/config/assets.yaml`, 2 connections times their `destination_tables` |
| Total per tick | 32 | one serial `bq.get_table()` each |

Each call is bounded at `timeout=10` with `retry=DEFAULT_RETRY.with_deadline(30)`,
so a single asset can consume about 30 seconds. 32 times 30 is roughly 960
seconds against a 600 second budget, and the loop has no notion of remaining
budget, so nothing can stop it.

The per-call bounds are themselves correct and follow the guidance in
`src/teamster/CLAUDE.md`. The defect is the absence of any bound on the aggregate.

Impact is low. The failure is non-terminal: no runs are launched, the sensor stays
RUNNING, and it recovers on the next tick. `minimum_interval_seconds` is 300, so a
failed tick delays AppSheet and Airbyte materializations by one interval rather
than dropping them. Filed as a correctness fix, not an incident.

## Steps to reproduce

1. Count occurrences over 30 days in Cloud Logging:

   ```text
   resource.type="k8s_container"
   resource.labels.namespace_name="dagster-cloud"
   "timed out due to taking longer than 600 seconds"
   ```

   Returns five events: 2026-07-21, 07-24, 08-12, 08-13, 08-18.

1. Inspect the failing tick directly: `mcp__dagster__get_tick_history` with
   `name: kipptaf__google__bigquery__table_modified_sensor`,
   `repository_location_name: kipptaf`, `statuses: ["FAILURE"]`, `limit: 3`.
   The 2026-08-18 tick ran 15:42:41 to 15:52:44 UTC, about 603 seconds.
1. Confirm the asset count is 32 by counting `google_appsheet` tables in
   `src/dbt/kipptaf/models/google/appsheet/sources-bigquery.yml` (10) and
   `destination_tables` across `src/teamster/code_locations/kipptaf/airbyte/config/assets.yaml`
   (22).

At 603 seconds across 32 assets the average call took roughly 19 seconds, far
above a normal `tables.get` metadata latency of 100 to 300 ms. Most calls were
hitting the 10 second timeout and retrying, so this is not merely a long loop.

## Where

- **Environment:** prod
- **Run, PR, or dashboard link (if any):** no run; sensor tick failures only.
  Five failures in 30 days, most recently 2026-08-18 15:52:44 UTC.

<details>
<summary>For Claude</summary>

### Relevant files

- `src/teamster/code_locations/kipptaf/google/bigquery/sensors.py` — the sensor
- `src/teamster/code_locations/kipptaf/google/appsheet/assets.py` — 10 specs
- `src/teamster/code_locations/kipptaf/airbyte/assets.py` plus
  `airbyte/config/assets.yaml` — 22 specs
- `tests/sensors/test_sensors_google_bigquery.py` — existing coverage

### Suspected fix

Chunk the asset list across ticks and carry the position in the cursor, as the
error message itself recommends. Each tick then processes a bounded slice whose
worst case fits inside 600 seconds, and the full set is still swept over a few
consecutive ticks.

Care needed on the cursor schema. The cursor is currently a flat mapping of
`python_identifier` to a modified timestamp, read via
`cursor.get(python_identifier, 0)`. A chunk position has to be stored without
colliding with a generated asset identifier and without breaking on cursors
already persisted by the deployed version. `src/teamster/CLAUDE.md` warns about
exactly this class of migration. A reserved key that cannot collide with a
generated `python_identifier`, or a nested versioned structure, both work; the
choice should be deliberate.

Worth considering alongside chunking, though not a substitute for it:

- Whether 32 serial round trips is the right shape at all. `list_tables()` does
  not return `modified`, and the only batch path is `INFORMATION_SCHEMA.TABLES`,
  which is billed at a 10 MB minimum per query while `tables.get` is free
  metadata. This tradeoff is already recorded in `src/teamster/CLAUDE.md`.
- Whether the per-call retry deadline should be lower than 30 seconds, so one
  unhealthy table cannot consume 5 percent of the tick budget.

### What fixed looks like

- The swept set still covers all 32 assets across consecutive ticks, with no
  asset starved.
- A cursor written by the currently deployed version is read without error on the
  first tick after deploy.
- No further `timed out due to taking longer than 600 seconds` events for this
  sensor.
- `tests/sensors/test_sensors_google_bigquery.py` updated for the chunking
  behavior.

</details>

## Out of scope

The agent-level gRPC unavailable errors and code location reload failures,
tracked in #4907. They share this exception class but have an unrelated cause.

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop one Dagster sensor from failing on a timeout.

The sensor watches 32 BigQuery tables to spot which ones have changed, checking them one at a time. Nothing limited how long the whole pass could take, so whenever BigQuery got slow the sensor ran past its 10 minute limit and the run was killed. That happened five times in the last 30 days.

It now stops once it has used half its time allowance, remembers where it got to, and carries on from there on the next run. When BigQuery is healthy the whole pass still finishes in about 15 seconds, so nothing changes day to day.

Refs #4908

## Reviewer Notes

- The sensor remembers its place in the same saved state that already holds each table's last-changed time. Worth a second look that a saved state written by the version running in production today still loads cleanly, since that is the one thing here that could misbehave on deploy rather than in review.
- Two new tests cover the stop-and-carry-on behaviour. They run against a stub instead of real BigQuery, so they are fast and give the same answer every time. I checked they actually fail when the stopping logic is removed, so they are genuinely testing it and not just passing.
- I declined one review finding: if someone reorders the table config without changing how many tables there are, a resumed run could check a different group than the one it skipped. Nothing is lost or double-counted and it corrects itself on the next pass, so I left the code alone and documented the limit instead.
- The time allowance and the gap between runs are both 300 seconds. That is deliberate rather than an oversight; reasoning below.

### General

      feedback; dismiss false positives with a brief reply explaining why. — done,
      per-finding verdicts posted as a comment: two accepted and fixed in
      6eeeebb, three declined with reasons

### Dagster

      satisfied via the pytest path below. Run directly it fails on an unset
      Illuminate credential, the known codespace limitation recorded in
      `src/teamster/CLAUDE.md`, at module scope in a file this PR does not touch.
      code location — `test_definitions_kipptaf`: `1 passed in 57.95s`
      with the correct asset key format and IO manager — n/a, no new integration

### dbt

Skipped, no dbt changes.

### Docs

      catalog — n/a, no sensor was added, renamed, or re-scheduled; only the body
      of an existing sensor changed, and the generator emits name and interval,
      neither of which moved
      `CLAUDE.md` — n/a, no new integration

## CI checks

- [ ] **Trunk** — passes. `trunk check --force` was clean on both changed files
      before each commit.
- [ ] **dbt Cloud** — passed on the first commit; no dbt paths changed.
- [ ] **Dagster Cloud** — passed on the first commit.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Found while debugging a separate report of code
location reload failures; that unrelated root cause is tracked in #4907 with a
design spec in #4914, and is deliberately untouched here even though it surfaces
the same `DagsterUserCodeUnreachableError` class.

### Why the timeout was guaranteed, not unlucky

`asset_selection` is 32 specs: 10 from dbt manifest sources where `source_name`
is `google_appsheet`, plus 22 from `destination_tables` across two connections in
`airbyte/config/assets.yaml`. Each gets one serial `bq.get_table()` bounded at
`timeout=10` with `retry=DEFAULT_RETRY.with_deadline(30)`, so a single asset can
burn about 30s. 32 x 30 is roughly 960s against a 600s tick budget, and the loop
had no notion of elapsed time, so nothing could stop it. The observed 2026-08-18
failure ran 603s, about 19s per call — most calls were hitting the 10s timeout
and retrying, not merely accumulating.

The per-call bounds were already correct per `src/teamster/CLAUDE.md`. The defect
was the absence of any aggregate bound.

### Time budget rather than a fixed chunk size

A count-based chunk (10 of 32) would split every sweep across four ticks and
multiply normal detection latency by four, for a sweep that really takes ~15s. A
time budget engages only once calls actually slow down, leaving the common case
untouched. `TICK_TIME_BUDGET_SECONDS` is 300 against a 600s tick timeout; the
deadline is checked before each call, so worst case is 300 + one 30s in-flight
call = 330s, comfortably under.

### Reviewer Notes flag 1: the cursor migration

The resume index lives under the reserved key `$offset`. A dollar sign is not
legal in a Python identifier, so it cannot collide with an
`AssetKey.to_python_identifier()` entry sharing the dict — verified empirically
against the live 32-spec list in `test_bigquery_table_sensor_cursor_offset`.
The offset is also range-checked, so an index left dangling by a deploy that
shortens `asset_selection` resets to 0 rather than raising.

The cursor is now returned on every tick, not only ticks that emit events. The
previous bare `SkipReason()` return discarded it, which under chunking would
throw away the resume index and leave the tail of `asset_selection` permanently
unpolled.

Note `next_offset = 0` is overloaded: it means "sweep finished", and a break at
index 0 also yields 0. Harmless, because both cases correctly want the next tick
to start at 0 — but it is why both new tests start from a non-zero offset, which
is the only way an assertion can distinguish "stopped here" from "finished".

### Reviewer Notes flag 2: the new tests

- `test_bigquery_table_sensor_budget_exhausted_polls_nothing` — budget 0 must
  break before the first call, leaving the resume index untouched. Asserts
  `get_table` was never called.
- `test_bigquery_table_sensor_resumes_mid_sweep` — a fake monotonic clock
  stepping 100s against a 300s budget puts the deadline at 400, so checks at 200
  and 300 pass and the check at 400 breaks. Asserts exactly two `get_table` calls
  and that the offset advanced by exactly two.

Both stub `BigQueryResource`, so no warehouse access and no flakiness.
Mutation-verified: disabling the budget check fails both, and the file restored
byte-exact afterward.

Implementation trap worth knowing: `bigquery/__init__.py` does
`from ...sensors import sensors`, rebinding the name to the sensor **list**. So
`import ...bigquery.sensors as x` hands back a list and monkeypatching it
silently no-ops. The tests reach the real module via `importlib.import_module`.

Local run of all four: `4 passed in 28.38s`.

### Reviewer Notes flag 3: declined reorder finding

`0 <= offset < len(asset_selection)` catches a deploy that shortens the list but
not one that reorders it at equal length. A resumed tick would then poll a
different slice than the one it skipped. Declined on impact: per-asset state is
keyed by `python_identifier`, not index, so nothing is lost or duplicated and the
worst case is a one-sweep (300s) detection delay that self-corrects. Closing it
properly means storing a resume identifier and scanning for it — more machinery
than a rare, benign, self-healing event justifies. The code comment now names
reordering explicitly so the guard is not read as stronger than it is.

### Reviewer Notes flag 4: budget equals interval

Both are 300s, and `minimum_interval_seconds` is a floor between tick starts
rather than a guaranteed idle gap, so during sustained BigQuery slowness ticks
can run near-continuously. That is the intended self-limiting behaviour: the
sensor keeps making forward progress through the sweep instead of stalling. If
continuous gRPC worker occupancy on the code server ever becomes a problem, the
lever is lowering the budget, not raising the interval.

### Also declined

The reviewer flagged that `deadline` is computed before entering
`get_client()`, so a slow client acquisition could in principle consume the
budget before the first check. `google.cloud.bigquery.Client()` construction is
local and non-blocking, and the failure mode would be no forward progress for one
tick rather than a regression. The reviewer marked it informational and I agree.

### Out of scope

Replacing 32 serial `tables.get` round trips with `INFORMATION_SCHEMA.TABLES`
(billed at a 10 MB minimum per query versus free metadata), and lowering the 30s
per-call retry deadline. Both recorded in #4908.

</details>

### DIFF
--- a/src/teamster/code_locations/kipptaf/google/bigquery/sensors.py
+++ b/src/teamster/code_locations/kipptaf/google/bigquery/sensors.py
@@ -51,7 +51,12 @@ def bigquery_table_modified_sensor(
     offset = cursor.get(OFFSET_CURSOR_KEY, 0)
 
     # tolerate a cursor written before this key existed, and an offset left
-    # dangling by a deploy that shortened asset_selection
+    # dangling by a deploy that shortened asset_selection. A deploy that REORDERS
+    # asset_selection without changing its length is not detectable here: the
+    # resumed tick polls a different slice than the one it skipped. That is
+    # bounded and self-healing -- per-asset state is keyed by identifier rather
+    # than index, so the worst case is one sweep's detection delay, never a lost
+    # or duplicated materialization.
     if not isinstance(offset, int) or not 0 <= offset < len(asset_selection):
         offset = 0
 
@@ -103,13 +108,23 @@ def bigquery_table_modified_sensor(
 
     cursor[OFFSET_CURSOR_KEY] = next_offset
 
+    if asset_events:
+        skip_reason = None
+    elif next_offset:
+        # a partial slice found nothing, but assets remain unpolled this sweep
+        skip_reason = SkipReason(
+            f"No modified tables in this slice; resuming at index {next_offset}."
+        )
+    else:
+        skip_reason = SkipReason("No modified tables.")
+
     # the cursor is returned on every tick, not just ticks that emit events --
     # otherwise the resume offset would be discarded and the tail of
     # asset_selection would never be polled
     return SensorResult(
         asset_events=asset_events,
         cursor=json.dumps(obj=cursor),
-        skip_reason=(None if asset_events else SkipReason("No modified tables.")),
+        skip_reason=skip_reason,
     )
 
 
--- a/src/teamster/code_locations/kipptaf/google/bigquery/sensors.py
+++ b/src/teamster/code_locations/kipptaf/google/bigquery/sensors.py
@@ -1,4 +1,5 @@
 import json
+import time
 
 from dagster import (
     AssetMaterialization,
@@ -23,6 +24,19 @@ from teamster.code_locations.kipptaf.google.appsheet.assets import (
 
 asset_selection: list[AssetSpec] = [*google_appsheet_asset_specs, *airbyte_asset_specs]
 
+# Wall-clock budget for one tick's polling loop. The tick is killed at 600s by
+# default_sensor_timeout, and a single get_table below can burn ~30s before it
+# gives up, so this leaves room for one in-flight call to finish past the check.
+# Polling every asset normally takes a few seconds, so the budget only bites once
+# BigQuery metadata calls start timing out -- at which point the loop stops and
+# resumes on the next tick instead of running the whole tick into the timeout.
+TICK_TIME_BUDGET_SECONDS = 300
+
+# Cursor key holding the resume position. "$" is not legal in a Python
+# identifier, so this can never collide with an AssetKey.to_python_identifier()
+# entry sharing the same dict.
+OFFSET_CURSOR_KEY = "$offset"
+
 
 @sensor(
     name=f"{CODE_LOCATION}__google__bigquery__table_modified_sensor",
@@ -34,8 +48,31 @@ def bigquery_table_modified_sensor(
     asset_events = []
     cursor: dict = json.loads(context.cursor or "{}")
 
+    offset = cursor.get(OFFSET_CURSOR_KEY, 0)
+
+    # tolerate a cursor written before this key existed, and an offset left
+    # dangling by a deploy that shortened asset_selection
+    if not isinstance(offset, int) or not 0 <= offset < len(asset_selection):
+        offset = 0
+
+    deadline = time.monotonic() + TICK_TIME_BUDGET_SECONDS
+
+    # 0 means the sweep finished; a break below overwrites it with the resume index
+    next_offset = 0
+
     with db_bigquery.get_client() as bq:
-        for assets_def in asset_selection:
+        for index in range(offset, len(asset_selection)):
+            if time.monotonic() >= deadline:
+                next_offset = index
+                context.log.info(
+                    msg=(
+                        f"Tick time budget of {TICK_TIME_BUDGET_SECONDS}s exhausted "
+                        f"after {index - offset} assets; resuming at index {index}"
+                    )
+                )
+                break
+
+            assets_def = asset_selection[index]
             python_identifier = assets_def.key.to_python_identifier()
 
             cursor_modified_timestamp = cursor.get(python_identifier, 0)
@@ -64,10 +101,16 @@ def bigquery_table_modified_sensor(
                 asset_events.append(AssetMaterialization(asset_key=assets_def.key))
                 cursor[python_identifier] = table_modified_timestamp
 
-    if asset_events:
-        return SensorResult(asset_events=asset_events, cursor=json.dumps(obj=cursor))
-    else:
-        return SkipReason()
+    cursor[OFFSET_CURSOR_KEY] = next_offset
+
+    # the cursor is returned on every tick, not just ticks that emit events --
+    # otherwise the resume offset would be discarded and the tail of
+    # asset_selection would never be polled
+    return SensorResult(
+        asset_events=asset_events,
+        cursor=json.dumps(obj=cursor),
+        skip_reason=(None if asset_events else SkipReason("No modified tables.")),
+    )
 
 
 sensors = [
--- a/tests/sensors/test_sensors_google_bigquery.py
+++ b/tests/sensors/test_sensors_google_bigquery.py
@@ -1,3 +1,5 @@
+import json
+
 from dagster import SensorResult, build_sensor_context
 from dagster_shared import check
 
@@ -23,3 +25,39 @@ def test_bigquery_table_sensor():
         context.log.info(msg=asset_event)
 
     context.log.info(msg=sensor_result.cursor)
+
+
+def test_bigquery_table_sensor_cursor_offset():
+    """Guards the cursor migration: a cursor written before chunking existed must
+    still load, the resume offset must round-trip, and the reserved key must not
+    be able to collide with a generated asset identifier.
+    """
+    from teamster.code_locations.kipptaf.google.bigquery.sensors import (
+        OFFSET_CURSOR_KEY,
+        asset_selection,
+        bigquery_table_modified_sensor,
+    )
+
+    # a cursor in the pre-chunking shape: asset identifiers only, no offset key
+    legacy_identifier = asset_selection[0].key.to_python_identifier()
+
+    context = build_sensor_context(
+        cursor=json.dumps({legacy_identifier: 1.0}),
+    )
+
+    sensor_result = check.inst(
+        obj=bigquery_table_modified_sensor(
+            context=context, db_bigquery=BIGQUERY_RESOURCE
+        ),
+        ttype=SensorResult,
+    )
+
+    cursor = json.loads(check.not_none(value=sensor_result.cursor))
+
+    assert OFFSET_CURSOR_KEY in cursor
+    assert isinstance(cursor[OFFSET_CURSOR_KEY], int)
+    assert 0 <= cursor[OFFSET_CURSOR_KEY] < len(asset_selection)
+
+    assert OFFSET_CURSOR_KEY not in {
+        spec.key.to_python_identifier() for spec in asset_selection
+    }
--- a/tests/sensors/test_sensors_google_bigquery.py
+++ b/tests/sensors/test_sensors_google_bigquery.py
@@ -1,9 +1,61 @@
+import contextlib
+import importlib
 import json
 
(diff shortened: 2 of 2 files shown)


cbini:
Verdicts on all five findings, verified against the code rather than taken on faith. Two accepted, three declined with reasons.

### 1. Test coverage gap on the core fix — ACCEPTED, fixed in 6eeeebb

Correct, and this was the finding worth making. Both existing tests complete a full 32-asset sweep in about 15s against a 300s budget, so `time.monotonic() >= deadline` never fired and `next_offset` was always 0. The chunking behavior the PR exists to add had no automated coverage at all.

Added two deterministic tests, both offline against a stubbed `BigQueryResource` so they need no warehouse access:

- `test_bigquery_table_sensor_budget_exhausted_polls_nothing` — budget of 0 must break before the first call, leaving the resume index untouched so the next tick retries the same slice instead of skipping it. Asserts `get_table` was never called.
- `test_bigquery_table_sensor_resumes_mid_sweep` — a fake monotonic clock stepping 100s against a 300s budget puts the deadline after exactly two assets. Asserts exactly two `get_table` calls and that the offset advanced by exactly two, which is the resume behavior itself.

I also took the second half of your suggestion: both start from a non-zero offset. That matters more than it looks, because with a starting offset of 0 a break at index 0 records `next_offset = 0`, which is indistinguishable from a completed sweep. A non-zero start is the only way the assertion can tell "stopped here" from "finished".

Rather than assume the new tests are meaningful, I mutation-checked them: disabling the budget check makes both fail, and the file restored byte-exact afterward. They are not vacuous.

One implementation note, since it would have bitten anyone extending these tests: `bigquery/__init__.py` does `from ...sensors import sensors`, which rebinds the name to the sensor **list**. So `import ...bigquery.sensors as x` hands back a list, and monkeypatching it silently no-ops. The tests reach the real module via `importlib.import_module`.

### 2. Offset validity bounds length, not order — DECLINED, comment broadened instead

The analysis is right and I have not tried to argue it away: a reorder that preserves the count is invisible to `0 <= offset < len(asset_selection)`, and a resumed tick would poll a different slice than the one it skipped.

Declining the code change on impact. Per-asset state is keyed by `python_identifier`, not by index, so nothing is lost or duplicated — the worst case is that a few assets are polled one sweep later than they otherwise would be, which at a 300s interval is a bounded, self-healing detection delay. Closing it properly means storing a resume identifier instead of an index and scanning for it, which is more machinery than a benign, rare, self-correcting event justifies.

What I did do is stop the code from implying the guard is stronger than it is. The comment now names reordering explicitly and states why it is bounded, so the next reader does not have to rediscover this.

### 3. `deadline` set before entering `get_client()` — DECLINED, no action

Agreed on both the mechanism and your own read that it is not actionable. `google.cloud.bigquery.Client()` construction is local and non-blocking, and the failure mode if it ever were slow is no forward progress for one tick rather than a regression — the next tick retries the same slice. Recording it here so the reasoning is on the record rather than silently dropped.

### 4. Budget equals `minimum_interval_seconds` — DECLINED, no action

Confirmed both are 300, and your reading of `minimum_interval_seconds` as a floor between tick starts rather than an idle gap is correct.

Leaving it. The budget's job is to fit inside the 600s tick timeout with margin, and 300s plus one 30s in-flight call is 330s, comfortably under. Back-to-back ticks during sustained BigQuery slowness are the intended self-limiting behavior: the sensor keeps making forward progress through the sweep instead of stalling. If the continuous gRPC worker occupancy ever becomes a problem, the lever is lowering the budget, not raising the interval — noted for whoever hits that.

### 5. `SkipReason` wording on a mid-sweep chunk — ACCEPTED, fixed in 6eeeebb

A partial slice now reports itself as partial and names the resume index, rather than saying "No modified tables." as though the full sweep had completed. Cheap, and it makes the Dagster UI honest about what happened. Confirmed independently that nothing parses the message, so this is safe to change.

---

Not re-requesting review — `claude-review` only fires on `opened` and `ready_for_review`, not `synchronize`, so it will not re-run on this push.

PR #4914, opened 2026-08-18
fix(k8s): code server scheduling resilience Phase 1, plus design spec and plan

### LINKED ISSUE #4907: fix(k8s): code server reload failures from deploymentStartupTimeout default plus scheduling over-constraint

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make code locations much less likely to fail to start, and it adds the written design and plan behind that work.

Background. When a code location fails to start, it stays failed until someone pushes a commit or clicks redeploy — the agent never retries on its own. That is why this keeps needing a person. The reason it fails is that the replacement pod cannot find a machine to run on quickly enough, and the agent gives up after 5 minutes while Kubernetes is still working on it.

Three changes, none of which reverse a decision anyone made deliberately:

- The agent now waits 15 minutes instead of 5 for a replacement to get a machine. The 5 minute limit was never set by us; it is the default, and it is shorter than how long getting a new machine normally takes.
- Code servers are now protected from being moved when the cluster shrinks. Every other pod type already had this protection; code servers were the only ones missing it, which cost 38 to 59 needless moves a week.
- One scheduling rule is relaxed from a hard requirement to a preference. As a hard rule it blocked the replacement from the single best machine available, forcing a brand-new one to be built on every redeploy.

Also fixes documentation that actively misled: the notes described only one of the three startup timeouts, which is why the wrong one was blamed first.

Worth knowing: the first two changes need a manual `helm upgrade` and do **not** take effect on merge. The third ships normally.

This is Phase 1 of three. Phases 2 and 3 get their own plans after this one's effect is measured, since that measurement decides whether they are still needed at the proposed scale.

Refs #4907

## Reviewer Notes

- The relaxed scheduling rule is the change most worth your judgement. My argument is that with one replica per location it spreads nothing, and its only real effect was excluding the outgoing pod's machine — the one machine already known to be the right type, sized correctly, and free of run pods. If that reasoning is wrong, the largest change here is wrong. Full detail below.
- The isolation between run pods and code servers is **untouched and still strict.** That is the rule from the earlier incident where run pods starved code servers, and there is an assertion in the plan that fails if it is ever weakened. Same for code server priority, which stays as it is.
- Relaxing that rule is a structural change, not a renamed key — the preferred form nests the term differently. I verified all five files parse into the right shape and that each selects its own location rather than a copied one, which is the mistake this kind of edit invites.
- The 5-to-15-minute change has one accepted downside: a genuinely broken deployment now takes 15 minutes to report instead of 5. Broken images are unaffected; they surface separately in 30 seconds.
- The spec deliberately marks two assumptions as unverified rather than asserting them, each gating the later phase that depends on it. Neither is exercised by this PR.

### General

- [x] Review the **Claude Code Review** comment posted on this PR — it did not run
      on this branch. `claude-code-review.yaml` fires on `opened` and
      `ready_for_review`, and this PR was opened while it was still docs-only, so
      markdown was excluded. If you want a bot pass over the config changes,
      toggling draft state re-fires it.

### Dagster

      **not applicable, deliberately.** `dagster-cloud.yaml` is agent-side
      deployment config; `definitions validate` does not read it, so the check
      cannot detect a defect in this change. Verified instead by asserting the
      parsed YAML structure of all five files (see the plan, Task 2 Step 3).
      location — same reason. It shells out to `definitions validate`, so it is
      equally blind to this change, and it needs five `prepare-and-package` runs
      in a fresh worktree to go green at all.
      [Library + Config pattern](https://teamschools.github.io/teamster/reference/adding-an-integration/) —
      n/a, no new integration

### dbt

Skipped, no dbt changes.

### Docs

      catalog — n/a, no schedule or sensor changed
      `CLAUDE.md` — n/a. `.k8s/CLAUDE.md` is updated, for the timeout
      documentation gap rather than for an integration.

The spec and plan land under `docs/superpowers/`, deliberately excluded from the
`mkdocs.yml` nav per `docs/CLAUDE.md`, so no nav entry is needed.

## CI checks

- [x] **Trunk** — passes. `trunk check --force --no-fix` was also run locally
      against each changed file from inside the worktree, with `pwd` confirmed,
      before every commit. Only prettier reflow was reported, which the
      pre-commit `fmt` hook applied and re-checked clean.
- [x] **dbt Cloud** — passes. No dbt paths changed; it triggers on PR events
      regardless.
- [x] **Dagster Cloud** — passes. All 32 checks are green, with the
      `dagster-cloud-deploy` jobs fanning out 5x (one per code location) because
      every `dagster-cloud.yaml` changed.

**What CI does NOT prove here.** A green deploy job is not evidence the agent
reloaded, per `.github/CLAUDE.md`. I could not confirm a branch deployment exists
(`list_deployments` returns only `prod`), so the relaxed affinity has **not** been
exercised against a live cluster by this PR. Verified separately that prod was not
touched: its locations still carry `main` commits, not this branch's head. The
first real test of the affinity change is the post-merge prod deploy, and of the
Helm changes, the manual `helm upgrade`.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Investigation, spec, plan, and Phase 1
implementation. Phases 2 and 3 are not implemented.

### Reviewer Notes flag 1: the relaxed rule, in full

Each location's code server carried
`requiredDuringSchedulingIgnoredDuringExecution` anti-affinity against its **own**
`location_name` on `kubernetes.io/hostname`, set per district in
`dagster-cloud.yaml`. With one replica per location it spreads nothing. Its only
live effect was forbidding the replacement pod from the outgoing pod's node during
a rollover — and that node is the single best candidate: it already has arm64
Scale-Out capacity sized for a code server, and the run-pod anti-affinity
guarantees no run pods are on it. So the rule specifically excluded the one node
guaranteed to satisfy every other constraint, forcing NAP to build a new node on
every redeploy.

Cross-location spreading is unaffected — it comes from
`topologySpreadConstraints` (`ScheduleAnyway` on zone and hostname) in the Helm
values. `affinity` is permitted per-location:
`onlyAllowUserDefinedK8sConfigFields.podSpecConfig.affinity` is `true`.

The structural point: `preferred` takes `WeightedPodAffinityTerm`, not
`PodAffinityTerm`, so the term nests under `podAffinityTerm` with a `weight`.
Weight is 100, the maximum, so separation is still preferred as strongly as
possible. A key rename alone would have produced config the agent rejects.

### Why the failure is terminal

Verified in `dagster_cloud/workspace/user_code_launcher/user_code_launcher.py`,
not inferred. `_should_trigger_recovery` returns `False` when
`location_key in control_plane_error_locations` — "control plane agrees there is
an error, don't retry". Recovery fires only when the agent holds a local error
while the control plane believes the location is healthy. Normal reconciliation
redeploys only when
`actual_entry.update_timestamp != desired_entry.update_timestamp`, which changes
on a new deploy, and the periodic health check needs a running endpoint a
never-started pod lacks. No setting changes this;
`DAGSTER_CLOUD_DISABLE_LOCAL_ERROR_SERVER_RECOVERY` only turns recovery off.

### The timeout, and my own wrong turn

`workspace.deploymentStartupTimeout` was unset, so 300s
(`DEFAULT_DEPLOYMENT_STARTUP_TIMEOUT` in the launcher). It governs
`wait_for_deployment_complete`, which is what raises on the observed failure.
`serverProcessStartupTimeout` (300) governs readiness *after* the Deployment
exists via `_wait_for_dagster_server_process`. I blamed the latter first and was
wrong; reading the installed package caught it before it reached the spec. That is
exactly the confusion the `.k8s/CLAUDE.md` fix in this PR removes.

### Measured baseline

Code-server pods only (`involvedObject.name` containing `-prod-`) in
`dagster-cloud`. 30-day retention, so nothing before 2026-07-19 is observable and
a June inflection cannot be ruled out.

| Week | gRPC errors | Killing | ScaleDown | Preempted | Evicted | FailedScheduling |
| Jul 20-27 | 85 | 122 | 42 | 21 | 0 | 472 |
| Jul 28-Aug 4 | 134 | 197 | 59 | 37 | 0 | 454 |
| Aug 4-11 | 106 | 118 | 48 | 22 | 0 | 213 |
| Aug 11-18 | 100 | 142 | 38 | 32 | 0 | 438 |

The reported recent uptick is **not** present: the peak was Jul 28 to Aug 4 and
the two most recent weeks are lower. Chronic, not a regression. Agent errors track
code-server kills at 0.70, 0.68, 0.90, 0.70, which is what ties the errors to pod
deaths rather than repo code. `Evicted` at 0 across all four weeks is the evidence
the isolation is working, and the reason it stays.

### Verification performed

Assertions rather than tests, because Helm values and YAML affinity have no test
harness:

- `deploymentStartupTimeout == 900`; `safe-to-evict` is the **string** `"false"`
  (a bare YAML boolean would be wrong); code-server requests still `500m` /
  `2.0Gi`, meeting the Autopilot extended-duration minimum the annotation
  requires; `runK8sConfig` anti-affinity still `required` with **both** terms
  (code-server labels and agent labels); no `priorityClassName` on
  `serverK8sConfig`.
- All five location files: `required` key gone, exactly one weighted term at
  weight 100, `topologyKey` unchanged, and each selects its own `location_name`.

### Still outstanding, for whoever picks this up

- The `helm upgrade` has not been run. Until it is, only the affinity change is
  live. Plan Task 4 Steps 4 to 6 carry the exact commands and what to check,
  including that the annotation only appears on code-server pods created after the
  upgrade.
- The post-change measurement (plan Task 4 Step 7) has not happened. Do not plan
  Phase 2 or 3 before it exists — if `FailedScheduling` collapses, Phase 2's
  continuous cost may no longer be justified.

### Loose ends found and deliberately not fixed here

- `kipptaf__google__bigquery__table_modified_sensor` is absent from
  `docs/reference/automations.md`; the generator appears to skip sensors that emit
  asset events rather than run requests.
- The `bug` label the new issue template applies is unused in practice — every
  real issue uses `fix`.
- Sibling PR #4913 fixes an unrelated sensor timeout that surfaces the same
  `DagsterUserCodeUnreachableError` class.

</details>

### DIFF
--- a/.k8s/CLAUDE.md
+++ b/.k8s/CLAUDE.md
@@ -28,9 +28,17 @@ on GKE Autopilot.
 - **Helm deploy is manual** — editing `values-override.yaml` is fine, but
   changes only take effect after `helm upgrade`. `git push` builds code location
   images, not Helm agent config.
+- `deploymentStartupTimeout` (Helm `workspace` key, chart default 300s) — time
+  the agent waits for the code server Deployment to become ready, i.e. for the
+  pod to be SCHEDULED. Currently set to 900s. This is the timeout that fires
+  during a NAP `FailedScheduling` wait. Raising it also raises the agent's
+  worst-case first reconcile, so the `readinessProbe` `failureThreshold` and
+  `DAGSTER_CLOUD_CLEANUP_SERVER_GRACE_PERIOD_SECONDS` must move with it, along
+  with `install.sh`'s `rollout status --timeout`.
 - `serverProcessStartupTimeout` (Helm `workspace` key, default 180s) — time the
-  agent waits for a code server gRPC ping after creating the Deployment.
-  Currently set to 300s in `values-override.yaml`.
+  agent waits for a code server gRPC ping after the Deployment exists. Fires on
+  slow definitions import, not on failed placement. Currently set to 300s in
+  `values-override.yaml`.
 
 ## Scheduling
 
--- a/.k8s/CLAUDE.md
+++ b/.k8s/CLAUDE.md
@@ -269,10 +269,21 @@ a different pod type.
 - **GCP Error Reporting investigation**: `list_group_stats` (find groups) →
   `list_log_entries` (reconstruct multi-line tracebacks from individual log
   entries) → `k8s_pod` events (find root cause: preemption, OOM, eviction).
-- **Timeout types** (do not conflate): Startup (`serverProcessStartupTimeout`,
-  default 180s) = agent→code server gRPC ping; failure → deployment removed +
-  replacement, causes churn. Sensor execution (300s) = sensor ran too long; code
-  server stays up, no churn.
+- **Timeout types** (do not conflate). Three distinct waits, in the order they
+  occur:
+  1. **Deployment startup** (`deploymentStartupTimeout`, Helm `workspace` key,
+     chart default 300s, set to 900s) — agent waits for the code server
+     Deployment to become ready, i.e. for the pod to be SCHEDULED. This is the
+     one that fires during a NAP `FailedScheduling` wait, and its failure
+     message is `Timed out waiting for deployment {name}` with
+     `Pod status: Pending`. Governs `wait_for_deployment_complete`, not
+     `_wait_for_dagster_server_process`.
+  1. **Server process startup** (`serverProcessStartupTimeout`, Helm `workspace`
+     key, default 180s, set to 300s) — agent waits for a gRPC ping AFTER the
+     Deployment exists. Fires when Dagster definitions are slow to import, not
+     when the pod cannot be placed.
+  1. **Sensor execution** (300s, Dagster+ deployment setting) — sensor ran too
+     long; the code server stays up and no churn results.
 - **Code server startup failure signals**: Check ALL pods for the deployment.
   `Aborted!` stderr = SIGABRT (native crash). `DagsterExecutionInterruptedError`
   = SIGTERM during import (rollover). Silent hang after "Starting Dagster code
@@ -293,9 +304,14 @@ a different pod type.
   `resource.type="gce_subnetwork"` +
   `logName=".../compute.googleapis.com%2Ffirewall"` → `instance.zone` +
   `remote_instance.zone`. Filter `dest_port=4000` for agent→code-server gRPC.
-- **Autopilot node pre-warming**: Not possible — no DaemonSets, no image
-  pre-pulling, no node lifecycle control. Only levers for cold-node startup
-  latency: image size reduction and Dagster timeout increases.
+- **Autopilot node pre-warming**: no node-lifecycle control — no DaemonSets, no
+  image pre-pulling, no cordon that holds. But capacity CAN be reserved at the
+  workload layer with balloon pods: a Deployment of placeholder pods on a
+  negative `PriorityClass` holds nodes that real pods preempt instantly, and the
+  displaced placeholder then triggers NAP to re-warm the buffer. Google
+  documents this for Autopilot. Not deployed here yet; see
+  `docs/superpowers/specs/2026-08-18-code-server-scheduling-resilience-design.md`
+  Phase 2, gated on verifying Warden admits a negative priority value.
 - **CPU limit alert sensitivity**: `GKE Container - High CPU Limit Utilization`
   fires at >90% `ALIGN_MEAN` over 60s with `count=1`. When an asset's `op_tags`
   CPU limit alerts, bump by 250m — re-measure peak with
@@ -317,6 +333,19 @@ a different pod type.
   Dagster+ full deployment settings (see Dagster+ Deployment Settings section)
   expose no tick-retry knob. Terminal `DagsterUserCodeUnreachableError` ticks
   remain terminal.
+- **A timed-out code server deployment is TERMINAL — the agent never retries
+  it.** `_should_trigger_recovery` in
+  `dagster_cloud/workspace/user_code_launcher/user_code_launcher.py` returns
+  `False` when the location is in `control_plane_error_locations` ("control
+  plane agrees there is an error, don't retry"). Recovery fires only when the
+  agent holds a local error while the control plane believes the location is
+  healthy. Normal reconciliation redeploys only when
+  `actual_entry.update_timestamp != desired_entry.update_timestamp`, which
+  changes on a new deploy, and the periodic health check needs a running
+  endpoint that a never-started pod does not have. So a location that fails to
+  schedule stays `ERROR` until someone pushes a commit or clicks redeploy. No
+  agent setting changes this;
+  `DAGSTER_CLOUD_DISABLE_LOCAL_ERROR_SERVER_RECOVERY` only turns recovery off.
 
 ## Dagster+ Deployment Settings
 
--- a/.k8s/dagster/install.sh
+++ b/.k8s/dagster/install.sh
@@ -8,12 +8,12 @@ set -euo pipefail
 # point back to setup.sh on any error (the clean "Aborted." exit 0 path skips
 # this trap).
 on_error() {
-  status=$?
-  echo "" >&2
-  echo "install.sh failed (exit ${status})." >&2
-  echo "Missing tool, unreachable cluster, or missing namespace? Run the" >&2
-  echo "environment bootstrap first, then retry:" >&2
-  echo "  bash .k8s/setup.sh" >&2
+	status=$?
+	echo "" >&2
+	echo "install.sh failed (exit ${status})." >&2
+	echo "Missing tool, unreachable cluster, or missing namespace? Run the" >&2
+	echo "environment bootstrap first, then retry:" >&2
+	echo "  bash .k8s/setup.sh" >&2
 }
 trap on_error ERR
 
@@ -24,22 +24,26 @@ helm show values dagster-cloud/dagster-cloud-agent >.k8s/dagster/values.yaml
 
 echo "Running dry-run..."
 helm upgrade \
-  --install user-cloud dagster-cloud/dagster-cloud-agent \
-  --namespace dagster-cloud \
-  -f .k8s/dagster/values-override.yaml \
-  --dry-run
+	--install user-cloud dagster-cloud/dagster-cloud-agent \
+	--namespace dagster-cloud \
+	-f .k8s/dagster/values-override.yaml \
+	--dry-run
 
 read -r -p "Dry-run succeeded. Apply to cluster? [y/N] " response
 if [[ ! ${response} =~ ^[Yy]$ ]]; then
-  echo "Aborted."
-  exit 0
+	echo "Aborted."
+	exit 0
 fi
 
 helm upgrade \
-  --install user-cloud dagster-cloud/dagster-cloud-agent \
-  --namespace dagster-cloud \
-  -f .k8s/dagster/values-override.yaml
+	--install user-cloud dagster-cloud/dagster-cloud-agent \
+	--namespace dagster-cloud \
+	-f .k8s/dagster/values-override.yaml
 
 # ConfigMap changes don't trigger a rollout — force restart to pick up new env vars.
 kubectl rollout restart deployment/user-cloud-dagster-cloud-agent-agent -n dagster-cloud
-kubectl rollout status deployment/user-cloud-dagster-cloud-agent-agent -n dagster-cloud --timeout=600s
+# Must match the agent readinessProbe budget in values-override.yaml
+# (failureThreshold x periodSeconds = 1300s). A shorter timeout here reports a
+# healthy-but-slow first reconcile as a failure and fires the on_error trap,
+# which then misdirects the reader to setup.sh.
+kubectl rollout status deployment/user-cloud-dagster-cloud-agent-agent -n dagster-cloud --timeout=1300s
--- a/.k8s/dagster/values-override.yaml
+++ b/.k8s/dagster/values-override.yaml
@@ -31,10 +31,14 @@ dagsterCloudAgent:
   env:
     # Reduce cleanup delay for orphaned code server deployments left by previous
     # agent IDs after Helm upgrades. Defaults: grace=3600s, interval=1800s.
-    # With replicas: 1, each upgrade leaves 1 orphaned deployment per location;
-    # the 10-min heartbeat expiry + 15-min grace period ensures replacements are
-    # reconciled before orphans are deleted.
-    DAGSTER_CLOUD_CLEANUP_SERVER_GRACE_PERIOD_SECONDS: "900"
+    # With replicas: 1, each upgrade leaves 1 orphaned deployment per location.
+    # The grace period must exceed the worst-case first reconcile so replacements
+    # are reconciled before orphans are deleted -- CLAUDE.md: "do not set grace
+    # period below reconciliation time". That worst case is now ~1200s (see the
+    # readinessProbe note below), so 900s would sit under it; 1500s restores the
+    # invariant. Cost is bounded: orphans linger ~10 min longer, and only after a
+    # helm upgrade.
+    DAGSTER_CLOUD_CLEANUP_SERVER_GRACE_PERIOD_SECONDS: "1500"
     DAGSTER_CLOUD_CLEANUP_SERVER_CHECK_INTERVAL: "600"
 
   readinessProbe:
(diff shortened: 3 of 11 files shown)


cbini:
Verdicts on the review. Two Important, four Minor, three Recommendations — **all nine accepted**, none declined. Fixed in 0c12472, 5155d1a, and 04bdb62.

The review caught a defect I would have shipped, and one of its own fixes was incomplete. Both noted below.

---

### Important 1 — `deploymentStartupTimeout: 900` breaks the agent's readiness budget — ACCEPTED, and the prescribed fix was incomplete

Verified from source rather than taken on faith:

- `readinessProbe`: `failureThreshold: 60 × periodSeconds: 10` = 600s
- `user_code_launcher.py:1533`: `if self._reconcile_count == 0 and self._requires_healthcheck: self._write_readiness_sentinel()` — the sentinel is written only after the first full `_reconcile` returns
- Worst case per location was `300 + 300` = 600s, exactly the budget. Now `900 + 300` = 1200s

This is the finding that mattered, and it bites on precisely the `helm upgrade` that ships the change. Raised `failureThreshold` to 130 (1300s).

**Your fix was incomplete, though.** Raising `failureThreshold` stops the pod being killed, but `install.sh:45` carries its own `kubectl rollout status --timeout=600s`. Left alone, the script still reports a healthy-but-slow upgrade as a failure and fires the `on_error` trap — which is the misleading "go run setup.sh" message you predicted, just from a different source. Raised that to 1300s to match, and added a comment at each site naming the other.

I also took the second half of your finding as a real change rather than a comment fix. You noted the orphan-cleanup reasoning ("10-min heartbeat expiry + 15-min grace period ensures replacements are reconciled before orphans are deleted") is no longer true of the worst case. `.k8s/CLAUDE.md` says not to set the grace period below reconciliation time, and 900s now sits under 1200s, so I raised `DAGSTER_CLOUD_CLEANUP_SERVER_GRACE_PERIOD_SECONDS` to 1500s rather than documenting a known-broken invariant. Cost is bounded: orphans linger ~10 min longer, and only after an upgrade.

All three budgets are now asserted in `tests/test_k8s_config.py`, so this cannot silently regress the next time someone moves a timeout.

### Important 2 — rollover co-location risks not modelled — ACCEPTED

Both consequences are real and both were missing from the failure-modes table, which modelled only requests. Added as separate rows:

- **A single run-pod preemption can now take BOTH code servers for a location.** The hard self-anti-affinity previously forced the incoming and outgoing pods onto separate nodes, so one preemption could only ever take one. That is a genuinely new exposure and it lands in the exact window the change exists to protect.
- **Two co-located pods requesting 500m / 2.0Gi can burst to their 1000m / 2.5Gi limits.** `safe-to-evict: "false"` does not block kubelet node-pressure eviction, and priority-0 code servers go first.

`Preempted` is promoted from informational to must-not-regress in the verification step, and a check that no node ends up hosting two code servers for one location once a rollover settles is added.

**On the vocabulary point you raised — you are right and it is worse than a wording nit.** `.k8s/CLAUDE.md` explicitly forbids describing the run-pod anti-affinity as "isolating" code servers or "preventing co-location", because a `required` anti-affinity authorizes preemption of the target pods. My spec used exactly that vocabulary, in the constraint the entire design is built around, and Task 3 was the documentation pass that should have caught it. The spec now states what the rule actually guarantees — that no node holds a run pod and a code server at the same instant — and that the mechanism is code servers being preempted, which is why `Preempted` runs at 21–37/week while `Evicted` stays at 0.

### Minor 3 — Helm section still listed only `serverProcessStartupTimeout` — ACCEPTED

Fair, and a bit embarrassing given closing that exact gap was the stated point of the commit. Someone editing Helm values reads the Helm section, not Troubleshooting. It now carries `deploymentStartupTimeout` with its default, its current value, what it governs, and the three budgets that must move with it.

### Minor 4 — 900s applies to branch deployments too — ACCEPTED

Added to the comment at the setting. A branch deployment that cannot schedule now takes 15 minutes to report instead of 5, which the "accepted downside" paragraph only discussed for prod crash-loops.

### Minor 5 — Task 2's top stated risk was never tested — ACCEPTED, and this was the sharpest process criticism

Correct, and it is the same failure mode you flagged on the sibling PR: the plan's own "critical structural note" said a wrong shape "produces config the agent rejects at deploy time", then verified it with raw YAML key assertions that would pass on `podAffinityTerms` (plural) or a stray `weight` inside `podAffinityTerm`.

`tests/test_k8s_config.py` now runs the coercion you suggested. I confirmed it is decisive before relying on it — both malformed forms are rejected with `Unexpected keys in model class V1WeightedPodAffinityTerm` — and included a negative case asserting that rejection so the positive tests cannot pass vacuously.

I did **not** add the allowlist validator as a second test. You established that the allowlist is field-granular at `pod_spec_config.affinity` and never inspects the nested shape, so a test around it would assert that `affinity: true` permits `affinity` — near-tautological. Your one-off verification that it passes is the useful part, and it is recorded here rather than re-implemented.

### Minor 6 — pre-existing `minAvailable: 1` doc error — ACCEPTED

Fixed. The comment claimed code servers are PDB-protected by `minAvailable: 1` while every code-server PDB in the same file uses `maxUnavailable: 1`, and `.k8s/CLAUDE.md` explicitly says not to switch. Cheap to correct in a PR whose third commit is a documentation-accuracy pass, as you say.

### Recommendation — `weight: 100` justification is vacuous — ACCEPTED

Right: with a single `preferred` term the InterPodAffinity plugin normalizes across candidate nodes, so `weight: 1` and `weight: 100` give identical ordering. Kept 100 as convention, dropped the "prefers separation as strongly as it can" claim so it cannot become the basis of a later decision.

### Recommendation — reword the "reuse the outgoing pod's node" claim — ACCEPTED, and it is the better argument

This was the most useful correction in the review. The `preferred` term still penalizes that node and the hostname `topologySpreadConstraint` penalizes it again, so the scheduler picks elsewhere whenever anything else fits. What the change buys is that the node becomes **feasible**, which is what removes the NAP wait. Corrected in the spec's Phase 1 table and in all five `dagster-cloud.yaml` comments.

### Recommendation — add `Preempted` to the must-not-regress list — ACCEPTED

Done, per Important 2.

---

Not re-requesting a bot pass: `claude-review` fires on `opened` / `ready_for_review`, not `synchronize`. It never ran on this PR at all, since it was opened while docs-only and markdown is excluded.

PR #4917, opened 2026-08-18
feat(gradebook-audit): pivot the expectations upload template wide

GabyRangelB:
# Pull Request

## Summary & Motivation

...pivot `rpt_gsheets__gradebook_audit_template` so the four assignment
categories are **columns** — `W`, `H`, `F`, `S` — instead of four rows.

This model feeds the Google Sheet T&L exports as the CSV they upload into
PowerSchool's `U_EXPECTATIONS` via the KIPP NJ Gradebook Audit plugin. It is an
upload grid a person fills in, not an analytical long table, so the wide shape
matches how it is actually used. It also means a category with no count now
shows up as a blank cell to fill in, rather than silently vanishing as a missing
row.

Alongside the pivot:

- Dropped `assignment_category_code`, `expectation`,
  `assignment_category_term`, and `assignment_category_name` — all lose their
  meaning once the categories are columns.
- Narrowed the uniqueness test to
  `region` + `school_level` + `quarter` + `week_number_quarter`.
- Removed the dead `where expectation is not null` filter. BigQuery `UNPIVOT`
  already excludes nulls by default, so that filter never dropped a row.
- Updated the published reference doc, whose stated grain and "category decode"
  description are now wrong, and recorded why there is deliberately no all-null
  guard so it does not get added later as a tidy-up.

Both `/* summer toggle: see skill */` markers are untouched.

### Validation

Built against prod upstreams with `--favor-state`, after confirming from the
compiled SQL that both refs resolve to `kipptaf_powerschool` rather than stale
personal dev copies (the first build read the dev copies, which would have made
the parity check meaningless).

| Check                                | Result                                                                                   |
| ------------------------------------ | ---------------------------------------------------------------------------------------- |
| `dbt build`, contract enforced       | PASS                                                                                     |
| Uniqueness test on the 4-column key  | PASS                                                                                     |
| Row count                            | 202 (was 808, i.e. 202 x 4)                                                              |
| Columns                              | `W` / `H` / `F` / `S` as `INT64`, ordinal positions 8-11                                 |
| Value parity vs the prod long output | 0 differences across all four categories and every passthrough column; 0 keys gained/lost |

### Rollback plan

Revert the commit. The model is a view, so the prior long shape is restored on
the next build with no backfill needed. No dbt model refs this one, so nothing
downstream in the project is affected.

### General

### dbt

- [x] Properties file updated — new `W` / `H` / `F` / `S` columns declared with
      `data_type` and `quote: true`, removed columns dropped, model description
      and uniqueness test updated.
- [x] Exposure is current — `rpt_gsheets__gradebook_audit_template` in
      `models/exposures/google-sheets.yml` already covers this model; no change
      needed since neither the model name nor its consumer changed.
- [x] **Breaking change — yes, deliberately.** This removes four columns from a
      contracted `rpt_` model and changes the sheet's headers, cutting it from
      808 rows to 202. Any formula or IMPORTRANGE built on the long shape will
      break. The exposure owner (Data Team - grangel) is the PR author and is
      aware; rollback plan is in the summary above.
- [ ] n/a — no external source added or modified.

## CI checks

- [ ] **Trunk** — `trunk check --force` was run locally on all three changed
      files before pushing: no issues.
- [ ] **dbt Cloud** — passes.
- [ ] **Dagster Cloud** — n/a, dbt-only change.

<details>
<summary>For Claude</summary>

Worth a close look: the removal of `where expectation is not null`. It is safe
because BigQuery's `UNPIVOT` excludes NULL values by default, so the predicate
could never have matched a row — but that reasoning is the crux of the change,
so please check it rather than assume it.

Deliberately out of scope, please do not flag:

- The summer toggle. Both `{{ var("current_academic_year") - 1 }}` expressions
  and their `summer toggle: see skill` markers are intentionally left as-is.
- Adding an all-null guard such as
  `where coalesce(cnt_w, cnt_h, cnt_f, cnt_s) is not null`. This was considered
  and rejected on purpose: it filters zero rows today (verified — all 205
  `U_EXPECTATIONS` rows carry a count in all four columns), and in the only
  future where it would fire it would hide exactly the gap this template exists
  to surface. The reasoning is recorded in the reference doc.
- The backticks on the `W` / `H` / `F` / `S` aliases. They are there because
  `src/dbt/` contains zero unquoted uppercase aliases and sqlfluff CP02 defaults
  to `consistent`; they do not change the emitted column names, and they match
  how `` `quarter` `` is already written in the same file.

No GitHub issue — the user explicitly declined one for this change.

</details>

---
  - https://app.asana.com/0/0/1217604903764513

### DIFF
--- a/docs/models/gradebook-audit-data-model.md
+++ b/docs/models/gradebook-audit-data-model.md
@@ -275,9 +275,11 @@ Three gotchas worth knowing before editing it:
   the gap was invisible in the sheet; wide, it surfaces as an empty cell for
   whoever is setting counts to fill in. Adding an all-null guard would hide
   exactly what the template exists to show, and re-uploading a blank row writes
-  back the blank state it came from. Measured 2026-08-18, all 205
-  `U_EXPECTATIONS` rows carry a count in all four columns, so no guard would
-  filter anything today anyway.
+  back the blank state it came from. Measured 2026-08-18, all 205 rows in
+  `stg_powerschool__u_expectations` carry a count in all four columns, so no
+  guard would filter anything today anyway. That 205 is the raw source count,
+  not this model's output — the template emits 202 rows, because `term_weeks`
+  keeps only weeks that have already started.
 
 Coverage follows the inner join to `stg_powerschool__u_expectations` — Newark
 and Camden at MS and HS, Paterson at MS, no ES and no Miami — so it needs no
--- a/docs/models/gradebook-audit-data-model.md
+++ b/docs/models/gradebook-audit-data-model.md
@@ -248,14 +248,16 @@ PowerSchool's `U_EXPECTATIONS` via the plugin — the write side of the same tab
 `rpt_gsheets__gradebook_audit_template` in
 `src/dbt/kipptaf/models/exposures/google-sheets.yml`.
 
-One row per
-`region × school_level × quarter × week_number_quarter × assignment_category_code`
-(uniqueness-tested). It shares the QTD unpivot's `term_weeks` CTE and category
-decode, but differs in two ways: it keeps **every** school week rather than
-collapsing to the most recently completed one per quarter, and it takes
-`school_week_end_date` (aliased `week_end_friday`) instead of `week_end_sunday`.
+One row per `region × school_level × quarter × week_number_quarter`
+(uniqueness-tested), carrying the four category counts as **columns** — `W`,
+`H`, `F`, `S` — rather than one row each. It shares the QTD unpivot's
+`term_weeks` CTE but differs in three ways: it keeps **every** school week
+rather than collapsing to the most recently completed one per quarter; it takes
+`school_week_end_date` (aliased `week_end_friday`) instead of `week_end_sunday`;
+and it stays wide, because the sheet is an upload grid a person fills in, not an
+analytical long table.
 
-Two gotchas worth knowing before editing it:
+Three gotchas worth knowing before editing it:
 
 - **`week_end_friday` is not always a Friday.** It is the last _in-session_ day
   of the school week, so it lands on Thursday or earlier when a holiday or PD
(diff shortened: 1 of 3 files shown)


GabyRangelB:
## Review verdicts

Each finding was verified against the codebase before responding. All three confirmed non-blocking; one fixed.

### 1. Breaking change not routed through dbt model versioning — declined, no action

Confirmed the exception independently rather than relying on the review's own self-resolution: grepping for `versions:`, `latest_version`, and `deprecation_date` across all sixteen dbt projects returns **zero** models. Versioning this extract would introduce a governance pattern the repo does not use anywhere, for a model whose only consumer is the Google Sheet in its own exposure — and whose exposure owner is the PR author.

The dbt Cloud run surfaces the matching warning, which is expected rather than a problem:

```text
[WARNING]: Breaking change to contracted, unversioned model
rpt_gsheets__gradebook_audit_template
  - Columns were removed: assignment_category_code, assignment_category_name
```

Accepting it knowingly. Rollback plan is in the PR description.

### 2. UNPIVOT / null-filter claim — confirmed, no action

Agreed. This was the one claim flagged for scrutiny rather than assertion, and it holds from two independent directions: the repo's standing note in `src/dbt/CLAUDE.md`, and BigQuery's documented default `UNPIVOT` behaviour. Separately, the removed predicate matched zero rows in prod at the time of the change.

### 3. "205 vs 202" ambiguity — valid, fixed in 5249457

Correct that these are two different populations: 205 raw rows in `stg_powerschool__u_expectations`, 202 after the `term_weeks` inner join drops weeks that have not started yet. The doc now names the staging model, states the 202 explicitly, and gives the reason the two differ, so a future reader comparing the doc against this PR does not read it as a contradiction.

### On the backtick-quoted aliases

Noted for the record, since the review checked this and arrived at the same conclusion from different evidence: `src/dbt/` carries **87** backtick-quoted uppercase aliases, concentrated in `extracts/` models feeding external tools — the same situation as this model. That is stronger precedent than the sqlfluff CP02 reasoning originally given in the PR body.

### Scope note on validation

The review disclosed it had no warehouse access, so the row-count and parity figures in the PR description were not independently re-verified by it. Those numbers come from two local builds against prod upstreams using `--favor-state`, after confirming from the compiled SQL that both refs resolved to `kipptaf_powerschool` rather than stale personal dev copies. The first build did read the dev copies and its results were discarded.

anthonygwalters (approved):

PR #4918, opened 2026-08-18
feat: ingest NJGPA summative record files from Cambium TIDE

### LINKED ISSUE #4899: feat: ingest NJGPA summative record files from Cambium TIDE
## Context

New Jersey moved NJGPA score reporting from Pearson Access Next to Cambium TIDE
beginning with the Spring 2026 administration. Two SY25-26 files are in hand:

| File                                                       | District code | Code location | Rows |
| ---------------------------------------------------------- | ------------- | ------------- | ---- |
| `2026_Spring_7325_District_Summative_Record_File_GPA.csv`   | 7325 (TEAM Academy Charter School) | `kippnewark` | 598 |
| `2026_Spring_1799_District_Summative_Record_File_GPA.csv`   | 1799 (KIPP Cooper Norcross)        | `kippcamden` | 252 |

Only **11 of 225** column names survive slugification into the existing NJGPA
Avro schema. Pearson shipped camel-case headers and Cambium ships spaced
headers, so `LastOrSurname` became `lastorsurname` while `Last or Surname`
becomes `last_or_surname`. This is a new source, not schema drift on
`src_pearson__njgpa`.

The Fall 2025 NJGPA administration (207 Newark rows, `period` of Fall) already
landed through Pearson under the same academic year, so the new feed must union
alongside the Pearson models rather than replace them.

The filename convention also changed, so the partition scheme needs redesigning:
the old pattern carried a 2-digit year plus `spr` or `fbk`, and the new pattern
carries a 4-digit year, a full season word, and the district code.

## Scope

Establish the ingestion pipeline for the Cambium NJGPA summative record file,
mirroring the existing Pearson setup:

- `libraries/cambium/schema.py` Pydantic model plus generated Avro schema
- `code_locations/kippnewark/cambium/` and `code_locations/kippcamden/cambium/`
  assets built with `build_sftp_file_asset`
- Couchdrop sensor registration in both regions
- New `src/dbt/cambium/` source-system project with an external source and a
  contract-enforced `stg_cambium__njgpa`
- Package registration in `kippnewark` and `kippcamden`
- kipptaf regional sources, union staging model, and wiring into
  `int_pearson__all_assessments`, `dim_assessments`,
  `dim_assessment_administrations`, and `fct_assessment_scores_enrollment_scoped`
- Deploy workflow path filters and the automations reference

NJSLA and NJSLA Science are out of scope. Those feeds may also move to Cambium,
but we have not seen the files and cannot assume the layout matches NJGPA. The
project should be structured so that adding a feed later is additive.

## Known behavior changes to resolve

1. **`test_score_complete` is entirely NULL in both files.**
   `int_students__graduation_path_codes` filters on `testscorecomplete = 1`.
   Copying that predicate drops every Cambium NJGPA score out of
   graduation-pathway determination with no error raised. Proposed replacement:
   `test_status` of completed, `test_attemptedness_flag` of Y, and a non-null
   `test_scale_score`.
2. **`assessment_grade` changed meaning.** Pearson reported Grade 11 for both
   subjects. Cambium reports Grade 10 for ELA and Grade 11 for Math, while
   `grade_level_when_assessed` is 11 for every row. `test_grade` becomes
   `grade_level` in both dimension models and the enrollment-scoped fact, so
   copying the regexp extract splits ELA history across two grade levels.
   Proposed fix: derive from `grade_level_when_assessed`.
3. **Newly null columns** that Pearson populated: reading and writing scale
   scores, reading and writing CSEM, `test_csem_probable_range`, `student_uuid`,
   `battery_form_id`. None are consumed downstream today, but they still need
   declaring in the enforced contract.
4. **Student crosswalk resets.** 8 Newark and 6 Camden rows have no
   `local_student_identifier`. `int_pearson__all_assessments` patches these from
   the Pearson student crosswalk sheet keyed on `student_test_uuid`, and Cambium
   mints its own identifiers, so Ops needs new crosswalk rows.

## Carries over unchanged

`test_code` of ELAGP and MATGP, `subject`, `test_performance_level` of 1 and 2
with the same Graduation Ready mapping, `period` of Spring, `assessment_year`
parsing to academic year 2025, the `student_with_disabilities` domain,
`student_test_uuid` unique per row, and 5 to 6 digit numeric local student
identifiers.

## Sequencing note

A brand-new external source cannot be staged until its asset has materialized at
least once, so the files must sit at their final Couchdrop SFTP path before the
dbt side can be validated. Two open questions gate that: which folder the files
land in, and whether the filename pattern is stable across administrations
(particularly the fall-block season token).

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will ingest NJGPA state assessment results from
**Cambium TIDE**, the vendor New Jersey moved to for the Spring 2026
administration, replacing Pearson Access Next.

The new files are structurally unrelated to the Pearson ones — only **11 of 225**
column names survive slugification into the existing NJGPA schema, because
Pearson shipped camel-case headers and Cambium ships spaced ones. So this adds a
new `cambium` source project rather than extending `pearson`. Both vendors now
coexist inside academic year 2025: Pearson carries the Fall 2025 administration,
Cambium carries Spring 2026 onward.

Everything downstream picks the new data up without code changes of its own — the
new results land in the same shared shape and join the existing union, so the
state assessments dashboard, graduation pathway codes, the DeansList extract and
the rest keep working as-is.

One reporting number does change on purpose: NJ 11th-grade state proficiency for
AY2025 on the topline weekly model is currently empty and becomes NJGPA-driven.

Refs #4899

## Reviewer Notes

Five choices worth a second look, in rough order of consequence:

1. **How the test date is derived.** Cambium only fills in the unit-level start
   timestamps for ELA, so deriving the date from those alone leaves it empty on
   every Mathematics row — which would have silently dropped about half of each
   administration from reporting while every test still passed. We fall back to
   the assessment session start instead.
1. **The reported grade level is asserted as 11, not read from the file.**
   Neither Cambium field reproduces what Pearson reported. Asserted values look
   like bugs, so please sanity-check the reasoning.
1. **This changes exactly one existing model** — the graduation pathway codes —
   because Cambium leaves the score-completeness field empty where Pearson set
   it. The predicate it changes is already a no-op for Pearson.
1. **The season tokens feeding the filename pattern and the partition list are
   deliberately one shared list.** A fall token that matches the filename but
   isn't a declared partition would stall every Couchdrop asset in that region
   until someone redeploys.
1. **Two residual items I chose not to fix**, both only biting from the Fall 2026
   administration onward, and both really one issue: the guard against an
   unexpected fall spelling is advisory rather than blocking. Detail in
   For Claude below — the fix is a pattern decision that affects every source
   package, not just this one, so it should be yours to make.

The design spec records all nine decisions with the evidence behind each:
`docs/superpowers/specs/2026-08-18-cambium-njgpa-ingestion-design.md`.

### General

### Dagster

      `kippnewark` and `kippcamden` both validate successfully
      location — passing for both, plus a new unit test for the generated schema

### dbt

- [ ] Include (or update) an
      — **not applicable.** No new model is consumed directly by a dashboard or
      application; the new data reaches existing dashboards through
      `int_pearson__all_assessments`, whose consumers already carry exposures.
- [x] **Breaking change?** No columns renamed or removed in any contracted model.
      Two column *descriptions* on `int_pearson__all_assessments` were corrected
      because Cambium changed what those columns contain.
      with **`--target staging`** — **outstanding, needs warehouse credentials.**
      See "Hand-offs" below.

### Docs

      catalog: `uv run scripts/gen-automations-doc.py` — **outstanding.**
      `docs/reference/automations.md` is stale at the Newark and Camden Couchdrop
      sensor rows. It cannot be regenerated in a Codespace: the script silently
      skips code locations that fail to import, and `kipptaf` plus `kippmiami`
      both do without credentials, so running it there would drop them from the
      catalog. Needs a full environment.
      `CLAUDE.md`

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

## Hand-offs

Three steps need warehouse credentials and are the reviewer's to run. The design
spec's rollout section has the full ordering and the exact commands.

1. **`stage_external_sources --target staging`** for both regions, with the
   `cloud_storage_uri_base` override pointing at the test bucket, so dbt Cloud CI
   can build. A brand-new external source cannot be staged until its asset has
   materialized at least once — Avro autodetect needs a file.
1. **`dbt build --select stg_cambium__njgpa --target staging`** in **both**
   district projects, plus seeding `zz_stg_kipptaf`. Staging the externals is not
   sufficient on its own: the kipptaf source points at a district *model*, and
   `dbt clone` cannot substitute because there is no prod relation to clone for a
   brand-new model. Without this, CI fails with `There were no columns found to
   union`.
1. **Student crosswalk rows** for the 9 students with no local identifier
   (3 Newark, 6 Camden), once their Cambium identifiers are visible in the
   warehouse. Note Cambium mints its own identifiers, so existing crosswalk rows
   keyed on Pearson ones will not match.

I can materialize both assets in the branch deployment myself once this PR is
open and the deployment builds — that is not blocked on credentials.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Design spec at
`docs/superpowers/specs/2026-08-18-cambium-njgpa-ingestion-design.md`, plan at
`docs/superpowers/plans/2026-08-18-cambium-njgpa-ingestion.md`. Nine numbered
decisions in the spec, each with alternatives and measured evidence.

### Full detail on the Reviewer Notes flags

**1. `test_date` coalesce.** Measured on both real files, post-filter: MATGP rows
carry **zero** unit-level online start timestamps (0 of 282 Newark, 0 of 124
Camden); ELAGP rows carry them on all 407. Unit-only derivation therefore nulls
`test_date` on 406 of 813 rows. Because
`int_assessments__resolved_section_enrollments` filters
`where test_date is not null` and `fct_assessment_scores_enrollment_scoped`
INNER JOINs it, every Cambium Mathematics score would have vanished from the
fact, Cube, and every fact-backed dashboard with all tests green and the row
counts matching the spec's own verification targets.
`safe_cast(... as timestamp)` returns NULL on the `MMDDYYYYHHMM` session field;
`safe.parse_datetime('%m%d%Y%H%M', ...)` parses it, and it is populated on 813 of
813 rows. Unit start wins in the coalesce so ELA behavior is unchanged. Measured:
where both sources exist, all 407 agree on the calendar date — so the ordering
has no observable effect on today's data and is kept for Pearson consistency and
precision.

**2. `test_grade`.** `case test_code when 'ELAGP' then 11 when 'MATGP' then 11
end`, keyed rather than a bare literal so an unrecognized code yields NULL rather
than a confident 11. Evidence: Pearson's `assessmentgrade` is `Grade 11` on all
4,130 post-filter rows across both regions and every administration, while the
student's grade is 12 for fall retakers. Cambium's `assessment_grade` is the test
*design* level — `Grade 10` for ELA, `Grade 11` for Math — and
`grade_level_when_assessed` is 11 or 12 by cohort. Neither reproduces Pearson, so
the value is asserted. It also prevents nondeterminism: `dim_assessments` dedups
with `order_by title`, and `title` is the constant `NJGPA`, so two candidate
grade levels for the ELAGP row would let `grade_level_tested` flip between
builds. Note `grade_level` is NOT in that dedup partition, so a row-count check
cannot detect the failure — the verification asserts `grade_level_tested = 11`
instead.

**3. `testscorecomplete`.** Cambium sends it entirely NULL; Pearson set it to 1
on all 4,130 post-filter rows, one distinct value, meaning the
`testscorecomplete = 1` predicate in `int_students__graduation_path_codes` was
already a no-op. Four options were weighed (spec D2). Chosen: coalesce at the
single consumer, so the shared column stays vendor-faithful and no cleanup is
deferred. Written as
`(n.testscorecomplete is null or n.testscorecomplete = 1)` rather than
`coalesce(...) = 1`, per the repo rule against functions applied to table
columns in `WHERE`. Grepped `src/dbt`, `src/cube` and `src/teamster`: that is the
only predicate consumer network-wide.

**4. The shared season list.** Verified in the installed Dagster source and
reproduced directly: an unrecognized partition key raises inside
`resolve_run_requests`, which processes every run request for a tick in one pass,
so the whole tick fails; and on `TickStatus.FAILURE` the daemon skips the cursor
write, so the offending file is re-listed forever. That stalls all six of the
region's Couchdrop assets until a redeploy. Deriving the filename alternation and
the partition values from one `ADMINISTRATIONS` list makes an unknown token fail
to *match* instead, which skips the file and leaves the sensor working. Sorted
longest-first so `FallBlock` is not shadowed by `Fall`. Relevant precedent: the
Pearson fall filename suffix was `_FallBlock` in 2024 and `_FALL` in 2025 — the
vendor changed spelling and case between consecutive years.

**5. The two residual items.** Both verified, neither fixed:

- `rpt_deanslist__state_test_scores` reads the raw `period` as `test_round`,
  bypassing the `admin` normalization. A Cambium `period` of `FALL` would reach
  that extract unnormalized.
- All four cambium tests resolve to `severity: warn` when built from either
  district, despite the package declaring `error`, because both district projects
  set an unscoped `data_tests: +severity: warn` which overrides installed-package
  test config. So the non-empty safety net below only warns.

Together: the `accepted_values` guard on the raw `period` is advisory rather than
blocking. Not fixed because the severity behavior is pre-existing and repo-wide,
affecting every source package equally, and the targeted fix — a scoped
`data_tests: cambium: +severity: error` in each district — is a pattern decision
worth making deliberately across all of them rather than unilaterally here.

### Other verified facts worth having

- **Removed-column drift is undetectable by the Avro check.** It computes
  `extras = record_fields - schema_fields` and warns only on extras, so a
  *removed* column produces no signal; every generated field defaults to null, so
  fastavro writes null silently. If Cambium dropped `Summative Flag`, the staging
  filter would return zero rows and `unique` / `not_null` would pass vacuously.
  Hence the new singular test `stg_cambium__njgpa__not_empty` — though see
  residual item 2 above for why it currently warns rather than fails.
- **`dim_assessment_administrations` needed a new branch; `dim_assessments` did
  not.** The administration key hashes academic year and administration period,
  and AY2025 with period Spring exists only in Cambium (Pearson's AY2025 is
  Fall-only), so without it every Cambium score orphans on the FK. The assessment
  key excludes grade level and year, so Cambium's rows hash identically to
  Pearson's. The branch was added anyway so the dimension does not silently
  depend on Pearson history persisting.
- **No fan-out risk in the topline weekly change.** Every AY2025 Pearson NJGPA
  row is `gradelevelwhenassessed = 12` while every Cambium row is grade 11, so no
  student appears in both and no key gains a second row. Structural note for
  later: from Fall 2026 a single academic year will hold both a Cambium Fall and
  a Cambium Spring administration, and that model's unfiltered join would
  double-count a student who sat both. Pre-existing shape — AY2024 Pearson
  already holds Spring and FallBlock — and only out of range today because of the
  model's academic-year window.
- **Union types cannot clash.** `dbt_utils.union_relations` casts every relation
  to the type declared by the first relation carrying the column, and Cambium is
  appended last, so it cannot change a type. `is_bl_fb` is the only include-list
  column it does not produce, matching `stg_pearson__njgpa`.
- **`_dbt_source_project` is the REGION**, not the vendor — the macro extracts
  `kipp\w+`. Counterintuitive given the name; Cambium adds no new values there.
- **Sensor cold start.** A newly added asset has no cursor entry, so the sensor's
  `min(cursor.values())` floor comes from the six existing assets and the first
  tick's Drive listing is mtime-bounded rather than unbounded. It self-heals on
  the next tick, and the explicit post-merge prod launch moots it. Pre-existing
  library behavior, not introduced here.
- **`int_pearson__all_assessments` now carries Cambium data under a Pearson
  name.** Deliberately deferred: the rename touches roughly 15 files across
  dashboards and extracts, and the Pearson-named student crosswalk sheet belongs
  in the same follow-up. Recorded in the spec's deferred work.

### Process notes

Ten tasks, each implemented by a fresh subagent and reviewed for spec compliance
and quality before the next started. Two adversarial reviews ran before any code
was written and caught three blockers in the design — including the `test_date`
one above, which the original spec had wrong. A final whole-branch review on the
most capable model returned two Important and six Minor findings, all fixed in
one wave and re-verified.

Corrections made to my own work during execution, for transparency: the spec
originally claimed the two test-date sources "agree on 363 of 407 rows and differ
on 44" — a statistic never measured, and false (all 407 agree). It also described
the ELA-only unit timestamps as "units 3 and 4 are null," which missed that
Mathematics has none at all. Both corrected.

Verification run on the final tree: pytest passing, `dbt parse` on all four
projects, `dagster definitions validate` on both code locations, and
`trunk check --force` clean over all 47 changed files.

</details>


anthonygwalters:
Verdict on each finding. Two accepted and fixed in `c8e90c3e8`; three declined with reasoning; one partially accepted with a correction to the proposed mechanism.

### 1. Missing `contract: enforced: true` on the kipptaf-level model — partially accepted, deferred with reason

The gap is real. Two corrections to the recommendation, though.

**The mechanism is wrong for this project.** `kipptaf/CLAUDE.md` lines 128-133 are explicit: "Contract-enforcement here is per-model, NOT directory-wide: the `powerschool:` block in `dbt_project.yml` sets only `+schema:` (no `staging: +contract`), so powerschool `staging/` union views are contract-enforced only where a model sets it in its own `properties.yml`." So adding `+contract: enforced: true` under `cambium: staging:` in `dbt_project.yml` would break the documented kipptaf pattern. The right place is the model's own `properties.yml`.

**I am declining to add it in this PR**, because doing it blind is the specific failure this repo warns about. `src/dbt/CLAUDE.md` says to regenerate contract `data_type` values "verbatim from `INFORMATION_SCHEMA.COLUMNS` of the staged table; don't hand-transcribe" — and I cannot, because the upstream relations do not exist yet. Worse, `dbt_utils.union_relations` casts every relation to the type declared by the *first* relation carrying each column, so the aligned model's output types are decided by the macro at compile time against real relations. Hand-declaring 34 types on inference is exactly how you get a contract that parses clean and fails at build (`numeric` vs `float64`), and the person who would hit it is whoever runs the staging builds, with no obvious cause.

Recommended sequencing instead: after the district `--target staging` builds land and this model has built once, pull the 34 types from `INFORMATION_SCHEMA.COLUMNS` and enable the contract in `properties.yml`. Happy to do that as an immediate follow-up on this branch if you'd rather it ship together — it just needs to come after the first successful build, not before.

On the "inconsistent spot between two patterns" point: the exception in `kipptaf/CLAUDE.md` is scoped to "pure `union_relations()` views ... functionally intermediates." This model is not pure — it carries every Cambium-to-shared translation (renames, the asserted grade level, case-insensitive season normalization, the proficiency mapping). So I think it correctly sits outside that exception, which is why it has `materialized: table` and real tests. The `period` / `administration_period` tests in particular cannot move upstream to the source project, because `administration_period` is derived here.

### 2. `CASE` placed before `IF`s (ST06 ordering) — declined

`trunk check --force --no-fix` on `src/dbt/kipptaf/models/cambium/staging/stg_cambium__njgpa.sql` reports **no issues**, and sqlfluff is the enforcer this finding names. Trunk Check is also green on this PR. The documented category ordering appears coarser in practice than the doc reads — the same question came up in internal review and resolved the same way.

I'd rather not reorder purely to match prose that the linter does not enforce, since it would move a comment block explaining the grade-level decision away from the expression it explains. Happy to reorder if you read the convention as binding regardless of sqlfluff.

### 3. `test_grade` has no test guarding the NULL-on-unrecognized-code case — accepted, fixed

This was the best finding of the six. The column was designed to yield NULL rather than a confident 11, but "fails loudly" needs a test, and there wasn't one — so a renamed Cambium `test_code` would have flowed a NULL grade level into both dimension models. Fixed in `c8e90c3e8`: added `not_null` plus `accepted_values: [11]`, both at `severity: error`, and moved `test_grade` up with the other tested columns per the sort-tested-columns-to-the-top convention.

One caveat on effectiveness, already disclosed in the PR body: both district projects carry an unscoped `data_tests: +severity: warn` that overrides installed-package test config, so these resolve to `warn` in the only way this project ever actually builds. The test is right; making it actually block is the repo-wide severity decision flagged in Reviewer Note 5.

### 4. `admin` re-derives what `administration_period` already computed — declined, deferred

The duplication is real and the observation is a good one. Declining here because the substitution changes the `admin` derivation for the three *other* assessment types in that union (PARCC, NJSLA, NJSLA Science), not just NJGPA, and `admin` feeds a Tableau dashboard. The two expressions are equivalent for every value Pearson has ever sent, but nothing in the test suite asserts that equivalence, and this PR is already large.

Logged onto the deferred rename of `int_pearson__all_assessments`, which restructures that model anyway — that is the right place to collapse it.

### 5. Silent slug-collision risk in the schema generator — accepted, fixed

Correct, and it was on our own deferred list. Fixed in `c8e90c3e8`: the generator now raises with the colliding slug and the distinct headers that produced it, rather than silently dropping a field. Verified the generator still produces byte-identical output against the real 225-column file after the change.

### 6. Possible ST06 issue with `parse_datetime` between two cast groups — declined

Same evidence as finding 2: `trunk check --force --no-fix` on `src/dbt/cambium/models/staging/stg_cambium__njgpa.sql` reports no issues, and Trunk Check passes on this PR. This finding was explicitly flagged low-confidence pending sqlfluff, and sqlfluff is clean.

---

### On the scope note

The dbt Cloud check is red, and that is expected rather than a regression. The failure is:

```text
Not found: Dataset teamster-332318:zz_stg_kippcamden_cambium was not found in location US
```

The kipptaf source points at a district *model*, and the `zz_stg_<region>_cambium` datasets do not exist until someone runs `dbt build --select stg_cambium__njgpa --target staging` in both district projects. That is documented in the PR body's hand-offs section and in the design spec's rollout sequence. `dbt clone` cannot substitute, because there is no prod relation to clone for a brand-new model.

Every other check is green: Trunk, CodeQL, and all Dagster Cloud deploy jobs for both code locations.

cbini:
## Review + five commits pushed

Charlie asked me to review this against repo patterns. Rather than leave a list, I pushed the fixes. Here is what changed and why, including the two things I flagged and then decided you had right.

The design spec made this reviewable — the numbered decisions with evidence are what let me check each one against the code instead of guessing at intent.

### Merge

`main` had moved. Three conflicts, all resolved toward `main`:

- `dim_assessments.sql` / `dim_assessment_administrations.sql` — `main` reworded a repeated comment in 13 and 14 places. Kept your new Cambium CTEs, took `main`'s wording.
- `src/dbt/CLAUDE.md` — your only change was adding `cambium` to an enumeration that `main` has since replaced with `ls src/dbt`. Took `main`'s version, so that maintenance is no longer needed.

### Fixed

| # | Finding | Fix |
|---|---|---|
| 1 | `scripts/gen-cambium-schema.py` duplicates `scripts/init_sftp_integration.py codegen`, which `docs/guides/sftp-integration.md` documents as the path for a new SFTP drop | Deleted the script and its `scripts/CLAUDE.md` row. The committed schema is unchanged. |
| 2 | The `not_empty` guard silently resolved to `warn` | Moved to `src/dbt/kipptaf/tests/`. Your Reviewer Note 5 called this a repo-wide pattern decision — it is, but kipptaf root-project tests already hold `error`, so this one test gets teeth without touching the pattern. Confirmed in the parsed manifest. |
| 3 | `accepted_values` on raw `period` omitted `FALL`, which `ADMINISTRATIONS` includes | Value list now mirrors `ADMINISTRATIONS` exactly. |
| 4 | `administration_year` was an unbounded four-digit match against a bounded partition list | Bounded via a shared `ADMINISTRATION_YEARS` list — the same technique, and the same reasoning, as your season list. |
| 6 | `rpt_deanslist__state_test_scores` read raw `period` | Reads `administration_period`, which all five union relations populate. |

Added `tests/libraries/test_cambium_assets.py`: every filename the regex matches must yield a declared partition, and an undeclared token must fail to match. I reverted the regex to confirm it actually fails on the old one.

### Two findings I withdrew

**Trimming the speculative season tokens.** I argued `["Spring", "Fall", "FALL", "FallBlock"]` should be only what you have observed. Then I read `libraries/couchdrop/sensors.py`: `context.log.info` sits inside the match loop and the cursor only advances on a match, so an unmatched file is logged nowhere and re-listed silently forever. Trimming would turn a possible hit on the Fall 2026 file into a guaranteed silent miss. The extra tokens cost only unused partition keys. Your call was better than mine.

**A missing package README.** All ten source-package READMEs are unmodified `dbt init` boilerplate. `src/dbt/cambium/CLAUDE.md` already does the real job.

### A bug only a real build could find

`accepted_values` on `test_grade` could never have passed. The column is `int64` and dbt renders the values as string literals unless told otherwise:

```
No matching signature for operator IN for argument types INT64 and {STRING}
```

Fixed with `quote: false`, matching five existing usages. It had never fired because the PR had never had a green dbt Cloud run.

### Comment sweep

`src/dbt/CLAUDE.md` keeps inline SQL comments to what a reader of that exact line cannot see, with rationale in the properties.yml `description`, and says outright that the existing multi-paragraph SQL comments are not a precedent to extend. Removed 47 comment lines.

Most of them were already in your yml near-verbatim — `test_score_complete`, `testscorecomplete`, `test_grade`, `administration_period` — so the deletions lose nothing. The `test_date` coalesce went from 11 lines to 2 because the yml description already carries the argument. The predicate comment in `int_students__graduation_path_codes` stayed: why a filter exists is an explicit keeper.

Verified comment-only the way the repo prescribes: strip comments, collapse whitespace, compare tokens. All five files token-identical.

Python is untouched and conformant — `src/teamster/CLAUDE.md` defers to the Google Python Style Guide and permits multi-line docstrings.

### Hand-offs, done

All three warehouse steps are complete:

1. Materialized both assets in the branch deployment, partition `Spring|2026`. Both SUCCESS.
2. `stage_external_sources` for both regions — 228 columns autodetected each.
3. District and kipptaf staging models built.

Results match your spec exactly: 813 rows (564 Newark, 249 Camden), `test_grade` 11 only, `administration_period` Spring only, `academic_year` 2025, `testscorecomplete` null on all 813.

The number worth calling out: **zero null `test_date`**. Reviewer Note 1 was the highest-stakes call in this PR, and it holds against real data — without the coalesce all 406 Mathematics rows would have vanished from the fact with every test still green.

dbt Cloud is green — the first green run this PR has had.

### One new warning, and it needs Ops, not code

The green run carries 8 test warnings. I checked each against prod; 7 are pre-existing and unrelated (the two gradebook uniqueness tests, the two kfwd survey tests, and 5 `Practice` FK orphans from the CARAT pipeline with periods `Jul 23` / `Aug 26`). `test_incorrect_student_number_pearson` reports 17, of which 9 are exactly the Cambium crosswalk gap you documented and 8 are pre-existing 2017/2018 rows.

The one that is new is `accepted_values` on `int_students__graduation_path_codes.grad_eligibility`: 52 cohort-2027 students land in the fallback string `New category. Need new logic.`, which prod does not produce.

It is not a defect in this branch — it is the branch working. `grad_eligibility` is a lookup into `stg_google_sheets__student_graduation_path_combos`, joined on 8 columns including `attempted_njgpa_ela` and `attempted_njgpa_math`. Admitting Cambium scores moves cohort 2027 from 38 students with both NJGPA attempts to 6,856, and surfaces 106 single-subject attempters (70 math-only, 36 ela-only) where prod had none. The sheet has no row for that shape, so the coalesce fallback fires.

The exact combination Ops needs to add:

```
grade_level=12  has_fafsa=false  njgpa_season_11th=false  fafsa_season_12th=false
attempted_njgpa_ela=false  attempted_njgpa_math=true  met_ela=false  met_math=false
```

Worth deciding before merge, since it is a reporting value users will see. I have not touched the sheet.

### Left to do

1. **CODEOWNERS approval.** `mergeable_state` is `blocked` on it — the only thing left before merge.
2. **Student crosswalk rows** for the 9 students with no local identifier. Their Cambium identifiers are queryable now.
3. **One row in the graduation-path-combos sheet**, per the combination above.
4. **Post-merge:** launch both `cambium/njgpa` assets in prod immediately. External sources bypass the deps gate, so the first tick otherwise stages against an empty prod prefix and fails.

Two things I found and deliberately left alone, both repo-wide and not yours to carry: the unscoped `data_tests: +severity: warn` in both district projects that downgrades every installed-package test, and CRLF line endings in 9 of 400 sampled `.sql` files with no `.gitattributes`.

cbini (dismissed):

Issue #4803, opened 2026-08-10
Miami attendance facts orphaned from dim_student_enrollments after the Focus cutover


cbini:
## What

`fct_student_attendance_daily` carries 153,739 rows whose `student_enrollment_key` no longer resolves to `dim_student_enrollments`. The `relationships` test warns at 155,863 total against a prod baseline of 2,286.

| Calendar year | Orphaned rows |
| --- | --- |
| 2016-2024 | ~2,124 (pre-existing) |
| 2025 | 76,237 |
| 2026 | 77,502 |

961 distinct enrollment keys, essentially all Miami AY2025.

## Why

#4775 makes Focus the sole source of Miami enrollment. Focus dates a returning student's stint to the real first day of school where PowerSchool used a July 1 administrative rollover, so 1,421 of Miami's 8,776 historical stints carry a different `entrydate` -- 973 of them in AY2025 alone. `entrydate` feeds `student_enrollment_key`.

`fct_student_attendance_daily` builds from `int_powerschool__ps_adaadm_daily_ctod`, so Miami's historical attendance still comes from the frozen PowerSchool archive and still keys off the archive's dates. The dimension moved; the fact did not.

The attendance rows themselves are not wrong -- they just no longer join to enrollment, so anything that traverses the fact through `dim_student_enrollments` loses a year of Miami attendance.

## Options

1. Drop Miami from the PowerSchool attendance source, consistent with the rest of the cutover. Miami loses historical attendance from the marts until Focus attendance is modeled.
2. Re-key the Miami attendance leg to Focus enrollment dates so the existing rows resolve.
3. Model Miami attendance from Focus directly -- the real fix, and the natural home for it is the next phase of #4731.

## Interim state

Accepted and tracked rather than fixed in #4775. The test is `severity: warn`, so it does not block CI, and Miami's first day of school is 2026-08-12.

Refs #4731
Refs #4775

cbini:
The `Refs #4731` pointer is dead. #4731 is closed, so option 3 — "model Miami attendance from Focus directly, and the natural home for it is the next phase of #4731" — currently has no home. Miami Focus attendance modeling is not tracked anywhere right now.

That matters for picking between the three options here, because option 3 is described as the real fix and the other two are stopgaps. Whoever picks this up should either open a Focus attendance modeling issue under #3584 and make option 3 depend on it, or commit to option 1 or 2 knowing the real fix is unscheduled.

[...]

Issue #3584, opened 2026-04-03
feat: integrate Focus SIS for KIPP Miami


cbini:
Integrate Focus SIS (PostgreSQL) into the data platform for KIPP Miami, replacing PowerSchool as the SIS.

## Scope

- **Extraction**: dlt `sql_table` → BigQuery (daily full-replace)
- **dbt**: New `focus` source-system project with staging + intermediate models
- **Dagster**: kippmiami code location wiring, schedule, 1Password/k8s config
- **kipptaf**: Regional sources to expose Focus data to the network data mart

## Design spec

[`docs/superpowers/specs/2026-04-03-focus-sis-integration-design.md`](https://github.com/TEAMSchools/teamster/blob/cbini/claude-/feat/focus-sis-integration/docs/superpowers/specs/2026-04-03-focus-sis-integration-design.md)

cbini:
Only Phase A completed

PR #4919, opened 2026-08-18
feat(dbt): Focus attendance and gradebook intermediate models

### LINKED ISSUE #4865: feat(dbt): remaining Focus intermediate models — attendance day, attendance period, gradebook grades
Split out of #4220, which is a register of **staging** coverage against the configured Focus tables. Intermediate models were riding along in a trailing section there and kept getting read as part of that wait-and-see tracker. They are their own work.

All three are built in PR #4919, along with ingestion this issue did not anticipate.

## Outstanding

| Model | Source table | State |
| --- | --- | --- |
| `int_focus__attendance_day` | `attendance_day` | built in #4919 |
| `int_focus__attendance_period` | `attendance_period` | built in #4919 |
| `int_focus__gradebook_grades` | `gradebook_grades` | built in #4919 |

## These are not the decode pattern

The established shape for `int_focus__{entity}` is staging columns plus decoded custom-field labels, left-joined from an `int_focus__{entity}__pivot` — see `int_focus__courses` and the pattern set by #4596.

**That pattern does not apply here.** None of the three source tables has a single `custom_*` column, so there are no labels to decode and no `__pivot` models were created. This part of the original issue held up.

## Correction: the gradebook model needed new ingestion

This issue said the three models require no new source tables, and that staging coverage under #4220 was the only gate. That is true for the two attendance models. It is wrong for gradebook.

`gradebook_grades` carries only `student_id` and `assignment_id`. Nothing in the tables we were pulling connects an assignment to a course period, so a grade had no path to a course. The Focus ERD names the link only as a glob, `gradebook_*_join_course_periods`. Probing the Focus Postgres resolved it to two tables that were never configured for extraction:

- `gradebook_assignments_join_course_periods` — assignment to course period, plus assigned, due and publish dates
- `gradebook_assignment_types_join_course_periods` — category to course period per marking period, carrying `final_grade_percent`, the category weight

#4220 tracked staging coverage of the tables already configured, so a table nobody had configured could not surface there. Both are now in `focus.yaml` and have staging models.

## Correction: the two attendance code columns resolve differently

The issue treated attendance codes as one problem. They are two.

`attendance_period.attendance_code` and `.attendance_teacher_code` hold `attendance_codes.id` and join directly, with zero orphans.

`attendance_day.daily_code` holds a `short_name` — values `U`, `AE`, `AD`, or null — not an id. Short names repeat across schools and years, and `attendance_day` has no `school_id`, so school is resolved from the marking period before the code join.

Also worth recording: `state_value` is already the present/absent classification. 6,804 of 7,699 day rows are `daily_code is null` with `state_value = 1`. The code join supplies the reason for an absence, not the classification.

## Correction: no attendance_calendar join

The issue asked for one, to tell a non-attendance day from a missing record. It was not built, for two measured reasons.

`attendance_calendar.minutes` is `999` on every matched 2026 row — one sentinel value, so there is nothing to compare `minutes_present` against. `bell_schedule_id` is populated on 7 percent of rows.

And a missing record is not visible at this grain. A row that exists is a day that happened. Finding absent rows needs a scaffold of enrolled students crossed with calendar days, which is a different grain and belongs to the consuming mart. The design document records how to scope the join correctly if a future consumer wants it: filter `attendance_calendars.default_calendar` to `Y`, the idiom `int_focus__school_year_first_day` already uses.

## Consumer waiting on attendance

#4803 records 155,863 Miami attendance rows orphaned from `dim_student_enrollments` after the Focus cutover, and names modelling Miami attendance from Focus directly as the real fix among its three options. That option routes through `int_focus__attendance_day`, which now exists.

The boundary from the original issue still holds: Focus carries AY2026-27 forward only. These models give Miami a correct current year. They do not re-key the orphaned AY2025 rows, which came from the frozen PowerSchool archive, so #4803 still needs a separate answer for its historical leg.

## Shipped earlier, for context

- `int_focus__student_enrollment` — shipped with #4213. Reshaped to staging-plus-labels in #4600, which also carries the `enroll_status` rollover-code fix absorbed from #4794.
- `int_focus__schedule` — shipped in PR #4725. Focus stores `0`, not null, for a full-year schedule (paired with `mp = 'FY'`), so the sentinel is normalized before the marking-period join.
- `int_focus__report_card_grades` — shipped in PR #4725, narrower than the design spec intended. Every Miami row is imported course history whose `report_card_grade_id` and `grade_scale_id` are empty, so the planned grade and grade-scale joins have no keys. It joins `marking_periods` and decodes the Course Flag fields instead. Revisit when Focus posts live report-card grades.

The `int_focus__report_card_grades` departure supersedes the Intermediate models section of the [design spec](https://github.com/TEAMSchools/teamster/blob/main/docs/superpowers/specs/2026-04-03-focus-sis-integration-design.md) for report cards. The three models in this issue are designed in [their own spec](https://github.com/TEAMSchools/teamster/blob/main/docs/superpowers/specs/2026-08-18-focus-attendance-gradebook-intermediates-design.md).

Refs #3584
Refs #4220
Refs #4803

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will add the three remaining Focus intermediate
models tracked in #4865, and the dlt ingestion the gradebook one needs.

`int_focus__attendance_day` and `int_focus__attendance_period` resolve Focus
attendance codes to readable titles and excused/tardy flags. These are what
#4803 needs to model Miami attendance from Focus directly, which that issue
names as the real fix for its 155,863 orphaned rows.

`int_focus__gradebook_grades` puts each gradebook score together with its
assignment, its category, that category's weight toward the final grade, and
the course period the score belongs to.

Getting a grade to its course period required ingesting two Focus tables that
were never configured for extraction. They exist in Focus, but nothing was
pulling them, so a gradebook grade had no path to a course. #4220 tracked
staging coverage of the tables we already pull, so it never surfaced this gap.

## Reviewer Notes

**The two attendance models resolve their codes differently, on purpose.**
`attendance_period` stores code ids. `attendance_day` stores a short name
(`U`, `AE`, `AD`), and short names repeat across schools and years, so that
model resolves the school from the marking period before joining. The two
joins look inconsistent side by side; they are not.

**A grade's course period is genuinely ambiguous, and the model resolves it
through the student's schedule.** Teachers assign the same work to several
sections, so an assignment maps to many course periods while a grade names
only the student and the assignment. 11 of the 12 assignments carrying grades
today are in that position. The model keeps the section the student is
actually scheduled into.

**No `attendance_calendar` join, though #4865 asks for one.** Its `minutes`
column is the same sentinel value on every current row, so there is nothing
to measure against, and spotting a missing attendance record needs a different
grain than these models have. The design document records how to scope the
join correctly if a future consumer wants it.

### General

### Dagster

      code location

### dbt

- [x] Include (or update) an
      applicable; these are internal models with no dashboard consumer yet
- [x] **Breaking change?** No. Every model here is new; nothing is renamed or
      removed, so no downstream exposure can break.
      with **`--target staging`** — not applicable; the two new Focus sources
      are BigQuery-native dlt landing tables, not external sources

### Docs

      catalog — not applicable; no schedule or sensor changed
      `CLAUDE.md` — not applicable; this extends the existing Focus integration

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Design document:
`docs/superpowers/specs/2026-08-18-focus-attendance-gradebook-intermediates-design.md`.
Implementation plan:
`docs/superpowers/plans/2026-08-18-focus-attendance-gradebook-intermediates.md`.

**This PR lands in two pushes.** The first carries the dlt config, the source
declarations, and both attendance intermediates. The two new gradebook staging
models and `int_focus__gradebook_grades` follow, after the branch deployment
loads the link tables — dbt CI cannot validate models whose source tables do
not exist yet. Toggle draft state before final review so `claude-review` sees
the complete change; it fires on `opened` and `ready_for_review`, not
`synchronize`.

**Measurements behind the design**, all taken 2026-08-18 against live data,
either in `dagster_kippmiami_dlt_focus` or by probing Focus Postgres from a
`kippmiami` code-server pod:

- `attendance_period.attendance_code` and `.attendance_teacher_code` resolve to
  `attendance_codes.id` with 0 orphans of 15,171 rows.
- `attendance_day.daily_code` resolves to `attendance_codes.short_name` scoped
  by year and school, 0 unresolved of 895 coded rows. `daily_pres_code` is null
  on every row. `state_value` is already the present/absent classification —
  6,804 of 7,699 rows are `daily_code is null` with `state_value = 1` — so the
  code join supplies the reason for an absence, not the classification.
- The assignment link has 428,009 rows over 355,583 distinct assignments;
  28,074 assignments map to more than one course period, including 11 of the 12
  that carry grades. Filtering by the student's schedule collapses every grade
  to exactly one course period, with 0 grades having a link row but no schedule
  match. 30 grades have no link row at all, which is why every join in that
  model is a LEFT join.
- `gradebook_assignment_types_join_course_periods` repeats an
  `(assignment_type_id, course_period_id)` pair 23,318 times — weights change
  per marking period — so the weight join keys on the triple including
  `marking_period_id`. 87,913 of its 89,206 rows carry `final_grade_percent`.
- `schedule` repeats a student and course-period pair on 299 combinations, so
  the model collapses it with `group by` and uses it as a lookup rather than
  joining it as a table. The earlier check counted distinct course periods and
  could not have caught this.
- `attendance_calendar.minutes` is `999` on all 7,697 matched 2026 rows;
  `bell_schedule_id` is populated on 544. Scoping to
  `attendance_calendars.default_calendar = 'Y'` reduces 1,554 duplicate
  `(syear, school_id, school_date)` keys to 177, all in `syear` 2016 — so the
  join is safe for current data, just not informative.

**Primary-key tests do not gate anything here.** All five models declare
`unique` and `not_null` at `severity: error`, matching the surrounding staging
models — but those resolve to `warn`. The root project sets
`data_tests: +severity: warn` in `src/dbt/kippmiami/dbt_project.yml`, and
root-project config outranks config declared inside an installed package, which
every `focus` model is. The manifest confirms it:
`unique_stg_focus__attendance_codes_id` resolves to `warn` despite its YAML
declaring `error`. This is pre-existing and repo-wide, not introduced here.

Consequence: a join that multiplied rows would warn, and the build would still
go green. Grain was therefore confirmed by direct count rather than inferred —
`int_focus__attendance_day` 7,701 rows and 7,701 distinct PK,
`int_focus__attendance_period` 15,633 / 15,633,
`int_focus__gradebook_grades` 697 / 697 against 697 staging rows,
`stg_focus__gradebook_assignments_join_course_periods` 428,025 / 428,025.
Whether the network wants that override lifted for primary-key tests is a
policy question this PR does not settle.

**One latent ambiguity is documented rather than fixed.** A grade resolves to a
single course period only while the student sits in exactly one of the
assignment's sections. Three student-and-assignment pairs currently match two
sections each; none carries a grade, so nothing duplicates today. Filtering the
match by assignment date against the schedule window was tested and rejected —
it drops a real grade belonging to a student added to a section on 2026-08-18
and back-graded on work assigned 2026-08-12, which is ordinary late-enrollment
behaviour. Trading a case that happens for one that has not seemed wrong. The
model description records this.

**Columns removed for carrying no signal.** `state_attendance_type` is null on
all 835 attendance codes, so `attendance_code_state_attendance_type` and
`daily_code_state_attendance_type` were dropped rather than mirrored onto the
teacher code as first suggested. `chronic_absenteeism` and `truancy` hold `Y`
on every code including `Present` — Focus config defaults, not classifications —
and `tardy` is null at day grain because `T` is period-level only, so all three
were dropped from `int_focus__attendance_day`. `daily_code_excused` stays; it
genuinely discriminates.

**Two traps worth knowing for anyone building on this locally.** The `focus`
project is a dbt package with `focus_schema: null` and no developer profile, so
it cannot be built standalone — build via `kippmiami`. And `--defer` falls back
to prod only for objects ABSENT from the dev schema, so a stale dev copy is
used silently and the model still builds green with wrong values. A 2026-06-22
copy of `stg_focus__attendance_codes` made 8 rows look unresolvable during
development. Both are recorded in the plan's Global Constraints.

Closes nothing on its own — #4865 also needs its body corrected, since it
states these models require no new ingestion.

Refs #4865
Refs #4803
Refs #3584

</details>


cbini:
## Review response

Verified each finding against the warehouse before acting. Two led to changes,
three are declined with reasons. Fixed in `7dd77f0ed`.

### 1. Teacher code missing `type` and `state_attendance_type` — split verdict

**`type`: valid, fixed.** Added `attendance_teacher_code_type`. It is genuinely
informative and not constant — teacher-code rows resolve to 15,263 `teacher`
and 1 `official`; official-code rows to 15,583 `teacher` and 50 `official`. The
asymmetry was an oversight, correctly caught.

**`state_attendance_type`: valid problem, opposite fix.**
`stg_focus__attendance_codes.state_attendance_type` is NULL on all 835 rows, so
adding it for the teacher code would have created a second column that can never
hold a value. The existing `attendance_code_state_attendance_type` and
`daily_code_state_attendance_type` had the same defect. Both are now removed
rather than mirrored.

### 2. `chronic_absenteeism` / `truancy` absent from the period model — declined

Both are populated on all 70 of the 2026 codes, but the value is `Y` on every
one of them, including `Present`. They are default-filled, not meaningful, and
adding them would carry no signal — the "population is not informativeness"
trap documented in `src/dbt/focus/CLAUDE.md`.

Their presence on `int_focus__attendance_day` is the inconsistency worth
noting, not their absence here. Left alone for now since removing them is a
change to a column set that was reviewed and approved on its own terms.

### 3. Fan-out risk on the `(syear, school_id, short_name)` join — declined, already measured

This was checked during design; the result is recorded in the implementation
plan's "Notes for the implementer" section rather than the design document,
which is why it read as unstated:

> `attendance_codes` has one duplicate `(syear, school_id, short_name)` in
> syear 2010.

That single duplicate is in `syear` 2010. `attendance_day` carries AY2026-27
only, so no row can reach it. And as the finding notes, the `unique` test at
`severity: error` on the PK is the standing guard if that ever changes.

### 4. PII tagging inconsistent on older sibling models — confirmed, out of scope

Correct on the facts: the two new models tag `student_id`, while
`int_focus__schedule` and `int_focus__report_card_grades` do not. Reconciling
those is a change to models this PR does not otherwise touch, so it belongs in
its own change rather than riding along here. Flagged to the author for a
follow-up issue.

### 5. Colon-space inside a folded YAML scalar — declined, verified fine

Block scalars do not treat `:` specially, as the finding itself reasoned. Now
confirmed rather than assumed: `dbt parse` resolves, the model builds, and the
dbt Cloud CI check on this PR passed with zero warnings.

### Note on scope

This PR lands in two pushes. The review above covers the first — the dlt config
and both attendance intermediates. The two gradebook staging models and
`int_focus__gradebook_grades` follow once the branch deployment has loaded their
source tables, which has now happened. Draft state will be toggled before final
review so this check re-fires on the complete change.

cbini:
## Review response — round 2

Five findings fixed in `8f8ccae54`, one answered with data. Grain re-checked
after the changes: 697 rows, 697 distinct PK, 30 unresolved — unchanged.

### 1. `group by` with no aggregation should be `distinct` — fixed, and the correction is mine

Valid, and I had this backwards. `src/dbt/CLAUDE.md` line 1121 prescribes
`DISTINCT` for precisely this case — a `GROUP BY` with no aggregation that is a
pure grain projection — annotated with a specific comment. I had instructed the
implementer to avoid `select distinct`, over-applying a rule that targets
dedup-masking rather than grain projection. Both projected columns here are the
grouping key, so it is the blessed case, not the banned one.

Now matches the precedent at `int_focus__student_contacts.sql:41-45`, with the
prescribed annotation. Semantics identical; the 697/697/30 numbers confirm it.

### 2. Tautological `attendance_teacher_code_type` description — fixed

Fair. The old wording was circular and buried the reason the column exists.
Reworded to say `type` is Focus's classification of whichever code row
resolved, independent of which column held the id — which is the whole point,
since teacher-code rows resolve to `type = 'official'` once and official-code
rows resolve to `type = 'teacher'` 15,583 times.

### 3. No uniqueness test on the weight-join triple — fixed

Added `dbt_utils.unique_combination_of_columns` over
`(assignment_type_id, course_period_id, marking_period_id)` on the staging
model, matching the shape used by `stg_focus__attendance_completed`. The triple
was already verified unique — zero duplicates across 89,206 rows — but nothing
enforced it. The test ran and passed on this build, so the one-time measurement
is now a standing check.

### 4. Grain claim vs `standard_id` — verified, description is accurate

Checked rather than assumed:

```text
n_rows 697 | distinct (student_id, assignment_id) 697 | standard_id not null 0
```

Every `(student_id, assignment_id)` pair is unique and `standard_id` is null on
every row — Miami is not using standards-based grading, so the documented grain
holds today. The concern is real in principle: if standards-based grading is
adopted, the natural grain becomes `(student, assignment, standard)` and this
description would need revisiting. `id` remains the correct PK either way, so no
code change.

### 5. Missing blank line between table groups in the CTE — fixed

Correct, `src/dbt/CLAUDE.md` line 1250. The final SELECT in the same file was
already right; the CTE was the oversight.

### 6. Missing null caveat on `assignment_type_drop_lowest_grades` — fixed

Correct — same LEFT JOIN as `assignment_type_final_grade_percent`, so null under
identical conditions. Caveat added, matching the sibling's wording.

### On the suggested singular test

Noted, not added. A test flagging a graded `(student_id, assignment_id)` pair
that matches two sections would be a real signal — but a caveat worth recording
alongside it: `data_tests: +severity: warn` in the root project outranks
package-level YAML, so every test on these models resolves to `warn` regardless
of declaring `error`. That includes the PK test this and finding 3 both lean on
as the standing safety net. It is a signal, not a gate. Raised separately with
the author as a network-wide policy question rather than patched here.

cbini:
Follow-ups from this review, now tracked:

- #4922 — package data tests silently downgrade to `warn`. The `unique`/`not_null` tests both review rounds treated as the fan-out safety net do not gate anything, in any package, in any district project. Includes the compiled-severity evidence and three options with their blast radius.
- #4923 — `student_id` untagged as PII on `int_focus__schedule` and `int_focus__report_card_grades`.

Neither blocks this PR. #4922 is the one worth a decision before the next model leans on a PK test the way this design originally did.

PR #4929, opened 2026-08-19
fix(clever): pin the ENR section teacher order to the DSO

### LINKED ISSUE #4928: fix: pin the ENR section teacher order in rpt_clever__sections
## What's happening

The auto-generated `ENR` ("Enroll") sections in `rpt_clever__sections` assign
their primary teacher arbitrarily. `teacher_id` flips between the school's DSO
and the school's School Leader from run to run, with no underlying change in
PowerSchool or ADP.

Clever alerted on 2026-08-19 that 11% of New Jersey sections had changed (~200
changes). Clever derives the section display name from the primary teacher's
last name, so a single flip writes two change rows (`name` and `teacher`) —
about 90 ENR sections times 2 fields accounts for the alert.

The flip is not directional. In the reported batch, two schools moved from
School Leader to DSO and one moved from DSO to School Leader. A real
org change would move one way; a re-rolled tie-break scatters.

Expected: the DSO is always `teacher_id` and the School Leader is always
`teacher_2_id`, stable across runs.

## Steps to reproduce

1. Query `kipptaf_extracts.rpt_clever__sections` filtered to `course_number = 'ENR'`.
2. Join `teacher_id` and `teacher_2_id` back to `int_people__staff_roster` on `powerschool_teacher_number`.
3. Observe that `teacher_id` resolves to the School Leader at 6 of 19 NJ schools and to the DSO at the other 13, despite the model comment declaring the DSO is the intended ENR teacher.

## Where

- **Code location / dbt project:** `kipptaf` — `src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** Clever NJ section-change alert, 2026-08-19

<details>
<summary>For Claude</summary>

Root cause is confirmed. The `dsos` CTE matches four job titles, one of which
is `School Leader`. Every one of the 19 NJ schools has exactly two matching
staff: one DSO-flavored title and one School Leader. The ENR branch of
`teachers_long` hardcodes `1 as sortorder` for both, so the
`row_number() over (partition by section_id order by sortorder asc)` in
`pre_pivot` is a pure tie and BigQuery resolves it by scan order.

The tie has existed since the model was created (2023-07-24); the title list is
byte-identical from then to now. The likely re-roll trigger is `54da6e6404`
(2026-08-11), which removed a `LEFT JOIN` from the `dsos` CTE and changed the
query plan without touching any teacher logic.

Fix: rank the titles in `dsos` so the DSO sorts ahead of the School Leader,
carry that through the ENR union branch instead of the hardcoded `1`, and add
`teachernumber` as a residual tie-break in the `row_number` so the
regular-section branch cannot hit the same class of bug when two teachers share
a `roledef` sortorder.

Out of scope for now: whether `School Leader` should be dropped from the title
list entirely. That is a pending decision with the Ops team; this issue only
makes the existing behavior deterministic.

Fixed means: rebuilding the model twice, or across a plan change, yields the
same `teacher_id` for every ENR section, and that `teacher_id` is the DSO at
all 19 schools.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop the "Enroll" (ENR) sections in the
Clever feed from randomly swapping their listed teacher between the school's
DSO and the school's School Leader.

Every NJ school has exactly two people who match the job-title list that picks
the ENR "teacher" — one DSO and one School Leader — and the model gave both the
same sort rank. With nothing to break the tie, BigQuery picked the primary
teacher by whatever order it happened to scan the rows, so the two could trade
places on any run with no change in PowerSchool or ADP. Clever names each
section after its primary teacher, so every swap wrote two change records and
eventually tripped Clever's section-churn alert (11% of NJ sections, ~200
changes, reported 2026-08-19).

This ranks the DSO ahead of the School Leader so the assignment is stable and
matches what the model always intended.

Refs #4928

## Reviewer Notes

- The School Leader stays in the title list. Whether to drop it entirely is a
  separate open question with the Ops team; this PR only makes the current
  behavior deterministic. Dropping it later is a one-line follow-up.
- There is a third change beyond the two that fix ENR: a residual tie-break on
  the pivot's row_number. That one guards the regular-section branch, which can
  hit the same class of bug whenever two teachers on a section share a roledef
  sort order.
- Expect one last burst of churn in Clever when this ships, then it stops. The
  6 ENR schools that are currently inverted snap back to the DSO, plus 5 regular
  sections whose co-teachers tie on roledef sort order. Measured, not estimated:
  845 regular sections carry such a tie, but only 5 of the 2,491 non-ENR
  sections actually change hands under the new order.
- A regression test now guards the invariant (`d60771dd6`). It returns 37 rows
  against prod's current pre-fix values and 0 against the fix.

### General

### dbt

      models; `rpt_clever__sections.yml` already exists and needs no change.
- [x] Include (or update) an
      change; the existing Clever extract exposure still applies.
- [x] **Breaking change?** No. No columns renamed, removed, or retyped. Only
      which teacher number lands in `teacher_id` vs `teacher_2_id` changes.
      not applicable, no source changes.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Root cause was traced from a Slack report, not
guessed.

Mechanism: the `dsos` CTE matches four job titles, one of which is
`School Leader`. All 19 NJ schools match exactly two staff — one DSO-flavored
title plus one School Leader. The ENR branch of `teachers_long` hardcoded
`1 as sortorder` for both, making
`row_number() over (partition by section_id order by sortorder asc)` in
`pre_pivot` a pure tie resolved by scan order.

The tie has existed since the model was created (`3ec16f4d4b`, 2023-07-24); the
title list is byte-identical from then to now, so `School Leader` was not added
recently. The likely re-roll trigger is `54da6e6404` (2026-08-11), which removed
a `LEFT JOIN` from the `dsos` CTE and changed the query plan without touching
any teacher logic.

Direction confirms it is a tie-break re-roll rather than an org change: in the
reported batch two schools moved School Leader to DSO while one moved DSO to
School Leader. A real org change moves one way.

Verification. `dbt build --select rpt_clever__sections` in the worktree
compiles and creates the view, and the
`dbt_utils_unique_combination_of_columns` test passes. The build proves
compilation only, not values — every `ref` resolved to a stale, empty
`zz_cbini_*` dev copy rather than deferring to prod, so the dev view returned
zero rows. The behavioral check therefore reran the compiled ENR path against
prod upstreams: all 90 ENR sections now place a DSO-flavored title in
`teacher_id` (82 Director School Operations, 8 Director Campus Operations) and
School Leader in `teacher_2_id`, versus 6 of 19 schools inverted before.

The `rpt_clever__school_id_resolves_to_schools_feed` test failed locally with 38
rows. All 38 are in the `teachers` and `staff` feeds and none in `sections`;
prod runs the same test clean. That is stale-dev drift on two models this PR
does not touch, not a regression.

`trunk check --force --no-fix` on the changed SQL: no issues.

</details>

### DIFF
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql
@@ -4,6 +4,11 @@ with
             sr.powerschool_teacher_number,
             sr.home_work_location_dagster_code_location,
             sr.home_work_location_powerschool_school_id as school_id,
+
+            -- The DSO is the intended ENR "teacher"; School Leader is the backup.
+            -- Every school matches both, so without an explicit rank the pivot's
+            -- row_number tie-break is arbitrary and the two swap between runs.
+            if(sr.job_title = 'School Leader', 2, 1) as sortorder,
         from {{ ref("int_people__staff_roster") }} as sr
         where
             sr.assignment_status != 'Terminated'
@@ -127,7 +132,7 @@ with
                 right('{{ var("current_fiscal_year") }}', 2)
             ) as terms_abbreviation,
 
-            1 as sortorder,
+            dsos.sortorder,
 
             dsos.powerschool_teacher_number as teachernumber,
 
@@ -171,7 +176,9 @@ with
 
             concat(
                 'teacher_',
-                row_number() over (partition by section_id order by sortorder asc),
+                row_number() over (
+                    partition by section_id order by sortorder asc, teachernumber asc
+                ),
                 '_id'
             ) as input_column,
         from teachers_long
--- a/src/dbt/kipptaf/tests/properties.yml
+++ b/src/dbt/kipptaf/tests/properties.yml
@@ -456,6 +456,23 @@ data_tests:
           ref:
             name: int_people__staff_roster
 
+  - name: rpt_clever__sections__enr_teacher_id_is_the_dso
+    description: >-
+      The auto-generated ENR sections must list the school's DSO as their
+      primary teacher, never the School Leader. Both titles match the same
+      filter in rpt_clever__sections, so the pick is stable only because
+      sortorder ranks them -- without that rank the two swap on BigQuery's scan
+      order, which is how ~90 sections silently changed hands and tripped
+      Clever's churn alert in August 2026. Errors rather than warns: Clever
+      derives each section's display name from its primary teacher, so a
+      regression here rewrites every ENR section name in the feed.
+    config:
+      severity: error
+      meta:
+        dagster:
+          ref:
+            name: rpt_clever__sections
+
   - name: rpt_clever__school_id_resolves_to_schools_feed
     description: >-
       Cross-feed referential integrity for the six Clever SFTP extracts -- every
--- a/src/dbt/kipptaf/tests/rpt_clever__sections__enr_teacher_id_is_the_dso.sql
+++ b/src/dbt/kipptaf/tests/rpt_clever__sections__enr_teacher_id_is_the_dso.sql
@@ -1,0 +1,20 @@
+-- The auto-generated ENR sections carry the school's DSO as their primary
+-- teacher and the School Leader as the backup. Both match the same job-title
+-- filter, so the pick is only stable because `sortorder` ranks them; drop that
+-- rank and the two swap on scan order, silently and without failing anything.
+-- Clever names each section after its primary teacher, so a swap rewrites every
+-- ENR section name and trips its churn alert. A teacher_id that resolves to no
+-- active roster row at all lands here too.
+select s.section_id, s.teacher_id, sr.job_title,
+from {{ ref("rpt_clever__sections") }} as s
+left join
+    {{ ref("int_people__staff_roster") }} as sr
+    on s.teacher_id = sr.powerschool_teacher_number
+    and sr.assignment_status != 'Terminated'
+where
+    s.course_number = 'ENR'
+    and coalesce(sr.job_title, 'unresolved') not in (
+        'Director of Campus Operations',
+        'Director Campus Operations',
+        'Director School Operations'
+    )


cbini:
## Review response

Verified all five findings against the warehouse and the repo before replying. Verdicts below; one finding changed the PR, one changed a number in the PR body, three declined with reasons.

### 1. `'Director of Campus Operations'` is dead — confirmed, declining in this PR

Correct, and stronger than "medium confidence": that exact string matches **0 of 4,762 rows** in `int_people__staff_roster`, across every assignment status. `'Director Campus Operations'` (no "of") matches 13. `rpt_clever__sections.sql:19` is also the only occurrence of the "of" variant anywhere in `src/` — the four models cited all use the no-"of" form, as described.

Declining here to keep the diff to the defect. It is dead weight, not a second live list, so it changes no behavior. Happy to strip it in a follow-up.

### 2. Reuse `int_people__leadership_crosswalk` — confirmed, follow-up

The citation checks out: `int_people__leadership_crosswalk.sql:37-47` does resolve one canonical `dso_employee_number` per school via that `coalesce(max(if(...)))` priority pattern, and it structurally cannot tie.

It also does not select `powerschool_teacher_number` or `home_work_location_dagster_code_location` — confirmed, neither appears in the file — so reuse means extending the crosswalk and rekeying this model onto it. That is a bigger change than the defect warrants while Clever is actively churning. Agreed as a follow-up, especially since `54da6e6404` was already moving toward the crosswalk owning this kind of lookup.

### 3. Blast radius of the `teachernumber` tiebreak — mechanism confirmed, impact measured at 5 sections

The mechanism is real and I had not accounted for it. Regular sections do tie on `roledef.sortorder`: **1,052 tied groups across 845 sections** in the current year.

But the tie existing is not the same as the pick changing. Comparing the new deterministic order against what prod currently ships, for all 2,491 non-ENR sections:

| | sections |
| --- | --- |
| unchanged | 2,486 |
| `teacher_id` changes | **5** |

So the extra churn is 5 sections, not 845 — BigQuery's scan order already agreed with `teachernumber asc` almost everywhere. Those 845 sections are unguarded today and will be pinned after this merges, which is the point of the tiebreak.

Updating the PR body with the measured number rather than leaving it unmentioned.

### 4. No rank between the three DSO-flavored titles — confirmed non-issue, declining

Correct that they all collapse to `sortorder = 1`. It cannot bite today for two reasons: the `'of'` variant matches zero rows (finding 1), leaving only two live DSO titles, and no school has staff in both — all 19 NJ schools match exactly one DSO-flavored title plus one School Leader. Adding a rank between titles that never co-occur would be inventing a policy no one has decided. Leaving it.

### 5. No regression test — accepted, added in `d60771dd6`

This is the one that mattered. The swap was invisible for three years precisely because nothing asserted the outcome.

Added `tests/rpt_clever__sections__enr_teacher_id_is_the_dso.sql` at `severity: error`: every ENR section's `teacher_id` must resolve to a DSO-flavored active roster row, and an id that resolves to no active row lands there too.

Confirmed the test actually catches the defect rather than passing vacuously — a local build passes it, but the dev view resolves to empty upstream copies, so that PASS proves nothing. Running its predicate against prod's current pre-fix values returns **37 rows**; against the fixed ordering it returns **0**.

PR #4932, opened 2026-08-19
fix(kipptaf): gate grad_calcs flags on the matching enrollment date

### LINKED ISSUE #4874: fix(kipptaf): grad_calcs flags gate on wrong enrollment date, plus BA/AA/CTE ever-earned metrics
## Summary

Six graduation flags in `int_kippadb__grad_calcs` test degree status on one enrollment but gate the completion date on a **different** enrollment. The mismatch silently drops real graduates, and it currently produces logically impossible results on the KIPP Forward dashboard: a student can count as graduating within 6 years while not counting as having graduated at all.

Separately, there is no way to ask whether a student **ever** earned a specific credential. The BA/AA/CTE flags are all window-bound (4/5/6/7 year), so completers outside the widest window are invisible to every metric.

Reported by a KIPP Forward analyst who noticed `Is Grad Ever` reading lower than expected against the 6-year BA flag for cohorts 2020 and 2021.

## Root cause

`int_kippadb__grad_calcs.sql` lines 109-118:

```sql
if(
    (
        e.ba_status = 'Graduated'
        or e.aa_status = 'Graduated'
        or e.cte_status = 'Graduated'
    )
    and e.ugrad_actual_end_date <= current_date('{{ var("local_timezone") }}'),
    1,
    0
) as is_grad_ever,
```

The status test is on `ba` / `aa` / `cte`, but the date gate is on `ugrad_actual_end_date`. In `int_kippadb__enrollment_pivot`, `ugrad_enrollment_id` is BA-or-AA only and is picked by **latest start_date**, so it drops two populations:

1. **CTE-only completers.** No BA/AA enrollment at all, so `ugrad_actual_end_date` is NULL. The NULL comparison makes the whole `and` NULL and `if()` returns 0.
2. **Graduates who later enrolled somewhere still in progress.** The pick selects the newer, in-progress BA/AA whose `actual_end_date` is NULL. Same NULL collapse. This includes students who genuinely earned a bachelor's and are now attending a community college.

The same `ugrad_*` pick is used by `is_4yr_ugrad_grad_int`, `is_5yr_ugrad_grad_int`, `is_6yr_ugrad_grad_int`, `is_6yr_ugrad_cte_grad_int`, and `is_24yo_ugrad_cte_grad_int`, so all six columns share the defect. The degree-specific `ba_` / `aa_` / `cte_` flags are unaffected.

## Evidence

All figures verified against `teamster-332318.kipptaf_kippadb` prod (10,100 students in `int_kippadb__grad_calcs`).

Undercounts:

| Column | Shipped | Correct | Undercount |
| --- | --- | --- | --- |
| `is_grad_ever` | 712 | 819 | **107** |
| `is_6yr_ugrad_grad_int` | 537 | 624 | **87** |

Logically impossible row counts today:

```text
is_6yr_grad_any_int = 1 and is_grad_ever = 0        103 students
is_6yr_aa_grad_int  = 1 and is_6yr_ugrad_grad_int = 0    50 students
is_6yr_ba_grad_int  = 1 and is_6yr_ugrad_grad_int = 0    37 students
is_6yr_ba_grad_int  = 1 and is_grad_ever = 0          5 students
ba_status Graduated     and is_grad_ever = 0          7 students
```

BA and AA are subsets of undergrad by definition, and graduating within 6 years is a subset of graduating ever, so these are hard contradictions rather than mere undercounts.

### The date gate

`current_date()` in `is_grad_ever` suppresses exactly **one** row network-wide, a record dated roughly 175 years in the future. That is a data-entry typo, not an anticipated completion. Removing the gate moves the count from 818 to 819 and changes nothing else, while eliminating a real problem: `grad_calcs` is a view, so `current_date()` evaluates at read time and two Tableau extracts taken either side of midnight can disagree with no upstream data change. No historical value is reproducible.

### Missing ever-earned metrics

604 students have ever earned a bachelor's. `is_7yr_ba_grad_int` captures 568, so **36 BA earners are invisible to every existing metric**. Years-to-BA distribution among completers:

| Years | 2 | 3 | 4 | 5 | 6 | 7 | 8 plus |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Students | 2 | 8 | 307 | 167 | 51 | 32 | **35** |

### 1. Replace the ugrad date gate with per-degree legs

For each of the six affected columns, test status and date on the same enrollment:

```sql
if(
    (e.ba_status = 'Graduated' and e.ba_actual_end_date <= date((c.contact_kipp_hs_class + 6), 08, 31))
    or (e.aa_status = 'Graduated' and e.aa_actual_end_date <= date((c.contact_kipp_hs_class + 6), 08, 31)),
    1,
    0
) as is_6yr_ugrad_grad_int,
```

A single-pick column such as `postsec_completed_enrollment_id` was considered and rejected for the windowed flags: 44 students earned both a BA and an AA and 43 of them finished the AA first, so one pick and one date systematically overstates time-to-completion. Roughly 30 students diverge at the 6-year window. `postsec_completed_*` remains the right column for reporting **which** credential a student completed.

### 2. Drop the date gate from `is_grad_ever`

```sql
if(
    e.ba_status = 'Graduated'
    or e.aa_status = 'Graduated'
    or e.cte_status = 'Graduated',
    1,
    0
) as is_grad_ever,
```

### 3. Add three ever-earned metrics

```sql
if(e.ba_status = 'Graduated', 1, 0) as is_ba_grad_ever_int,
if(e.aa_status = 'Graduated', 1, 0) as is_aa_grad_ever_int,
if(e.cte_status = 'Graduated', 1, 0) as is_cte_grad_ever_int,
```

Safe without a date gate: zero graduated BA/AA/CTE enrollments have a NULL `actual_end_date`, and the pick in `int_kippadb__enrollment_pivot` orders `is_graduated desc`, so a graduated enrollment always wins over a non-graduated one at the same degree level. Naming follows rubric R3 plus the family convention of an `_int` suffix on countable INT64 flags.

These three make `is_grad_ever` exactly the greatest of the three, which is a self-documenting invariant.

## Tests

`int_kippadb__grad_calcs.yml` currently has **zero** `data_tests`, including no uniqueness test on `student` despite the repo requiring one on every intermediate model. Nothing would have caught this class of bug.

Add:

- `unique` on `student`
- Window nesting per degree, for example:

  ```text
  is_4yr_ba_grad_int <= is_5yr_ba_grad_int <= is_6yr_ba_grad_int <= is_7yr_ba_grad_int
  ```

- Subset invariants:

  ```text
  is_7yr_ba_grad_int  <= is_ba_grad_ever_int
  is_6yr_ba_grad_int  <= is_6yr_ugrad_grad_int
  is_6yr_aa_grad_int  <= is_6yr_ugrad_grad_int
  is_6yr_grad_any_int <= is_grad_ever
  is_ba_grad_ever_int <= is_grad_ever
  ```

- Warn-level test on `stg_kippadb__enrollment` for `status = 'Graduated'` with a future-dated `actual_end_date`, so the typo surfaces instead of being silently suppressed by the removed gate.

`grad_calcs` is a view, so per `src/dbt/CLAUDE.md` these tests only re-run at build time. Add `config: materialized: table` in its properties YAML if regular refresh is wanted.

## Sequencing

Window nesting and `is_7yr_ba_grad_int` at most `is_ba_grad_ever_int` pass today. The two invariants involving `is_grad_ever` fail until change 2 lands, so the new metrics and the fix should ship in the same PR.

## Impact

- `is_grad_ever` moves 712 to 819, about **plus 15 percent** on a published KIPP Forward dashboard. Needs stakeholder comms before merge.
- Only consumer of these columns is `rpt_tableau__kfwd_dashboard` (contract enforced, so the three new columns need YAML entries with `data_type: int64`). No `src/cube` references. No existing test or contract breaks; `data_type` on changed columns is unchanged.
- `rpt_tableau__kfwd_spi.sql:146` carries a separate, independently derived `is_grad_ever_any` at enrollment grain that does not read `grad_calcs`. Two dashboards will continue to report two different definitions of graduating ever. Out of scope here; worth a follow-up.
- The `is_grad_ever` description in the properties YAML documents the mismatched gate as intended behavior and must be rewritten.

## Ruled out

- **Cohort key divergence.** `int_kippadb__roster.sql:284` uses `coalesce(contact_kipp_hs_class, se.cohort)` for `ktc_cohort` while `grad_calcs` uses the raw `contact_kipp_hs_class`, which would silently zero every window flag on a NULL. In practice `contact_kipp_hs_class` is never NULL across all 10,100 students and never diverges from `ktc_cohort`, so the fallback never fires for this population.
- **Degree scope confusion alone.** `is_grad_ever` being any-degree while `is_6yr_ba_grad_int` is BA-only explains part of the reported gap, but not the 5 students whose BA graduation is excluded from `is_grad_ever` outright.

CGibson17:
# Pull Request

## Summary & Motivation

When merged, this pull request will correct six graduation flags in
`int_kippadb__grad_calcs` that tested degree status on one enrollment but gated
the completion date on a **different** one, and will add three unbounded
ever-earned metrics.

A KIPP Forward analyst noticed `Is Grad Ever` reading lower than expected
against the 6-year BA flag for cohorts 2020 and 2021. The root cause is line 115
of the old model: the status test is on `ba` / `aa` / `cte`, but the date gate is
on `ugrad_actual_end_date` — a BA-or-AA-only enrollment picked by latest start
date. Where that pick is NULL or still in progress, the NULL comparison makes
the whole predicate NULL and the flag reads 0.

The result was logically impossible output on a published dashboard: a student
could count as graduating within 6 years while not counting as having graduated
at all. On prod today that contradiction covers 103 students, plus 87 more where
a BA or AA graduate is excluded from the undergraduate flag that contains it by
definition.

Each flag now gates each degree on its own end date:

| Column | Before | After |
| --- | --- | --- |
| `is_grad_ever` | 712 | **819** |
| `is_6yr_ugrad_grad_int` | 537 | **624** |
| `is_4yr_ugrad_grad_int` | 327 | **386** |
| `is_6yr_ugrad_cte_grad_int` | 662 | **749** |

New unbounded metrics — the BA family previously topped out at a 7-year window,
so 36 bachelor's earners were invisible to every existing flag:

| Column | Students |
| --- | --- |
| `is_ba_grad_ever_int` | 604 |
| `is_aa_grad_ever_int` | 127 |
| `is_cte_grad_ever_int` | 149 |

Two further decisions:

- **The `current_date()` gate is removed from `is_grad_ever`.** It suppressed
  exactly one record network-wide — a completion date roughly 175 years in the
  future, i.e. a data-entry typo — while making an ever metric depend on read
  time. `grad_calcs` is a view, so two Tableau extracts either side of midnight
  could disagree with no upstream data change. That record now surfaces as a
  warn-level test on `stg_kippadb__enrollment` instead of being silently hidden.
- **`is_6yr_ugrad_cte_grad_int` is identical to `is_6yr_grad_any_int`** once each
  degree is gated on its own end date — BA-or-AA-or-CTE is the same set as
  any-degree, and 0 of 10,100 students differ. `grad_calcs` now carries one
  derivation; the dashboard aliases it under both names so no Tableau field
  breaks. Retiring the legacy name is left as follow-up.

### Rollback plan

No column is renamed or removed from any contracted model — the three new
columns are additive and `is_6yr_ugrad_cte_grad_int` is preserved as an alias.
Reverting this commit restores the prior values with no contract change and no
coordination needed.

### AI Assistance

Claude Code authored the investigation, the fix, and the tests, working from the
analyst's report. Human-directed: the decision to open #4874, the choice to fix
the whole `ugrad` family rather than `is_grad_ever` alone, the decision to add
all three ever-earned metrics, and the decision to resolve the duplicate column
in this PR rather than track it separately.

An adversarial review pass was run against the fix plan before implementation.
It overturned the original approach — the plan had proposed gating on the
single-pick `postsec_completed_*` column, which is correct for `is_grad_ever`
today but diverges from correct on roughly 30 students for the windowed flags,
because 43 of the 44 students who earned both a BA and an AA finished the AA
first. Per-degree legs are the only shape that works for all six columns.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — the
      three new columns are documented, and `is_6yr_grad_any_int` gained the
      column entry it was previously missing
- [x] Include (or update) an exposure — `rpt_tableau__kfwd_dashboard` is already
      covered by the KIPP Forward Data Suite exposure; no new consumer
- [x] **Breaking change?** No. Additive columns only; the one redundant column
      is kept as an alias rather than dropped. Rollback plan above.
- [x] External sources — not applicable, no external source added or modified

## Validation

All run locally against deferred prod upstreams.

```text
dbt parse --no-partial-parse                              clean
dbt build int_kippadb__grad_calcs + 4 tests               PASS=5, 0 errors
dbt build rpt_tableau__kfwd_dashboard (contract enforced) PASS, 3 new columns
dbt test enrollment_graduated_end_date_not_future         WARN=1 (the typo)
trunk check --force (5 files)                             No issues
```

The three new invariant tests are not decorative — run against the unfixed prod
model they fail at 107, 103 and 87 rows respectively:

```text
greatest(is_ba_grad_ever_int, is_aa_grad_ever_int, is_cte_grad_ever_int) = is_grad_ever
is_6yr_grad_any_int <= is_grad_ever
is_6yr_ba_grad_int <= is_6yr_ugrad_grad_int and is_6yr_aa_grad_int <= is_6yr_ugrad_grad_int
```

`int_kippadb__grad_calcs.yml` previously had **zero** `data_tests`, including no
uniqueness test on `student` despite the repo requiring one on every
intermediate model. That is why nothing caught this before a human noticed it on
a dashboard.

## Deliberate deviations

- **No `materialized: table` flip.** Making these tests refresh continuously in
  prod would need it, but per `src/dbt/CLAUDE.md` a view-to-table flip requires
  mapping every consumer's refresh cadence first, and this dashboard refreshes
  twice daily. The tests still run on every CI build of the model, which is the
  regression guard that matters here. Left as follow-up.
- **The new staging test is `severity: warn`, not `error`.** Repo convention says
  staging tests must be `error`, but the violating row exists in prod today, so
  `error` would fail the build for everyone immediately. Set explicitly so it is
  a deliberate warn rather than a silent degrade to the project default.

## Notes for reviewers

- `is_grad_ever` moves **712 to 819, about plus 15 percent**, on a published
  KIPP Forward dashboard. The old number was undercounting 107 real graduates.
  Worth telling the KIPP Forward team before merge rather than having them
  discover it.
- I could not verify whether the Tableau workbook references
  `is_6yr_ugrad_cte_grad_int` — it uses an embedded datasource, which the REST
  API does not expose. That is why the column is aliased rather than dropped.
- `rpt_tableau__kfwd_spi.sql` carries a separate, independently derived
  `is_grad_ever_any` at enrollment grain that does not read `grad_calcs`. Two
  dashboards will continue to report two different definitions of graduating
  ever. Out of scope here; worth a follow-up.
- A cohort-key divergence theory was investigated and ruled out:
  `contact_kipp_hs_class` is never NULL across all 10,100 students and never
  diverges from `ktc_cohort`, so the `coalesce` fallback in
  `int_kippadb__roster` never fires for this population.

Refs #4874


anthonygwalters (dismissed):

CGibson17:
## Review responses

Verdict per finding. Two fixed in b490b2ad6, one declined with reasoning, one confirmed and headed for a follow-up issue.

### 1. `is_grad_ever` lost its end-date sanity check — valid concern, verified, keeping the change

I ran your query against prod. The answer is **1 row**:

```text
status = 'Graduated' and actual_end_date > current_date  ->  1 student
```

That record's end date is roughly **175 years** in the future (64,188 days) — a data-entry typo, not an anticipated completion. So the practical effect of removing the gate is that `is_grad_ever` reads 819 instead of 818.

Keeping the removal, because the old gate was not a data-quality guard — it was an accidental one-row filter with a real cost. `grad_calcs` is a view, so `current_date()` evaluated at read time, meaning two Tableau extracts taken either side of midnight could disagree with no upstream data change, and no historical value was reproducible. The typo is now visible via `enrollment_graduated_end_date_not_future` rather than silently absorbed.

You are right that a warn test does not block the row. That is the intended trade: surface the bad record to whoever can fix it in kippadb instead of hiding it inside a flag. If the preference is to block instead, the lever is `severity: error` on that staging test — but that fails the build for everyone until Ops corrects the record, which is why it is a warn.

### 2. Inconsistent multi-line YAML scalar style — fixed

Fixed in b490b2ad6. `grad_calcs_grad_ever_decomposes_by_degree` now uses the literal block, matching its sibling and the `src/dbt/CLAUDE.md` rule. Confirmed the plain scalar was in fact inconsistent with the sibling three tests down.

### 3. Same root-cause pattern persists elsewhere — confirmed accurate, follow-up warranted

Verified rather than taken on assertion. `eis.ugrad_status` appears in **10 places** in `int_kippadb__persistence.sql` (lines 112, 133, 154, 158, 161, 242, 263, 284, 288, 291), and `rpt_tableau__kfwd_aid_report.sql` (76, 78) plus `rpt_tableau__kfwd_spi.sql` (42) project the `ugrad_*` columns straight through.

`int_kippadb__enrollment_pivot`'s `ugrad_enrollment_id` pick is unchanged by this PR, so the single-pick-by-start-date problem does still reach those consumers. Out of scope here — this PR deliberately stops at `grad_calcs` so the diff stays reviewable against one Tableau consumer — but it is a real follow-up and will be tracked.

### 4. Derive `is_grad_ever` from the three new columns — declined, would not compile

This one does not work on BigQuery. `src/dbt/CLAUDE.md` is explicit:

> **No lateral column aliases.** BigQuery rejects a `SELECT`-list alias referenced by another item in the same list — `select 1 as a, case when a = 1 then 'x' end as b` fails `Unrecognized name: a`.

`is_ba_grad_ever_int` and its siblings are aliases in the same `SELECT` list as `is_grad_ever`, so referencing them there fails at build. Making the invariant true by construction would need a CTE, which trades a cheap test for an extra layer on a view that already re-expands the whole kippadb chain.

Keeping the independent derivation plus the cross-check test. Worth noting the invariant is not merely stylistic — it failed on 107 rows against the pre-fix model, so the test is a real tripwire, not a tautology guard.

### 5. Model description isn't fully exhaustive — fixed

Correct, and the description was mine. `is_24yo_ugrad_cte_grad_int` is measured from `contact_birthdate` plus 25 years, not from the class-year August 31. Reworded in b490b2ad6 to say most windows are class-year-anchored and to name that column as the exception.

### 6. `severity: warn` on the new staging test — agreed, and the precedent check matches

Same conclusion reached independently before pushing, for the same reason. Recording the rationale explicitly: the violating row exists in prod today, so `error` would red the build immediately; `warn` is set explicitly rather than left to degrade to the project default silently.

### 7. Missing `description:` on new Tableau extract columns — agreed, pre-existing

Confirmed no column in that file carries a description. Not introducing a new gap, and not fixing ~230 columns in this PR.

---

Fixes are in b490b2ad6. Findings 1 and 4 are deliberate declines with the reasoning above rather than silent dismissals.

Issue #4927, opened 2026-08-19
feat(dbt): source Miami day-grain tardy and suspension flags for the attendance facts


cbini:
## What and why

`int_students__attendance_daily`, introduced by #4924, leaves six flags null for
Miami because Focus carries no day-grain source for them.

| Flag | Why it is null |
| --- | --- |
| `is_tardy` | Focus records tardies only at period grain |
| `is_ontime` | derived from `is_tardy` |
| `is_oss` | Focus attendance has no suspension codes |
| `is_iss` | Focus attendance has no suspension codes |
| `is_suspended` | Focus attendance has no suspension codes |
| `is_absent_non_susp` | depends on the suspension codes |

Null rather than zero was deliberate. A zero would read as a verified zero
suspension rate and a verified zero tardy rate for Miami, and the data does not
support either claim. The consequence is that Miami is excluded from network
tardy and suspension metrics rather than quietly diluting them.

These are two independent gaps with different fixes and different owners. They
are filed together because both are the same deferral from #4924.

### Tardy

Focus day grain carries 4 codes for AY2026: null (present, 8,284 rows), `U`
absent unexcused (943), `AE` absent excused (19), `AD` absent documented (7).
None is a tardy.

Period grain does carry it — 528 AY2026 rows titled `Tardy`, state code `P`. So
a day-grain flag needs a period-to-day rollup, and the rollup rule is the open
question:

- any period tardy marks the day tardy — cannot understate, but a student tardy
  to one mid-day class then reads the same as one who arrived late to school
- only the day's first scheduled period — matches the arrived-late-to-school
  meaning NJ's daily code carries, but Focus period ordering needs verifying
  before anything relies on it

### Suspension

Focus attendance carries no suspension codes at any grain. Miami suspension data
lives in DeansList, already modeled as `int_deanslist__*` and consumed elsewhere
in kipptaf.

## Proposed approach

Ship separately. Tardy does not depend on suspensions.

**Tardy**: roll `int_focus__attendance_period` up to day grain, settle the rule
above, and populate `att_code` so the existing `att_code like 'T%'` flag logic in
`int_students__attendance_daily` picks it up with no new branching. Note this
also activates the 0.67 tardy weight in `is_present_weighted` for Miami, so
expect Miami ADA to fall slightly when it lands.

**Suspension**: source from DeansList at Miami's district altitude and conform to
the PowerSchool code vocabulary the flags already read — `OS` / `OSS` / `OSSP` /
[...]

PR #4936, opened 2026-08-19
fix(dbt): add Focus attendance models to the focus package for Miami

### LINKED ISSUE #4924: fix(dbt): wire Focus attendance into the kipptaf network layer for Miami AY2026
## What's happening

Miami has been in session since 2026-08-12 and contributes **zero attendance rows** to the kipptaf network layer.

`int_powerschool__ps_adaadm_daily_ctod` holds 229,463 Miami AY2025 rows from the frozen `kippmiami_powerschool` archive and 0 rows for AY2026. Everything downstream reads empty for Miami: `fct_student_attendance_daily`, `fct_student_attendance_streaks`, `fct_student_attendance_interventions`, `dim_school_calendars`, and `bridge_survey_expectations` — so Miami ADA, chronic absenteeism, and topline Total Enrollment all understate the network.

Focus has the data and the Miami-side models exist. Nothing at kipptaf level consumes them.

| Model | Rows | Note |
| --- | --- | --- |
| `stg_focus__attendance_day` | 9,253 | grows every school day |
| `stg_focus__attendance_period` | 18,156 | grows every school day |
| `int_focus__attendance_day` | 9,253 | shipped in PR #4919 |
| `int_focus__attendance_period` | 18,156 | shipped in PR #4919 |

Both prerequisites are closed — #4220 for staging coverage and #4865 for the intermediates. What was never built is the kipptaf branch, so the closed issues read as "attendance done" when the network layer was never wired.

## Steps to reproduce

1. Confirm Miami is absent from the network attendance scaffold:

   ```sql
   select academic_year, count(*) as rows
   from `teamster-332318.kipptaf_powerschool.int_powerschool__ps_adaadm_daily_ctod`
   where regexp_contains(_dbt_source_relation, 'kippmiami')
   group by 1
   order by 1
   ```

   No AY2026 row is returned. Newark and Camden both carry full AY2026-27 scaffolds.

1. Confirm no Focus branch exists: `grep -rl focus src/dbt/kipptaf/models/powerschool/` returns nothing.

1. Confirm the Focus data is there: `select count(*) from` `teamster-332318.kippmiami_focus.int_focus__attendance_day`.

## Where

- **Code location / dbt project:** `kipptaf`, plus a source declaration and wrapper step in the kipptaf focus layer
- **Environment:** prod
- **Related:** Phase 2 of the plan on #4729. Resolves the current-year leg of #4803.

<details>
<summary>For Claude</summary>

### Prerequisite step

`int_focus__attendance_day` and `int_focus__attendance_period` are not declared in `kipptaf/models/focus/sources-kippmiami.yml` and have no kipptaf wrapper. Per `kipptaf/CLAUDE.md`, every source added to a `sources-kipp*.yml` needs a matching `union_relations` passthrough, and consumers read the wrapper. That is a two-PR sequence per `src/dbt/CLAUDE.md` — the package model must materialize in prod before kipptaf reads it, and the kipptaf PR needs `dbt clone --select ... --target staging` to refresh `zz_stg_kippmiami_focus` or CI reads a stale copy.

| Model | Role |
| --- | --- |
| `int_powerschool__ada` | daily attendance grain, feeds `int_students__attendance_interventions` |
| `int_powerschool__calendar_week` | direct parent of `ps_adaadm_daily_ctod` |
| `int_powerschool__calendar_rollup` | calendar aggregation |
| `int_powerschool__attendance_streak` | feeds `fct_student_attendance_streaks` |
| `int_powerschool__ps_adaadm_daily_ctod` | builds from `calendar_week` plus `int_students__terms` |
| `stg_powerschool__attendance`, `stg_powerschool__attendance_code`, `stg_powerschool__calendar_day` | staging passthroughs unioning the archive |

### Open design decision

Phase 1 did not add Focus branches to same-named `int_powerschool__*` models — it created SIS-neutral `int_students__*` models instead (`int_students__student_enrollment_union`, `int_students__terms`, `int_students__schools`, `int_students__students`). Decide whether the attendance branch follows that precedent as `int_students__*` models or branches in place. Following Phase 1 is probably right, but it widens the diff and touches the `student_attendance_daily_key` and `student_enrollment_key` hash inputs — worth settling before writing code.

### Known constraints from #4865

- `attendance_day.daily_code` holds a `short_name`, not an id. Short names repeat across schools and years and `attendance_day` has no `school_id`, so school resolves from the marking period before the code join. `attendance_period.attendance_code` joins directly on id with zero orphans.
- `state_value` is already the present or absent classification; 6,804 of 7,699 day rows are `daily_code is null` with `state_value = 1`. The code join supplies the reason for an absence, not the classification.
- `attendance_calendar.minutes` is the sentinel `999` on every matched 2026 row, so there is nothing to compare `minutes_present` against.
- A missing attendance record is not visible at this grain. Finding absent days needs an enrolled-students-crossed-with-calendar-days scaffold, which belongs to the consuming mart.

### Scope boundary

Focus carries AY2026-27 forward only. This restores Miami's current year. It does not re-key the AY2025 rows that came from the frozen archive, so #4803 still needs a separate answer for its historical leg — expect that orphan count to move, not vanish.

### Validation

1. Miami AY2026 presence: `int_powerschool__ps_adaadm_daily_ctod` carries Miami rows from 2026-08-12 forward where there are 0 today.
1. `fct_student_attendance_daily` reports Miami for AY2026, and Miami ADA is non-null.
1. Re-measure the #4803 orphan count and record the new number there.

Refs #4729
Refs #4803
Refs #4865

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will add six models to the `focus` dbt package that
shape KIPP Miami's Focus attendance data for downstream use.

Miami has been in session since 2026-08-12 and currently contributes **zero
attendance rows** to the network reporting layer. Focus holds the data and the
Miami-side intermediates shipped in #4919, but nothing consumes them, so Miami ADA,
chronic absenteeism, and topline Total Enrollment all understate the network.

This PR builds the Miami side only. **Nothing downstream changes yet** — no existing
model, dashboard, or number moves. A follow-up PR wires these into the kipptaf
network layer, and it cannot start until these six models have materialized in prod.

The six models:

| Model                          | Grain                                             |
| ------------------------------ | ------------------------------------------------- |
| `int_focus__attendance_daily`  | one row per enrolled student per in-session day   |
| `int_focus__ada`               | one row per student per academic year             |
| `int_focus__attendance_streak` | one row per contiguous attendance streak          |
| `int_focus__calendar_day`      | one row per school per in-session day             |
| `int_focus__calendar_week`     | one row per school per week                       |
| `int_focus__calendar_rollup`   | one row per school per academic year              |

## Reviewer Notes

**These models speak Focus, not PowerSchool.** Column names come from Focus's own
source columns — `school_date`, `daily_code`, `state_value`, `academic_year`,
`startdate` — matching what the sibling `int_focus__student_enrollment` already does.
There is deliberately no `studentid`, no `yearid`, no `att_code`, and no
constant-NULL PowerSchool scaffolding. Every translation into the network's
PowerSchool-derived shape happens in the kipptaf blended-source layer in the
follow-up PR, which is the layer whose job that is. An earlier revision of this PR
had it the other way round; see the fold-out.

**This PR is not validated by CI, and the green checks do not mean what they usually
mean.** The dbt Cloud CI job builds the `kipptaf` project only, so a change confined
to the `focus` package selects zero models and passes in about 30 seconds without
building anything. The real validation was a local `dbt build` of the whole
six-model subgraph: 6 models, 11 data tests, 1 unit test, all passing. Please don't
read the dbt Cloud check as evidence.

**A day nobody took attendance for is distinguished from a recorded presence.**
`daily_code` is NULL for both, and `is_attendance_recorded` separates them. That
distinction is the one thing Focus can express and PowerSchool cannot.

**Every data test in this package can only warn, never fail a build.** Pre-existing
repo behavior, not introduced here, and it is why the follow-up PR carries the real
enforcement.

**Two of Miami's Focus schools have misconfigured calendars.** These models surface
those rows rather than hiding them, deliberately — the fix belongs in Focus
configuration and is already with Ops.

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Verdicts on all
      six findings are posted as a PR comment — three fixed, three declined with
      reasons.

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — all six have
      one, with model and column descriptions.
      analysis, or application — not applicable; internal-only intermediates with no
      consumer yet.
- [x] **Breaking change?** No. Every model is new and nothing existing reads them.
      Rollback is reverting the PR.
- [x] If adding or modifying an external source, run `stage_external_sources` with
      **`--target staging`** — not applicable; no external sources added or changed.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes, but see Reviewer Notes: it validates nothing here.
- [ ] **Dagster Cloud** — not triggered; no Dagster paths change.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed, executed task-by-task against a written plan with a
fresh implementer and an independent reviewer per task. Design spec and
implementation plan are both in this PR under `docs/superpowers/`.

### Why the package carries no PowerSchool vocabulary

An earlier revision of this PR pushed `studentid`, `yearid`, the PowerSchool
attendance-code vocabulary (`U` mapped to `A`), and constant-NULL `fteid`,
`attendance_conversion_id`, `ontrack`, `offtrack`, `student_track` and `track` columns
down into these models, so the downstream union could be a trivial `select *`. That
was the wrong trade — it left a source-system package unreadable without PowerSchool
knowledge, and put conform logic in the one layer with no business knowing the target
shape. The refactor is projection-only: every row count is byte-identical before and
after (307,638 daily rows for 2026; 1,628 ADA rows; 4,474 streaks; 13,552 calendar
days; 2,571 weeks; 68 rollups).

The kipptaf `focus_conformed` CTE in the follow-up PR absorbs the whole translation:
the `U` to `A` mapping, `yearid = academic_year - 1990`, the `state_value` to
`attendancevalue` cast, the membership constants, and typed NULLs for the
PowerSchool-only columns. `int_students__*` models will dual-expose — system-agnostic
names as primary columns plus legacy PowerSchool-named aliases — so all 37 existing
consumer refs keep working and new work can adopt the neutral names without a
coordinated migration.

### The no-record encoding

`daily_code` NULL plus `is_attendance_recorded` false. `state_value` coalesces to 1,
matching how PowerSchool resolves a day with no absence record to `Present`.

An earlier revision mapped no-record onto PowerSchool's `M` code, reasoning that `M`
means Missing Attendance and was in-vocabulary. That was wrong: PowerSchool's `M` is a
rare, explicitly entered code averaging 0.741 `attendancevalue`, and
`rpt_gsheets__absence_streak_roster` filters `att_code in ('A', 'AD', 'M')`, counting
it as an absence. 1,127,834 Focus present-by-default rows carried `M`, so it would have
published fake multi-month Miami absence streaks. Under the Focus-native design the
question does not arise in this package at all.

Focus's no-record rate is material: 119 of 7,773 completed-day rows for 2026, 1.5%.
The kipptaf union leaves `is_attendance_recorded` NULL on PowerSchool rows rather than
true, since PowerSchool records only absences and cannot express the distinction, so a
Focus-versus-NJ comparison cannot read PowerSchool as fully compliant.

### Surrogate keys hash student_number

`int_focus__attendance_daily` has no `studentid` at all now, and Phase 1 left it NULL
for Focus in `int_students__student_enrollment_union` (all 10,047 Miami rows).
`dbt_utils.generate_surrogate_key` maps NULL to a single constant, so hashing it would
collapse every Focus student into one streak per year and code — silently, while still
building green. Every hash input and window partition uses `student_number`. Verified:
4,474 streak rows, 4,474 distinct `streak_id`.

Because `studentid` is NULL network-side, the follow-up PR must repoint two existing
joins from `studentid` to `student_number`. They are 1:1 in every NJ region (Newark
18,148, Camden 5,510, Paterson 1,051 — identical distinct and pair counts), so that
repoint cannot fan out or drop an NJ row.

`int_focus__attendance_streak` also replaces the district model's single overloaded
`att_code` column with `streak_type` (`'daily_code'` or `'state_value'`) plus
`streak_value`. The district model uses one column for both union branches with nothing
to tell them apart; a code review flagged that latent ambiguity. The conform layer
reassembles the old shape.

### The transfer-day trim in int_focus__attendance_daily

Focus dates a school transfer with the departing stint's `exitdate` EQUAL to the
arriving stint's `startdate`, so an inclusive range counts that day twice and breaks
the `(student_number, school_date)` grain. The model trims only that day, via a
`stint_starts` semi-join guarded by `e.startdate <> s.startdate`.

Both halves matter. A blanket half-open range would drop 1,755 stints that legitimately
end on an in-session day the student attended, to fix 78 transfer boundaries. And
without the self-match guard, a single-day stint (`startdate = exitdate`) matches its
own row, gets `exitdate = startdate - 1`, and loses its only membership day — 73 stints
in 2026 are single-day and all fall on in-session days.

### distinct is grain projection, not dup-masking

Both `int_focus__attendance_daily` and `int_focus__calendar_day` select `distinct` from
`stg_focus__attendance_calendar`, which holds 15,067 rows against 13,512 distinct
`(school_id, syear, school_date)` keys. Including `minutes` still yields 13,512, so the
duplication is total and the collapse is lossless. A Focus-side data-quality issue, not
one introduced here. The `stint_starts` `DISTINCT` is annotated as defensive rather than
collapsing — it collapses nothing today, but becomes load-bearing if the upstream dedupe
in `int_focus__student_enrollment` is removed per its own TODO (#4905).

### Test severity ceiling

`src/dbt/kippmiami/dbt_project.yml` sets an unscoped `data_tests: +severity: warn`, and
dbt lets a root project override configs on resources defined in an installed package.
Every test here resolves to `warn` despite declaring `error` — confirmed with `dbt ls`.
The declarations stay because they express correct intent and would bind if that project
config were ever scoped. Real enforcement lands in the follow-up PR at the kipptaf level,
where kipptaf is the root project, and that PR carries an explicit step verifying its
grain test resolves to `error`.

### Calendar grain includes academic_year

`int_focus__calendar_day`'s uniqueness test is `(schoolid, academic_year, school_date)`.
Focus reuses a calendar date across school years for the same school: 13,552 rows carry
only 12,484 distinct `(schoolid, school_date)` pairs, while all 13,552 are distinct once
the year is added.

### Known cosmetic item, deliberately not fixed

`int_focus__calendar_week.date_count` uses `count()` rather than `count(distinct)`. Focus
quarter marking periods overlap in pre-2020 years only — 4 days in 2019, 2 in 2018, 1 in
2015 — so those weeks inflate slightly. 2026 and every year back to 2020 are clean, and
the district model has the same non-distinct count.

### Ops handoff

Five of Miami's ten Focus schools carry a 212-day 2026 attendance calendar including
Labor Day, Thanksgiving, Christmas, New Year's Day, and the whole winter break. Three are
non-instructional and have no locations-sheet row, so the network crosswalk drops them.
The two closed schools (Sunrise 2008Z, Liberty 2332D) do map to live location keys. Handed
to Ops as an Asana task; these models surface the rows rather than filtering them, and the
follow-up PR adds a warn-severity test that makes them visible.

</details>

Refs #4924
Refs #4927


cbini:
Verdicts on all six findings, verified against the code before replying. Three fixed in `e700a1c7d0`, three declined with reasons.

**#1 ST06 column ordering in `int_focus__attendance_streak.sql` — declined, false positive.**

You flagged this as "likely a Trunk/CI blocker" while disclosing you had no trunk access. It isn't. `trunk check --force --no-fix` on that exact file returns `No issues`, and the same check over all 15 changed files returns `No issues`. The Trunk Check CI run on this PR also passed. Nothing to fix.

**#2 `not_null` on `generate_surrogate_key` output — fixed.**

Correct, and it violates a written rule I missed: `src/dbt/CLAUDE.md:1008-1009` says never add `not_null` to `generate_surrogate_key` output because it cannot fail. Removed from the `streak_id` column. The `not_null` on `student_number` stays — that is a real column and the test can fail.

**#3 Uniqueness weaker than the mirrored model — fixed.**

Correct. Replaced the `unique_combination_of_columns(streak_id, att_code)` with a plain `unique` on `streak_id`, matching `int_powerschool__attendance_streak.yml`. Your reasoning holds: the hash carries a literal branch discriminator (`'code'` / `'att'`), so `streak_id` is globally unique on its own, and testing the pair would only catch a cross-family collision if `att_code` also matched.

**#4 Unannotated `DISTINCT` in `stint_starts` — fixed, but kept rather than dropped.**

Your analysis is right that it collapses nothing today: `int_focus__student_enrollment` dedupes on exactly `(student_number, academic_year, startdate)` at line 113, the same three columns selected here. I annotated it instead of removing it, because that upstream dedupe carries a TODO to be removed once Focus stops accepting duplicate open stints (#4905) — and without the `DISTINCT`, duplicate rows here would fan out the `enrollments` semi-join. The comment now says it is defensive rather than collapsing, so the next reader knows which it is.

**#5 Recomputed `yearid + 1990` in a JOIN predicate — declined.**

Your own grep found the identical shape in at least three production files, and Trunk is clean. Hoisting it would add a CTE for no functional change. Noting it as a fair DRY observation, not acting on it.

**#6 ST06 tier ordering in `int_focus__attendance_daily.sql` — declined.**

You flagged this low-confidence and asked for a local Trunk check to settle it. Settled: clean.

---

Separately, and unrelated to your findings: this PR is now a draft. A design change came in after you reviewed — the Focus package should not carry PowerSchool-compatibility scaffolding at all (`studentid`, `yearid`, the null `fteid` / `attendance_conversion_id` / track columns, and the PowerSchool attendance-code vocabulary). Those exist in this diff only to make the downstream union cheap, which is the wrong trade: the conforming belongs in the kipptaf blended-source layer, which is where a system-agnostic model and any transitional legacy aliases should live. I am reworking the six models to be Focus-native before this merges, so several of the columns you reviewed will disappear.

cbini:
Verdicts on the four findings from this round. Two fixed in `ca0d6d1776`, two declined.

**#1 No test covers the transfer-day trim or the single-day-stint guard — fixed, and this was the most useful finding either round.**

You were right that the existing test never exercised `stint_starts`: every fixture student had a single stint with identical `startdate` and `exitdate`. Worse than you knew — both of those code paths have already been wrong once during development. The guard's absence silently deleted 63 real attendance days, and an earlier blanket half-open range would have dropped 1,755 legitimately-attended final days. Exactly the regression class you described, already realised once.

Added `test_int_focus__attendance_daily_transfer_day_and_single_day_stint` with three fixtures: a student transferring 58 to 59 where the departing stint's `exitdate` equals the arriving stint's `startdate` (asserts the shared day appears once, attributed to the ARRIVING school), a single-day stint (asserts its only day survives), and a normal multi-day stint whose `exitdate` no other stint starts on (asserts it is NOT trimmed).

I did not accept a green run as evidence the assertions bind. Both failure modes were proven by deliberately breaking the model:

- guard removed → FAILS, naming the single-day student's lost row
- trim disabled → FAILS, naming the transfer day duplicated at the departing school

Neither broken version was committed.

**#2 `int_focus__calendar_day.yml` column sort order — fixed.**

Correct, and correct that it survived the rename from `date_value`. `school_date` now sits above `schoolid` so the tested column comes first.

**#3 ST06 literal interleaving in `int_focus__attendance_streak.sql` — declined.**

Settled the way you suggested: `trunk check --force` over all 59 files in the package returns `No issues`, and Trunk Check passes on this PR. Credit for flagging your own prior false positive on this file and framing this as "worth a local trunk check" rather than a blocker — that framing is what made it cheap to resolve.

**#4 Pass-through `attendance` CTE — declined.**

Soft match, as you noted. The rule's example is a `select *` alias; this narrows nine columns to five, which is closer to shaping than aliasing, and it keeps the final `left join` readable against a three-key condition. Not worth the churn on a model this heavily reviewed.

PR #4940, opened 2026-08-20
fix(overgrad): declare university_type and public_school on University

### LINKED ISSUE #4939: fix(overgrad): declare university_type and public_school on the University schema
## What's happening

Overgrad's universities endpoint started returning two fields we do not declare: `university_type` and `public_school`. Because the `University` pydantic model drives the generated Avro schema, `fastavro.writer` silently drops both, and the `avro_schema_valid` asset check warns on every materialization of `kipptaf/overgrad/universities`.

The check is WARN severity, so runs still succeed — but each one fires a Dagster+ alert, and the two fields never reach the warehouse.

First failing execution was run `0ed20d84-6972-4b93-86c8-8f9d2c25b459` (approx. 2026-08-09). The four executions before it all passed with empty `extras`. No repo change is involved; the API changed.

This is the same class of drift fixed for the students endpoint in 5b733b4ddc.

## Steps to reproduce

1. Open any recent materialization of `kipptaf/overgrad/universities`, e.g. run `ee8cf7e5-f411-41bd-ad13-0cddc7ac1447`.
2. The run reports SUCCESS, but the `avro_schema_valid` check evaluation reports `passed: false` with metadata `extras: "university_type, public_school"`.
3. Query `kipptaf_overgrad.stg_overgrad__universities` and confirm neither column exists.

## Where

- **Code location / dbt project:** `kipptaf` (Dagster) and `src/dbt/kipptaf` (dbt)
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** https://kipptaf.dagster.cloud/prod/runs/ee8cf7e5-f411-41bd-ad13-0cddc7ac1447

<details>
<summary>For Claude</summary>

Root cause: `University` in `src/teamster/libraries/overgrad/schema.py` is missing the two fields. `check_avro_schema_valid` in `src/teamster/core/asset_checks.py` computes `record_fields - schema_fields`, so any undeclared key surfaces as an extra.

Types verified against the live API across a random sample of 40 universities:

```text
university_type: str | None   -- four_year, two_year, less_than_two_year, null
public_school:   bool | None  -- true, false, null
```

Both keys are present on all 40 records, so the addition is permanent rather than sporadic.

Fix touches two files, and both are required in the same change:

1. `src/teamster/libraries/overgrad/schema.py` -- declare both fields on `University`.
2. `src/dbt/kipptaf/models/overgrad/staging/properties/stg_overgrad__universities.yml` -- the model is a bare `select *` under an enforced contract, so the two new external-table columns break the contract unless they are declared here too (`university_type` string, `public_school` bool).

Fixed means a materialization of `kipptaf/overgrad/universities` reports `avro_schema_valid` passing with empty `extras`, and `stg_overgrad__universities` builds with both new columns populated.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop the alert that fires on every
`kipptaf/overgrad/universities` materialization, and start capturing two fields
Overgrad added to its universities endpoint: `university_type` and
`public_school`.

Overgrad began returning both fields around 2026-08-09. We never declared them,
so the Avro writer silently dropped them and the `avro_schema_valid` check
warned on every run. The check is WARN severity, so runs stayed green while
firing an alert each time. This is the same drift that hit the students endpoint
in 5b733b4dd.

Refs #4939

## Reviewer Notes

- **Field types were checked against the live API, not guessed.** A random
  sample of 40 universities gives `university_type` as a string enum
  (`four_year`, `two_year`, `less_than_two_year`) and `public_school` as a bool.
  Both keys appear on all 40 records, sometimes null.
- **The dbt change is not optional.** `stg_overgrad__universities` is a bare
  `select *` under an enforced contract, so two new external-table columns break
  it unless they are declared in the properties file in the same change.
- **The 2,617 universities already in the warehouse will read null for both new
  fields until they are backfilled.** The asset uses
  `AutomationCondition.newly_missing()`, so existing partitions never
  re-materialize on their own. See the rollout note below.

## Rollout

After merge, back-fill all partitions of `kipptaf/overgrad/universities` from
the Dagster UI. Until that finishes, only newly-added universities carry the two
fields. Nothing breaks in the meantime, because BigQuery unions the Avro schemas
across the wildcard rather than reading one representative file.

### General

### Dagster

- [x] Ran `uv run pytest tests/assets/test_assets_overgrad.py::test_universities_kipptaf`
      — passes, and that test asserts the check's `extras` metadata is empty.

### dbt

- [x] Properties file updated for the two new contract columns, with
      descriptions.
- [x] No columns renamed or removed, so no downstream exposure breakage.
- [x] Ran `stage_external_sources --target staging` against
      `gs://teamster-test/dagster/kipptaf` so dbt Cloud CI sees the new schema
      before the prod Avro files are updated. Re-stage to the prod location
      after merge, once the backfill lands.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Investigated with
`superpowers:systematic-debugging` from run
`ee8cf7e5-f411-41bd-ad13-0cddc7ac1447`.

**Root cause.** `University` in `src/teamster/libraries/overgrad/schema.py`
drives the generated Avro schema. `check_avro_schema_valid` in
`src/teamster/core/asset_checks.py` computes `record_fields - schema_fields`, so
any undeclared key surfaces as an extra. Check history: first failure was run
`0ed20d84-6972-4b93-86c8-8f9d2c25b459`; the four executions before it passed
with empty `extras`. No repo commit is implicated — the API changed.

**Type evidence.** A throwaway pytest probe fetched 40 randomly-sampled
university ids through `OvergradResource` (the conftest bootstraps credentials).
Observed Python types were `str`/`NoneType` for `university_type` and
`bool`/`NoneType` for `public_school`, with distinct values `four_year`,
`two_year`, `less_than_two_year` and `true`/`false`. The probe was deleted after
the run.

**Mixed-schema behavior, verified rather than assumed.** The concern was that
2,617 existing per-id Avro files keep the old 7-field schema while new
partitions write 9, and Google's documentation does not state how an external
Avro table resolves that. Tested it directly: wrote one old-schema file and one
new-schema file to the test bucket, staged an external table over just those
two, and read it back. BigQuery **unions** the fields — all nine columns
present, with the old file's row reading null for both new fields. So the
contract holds as soon as the first new-schema partition lands, and the backfill
is a data-completeness follow-up rather than a merge blocker.

**Column order.** The properties file lists the two new columns in the same
positions the pydantic model emits them (alphabetical within `University`).
Contract enforcement matches on name plus type rather than order, so this is
cosmetic consistency, not a requirement.

</details>

### DIFF
--- a/src/dbt/kipptaf/models/overgrad/staging/properties/stg_overgrad__universities.yml
+++ b/src/dbt/kipptaf/models/overgrad/staging/properties/stg_overgrad__universities.yml
@@ -19,7 +19,17 @@ models:
         data_type: string
       - name: object
         data_type: string
+      - name: public_school
+        description: >-
+          Whether Overgrad classifies the institution as publicly controlled.
+          Null where Overgrad has not classified it.
+        data_type: boolean
       - name: state
         data_type: string
       - name: status
+        data_type: string
+      - name: university_type
+        description: >-
+          Overgrad's institution length category: `four_year`, `two_year`, or
+          `less_than_two_year`. Null where Overgrad has not categorized it.
         data_type: string
--- a/src/teamster/libraries/overgrad/schema.py
+++ b/src/teamster/libraries/overgrad/schema.py
@@ -121,8 +121,10 @@ class University(UniversityID):
     city: str | None = None
     name: str | None = None
     object: str | None = None
+    public_school: bool | None = None
     state: str | None = None
     status: str | None = None
+    university_type: str | None = None
 
 
 class DueDate(BaseModel):


cbini:
## Review verdicts

All three notes are non-blocking; declining all three, with reasons.

**1. `accepted_values` test on `university_type` — declined.**

Two reasons, one of which the review reached itself:

- Precedent: `grep -c accepted_values` returns 0 across all ten overgrad staging
  properties files. Adding it here would make this model the sole exception, not
  bring it into line.
- Sample size: the three observed values come from 40 randomly-sampled
  universities out of 2,617, so roughly 1.5% of the population. The column does
  not exist in the warehouse yet, so there is no way to enumerate the full value
  set until the backfill lands. An `accepted_values` test written from a 1.5%
  sample is likely to fail on the other 98.5%, and a test that fires on
  legitimate data is worse than no test.

Worth revisiting after the backfill, when the full distribution is queryable.
Filing that is not worth an issue until the data exists.

**2. Missing model-level and existing-column descriptions — declined, agreed
non-issue.** The review verified this itself against `5b733b4dd`, which likewise
documented only the newly-declared columns. Retro-documenting the six
pre-existing columns is unrelated to this fix and belongs in its own change.

**3. `public_school` not prefixed `is_` — declined, agreed non-issue.** Also
self-verified. Confirmed independently: every raw boolean in this model family
keeps the source's native name (`fafsa_completed`, `first_generation`,
`completed_account_setup`, `waitlisted`, `deferred`). Renaming one of them would
break the passthrough convention rather than follow it.

## On the claims the review could not re-derive

The review correctly flagged that it has no warehouse or API access and took the
data-dependent claims as given. For the record, both were tested rather than
reasoned about:

- Field types: 40 randomly-sampled university ids pulled live through
  `OvergradResource`.
- BigQuery's cross-file Avro behavior: an external table staged over exactly one
  old-schema file and one new-schema file, then read back. It exposed all nine
  columns, with the old file's row reading null for both new fields.

PR #4945, opened 2026-08-21
fix(dbt): correct the gradebook audit exposure targets

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will point the two gradebook audit exposures at
the right places. Two independent one-line corrections, found while reviewing an
unrelated exposure.

### rpt_gsheets__gradebook_audit_student_flags

The exposure pointed at `1e52D42Zhlhasj…`, titled "Gradebook - Student Flags",
which lives in the **Reports** folder — that is the user-facing IMPORTRANGE copy.
Repointed to `1d9hApv…`, titled `rpt_gsheets__gradebook_audit_student_flags`, in
the **IMPORTRANGE Sources** folder, which is the sheet the model actually loads
into.

Per the `rpt_gsheets__` convention the exposure carries the source sheet, not the
IMPORTRANGE copy. The sibling `rpt_gsheets__gradebook_audit_template` exposure
was already correct, which is what made the mismatch visible.

### rpt_gsheets__gradebook_audit_template

Added the missing `ref("stg_powerschool__u_expectations")`. That sheet reads
current expectations from the staging model alongside the template rows, so the
exposure was declaring only half its lineage.

## AI Assistance

Claude found both mismatches by checking which Drive folder each exposure URL
actually resolves to, and applied the corrections. The convention itself —
source sheet in IMPORTRANGE Sources, user-friendly copy in Reports, exposure
carries the source — was human-supplied, as was the confirmation of which refs
each sheet should declare.

### dbt

- [x] Exposure-only change. No models, tests, or contracts touched
- [x] `ref("stg_powerschool__u_expectations")` resolves — the model exists at
      `src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__u_expectations.sql`
- [x] Not a breaking change — no columns renamed or removed
- [x] `trunk check --force` clean

### Notes for the reviewer

Both exposures carry `owner.name: Data Team - grangel` where the convention in
`src/dbt/kipptaf/CLAUDE.md` is plain `Data Team`. Left alone deliberately to keep
this diff to the two reported problems; happy to fold it in if wanted.

## CI checks

- [ ] **Trunk**
- [ ] **dbt Cloud**
- [ ] **Dagster Cloud** — not expected to trigger, no Dagster paths changed


GabyRangelB:
## Review verdicts

Verified each finding against the code before replying. Three need no action; one is declined with reasoning.

### 1. Missing `int_powerschool__calendar_week` in `depends_on` — declined

Both citations are accurate — `rpt_gsheets__gradebook_audit_template.sql` refs `int_powerschool__calendar_week` at line 19 and `stg_powerschool__u_expectations` at line 49. Declining on two independent grounds.

**`depends_on` on these sheet exposures declares the sheet's own BigQuery connections — one entry per connected tab — not the exposed model's upstream lineage.** Of the 66 exposures in this file, 6 declare more than one ref, and in none of them is one ref an upstream of another ref in the same exposure. The two closest analogues:

| Exposure | Refs | One an upstream of the other? |
| --- | --- | --- |
| `tech_data_feeds` | `rpt_idauto__staff_roster` + `stg_ldap__user_person` | No — that rpt refs only `int_people__staff_roster` |
| `che_camden_workspace_feed` | `int_powerschool__ps_adaadm_daily_ctod` + `base_powerschool__student_enrollments` | No — that int model does not ref the base model |

`tech_data_feeds` is structurally identical to this case (a `rpt_` plus a `stg_`), and its two refs are independent tabs. This sheet has exactly two BigQuery connections, which is why exactly two refs are declared. `int_powerschool__calendar_week` is not one of them, so declaring it would assert a third connected tab that does not exist.

That the template model happens to ref `u_expectations` directly is incidental — the reason it belongs here is that the sheet reads it, not that the model does.

**The stated failure scenario does not occur.** "A reviewer relying on the exposure's `depends_on` for impact analysis would miss this consumer" is already false:

```text
$ dbt ls --select int_powerschool__calendar_week+ --resource-type exposure --project-dir src/dbt/kipptaf
...
exposure:kipptaf.gradebook_audit
exposure:kipptaf.rpt_gsheets__gradebook_audit_student_flags
exposure:kipptaf.rpt_gsheets__gradebook_audit_template
```

The exposure is already reachable downstream of `int_powerschool__calendar_week` through the rpt model. Adding the ref would duplicate an edge the graph already has, while falsely implying a direct sheet connection.

### 2. Sheet-ID correctness — verified, no change needed

Resolved via the Drive API, which is available locally even though the CI job has no Drive access:

| ID | Title | Parent folder |
| --- | --- | --- |
| `1d9hApv…` (new) | `rpt_gsheets__gradebook_audit_student_flags` | `1kwng_xGxQzIRNTVueiJqa_VHsd1tSQgO` — IMPORTRANGE Sources |
| `1e52D42…` (old) | `Gradebook - Student Flags` | `17PG7aMo6f3JxqtSvx7bm0w7FESf0Qbpy` — Reports |

This matches the PR description exactly: the old value pointed at the user-facing IMPORTRANGE copy in Reports; the new one is the source sheet in IMPORTRANGE Sources, which is what the model loads into. The correction stands as written.

The severity read is also right — `url` feeds `AssetSpec(metadata={"url": ...})` only, so this was documentation/lineage metadata, never a write target.

### 3. `owner.name: Data Team - grangel` — no action

Confirmed pre-existing and confirmed the count: 10 exposures in this file carry `Data Team - grangel`, so 8 besides the two this PR touches, against 61 using plain `Data Team`. Out of scope for a two-line fix, as flagged in the PR description. Left as-is.

### 4. Reuse / simplification / dead code — no action

No disagreement.

GabyRangelB:
@claude - we updated. please review

src/dbt/kipptaf/models/exposures/google-sheets.yml
```diff
@@ -814,6 +814,7 @@ exposures:
     url: https://docs.google.com/spreadsheets/d/1ofCxW0pLniywn_XZT69S23vhcDs6y9ElAtJa5fTtiT0
     depends_on:
       - ref("rpt_gsheets__gradebook_audit_template")
+      - ref("stg_powerschool__u_expectations")
```
cbini: @GabyRangelB this needs a rpt_ view buffer between the tool and the source table. even if it's a simple passthrough.

cbini (changes requested):

src/dbt/kipptaf/models/exposures/google-sheets.yml
GabyRangelB: Agreed, fixed in 53f511f. Added `rpt_gsheets__gradebook_audit_current_expectations` as a passthrough view over `stg_powerschool__u_expectations` (contract enforced, `unique`/`not_null` on `id`), and repointed the exposure's `depends_on` at it instead of the staging model directly. Gaby will reconnect the sheet's current-expectations tab to the new view.

cbini (approved):

PR #4961, opened 2026-08-24
feat(kipptaf): put Miami Focus attendance into the network layer

### LINKED ISSUE #4924: fix(dbt): wire Focus attendance into the kipptaf network layer for Miami AY2026

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will put Miami's attendance data into the network
layer for the first time.

Miami moved from PowerSchool to Focus at the start of AY2026. The Focus data
reached the warehouse, but no network model read it. Miami contributed 0
attendance rows to `kipptaf`, so network average daily attendance, chronic
absenteeism, and enrollment totals all understated the network.

This pull request adds 6 system-neutral `int_students__*` models. Each one
unions the frozen PowerSchool archive with live Focus data, and splits Miami on
one shared cutover year. It then repoints about 40 consumer references onto
those models.

Miami's rows in each surface, before and after:

| Surface                                  | Before | After                    |
| ---------------------------------------- | ------ | ------------------------ |
| `rpt_tableau__attendance_dashboard`      | 0      | 239,157 rows             |
| `int_topline__ada_running_weekly`        | 0      | 121,083 populated        |
| `rpt_gsheets__kippfwd_miami_roster`      | 0      | 6,709 students with ADA  |
| `rpt_tableau__attendance_chronic_absenteeism_log` | 0 | 374 students / 743 absences, but see below |
| `fct_student_attendance_streaks`         | 0      | all years, AY2018 to AY2026 |

New Jersey numbers do not move. Every model gates on that.

Two rows in that table are latent, not live.
`rpt_tableau__attendance_chronic_absenteeism_log` and `int_students__fldoe_fte`
both carry `enabled: false` and predate this pull request. Their fixes are real
and verified, but neither model builds until someone re-enables it. Re-enabling
is a separate decision. Every other row in the table is live on merge.

Closes #4924

## Reviewer Notes

**dbt Cloud is red on purpose. Do not read it as a broken branch.** One Miami
student has a Focus enrollment stint that fully contains another stint for the
same year. Both stints match the same 38 attendance streaks, so
`unique_fct_student_attendance_streaks_student_attendance_streak_key` fails with
38 results. Charlie chose to correct the Focus record at source rather than work
around it in a model. The test stays at `severity: error` until someone fixes
that record. It is the only failing test.

**The cutover year comes from data, not a hardcoded 2026.**
`int_students__sis_cutover` returns `min(academic_year)` over rows that carry
recorded attendance. It reads 2026 today. Deriving it from row presence would
have been wrong: `int_focus__attendance_daily` holds rows back to AY2020, and
827,851 of them are present-by-default scaffold with no recorded attendance.
Using presence would have replaced 6 years of real Miami PowerSchool attendance
with fabricated perfect attendance.

**One bug shape caused most of the work here.** Focus has no `studentid`, and
Miami rows carry `NULL` for it. Any join, `group by`, or `partition by` keyed on
`studentid` silently drops or collapses every Miami row. This shape appeared 8
times. The worst case was `int_students__fldoe_fte`, a Florida state reporting
model, which collapsed about 1,600 Miami students into 1 row and took `max()`
across the whole district. Every site now keys on `student_number`.

**`int_powerschool__ps_adaadm_daily_ctod` is now a thin union.** About 200 lines
of derived calculations moved to `int_students__attendance_daily`. Every
in-repo consumer is repointed. Anything outside this repo that reads a derived
column from `int_powerschool__ps_adaadm_daily_ctod` needs to move to
`int_students__attendance_daily`. The moved calculations are byte-identical, and
26 rows of aggregate parity confirm that.

**`int_students__calendar_day` nearly narrowed an existing dimension.** Its
PowerSchool branch joins `stg_powerschool__terms` to source `yearid`. As an
`INNER JOIN` it dropped 423 New Jersey school-days that no year record covers,
mostly August pre-service dates. `dim_school_calendars` used to hold those days.
It is a `LEFT JOIN` now, with a null-safe cutover predicate.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — every
      new model has one, including the 6 `int_focus__*` wrappers.
- [ ] Include (or update) an
- [x] **Breaking change?** Yes. `int_powerschool__ps_adaadm_daily_ctod` no longer
      emits its derived columns. All 11 in-repo consumers are repointed to
      `int_students__attendance_daily`. Rollback is a revert of this branch: no
      model is dropped and no column is renamed in a contracted model.
      with **`--target staging`** — not applicable, no external sources change.

## CI checks

- [x] **Trunk** — passes. `trunk check --force` run over all 60 changed files.
- [ ] **dbt Cloud** — fails, on purpose. See the first Reviewer Note.
- [x] **Dagster Cloud** — not triggered, no Dagster paths change.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Executed with subagent-driven development: a
fresh implementer per task, a spec-and-quality review after each, and a
whole-branch review at the end.

**Cutover mechanics.** `int_students__sis_cutover` is one row, one column,
`focus_start_academic_year`. It is a floor, not a year set: an `in (select ...)`
set can punch a hole mid-history if a Focus year records no attendance
exceptions, and that year would then fall back to a PowerSchool archive holding
nothing for it. Both branches of every conform are floored against it. The
PowerSchool side excludes `_dbt_source_project = 'kippmiami'` at or above the
floor; the Focus side admits only rows at or above it. Omitting the Focus-side
floor lets Focus's pre-2026 scaffold rows land beside PowerSchool's real rows
and break each model's own grain test. The model carries a `not_null` test at
`severity: error` and that test must never be downgraded: a `NULL` cutover makes
every Miami predicate evaluate to `NULL` and erases Miami from all 6 models.

**The 8 null-key sites.** `int_powerschool__ada_term` (`group by`, plus 7 window
`partition by`), `int_students__fldoe_fte` (`group by`),
`rpt_tableau__attendance_dashboard` (join), `rpt_tableau__ops_dashboard`
(`group by` and join), `int_topline__ada_running_weekly` (`group by`, join, and
4 `partition by`), `int_reporting__promotional_status` (2 sites),
`fct_student_attendance_streaks` (`yearid` join), and the 4
`int_powerschool__ada_term_pivot` consumer joins. `student_number` to
`studentid` is strictly 1:1 within every project, verified, so the
`int_powerschool__ada_term` re-key keeps `studentid` as `max(studentid)` and
every existing consumer join keeps working unchanged.

**PowerSchool duplicate rows.** `int_students__attendance_daily` dedupes its
PowerSchool branch. Prod holds 561 duplicate `(project, studentid, calendardate)`
keys and 1,301 excess rows, of which 18 are genuine conflicts with differing
values. 558 of the keys are Newark, spanning 2026-08-19 to 2027-06-17, so this
is an ongoing double-write and not a closed historical defect. The dedupe is
`dbt_utils.deduplicate` per `src/dbt/CLAUDE.md`, and it carries a `TODO` naming
the upstream fix. The dev relation holds zero duplicates, so this only bites in
CI and prod.

**Two models in this diff are disabled.**
`rpt_tableau__attendance_chronic_absenteeism_log` has `enabled: false` and could
not compile at all after PR #3142, which referenced `ac._dbt_source_project`
without projecting it. This branch fixes that and proves the model compiles.
`int_students__fldoe_fte` is also `enabled: false`. Both rewrites are correct
but deliver nothing until someone re-enables them. Re-enabling is a separate
decision.

**Follow-ups, ranked.**

1. `int_students__student_enrollment_union` and `int_students__student_enrollments`
   both emit `yearid = NULL` for 100% of Miami rows, while `academic_year` is
   populated. That one defect caused 4 of the silent-drop bugs above. The fix
   pattern already exists in `int_students__terms.sql`:
   `yearid = academic_year - 1990`. It does not subsume the `studentid` repoints
   in this branch, because Focus has no `studentid` at all. Its blast radius is
   every consumer of 2 shared models, so it needs its own pull request.
2. Re-enable the 2 disabled models named above, or retire them.
3. `int_students__attendance_interventions` still joins
   `int_powerschool__spenrollments` on `studentid`. That source has no
   `student_number`, so there is no key to switch to. It is a `LEFT JOIN`, so no
   row drops; Miami's `is_ca_exception` stays false. It needs a data decision.

**Verification notes for a future `@claude`.** Gate a live-year model on an
invariant, never on an absolute row count: AY2026 is in session and prod Focus
attendance moved 3 times inside one working session. Compare a parity gate
against the raw source, never against a restatement of the model's own logic:
the first `int_students__calendar_day` gate embedded the same `INNER JOIN` it
should have tested, so it passed while dropping 423 school-days. A `--state`
manifest compiled from this branch contains the branch's own unmerged models
pointed at prod tables that do not exist, so `--favor-state` returns 404; compile
the manifest from `main` instead.

</details>


cbini:
Verdict per finding. Fixed in `5c450946fe` unless noted.

**Fixed**

1. `int_topline__attendance_contacts.sql:77` — confirmed. `att.is_suspended = 0` in an `INNER JOIN` `ON` clause, and `is_suspended` is NULL for Focus rows, so every Miami row failed the join. Now null-safe. Good catch: this site was not on our list, so it is a genuine new find.
2. `rpt_tableau__attendance_dashboard.sql:115` — confirmed. The only 1 of 7 window functions in that select still keyed on `studentid`. Now `student_number, academic_year`, matching its 6 siblings.
3. Materialization regression on `int_students__ada`, `int_students__calendar_rollup`, `int_students__calendar_week` — confirmed. All three now set `materialized: table`, matching their PowerSchool predecessors. Verified `BASE TABLE` in `INFORMATION_SCHEMA.TABLES` and grain tests still pass. Your exclusion of `int_students__attendance_streak` was right — its predecessor was already a view, so it is not a regression.
4. PR body overstated a result — confirmed and corrected. The Before/After table now labels the 2 latent rows, and the summary says plainly that `rpt_tableau__attendance_chronic_absenteeism_log` and `int_students__fldoe_fte` both carry `enabled: false` and build nothing until someone re-enables them.
6. `int_students__fldoe_fte.sql:29` — confirmed. Now reads the pass-through `att._dbt_source_project` instead of re-deriving it with `extract_source_project`.
14. `fct_student_attendance_daily.yml` — confirmed, both halves. The `config.meta` for `attendance_code` / `attendance_value` / `membership_value` now names the real source, and `is_tardy` / `is_ontime` / `is_oss` / `is_iss` / `is_suspended` now carry the "Null for Focus-sourced Miami rows" caveat. The second half was the one that mattered: a `sum()` over those in Cube would have read Miami as a verified zero rather than unknown.

**Already answered by data**

15. Post-cutover Miami PowerSchool days with no covering term coexisting with a Focus row for the same date — checked, and it does not happen. For `kippmiami` the model returns 14,467 rows, 14,467 distinct `(schoolid, yearid, date_value)`, and 14,467 distinct `(schoolid, date_value)`. That third count is the one your finding predicts would be lower. All 3 are equal, so no physical Miami school-day appears twice.

**Declined, with reasons**

7. `rpt_tableau__nj_school_register.sql:20` grouping on `studentid` — declining. `where co.region != 'Miami'` excludes every Miami row from the output regardless of the grouping, so there is no defect to fix and a rekey would be churn in a file this PR otherwise only ref-swaps.
9. No uniqueness test on the 6 `int_focus__*` wrappers — declining. These are pure `union_relations` passthroughs, and the established precedent for that shape (`int_focus__users`, `int_focus__students`) omits it. Adding it here would diverge from 11 sibling files.
10. No `not_null` on `int_powerschool__ada_term.student_number` — declining for this PR. The old `studentid` column had the same gap, so this is not a regression, and the grain test already fails loudly if the key goes null.
11. No uniqueness test on `int_students__sis_cutover` — declining. The model is an unfiltered `min()` with no `GROUP BY`, so it structurally returns exactly 1 row and a uniqueness test could never fail. The `not_null` test at `severity: error` is the guard that matters here, and it is the one that catches the failure mode we care about: a NULL cutover erases Miami from all 6 conform models.
12. The repeated `cutover` CTE across 6 models — declining. The "no pass-through import CTE" rule targets CTEs that only rename a `ref()`. This one carries the comment explaining why the boundary is a floor derived from recorded attendance, which is the single least obvious decision in the PR. Inlining `cross join ref(...)` 6 times would save 18 lines and delete the explanation from every site.

**Acknowledged, not fixed here**

5. No tracking issue number against the deliberately-failing PK test — fair. The failure is a Focus source-data defect in 1 enrollment record, and the decision to fix it at source rather than downgrade the test to `warn` was deliberate. You are right that the repo precedent for this shape pairs the choice with a `TODO(#issue)`. Raising the tracking issue is with the repo owner.
8. No test would catch a revert of the `stg_powerschool__terms` join from `LEFT` back to `INNER` — correct, and the sharpest of the test-coverage points. A day-count parity check of the PowerSchool branch against `stg_powerschool__calendar_day` per school-year is the right test. Noting it rather than adding it, because this PR is already large and that test belongs with the `yearid` follow-up work.

Two caveats on the verification above. The BigQuery and dbt MCP servers are both disconnected right now, so the dbt Cloud run for `5c450946fe` has not been read — dbt Cloud is expected to fail on the 38-row streak test, but whether that is still the *only* failure is unconfirmed. Separately, findings 1 and 2 move no live row count today: Miami's AY2026 enrollment-weeks are gated behind a known pre-existing defect where both enrollment models emit `yearid = NULL` for 100% of Miami rows. Both fixes are still correct, and both would fail silently once that gating clears.

PR #4964, opened 2026-08-25
feat(dbt): retrofit DIBELS foundation goals for Bright Spots tracker

### LINKED ISSUE #4952: feat(dbt): DIBELS Bright Spots tracker against Foundation benchmark goals
## What and why

T&L asked for a DIBELS Bright Spots tracker: a view of how each school and grade
is performing against KIPP Foundation's benchmark goals, modeled on the existing
Bright Spots tracker on the DDI suite landing page. Nothing in the model supports
it today.

This was originally folded into #3834 (the DIBELS PM aimline migration). It is
split out here because it reads a different source at a different grain —
Foundation goal **rates** joined to benchmark composites, which
`rpt_gsheets__dibels_bm_goals_calculations` already does — and because it is not
blocked on the aimline work. The two can proceed in parallel.

## What the tiers actually measure

Worth stating plainly, because the Well Below tiers read as inverted until you
see it.

`stg_google_sheets__dibels_foundation_goals` holds two goal types per region,
grade and period, and each is a **rate**, not a score:

- `At/Above` — the share of students who should be At or Above Benchmark
- `Well Below` — the share of students expected to remain Well Below

For example, Camden kindergarten at MOY carries a Well Below goal of `0.23`, or
23% of students.

The Bright Spots tiers are expressed in percentage points against those rates.
For At/Above, landing above the goal is good. For Well Below, landing below the
goal is good — that means fewer students are struggling. Both directions are
correct as written. They only look backwards if the goal is read as a student
score rather than a rate.

## Tier definitions

From T&L's Bright Spot Brainstorm notes. Thresholds are in percentage points.

GK-5, and G6-8 at EOY — measured against the goal for that period:

| Tier | At/Above | Well Below |
| --- | --- | --- |
| Bright spot | 5 or more points over goal | 5 or more points below goal |
| On Track | 1-4 points above goal | 1-4 points below goal |
| In Range | 1-5 points below goal | 1-5 points above goal |
| Off Track | more than 5 points from goal | more than 5 points above goal |

G6-8 at MOY is the exception. It is measured against the **EOY** goal, not the
MOY goal, and uses a different spread:

| Tier | Both goal types |
| --- | --- |
| Bright spot | meeting or exceeding EOY goal |
| On Track | 1-3 points from EOY goal |
| In Range | 4-5 points from EOY goal |
| Off Track | more than 5 points from EOY goal |

Foundation does publish MOY goals for grades 6-8, so this is a deliberate choice
rather than a data gap: middle school drops PM rounds mid-year for test prep, so
those grades are tracked against EOY throughout. Worth encoding as a named rule,
since it otherwise reads as a bug.

## Reference implementation

T&L named the Bright Spots tracker on the DDI suite landing page as the thing to
copy. Its structure, read off the live view:

**This is the layout reference, not the tiering reference.** DIBELS keeps T&L's
thresholds as written above. Noting the difference so nobody later "corrects"
one to match the other: the DDI view uses plus or minus 10 points and calls the
bottom tier "Concern Area", where DIBELS uses plus or minus 5, calls it "Off
Track", and adds the G6-8 MOY variant. Same four-tier shape, different numbers
and names, deliberately.

**Grain.** One row per school, discipline, subject/grade and module code — that
is, per assessment.

**Columns.** Grade Goal, % Mastery (Overall), % Mastery for four numbered
buckets, % Mastery (Gen Ed), % Mastery (IEP), and % B/FB.

**Filters.** Term, Region, School, Discipline, Bright Spot Status as a
multi-select, and Module Type.

**Encoding.** Status is cell shading across four values: Bright Spot in dark
blue, On-Track in light blue, In Range in grey, Concern Area in orange.

Four consequences for the build:

1. **Every mastery cell is tiered independently.** Overall, each bucket, Gen Ed
   and IEP each get their own status against the same Grade Goal on that row —
   a single row routinely shows Bright Spot in one column and Concern Area in
   another. So status is computed per population, not once per row.
1. **That implies a long underlying model pivoted wide in the view.** One row
   per school, subject/grade, assessment and population, each carrying its own
   mastery rate and status, with Tableau pivoting population into columns.
1. **Bright Spot Status has to be a materialized column**, not a
   visualization-level calculation, because it is a multi-select filter.
1. **A % B/FB column sits alongside mastery** — the Below and Far Below share.
   It appears to be oriented inversely, with low values shaded favourably, which
   makes sense for a rate you want to reduce. The DIBELS analogue is the Below
   plus Well Below share, which is exactly what Foundation's Well Below goal
   type measures — so the existing two-goal-type structure already covers it.
   Confirm the inverse orientation.

## Demographic splits

All students and SPED. T&L's threshold definitions are identical for both, so
SPED does not need separate rules or a separate goal source — but per the
reference layout it does need to be a pivotable population dimension, since
Overall, Gen Ed and IEP appear as adjacent columns on the same row.

## Padding

Two decisions here that are easy to conflate.

`rpt_gsheets__dibels_bm_goals_calculations` currently applies two T&L planning
buffers: `+ 5` on the expected At/Above count, and `x 1.5` on the raw gap. Those
exist to pad intervention planning targets.

T&L asked that the Bright Spots tracker report attainment against Foundation
goals **with no padding at all**. So this needs an unpadded read of the same
join — whether that is a separate model, a flag, or unpadded columns alongside
the padded ones is a design decision.

Separately, T&L confirmed a change to the planning buffers themselves, and the
wording maps exactly onto the existing code. "Double padding" means `+ 5` on the
benchmark goal count, then `x 1.5` on the resulting number of students to move.
"Single padding" means the `x 1.5` alone. BOY to MOY keeps double padding; MOY
to EOY drops the `+ 5` and keeps the `x 1.5`, across K-8. That changes the
planning output rather than this tracker, but it is the same two expressions.

## Proposed approach

Build on the benchmark path, not the PM path:

1. A **long** intermediate — one row per school, grade, period, goal type and
   population — joining benchmark composite counts to
   `stg_google_sheets__dibels_foundation_goals` and carrying attained rate, goal
   rate, and the gap in percentage points, unpadded.
1. Tier assignment on each of those rows, so every population is tiered
   independently, with the G6-8 MOY branch keyed on grade band and period.
   Materialize the status as a column so it can drive a multi-select filter.
1. A reporting model shaped for the DDI layout: population pivotable into
   columns, plus the Below and Well Below share alongside the At/Above rate.

## Open questions

- **Which columns does T&L actually want?** This is the main thing to ask. The
  DDI view carries Grade Goal, % Mastery (Overall), four numbered buckets,
  % Mastery (Gen Ed), % Mastery (IEP), and % B/FB — every one of them tiered
  independently. Confirm whether DIBELS should reproduce that full set or a
  subset, since each column is a separate population to compute and tier.
  Included in that: **what the four buckets should map to.** The obvious DIBELS
  candidate is the four benchmark bands — Well Below, Below, At, Above — but
  that is a guess, and it partly overlaps the two goal types.
- **What plays the Module Code role.** A DDI row is one assessment. The closest
  DIBELS equivalent is a benchmark administration, so school / grade / period.
  Confirm the intended row grain.
- **Which goal sets.** T&L's answer named Foundation At/Above and Well Below for
  all students and SPED. Confirm topline and internal goals are out of scope.
- **Units.** This reads the tiers as percentage points throughout. Confirm.
- **Teacher drill-down.** T&L asked for a teacher-level drill-down like the DDI
  tracker's. Confirm benchmark data supports reliable teacher attribution before
  committing to it.
- **AY2026 Foundation goals are not loaded.**
  `stg_google_sheets__dibels_foundation_goals` holds AY2024 and AY2025 only.
  This gates both this work and the existing BM goals calculation.

## For Claude

The repo template puts this in a fold-out; the GitHub MCP write tools strip HTML
tags, so it is a plain heading here.

Relevant models:

- `stg_google_sheets__dibels_foundation_goals` — the rate goals, by region,
  grade, period, goal type
- `rpt_gsheets__dibels_bm_goals_calculations` — existing join of composites to
  Foundation rates, with the padding applied; #3962 narrowed it to the At/Above
  goal type, so re-reading both types is part of this work
- `int_amplify__all_assessments` — benchmark composites, the single safe read
  point for valid assessment scores
- `stg_google_sheets__dibels_bm_goals` — the manually frozen copy-paste output
  that currently feeds the dashboard's BM branch

Constraints:

- Do not route this through `int_amplify__pm_met_criteria` or any PM model. The
  grain is school/grade aggregate against a rate, not student against a score.
- The Well Below direction is correct as specified. Do not "fix" it.
- The DIBELS thresholds are correct as specified. Do not replace them with the
  DDI tracker's plus or minus 10.
- Benchmark history is expected to remain complete; this is not a
  current-year-only model.

Acceptance criteria: for a given region, grade and period, the tracker's attained
rate and gap reconcile against `rpt_gsheets__dibels_bm_goals_calculations` with
the padding removed, tier assignment matches the tables above including the G6-8
MOY exception, and status is materialized per population rather than per row.

GabyRangelB:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..."

...lay the groundwork for the new DIBELS Bright Spots tracker (#4952) by
reshaping how DIBELS foundation goals are stored, and by fixing a bug this
reshaping would otherwise have caused downstream.

Today, `stg_google_sheets__dibels_foundation_goals` holds one goal number per
region/grade/period, with no way to tell "all students" apart from "students
with IEPs." T&L's actual goal-setting sheet has separate numbers for each
group starting this year, and those numbers genuinely differ -- so the old
shape can no longer represent what T&L is setting. This PR reshapes that
table to carry a population column (All / IEP today, MLL to follow), and adds
a second new table holding the Bright Spot tier boundaries (Bright Spot / On
Track / In Range / Off Track) that the tracker will use to grade a school
against its goal.

While checking for side effects, we found that the one existing report reading
the old goals table (`rpt_gsheets__dibels_bm_goals_calculations`) would have
silently double-counted its goal comparisons once this year's goals load,
because its join didn't know about the new population column. That's fixed
here too.

Also included: a new `is_above_average_growth` flag on
`int_amplify__all_assessments`, needed for an upcoming Bright Spots column.

## Reviewer Notes

- The foundation goals table's columns changed (new: `Grade_Band`,
  `Population`, `Grade_Goal_Low`, `Grade_Goal_High`; nothing renamed or
  removed). Worth a second look: the join fix in
  `rpt_gsheets__dibels_bm_goals_calculations` -- see "For Claude" for how the
  bug was confirmed and how the fix was verified.
- `stage_external_sources --target staging` has **not** been run yet for
  either Sheets source in this PR -- needed before dbt Cloud CI can find them.
  Flagging per the dbt checklist below rather than checking that box.

### dbt

- [x] Include a properties file for all new models
- [ ] Include (or update) an exposure for all models consumed by a dashboard --
      not applicable yet; neither new/changed table is read directly by a
      dashboard today
- [x] Breaking change? -- no columns renamed or removed on a contracted model,
      additive only; see Reviewer Notes for the one behavior change found
- [ ] `stage_external_sources --target staging` -- **not yet run**, see
      Reviewer Notes

## CI checks

- [ ] Trunk
- [ ] dbt Cloud
- [ ] Dagster Cloud

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed (grangel).

**Schema retrofit** -- `stg_google_sheets__dibels_foundation_goals` moves from
one row per (academic_year, region, grade_level, period, goal_type) holding a
single `Grade_Goal` value, to the same grain plus `population`, with
`Grade_Goal_Low`/`Grade_Goal_High`/`Grade_Goal` all carried. The
low/high-to-single-value rule (`Grade_Goal` = low for At/Above, high for Well
Below) was verified against every existing Newark/Camden row across AY2024
and AY2025 with zero exceptions before being encoded -- not assumed. Full
generation procedure, named-range gotchas hit while building this (a stale
`src_google_sheets__dibels_foundation_goals` single-underscore range pointed
at a deprecated "Foundation Goals V1" tab; a row-bounded named range silently
truncated a paste), and the population/MLL open item are documented in the
`dibels-dashboard` skill (`.claude/skills/dibels-dashboard/SKILL.md`).

**New table** -- `stg_google_sheets__dibels_brightspot_goals` is the Bright
Spot tier boundary lookup: `(academic_year, grade_band, period, population) ->
tier boundaries in percentage points`. Boundaries are currently identical
across All/IEP/MLL and both years, confirmed against T&L's own Bright Spot
Brainstorm doc (byte-identical threshold text between "Overall" and "Sped"
rows) -- included as real dimensions anyway since the equivalent assumption
for goal values (that IEP would match All) turned out to be wrong once
already this build.

**The join fan-out bug, confirmed and fixed** --
`rpt_gsheets__dibels_bm_goals_calculations` joins to the foundation goals
table on `(academic_year, region, grade_level, period, goal_type)`, with no
population filter, because population didn't exist when it was written. Once
a population split exists for whatever year that model's
`current_academic_year` var points at, this join silently matches 2 goal rows
per key instead of 1 (e.g., Newark grade 0 MOY At/Above AY2025: All=0.62,
IEP=0.51 both match), doubling `n_admin_season_school_gl_*` and related
counts. Not live in prod today only because `current_academic_year` is
already 2026 and no AY2026 foundation goals exist yet -- first load of
AY2026 goals with a population split would have triggered it. Fixed by
scoping the join to `f.population = 'All'` in the `ON` clause (not `WHERE`,
which would have turned the `LEFT JOIN` into an `INNER JOIN` and dropped
every currently-unmatched AY2026 row). Verified both the bug and the fix
directly: queried the dev-rebuilt foundation goals table to confirm the
duplicate match exists pre-fix, then built
`rpt_gsheets__dibels_bm_goals_calculations` into dev with
`--vars '{current_academic_year: 2025}'` (AY2025 has real population data,
standing in for the eventual AY2026 case) and confirmed only the All-population
goal value (0.62) appears in the output, never the IEP one (0.51).

**`is_above_average_growth`** -- boolean on `int_amplify__all_assessments`,
`true` when `measure_semester_growth` is `Above Average` or
`Well Above Average`, `false` for the other three categories, `null` on BOY
rows and on PM rows (concept is Benchmark-only there). Verified against a
real student's full AY2025 BOY/MOY/EOY history and against the full
distribution of `measure_semester_growth` values for Composite rows (each of
the 5 categories maps to exactly the expected boolean, counts sum correctly).
Hit a BigQuery `UNION ALL` type-mismatch (`BOOL` vs `INT64`) from a bare
`null` in the PM branch -- fixed with `cast(null as bool)`.

Note this flag answers "did Amplify call this above its own national growth
norm," not "above our own population's average growth" -- the latter is a
distinct, unresolved ask (needs a reference population decision from T&L) and
is intentionally not attempted here. Documented in the skill so it isn't
re-litigated from scratch next time it comes up.

**Downstream impact confirmed**: `stg_google_sheets__dibels_foundation_goals`
has exactly one consumer in the codebase
(`rpt_gsheets__dibels_bm_goals_calculations`, confirmed via repo-wide grep),
addressed above.

**Outstanding before this is fully live**: `stage_external_sources --target
staging` for `google_sheets.src_google_sheets__dibels__foundation_goals` and
`google_sheets.src_google_sheets__dibels__brightspot_goals` -- dev-schema
restage and rebuild only were performed (not classifier-blocked); the
`--target staging` restage recreates the shared `zz_stg` external table and
needs direct authorization to run.

</details>

---
  - https://app.asana.com/0/0/1217834968249579


GabyRangelB:
Verdicts on the review findings (verified each against the code before acting, per repo convention):

1. **Missing `severity: error` on new uniqueness tests** — CONFIRMED, FIXED. Checked `dbt_project.yml`'s project-level default (`+severity: warn`), confirmed no override in the `google.sheets.staging` block, and confirmed `stg_google_sheets__gpa_goals.yml` sets `config: severity: error` explicitly on its own test. Added the same override to both new uniqueness tests, plus rebuilt both models in dev to confirm they still pass clean with `severity: error` in place.

2. **Enum-like columns have no `accepted_values` coverage** — FIXED. Added `accepted_values` (with `severity: error`) on `Grade_Goal_Type`/`Period`/`Population` (foundation_goals) and `Period`/`Population`/`Tier`/`Measured_Against` (brightspot_goals), matching the same `gpa_goals` sibling convention cited in the finding. Reordered those columns to the top of each `columns:` list per the repo's "tests sort to the top" rule.

3. **Passthrough models with no null-key filter** — DECLINED, already resolved. Both new tables' uniqueness tests already ran clean in the dev rebuild (203 rows / 96 rows, zero failures) before this comment was even posted — no phantom null-key rows. Confirmed again after adding the severity:error + accepted_values tests in fix #1/#2: still clean.

4. **No explicit encoding on file I/O** — FIXED, both call sites (`read_text` and the output `open`) now pass `encoding="utf-8"`.

5. **Nesting depth at the limit** — Acknowledged, no action now (correctly not a current violation). Already flagged in the `dibels-dashboard` skill's "Open: MLL population" section as something to watch when that generalization happens — will extract a helper then rather than pre-emptively refactoring working code today.

All fixes pushed in 3c99683.

PR #4969, opened 2026-08-25
feat(dbt): add salesforce_contact_id to rpt_gsheets__gpa_roster

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will add `salesforce_contact_id` as the final
column of `rpt_gsheets__gpa_roster`.

The value is passed through from `int_extracts__student_enrollments` (where it is
exposed as `salesforce_id`, itself an alias of the upstream
`salesforce_contact_id`), so no new join or ref was needed — just one extra
column carried through the `roster` CTE and emitted last.

Purpose: the data team needs the Salesforce contact ID alongside the GPA roster
to support periodic uploads into other systems. It is not intended for display
to sheet readers.

Not a breaking change — additive column only, no renames or removals. Rollback
is a straight revert of this commit.

### Validation

Built in dev with `--defer --favor-state` so all three upstream refs resolved to
prod relations (an initial build without `--favor-state` silently read stale
personal `zz_` copies and undercounted by ~600 rows — worth flagging for anyone
validating extract models locally):

- 6,227 rows / 6,227 distinct students, exactly matching the current prod
  relation — grain preserved, no fan-out from the added column through the
  `pivot`
- 1,802 populated IDs, all distinct, all exactly 18 characters (valid Salesforce
  18-char IDs)
- `trunk check --force` clean on both changed files

Population is high-school-only by nature (0% in grades 5-7, 78% in grade 9,
rising to 99.5% in grade 12) — Salesforce contacts are created as students reach
high school, so middle-school rows are legitimately blank. Confirmed with the
requester that blank values are expected for their use case.

## AI Assistance

Human-directed: the request, the output column name (`salesforce_contact_id`,
chosen over `salesforce_id` / `contact_id` to match the existing gsheets
precedent), and the decision to accept null middle-school rows.

AI-assisted (Claude Code): locating the upstream column, the SQL and YAML edits,
the dev build, and the validation queries above.

## Self-review

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — the
      existing `properties/rpt_gsheets__gpa_roster.yml` is updated with the new
      column and a description
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — this
      model's exposure already exists in `models/exposures/google-sheets.yml`;
      adding a column does not change it
- [x] **Breaking change?** No — additive column only. No contracted column was
      renamed or removed, so no downstream exposure breaks.
- [ ] If adding or modifying an external source, run `stage_external_sources` —
      N/A, no external source touched

### Dagster / Docs

Skipped — no Dagster, schedule, sensor, or integration changes. No column list
exists in Dagster config for this extract, so nothing outside dbt needs updating.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

### DIFF
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__gpa_roster.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__gpa_roster.yml
@@ -70,3 +70,8 @@ models:
           Weighted projected cumulative Y1 GPA, incorporating current
           in-progress grades not yet stored.
         data_type: float64
+      - name: salesforce_contact_id
+        description:
+          Salesforce contact ID for the student, used to join this roster to
+          KIPP Forward records in Salesforce.
+        data_type: string
--- a/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__gpa_roster.sql
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__gpa_roster.sql
@@ -8,6 +8,7 @@ with
             co.team,
             co.student_number,
             co.student_name,
+            co.salesforce_id,
             co.iep_status,
             co.gender,
             co.lep_status,
@@ -83,4 +84,6 @@ select
     cumulative_y1_gpa_projected_s1_unweighted,
     cumulative_y1_gpa,
     cumulative_y1_gpa_projected,
+
+    salesforce_id as salesforce_contact_id,
 from roster pivot (max(gpa_term) as gpa for term in ('Q1', 'Q2', 'Q3', 'Q4'))


anthonygwalters:
Verdicts on the three review items, each checked against the code before replying.

### 1. Model-level `description:` missing — CONFIRMED, deferred

The rule is real: `src/dbt/CLAUDE.md:1277` — "All new or modified models require
`description:` on the model and every column." This model is modified by this PR,
so it applies. Our properties file goes straight from `- name:` to `columns:`
with no model-level description. Both named counterexamples check out
(`rpt_gsheets__staff_roster.yml`, `rpt_gsheets__community_service_upload.yml`
have zero).

One correction to the supporting stat: adherence in
`models/extracts/google/sheets/properties/` is **17 of 62 files (27%)**, not
"roughly half." That cuts in the same direction as the finding's conclusion — the
rule is even less consistently enforced here than stated.

Deferred rather than declined: the same rule sentence also requires a description
on *every column*, and about 20 columns in this file have none. Adding only the
model-level line is partial compliance; doing it properly is a documentation pass
well outside a one-column addition. Raising with the requester to decide whether
to do it here or as a follow-up.

### 2. No uniqueness test — CONFIRMED, declined for this PR

Quoted rule verified verbatim at `src/dbt/CLAUDE.md:780-783`. I also checked the
exception immediately below it (`:785-789`, thin cross-project `extracts/`
wrappers sourcing `kipptaf_extracts` are contract-columns-only): it does **not**
cover this model, which refs `int_extracts__student_enrollments` and
`int_powerschool__*` directly rather than wrapping a kipptaf extract. So the
requirement genuinely applies and is genuinely unmet.

Declined for this PR as correctly-scoped pre-existing debt — the diff adds no
tests and removes none, and 19 of 62 files in this directory have any
`data_tests:` block at all. Worth noting the follow-up would lock in exactly the
grain I validated by hand (6,227 rows / 6,227 distinct students), so
`dbt_utils.unique_combination_of_columns` on `academic_year` + `student_number`
is the concrete shape.

### 3. PII classification — evidence is inverted; underlying concern stands

The finding states that neither `rpt_gsheets__student_contact_info.yml` nor
`rpt_gsheets__kfwd_student_salesforce_ids.yml` tags `salesforce_contact_id` with
`contains_pii`, and cites `rpt_gsheets__kfwd_student_salesforce_ids.yml:19` as
support. That file **does** tag it: `config: meta: contains_pii: true` sits at
lines 15-17, four lines above the cited column, and the file carries seven
`contains_pii` occurrences plus a `unique` test on `salesforce_contact_id`.

Actual precedent across the three models that emit this column:

| Model | `contains_pii` | `data_tests` |
| --- | --- | --- |
| `rpt_gsheets__kfwd_student_salesforce_ids.yml` | yes (7) | yes (4) |
| `rpt_gsheets__student_contact_info.yml` | no | no |
| `rpt_gsheets__gpa_roster.yml` (this PR) | no | no |

So precedent is mixed (1 of 3), not uniformly untagged — and the one model whose
entire purpose is exposing student Salesforce contact IDs is the one that tags
it. The finding's "consistent with existing repo precedent, not a new gap"
conclusion rests on the misread, so I'm not accepting it as-is.

The underlying question is still worth answering, and it is a policy call rather
than a code call: this model already emitted `student_name` and `student_number`
untagged before this PR, so the gap predates the change, but the closest
precedent argues for tagging. Raising with the requester.

### On the "checked clean" items

Contract enforcement is confirmed — `src/dbt/kipptaf/dbt_project.yml:110-113`
sets `+contract: enforced: true` on `extracts/`. Worth making explicit that this
is what makes the green dev build load-bearing rather than decorative:
`assert_columns_equivalent` runs inside `dbt build`, so the build passing is
positive proof the YAML column entry matches the SQL output. A prod SELECT would
not have shown that.

Noted and accepted: the review had no warehouse access, so the row-count,
null-rate and distinctness numbers in the PR body remain self-reported from my
local `--favor-state` build.

Issue #4970, opened 2026-08-25
feat(dbt): attach the remaining Miami course enrollments to a student stint


cbini:
## What and why

`dim_student_section_enrollments` attaches each course enrollment to a student
enrollment stint by overlapping their date ranges. After PR 1 of #4925, 2,025 of
Miami's 19,295 AY2026 rows (10.5%) still attach to no stint and carry a null
`student_enrollment_key`.

This is not a regression. Before that PR, Miami had zero AY2026 rows in this
mart at all. It is the residual of a fix that took the orphan count from 19,207
down to 2,025.

### Why the first 17,182 resolved

Focus leaves the schedule `end_date` null while a schedule is open, so
`cc_dateleft` is null on 18,706 of Miami's 19,295 rows. The join predicate
`cc_dateleft {greater than} entrydate` then evaluates to null rather than true,
and the row silently drops. PR 1 coalesces a null `cc_dateleft` to an
open-ended sentinel, matching a pattern the same model already used.

That change is provably safe for New Jersey: `cc_dateleft` is null on zero rows
in all three NJ regions, so the coalesce is a no-op there.

### What is left

| Cause | Rows |
| --- | --- |
| Student has a stint at that school and year, but the schedule row falls outside the stint window | 1,839 |
| Student has an AY2026 stint at a different school | 186 |

Of the 1,839, some 1,383 have a `cc_dateenrolled` on or after the stint's
`exitdate` — the schedule row starts after the student left.

### Why 10.5% was judged acceptable to ship

A null `student_enrollment_key` is an expected outcome in this model, not a
defect on its own. The model's own comment documents alumni placeholder rows
that match no stint by design. Measured in prod:

| Academic year | Rows | Orphaned | Rate |
| --- | --- | --- | --- |
| 2024 | 85,886 | 15,597 | 18.16% |
| 2025 | 91,901 | 18,153 | 19.75% |
| 2026 | 63,175 | 594 | 0.94% |

AY2024 and AY2025 are complete years and carry 18 to 20%. AY2026 is low only
because the year just started. Miami at 10.5% sits inside that band rather than
outside it.

## Proposed approach

Two independent questions, neither settled.

**The 1,839 window mismatches.** Establish what a Focus schedule `start_date`
actually means. If Focus dates a schedule row to the course period's term start
rather than the day the student joined the section, then a student who enrolls
mid-term gets a `cc_dateenrolled` before their own stint entry, and the overlap
[...]

Issue #4968, opened 2026-08-25
feat(dbt): source Miami drop status for the course enrollment facts


cbini:
## What and why

`int_students__course_enrollments`, introduced by PR 1 of #4925, leaves two
flags null for Miami because Focus carries no source for them.

| Flag | Why it is null |
| --- | --- |
| `is_dropped_section` | Focus has no drop convention |
| `is_dropped_course` | derived from `is_dropped_section` |

PowerSchool derives both from its own convention: a course enrollment row whose
`sectionid` is negative is a dropped section. Focus records nothing equivalent.
A schedule row is simply present or absent.

Null rather than false was deliberate. A false would read as a verified
zero drop rate for Miami across all 19,594 AY2026 Focus schedule rows, and the
data does not support that claim. The consequence is that Miami is excluded from
network drop-rate metrics rather than quietly diluting them.

This matters beyond the flags themselves. `dim_student_section_enrollments`
ranks course enrollments with `(is_dropped_section or is_dropped_course) asc` to
pick the current one per student per course. With both null for Miami, that
ordering term is null for every Miami row, so the rank falls through to the
date tiebreakers. That is acceptable while Miami has one schedule row per course
period, but it stops being acceptable the moment Focus starts recording
schedule changes.

## Proposed approach

Two candidate sources, neither verified yet.

**Schedule dates.** `int_focus__schedule` carries `start_date` and `end_date`
per row. A row whose `end_date` lands before its marking period ends is
plausibly a drop. This is an inference PowerSchool does not make, so adopting it
means the same column name would mean two different things by region. Settle
that before building on it.

**A Focus audit trail.** Focus may record schedule changes somewhere the current
staging models do not read. Check the raw dlt tables before assuming the
schedule row is all there is.

Whichever source wins, the acceptance bar is the same: the Miami drop rate has
to be reconcilable against something outside the model.

For Claude

Context, not instructions.

The nulls are marked in place with a `TODO` comment pointing at this issue. Find
them with `git grep -n 'TODO' -- src/dbt/kipptaf/models/students/intermediate/int_students__course_enrollments.sql`.

Relevant models:

- `src/dbt/kipptaf/models/students/intermediate/int_students__course_enrollments.sql`
  — the `focus_conformed` CTE, where the nulls live
- `src/dbt/focus/models/intermediate/int_focus__schedule.sql` — the Focus
  schedule grain, one row per student per course period
[...]

Issue #3999, opened 2026-05-22
refactor(dbt): migrate remaining base_powerschool__student_enrollments consumers to int_powerschool__student_enrollment_union


cbini:
## Context

PR #3985 migrated 12 marts consumers
(`dim_student_enrollments`, `dim_student_section_enrollments`, 9 `fct_*`) from
`base_powerschool__student_enrollments` to
`int_powerschool__student_enrollment_union` (now exists at kipptaf level as a
union of district-level int_ models). After that PR:

- **0** mart consumers of `base_powerschool__student_enrollments`
- **17** non-mart consumers still on base_

Tracking the remaining consumers here as a sub-task of #2541 (legacy `base_`
prefix rename).

## Why this matters now

PR #3985 introduced two relationships test warnings because
`int_extracts__student_enrollments` (consumed by `fct_survey_submissions` and
`bridge_survey_expectations`) still reads from `base_`, producing
`student_enrollment_key` hashes that don't match `dim_student_enrollments`
(which now hashes via int_'s `_dbt_source_relation`). These warnings clear
when `int_extracts__student_enrollments` migrates to int_, OR when all
`student_enrollment_key` hash compositions switch to `_dbt_source_project`
(more architecturally robust).

| Test | Rows |
| --- | --- |
| `relationships_fct_survey_submissions.student_enrollment_key → dim_student_enrollments` | 27,482 |
| `relationships_bridge_survey_expectations.student_enrollment_key → dim_student_enrollments` | 22,108 |

### Intermediate models (7)

- `int_assessments__fast_previous_year`
- `int_assessments__course_enrollments`
- `int_kippadb__roster`
- `int_topline__school_community_diagnostic`
- `int_extracts__student_enrollments` ← affects two marts relationships tests
- `int_finance__enrollment_targets`
- `int_reporting__promotional_status`

### Extracts / reports (3)

- `rpt_gsheets__gpa_flags_report`
- `rpt_gsheets__award_ceremony_gpa`
- `rpt_gsheets__school_metrics_extract`

### Source declarations (4)

- `kipptaf/models/powerschool/sources-kipp{newark,camden,miami,paterson}.yml`
  still declare `base_powerschool__student_enrollments` as a source — keep
  until all consumers migrated, then remove the source entries alongside the
  base_ model itself.

### Exposures (1)

- `kipptaf/models/exposures/google-sheets.yml` declares an exposure that
  depends on `base_powerschool__student_enrollments` — update once consumers
[...]

PR #4972, opened 2026-08-25
fix(dbt): wire Focus course enrollments into the kipptaf network layer for Miami AY2026

### LINKED ISSUE #4925: fix(dbt): wire Focus course enrollments and grades into the kipptaf network layer for Miami AY2026
## What's happening

Miami contributes zero course-enrollment and grades rows to the kipptaf network layer for AY2026. `base_powerschool__course_enrollments`, `base_powerschool__sections`, and `base_powerschool__final_grades` still union the frozen `kippmiami_powerschool` archive as their Miami branch, so five marts read empty for Miami: `fct_grades_assignments`, `fct_grades_category`, `fct_grades_gpa`, `fct_grades_term`, and `dim_student_section_enrollments`.

This splits into two halves with different readiness.

**Course enrollments — buildable now, no prerequisite.** `int_focus__schedule` shipped in PR #4725, carries 19,395 rows across 1,676 students, and is already declared in `kipptaf/models/focus/sources-kippmiami.yml` *and* already has a kipptaf wrapper. Same for `stg_focus__marking_periods`. It is a direct analogue of `base_powerschool__course_enrollments`: grain is student by course period by marking period, joined to `stg_focus__course_periods`, `stg_focus__courses`, and `stg_focus__marking_periods`, with the Focus full-year `marking_period_id = 0` sentinel normalized to null.

**Report card grades — buildable, but needs the focus package widened first.** This issue originally said every Miami row is imported course history with empty `report_card_grade_id` and `grade_scale_id`. That is true of the history rows and backwards for the current year. Re-measured 2026-08-20:

| `syear` | Rows | `report_card_grade_id` | `grade_scale_id` | `course_period_id` | `percent_grade` |
| 2026 | 1,432 and growing daily | 100% | 66% | 100% | 100% |
| 2025 and earlier | 15,278 | 0% | 0% | 0% | 0% |

The live AY2026 postings carry full keys. All four columns already exist in the raw dlt table `dagster_kippmiami_dlt_focus.student_report_card_grades` — no ingestion change is needed. What blocks them is `stg_focus__student_report_card_grades` not selecting them, plus the marking-period token described below.

**Gradebook grades — still thin.** `int_focus__gradebook_grades` shipped in PR #4919 with 729 rows and now carries 1,696 across 768 students. Real and growing, but too early for NJ-parity validation to mean much. `student_standard_grades` is still empty upstream and stays out of scope.

## Steps to reproduce

1. Confirm the archive is still the Miami branch: `grep -l kippmiami_powerschool src/dbt/kipptaf/models/powerschool/base/base_powerschool__course_enrollments.sql`.
1. Query any of the five marts filtered to Miami for `academic_year = 2026` — no rows.
1. Confirm Focus has the schedule data: `select count(*) from` `teamster-332318.kippmiami_focus.int_focus__schedule`.
1. Confirm the AY2026 report card rows exist and lose their marking period: group `int_focus__report_card_grades` by `academic_year` and count `marking_period_id is null`. AY2026 is 100% null; every prior year is 0% null. Note that `marking_period_academic_year` is null for all AY2026 rows, so filtering on it hides them entirely.

## Where

- **Code location / dbt project:** `focus` package plus `kipptaf`
- **Environment:** prod
- **Related:** Phase 3 of the plan on #4729. Covers the Base Students Course Enrollments and Little SIS items listed on #4257.

For Claude

### Prerequisite: widen the focus package report card models

`stg_focus__student_report_card_grades` drops four columns the raw dlt table already carries. Add them: `report_card_grade_id`, `grade_scale_id`, `course_period_id`, `percent_grade`.

Then resolve the `TODO` in that model on the token-prefixed marking period. The comment says the token vocabulary is unknown. It is now knowable — only 5 distinct values exist for AY2026 and every one decodes cleanly against `stg_focus__marking_periods`:

| Raw | Tail id | Resolves to | School | Dates |
| `DT7057`, `DY7057` | 7057 | Quarter 1 (quarter) | 15 | 2026-08-12 to 2026-10-16 |
| `DT7181`, `DY7181` | 7181 | Quarter 1 (quarter) | 58 | 2026-08-12 to 2026-10-16 |
| `DT7178` | 7178 | Full Year (year) | 58 | 2026-08-12 to 2027-06-03 |

Right school, right dates, zero orphans, zero collisions — the collision risk the comment hedges against does not materialize.

**The prefix is load-bearing, not noise.** A student gets both a `DT` and a `DY` posting against the same marking period, so stripping the token collapses the grain: 1,415 rows reduced to 933 distinct `(student_id, course_period_id, marking_period_id)` at time of measurement, against 1,415 distinct with the prefix retained. Decode the digits into `marking_period_id` and keep the alpha token as its own column, then include it in the grain. Discarding it silently creates roughly 480 duplicate rows.

**Open external dependency:** what `DT` versus `DY` actually means is not established from the data, and both are needed to keep the grain unique. This needs a Focus support answer before the column can be named, and it is the one thing on this issue that cannot be settled from the warehouse.

### Prerequisite: grade-scale resolution path

Do not join on the raw `grade_scale_id` column — it is null on 488 of the AY2026 rows, split across both schools (28% at school 58, 45% at school 15), so it is not a school-scoped gap that can be filtered around.

Resolve the scale through `report_card_grade_id` instead. It joins to `stg_focus__report_card_grades` with zero orphans, and that table's `scale_id` is non-null on every matched row — coverage goes 66% to 100%. From there join `stg_focus__report_card_grade_scales`. Both staging models already exist and are currently unused by any intermediate. `course_period_id` also joins to `stg_focus__course_periods` with zero orphans and no fan-out.

Course enrollments half:

| Model | Marts affected |
| --- | --- |
| `base_powerschool__course_enrollments` | `fct_grades_assignments`, `fct_grades_category`, `dim_student_section_enrollments` |
| `base_powerschool__sections` | section identity for the above |
| `stg_powerschool__cc`, `stg_powerschool__courses` | staging passthroughs unioning the archive |

Grades half:

| Model | Marts affected |
| --- | --- |
| `base_powerschool__final_grades` | `fct_grades_term` |
| `int_powerschool__gradebook_assignments`, `int_powerschool__gradebook_assignments_scores` | `fct_grades_assignments` |
| `int_powerschool__category_grades`, `int_powerschool__category_grades_pivot` | `fct_grades_category` |
| `int_powerschool__gpa_term`, `int_powerschool__gpa_cumulative`, `int_powerschool__gpa_cumulative_year` | `fct_grades_gpa` |
| `int_powerschool__final_grades_pivot`, `int_powerschool__section_grade_config`, `int_powerschool__gradescaleitem_lookup` | grade-scale resolution |

### Ingestion context from #4865

`gradebook_grades` carries only `student_id` and `assignment_id` — nothing connects an assignment to a course period on its own. Two tables were added to `focus.yaml` to resolve it, and both now have staging models:

- `gradebook_assignments_join_course_periods` — assignment to course period, plus assigned, due, and publish dates
- `gradebook_assignment_types_join_course_periods` — category to course period per marking period, carrying `final_grade_percent`, the category weight

### Open design decision

Same question as the attendance phase: Phase 1 landed SIS-neutral `int_students__*` models rather than branching same-named `int_powerschool__*` models. Decide whether this follows that precedent. It matters more here because the `base_` models in scope are the ones #3999 plans to split into focused intermediates and #2541 plans to rename — sequencing this against those avoids churning the Focus branches twice.

The attendance phase settled two related points that carry over — see the design comment on #4924. Branch at the altitude PowerSchool derives the model rather than recomputing hashes at network altitude, and year-scope the union rather than excluding Miami wholesale, because the frozen archive holds Miami AY2020 through AY2025 for these models too.

### Sequencing

**Three PRs. A single PR is mechanically possible but lands on a known, non-self-healing failure mode.**

The single-PR cross-project workflow in `src/dbt/kipptaf/CLAUDE.md` handles CI: add a `target=staging` branch to `sources-kippmiami.yml` and seed `zz_stg_kippmiami_focus`. Because the focus package models are MODIFIED here rather than merely read, `dbt clone` would copy the old schema — use `stage_external_sources --target staging` then `dbt build --select ... --target staging`, plus clone and build `zz_stg_kipptaf` itself.

What it does not handle is the post-merge prod deploy. Checked rather than assumed:

- kipptaf's focus wrappers are views — the `focus` block in `kipptaf/dbt_project.yml` sets only `+schema`, no `+materialized`, so dbt's view default holds.
- The translator therefore assigns `dbt_union_relations_automation_condition()` via the `union_relations in raw_code` branch in `libraries/dbt/dagster_dbt_translator.py`.
- That condition fires on recursive ancestor `code_version_changed`, not `any_deps_updated`.
- The dep-code-version gate that would serialize a cross-code-location deploy was deliberately removed for causing permanent deadlocks. Its removal note in `core/automation_conditions.py` scopes the coverage claim to the in-code-location case and states that cross-code-location races fail at BigQuery query time.

This is #4290. If the kipptaf wrapper materializes at deploy before the kippmiami upstream rebuilds with the new schema, `.since(newly_updated)` consumes the trigger, the wrapper stays compiled against the old column set, and downstream reads fail `failed to parse view`. It does **not** self-heal — recovery is a manual `launch_run` of the wrapper plus every consumer. Diagnose by comparing the wrapper's stored `input_data_version` materialization tag against the upstream's current `data_version`.

Five marts read this wrapper. A third PR costs less than babysitting that deploy.

1. **kipptaf, course enrollments.** Ships now. Nothing here changes a package model's columns, so no staging seed, no cross-project coordination, and no exposure to the wrapper race.
1. **focus package, report card grades.** Widen `stg_focus__student_report_card_grades`, decode the marking-period token, wire the grade-scale joins into `int_focus__report_card_grades`. Blocked on the DT versus DY answer. Let Dagster materialize it in prod before opening the next PR.
1. **kipptaf, report card grades.** Declare and wrap `int_focus__report_card_grades`, branch `base_powerschool__final_grades` and the GPA chain. Needs `dbt clone --select ... --target staging` to refresh `zz_stg_kippmiami_focus` or CI reads a stale copy.

Hold the `gradebook_grades` branch of `fct_grades_assignments` and `fct_grades_category` until volume supports validation — re-measure with `count(*)`, not `__TABLES__.row_count`.

### Validation

1. Miami AY2026 presence in `dim_student_section_enrollments` and each restored fact.
1. Miami historical reconciliation against the archive for the overlapping years, where Focus carries history.
1. Uniqueness test on `(student_id, course_period_id, marking_period_id, grade_type_token)` for the report card grain — the prefix-collapse above is the specific failure this catches.
1. After the PR 3 deploy, confirm the wrapper did not lose the race: compare its stored `input_data_version` tag to the upstream's current `data_version` before trusting a green build.
1. Uniqueness test on each modified model's key.

Refs #4729
Refs #4257
Refs #4865
Refs #4290

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will restore Miami's course enrollments and
sections to the kipptaf network layer for AY2026.

Miami moved from PowerSchool to the Focus SIS. The network layer still read a
frozen PowerSchool archive for Miami, and that archive stops at AY2025, so Miami
contributed zero course-enrollment and zero section rows for the current school
year. Five marts read empty for Miami as a result.

Two new SIS-neutral models now union the four PowerSchool districts with a Focus
branch for Miami: `int_students__course_enrollments` and
`int_students__course_sections`. A third, `int_students__courses`, does the same
for the course catalog so Miami's sections can resolve their course. The union is
scoped by year rather than excluding Miami outright, so the frozen archive keeps
serving Miami AY2020 through AY2025.

`base_powerschool__course_enrollments` and `base_powerschool__sections` become
one-line compatibility passthroughs. More than 50 models read them and none
needed editing. Migrating those consumers is tracked separately in #3999.

New Jersey is unchanged. Every frozen-year count matches prod exactly.

**Scope, stated precisely.** This restores Miami at the model layer and in
`dim_student_section_enrollments`, `dim_course_sections` and `dim_courses`. It
does NOT yet reach roughly 40 downstream reports. Those filter with a bare
`not is_dropped_section`, and that column is null for Focus by design, so
`not null` is null and every Miami row is removed. A further five consumers still
join on `cc_studentid` and `cc_yearid`, which Focus never populates. Neither is a
regression -- those reports held no Miami rows before this branch either -- but
the data now exists one layer below them. Tracked in #4973.

## Reviewer Notes

**One thing must happen after merge, or production serves stale data silently.**
`base_powerschool__sections` converts from a table to a view. In BigQuery
`create or replace view` does not drop a pre-existing table at the same path — it
keeps serving the old table with no error anywhere. Someone has to run this once:

```bash
uv run dbt build --select base_powerschool__sections --full-refresh \
  --project-dir src/dbt/kipptaf --target prod
```

`base_powerschool__course_enrollments` needs no such step. It is already a view.

**The student-stint join moved off PowerSchool internal ids.**
`dim_student_section_enrollments` joined on `studentid` and `yearid`, which are
null for every Focus row, so Miami rows would have landed orphaned from the
student dimension. It now joins on `student_number`, `schoolid` and
`academic_year`. The two keys have identical selectivity in all three New Jersey
regions, so nothing there moves.

**A tenth of Miami's rows still carry no student enrollment key, and that is
expected.** 2,025 of 19,295 attach to no enrollment stint. That is inside this
model's normal range — New Jersey runs 18% to 20% in completed years. Tracked in
#4970, with a warn-severity test that fires if the rate climbs past 15%.

**Two flags land null for Miami rather than false.** Focus has no equivalent of
PowerSchool's dropped-section convention, so `is_dropped_section` and
`is_dropped_course` are null with a `TODO(#4968)`. A false would read as a
verified zero drop rate for Miami across every row, which the data does not
support.

**Two New Jersey values do change, both corrections.**
`is_current_homeroom` and `is_current_section_enrollment` each gain rows per New
Jersey region. Orphaned rows previously shared one network-wide ranking
partition, and Miami's null drop flags sorted ahead of New Jersey's real ones,
taking rank 1. Adding the source project to the partition gives each region its
own. Only stint-orphaned rows are affected.

### General

No Dagster changes.

### dbt

- [x] Include (or update) an
      summary above.
      See

No columns were renamed or removed. Both passthroughs keep their full column set,
verified by an `INFORMATION_SCHEMA` comparison against prod that returned zero
dropped columns. No external sources changed.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Design spec and implementation plan are in the
diff under `docs/superpowers/`.

**Verified numbers, measured 2026-08-25 against prod and the dev build.**

Frozen years, which are the meaningful parity check because the live year drifts
during the day:

| Source | AY2025 course enrollments |
| --- | --- |
| `kippnewark` | 51,575 |
| `kippcamden` | 18,730 |
| `kipppaterson` | 4,531 |
| `kippmiami` | 17,065 |

Do not compare AY2026 counts against any number written down. Newark moved
44,481 to 44,523 and Camden 14,655 to 14,689 within a single day of development.
Compare against live prod in the same session.

Other verified values: Miami AY2026 course enrollments 19,295; Miami AY2026
sections 823; `dim_course_sections` 33,680 rows with 0 unresolved
`course_section_key`; Miami orphaned `student_enrollment_key` 2,025 of 19,295;
New Jersey orphan counts 5,587 / 2,464 / 128; full build PASS=383 WARN=17
ERROR=0.

**Row-count gotcha.** `int_focus__schedule` means two different relations. The
`kippmiami_focus` package model holds 19,594 rows; the kipptaf wrapper holds
19,295 because it deduplicates superseded same-day stints on
`(student_id, course_period_id)`. Every kipptaf model reads the wrapper. A
validation written against 19,594 reads as a 299-row loss that is not happening.

**Build recipe, both halves required.** Local builds of this graph need
`--defer --favor-state --state <prod manifest>` AND a `--select` naming the
changed models with `+`. Omit the first and the `int_focus__*` upstreams resolve
to empty personal-dev copies, dropping the entire Focus branch with every test
still green. Omit a changed model from the second and it resolves to prod's
pre-branch copy, dropping this branch's changes instead. Both failure modes fired
during development and both produce a green build with wrong data.

**Empty-relation guard.** Both year-boundary CTEs use
`coalesce(min(academic_year), 9999)`. Without the coalesce, an empty
`int_focus__schedule` yields a null boundary, and `not (TRUE and NULL)` is NULL,
which silently drops every Miami row of every year while New Jersey survives —
including six years of archive history, with a green build and no test firing.
The `--defer` trap above is a live trigger for exactly that state.

**Teacher resolution.** Focus `teacher_id` resolves through
`int_focus__users.staff_id` to the staff roster on `ein`, coalesced with email.
The email fallback is load-bearing: of 771 course periods with a user, 711
resolve on `ein`, 32 only on email, and 28 on neither. A warn-severity test
carries a documented floor of 80 — 67 course periods with no teacher assigned in
Focus at all, plus 13 unresolvable. Both are source data gaps.

**Deliberate asymmetry, do not "fix" it.** `int_students__courses` does NOT
year-scope Miami's PowerSchool catalog out, unlike its two sibling models.
Archive sections hash their PowerSchool `courses_course_number` into
`dim_course_sections.course_key`, so removing those catalog rows would orphan the
course FK on every Miami archive section. This is documented in that model's
properties file.

Refs #4925
Refs #4968
Refs #4970
Refs #4973

</details>


cbini:
## dbt Cloud CI warnings, triaged

CI passed with 18 warnings. Here is what each one is, so nobody has to re-derive
it. Two are this branch's own and expected; the rest are pre-existing or CI
environment artifacts.

### The one that looks alarming and is not

`relationships_fct_grades_term_student_section_enrollment_key` reports **89,474**
orphans. In prod the same test reports **0**. That difference is a stale CI
source copy, not data loss.

The chain, all measured:

| Relation | Newark AY2026 rows |
| --- | --- |
| prod district source (`kippnewark_powerschool`) | 44,474 |
| `zz_stg_kipptaf_powerschool` — what the staging marts were built from | 44,474 |
| `zz_stg_kippnewark_powerschool` — what this CI run reads | **32,426** |
| this PR's `int_students__course_enrollments` | **32,426** |

The PR model reproduces the staging district source exactly. The staging marts,
including the deferred `fct_grades_term` this test reads as its child, were built
when those source copies were fuller. So the child carries keys a freshly rebuilt
dimension cannot contain.

Confirmation that the environments themselves are consistent: staging and prod
`dim_student_section_enrollments` both hold exactly 63,079 AY2026 rows, with zero
keys in one and not the other. The skew is entirely in the district source copies
CI reads.

Any PR that rebuilds `dim_student_section_enrollments` against these source
copies would produce the same warning. After merge, the prod build reads prod
sources and this resolves on its own.

### This branch's own warnings, both expected

`test_focus_course_sections_teacher_resolves` reports **80**. That is the
documented floor, stated in `tests/properties.yml`: 67 Focus course periods with
no teacher assigned in Focus at all, plus 13 whose user reaches the staff roster
through neither `ein` nor email. Both are source data gaps, not modeling defects.

`dim_student_section_enrollments__no_multi_stint_overlap` reports **147**. Miami
contributes **16** of that, which is the number this branch is responsible for and
which reproduces exactly in every environment I measured.

One correction to my own commit here. The description I shipped states the
baseline as "154 rows — 16 Miami, 135 Newark, 3 Camden". Miami and Camden hold at
16 and 3, but Newark read 135 in my dev build and 128 in CI. The New Jersey
component drifts with the live school year, exactly as this branch's own plan
warns about live-year counts. Pinning it to 135 was my error and I am correcting
the description to state Miami's stable 16 and describe the NJ component as
drifting rather than fixed.

### Pre-existing, unchanged by this branch

`test_base_powerschool__course_enrollments_no_studyear_course_overlap` reports
14,514. Prod reports **14,874** for the same test, so this branch slightly
reduces it rather than causing it.

The remaining warnings are on models this branch does not touch: survey
submissions and responses uniqueness, the Tableau gradebook and AP dashboard
uniqueness, the Clever enrollments feed, KFWD survey uniqueness, DeansList mod
audit, the assessment relationships pair, graduation path accepted values, and
the assessment scaffold expression. The one log warning names a breaking change
to `rpt_gsheets__gpa_roster`, which is not in this diff.

cbini:
## Response to the automated review, finding by finding

Pushed in bc80b28312. Verdict on each of the 7, including the two I declined and
why.

### 1. `not is_dropped_section` removes every Miami row — ACCEPTED, tracked, not fixed here

Correct, and the blast radius is as described. I measured it: 59 files under
`src/dbt/kipptaf/models` use the bare idiom, and **40** of them also read the
course-enrollment or section models. Since the column is null for Focus and
`not null` is null, all 40 silently exclude Miami.

Not fixed in this PR, by decision. Two reasons. It is not a regression — those
reports held zero Miami AY2026 rows before this branch too, because the source
had none. And the fix is not mechanical: some of the 40 are New Jersey-scoped by
design and should keep excluding Miami, so each needs a per-report judgment
rather than a sweep.

Tracked in #4973, which carries the grep, the affected model list, and an
explicit constraint against resolving it by emitting false for the drop flags.

The review was right that the PR body overclaimed. I have rewritten the summary
to state that this restores Miami at the model layer and in the three dimensions,
and does not yet reach the reports.

### 2. Other consumers still join on PowerSchool internal ids — ACCEPTED, tracked

Also correct, and the same silent-orphan class this PR fixed for
`dim_student_section_enrollments`. `int_assessments__course_enrollments`,
`int_extracts__course_enrollments_by_term`, `fct_grades_assignments` and
`fct_grades_category` all join on `cc_studentid` or `cc_yearid`, which Focus
never populates. `bridge_course_section_teachers` is a different shape, as noted
— it inner-joins a PowerSchool-only staging table with structurally zero Miami
rows, so null handling would not help it.

Same reasoning as finding 1, same tracker: #4973.

### 3. Missing `contains_pii` on `students_student_number` — FIXED

Correct and inconsistent with the sibling model this one follows. Tagged.

### 4. `credit_hours` numeric versus float64 — VERIFIED, no change needed

Measured rather than reasoned about. The built column in the PR schema resolves
to **FLOAT64**, so BigQuery's set-operation supertype rules widen the Focus
`numeric` as the finding predicted. The `float64` contract on `dim_courses` is
satisfied. No cast added — an explicit one would be redundant.

### 5. Email-fallback join fan-out — VERIFIED, no change needed

`select lower(google_email), count(*) ... having count(*) greater than 1` against
the staff roster returns **0 rows**. No duplicates, so the join cannot fan out
today. The finding's own note holds: the model's error-severity
`unique_combination_of_columns` test would catch it loudly if that ever changed,
so this is a build-time signal rather than silent corruption. Adding a uniqueness
test to `google_email` is reasonable but belongs to that model, not this PR.

### 6. Duplicated expression in the orphan-rate test — FIXED

The `HAVING` clause now references the `orphan_rate` alias instead of repeating
`safe_divide(...)`. Confirmed BigQuery resolves select-list aliases in `HAVING`,
and re-ran the test: PASS.

### 7. Undocumented `course_name` and `credit_hours` — DECLINED

The finding notes this matches existing precedent for SIS-neutral wrapper models,
and that is the reason. `int_students__student_enrollment_union` documents 11 of
its 20-plus output columns. Documenting the columns that carry branch-specific
semantics, and leaving pass-through columns to the upstream model, is the pattern
here. Both columns pass through unchanged from `stg_powerschool__courses` and
`int_focus__courses`.

### One correction of my own, not from the review

The multi-stint test description I shipped pinned the baseline at "154 rows — 16
Miami, 135 Newark, 3 Camden". CI reported 147. Miami held at 16 and Camden at 3;
Newark moved 135 to 128. The New Jersey component drifts with the live school
year, which is a thing this branch's own plan warns about, and I pinned it
anyway. The description now names Miami's stable 16 and describes the NJ
component as drifting.

src/dbt/kipptaf/models/students/intermediate/int_students__course_enrollments.sql
```diff
@@ -0,0 +1,180 @@
+with
+    union_relations as (
```
cbini: should be its own model

src/dbt/kipptaf/models/students/intermediate/int_students__course_sections.sql
```diff
@@ -0,0 +1,115 @@
+with
+    union_relations as (
```
cbini: should be its own model

src/dbt/kipptaf/models/students/intermediate/int_students__courses.sql
```diff
@@ -0,0 +1,58 @@
...
+            syear,
+            course_id,
+            short_name as course_number,
+            title as course_name,
+            credit_hours,
+
+            -- Focus has no credit-type field. Null rather than a guess.
+            cast(null as string) as credittype,
+
+            'kippmiami' as _dbt_source_project,
```
cbini: should come from the union_relations model

src/dbt/kipptaf/models/students/intermediate/int_students__course_enrollments.sql
cbini: Extracted to `int_powerschool__course_enrollments_union`, matching the shape of `int_powerschool__student_enrollment_union` — a model that does nothing but the union plus `_dbt_source_project`. This model now refs it.

The year-scope filter and the `coalesce(min(academic_year), 9999)` boundary stayed here rather than moving into the union model. They are Focus-boundary logic, not union logic, and the coalesce is what stops an empty Focus relation from silently deleting every Miami row of every year.

src/dbt/kipptaf/models/students/intermediate/int_students__course_sections.sql
cbini: Extracted to `int_powerschool__sections_union`, same shape. This model now refs it.

src/dbt/kipptaf/models/students/intermediate/int_students__courses.sql
cbini: Fixed. `int_focus__courses` is itself a `union_relations` wrapper and already carries `_dbt_source_project`, so it is selected from the ref instead of written as a literal.

Applied the same fix to the two places you did not flag, so the three Focus branches agree: `int_students__course_sections` had the same hardcoded `kippmiami`, and `int_students__course_enrollments` had both that and a hardcoded `Miami` region — the latter now derives through `extract_region`, matching `int_focus__student_enrollments`.

Verified behavior-preserving on the dev build: zero null `_dbt_source_project` on all three models, `region` still resolves to one correct value per project, frozen AY2025 unchanged at newark 51,575 / camden 18,730 / paterson 4,531 / miami 17,065, and Miami AY2026 still 19,295 course enrollments and 823 sections.

PR #4988, opened 2026-08-26
feat: add the EdPlan feed to kipppaterson

### LINKED ISSUE #4984: feat: add the EdPlan feed to kipppaterson
## What and why

Paterson has no EdPlan feed. Newark and Camden both pull the NJSMART-PowerSchool export from EdPlan over SFTP, so their IEP data is real. Paterson fakes it: the kipptaf marts fall back to PowerSchool's `spedlep` field for Paterson students, which carries none of the placement codes, consent dates, or related-services flags the NJ regions get.

Add the EdPlan pull for Paterson, then cut Paterson's downstream models over to it.

## Proposed approach

Four layers, following the Camden pattern:

1. **Ingest** — a `kipppaterson/edplan/` Dagster module (asset, sensor, schema), the `SSH_EDPLAN` resource wired into `definitions.py`, and the Paterson EdPlan SFTP secret. The 1Password entry already exists.
1. **Regional dbt** — add `../edplan` to `kipppaterson/packages.yml` and an `edplan` block to `dbt_project.yml`.
1. **kipptaf union** — a `sources-kipppaterson.yml`, plus a Paterson leg in `int_edplan__njsmart_powerschool_union`.
1. **Downstream cutover** — drop Paterson from the two PowerSchool fallbacks so the marts read real EdPlan data.

<details>
<summary>For Claude</summary>

> Context, not instructions.

Two blockers found while tracing this:

**Paterson has no archive table.** The shared regional model `src/dbt/edplan/models/intermediate/int_edplan__njsmart_powerschool_union.sql` hard-references `source("edplan", "stg_edplan__njsmart_powerschool_archive")`. That table exists as a real BigQuery table only in `kippcamden_edplan` and `kippnewark_edplan` (one-time historical loads; the model that builds it is disabled everywhere). Paterson's build would fail on a missing table. Gate the archive `union all` behind a var and set it false for Paterson.

**Two Paterson-specific fallbacks would double-count.** Once Paterson enters the union, both need Paterson removed:

- `kipptaf/models/students/intermediate/int_students__student_enrollments.sql` — two `ar.region in ('Miami', 'Paterson')` conditionals that read `ar.spedlep` instead of the EdPlan join. Miami stays; Paterson comes out. The comment above them says the exceptions are not stale, so update it too.
- `kipptaf/models/marts/dimensions/dim_student_iep_status.sql` — the `pm_leg` CTE filters `_dbt_source_project in ('kipppaterson', 'kippmiami')`. Paterson would land in both `nj_leg` and `pm_leg` and collide on `student_iep_status_key`. Miami stays.

Out of scope: the Paterson `rpt_powerschool__autocomm_students_iep` extract, noted as a possible follow-on in `docs/superpowers/specs/2026-07-14-paterson-autocomm-extracts-design.md`.

Also update `src/teamster/code_locations/kipppaterson/CLAUDE.md` (its "No `edplan`" line), `src/dbt/kipppaterson/CLAUDE.md`, and the paths list in `.github/workflows/deploy-prod-kipppaterson.yaml`.

Done when `dbt build --select +int_edplan__njsmart_powerschool_union +dim_student_iep_status` passes in all three NJ projects, and `dim_student_iep_status` has no duplicate `student_iep_status_key` for Paterson.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will pull Paterson's IEP data from EdPlan and use
it in the marts, instead of faking it from PowerSchool.

Newark and Camden both pull a special-education export from EdPlan over SFTP.
Paterson had no such feed, so the kipptaf marts fell back to PowerSchool's
`spedlep` field for Paterson students. That field carries none of the placement
codes, consent dates, or related-services flags the other two NJ regions get.

Four layers, following the Camden pattern:

1. A `kipppaterson/edplan/` Dagster module — asset, sensor, schema — plus the
   `SSH_EDPLAN` resource, the Paterson EdPlan SFTP secret, and the new paths in
   the Paterson deploy workflow.
1. `../edplan` added to `kipppaterson/packages.yml`, with an `edplan` block in
   `dbt_project.yml`.
1. `kipppaterson_edplan` added as a kipptaf source and unioned into
   `int_edplan__njsmart_powerschool_union`.
1. Paterson dropped from the two PowerSchool IEP fallbacks, so the marts read
   real EdPlan data.

Closes #4984

## Reviewer Notes

**Paterson's export has a different file name.** Its EdPlan account does not
produce Newark's and Camden's `NJSMART-PowerSchool.txt`. It writes one dated file
per day under a different name. I checked the live account: same 37 columns, same
comma-delimited shape, every column maps onto the existing Avro schema with
nothing left over. So no new schema and no new staging model — only the regex
changed. Details in the fold-out.

**The archive leg of the shared union is now optional.** Newark and Camden hold a
one-time NJSMART archive table that Paterson has none of. Without a switch,
Paterson's build fails on a missing table.

**Two Paterson carve-outs had to come out or the marts double-count.** Both were
written when Paterson had no EdPlan rows. `dim_student_iep_status` would have
broken its own uniqueness test.

**The external table points at the test bucket, not Paterson's.** Paterson has no
production EdPlan file yet, and BigQuery cannot create an external table over an
empty prefix. This is the documented workaround (`docs/guides/sftp-integration.md`,
"Once the asset is materialized in production, drop `--test`"). **After the first
production materialization, someone must re-stage the source without the
test-bucket override.**

### General

      `kipppaterson` passes.
      location — `test_definitions_kipppaterson` passes. `kippmiami` and
      `test_definitions_all` fail, but they fail identically on `main` in this
      codespace (missing credentials), so they are not from this branch.
      with the correct asset key format and IO manager — reuses
      `build_sftp_file_asset` and `build_edplan_sftp_sensor`; asset key is
      `kipppaterson/edplan/njsmart_powerschool`; `io_manager_gcs_avro`.

      — no new models. Paterson picks up the shared `edplan` package's existing
      models and their properties files.
- [x] Include (or update) an
      exposure needed; the existing kipptaf exposures downstream of
      `dim_student_iep_status` are unchanged.
      summary above. — No column renamed or removed. Paterson's *values* in
      `dim_student_iep_status` and `int_students__student_enrollments` do change,
      from PowerSchool `spedlep` to real EdPlan data. Rollback is a revert: the
      two carve-outs go back and Paterson leaves the union.
      See
      — done, against the test bucket. See Reviewer Notes.

      catalog: `uv run scripts/gen-automations-doc.py` — **not done.**
      `scripts/CLAUDE.md` says not to run it in a codespace: it silently drops
      `kipptaf`, and a trial run here also dropped `kippmiami`. Needs a
      bootstrapped terminal. Paterson's existing row in that catalog is already
      stale for an unrelated reason (it still lists the retired PowerSchool SFTP
      assets).
      `CLAUDE.md` by running `/claude-md-management:revise-claude-md` — updated
      `code_locations/kipppaterson/CLAUDE.md`, `dbt/kipppaterson/CLAUDE.md`, and
      `dbt/edplan/CLAUDE.md` by hand.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

**The file-name finding.** Paterson's `Reports/` directory holds only:

```text
9754  EDPlan Special Education Data Export-PCG Support-2026-08-24.txt
9761  EDPlan Special Education Data Export-PCG Support-2026-08-25.txt
```

The copied `NJSMART-Power[Ss]chool\.txt` regex matched nothing and the asset
raised `FileNotFoundError`. Column comparison against the `NJSmartPowerschool`
Pydantic model: 37 header columns, 37 schema data fields, zero on either side
without a counterpart (`source_file_name` is the 38th, added by the loader).
`file_sep` already defaults to `,` and `slugify_cols=True` turns
`Student Number` into `student_number`, so the header's title-case spacing needs
no handling.

The name carries the export date and one file accumulates per day, so a plain
regex trips `build_sftp_file_asset`'s multiple-match guard. The fix uses a
`date` named group, which `compose_regex` substitutes with the partition key, so
exactly one file matches per partition. That is only sound because each file's
mtime date equals the date in its name — verified on both files:

```text
mtime (UTC)                partition key   filename date
2026-08-24T16:39:10+00:00  2026-08-24      2026-08-24
2026-08-25T16:38:29+00:00  2026-08-25      2026-08-25
```

`start_date` is `2026-08-24`, the first file EdPlan produced. The earlier
`2026-07-01` would have created seven weeks of partitions that can never resolve
a file.

**The archive switch.** `src/dbt/edplan/models/intermediate/int_edplan__njsmart_powerschool_union.sql`
hard-referenced `source("edplan", "stg_edplan__njsmart_powerschool_archive")`.
Confirmed against `INFORMATION_SCHEMA`: that table exists as a real table only in
`kippcamden_edplan` and `kippnewark_edplan`, one-time historical loads whose
building model is disabled everywhere. The archive leg is now behind
`var("edplan_has_archive")`, defaulted `true` in the package so Newark, Camden,
and kipptaf are untouched, and set `false` in `kipppaterson/dbt_project.yml`.
Verified by compiling all three: Paterson's compiled SQL has zero references to
the archive relation, Newark's and Camden's still have one.

**The two carve-outs.**

- `int_students__student_enrollments.sql` — two
  `ar.region in ('Miami', 'Paterson')` conditionals reading `ar.spedlep` instead
  of the EdPlan join, now `ar.region = 'Miami'`. The comment above them
  explicitly said the exceptions were not stale; it now says Paterson's no longer
  applies and why. Miami stays: it genuinely has no rows in the union.
- `dim_student_iep_status.sql` — `pm_leg` filtered
  `_dbt_source_project in ('kipppaterson', 'kippmiami')`, now `= 'kippmiami'`.
  Left alone, Paterson would land in both `nj_leg` and `pm_leg` and produce two
  rows per enrollment stint, colliding on `student_iep_status_key`, which carries
  a `unique` data test.

**Verification run.** `pytest tests/assets/test_assets_edplan_sftp.py -k kipppaterson`
passes against the live account: 67 rows, Avro schema check clean (no extra
fields). Then `stage_external_sources` plus
`dbt build -s stg_edplan__njsmart_powerschool+ --target staging` gives PASS=3
WARN=0 ERROR=0, with 67 rows through both models and the union's
uniqueness test passing.

**Two follow-ups, neither in this PR.**

1. `scripts/dbt-sxs.py` does not exist. `docs/guides/sftp-integration.md` and
   `scripts/CLAUDE.md` both document it as the way to stage a source against the
   test bucket. I used a `cloud_storage_uri_base` var override instead, which
   does the same thing. Worth its own issue.
1. The Paterson `rpt_powerschool__autocomm_students_iep` extract, flagged as a
   possible follow-on in
   `docs/superpowers/specs/2026-07-14-paterson-autocomm-extracts-design.md`, is
   deliberately out of scope. Newark and Camden each opt in with a regional
   model plus Dagster extract config; Paterson has neither, so nothing writes
   back to its PowerSchool.

</details>


cbini:
## Review verdicts

### 1. Stale "NULL for Paterson" documentation — **accepted, fixed in `ce3d8374e3`**

Correct, and I missed it. I verified all 5 cited locations before changing anything; every quoted line was accurate:

| File | Lines | Claim |
| --- | --- | --- |
| `dim_student_iep_status.yml` | 7-9 | model description: Paterson derives a current-state span from PowerSchool |
| `dim_student_iep_status.yml` | 147, 153, 159 | code/name/placement "null for Paterson + Miami" |
| `dim_student_enrollment_status.yml` | 36, 42, 48 | same 3 columns, "NULL for Miami and Paterson" |
| `stg_powerschool__studentcorefields.yml` | 7 | "`dim_student_iep_status` (Paterson + Miami leg)" |
| `student_enrollment_status.yml` (Cube) | 43, 51, 59 | same 3 fields, `public: true` |

All 11 lines now say the NJ regions draw from EdPlan and Miami does not. I also swept the rest of `src/dbt/kipptaf` and `src/cube` for related region-scoped SPED claims — nothing else stale.

### 2. No regression test for `edplan_has_archive` — **declined, with reasoning**

A dbt unit test is the wrong instrument here. Unit tests mock a model's `ref`/`source` inputs and assert on output rows; this toggle decides at **compile** time whether a `source()` is referenced at all. With the var false the archive relation is absent from the compiled SQL, so there is no input to mock — a unit test would be asserting on a leg that does not exist in the graph.

What actually pins the behavior is the compiled artifact, which I checked per project:

```text
grep -c njsmart_powerschool_archive target/compiled/.../int_edplan__njsmart_powerschool_union.sql
  kipppaterson  0
  kippnewark    1
  kippcamden    1
```

The failure mode a test would guard against — a refactor re-coupling Paterson to the archive relation — surfaces loudly rather than silently: Paterson has no `kipppaterson_edplan.stg_edplan__njsmart_powerschool_archive` table, so re-coupling fails the build with "table not found". That is a louder signal than a warn-severity test in this repo, where `data_tests: +severity: warn`.

Happy to add one if a maintainer disagrees.

### 3. Devcontainer template unreviewable in the sandbox — **confirmed, no action**

Your inference is right. The change is exactly 2 lines, the district-suffixed EdPlan SFTP username and password entries matching the existing Camden and Newark shape. I can confirm because the script that applied it asserted the patch touched 1 file with +2/-0 lines and that both added lines carried the Paterson suffix, aborting otherwise.

### On the 2 dbt Cloud test warnings against `dim_student_iep_status`

Not raised in the review, but worth recording since they touch the changed model. Both warnings are the same `kippnewark` row, and neither comes from this branch:

- The 2 duplicate rows are byte-identical on every column including `student_enrollment_key` — an exact duplicate, not two distinct IEP legs.
- Root cause is upstream: prod `int_students__student_enrollment_union` holds duplicate `(project, student, academic_year, entrydate)` tuples (4 Newark, 1 Camden, 1 Paterson). `nj_leg` inner-joins enrollments, so those fan out. That model is unmodified here, so CI deferred it to prod.
- Ruling out this branch: Newark has **fewer** rows in the PR edplan union than prod (38,373 vs 38,501), not double; Camden likewise (15,622 vs 15,662). Appending the Paterson relation duplicated nothing. Paterson contributes 67 rows, 0 duplicate keys, 0 overlapping spans.

The change also **removes** a live production defect: prod `dim_student_iep_status` currently has a duplicate `kipppaterson` key (`1ec4f918cc2b95f502367e81103930f4`, 2 rows) from the `pm_leg` double-count. That key is absent from the PR build.

### Note on CI

The first 5 checks on this PR ran, but `Trunk Check`, `Claude Code Review`, and the `kipppaterson` Dagster deploy did not fire, because the PR was opened with a Codespace installation token and GitHub does not trigger workflow runs from events generated by it. A draft toggle re-fired them; all 3 then passed. The one `cancelled` `kipppaterson` run was superseded by the second batch, not a failure.

cbini:
## Follow-up: the sensor-cursor finding

I missed one item from the second review round and want to close it out rather than leave it silent.

**Finding: `build_edplan_sftp_sensor` stores `now.timestamp()` as its cursor instead of max matched mtime** (`src/teamster/libraries/edplan/sensors.py:36,78`), against the guidance in `src/teamster/CLAUDE.md`. **Accepted as a real pre-existing weakness. Declining to fix it in this PR, and correcting the direction of the Paterson-specific risk.**

The underlying issue is real: `cursor[asset_identifier] = now_timestamp` means a matched file whose `st_mtime` predates the previous tick's wall clock never satisfies `f.st_mtime > last_run` and is skipped permanently.

The Paterson-specific reasoning does not hold up, though. The claim was that Paterson is the first EdPlan region whose reports directory accumulates dated files rather than holding one perpetually-overwritten file, and that this is "the shape where clock-skew-driven cursor gaps would first bite." Reading the loop, the accumulating shape does not change that exposure:

- The skip depends only on the *new* file's mtime versus the last tick's wall clock. In both shapes the newly-arrived file has an mtime of roughly now, so both are equally exposed — and equally safe under normal clock behavior.
- Retained older files change behavior only in the opposite direction. On a cursor reset, `last_run` is `0`, so every retained dated file matches and the sensor emits one `RunRequest` per file, each with its own `partition_key`. That is an unintended backfill, not a gap. It is bounded by `start_date="2026-08-24"` and by the composed regex resolving to exactly one file per partition, so the blast radius today is small — but it is worth knowing it is a re-materialization risk rather than a skip risk.

So the shape observation is correct and the consequence is inverted from what was flagged.

Not fixing it here for two reasons. The factory is shared: `kippnewark` and `kippcamden` both call it, so changing the cursor semantics changes their production behavior in a PR scoped to onboarding Paterson. And the fix is not a one-liner — switching to max-matched-mtime needs care about the first tick and about files that match but are skipped for `st_size == 0`, or it silently advances past a file that was never materialized.

Happy to open a separate issue against `libraries/edplan/sensors.py` for it if a maintainer wants it tracked.

### Remaining items from that round

- **Human eyeball on the local template** — confirmed. Exactly 2 lines, the district-suffixed EdPlan SFTP username and password entries matching the Camden and Newark shape. The applying script asserted 1 file and +2/-0 lines with both added lines carrying the Paterson suffix, aborting otherwise.
- **dbt Cloud status, which that job could not query** — `success` on both commits. Run 1: 19 warnings. Run 2 (head): 17 warnings, and both `dim_student_iep_status` warnings are gone. That model now has 0 duplicate keys across 34,697 and 0 overlapping spans, confirming the 2 warnings in run 1 were a transient upstream artifact rather than anything this branch introduced.
- **Post-merge re-stage** — still the one action item after merge, and it is in the PR body.

Issue #5002, opened 2026-08-26
feat(dbt): establish what a Focus schedule start_date means, and resolve the 1,839 window mismatches


cbini:
## What and why

The larger half of #4970. Read that issue for the measured analysis.

1,839 Miami course enrollments attach to no student stint because the schedule row
falls outside the stint window. Of those, 1,383 have a `cc_dateenrolled` on or
after the stint's `exitdate` — the schedule row starts after the student left.

That is either a real data problem or a Focus dating convention, and the two call
for opposite fixes. Establish which before changing any join.

## The question

Does a Focus schedule `start_date` mean the day the student joined the section, or
the start of the course period's term?

If Focus dates a schedule row to the term start, a student who enrolls mid-term
gets a `cc_dateenrolled` before their own stint entry, and the overlap legitimately
fails. That would be a conforming problem in the Focus branch, not a join problem.

## Blocked on

#4987. `int_focus__schedule` currently nulls the marking period on 16,104 of
Miami's 19,636 AY2026 schedule rows, so 82% of rows have no term bounds to compare
the schedule dates against. That comparison is the whole method here.

## Constraints

Do not close this by widening the overlap predicate before the question is
answered. A looser join attaches rows to the wrong stint and produces a
confidently wrong `student_enrollment_key`, which is worse than a null one.

For Claude

Context, not instructions.

Relevant models:

- `src/dbt/focus/models/intermediate/int_focus__schedule.sql` — the source of
  `start_date` and `end_date`
- `src/dbt/kipptaf/models/students/intermediate/int_students__course_enrollments.sql`
  — the `focus_conformed` CTE, where `cc_dateenrolled` and `cc_dateleft` are set
- `src/dbt/kipptaf/models/marts/dimensions/dim_student_section_enrollments.sql` —
  the `enrollment_overlap` CTE holds the join

Constraints:

- NJ counts must not move. `cc_dateleft` is null on zero NJ rows, so any change
  keyed on that null is NJ-safe by construction; anything else needs measuring.
- A warn-severity test on the Miami orphan rate shipped with PR #4972. Update its
  threshold rather than deleting it if the rate legitimately changes.
- 10.5% orphaned is inside the normal band for this model, not outside it. Prod
  runs 18 to 20% on complete years. This is not an anomaly to drive to zero.

Acceptance criteria:

- The meaning of a Focus schedule `start_date` is established and documented in the
  model YAML.
[...]

Issue #4985, opened 2026-08-26
chore(dbt): working order for the Focus dbt modeling backlog


cbini:
## What and why

9 open issues are Focus dbt modeling work for Miami. Each is filed on its own,
with no shared order. This issue holds the order. Every item keeps its own issue
and its own PR.

Revised 2026-08-26. What changed is at the bottom.

### Scope

| Issue | Title | Kind |
| --- | --- | --- |
| #4987 | `int_focus__schedule` discards the full-year marking period | fix |
| #4973 | make the downstream consumers actually show Miami course enrollments | fix |
| #4970 | attach the remaining Miami course enrollments to a student stint | feat |
| #4968 | source Miami drop status for the course enrollment facts | feat |
| #4986 | wire Focus report card grades into the kipptaf network layer | feat |
| #4803 | Miami attendance facts orphaned from `dim_student_enrollments` | fix |
| #4600 | reshape `int_focus__student_enrollment` to the staging-plus-labels pattern | refactor |
| #4670 | replace hand-copied frozen-region exclusion filters with a shared var | refactor |
| #4795 | source an advisory assignment for Miami grades 7-8 | feat, ops-tracked |

### #4987 gates two issues

`int_focus__schedule` nulls the marking period on 16,104 of Miami's 19,636 AY2026
schedule rows, because Focus writes `marking_period_id = 0` to mean the full year
and the model reads that sentinel as missing data. All 5 marking-period columns go
null together, including `start_date` and `end_date`.

Both downstream issues need those term bounds:

- **#4970** has 1,839 course enrollments whose schedule row falls outside the
  student's stint. Telling a real mismatch from a Focus dating convention means
  comparing the schedule dates to their term. On a full-year row there is no term
  to compare to.
- **#4968** proposes inferring a drop from an `end_date` that lands before its
  marking period ends. That rule reads the marking period's `end_date` directly,
  so it cannot be written at all today.

#4987 is also the smallest item here, the fix is vendor-documented, and
`rpt_littlesis__enrollments` already contains a working reference implementation.
Do it first.

### #4968 has to land before half of #4973

#4973 has 2 independent causes. Only the first touches #4968.

- **Cause 1, the null-versus-false filter trap.** About 40 models filter with
  `where not is_dropped_section`, and `not null` is null, so every Miami row drops
  out. #4973 forbids resolving this by emitting `false`, because #4968 owns that
  decision. If #4968 lands first and yields non-null flags, the `coalesce` becomes
[...]

Issue #4290, opened 2026-07-01
fix(dagster): union_relations automation condition races upstream rebuild and freezes stale view


cbini:
## Summary

`dbt_union_relations_automation_condition()` (`src/teamster/core/automation_conditions.py:167`) can permanently freeze a `union_relations` view against its pre-deploy column set when the view materializes before its upstream table rebuilds with a changed schema. This stranded 5 prod `rpt_focus__*` extracts after PR #4285.

## Incident

PR #4285 changed the finalsite package model output schema (dropped `finalsite_status`, added `focus_student_id_prefixed`). After deploy, the kipptaf `union_relations` wrapper views compiled against the OLD Miami schema and never recompiled, so BigQuery view expansion failed at read time:

```text
Unrecognized name: finalsite_status; failed to parse view
'teamster-332318.kipptaf_finalsite.int_finalsite__enrollment_lifecycle'
```

Failed run: https://kipptaf.dagster.cloud/prod/runs/fdbcca17-6787-4ebc-8cda-95374084819b

Timeline (2026-07-01 UTC), from materialization `code_version` / `input_data_version` tags on `int_finalsite__enrollment_lifecycle`:

| Time  | Asset          | code_version | Note |
| ----- | -------------- | ------------ | ---- |
| 17:18 | kipptaf wrapper | 047a07ab (new) | read OLD Miami version, compiled stale |
| 17:20 | kipptaf wrapper | 047a07ab (new) | read OLD Miami version again |
| 17:23 | Miami lifecycle | 3f6cbc2f (new) | first NEW-schema rebuild |
| 17:36 | Miami lifecycle | 3f6cbc2f (new) | current |

The wrapper materialized ~5 min BEFORE its upstream rebuilt, and never re-ran.

## Root cause

The condition triggers on ancestor `code_version_changed` (not `any_deps_updated`, deliberately, to save credits) and resets that trigger via `.since(newly_updated())` when the wrapper itself materializes.

1. Ancestor `code_version_changed` is true only during the deploy window: from deploy until the ancestor rematerializes with the new code version (which clears it).
2. At deploy the wrapper had BOTH its own code-version bump (partly from the commit `4020abcf` doc-comment intended to force a rebuild) and a changed ancestor code version, so its trigger fired.
[...]

PR #5006, opened 2026-08-26
fix(dbt): resolve the Focus full-year marking-period sentinel in int_focus__schedule

### LINKED ISSUE #4987: fix(dbt): int_focus__schedule discards the full-year marking period instead of resolving it
## What's happening

`int_focus__schedule` nulls the marking period on 82% of Miami's AY2026 schedule
rows, so those rows carry no term at all.

Focus writes `marking_period_id = 0` on a schedule row to mean **the full year.**
It is a sentinel, not missing data, and it resolves to the `marking_periods` row
whose `type = 'year'`. The model treats it as missing instead:

`nullif(marking_period_id, 0) as marking_period_id` at
[`int_focus__schedule.sql:27`](https://github.com/TEAMSchools/teamster/blob/main/src/dbt/focus/models/intermediate/int_focus__schedule.sql#L27),
then a left join on the nulled id.

Because 5 output columns come from that join, all 5 go null together on every
affected row.

| Measure, AY2026 | Rows |
| --- | --- |
| `stg_focus__schedule` rows | 19,636 |
| Of those, `marking_period_id = 0` | **16,104 (82%)** |
| Already null before the `nullif` | 0 |
| `int_focus__schedule` null `marking_period_id` | 16,104 |
| Null `marking_period_title`, `_short_name`, `_type`, `_start_date`, `_end_date` | 16,104 each |
| `marking_periods` rows with `type = 'year'` | 10, one per school |

Every null traces to the `nullif`, and the row it should resolve to exists for
all 10 schools.

## This is already known, and already fixed in the wrong place

`rpt_littlesis__enrollments` works around it locally. Its comment names the
problem outright:

```text
Full-year course periods carry Focus's marking_period_id = 0 sentinel,
which int_focus__schedule normalizes to null. Little SIS gets the
school's year-level marking period for those instead of a null term, so
every row carries a term the way the PowerSchool branch does.
```

It builds a `focus_year_terms` CTE from `stg_focus__marking_periods` filtered to
`type = 'year'`, then does
`coalesce(sch.marking_period_id, fyt.marking_period_id) as term_id`.

So one consumer out of six resolves the sentinel correctly. The other five get
nulls. The resolution belongs in the model that owns the grain.

## Steps to reproduce

1. Count the sentinel at the source: rows in
   `kippmiami_focus.stg_focus__schedule` where `syear = 2026` and
   `marking_period_id = 0`. Expect 16,104 of 19,636.
1. Count the loss downstream: rows in `kippmiami_focus.int_focus__schedule` where
   `academic_year = 2026` and `marking_period_start_date is null`. Expect the same
   16,104.
1. Confirm the resolution target exists: rows in
   `kippmiami_focus.stg_focus__marking_periods` where `syear = 2026` and
   `type = 'year'`. Expect 10, one per school.

## Where

- **Code location / dbt project:** `focus` package, consumed by `kipptaf`
- **Environment:** prod

For Claude

Context, not instructions.

### The vendor documents the join

Focus SIS Level 1 Certification, Day 2, notes on `student_report_card_grades`:

```sql
JOIN marking_periods mp
  ON mp.syear = cp.syear
 AND mp.school_id = cp.school_id
 AND (CASE WHEN mp.type = 'year' THEN '0' ELSE mp.marking_period_id END)
     = cp.marking_period_id
```

The same doc gives the pattern a second time in its Day 4 cron-email example,
joining `schedule` directly to `marking_periods` on `mp.type = 'year'`. Note that
the join is scoped on **both** `syear` and `school_id` — the current left join in
`int_focus__schedule` matches on `marking_period_id` alone.

### Blast radius

Consumers of `int_focus__schedule`:

- `int_focus__report_card_grades` (focus package)
- kipptaf wrapper `int_focus__schedule`, and through it
  `int_students__course_enrollments`, `int_students__course_sections`,
  `int_students__teacher_grade_levels`, `int_focus__advisory`,
  `rpt_littlesis__enrollments`

`int_students__course_enrollments` and `int_focus__advisory` do not read the
marking-period columns today, so nothing downstream is currently emitting a wrong
value. The harm is that the term context is unavailable to any consumer that wants
it, and one consumer had to rebuild it.

### Why this gates #4970

#4970 has 1,839 Miami course enrollments whose schedule row falls outside the
student's enrollment stint, and it cannot tell a real mismatch from a Focus dating
convention without comparing the schedule dates to their term bounds. On a
full-year row there are no term bounds to compare against right now. Fixing this
supplies them for 82% of the AY2026 rows.

### Constraints

- Resolve the sentinel; do not drop the `nullif` and leave a raw `0` in the
  output. A literal `0` is not a real `marking_period_id` and would orphan the
  join.
- Scope the join on `syear` and `school_id` as well as the id, per the vendor
  pattern. 10 schools each have their own `type = 'year'` row.
- Keep a way to tell a resolved full-year term from a natively-termed row. The
  schedule's own `mp` column already carries the term code, 3 distinct values in
  AY2026 — check whether it already serves this before adding a flag.
- Delete the `focus_year_terms` workaround from `rpt_littlesis__enrollments` in
  the same change, and confirm its `term_id` output does not move.
- Miami is the only region on Focus, so there is no NJ parity risk here. Verify
  the AY2025-and-earlier Focus rows anyway, since the sentinel predates AY2026.

### Acceptance criteria

- `int_focus__schedule` AY2026 rows with a null `marking_period_start_date` drops
  from 16,104 toward 0.
- Every resolved row's marking period belongs to the same school and `syear` as
  the schedule row.
- `rpt_littlesis__enrollments.term_id` is row-identical to prod, with the local
  workaround removed.
- Row count of `int_focus__schedule` unchanged at 19,636 for AY2026, proving the
  join does not fan out.

Refs #4970
Refs #4986

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will give every Miami full-year course schedule
row its marking period back.

Focus writes `marking_period_id = 0` on a schedule row to mean "the full year."
It is a code, not missing data, and it points at the school's year-level marking
period. `int_focus__schedule` read it as missing and threw it away. Because 5
output columns all come from that one join, all 5 went null together on 16,119
of Miami's 19,649 AY2026 schedule rows — 82% of them.

This joins the marking period the way the Focus vendor documents it, so the code
resolves instead of being discarded. Measured against prod:

| Measure, AY2026 | Before | After |
| --- | --- | --- |
| Rows with no marking-period dates | 16,119 | **0** |
| Row count | 19,649 | 19,649 |
| Distinct schedule keys | 19,649 | 19,649 |
| Resolved period in a different school or year | — | 0 |

One report, `rpt_littlesis__enrollments`, had already built this fix for itself
in a local block. That block is now deleted. Its `term_id` output does not move.

## Reviewer Notes

- **The join gained two conditions, `syear` and `school_id`.** Each of the 10
  schools has its own year-level marking period, so matching on the id alone is
  not enough. All 3,530 rows that already matched still match — the extra scope
  drops nothing.
- **Telling the two kinds of row apart still works.** The schedule's own `mp`
  column reads `FY` on exactly the 16,119 rows that carry the code, so no new
  flag was needed.
- **One change is outside the issue.** The kipptaf wrapper picked which
  duplicate row to keep partly by asking whether `marking_period_id` was null,
  which meant "prefer the full-year row." Nothing is null now, so that question
  is dead. It now asks `mp is distinct from 'FY'`, which means the same thing.
  It decides none of the 299 duplicate pairs either way.
- **A unit test guards the fix.** It puts a decoy year-level marking period on a
  second school, so it fails if the `school_id` scope is ever dropped.

### General

### dbt

Only the first item applies. This pull request adds no model, adds no column,
renames nothing, and touches no external source.

- [ ] Include (or update) an
      summary above.
      See

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. This is P1 of the Focus dbt modeling backlog
(#4985). Value-only across both projects, no column-set change, so it is not
exposed to the #4290 deploy race and ships as one PR.

### The join

```sql
with
    marking_periods as (
        select
            marking_period_id,
            syear,
            school_id,
            ...
            if(type = 'year', 0, marking_period_id) as schedule_marking_period_id,
        from {{ ref("stg_focus__marking_periods") }}
    )
...
left join
    marking_periods as mkp
    on s.syear = mkp.syear
    and s.school_id = mkp.school_id
    and s.marking_period_id = mkp.schedule_marking_period_id
```

This is the vendor's `student_report_card_grades` pattern from Focus SIS Level 1
Certification, Day 2, quoted on #4987. The sentinel expression sits in a CTE
rather than the `ON` clause, per the "no one-sided calculations in join
predicates" rule; the two forms are value-identical against prod, 19,649 rows
each with `except distinct` returning 0 both ways.

The output column is `mkp.marking_period_id`, not `s.marking_period_id`, so the
raw `0` can never reach a consumer. The `nullif` and the `student_schedule` CTE
that held it are both gone; the final select already enumerated every column, so
that CTE was a redundant projection.

### Verification against prod, not the dev build

The dev build is not the evidence here. It confirms the SQL compiles and
materializes 19,649 rows with 0 null `marking_period_start_date` and 19,649
distinct keys, but `zz_cbini_kipptaf_extracts.rpt_littlesis__enrollments` built
only 5,711 of prod's 18,878 Focus rows — stale personal copies of upstream
staging models shadowed `--defer`. Every number below comes from running the old
and new logic side by side against prod staging tables.

Model level, AY2026:

| Measure | Value |
| --- | --- |
| Rows | 19,649, unchanged |
| Distinct `student_schedule_id` | 19,649 — no fan-out |
| Null `marking_period_start_date` | 16,119 to 0 |
| Resolved period in a different `school_id` or `syear` | 0 |
| `mp = 'FY'` rows resolving to `type = 'year'` | 16,119 of 16,119 |
| Non-FY rows resolving to `type = 'year'` | 0 |

`term_id`, old expression versus new, through the wrapper dedupe, over the full
prod population: 19,350 rows each side, `except distinct` returns 0 in both
directions, 0 nulls. That covers all 18,878 prod Focus rows, which the dev build
could not.

### Three checks the issue's reproduce steps did not cover

1. **`mp` is a clean discriminator.** `mp = 'FY'` matches the sentinel exactly,
   16,119 for 16,119, so #4987's "check whether it already serves this before
   adding a flag" resolves to yes.
1. **No fan-out risk in any year.** Exactly one `type = 'year'` row per
   (`syear`, `school_id`) across `syear` 1980 to 2026.
1. **No earlier-year Focus rows to verify.** `stg_focus__schedule` holds only
   `syear = 2026`, so #4987's "verify AY2025 and earlier" is vacuous rather than
   passing.

Prod counts run today are 13 rows above #4987's, which was written against an
earlier snapshot: 19,649 not 19,636, and 16,119 not 16,104. Same ratio, same
conclusion.

### The unit test

`test_int_focus__schedule_full_year_sentinel_resolves` mocks 2 schedule rows —
one carrying the 0 sentinel, one term-specific — against 3 marking periods: the
school's year row, a quarter row, and a decoy year row on a second school.
Removing `and s.school_id = mkp.school_id` makes it fail, which I ran rather
than assumed.

### Warnings seen locally, both pre-existing

- `relationships_stg_focus__schedule_student_id` (475) and
  `relationships_stg_focus__co_teachers_course_period_id` (4) are stale-dev
  `--defer` drift. Prod has 0 orphans for both.
- `test_focus_course_sections_teacher_resolves` warns 80. The same query run
  entirely against prod also returns 80.

### The first dbt Cloud failure was not this change

Run 70403222330862 failed on `unique_int_kippadb__roster_student_number` with
`Not found: Table zz_stg_kipptaf_edplan.int_edplan__njsmart_powerschool_union`.
That shared staging table was created at 18:59:29 with 54,163 rows, after the
test ran. `int_kippadb__roster` reaches it through
`int_extracts__student_enrollments` and landed in the build set only because this
change sits far upstream. The re-run passed with no code change to that path.

### Blast radius

`rpt_littlesis__enrollments` is the only consumer that reads any marking-period
column. `int_students__course_enrollments`, `int_students__course_sections`,
`int_students__teacher_grade_levels`, and `int_focus__advisory` reference none of
them, so nothing else changes value. `int_focus__report_card_grades` joins
marking periods off its own key, not off the schedule.

`marking_period_id` moved into the `mkp` column group in the second commit, on
review. Consumers resolve columns by name — `dbt_utils.union_relations` builds
its list from the relation schema, and `rpt_littlesis__enrollments` names every
column it reads — so the position carried no weight.

### Gates #4970

#5002 needs schedule dates comparable to their term bounds. This supplies real
`marking_period_start_date` and `marking_period_end_date` on 82% of Miami's
AY2026 schedule rows, which previously had none.

Refs #4985
</details>

### DIFF
--- a/src/dbt/focus/models/intermediate/int_focus__schedule.sql
+++ b/src/dbt/focus/models/intermediate/int_focus__schedule.sql
@@ -1,3 +1,19 @@
+with
+    marking_periods as (
+        select
+            marking_period_id,
+            syear,
+            school_id,
+            title,
+            short_name,
+            type,
+            start_date,
+            end_date,
+
+            if(type = 'year', 0, marking_period_id) as schedule_marking_period_id,
+        from {{ ref("stg_focus__marking_periods") }}
+    )
+
 select
     s.id as student_schedule_id,
     s.syear as academic_year,
@@ -5,9 +21,6 @@ select
     s.student_id,
     s.course_id,
     s.course_period_id,
-    -- resolved, never the raw sentinel: mkp supplies the school's year-level id
-    -- on a full-year row and the schedule's own id on a term-specific one
-    mkp.marking_period_id,
     s.mp,
     s.course_weight,
     s.days,
@@ -48,6 +61,8 @@ select
     c.course_hours,
     c.homeroom,
 
+    -- aliased mkp, not mp: schedule has its own mp column (the FY/SEM/QTR code)
+    mkp.marking_period_id,
     mkp.title as marking_period_title,
     mkp.short_name as marking_period_short_name,
     mkp.type as marking_period_type,
@@ -59,18 +74,8 @@ inner join
     {{ ref("stg_focus__course_periods") }} as cp
     on s.course_period_id = cp.course_period_id
 inner join {{ ref("stg_focus__courses") }} as c on s.course_id = c.course_id
-/*
-   Focus writes marking_period_id = 0 on a schedule row to mean the full year.
-   It resolves to the school's type = 'year' marking period, per the vendor's
-   documented join. Scoped on syear and school_id as well as the id because
-   each school gets its own year-level row.
-
-   Aliased mkp, not mp: schedule has its own mp column (the FY/SEM/QTR code),
-   which is how a resolved full-year row stays distinguishable from a
-   natively-termed one.
-*/
 left join
-    {{ ref("stg_focus__marking_periods") }} as mkp
+    marking_periods as mkp
     on s.syear = mkp.syear
     and s.school_id = mkp.school_id
-    and s.marking_period_id = if(mkp.type = 'year', 0, mkp.marking_period_id)
+    and s.marking_period_id = mkp.schedule_marking_period_id
--- a/src/dbt/focus/models/intermediate/int_focus__schedule.sql
+++ b/src/dbt/focus/models/intermediate/int_focus__schedule.sql
@@ -1,33 +1,3 @@
-with
-    student_schedule as (
-        select
-            id,
-            syear,
-            school_id,
-            student_id,
-            course_id,
-            course_period_id,
-            mp,
-            course_weight,
-            days,
-            rotation_days,
-            start_date,
-            end_date,
-            fefp_number,
-            dual_enrollment_indicator,
-            class_minutes_weekly,
-            reading_intervention_component,
-            basic_skills_exam,
-            location_of_student,
-            eoc_exam_term,
-            exempt_from_total_clock_hours,
-            exclude_from_fte,
-            pmrn,
-
-            nullif(marking_period_id, 0) as marking_period_id,
-        from {{ ref("stg_focus__schedule") }}
-    )
-
 select
     s.id as student_schedule_id,
     s.syear as academic_year,
@@ -35,7 +5,9 @@ select
     s.student_id,
     s.course_id,
     s.course_period_id,
-    s.marking_period_id,
+    -- resolved, never the raw sentinel: mkp supplies the school's year-level id
+    -- on a full-year row and the schedule's own id on a term-specific one
+    mkp.marking_period_id,
     s.mp,
     s.course_weight,
     s.days,
@@ -82,12 +54,23 @@ select
     mkp.start_date as marking_period_start_date,
     mkp.end_date as marking_period_end_date,
 
-from student_schedule as s
+from {{ ref("stg_focus__schedule") }} as s
 inner join
     {{ ref("stg_focus__course_periods") }} as cp
     on s.course_period_id = cp.course_period_id
 inner join {{ ref("stg_focus__courses") }} as c on s.course_id = c.course_id
--- aliased mkp, not mp: schedule has its own mp column (the FY/SEM/QTR term code)
+/*
+   Focus writes marking_period_id = 0 on a schedule row to mean the full year.
+   It resolves to the school's type = 'year' marking period, per the vendor's
+   documented join. Scoped on syear and school_id as well as the id because
+   each school gets its own year-level row.
+
+   Aliased mkp, not mp: schedule has its own mp column (the FY/SEM/QTR code),
+   which is how a resolved full-year row stays distinguishable from a
+   natively-termed one.
+*/
 left join
     {{ ref("stg_focus__marking_periods") }} as mkp
-    on s.marking_period_id = mkp.marking_period_id
+    on s.syear = mkp.syear
+    and s.school_id = mkp.school_id
+    and s.marking_period_id = if(mkp.type = 'year', 0, mkp.marking_period_id)
--- a/src/dbt/focus/models/intermediate/properties/int_focus__schedule.yml
+++ b/src/dbt/focus/models/intermediate/properties/int_focus__schedule.yml
@@ -6,8 +6,9 @@ models:
       as stg_focus__schedule. Course period and course are inner-joined because
       a schedule row without them is not interpretable; stg_focus__schedule
       carries relationships tests on both keys so an orphan fails there rather
-      than silently disappearing here. Marking period is left-joined and is null
-      for full-year schedules, which Focus records with a 0 marking-period id.
+      than silently disappearing here. Marking period is left-joined on syear
+      and school_id as well as the id, resolving Focus's full-year sentinel
+      (marking_period_id 0) to the school's year-level marking period.
       Internal-only — a rpt_ view must sit between this model and any external
       consumer.
     columns:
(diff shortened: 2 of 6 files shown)


cbini:
## Review response — all 6 findings accepted, all 6 fixed in ff34bbe

I checked each rule citation against the guide text and `git grep` before acting
on it. All 6 quote `src/dbt/CLAUDE.md` accurately.

| # | Verdict | What changed |
| --- | --- | --- |
| 1 | Accepted | `stg_focus__schedule.yml` now describes the resolution instead of the old null-normalizing |
| 2 | Accepted | The sentinel expression moved into a `marking_periods` CTE as `schedule_marking_period_id`; the `ON` clause matches plain columns |
| 3 | Accepted | `mkp.marking_period_id` moved into the `mkp` group |
| 4 | Accepted | The 9-line block comment is gone; only the "aliased `mkp`, not `mp`" line stays |
| 5 | Accepted | The tiebreak reads `mp is distinct from 'FY'` |
| 6 | Accepted | Added `test_int_focus__schedule_full_year_sentinel_resolves` |

### On #2, the hoisted expression

The two forms are value-identical. Run side by side against prod, each returns
19,649 rows and `except distinct` returns 0 in both directions, over
`(id, marking_period_id, start_date, end_date, type)`.

### On #5, the null check you asked for

`select countif(mp is null) from stg_focus__schedule` returns **0 of 19,649**. So
no row hits this today. `mp` carries no `not_null` test, though, so the null
ordering was reachable in principle — `is distinct from` closes it either way.

### On #6, the test earns its place

It mocks 2 schedule rows, one carrying the 0 sentinel and one term-specific,
against 3 marking periods: the school's year row, a quarter row, and a **decoy
year row belonging to a second school**. The decoy is the point. I removed
`and s.school_id = mkp.school_id` and confirmed the test fails, then restored it
— a test that passes both ways would be worth nothing.

### On your scope note

Correct that this job had no warehouse access. The row counts in the fold-out
come from running the old and new logic side by side against prod staging tables,
not from the dev build — the dev build of `rpt_littlesis__enrollments` reached
only 5,711 of prod's 18,878 Focus rows, because stale personal copies of upstream
staging models shadowed `--defer`.

Issue #4981, opened 2026-08-26
fix(clever): grant School Leader in Residence staff access to Paterson schools


ldesimon:
## Summary

Staff holding the "School Leader in Residence" role, based at Newark's "Room 9" central office location, need Clever access to Paterson schools in addition to their home Newark schools. `rpt_clever__staff` has no branch that expresses "home region plus one additional region" — every existing assignment path is either single-school, CMO-wide, home-region-wide, or driven by the `Group Staff NJ Regional` AD group (which would also grant Camden, not requested).

## Root cause

In `src/dbt/kipptaf/models/extracts/clever/rpt_clever__staff.sql`, the `assignments` CTE has four branches (single school, CMO-wide T&L/EDs/Data, home-region-wide via `home_work_location_powerschool_school_id = 0`, all-NJ via LDAP group). None cover "Newark home region + Paterson" without also granting Camden.

## Fix

Add a fifth branch to `assignments` that grants Newark + Paterson schools to any staff member with `job_title = 'School Leader in Residence'` and `home_work_location_reporting_name = 'Room 9'`.

```sql
union all

/* School Leader in Residence (Room 9, Newark) also covers Paterson schools */
select
    sr.powerschool_teacher_number,
    sr.user_principal_name,
    sr.given_name,
    sr.family_name_1,
    sr.home_department_name,
    sr.sam_account_name,

    sch.school_id,
from staff_roster as sr
inner join
    schools as sch
    on sch.dagster_code_location in ('kippnewark', 'kipppaterson')
where
    sr.job_title = 'School Leader in Residence'
    and sr.home_work_location_reporting_name = 'Room 9'
```

## Scope verified

Queried prod `int_people__staff_roster`: exactly 3 active, non-prestart employees currently match this criteria (all home region kippnewark). The existing "all region" branch already grants these 3 all Newark schools; `select distinct` at the end of the model absorbs that overlap, so this change only adds Paterson access for those 3 — no other staff affected.

## Source systems

Clever, PowerSchool, ADP/LDAP (`int_people__staff_roster`) — kipptaf, kippnewark, kipppaterson.

PR #5007, opened 2026-08-26
refactor(clever): replace 12 hardcoded region filters with one shared macro

### LINKED ISSUE #4670: refactor(dbt): replace hand-copied frozen-region exclusion filters in current-state extracts with a shared var

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make adding or removing a region from the
six Clever feeds a one-line edit.

The exclusion that keeps Miami out of the Clever feeds was written out 12 times
across the six `rpt_clever__*` models, in two different syntaxes. #4653 happened
because those copies drifted apart. #4667 fixed the drift but left all 12
copies in place, so the same bug stayed available.

Now there is one list and one macro:

- `frozen_powerschool_code_locations: [kippmiami]` — a var in
  `src/dbt/kipptaf/dbt_project.yml`. This is the one place to edit. It names the
  PowerSchool instance, not the region: Miami itself is not frozen, it moved to
  Focus in AY2026 and rosters into Clever from there. Only its PowerSchool stops
  at AY2025, and that is why the PowerSchool-derived extracts drop it.
- `exclude_frozen(column)` — a macro in `src/dbt/kipptaf/macros/utils.sql`, next
  to the existing `extract_source_project`.

Every one of the 12 sites now reads
`and {{ exclude_frozen("cc._dbt_source_project") }}`.

**Why option 2 (a macro) and not option 1 (a var alone).** A var alone cannot
do the job. A dbt var holding a list renders as `['kippmiami']`, which
BigQuery rejects, so every call site needs Jinja to turn the list into
`('kippmiami')`. That Jinja is the macro. A var alone also cannot reconcile the
two syntaxes the issue names, because one is a `LIKE` on a relation name and
the other is an equality on a code-location column. The var is still the
one-place edit — the macro just reads it.

**The two syntaxes collapse into one.** `_dbt_source_project` is defined as
`regexp_extract(_dbt_source_relation, r'(kipp\w+)_')`, so it already holds
exactly the code location. The seven
`_dbt_source_relation not like '%kippmiami%'` sites therefore become the same
`not in (...)` form the other five already used. One syntax network-wide, and
it matches the `_dbt_source_project != 'kippmiami'` idiom the rest of the
project already prefers.

The change also deletes the 12 duplicated
`-- Miami rosters into Clever directly from Focus` comments. The reason now
sits once, on the var.

The wider audit the issue asks for is below, under "Reviewer Notes". Per the
issue's scope control it is a report only — no models outside
`models/extracts/clever/` changed.

## Reviewer Notes

**One region literal stays in `rpt_clever__staff`, on purpose.** Line 157 reads
`on sch.dagster_code_location in ('kippnewark', 'kipppaterson')`. That is the
School Leader in Residence rule from #4981 — a Newark-based leader who also
covers Paterson schools. It is an inclusion for one job title, not a
frozen-PowerSchool exclusion, so the shared var cannot express it and should not
try.

**The audit found 68 more region literals in `kipptaf`, and they are not all
the same thing.** Sorting them matters more than counting them, because
pointing the wrong ones at this var would break reporting.

Adopt next — same policy, same shape, 19 sites in 12 models:

| feed | models | sites |
| --- | --- | --- |
| `rpt_illuminate__*` | courses, roles, sites, terms, users | 9 |
| `rpt_parentsquare__*` | emergency_contacts, parents, rosters, schools, sections, staff, students | 10 |

Both are current-state rostering feeds excluding Miami for the same reason
Clever does. Both are also feed *sets* that have to agree with each other, so
they carry the exact #4653 bug class. `rpt_illuminate__roles` and
`rpt_illuminate__users` even use the same two syntaxes this PR just merged.
Illuminate and ParentSquare are the highest-value follow-ups.

Adopt with care — 16 sites in 14 models: the shared intermediates
(`int_students__students`, `__schools`, `__terms`, `__student_enrollment_union`,
`__student_core_fields`, `__student_user_fields`, `__teacher_grade_levels`,
`int_powerschool__gradebook_assignment_scores_rollup`,
`stg_people__student_logins`, `int_extracts__gradebook_audit_student_flags`,
`int_tableau__fresh_enrollment_scaffold`, `rpt_tableau__gradebook_audit`,
`rpt_tableau__okrts_behavior`, `rpt_deanslist__missing_assignments`). Same
reason as Clever, but these sit on the shared spine, so one edit moves many
marts. Worth its own PR with its own row-count checks.

Do not adopt — 22 sites: everything spelled `region != 'Miami'`. Two reasons.
The column holds `Miami`, not `kippmiami`, so the var does not fit. More
importantly, several of these exclude Miami because Miami is in Florida, not
because its PowerSchool is frozen: `rpt_tableau__nj_school_register`,
`rpt_gsheets__nj_state_test_roster`, `rpt_gsheets__njsmart_transfer_unverified`
and `int_topline__state_assessments_weekly` are New Jersey state reporting.
Pointing them at this var would silently pull Miami back into NJ reporting on
the day Miami's PowerSchool unfreezes. These should stay literal, or get a
separate `nj_only` list.

Separate list — 9 sites: the Paterson gates (`region != 'Paterson'`,
`home_work_location_dagster_code_location != 'kipppaterson'`) in
`rpt_tableau__sight_words_dashboard`, `rpt_tableau__assessment_dashboard`,
`rpt_schoolmint_grow__users`, `rpt_gsheets__historical_suspensions`,
`rpt_gsheets__athletic_eligibility`, `int_students__athletic_eligibility` and
two `int_topline__*` models. "Paterson is not live yet" is a different policy
from "this region's PowerSchool is frozen". If those get consolidated they need
their own var, not this one.

One bug found while auditing (the last 2 of the 68): `rpt_tableau__crdc_roster`
lines 181 and 247 filter with
`regexp_extract(_dbt_source_relation, r'(kipp\w+)_') != 'kippmiami'` — a
function applied to a column inside `WHERE`, which the repo's dbt conventions
forbid. `_dbt_source_project` is the same value as a plain column. Not fixed
here; it is outside this PR's scope.

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Verdicts on
      both Minor findings are posted as a PR comment — one declined with
      reasoning, one agreed with no action needed.

### dbt

- [x] No new models, so no new properties files. The 6 models and their
      properties files are unchanged apart from the filter lines.
- [x] No exposure change. These views feed the Dagster BigQuery-to-SFTP extract
      assets, not a dashboard.
- [x] **Not a breaking change.** No column is renamed, removed, or retyped. Only
      `WHERE` clauses changed, and they select the same rows — see the proof
      below. Rollback is a plain revert.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes. This PR is kipptaf-only. `dbt_project.yml` and
      `macros/utils.sql` are project-wide files, so `state:modified+` will
      select far more than the 6 models.
      `src/dbt/`.

<details>
<summary>For Claude</summary>

**Claude-authored, human-directed.** Charlie picked P5 off the #4985 backlog and
set the scope: 6 clever models only, pick between the issue's options 1 and 2,
report the wider audit rather than expanding the PR, and leave the powerschool
`union_relations` views alone. Charlie also caught the first var name,
`frozen_code_locations`, as asserting something false about Miami; it became
`frozen_powerschool_code_locations`. Claude chose option 2, verified
equivalence, and ran everything below.

**Proof the predicate swap keeps the same rows.** The risk in this refactor is
the 7 sites that moved from `_dbt_source_relation not like '%kippmiami%'` to
`_dbt_source_project not in ('kippmiami')`. Both predicates were counted
against every affected prod upstream:

| upstream | total rows | old kept | new kept | delta | null `_dbt_source_project` |
| --- | --: | --: | --: | --: | --: |
| `stg_powerschool__cc` | 868,495 | 769,630 | 769,630 | 0 | 0 |
| `stg_powerschool__students` | 30,909 | 26,964 | 26,964 | 0 | 0 |
| `stg_powerschool__schools` | 32 | 27 | 27 | 0 | 0 |
| `base_powerschool__sections` | 33,663 | 29,407 | 29,407 | 0 | 0 |
| `int_extracts__student_enrollments` | 134,291 | 124,216 | 124,216 | 0 | 0 |

1,067,390 rows, zero delta, no nulls. `_dbt_source_project` cannot be null
here: `dbt_utils.union_relations` always populates `_dbt_source_relation`, and
`extract_source_project` derives the project from it.

`base_powerschool__sections` needed a second look because it passes through
`int_students__course_sections`, which unions PowerSchool and Focus. Focus rows
carry `_dbt_source_relation` = `...kippmiami_focus...` and
`_dbt_source_project` = `kippmiami`, so both predicates drop them. The delta of
0 above confirms it on real data.

**Build and tests**, deferred to the prod manifest:

```text
Done. PASS=11 WARN=1 ERROR=0 SKIP=0 NO-OP=0 TOTAL=12
```

`rpt_clever__school_id_resolves_to_schools_feed` passes at `severity: error`.
`rpt_clever__enrollments__student_id_resolves_to_students_feed` warns with 3
orphan student ids locally and 4 in dbt Cloud CI. Those are pre-existing and
unrelated: the same orphans come back from the same query against the prod views
right now, and that test is deliberately `severity: warn` for the vintage-gap
reason its own `tests/properties.yml` description spells out. The zero row delta
above means this change cannot alter feed membership, so it cannot add an
orphan. Ids withheld here — they are student numbers.

**One-place-edit proof.** Compiling with
`--vars '{frozen_powerschool_code_locations: [kippmiami, kipppaterson]}'` renders
every site as `not in ('kippmiami', 'kipppaterson')`, with no other edit.

**Macro name.** `exclude_frozen` is short on purpose.
`sr.home_work_location_dagster_code_location` at 12 spaces of indent leaves 19
characters for the macro name before the 88-column limit, so
`exclude_frozen_powerschool` would force sqlfmt to wrap that call across 5 lines
— harder to read than the literal it replaces. The macro's doc comment names
PowerSchool in its first line instead.

**Lint.** `trunk check --force` over all changed files: no issues. sqlfmt wanted
one `WHERE` in `rpt_clever__schools.sql` collapsed onto one line; that fix was
applied by hand before committing, then re-checked clean.

</details>

Refs #4670

### DIFF
--- a/src/dbt/kipptaf/dbt_project.yml
+++ b/src/dbt/kipptaf/dbt_project.yml
@@ -35,11 +35,12 @@ vars:
   # fresh-dashboard skill's "Update the Finalsite recruitment year" procedure
   # before bumping this.
   finalsite_recruitment_year: 2026
-  # Code locations that no longer roster through PowerSchool current-state, so
-  # the current-state extracts exclude them. Miami moved to Focus in AY2026 and
-  # rosters into Clever from there. Historical models must NOT read this var --
+  # Code locations whose PowerSchool instance is frozen. Miami itself is not
+  # frozen -- it moved to Focus in AY2026 and rosters into Clever from there --
+  # but its PowerSchool stops at AY2025, so extracts built on PowerSchool
+  # current-state have to drop it. Historical models must NOT read this var:
   # Miami's frozen PowerSchool history has to stay readable. See #4670.
-  frozen_code_locations: [kippmiami]
+  frozen_powerschool_code_locations: [kippmiami]
 
 models:
   kipptaf:
--- a/src/dbt/kipptaf/dbt_project.yml
+++ b/src/dbt/kipptaf/dbt_project.yml
@@ -35,6 +35,11 @@ vars:
   # fresh-dashboard skill's "Update the Finalsite recruitment year" procedure
   # before bumping this.
   finalsite_recruitment_year: 2026
+  # Code locations that no longer roster through PowerSchool current-state, so
+  # the current-state extracts exclude them. Miami moved to Focus in AY2026 and
+  # rosters into Clever from there. Historical models must NOT read this var --
+  # Miami's frozen PowerSchool history has to stay readable. See #4670.
+  frozen_code_locations: [kippmiami]
 
 models:
   kipptaf:
--- a/src/dbt/kipptaf/macros/utils.sql
+++ b/src/dbt/kipptaf/macros/utils.sql
@@ -8,10 +8,12 @@
     initcap(regexp_extract({{ table }}._dbt_source_project, r'kipp(\w+)'))
 {% endmacro %}
 
-{# Current-state exclusion for regions whose SIS is frozen. `column` is any
-   code-location column -- `_dbt_source_project` on a union model, or
-   `dagster_code_location` / `home_work_location_dagster_code_location` on the
-   staff roster. Add or remove a region in the frozen_code_locations var. #}
+{# Drops code locations whose PowerSchool instance is frozen, for extracts
+   built on PowerSchool current-state. `column` is any code-location column --
+   `_dbt_source_project` on a union model, or `dagster_code_location` /
+   `home_work_location_dagster_code_location` on the staff roster. Add or remove
+   a region in the frozen_powerschool_code_locations var. #}
 {% macro exclude_frozen(column) -%}
-    {{ column }} not in ('{{ var("frozen_code_locations") | join("', '") }}')
+    {%- set locations = var("frozen_powerschool_code_locations") -%}
+    {{ column }} not in ('{{ locations | join("', '") }}')
 {%- endmacro %}
--- a/src/dbt/kipptaf/macros/utils.sql
+++ b/src/dbt/kipptaf/macros/utils.sql
@@ -7,3 +7,11 @@
 {% macro extract_region(table) %}
     initcap(regexp_extract({{ table }}._dbt_source_project, r'kipp(\w+)'))
 {% endmacro %}
+
+{# Current-state exclusion for regions whose SIS is frozen. `column` is any
+   code-location column -- `_dbt_source_project` on a union model, or
+   `dagster_code_location` / `home_work_location_dagster_code_location` on the
+   staff roster. Add or remove a region in the frozen_code_locations var. #}
+{% macro exclude_frozen(column) -%}
+    {{ column }} not in ('{{ var("frozen_code_locations") | join("', '") }}')
+{%- endmacro %}
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__enrollments.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__enrollments.sql
@@ -14,8 +14,7 @@ inner join
     and s.enroll_status in (0, -1)
 where
     cc.dateleft >= current_date('{{ var("local_timezone") }}')
-    -- Miami rosters into Clever directly from Focus; excluded from all six feeds
-    and cc._dbt_source_relation not like '%kippmiami%'
+    and {{ exclude_frozen("cc._dbt_source_project") }}
 
 union all
 
@@ -30,4 +29,4 @@ select
 
     student_number as student_id,
 from {{ ref("stg_powerschool__students") }}
-where enroll_status in (0, -1) and _dbt_source_relation not like '%kippmiami%'
+where enroll_status in (0, -1) and {{ exclude_frozen("_dbt_source_project") }}
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__schools.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__schools.sql
@@ -22,9 +22,7 @@ select
 from {{ ref("stg_powerschool__schools") }}
 where
     /* filter out summer school and graduated students */
-    state_excludefromreporting = 0
-    -- Miami rosters into Clever directly from Focus; excluded from all six feeds
-    and _dbt_source_relation not like '%kippmiami%'
+    state_excludefromreporting = 0 and {{ exclude_frozen("_dbt_source_project") }}
 
 union all
 
--- a/src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql
+++ b/src/dbt/kipptaf/models/extracts/clever/rpt_clever__sections.sql
@@ -12,9 +12,7 @@ with
         from {{ ref("int_people__staff_roster") }} as sr
         where
             sr.assignment_status != 'Terminated'
-            -- Miami rosters into Clever directly from Focus; excluded from all
-            -- six feeds
-            and sr.home_work_location_dagster_code_location != 'kippmiami'
+            and {{ exclude_frozen("sr.home_work_location_dagster_code_location") }}
             and sr.job_title in (
                 'Director of Campus Operations',
                 'Director Campus Operations',
@@ -38,7 +36,7 @@ with
         from {{ ref("stg_powerschool__schools") }}
         where
             state_excludefromreporting = 0
-            and _dbt_source_relation not like '%kippmiami%'
+            and {{ exclude_frozen("_dbt_source_project") }}
     ),
 
     teachers_long as (
@@ -109,9 +107,7 @@ with
             and st._dbt_source_project = t._dbt_source_project
         where
             sec.terms_yearid = ({{ var("current_academic_year") - 1990 }})
-            -- Miami rosters into Clever directly from Focus; excluded from all
-            -- six feeds
-            and sec._dbt_source_relation not like '%kippmiami%'
+            and {{ exclude_frozen("sec._dbt_source_project") }}
 
         union all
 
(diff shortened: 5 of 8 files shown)


cbini:
Verdicts on the two Minor findings. I verified both claims against the code before answering, including the file citations.

**Minor 1 — empty-list guard in `exclude_frozen`. Declined.**

The rendering claim is correct. I confirmed it: `[] | join("', '")` produces an empty string, so `frozen_code_locations: []` would render `column not in ('')`.

I am not adding the guard, because that render is not a coincidence — it is the right answer. An empty frozen list means "exclude no region". `col not in ('')` excludes only rows whose code location is the empty string, and no code location is ever the empty string, so the predicate excludes nothing. That is exactly what an empty list should do.

The null case is the same story. `regexp_extract` returns NULL, never `''`, on no match, and a NULL column fails `NOT IN`, `!=` and `NOT LIKE` identically — so an empty list drops NULL-code-location rows exactly as the pre-refactor literals did. It changes nothing, and there are no such rows: 0 NULL `_dbt_source_project` across 1,067,390 prod rows on all 5 upstreams.

So the guard would add a Jinja branch to a 3-line macro to make already-correct behavior look correct, and it would cost a full `state:modified+` dbt Cloud re-run, which this PR triggers across most of the project. Not worth it at that price.

If you want it anyway it is one line, and I will bundle it with any other change rather than push it alone.

**Minor 2 — the one-line `WHERE` in `rpt_clever__schools.sql:25`. Agreed, no action.**

Your read of the mechanism is right, and the citation is real — `src/dbt/CLAUDE.md:1278` is the "Removing comments changes lint/format behavior" section. Deleting the 2-line comment dropped the line under 88 characters, so sqlfmt rejoined it.

One correction to your description: I did not leave it as sqlfmt produced it by choice. `trunk check --force` flagged it as an autofix, and I applied that fix by hand before committing, then re-ran the check to confirm clean. The result is the same, but nothing here was left unformatted.

**On the data-dependent claims you flagged as unverified.** That is the right call given no warehouse access, and the equivalence proof is the load-bearing part of this PR, so here is how to re-run it directly rather than through a dbt build. This is the check that matters — it compares the old and new predicates row-for-row on prod:

```sql
select
  countif(_dbt_source_relation not like '%kippmiami%') as old_kept,
  countif(_dbt_source_project not in ('kippmiami')) as new_kept,
  countif(_dbt_source_project is null) as null_project,
  count(*) as total
from `teamster-332318.kipptaf_powerschool.stg_powerschool__cc`
```

Repeat for `stg_powerschool__students`, `stg_powerschool__schools`,
`base_powerschool__sections` (in `kipptaf_powerschool`) and
`int_extracts__student_enrollments` (in `kipptaf_extracts`). Every row of the
PR body's table came from that query.

PR #5008, opened 2026-08-26
fix(dbt): keep the primary stint when a Focus student has two schools

### LINKED ISSUE #5003: feat(dbt): explain the 186 Miami course enrollments whose stint is at another school
## What and why

The smaller half of #4970. Read that issue for the measured analysis.

186 Miami course enrollments attach to no stint because the student holds their
AY2026 stint at a **different school** than the section's. Two candidate causes,
neither established:

1. The schedule row is stale after a transfer — the student moved schools and the
   old section assignment was never ended.
1. The Focus school-to-network mapping resolves the wrong school for these
   sections.

These are distinguishable. Cause 1 shows up as a real transfer in the enrollment
history; cause 2 shows up as a consistent offset for particular sections or
schools.

## One hypothesis already ruled out

Focus documents a null `student_enrollment.custom_9` as marking the student's
primary or home-campus enrollment record, which suggested these might be
non-primary enrollment rows. **They are not.** Only 1 of 1,758 AY2026 Miami
enrollment rows has `custom_9` set, measured 2026-08-26. `custom_9` is also not
mapped in `stg_focus__student_enrollment` today — worth fixing on #4600 for
completeness, but not the cause here.

## Independent of the other half

This does not need #4987 and does not need the schedule-date question answered. It
can run at any time.

For Claude

Context, not instructions.

Relevant models:

- `src/dbt/kipptaf/models/marts/dimensions/dim_student_section_enrollments.sql` —
  the `enrollment_overlap` CTE
- `src/dbt/kipptaf/models/students/intermediate/int_students__course_enrollments.sql`
  — `focus_conformed` maps the school through
  `stg_google_sheets__people__locations.focus_school_id` to
  `powerschool_school_id`. If the mapping is the cause, it is here.

Constraints:

- 186 rows out of 19,295 is small enough that a per-row inspection is feasible.
  Check the actual enrollment histories before designing a rule.
- NJ counts must not move.

Acceptance criteria:

- The 186 are attributed to stale schedule rows, a mapping defect, or a mix, with
  counts for each.
- Whatever the cause, the handling is recorded in the model description rather than
  left as a silent null.

Refs #4970
Refs #4600

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop attributing a student to a school they
do not attend.

Focus lets a student hold two enrollments at once — a home campus plus a second
school for a few courses. It marks the second one with a checkbox called "Second
School." `int_focus__student_enrollment` collapses two enrollments that share a
start date down to one, but it never read that checkbox, so it kept whichever
was created later.

For the one student this affects, that is the second school. The model records
them at KIPP Miami Technical High when they attend KIPP Courage Academy, and the
4 courses they take at Courage lose their link to any enrollment.

The fix reads the checkbox and keeps the home-campus enrollment.

This started as #5003, which asked why 186 Miami course enrollments attach to no
enrollment stint. Answering that turned up this bug.

Refs #5003

## Reviewer Notes

**One student changes, and nothing else.** Verified against prod raw data:
10,075 surviving enrollments before and after the change, so the grain does not
move and no uniqueness test is affected. Exactly 1 row differs.

**The 4 courses stay unlinked, and that is now correct.** They are taken at the
second campus, so no enrollment there covers them. Before the fix the *other* 4
were unlinked — the ones at the student's real school. Same count, opposite and
correct half.

**The signal is a single row, but its meaning is not guessed.** Only one
enrollment in nine years of Focus data has the checkbox set. The field's meaning
comes from Focus's own custom-field catalog, which titles it "Second School" —
not from inference. Two independent facts agree: the home-campus record carries
an enrollment code and the daily attendance, the second-school record carries
neither.

**A new column on a contracted staging model.** `second_school` is additive, and
no kipptaf model names it, so nothing downstream needs to change with it.

**New Jersey cannot be affected.** The `focus` package is imported only by
`kippmiami` (`grep -l 'local: ../focus' src/dbt/*/packages.yml`).

## Self-review

### General

      (Claude is advisory — use your judgement, but don't ignore it.) — all 4
      actionable findings accepted; per-finding verdicts posted as a comment.

Skipped — no Dagster changes.

      models. The new column is declared in both properties files, and the new
      test has its entry in `src/dbt/focus/tests/properties.yml`.
- [x] Include (or update) an
      applicable, no new or changed models.
      (`stg_`, `rpt_`, mart) will break downstream exposures — no. One column is
      added to a contracted staging model; none renamed or removed.
      with **`--target staging`** — not applicable. Focus is dlt-loaded into
      BigQuery, not an external table.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The full #5003 attribution, measured 2026-08-26
against prod, is a comment on that issue; this covers the diff.

**How the dedupe was picking wrong.** `int_focus__student_enrollment`
deduplicates on `(student_number, academic_year, startdate)` with tie-break
`(schoolid is null) asc, enrollment_created_at desc, student_enrollment_id desc`.
For the affected pair both schools are non-null, so the first term is inert and
`enrollment_created_at desc` decides — the second-school record was created
later (2026-08-12 vs 2026-07-13), so it won. `is_second_school asc` now precedes
both.

**Why the field lookup nearly missed.** `stg_focus__custom_fields` stores
`column_name` UPPERCASE for `StudentEnrollment` (`CUSTOM_9`) and lowercase for
`SISDisciplineIncident` (`custom_9`), so a lowercase filter returns only the
discipline row ("Alcohol Related") and reads as though enrollments have no
`custom_9` definition. `src/dbt/focus/CLAUDE.md` already documents the
`lower(column_name)` join for entity-table joins; this is the same trap on the
catalog side.

**Dead code kept deliberately.** Zero Focus enrollment rows carry a null
`school_id` in any year, so `(schoolid is null) asc` never fires. Left in place
as a guard, since a null-school stub is what motivated the dedupe originally
(#4904).

**What I deliberately did not do.** Keeping BOTH enrollments would attach all 8
of that student's course enrollments rather than 4. It needs the `rn_year`
`row_number()` tie broken (both enrollments default `exitdate` to June 30) and
the uniqueness keys on `int_students__student_enrollment_union` (warn) and
`int_extracts__student_enrollments` (**error**, and Miami flows 1:1 into that
extract at 1,760 rows) widened to include `schoolid`. Measured: that widening
looks correct rather than loosening — 6 of the union's existing violations are
all New Jersey, and one Newark group is a legitimate different-school
same-entrydate pair the 4-column key wrongly flags. But it changes the
enrollment grain network-wide for 4 course-enrollment rows, so it is a separate
call.

**Stale trackers cleaned up.** 3 in-code references pointed at #4905, closed as
a duplicate of #4904, itself closed as completed by merged PR #4910 — the PR
that created this dedupe. They now name the condition they track instead of a
dead issue, so they cannot rot the same way.

**The other 182 of #5003's 186.** Stale schedule rows at a campus the student
was reassigned away from, left open in Focus. Upstream Ops cleanup, not a model
problem; a null key is correct, since attaching a section at one campus to a
stint at another emits a confidently wrong key (#4970). The Focus
school-to-network mapping was ruled out as a cause three ways: the locations
crosswalk is 1:1 on `focus_school_id` with every Focus school title matching its
crosswalk `location_name`; `schedule.school_id` equals
`course_periods.school_id` on all 19,649 AY2026 rows; and the cross-school pairs
run in both directions within a grade band, which no per-school offset produces.

**Verification.** `dbt parse --target prod` clean on `kippmiami` and `kipptaf`
after every commit. The new test compiles against
`teamster-332318.kippmiami_focus.stg_focus__student_enrollment` and returns 0
rows in prod, so it ships green at `severity: error`. Simulated the new
tie-break over prod raw data: 10,075 stints both sides, 1 row changed, school 14
to 15. Simulated the mart join with the corrected stint: the 4 Courage sections
attach, the 4 Miami Tech sections do not. `trunk check --force` clean.

**PII.** Student identifiers stayed in a local scratch file. Nothing in this PR,
the issue comment, or the committed comments carries a student number or name.

</details>


cbini:
## Review verdicts

Per-finding, against the code and the warehouse. Note the review ran on the
first commit; two later commits changed the approach substantially, so a couple
of findings resolved on the way.

| # | Finding | Verdict |
| --- | --- | --- |
| 1 | NULL `school_id` invisible to `count(distinct school_id)` | **Accepted, documented** |
| 2 | `severity: warn` violates the staging-tests-must-error rule | **Accepted, fixed** |
| 3 | Duplicated narrative between SQL comment and YAML description | **Accepted, fixed** |
| 4 | Issue refs in a mart column `description:` | **Accepted, fixed** |
| 5, 6 | No violations found / claims spot-checked | No action |

### 1. NULL `school_id` gap — accepted, and it is a real gap

The SQL reasoning is right: `COUNT(DISTINCT ...)` ignores nulls, so a
null-school + real-school pair reads as `n_schools = 1` and is not flagged. Ran
your query against prod:

```text
0 rows
```

Zero enrollment rows in Focus carry a null `school_id`, measured back to AY2018
— so the gap is theoretical rather than live, and the dedupe's
`(schoolid is null) asc` tiebreak is currently dead code. Your point that the
limit was unstated stands, so the test and its description now say it explicitly:
with no school there is no second school to be primary over.

### 2. `severity: warn` on a staging test — accepted, already resolved

You were right that this violates `src/dbt/CLAUDE.md` → _Test config defaults_
("Staging-layer tests MUST set `config: severity: error`"), and I verified the
rule text rather than taking the citation on trust. It is not the well-precedented
exception you generously read it as — the `TODO(#4905)` I anchored it to points
at a **closed** issue, so nothing tracked restoring `error`.

Resolved by reworking the assertion instead of the severity. See the note below.

### 3 and 4. Description vs comment — both accepted

`src/dbt/CLAUDE.md` → _YAML conventions_ is explicit: descriptions carry what and
why a column computes, and "TODOs, history, migration plumbing, or tracking-issue
refs (`#3142`, etc.)" go in inline SQL comments at the derivation site. Same
section caps a SQL comment at "what a reader of that line cannot see."

You declined to treat 4 as a violation on precedent grounds. I am overriding that:
the same section says the repo's existing multi-paragraph SQL comments "are not a
precedent to extend," which reads as the written rule beating the drift. So the
counts and refs moved to the comment, the description kept the causes and why each
null is correct, and the comment shrank to fit its own rule. That collapses 3 as a
side effect.

## What changed underneath the review

Worth flagging, since it invalidates the premise of your finding 6 rather than
contradicting it. Your spot-check of my `rn_year` tie claim was correct on the
code as written. The claim itself turned out to rest on a missing fact.

`StudentEnrollment.CUSTOM_9` is titled **"Second School"** in Focus's own
custom-field catalog, a checkbox. The one AY2026 row carrying it is the stint the
dedupe was **keeping** — so the model attributed that student to a campus they do
not attend, and stranded the course enrollments at their real one. The fix maps
that field and makes it the dedupe's first tiebreak.

Verified against prod: 10,075 surviving stints before and after, so no grain
change and no uniqueness test affected; exactly 1 row changes. That is why the
test is now an assertion that passes at `severity: error` with no TODO, rather
than a warning against a closed issue.

cbini:
## dbt Cloud is red here for a reason outside this PR

This PR's dbt Cloud check fails on a `main` regression, not on anything in these
9 files:

```text
Unrecognized name: student_first_name; failed to parse view
'teamster-332318.kipptaf_focus.int_focus__student_enrollments'
```

`int_assessments__course_enrollments` reads through that view, and the view has
been unqueryable in prod since #5009 reshaped its upstream. Filed as #5013 and
fixed in #5014.

**Merge #5014 first**, then this PR's CI will pass on a re-run. Nothing to change
here.

For the record, the earlier failure on this branch was a different and genuinely
transient cause: the four `zz_stg_kipptaf_pearson` staging tables were dropped
and recreated at 20:06:32 by a concurrent staging clone, and the build failed at
20:07:20 having resolved a reference that had just been replaced. Two failures,
two unrelated causes.

Issue #5001, opened 2026-08-26
refactor(dbt): repoint the kipptaf Focus enrollment wrappers and delete the 3 superseded ones


cbini:
## What and why

PR 2 of #4600. Read that issue for the column-containment analysis.

## Blocked on

The focus package PR must be **merged and materialized in prod** first. Not merged
— materialized. See the sequencing comment on #4600 for the #4290 mechanism.

## Proposed approach

1. **Add a source entry and wrapper** for the reshaped
   `int_focus__student_enrollment`.
1. **Repoint kipptaf `int_focus__student_enrollments`** at the new roster model. It
   keeps its name — it is a wrapper plus enrichment (locations crosswalk, region,
   district), not a pure passthrough, so the same-name convention does not bind it.
1. **Delete 3 superseded staging wrappers**, verified by column containment rather
   than by name:

   | kipptaf model | superseded by | staging columns not covered |
   | --- | --- | --- |
   | `stg_focus__schools` | `int_focus__schools` | 0 of 24 |
   | `stg_focus__students` | `int_focus__students` | 0 of 34 |
   | `stg_focus__student_enrollment` | `int_focus__student_enrollment` | 0 of 36 |

   All 3 have zero consumers today — the only
   `source("kippmiami_focus", "stg_focus__*")` calls are the wrappers themselves.
   The two attendance-calendar wrappers stay; no int equivalent exists for them.

For Claude

Context, not instructions.

### Consumers to verify

`int_focus__student_enrollments` feeds `int_students__student_enrollments`,
`int_students__students`, `int_students__student_enrollment_union`,
`int_students__teacher_grade_levels`, `int_tableau__fresh_enrollment_scaffold`,
`int_tableau__finalsite_student_scaffold`, `rpt_gsheets__kippfwd_miami_roster`, and
`rpt_google_directory__users_import`. Through
`int_students__student_enrollment_union` it reaches `dim_student_enrollments`.

#4600 says only the two `int_tableau__` scaffolds are affected. That list is
incomplete — verify all 8 against prod.

### Constraints

- `dbt clone` gives the old schema for the package model modified in PR 1. Refresh
  `zz_stg_kippmiami_focus` with `stage_external_sources --target staging` then
  `dbt build --select ... --target staging`, per the single-PR cross-project
  workflow in `src/dbt/kipptaf/CLAUDE.md`. Also clone and build `zz_stg_kipptaf`.
[...]

Issue #4999, opened 2026-08-26
feat(dbt): branch the kipptaf final-grades and GPA chain onto Focus for Miami


cbini:
## What and why

PR 2 of #4986. Read that issue and its comments first.

`fct_grades_term` and `fct_grades_gpa` read empty for Miami AY2026 because
`base_powerschool__final_grades` and the GPA chain still union the frozen
`kippmiami_powerschool` archive as their Miami branch.

## Blocked on

The focus package PR must be **merged and materialized in prod** first. Not merged
— materialized. The kipptaf wrappers are views, and if one materializes before its
package upstream rebuilds it stays compiled against the old column set and
downstream reads fail `failed to parse view`. That is #4290 and it does not
self-heal.

## Proposed approach

1. Declare and wrap `int_focus__report_card_grades` in
   `kipptaf/models/focus/sources-kippmiami.yml`.
1. Branch `base_powerschool__final_grades`.
1. Branch the GPA chain: `int_powerschool__gpa_term`,
   `int_powerschool__gpa_cumulative`, `int_powerschool__gpa_cumulative_year`.
1. Branch the grade-scale resolution: `int_powerschool__final_grades_pivot`,
   `int_powerschool__section_grade_config`,
   `int_powerschool__gradescaleitem_lookup`.

## An open scoping question, settle before designing

Focus computes GPA and class rank itself, in `student_gpa_calculated`, with
`marking_period_id = -1` marking the course-history GPA. That is vendor-documented.

**That table is not ingested** — absent from `dagster_kippmiami_dlt_focus` and
absent from the focus dlt config, verified 2026-08-26.

Sourcing the vendor's own GPA is likely better than recomputing it through a
branched PowerSchool chain, but it needs an ingestion change first. Decide which
way this goes before writing the GPA branches, because the two designs share almost
nothing.

For Claude

Context, not instructions.

### CI

This reads a package model modified in the previous PR, so `dbt clone` gives the
old schema. Use `stage_external_sources --target staging` then
`dbt build --select ... --target staging` to refresh `zz_stg_kippmiami_focus`, per
the single-PR cross-project workflow in `src/dbt/kipptaf/CLAUDE.md`. Also clone and
build `zz_stg_kipptaf` itself.

### Constraints

- Year-scope the union rather than excluding Miami wholesale. The frozen archive
  holds Miami AY2020 through AY2025 for these models and that history has to stay
  readable.
[...]

Issue #4986, opened 2026-08-26
feat(dbt): wire Focus report card grades into the kipptaf network layer for Miami AY2026


cbini:
## What and why

`fct_grades_term` and `fct_grades_gpa` read empty for Miami AY2026. #4925 planned
3 PRs to fix that. PR #4972 shipped PR 1, the course enrollments half, and #4925
closed as completed on 2026-08-26. PRs 2 and 3, the report card grades half,
never shipped and no open issue tracked them. This issue does.

Nothing here is new analysis. #4925 and its re-measure comment of 2026-08-25 hold
the full design. This issue carries the work forward so it stays visible.

Verified against `main` on 2026-08-26:

- `stg_focus__student_report_card_grades` still omits all 4 columns the raw dlt
  table carries, and its marking-period `TODO` is intact.
- `base_powerschool__final_grades` still unions `kippmiami_powerschool`, the
  frozen archive, with no Focus branch.

## Proposed approach

Two PRs, in this order, for the reason #4925 gives: the kipptaf wrappers are
views, so the translator assigns `dbt_union_relations_automation_condition()`,
and a cross-code-location deploy race (#4290) can leave a wrapper compiled
against the old column set. That failure does not self-heal.

### PR 1, the focus package

1. Widen `stg_focus__student_report_card_grades` with `report_card_grade_id`,
   `grade_scale_id`, `course_period_id`, and `percent_grade`.
1. Decode the marking-period token generically: strip the alpha prefix, cast the
   digit tail. Do not enumerate the token values.
1. Keep the alpha token as its own column and include it in the grain.
1. Wire the grade-scale joins into `int_focus__report_card_grades` through
   `report_card_grade_id`, not the raw `grade_scale_id`.

Let Dagster materialize this in prod before opening PR 2.

### PR 2, kipptaf

Declare and wrap `int_focus__report_card_grades`, then branch
`base_powerschool__final_grades` and the GPA chain. Needs
`dbt clone --select ... --target staging` to refresh `zz_stg_kippmiami_focus`, or
CI reads a stale copy.

### Blocked externally

What `DT` versus `DY` means on the marking-period token is not established from
the data. Both tokens appear at all 4 schools with near-identical counts, and both
are needed to keep the grain unique, so the question cannot be settled from the
warehouse. It needs a Focus support answer before the token column can be named.

PR 1 can be built while that answer is outstanding. Name the column provisionally
and rename it when the answer arrives.

For Claude

[...]

Issue #4600, opened 2026-07-28
Reshape int_focus__student_enrollment to the staging-plus-labels pattern


cbini:
Follow-up to #4596, which established a consistent shape for the `focus` package: one `int_focus__{entity}` per staging model, carrying that model's columns plus its decoded custom-field labels, with the `__pivot` models kept internal.

`int_focus__student_enrollment` is the one model that does not follow it. It is a denormalized enrollment spine — student attributes, school attributes, decoded grade level, resolved entry and exit codes, point-in-time flags, and year counters — that happens to occupy the name the pattern wants.

The gap is total, not partial. Its column set shares **zero of the 36 columns** in `stg_focus__student_enrollment`: every key is renamed (`id` becomes `student_enrollment_id`, `school_id` becomes `schoolid`, `start_date` becomes `startdate`, `syear` becomes `academic_year`) and the raw operational fields are dropped entirely — `calendar_id`, `grade_id`, `next_grade`, `next_school`, `fl_days_present`, `mid_year_promotion`, `include_in_class_rank`, and 29 more. So there is currently no model exposing raw Focus enrollment attributes with labels.

## Changes

1. **Reshape `int_focus__student_enrollment`** to `stg_focus__student_enrollment` columns plus its five decoded labels, matching its six siblings.
2. **Add `int_focus__student_enrollment_roster`** carrying the current denormalized logic. Compose it from `int_focus__student_enrollment`, `int_focus__students`, and `int_focus__schools` rather than reaching back to raw staging — the combined models exist now, so the roster should not re-derive what they already provide.
3. **kipptaf `int_focus__student_enrollments` keeps its name** and repoints its source to the roster. It is a wrapper plus enrichment (locations crosswalk, region, district), not a pure passthrough, so the same-name convention does not bind it. Its two consumers — `int_tableau__fresh_enrollment_scaffold` and `int_tableau__finalsite_student_scaffold` — are untouched.
4. **kipptaf gains a wrapper and source entry** for the reshaped `int_focus__student_enrollment`.

## Deletions this unlocks

Three kipptaf staging wrappers become strict subsets of an int and can go, along with their properties and source entries. Verified by column containment, not by name:

| kipptaf model | superseded by | staging columns not covered |
| --- | --- | --- |
| `stg_focus__schools` | `int_focus__schools` | 0 of 24 |
| `stg_focus__students` | `int_focus__students` | 0 of 34 |
[...]

PR #5009, opened 2026-08-26
feat(dbt): widen the Focus report card models and reshape int_focus__student_enrollment

### LINKED ISSUE #4997: feat(dbt): widen the Focus report card grade models in the focus package
## What and why

PR 1 of #4986. Read that issue and its comments for the measured analysis and the
vendor documentation.

`stg_focus__student_report_card_grades` drops 4 columns the raw dlt table already
carries, and `int_focus__report_card_grades` never resolves the grade scale. Fix
both inside the `focus` package, before anything at kipptaf reads them.

## Proposed approach

1. **Widen the staging model** with `report_card_grade_id`, `grade_scale_id`,
   `course_period_id`, and `percent_grade`.
1. **Decode the marking-period token generically** — strip the alpha prefix, cast
   the digit tail. Keep the token as its own column and include it in the grain.
1. **Resolve the grade scale through `report_card_grade_id`**, not the raw
   `grade_scale_id`.
1. **Filter on the token's meaning.** Keep `DT` and `E`, drop `DY`.

## Why this is its own PR

This changes a focus package model's **column set**, and kipptaf reads it through
a `union_relations` view. Merging the kipptaf half in the same PR exposes the
#4290 deploy race, which does not self-heal — recovery is a manual `launch_run` of
the wrapper plus every consumer. Let Dagster materialize this in prod before
opening the kipptaf PR.

For Claude

Context, not instructions.

### The token is vendor-documented

Focus SIS Level 1 Certification, Day 2: `marking_period_id` is VARCHAR, and
`DT` = running gradebook, `DY` = yesterday's grade, `E` = exam.

Verified against prod 2026-08-26:

- `DT` is written by individual staff accounts (`created_by_class = SISUser`), and
  is the live grade.
- `DY` is written entirely by one account, username `cron`, profile `admin`
  (`created_by_class = FocusUser`), and mirrors the DT grade on 1,508 of 1,510
  pairs. It is a daily snapshot.
- No `E` row exists yet. All-time distribution is DT 1,536 and DY 1,516, both
  AY2026 only, plus 15,278 unprefixed rows spanning AY2012 to AY2025.

So `E` is a distinct measure, not a duplicate. Do not dedupe by picking one row per
`(student_id, course_period_id, marking_period_id)` — that silently drops exam
grades the moment they post. Do not filter to `DT` alone either.

The token vocabulary grew from 5 values across 2 schools to 8 across 4 schools in
5 days, and a `Full Year` entry disappeared. Write a generic decode, not an
enumeration.

### The grade-scale path

Raw `grade_scale_id` is populated on 100% of DT rows and 0% of DY rows, which is
why its apparent coverage reads about 50% and appeared to be declining. It is not
decaying — it tracks the DT share of rows.

`report_card_grade_id` joins `stg_focus__report_card_grades.id` with 0 orphans, and
`scale_id` is non-null on every matched row. `course_period_id` joins
`stg_focus__course_periods` with 0 orphans and no fan-out. Both verified 2026-08-26.

### Constraints

- Miami is the only region on Focus, so there is no NJ parity risk. Verify the
  AY2025-and-earlier rows anyway — they are unprefixed and must keep decoding.
- The `TODO` in `stg_focus__student_report_card_grades` points at the old unknown
  token vocabulary. Remove it with the fix.
- Land #4987 first if you can. It corrects the marking periods that
  `int_focus__report_card_grades` reads through `int_focus__schedule`.

### Acceptance criteria

- A uniqueness test on
  `(student_id, course_period_id, marking_period_id, grade_type_token)`. That test
  is what catches the prefix collapse.
- Grade-scale coverage reaches 100% on AY2026 rows.
- Marking period resolves on every AY2026 row, 0 orphans.

Refs #4986
Refs #4290

### LINKED ISSUE #5000: refactor(dbt): reshape int_focus__student_enrollment and add the roster model
## What and why

PR 1 of #4600. Read that issue for the column-containment analysis.

`int_focus__student_enrollment` is the one model in the `focus` package that does
not follow the staging-plus-labels pattern its six siblings follow. It is a
denormalized enrollment spine occupying the name the pattern wants, and it shares
**zero of the 36 columns** in `stg_focus__student_enrollment`. So no model exposes
raw Focus enrollment attributes with labels.

## Proposed approach

1. **Reshape `int_focus__student_enrollment`** to `stg_focus__student_enrollment`
   columns plus its 5 decoded labels, matching its siblings.
1. **Add `int_focus__student_enrollment_roster`** carrying the current denormalized
   logic. Compose it from `int_focus__student_enrollment`, `int_focus__students`,
   and `int_focus__schools` rather than reaching back to raw staging.
1. **Map `custom_9`** while reshaping. Focus documents a null `custom_9` as marking
   the student's primary or home-campus enrollment record, so it is a meaningful
   filter. Only 1 of 1,758 AY2026 Miami rows has it set, so mapping it changes
   nothing today — it stops the column being silently dropped.

## Why this is its own PR

The reshape replaces the model's column set entirely, and kipptaf reads it through
a `union_relations` view. Merging the kipptaf changes in the same PR exposes the
#4290 deploy race — the wrapper can materialize before the upstream rebuilds, stay
compiled against the old column set, and fail `failed to parse view` on every read.
It does not self-heal. See the sequencing comment on #4600.

Let Dagster materialize this in prod before opening the kipptaf PR.

For Claude

Context, not instructions.

Its stated dependency, #4596, merged on 2026-07-28. Nothing blocks this but the
staging reseed.

Constraints:

- `zz_stg_kippmiami_focus` needs a `--target staging` reseed of the reshaped model
  and the new roster before CI can resolve them. That reseed needs explicit
  authorization — ask before running it.
- Compose the roster from the combined `int_focus__*` models. Do not re-derive what
  they already provide.
- Miami is the only region on Focus, so there is no NJ parity risk here.

Acceptance criteria:

- `int_focus__student_enrollment` carries all 36 `stg_focus__student_enrollment`
  columns plus its 5 labels, and matches the staging row count exactly.
- `int_focus__student_enrollment_roster` reproduces the current denormalized output
  — parity against prod on row count and on `school_level`, by full outer join on
  `student_enrollment_id`.
- A `unique` plus `not_null` test on each model's key.

Refs #4600
Refs #4290
Refs #4596

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make Focus report card grades and Focus
enrollments usable by the network layer, entirely inside the `focus` package.
It is P3 on #4985 and covers two issues.

**Report card grades (#4997).** `stg_focus__student_report_card_grades` now
projects the 4 columns the raw table already carried — `report_card_grade_id`,
`grade_scale_id`, `course_period_id` and `percent_grade` — and decodes the
marking-period token generically: strip the alpha prefix, cast the digit tail.
Focus documents that token as the grade type, so it is kept as its own column,
`grade_type_token`, and is part of the grain. Every row now resolves a marking
period, including the 15,278 imported rows from AY2012 to AY2025 that were
already unprefixed. The stale `TODO` is gone.

`int_focus__report_card_grades` resolves the grade scale through
`report_card_grade_id` instead of the raw `grade_scale_id` column, which takes
AY2026 coverage from 50.3% to 100%. It drops the `DY` grade type and keeps `DT`
and `E`.

**Enrollment (#5000).** `int_focus__student_enrollment` now follows the
staging-plus-labels pattern its 6 siblings follow: every
`stg_focus__student_enrollment` column plus the 5 pivot labels. It shared zero
columns with staging before, so nothing exposed raw Focus enrollment attributes
with labels.

The denormalized spine that used to occupy that name moves to
`int_focus__student_enrollment_roster`, composed from
`int_focus__student_enrollment`, `int_focus__students` and `int_focus__schools`
rather than reaching back to staging. `int_focus__attendance_daily` and 3 unit
tests repoint to it.

`stg_focus__student_enrollment` also maps `custom_9`. Focus titles that field
Second School and documents a null value as the student's primary or home-campus
record, so it is a real filter. 1 of 10,076 rows has it set — mapping it changes
nothing today, it just stops the column being dropped.

**No kipptaf changes here, deliberately.** Both halves change a column set that
kipptaf reads through a `union_relations` view, which exposes the #4290 deploy
race. Let Dagster materialize this in prod, then open P4 (#4999 plus #5001).

## Reviewer Notes

- **A Miami-side spot check is wanted before merge.** Dropping `DY` rests on
  vendor documentation plus audit columns, not on a real report card. Please open
  one AY2026 student's Focus report card and confirm the grade shown matches the
  `DT` row. Detail in the fold-out.
- **The uniqueness test is scoped, not unscoped as the issue asked.** The grain
  the acceptance criteria name is not unique across imported course history.
  Numbers in the fold-out.
- **The issues say 36 staging columns. It is 38, now 39.** No action needed,
  just so the count in review does not read as a miss.
- **2 pivot labels stay unexposed.** The reshape carries the 5 labels the model
  carried before, per the issue. `resident_status_state_county_label` and
  `disaster_affected_student_label` exist in the pivot and reach no consumer. Say
  the word and I will add them.
- **An unrelated prod model is stale.** `int_focus__school_year_first_day` reads
  2026-08-12 in prod and 2026-08-10 from current code and current data. Found
  while checking roster parity; not touched here.
- **5 orphan grade-scale references** exist on `stg_focus__report_card_grades`,
  a model this PR does not change. They do not touch the resolution path used
  here. Detail in the fold-out.

### General

### dbt

- [ ] Include (or update) an
      applicable, both models are internal-only
- [x] **Breaking change?** Yes — `int_focus__student_enrollment` replaces its
      whole column set. Its only consumers are the kipptaf wrapper (handled in
      P4, gated on this materializing in prod) and
      `int_focus__attendance_daily`, repointed here. Rollback is a revert; no
      external consumer reads either model directly.
- [x] External sources: none added or modified. The `focus` sources are
      BigQuery-native dlt tables on a hardcoded schema, so they read prod under
      every target and need no staging. `zz_stg_kippmiami_focus` was seeded
      directly with `dbt build --target staging` — see the fold-out.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes. Note this will be a trivial no-op: dbt Cloud CI
      builds the kipptaf project only, and this PR touches no kipptaf model. The
      local build is the real validation.
- [ ] **Dagster Cloud** — not triggered, no Dagster paths changed.

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Every number below was measured against prod on
2026-08-26, not carried over from the issue bodies.

### The DY spot check, and why it is still wanted

The case for dropping `DY` is strong but indirect. Vendor documentation (Focus
SIS Level 1 Certification, Day 2) says `DT` is the running gradebook, `DY` is
yesterday's grade, `E` is an exam. The audit columns agree with no crossover:
all 1,516 `DY` rows are written by one account, username `cron`, profile
`admin`, and the `DT` rows by 10 individual staff accounts. Self-joining the two
on `(student_id, course_period_id, tail_id)` gives 1,510 pairs agreeing on
`grade_title`, `percent_grade` and `gpa_points` on 1,508 of them.

What is missing is a human confirming that the grade Focus SHOWS on a report card
is the `DT` value. The 2 disagreeing pairs are the interesting case — a teacher
changed the grade after the snapshot. If Focus renders `DY` on the printed report
card, this filter is backwards and `fct_grades_term` will read a day stale in P4.
Nothing in the warehouse can settle it: `modified_date` is null on all 3,074
live rows, and row `id` ordering is mixed (`DY` holds the higher id on only 206
of 1,510 pairs).

The filter is written as `where coalesce(grade_type_token, '') != 'DY'`, so
reversing it is a one-line change.

### Why the uniqueness test is scoped

The acceptance criteria ask for a uniqueness test on
`(student_id, course_period_id, marking_period_id, grade_type_token)`. Unscoped,
that fails. Measured on the rows this PR keeps (18,352 raw minus 1,516 `DY`):

| Row set | Rows | Distinct on the 4-column grain |
| --- | --- | --- |
| All kept rows | 16,836 | 3,718 |
| Imported course history only | 15,278 | 2,160 |
| Live-posted rows only (token not null) | 1,558 | 1,558 |

Imported course history carries no token and a null `course_period_id`, and
legitimately repeats a `(student, marking period)` pair across transcript
courses — adding `syear` and `course_num` still leaves 12,780 distinct on 15,278.
So the test carries `where: grade_type_token is not null` on both the staging and
the intermediate model. It still catches the prefix collapse the criteria are
aimed at: strip the token and the same 3,074 tokened rows collapse to 1,542.

### Verification

Local build through the `kippmiami` project, selection widened to rebuild the
upstream staging models (a first pass showed 61 `student_id` orphans that were
purely a stale dev copy of `stg_focus__students`, and 98 `course_period_id`
orphans that were a stale `zz_stg` copy of `stg_focus__course_periods` — both
gone once rebuilt):

- `PASS=92 WARN=3 ERROR=0` on the dev target. All 7 unit tests pass, including
  the 3 repointed ones.
- `PASS=111 WARN=2 ERROR=0` on the staging target.
- The remaining warnings are pre-existing and unrelated: 2 duplicate-open-stint
  warnings on `stg_focus__student_enrollment` (#4905) and, in the first pass
  only, 4 `course_period_id` orphans on `stg_focus__co_teachers`. Note
  `kippmiami/dbt_project.yml` sets `data_tests: +severity: warn` project-wide,
  which overrides the per-test `severity: error` in the package properties files
  — so nothing in this package fails a district build regardless.

Three foreign keys on this model verified 0 orphans against the raw table over
all 18,352 rows, and each got a `relationships` test: `report_card_grade_id` to
`report_card_grades.id`, `course_period_id` to
`course_periods.course_period_id`, and raw `grade_scale_id` to
`report_card_grade_scales.id`.

The resolved scale path — `report_card_grades.scale_id` to
`report_card_grade_scales.id` — is measured but deliberately NOT tested, which
corrects an earlier version of this sentence. It has 0 orphans among the grade
definitions this model actually reaches, but 5 orphans across the full
1,357-row `report_card_grades` table. Those 5 definitions are referenced by no
student grade row, so the resolution path this PR relies on is clean while the
parent FK as a whole is not. A `relationships` test there would be a
knowingly-failing test on a model this PR does not otherwise change, so the
measurement is recorded here instead.

Acceptance criteria, measured on the built model:

- Grade-scale coverage: 0 null `grade_scale_id` and 0 null `grade_scale_title`
  on the 1,558 AY2026 rows.
- Marking period: 0 nulls across all 16,836 rows, so the historical rows still
  decode.
- `percent_grade` and `course_period_id`: 0 nulls on AY2026.
- `DY`: 0 rows. Only `DT` is present; no `E` row exists yet.
- `int_focus__student_enrollment`: 44 columns (39 staging plus 5 labels), 10,076
  rows, matching staging exactly.
- Roster: 10,075 rows and 40 columns, both matching prod.

### Roster parity, and the first-day-of-school drift

Full outer join on `student_enrollment_id`, roster against prod
`int_focus__student_enrollment`: 0 rows on either side only, and 0 diffs on every
column except `first_day_of_school` (1,760) and the 2 flags derived from it,
`is_enrolled_fdos` (1,680) and `is_pre_year_withdrawal` (18).

That is not this change. Substituting prod's `int_focus__school_year_first_day`
value into the roster's flag expressions drops all 3 to 0 diffs, which isolates
the delta entirely to that one upstream model. Prod holds 2026-08-12 for
`syear = 2026`; a fresh build of the unmodified model against current data gives
2026-08-10. The model joins `attendance_calendar` to `attendance_calendars` on
`calendar_id` alone and filters `minutes > 0`, so it picks up a school whose
calendar starts earlier than the default-calendar minimum. Worth its own issue;
out of scope here.

### The staging seed

Authorized on #5000, 2026-08-26. `stage_external_sources` was NOT run and is not
needed: `kippmiami/dbt_project.yml` pins `focus_schema` to
`dagster_kippmiami_dlt_focus`, a hardcoded BigQuery-native dataset, so the focus
sources read prod under every target. The seed was
`dbt build --target staging --indirect-selection=cautious --select +int_focus__student_enrollment_roster+ +int_focus__report_card_grades+ +int_focus__attendance_daily+ stg_focus__course_periods`.

`--indirect-selection=cautious` matters. Without it, eager indirect selection
pulls in `relationships` tests belonging to unrelated staging models
(`stg_focus__students_join_groups`, `stg_focus__gradebook_grades`,
`stg_focus__discipline_referrals`, `stg_focus__attendance_period`,
`stg_focus__scheduling_teams`, `stg_focus__students_join_students`) because their
parent is selected, then fails all 8 because those models are not built in
`zz_stg`.

`zz_stg_kippmiami_focus` now holds all 5 changed models at their new shapes, so
P4 has what it needs on the CI side. One caveat for whoever runs P1 or P2
concurrently: that schema is shared, and a kipptaf CI run that rebuilds
`int_focus__student_enrollments` under `--target staging` will now see the
reshaped column set and fail on the `except` list until P4 lands.

### Column count and label scope

The issues say 36 staging columns. The model projected 38 before this change and
39 after. Verified by counting the select list and by
`INFORMATION_SCHEMA.COLUMNS`.

On labels: `int_focus__student_enrollment__pivot` produces 7, and the 6 sibling
models all carry every label their pivot produces. The reshape carries 5, per the
issue and the direction given. `resident_status_state_county_label` and
`disaster_affected_student_label` therefore stay with no consumer.

### Small in-blast-radius fixes

- `tests/properties.yml` and `tests/test_focus__school_level_labels_known.sql`
  both said the `school_level` decode lives in `int_focus__student_enrollment`.
  It lives in `int_focus__schools`. Corrected.
- `src/dbt/focus/CLAUDE.md` said `int_focus__student_enrollment.student_number`
  holds the 8400-prefixed id. That column now lives on the roster. Repointed.
- The dedupe unit test is renamed to
  `test_int_focus__student_enrollment_roster_dedupes_same_start_date_stints` and
  its fixtures moved from `stg_focus__students` plus
  `stg_focus__student_enrollment` plus the pivot to `int_focus__students` plus
  `int_focus__student_enrollment`, matching the roster's new upstreams.

Closes #4997
Closes #5000
Refs #4986
Refs #4600
Refs #4985
Refs #4290

</details>


cbini:
## Review verdicts

Verified each finding against the code before acting. 3 fixed, 2 declined with
evidence, 1 answered by correcting the PR body instead. Pushed as `7218ab8b00`.

| # | Finding | Verdict |
| --- | --- | --- |
| 1 | Column order fails sqlfluff ST06 | **Declined** — does not fire |
| 2 | Column-test sort order in properties YAML | **Fixed** |
| 3 | Missing model-level `contains_pii` on the roster | **Declined** — matches precedent |
| 4 | Missing `contains_pii` on `second_school` | **Fixed** |
| 5 | PR-claimed relationships test not present | **PR body corrected**, test deliberately not added |
| 6 | Published doc goes stale on merge | **Fixed** |

### 1. Declined — ST06 does not fire

You flagged this as Confidence: Medium and asked for a local lint run. Ran it:

```text
trunk check --force --no-fix \
  src/dbt/focus/models/staging/stg_focus__student_report_card_grades.sql \
  src/dbt/focus/models/intermediate/int_focus__report_card_grades.sql \
  src/dbt/focus/models/intermediate/int_focus__student_enrollment_roster.sql

Checked 3 files
No issues
```

The `Trunk Check` check-run on this PR is also green on that exact file, which
is the full CI check rather than a diff-scoped one.

The convention you quoted is real but does not reach this case. It governs a
plain `cast(col as type)`, which ST06 buckets as a simple target and which
therefore has to precede any function call. Here the argument is itself a
function — `safe_cast(regexp_extract(...) as int)` — so ST06 buckets it as a
calculation, the same bucket as the bare `regexp_extract` above it. Two
calculations, so their relative order is unconstrained.

### 2. Fixed

`grade_type_token` now sits below the 6 columns that carry `data_tests`, ahead
of `percent_grade` and the remaining untested columns. Rebuilt to confirm the
contract still holds after the reorder — contract matching is by name, not
position, and the build is clean.

### 3. Declined — this matches precedent, and nothing was lost in the move

The premise that a tag was dropped during the move is not what the diff shows.
The pre-PR `int_focus__student_enrollment.yml` had **no** model-level
`contains_pii` either — it carried per-column tags only, exactly as the roster
now does. The roster reproduces that tagging unchanged.

The model-level tag on the reshaped `int_focus__student_enrollment` is not the
same decision. That model is `select s.*` plus labels, so there are no
enumerated columns to tag; the model-level tag is the only place it can go.
That is why `int_focus__students` and `int_focus__users`, both `s.*` models,
carry it too.

The split across the package is by SQL shape, not by PII density:

| Tagging | Models |
| --- | --- |
| Model-level | `int_focus__students`, `int_focus__users`, the 3 `__pivot` models, `int_focus__custom_field_log__unpivot`, and the reshaped `int_focus__student_enrollment` — every one a `select *` model |
| Column-only | `int_focus__student_contacts` (13 tags), `int_focus__attendance_day` (3), `int_focus__attendance_period`, `int_focus__attendance_daily`, `int_focus__attendance_streak`, `int_focus__ada`, `int_focus__gradebook_grades` — every one an explicit-column model |

`int_focus__student_contacts` is the closest analogue: explicit-column,
PII-dense, 13 per-column tags, no model-level tag. The roster is explicit-column
with 8 per-column tags. Per-column is also the more precise statement — it says
which 8 of 40 columns are PII rather than painting the whole relation.

Happy to add the model-level tag as belt-and-braces if you would rather have
both, but it is not a gap against the pattern.

### 4. Fixed

`second_school` now carries `contains_pii: true`. The sibling precedent you
cited is 7 of 7 on that model, and a non-home-campus flag is a geographic
indicator under the indirect-identifier list. Cheap and errs the right way.

Worth noting for calibration: the "100% of custom fields tagged" observation is
true, but the model's tagging is not "all indirect identifiers" — `school_id`,
`grade_id`, `start_date` and `end_date` sit untagged on the same model and are
stronger indirect identifiers than this flag. Tagged anyway.

### 5. PR body corrected; the test is deliberately not added

You are right that the test does not exist and that the PR body claimed it did.
That sentence was wrong and is now corrected.

Adding the test would have been the wrong fix, though. Measured the FK across
the whole parent table, not just the rows this model reaches:

| Scope | Rows | Orphans |
| --- | --- | --- |
| Grade definitions reached via `report_card_grade_id` | 1,357 matched | **0** |
| Full `report_card_grades` table | 1,357 | **5** |

`scale_id` is non-null on all 1,357 rows, and 5 of them point at a
`report_card_grade_scales.id` that does not exist. Those 5 definitions are
referenced by no student grade row, so the resolution path this PR depends on is
clean while the parent FK as a whole is not.

So a `relationships` test there would be a knowingly-failing test on a model
this PR does not otherwise change. The measurement is recorded in the PR body
instead, and the 5 orphans are now flagged in Reviewer Notes as a separate
data-quality observation for @cbini to decide on.

### 6. Fixed

`docs/models/fresh-dashboard-data-model.md` no longer names
`int_focus__student_enrollment` as the owner of the full derivation.

Fixed here rather than deferred to P4, for the reason you gave: the sentence
goes wrong the moment Dagster rematerializes this in prod, which is ahead of
P4. The replacement is worded to be accurate at every point in the sequence —
it names the roster as the derivation owner, says what the singular model is
now, and notes that the wrapper is repointed in #5001 rather than implying it
already is.

The other hits for that model name are all under `docs/superpowers/`, which is
excluded from the MkDocs nav. Per `src/dbt/CLAUDE.md` a spec or plan describing
a superseded design is expected; only a wrong published page is a bug. Left
those alone.

### On the non-issues list

Checked the `coalesce(...) != ...` in `WHERE` dismissal independently rather
than taking it — the precedent you cite holds, and the rule it might have
violated exempts row-independent expressions on the other side of the
comparison, which is this shape.

The "not evaluated" scoping is accurate. The data-dependent claims are my
measurements; the DY rendering question is now tracked as an Asana task due
2026-08-27 and gates merge.

Issue #5010, opened 2026-08-26
fix(dbt): wire Focus gradebook and category grades into the kipptaf grades facts for Miami


cbini:
## What's happening

`fct_grades_assignments` and `fct_grades_category` hold no Miami rows for AY2026,
and they will not until Focus gradebook data reaches the kipptaf layer.

#4995 set out to fix these two by repointing their join keys off `cc_studentid`
and `cc_yearid`. Measuring first showed the join keys are not what blocks them.
Both models inner-join a source that has no Miami AY2026 rows at all, so no
change to a join key can produce one.

Measured against prod on 2026-08-26:

| Source model | Miami AY2025 | Miami AY2026 |
| --- | --- | --- |
| `int_powerschool__gradebook_assignments_scores` | 523,021 | 0 |
| `int_powerschool__category_grades` | 255,668 | 0 |

Both source models union the district PowerSchool packages. Miami left that
package when it moved to Focus, so its archive stops at AY2025.

Focus does hold the data. `int_focus__gradebook_grades` and
`int_focus__report_card_grades` are built in the focus package. Neither has a
kipptaf wrapper, so nothing reaches the network layer. That wrapper is the work.

## Steps to reproduce

1. Count Miami AY2026 rows in `kipptaf_powerschool.int_powerschool__category_grades`. Expect 0.
1. Count Miami AY2026 rows in `kipptaf_powerschool.int_powerschool__gradebook_assignments_scores`. Expect 0.
1. Confirm the data exists upstream: count rows in the `kippmiami_focus` copy of `int_focus__gradebook_grades`.

## Where

- **Code location / dbt project:** `kipptaf`
- **Environment:** prod
- **Related:** #4995 repointed the two joins that a key swap could fix and documented these two as out of reach. #4973 is the parent. #4972 introduced the Focus branch upstream.

<details>
<summary>For Claude</summary>

> Context, not instructions.

The shape to follow is `int_students__course_enrollments` and
`int_students__course_sections` from #4972: a SIS-neutral kipptaf model that
unions the PowerSchool districts with a Focus branch, scoped by year off a
`focus_academic_year_boundary` CTE so the frozen Miami archive keeps serving
AY2020 through AY2025.

Two known obstacles beyond the wrapper itself:

- `fct_grades_assignments` joins on `students_dcid` and `fct_grades_category`
  joins on `cc_abs_sectionid`. Both are null on all 19,350 Miami AY2026 rows in
[...]

Issue #4996, opened 2026-08-26
fix(dbt): decide Miami scope per report across the 40 models filtering on the drop flags


cbini:
## What's happening

Cause 1 of #4973. Read that issue for the original analysis.

**Rewritten 2026-08-27.** The original framing has been superseded by
measurement. It said 40 models filter with the bare idiom
`where not is_dropped_section`, that Focus leaves those flags null, and that the
fix is `not coalesce(is_dropped_section, false)` in each. The first half was
right when it was written. The prescription is now dead, and the real defect is
one model up.

### The drop-flag coalesce is a no-op on all 40 files

#4968 shipped, and it made both Focus drop flags non-null. Measured against prod
2026-08-27, in `base_powerschool__course_enrollments`:

| Project | AY2026 rows | `is_dropped_section` null | `is_dropped_course` null |
| --- | --- | --- | --- |
| kippmiami | 19,363 | 0 | 0 |
| kippnewark | 45,262 | 0 | 0 |
| kippcamden | 15,406 | 0 | 0 |
| kipppaterson | 3,937 | 0 | 0 |

All 40 files read both flags from that relation and nowhere else, checked per
file alias by alias. `not is_dropped_section` already passes 18,787 of Miami's
19,363 AY2026 rows. Adding `coalesce(..., false)` would change zero rows in zero
files, and in 18 of the 40 the filter sits only on a LEFT JOIN, where it cannot
remove a row at all. This issue's own Sequencing section predicted this outcome
if #4968 landed first.

### The same trap is live on the enrollment flags

Folded into this issue's scope 2026-08-27. In `int_students__student_enrollments`
these are 100% null on Miami and non-null on every New Jersey row, because the
Focus branch never produced them and `full union all corresponding` fills null:

| Flag | Miami AY2026 null | Used bare in |
| --- | --- | --- |
| `is_enrolled_y1` | 1,760 / 1,760 | `rpt_tableau__miami_fast` where clause |
| `is_enrolled_recent` | 1,760 / 1,760 | both `rpt_gsheets__csgf_hs_*` extracts |
| `is_out_of_district` | 1,760 / 1,760 | 7 of the 40 files |
| `is_self_contained` | 1,760 / 1,760 | 3 of the 40 files |
| `is_504` | 1,760 / 1,760 | nothing — deliberate, already documented |

**`rpt_tableau__miami_fast` held zero rows of any kind** — not zero Miami rows,
[...]

Issue #4973, opened 2026-08-25
fix(dbt): make the downstream consumers actually show Miami course enrollments


cbini:
## What's happening

PR #4972 puts Miami's Focus course enrollments and sections into the kipptaf
models. 19,295 course enrollments and 823 sections for AY2026 are now there, and
`dim_student_section_enrollments`, `dim_course_sections` and `dim_courses` all
resolve them.

About 40 downstream models still show nothing for Miami. This is not a
regression — they showed nothing before that PR either, because the source held
no Miami rows at all. But it means the data now exists one layer below every
report that is supposed to use it.

Two independent causes.

### Cause 1: the null-versus-false filter trap

The Focus branch leaves `is_dropped_section` and `is_dropped_course` null,
deliberately. Focus has no drop convention, and a false would assert a verified
zero drop rate across 19,295 rows that the data does not support. That decision
is tracked in #4968 and should stand.

The consequence is wider than the drop-rate metrics that decision was about.
About 40 models filter with the bare idiom:

```sql
where not is_dropped_section
```

In SQL `not null` is null, and a `where` clause treats null as false. So every
Miami row is silently removed wherever that idiom appears.

Confirmed by grep: 59 files across `src/dbt/kipptaf/models` use the idiom, and 40
of them also read `base_powerschool__course_enrollments`,
`base_powerschool__sections`, or one of the new `int_students__` models. They
span `rpt_tableau__`, `rpt_gsheets__`, `rpt_deanslist__`, `rpt_illuminate__`,
`rpt_littlesis__`, `fct_grades_assignments`, and several `int_extracts__` models.

The fix is mechanical — `not coalesce(is_dropped_section, false)` — but the
judgment is not. Some of those reports are New Jersey-scoped by design and should
keep excluding Miami. Each needs a decision, not a sweep.

### Cause 2: joins on PowerSchool internal ids

PR #4972 fixed this for `dim_student_section_enrollments` by moving its student
join to `(student_number, schoolid, academic_year)`. The same pattern survives
elsewhere, and `cc_studentid` and `cc_yearid` are null on every Focus row:

- `int_assessments__course_enrollments` — joins `ce.cc_studentid = co.studentid`,
  and partitions a window on the same column
- `int_extracts__course_enrollments_by_term` — joins `e.studentid = s.cc_studentid`
- `fct_grades_assignments` — joins on `cc_studentid` and on
  `cc_academic_year - 1990 = enr.yearid`
[...]

PR #5011, opened 2026-08-26
fix(dbt): repoint the Miami course-enrollment joins off PowerSchool internal ids

### LINKED ISSUE #4995: fix(dbt): repoint the Miami course-enrollment joins off PowerSchool internal ids
## What's happening

Cause 2 of #4973. Read that issue for the measured analysis; this one is the
buildable half.

`cc_studentid` and `cc_yearid` are null on every Focus row, so any model joining on
them silently drops Miami. PR #4972 fixed this for
`dim_student_section_enrollments` by moving its student join to
`(student_number, schoolid, academic_year)`. The same pattern survives in 4 models:

| Model | The join |
| --- | --- |
| `int_assessments__course_enrollments` | `ce.cc_studentid = co.studentid`, and a window partitioned on the same column |
| `int_extracts__course_enrollments_by_term` | `e.studentid = s.cc_studentid` |
| `fct_grades_assignments` | `cc_studentid`, plus `cc_academic_year - 1990 = enr.yearid` |
| `fct_grades_category` | `cc_studentid` and `cc_yearid` |

`bridge_course_section_teachers` is a different shape. It inner-joins
`stg_powerschool__sectionteacher`, a PowerSchool-only staging table with no Miami
rows now that Miami is off that package. Null handling does not help it — it needs
a Focus branch or an explicit, documented exclusion.

## Why this is separate from cause 1

This half touches #4968 not at all. Cause 1 of #4973 is the drop-flag filter trap,
which #4968 owns the decision for and which needs a per-report judgment across 40
models. This half has no open questions and no dependency. Ship it on its own.

## Steps to reproduce

1. Confirm the upstream data exists: count rows in
   `int_students__course_enrollments` where `_dbt_source_project = 'kippmiami'` and
   `cc_academic_year = 2026`.
1. Count Miami AY2026 rows in each of the 4 models above. Expect zero.
1. Repoint one join to `(student_number, schoolid, academic_year)` and re-count.

## Where

- **Environment:** prod

For Claude

Context, not instructions.

Follow the join pattern PR #4972 established on
`dim_student_section_enrollments` rather than inventing a second one.

Constraints:

- New Jersey output must not move. `cc_studentid` and `cc_yearid` are non-null on
  every NJ row, so a rekeyed join has to produce identical NJ results — verify per
  model rather than assuming.
- `fct_grades_assignments` also carries `cc_academic_year - 1990 = enr.yearid`. The
  1990 offset is a PowerSchool `yearid` convention with no Focus equivalent, so
  that term needs its own handling, not just a student-key swap.
- For `bridge_course_section_teachers`, decide between a Focus branch and an
  explicit exclusion, and record the choice in the model description either way.

Acceptance criteria:

- Miami AY2026 rows present in all 4 models, verified by row count.
- `bridge_course_section_teachers` either resolves Miami or documents why it does
  not.
- NJ counts unchanged against prod, on `count(*)` plus a distinct count of the key
  columns.

Refs #4973
Refs #4972

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will put Miami back into three kipptaf models that
had been dropping it silently.

Miami moved to the Focus SIS. Focus does not populate `cc_studentid`, PowerSchool's
internal student id, so any model joining on it matched no Miami row and quietly
returned nothing. #4995 listed four such models plus a bridge table.

Measuring each one first changed the plan. Only two of the four are fixable by
swapping the join key. The other two are blocked by something a key swap cannot
reach, so this pull request documents that instead of guessing.

**Two models repointed.** `int_assessments__course_enrollments` and
`int_extracts__course_enrollments_by_term` now join on `student_number`, which both
SIS branches populate. New Jersey does not move. Miami gains 59,188 and 169,597
rows.

**The bridge gains a Focus branch.** `bridge_course_section_teachers` inner-joins a
PowerSchool-only table that holds no Miami rows for AY2026. It now reads the lead
teacher that `int_students__course_sections` already resolves for Focus. That adds
743 rows for Miami AY2026 — the only AY2026 win in this pull request.

**Two models documented, not changed.** `fct_grades_assignments` and
`fct_grades_category` read gradebook sources that hold zero Miami AY2026 rows.
Miami left the PowerSchool package, and Focus gradebook data has no kipptaf
wrapper, so no join key can produce a Miami row. Both model descriptions now say
this. #5010 tracks the real fix.

## Reviewer Notes

**The issue's premise was half wrong, and that is the main thing to check.** #4995
says repointing four joins will bring back Miami AY2026. It will not. Two of the
four inner-join a source with no Miami AY2026 rows at all. Please sanity-check that
call before merging — the numbers are in the fold-out below.

**Miami AY2026 is still missing from the two repointed models, and that is
expected.** Their `is_dropped_section` and `is_dropped_course` filters still remove
it, because Focus leaves those flags null. #4996 owns that decision and #4995 put it
out of scope. What this pull request restores is Miami's archive years, which were
also broken.

**One join deliberately does not follow the pattern #4972 set.**
`dim_student_section_enrollments` keys on `(student_number, schoolid, academic_year)`.
The assessments join never carried `schoolid`, and adding it drops New Jersey rows.
Reasoning and measured counts are in the fold-out.

**The Focus branch of the bridge carries less than the PowerSchool branch.** It has
the lead teacher only, and its effective dates are null. Both are source-data limits,
not shortcuts, and both are written into the model description.

### General

No Dagster changes.

### dbt

- [x] Include (or update) an
      summary above.
      See

No new models, so no new properties file. No exposure changes: the three touched
models keep their column sets, so nothing downstream needs coordinating. No column
was renamed or removed — `int_extracts__course_enrollments_by_term` changes an
internal CTE column only, and its output column list is unchanged. No external
sources changed.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. All values measured against prod on 2026-08-26.

### Why two of the four models were not repointed

#4995 inherited its four-model list from #4973's analysis, which was not
re-verified. Two of the four inner-join a source that holds no Miami AY2026 rows,
so the join key is not what blocks them:

| Source model | Miami AY2025 | Miami AY2026 |
| --- | --- | --- |
| `int_powerschool__gradebook_assignments_scores` | 523,021 | 0 |
| `int_powerschool__category_grades` | 255,668 | 0 |

Both union the district PowerSchool packages. `src/dbt/kippmiami/packages.yml` no
longer lists `powerschool`, so Miami's archive stops at AY2025. Focus holds the
data (`int_focus__gradebook_grades`, `int_focus__report_card_grades` in the focus
package) but neither has a kipptaf wrapper. Tracked in #5010.

Their other join keys are null for Miami too, so a `cc_studentid` swap alone would
not have been enough even with a source: `students_dcid` and `cc_abs_sectionid` are
null on all 19,350 Miami AY2026 rows of `base_powerschool__course_enrollments`,
alongside `cc_studentid`, `cc_yearid`, `is_dropped_section`, `is_dropped_course`
and `sections_no_of_students`. All six are non-null on every NJ row.

`fct_grades_assignments` also carries `cc_academic_year - 1990 = enr.yearid`. That
term is untouched here. The 1990 offset is a PowerSchool `yearid` convention with
no Focus equivalent, and it only needs solving once #5010 supplies a Focus branch.

### The breakage is wider than AY2026

`int_assessments__course_enrollments` holds zero Miami rows in **every** year, not
just AY2026. #4972 moved Miami's student enrollments wholesale onto Focus back to
AY2018, and Focus leaves `studentid` null, so the archive years broke too. The
prod table was rebuilt at 19:03 on 2026-08-26, after #4972 merged at 14:26, so
that state is post-merge. `int_extracts__course_enrollments_by_term` shows the
same zero-Miami shape.

### NJ parity method

A straight new-vs-prod count is not a valid check here: the live year drifts within
a single day, exactly as #4972 warned. A first comparison showed Newark -24 and
Camden -8, both confined to the live year.

So both join variants were run against the **same live upstream** in one query,
which removes drift entirely. Result: zero differing rows for Camden, Newark and
Paterson in every year, on `count(*)` and on a distinct count of the grain key.
The only differences are Miami.

Key-level parity backs this up. In `int_students__student_enrollments` at
`rn_year = 1`, `(academic_year, studentid)` and `(academic_year, student_number)`
have identical distinct counts equal to the row count in all three NJ regions. In
`base_powerschool__course_enrollments`, distinct `cc_studentid` = distinct
`students_student_number` = distinct pairs for all three.

Miami gains, by model:

| Model | Miami rows gained | Years |
| --- | --- | --- |
| `int_assessments__course_enrollments` | 59,188 | AY2018-AY2025 |
| `int_extracts__course_enrollments_by_term` | 169,597 | AY2020-AY2025 |
| `bridge_course_section_teachers` | 743 | AY2026 |

### Why the assessments join does not gain schoolid

Adding `schoolid` to that join drops NJ rows where a student's course school
differs from their enrollment school: Newark -2,271, Camden -389, Paterson -63.
The student-key swap alone is exactly NJ-neutral. `dim_student_section_enrollments`
can carry `schoolid` because its join always had it;
`int_assessments__course_enrollments` never did.

`int_extracts__course_enrollments_by_term` already joined on `schoolid` and
`academic_year`, so only the student key changed there.

### Bridge verification

Running the compiled model against prod:

- PowerSchool branch: 87,557 rows, matching the prod table exactly. Zero rows lost.
- Focus branch: 743 rows.
- Primary-key duplicates on `(course_section_key, staff_key, effective_start_date)`: 0.
- Rows in prod but not in the new model: 0.
- Rows in the new model but not in prod: 743, all Focus.

743 of Miami's 823 AY2026 sections resolve a lead teacher. The 80 that do not are
#4972's documented floor: 67 course periods with no teacher assigned in Focus, plus
13 whose user resolves to no staff roster row.

The Focus branch is year-scoped off the same `focus_academic_year_boundary` CTE the
sibling models use. Without it the branch would emit a duplicate Lead Teacher row
for every Miami archive section, since those sections carry both a `sectionteacher`
row and a resolved `teachernumber`. `stg_powerschool__sectionteacher` does hold
19,529 Miami rows — #4995 says it has none, which is wrong for the archive years
though right for AY2026.

### Checked and clear

`base_powerschool__sections` is a VIEW in prod as of 14:53 on 2026-08-26, so
#4972's required post-merge full-refresh did happen. The bridge comparison above
would have been meaningless against a stale table.

### Not done

The `is_dropped_section` / `is_dropped_course` filters are untouched, per #4995 and
#4996. That is why Miami AY2026 is still absent from the two repointed models.

</details>

Closes #4995
Refs #4973
Refs #4996
Refs #5010
Refs #4972

### DIFF
--- a/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__course_enrollments.sql
+++ b/src/dbt/kipptaf/models/assessments/intermediate/int_assessments__course_enrollments.sql
@@ -19,20 +19,41 @@ with
 
             co.grade_level + 1 as illuminate_grade_level_id,
 
+            -- partitioned on students_student_number rather than cc_studentid:
+            -- Focus leaves cc_studentid null on every Miami row, so a partition
+            -- on it collapses all Miami rows into a single group. The two keys
+            -- are exactly 1:1 within every NJ region (verified against prod:
+            -- distinct cc_studentid = distinct students_student_number =
+            -- distinct pairs, in all three), so NJ output does not move.
             max(ce.is_advanced_math) over (
                 partition by
                     ce._dbt_source_project,
-                    ce.cc_studentid,
+                    ce.students_student_number,
                     ce.cc_academic_year,
                     ce.courses_credittype
             ) as is_advanced_math_student,
         from {{ ref("base_powerschool__course_enrollments") }} as ce
+        -- cc_studentid is null on every Focus row, and #4972 moved Miami's
+        -- student enrollments wholesale onto Focus for every year back to
+        -- AY2018 -- so this join dropped all 93,858 Miami rows, archive years
+        -- included, not just AY2026. student_number carries both SIS branches.
+        --
+        -- Deliberately NOT keyed on schoolid, unlike the (student_number,
+        -- schoolid, academic_year) join in dim_student_section_enrollments:
+        -- this join never carried schoolid, and adding it drops NJ rows where a
+        -- student's course school differs from their enrollment school --
+        -- Newark -2,271, Camden -389, Paterson -63, measured against prod. The
+        -- student-key swap on its own is exactly NJ-neutral.
         inner join
             {{ ref("base_powerschool__student_enrollments") }} as co
-            on ce.cc_studentid = co.studentid
+            on ce.students_student_number = co.student_number
             and ce.cc_academic_year = co.academic_year
             and ce._dbt_source_project = co._dbt_source_project
             and co.rn_year = 1
+        -- TODO(#4996): is_dropped_course is null on every Focus row and
+        -- `not null` is null, so this still removes Miami AY2026 even with the
+        -- join above repointed. Miami's archive years carry real flags and do
+        -- flow through. #4996 owns the coalesce decision for this filter.
         where not ce.is_dropped_course
 
         union all
--- a/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__course_enrollments.yml
+++ b/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__course_enrollments.yml
@@ -6,6 +6,14 @@ models:
       Illuminate assessments are joined against. One row per (student, academic
       year, subject area, date enrolled), deduplicated to the latest exit per
       enrollment span.
+
+
+      Covers Miami through AY2025 only. The student join keys on
+      `student_number` rather than PowerSchool's internal `cc_studentid`, which
+      Focus leaves null -- before that change this model held no Miami rows in
+      any year. Miami AY2026 forward is still excluded, because the `not
+      is_dropped_course` filter drops rows whose flag is null and Focus never
+      sets it. #4996 owns that filter.
     config:
       materialized: table
       meta:
--- a/src/dbt/kipptaf/models/marts/bridges/bridge_course_section_teachers.sql
+++ b/src/dbt/kipptaf/models/marts/bridges/bridge_course_section_teachers.sql
@@ -1,31 +1,103 @@
-select
-    {{
-        dbt_utils.generate_surrogate_key(
-            ["sec.sections_dcid", "sec._dbt_source_project"]
-        )
-    }} as course_section_key,
+-- Focus, Miami's system of record from AY2026 forward, has no equivalent of
+-- PowerSchool's sectionteacher table, so the PowerSchool branch below resolves
+-- no teacher for a Miami section in those years. Focus models a single lead
+-- teacher per course period, already resolved to the staff roster in
+-- int_students__course_sections, so the Focus branch reads that resolved
+-- teachernumber rather than re-deriving it.
+--
+-- Two Focus limitations, both source-data shape rather than modeling gaps:
+--
+-- * Lead teacher only. Focus carries co-teachers in stg_focus__co_teachers,
+-- which is not surfaced into kipptaf, so the Co-teacher, Gradebook Access
+-- and Blended Learning roles PowerSchool supplies have no Miami
+-- counterpart from AY2026 forward.
+-- * effective_start_date and effective_end_date are null. Focus attaches a
+-- teacher to a course period without dating the assignment; inventing the
+-- term dates would assert an assignment history the source does not
+-- record.
+--
+-- 743 of Miami's 823 AY2026 sections resolve a lead teacher. The 80 that do
+-- not are the documented Focus floor from #4972: 67 course periods with no
+-- teacher assigned in Focus at all, plus 13 whose user resolves to no staff
+-- roster row.
+with
+    -- The year boundary keeps the two branches disjoint. Miami's frozen
+    -- PowerSchool archive still holds sectionteacher rows through AY2025, and
+    -- those sections also carry a resolved teachernumber, so an unscoped Focus
+    -- branch would emit a duplicate Lead Teacher row for every archive
+    -- section. coalesce to 9999 fails toward emitting nothing rather than
+    -- duplicating, when int_focus__schedule is empty in an unbuilt --defer dev
+    -- copy.
+    focus_academic_year_boundary as (
+        select coalesce(min(academic_year), 9999) as min_academic_year,
+        from {{ ref("int_focus__schedule") }}
+    ),
 
-    {{ dbt_utils.generate_surrogate_key(["sr.employee_number"]) }} as staff_key,
+    powerschool_teachers as (
+        select
+            {{
+                dbt_utils.generate_surrogate_key(
+                    ["sec.sections_dcid", "sec._dbt_source_project"]
+                )
+            }} as course_section_key,
 
-    r.name as `role`,
+            {{ dbt_utils.generate_surrogate_key(["sr.employee_number"]) }} as staff_key,
 
-    cast(st.start_date as date) as effective_start_date,
-    cast(st.end_date as date) as effective_end_date,
+            r.name as `role`,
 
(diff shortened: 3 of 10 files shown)


cbini:
Verdicts on all six findings, verified against the code before acting. Four
fixed in 8400c2995; two need no change.

**1. Stale `rn` column description — no change needed, already fixed.**
Corrected in 67059ac04, which this review predates. That commit repointed the
model's uniqueness test off `studentid` (see below) and updated the `rn`
description in the same file. It now reads
`(project, academic_year, student_number, course_number, quarter)`.

**2. `source_system` not updated for the Focus branch — fixed.** Confirmed the
precedent: `dim_courses.yml` uses `source_system: PowerSchool and Focus` at
lines 36, 46, 57 and 91. All three bridge columns now carry that tag.
`source_model` / `source_column` still name the PowerSchool relations, matching
how `dim_courses` keeps a single upstream pointer while the description
qualifies the per-branch behavior.

**3. No properties description on the two changed models — fixed.** Fair point,
and the asymmetry was the wrong way round: the two models whose SQL did not
change got a multi-paragraph note while the two that did change got nothing.
Both now state that they cover Miami through AY2025 only, why the student join
moved off `cc_studentid`, and that Miami AY2026 stays excluded pending #4996 —
so a docs-site reader sees it, not just someone reading the inline `TODO`.

**4. Possible ST06 ordering deviation — declining, verified non-issue.** Ran
`trunk check --force --no-fix` on `bridge_course_section_teachers.sql`
directly: no issues, only a `sqlfmt` formatting item that the pre-commit hook
resolved. Your own low-confidence read was right — the Jinja-templated
`generate_surrogate_key` slices are why it does not fire.

**5. `select *` in the UNION ALL branches — fixed, and going further than you
suggested.** You flagged this at low-medium confidence and did not require a
change, citing precedent in `dim_surveys.sql` and `bridge_survey_questions.sql`.
Taking it anyway, because `src/dbt/CLAUDE.md` is explicit that enumerating is
the *correctness* fix and not just the lint fix — BigQuery binds UNION ALL
branches by position, so a future edit reordering one branch binds the wrong
columns, and `role` / `effective_start_date` are not typed differently enough to
guarantee a loud failure. Precedent existing elsewhere is not a reason to add
another instance. Five columns, enumerated in both branches, with a comment
saying why.

**6. No regression guard — fixed.** This is the finding I would rank highest,
because the failure mode is precisely the one this PR exists to fix: both models
returned zero Miami rows in every year while every test stayed green. Added
`tests/test_focus_course_enrollment_joins_resolve_miami.sql`, following the
shape of `test_focus_course_sections_teacher_resolves.sql`.

It asserts row presence rather than a per-year count floor. A floor would either
drift stale as enrollment moves or fire on legitimate change, whereas zero is
always wrong. It is also deliberately not year-scoped: Miami AY2026 is still
excluded from both models by their `is_dropped_*` filters until #4996 lands, so
a per-year assertion would fail today by design. Severity `error` — a silent
re-drop should not be a warning.

Verified it passes against the PR-branch build: 59,188 Miami rows in
`int_assessments__course_enrollments` and 169,597 in
`int_extracts__course_enrollments_by_term`, so the test returns no rows.

---

On the data-dependent claims you flagged as unverifiable without warehouse
access — the NJ-neutrality evidence is worth restating, because a naive check
produces a false positive. Comparing the new build to the prod table shows
Newark -24 and Camden -8, which looks like the repoint moving NJ. It is not: it
is upstream drift in the live year between prod's build time and the query. The
real check runs both join variants against the *same* live upstream in one
query, which removes drift. That returns zero differing rows for Camden, Newark
and Paterson in every year, on `count(*)` and on a distinct count of the grain
key.

Issue #5013, opened 2026-08-26
fix(dbt): int_focus__student_enrollments is unqueryable in prod after the enrollment reshape


cbini:
## What&#39;s happening

`kipptaf_focus.int_focus__student_enrollments` cannot be read in prod. Any query
against it fails before it runs:

```text
Unrecognized name: student_first_name; failed to parse view
&#39;teamster-332318.kipptaf_focus.int_focus__student_enrollments&#39; at [15:26]
```

PR #5009 turned `int_focus__student_enrollment` into a thin model and moved the
denormalized columns to the new `int_focus__student_enrollment_roster`. The
kipptaf model still reads the thin one with `e.* except (...)`, so the columns it
expects are gone.

The stored prod view was created at 20:04:55 on 2026-08-26, while the upstream
was still wide. It lists column names that no longer exist, so it is now
invalid. Rebuilding it will not help, because the code it would rebuild from
points at the wrong upstream.

`int_students__student_enrollment_union` still returns 134,301 rows only because
it is a materialized table holding pre-break data. **Its next rebuild will
fail.** That model is the enrollment spine for all four regions, not just Miami.

## Steps to reproduce

1. Run `select count(*) from kipptaf_focus.int_focus__student_enrollments` against
   prod. It fails to parse.
1. Confirm the upstream lost the columns: `int_focus__student_enrollment` in
   `kippmiami_focus` has 44 columns and none of them is `student_first_name`.
1. Confirm the code still points at the thin model: line 6 of
   `src/dbt/kipptaf/models/focus/intermediate/int_focus__student_enrollments.sql`
   sources `int_focus__student_enrollment`.

## Where

- **Code location / dbt project:** `kipptaf`, with the upstream in the `focus`
  package
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** introduced by #5009. Surfaced by a dbt
  Cloud CI failure on #5008, where `int_assessments__course_enrollments` reads
  through the broken view.

<details>
<summary>For Claude</summary>

Context, not instructions.

Relevant models:

- `src/dbt/kipptaf/models/focus/intermediate/int_focus__student_enrollments.sql`
  — sources `int_focus__student_enrollment` and projects
  `e.* except (school_number, school_title, school_state_school_id, grade_level_short_name, school_level)`.
  Every one of those five columns now lives on
  `int_focus__student_enrollment_roster`, so repointing the `source()` is the
  fix.
[...]

PR #5020, opened 2026-08-27
feat(dbt): branch kipptaf term grades onto Focus and retire 3 superseded wrappers

### LINKED ISSUE #4999: feat(dbt): branch the kipptaf final-grades and GPA chain onto Focus for Miami

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will put Miami's AY2026 grades into
`fct_grades_term` for the first time, and finish the Focus enrollment-wrapper
cleanup that #5013 started.

Miami reads empty in `fct_grades_term` today. The fact reads only
`base_powerschool__final_grades`, and the frozen `kippmiami_powerschool` archive
holds zero rows for that model. Miami has in fact never had a row in
`fct_grades_term` or `fct_grades_gpa`, in any year. This is new coverage, not a
repair.

This adds `int_students__final_grades`, a SIS-neutral term final-grade spine. It
unions the PowerSchool branch with Miami's Focus report card grades, conformed to
the PowerSchool column names. The PowerSchool branch is scoped by year off
`int_students__sis_cutover`, the same boundary the five `int_students__`
attendance models already share, so the frozen archive stays readable.
`fct_grades_term` now reads that spine.

It also retires 3 staging wrappers that nothing reads any more —
`stg_focus__schools`, `stg_focus__students` and `stg_focus__student_enrollment`.
They are disabled, not deleted, so their prod tables stay readable. And it adds
the wrapper for the reshaped `int_focus__student_enrollment`, whose source entry
#5013 left without one.

`fct_grades_gpa` is deliberately left out. See Reviewer Notes.

## Reviewer Notes

**The enrollment join changes key.** It moves from `(studentid, yearid)` to
`(student_number, academic_year)`. It has to:
`int_students__student_enrollment_union` leaves `studentid` and `yearid` null on
all 10,075 Miami rows, so the old key can never match a Focus row. The swap is
safe, and measured rather than assumed — the two key pairs are exactly 1:1 inside
every NJ district, and the rewritten join returns identical counts.

**NJ output is unchanged.** 232,887 rows and 232,856 distinct `grades_term_key`.
Both are exact matches for what `fct_grades_term` holds in prod today.

**77 Focus rows do not reach the fact,** and all 77 are accounted for. 65 belong
to students who enrolled after Q1 began, and 12 to students who left before it.
The fact's covering-enrollment filter treats NJ the same way.

**Miami `term_key` will be null at first.** The reporting-terms sheet has no
Miami rows for AY2026 yet. That is a sheet gap, not a code gap, and the
relationships test passes on null.

**GPA is split out.** Focus computes GPA and class rank itself, in a table we do
not ingest. Recomputing it through a branched PowerSchool chain has almost no
inputs to work from. A follow-up issue covers ingesting the vendor's own numbers.

**No post-merge cleanup.** An earlier revision of this PR deleted the 3 wrappers.
`src/dbt/CLAUDE.md` says to disable rather than delete, and says so for exactly
this case — a model with no exposure and zero remaining `ref()`s. It also says
not to drop the retired relation. So the 3 prod tables stay in `kipptaf_focus`,
frozen, and there is nothing to clean up after merge.

### General

### dbt

- [x] Include (or update) an
      external consumer; `fct_grades_term` keeps its existing Cube exposure
- [x] **Breaking change?** No contracted column is renamed or removed.
      `fct_grades_term` keeps its exact column list.
      with **`--target staging`** — not needed. No external source changed, and
      all 10 relations CI defers to were verified present and current in the
      `zz_stg_*` schemas.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Every number below was measured against prod, by
inlining the compiled model definition and rewriting the dev schemas to prod. None
is inferred.

### Why the branch sits where it does

`int_students__final_grades` copies `int_students__course_enrollments` exactly:
`powerschool_conformed full union all corresponding focus_conformed`, year-scoped
off `int_students__sis_cutover`. That is the altitude the attendance phase set on
#4924 — union two derived models, do not recompute hashes at network altitude.
Branching `base_powerschool__final_grades` in place, as #4999 proposed, would put a
differently-shaped Focus arm inside a `base_` model that is a pure 4-way
`union_relations` over district PowerSchool sources.

### Why the archive year-scope is a no-op here, but still correct

`kippmiami_powerschool.base_powerschool__final_grades` holds **0 rows**, as does
`int_powerschool__final_grades_pivot`. Only `stg_powerschool__storedgrades`
(239,545) and the GPA chain survived the freeze. `base_powerschool__final_grades`
is current-year-only in every region, which is why the archive kept nothing. The
year-scope costs nothing and is the shape every sibling conform uses, so it stays.

Corollary: the "Miami historical reconciliation against the archive" acceptance
criterion on #4999 has no target for term grades. Confirmed from the other
direction too — `fct_grades_gpa` in `kipptaf_marts` holds Camden, Newark and
Paterson only, for AY2024 through AY2026.

### The join-key swap, measured

`(studentid, yearid)` vs `(student_number, academic_year)` in
`int_students__student_enrollment_union`, NJ only:

| region | rows | distinct by PS id | distinct by number | distinct studentid | distinct student_number |
| kippcamden | 23,964 | 23,964 | 23,964 | 5,526 | 5,526 |
| kippnewark | 98,199 | 98,199 | 98,199 | 18,225 | 18,225 |
| kipppaterson | 2,084 | 2,084 | 2,084 | 1,042 | 1,042 |

`distinct yearid` equals `distinct (yearid, academic_year)` pairs in all three. So
the mapping is 1:1 on both halves of the key.

Running both join variants over `base_powerschool__final_grades`: kippcamden
53,480 both ways, kippnewark 164,097 both ways, kipppaterson 15,306 both ways.
Delta 0 everywhere.

`stg_powerschool__students` with `dcid >= 1` is unique on
`(id, _dbt_source_project)` in all 4 regions with zero null `student_number`, so
resolving the number cannot fan out. The join is nonetheless LEFT, not INNER —
INNER would silently drop any final-grade row whose student fails the placeholder
filter, and the join type is what keeps the NJ population stable rather than the
current measurement of it.

### Full fact parity

Built `fct_grades_term` from the new spine and compared to prod:

| | built | prod |
| --- | --- | --- |
| NJ rows | 232,887 | 232,887 |
| NJ distinct `grades_term_key` | 232,856 | 232,856 |
| Miami rows | 1,516 | 0 |

Per region built: kippcamden 53,480, kippnewark 164,101, kipppaterson 15,306,
kippmiami 1,516. The Newark figure runs 4 above the pre-`reporting_terms` count
because the existing `left join` to reporting terms fans out on 4 rows. Prod does
the same, which is why the totals still match exactly.

### cc_dcid parity

`cc_dcid` maps to Focus `student_schedule_id`, matching
`int_students__course_enrollments`. That is what makes
`student_section_enrollment_key` resolve against `dim_student_section_enrollments`.
Verified: all 1,593 AY2026 grade rows join `int_focus__schedule` on
`(student_id, course_period_id, academic_year)` to exactly one schedule row — 0
orphans, 0 fan-out.

### The 77 dropped rows

Attributed completely, with no residual:

| cause | rows |
| --- | --- |
| student enrolled after the marking period began | 65 |
| student withdrew before the marking period began | 12 |
| no AY2026 stint at all | 0 |
| stint only at a different school | 0 |

Q1 begins 2026-08-12; the late entry dates run through 2026-08-27. This is the
fact's existing `termbin_start_date >= entrydate` semantics, which drops the same
class of row in NJ.

### term_key

`stg_google_sheets__reporting__terms` has 80 Miami RT rows but
`max(powerschool_year_id)` is 35, meaning AY2025. AY2026 is 36. So every Miami row
lands with `term_key` null until the sheet gains AY2026 rows.
`relationships_fct_grades_term_term_key__term_key__ref_dim_terms_` excludes nulls
and passes. `storecode` needs no translation — Focus `marking_period_short_name`
reads `Q1`, the same domain as the sheet's `name`.

### Focus column mapping

Nulls are explicit casts, never a copied term value: a YTD column that silently
equals the term grade reads as a real rollup and is not one.

| PowerSchool | Focus |
| --- | --- |
| `cc_dcid` | `int_focus__schedule.student_schedule_id` |
| `yearid` | `academic_year - 1990` |
| `storecode` | `marking_period_short_name` |
| `termbin_start_date` / `_end_date` | `marking_period_start_date` / `_end_date` |
| `term_letter_grade` | `grade_title` |
| `term_percent_grade` | `percent_grade` |
| `term_grade_points` | `gpa_points` |
| `potential_credit_hours` | `credits` |
| `exclude_from_gpa` | `if(affects_gpa = 'Y', 0, 1)` — PS stores the exclusion, Focus the inclusion |
| `studentid`, all `*_adjusted`, all `y1_*`, `citizenship`, `lastgradeupdate` | null, no analog |

`lastgradeupdate` is null because `modified_date` is null on every
`student_report_card_grades` row, as recorded on #4986.

The Focus branch is also year-scoped, from the other side: imported course history
runs back to AY2012 and would double-count against the archive for the years the
archive covers.

### The GPA split, and why

The evidence against recomputing:

- The chain is ~1,000 lines across `int_powerschool__gpa_term`, `_cumulative` and
  `_cumulative_year`, and reads `stg_powerschool__storedgrades` and
- Focus has no `storedgrades` analog. The nearest thing,
  `int_focus__report_card_grades where course_history = 'Y'`, is **0 rows for
  AY2026**, so cumulative and prior-year GPA have no input at all for the cutover
  year.
- Only 145 of 1,593 Miami AY2026 rows carry `affects_gpa = 'Y'`.
- Miami is absent from `fct_grades_gpa` for every year, so there is no regression
  pressure to ship something approximate.

`student_gpa_calculated` remains absent from
`src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml`, re-verified on
this branch. Ingesting it also brings class rank, which the PowerSchool chain does
not produce at all.

### Gate and CI state

P3 gate re-verified 2026-08-27, not taken on trust:
`kippmiami_focus.int_focus__report_card_grades` (35 columns),
`int_focus__student_enrollment` (44) and `int_focus__student_enrollment_roster`
(41) all materialized in prod at 13:30–14:01 UTC that day.

No `zz_stg` reseed was needed. All 10 relations CI defers to were checked present:
`zz_stg_kipptaf_focus.int_focus__{schedule,students,schools}`,
`zz_stg_kipptaf_powerschool.{stg_powerschool__students,base_powerschool__final_grades}`,
`zz_stg_kipptaf_students.{int_students__sis_cutover,int_students__student_enrollment_union}`,
`zz_stg_kipptaf_google_sheets.{stg_google_sheets__people__locations,stg_google_sheets__reporting__terms}`,
and `zz_stg_kippmiami_focus.int_focus__report_card_grades`.

One staleness exists and does not affect this PR:
`zz_stg_kippmiami_focus.int_focus__student_enrollment_roster` has 40 columns to
prod's 41. The missing column is `is_second_school`, added by #5008 earlier the
same day. Nothing here reads it, and
`zz_stg_kipptaf_students.int_students__student_enrollment_union` already matches
prod at 52 columns with all 5 join columns present.

### Local verification

`dbt parse` clean. `dbt compile` clean over all 4 changed or new models.
`dbt build --empty` clean across the descendant graph — PASS=10, ERROR=0,
including the new grain test and all 5 existing `fct_grades_term` tests.
`trunk check --force` clean on all 8 files.

### After the deploy

Compare each wrapper's stored `input_data_version` materialization tag against the
upstream's current `data_version` before trusting a green build. The kipptaf
wrappers are views, and #4290 does not self-heal.

</details>


cbini:
## dbt Cloud failure is staging drift, not this branch

The failing node is
`int_powerschool__ada_term_pivot__weighted_year_matches_source`, a singular test
on the NJ PowerSchool ADA chain. Nothing in this branch touches ADA.

**The same test failed identically on #5017 and #5018**, both run within 30
minutes of this one, with the same count of 3,535. It is a repo-wide condition
today.

### Where the mismatch lives

| relation set | mismatches |
| --- | --- |
| prod `kipptaf_powerschool` | **0** |
| staging `zz_stg_kipptaf_powerschool` | **3,535** |

3,535 is exactly what CI reported. The test reads
`zz_stg_kipptaf_powerschool` on **both** sides of its comparison, so it never
touched anything this PR built.

All 3,535 rows are Camden, Newark and Paterson. Zero Miami, zero grades.

### Why staging disagrees with prod

The staging pair was seeded 2026-08-26 22:51, when prod itself was still
inconsistent. Prod's `int_powerschool__ada_term_pivot` rebuilt today at
2026-08-27 14:17 and prod now agrees. Staging still holds the older mismatched
pair.

| relation | prod built | staging built |
| --- | --- | --- |
| `int_powerschool__ada_term` | 2026-08-25 16:55 | 2026-08-26 22:51 |
| `int_powerschool__ada_term_pivot` | 2026-08-27 14:17 | 2026-08-26 22:51 |

### The fix is a staging refresh, not a code change

Reseed `zz_stg_kipptaf_powerschool.int_powerschool__ada_term` and
`int_powerschool__ada_term_pivot` from prod, then re-run CI with a fresh build
rather than a retry — `dbt retry` replays the prior run's compiled SQL and would
not pick up the changed relations. That should clear #5017 and #5018 as well.

### This PR's own nodes all built and passed

CI created exactly the 4 models this branch adds or changes, and every test on
them passed:

- `dbt_cloud_pr_..._5020_focus.int_focus__report_card_grades`
- `dbt_cloud_pr_..._5020_focus.int_focus__student_enrollment`
- `dbt_cloud_pr_..._5020_students.int_students__final_grades`
- `dbt_cloud_pr_..._5020_marts.fct_grades_term`

Passing: the new `(cc_dcid, storecode, _dbt_source_project)` grain test, plus all
5 pre-existing `fct_grades_term` tests. It did **not** build
`int_powerschool__ada_term` or its pivot — no such relations exist in the PR
schema, which is what confirms the failing test read the staging copies.

### Separately: `claude-review` errored

It posted "Claude encountered an error" rather than findings, so there is no
review to address yet. It only fires on `opened` and `ready_for_review`, so a
draft toggle is needed to re-run it.

cbini:
## dbt Cloud is green; both warnings triaged, neither is from this branch

The staging reseed cleared it. `zz_stg_kipptaf_powerschool` ADA mismatches went
3,535 to 0, and the run passed on `e4f81fefcc`.

Two test warnings remain on the passing run. Both are pre-existing, and neither
involves a Miami row.

### 1. `unique_fct_grades_term_grades_term_key` — 31 results

Prod `fct_grades_term` holds 232,887 rows against 232,856 distinct
`grades_term_key`. That is 31 duplicates, already there.

This branch adds zero. Miami contributes 1,516 rows and 1,516 distinct keys, so
the duplicate count is unchanged at 31.

### 2. `relationships_..._student_section_enrollment_key` — 166 results

Every one of the 166 is a NJ row. **Miami contributes zero.**

| region | orphan fact rows | distinct orphan keys |
| --- | --- | --- |
| kippcamden | 44 | 11 |
| kippnewark | 106 | 29 |
| kipppaterson | 16 | 4 |
| kippmiami | **0** | **0** |

44 + 106 + 16 = 166, matching the warning exactly.

Zero Miami orphans is the design goal met, not a coincidence: `cc_dcid` maps to
the Focus `student_schedule_id`, the same value
`int_students__course_enrollments` uses, so all 1,516 Miami keys resolve against
`dim_student_section_enrollments`.

The same test run against **prod** dims returns 56, not 166. The gap is the
stale `zz_stg_kipptaf_marts` clone the CI test reads as its parent, which is the
known staging-defer drift, not a real orphan increase.

cbini:
## The red `claude-review` check is a broken workflow, not a review

Filed as #5028. It fails the same way on #5017 and #5018.

Claude never runs. The job dies installing plugins:

```text
Installing plugin: dignified-python@dagster
✘ Failed to install plugin "dignified-python@dagster":
  Plugin "dignified-python" not found in marketplace "dagster"
```

`dagster-io/skills` dropped that plugin from its marketplace. Checked all four
marketplaces the workflow declares — `dignified-python` is in none of them. A
Codespace plugin cache from 2026-08-05 still holds it under `cache/dagster/`, so
the removal happened upstream and nothing in this repo changed.

I re-fired the check once with a draft toggle to rule out a transient failure. It
failed identically the second time.

**So this PR has no Claude review, and cannot get one until #5028 is fixed.** The
self-review checkbox for it stays unchecked deliberately. Worth a human eye on
the two things that would normally draw review attention:

1. The enrollment join key swap in `fct_grades_term`, from `(studentid, yearid)`
   to `(student_number, academic_year)`. Measured row-identical for all 3 NJ
   regions, but it is the one behavior-affecting change to an existing model.
1. The `left join` to `stg_powerschool__students` in
   `int_students__final_grades`. It is deliberately not an inner join — see the
   comment on the CTE.

cbini:
## Review responses — 4 fixed, 2 informational, 0 declined

Pushed in `63e9ff79a7`. I verified each cited convention against
`src/dbt/CLAUDE.md` and the codebase before acting; every citation checked out.

### 1. Models deleted instead of disabled — **fixed**

Correct, and it is a real violation. `src/dbt/CLAUDE.md:830`:

> Retiring a model is always a disable, never a delete... Applies even when the
> model has no exposure and zero remaining `ref()`s... Do not propose deletion as
> the tidier option, and do not issue a `drop view` for the orphaned relation.

#5001 asked for a delete. The repo convention outranks the issue text, and I
should have caught that.

Restored all 6 files with `config: enabled: false` on each model, plus the 3
source entries so the disabled models still resolve their `source()` calls. None
of the 3 carries data tests, so the "disable the tests too" clause has nothing to
act on here — confirmed by reading each properties file.

`dbt parse` proves the outcome rather than the intent: all 3 land in the
manifest's `disabled` collection, and zero enabled nodes carry those names.

This also reverses the PR body, which claimed post-merge cleanup was needed. The
same rule forbids the `drop view`, so the 3 relations stay frozen and there is
nothing to clean up. PR body corrected.

### 2. ST06 column ordering — **fixed**

`academic_year - 1990 as yearid` is a calculation and was sitting between plain
column refs. Convention confirmed at `src/dbt/CLAUDE.md:1207`, and the sibling
this model copies does keep every plain ref first.

Worth noting for accuracy: sqlfluff does **not** flag it. `trunk check --force`
was clean before this change and is clean after, on all 9 files — ST06 buckets
`cast()` as a simple target but does not appear to catch arithmetic. So the
finding is right about the convention and wrong about the CI risk; the "trunk
check clean" claim was not stale. Fixed on the convention's merits.

### 3. Comments longer than the code — **fixed**

Confirmed at `src/dbt/CLAUDE.md:1255`, including the clause that pre-empts the
sibling-model defence:

> A comment longer than the expression it annotates belongs in the properties
> file; the repo's existing multi-paragraph SQL comments are not a precedent to
> extend.

Trimmed the 8-line `powerschool_students` comment to 2 lines and the 6-line
`cc_dcid` comment to 2. The join-key rationale and the `cc_dcid` 1:1 result moved
into the model `description:`, which is where they belonged.

### 4. `student_number` nullable into a downstream inner join — **fixed**

Fair, and the reasoning is right: measured-at-zero is not test-enforced. Added a
`not_null` test at `severity: warn`, with a comment explaining the failure mode.
Warn rather than error because a dropped row is a coverage gap, not a corrupt
row.

Answering the verification you could not run: `student_number` is null on **0**
rows today, on both branches.

### 5. `int_powerschool__final_grades_rollup` still Miami-blind — **no change**

Agreed on all counts, including that it is not a regression. That model reads
`base_powerschool__final_grades` directly, and the Miami archive holds 0 rows
there, so it was already Miami-blind before this branch. Y1 columns are null on
the Focus branch, so wiring it up would produce empty rollups rather than useful
ones.

Not filing a follow-up yet — #5021 already covers the Focus GPA work, and
whether Miami wants failing-course tracking is a question for the Miami team, not
an engineering gap.

### 6. Schedule inner-join invariant is untested — **noted in code**

The mechanism is correct, and I verified it against prod: **all 15,278**
course-history rows have a null `course_period_id`, against **0** of the AY2026
rows. So those rows would drop out of the schedule inner join even without the
year filter.

Added a comment recording the invariant and the measurement. Not adding a test:
the year filter is the real guard, and a test asserting "no course-history rows
at or after the cutover" would fire on a legitimate backfill. The comment tells
whoever does that backfill what to re-check.

---

Verification after the fixes: `dbt parse` clean, `dbt build --empty` clean across
the descendant graph (PASS=11, ERROR=0, including the new `not_null` test),
`trunk check --force` no issues over all 9 files.

Issue #5024, opened 2026-08-27
fix(dbt): 185 Miami AY2026 stints have an exitdate on or before their entrydate


cbini:
## What's happening

185 of Miami's 1,760 AY2026 student stints (10.5%) have an exitdate on or before
their entrydate. 91 are the same day. 94 have an exitdate that falls before the
entrydate.

A stint is supposed to be a window a student attended. A window that ends
before it starts, or lasts zero days, is not one.

This is Focus-only and new. Zero stints have this shape in any of the three New
Jersey regions, and zero have it in Miami's PowerSchool archive years.

| Region | AY2024 | AY2025 | AY2026 |
| --- | --- | --- | --- |
| kippmiami | 0 | 0 | 185 of 1,760 (10.5%) |
| kippnewark | 0 | 0 | 0 |
| kippcamden | 0 | 0 | 0 |
| kipppaterson | 0 | 0 | 0 |

### It breaks the course enrollment join

`dim_student_section_enrollments` attaches a course enrollment to a stint by
overlapping their date ranges. Nothing can overlap a window that has no
duration.

Measured 2026-08-27: 1,332 of the 1,886 orphaned Miami AY2026 course
enrollments (71%) come from these 185 stints. That is the largest single cause
of the orphan residual tracked on #4970, and it is bigger than the two causes
already split out into #5002 and #5003 combined.

Widening the join will not help and should not be attempted. #4970 rules it
out, and there is no date that overlaps an empty window anyway. The stint dates
are what need fixing.

## Steps to reproduce

1. Count the degenerate stints per region and year:

```sql
select
  _dbt_source_project,
  academic_year,
  count(*) as stints,
  countif(exitdate <= entrydate) as degenerate,
  countif(exitdate = entrydate) as zero_length,
  countif(exitdate < entrydate) as negative_length,
from `kipptaf_students.int_students__student_enrollment_union`
where academic_year between 2024 and 2026
group by 1, 2
order by 1, 2
```

1. Expect 185 for kippmiami AY2026 and 0 everywhere else.

1. Compare an affected stint's dates against the same student's raw Focus
   enrollment rows to see which date the model picked.

## Where

- **Code location / dbt project:** `kipptaf`, and the `focus` package
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** measured against
  `kipptaf_students.int_students__student_enrollment_union` in prod

<details>
<summary>For Claude</summary>

> Context, not instructions.

Relevant models:

[...]

Issue #5023, opened 2026-08-27
fix(dbt): ada_term_pivot weighted-year guard test fails CI on unrelated PRs (3,535 rows)


cbini:
## What's happening

`int_powerschool__ada_term_pivot__weighted_year_matches_source` fails in dbt Cloud CI with 3,535 rows, at severity error, so it blocks any pull request whose build reaches it. It is failing on unrelated branches with an identical row count, so it is a shared upstream state rather than anything in a particular diff.

I found it because it blocked a pull request that changes one Python file and no dbt at all. I have not touched the models or the test.

## Steps to reproduce

1. Open a pull request whose `state:modified+` build reaches the `int_powerschool__ada_term` models.
1. dbt Cloud fails on `int_powerschool__ada_term_pivot__weighted_year_matches_source` with "Got 3535 results, configured to fail if != 0".

## Where

- **Code location / dbt project:** `kipptaf` — `src/dbt/kipptaf/tests/int_powerschool__ada_term_pivot__weighted_year_matches_source.sql`
- **Environment:** dbt Cloud CI (staging target)
- **Run, PR, or dashboard link (if any):** run `70403222434767` on #5017, run `70403222436325` on #5018. Both report the same test, same 3,535 rows.

<details>
<summary>For Claude</summary>

### Why this is upstream, not either pull request

| Pull request | Content | Test result |
| --- | --- | --- |
| #5018 | one Python file, no dbt | fail, 3,535 rows |
| #5017 | unrelated, different author | fail, 3,535 rows |

An identical count on two unrelated branches points at shared state. #5018 in particular changes only `src/teamster/code_locations/kipppaterson/edplan/assets.py`, so it cannot put a dbt model into the selection at all.

### A theory, explicitly not verified

Two merged commits re-keyed both models onto `student_number`:

```text
fb23c9e3ba  2026-08-20  fix(kipptaf): re-key ada_term on student_number, harden Focus discriminator
d8620057c1  2026-08-20  fix(kipptaf): key ada_term_pivot on student_number so Miami does not collapse
```

The test still keys entirely on `studentid` — 5 occurrences, including the source aggregation (`group by _dbt_source_project, studentid, academic_year`) and the join (`and p.studentid = s.studentid`). If a `studentid` no longer maps one-to-one to the grain the models now carry, aggregating the source by `studentid` and comparing against a pivot keyed on `student_number` would not line up, which fits a stable non-zero count.

[...]

PR #5027, opened 2026-08-27
feat(dbt): establish Focus schedule date semantics and source Miami section status

### LINKED ISSUE #4968: feat(dbt): source Miami drop status for the course enrollment facts

### LINKED ISSUE #5002: feat(dbt): establish what a Focus schedule start_date means, and resolve the 1,839 window mismatches

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will settle what a Focus schedule date means, and
give Miami's course enrollments a real section-status flag instead of a null one.

**A Focus schedule `start_date` is the marking period's start date.** It is not
the day the student joined the section. On Miami AY2026 it equals the marking
period start on 90.7% of rows and falls before it on none. It takes only 16
distinct values across 19,363 rows, against 3 distinct marking-period starts.
Every Semester 2 row is dated to the Semester 2 start, months in the future, so
it cannot be read as an event date.

**The join that attaches course enrollments to student stints does not change.**
The cause #4970 proposed for the 1,886 unmatched rows -- a student who joins
mid-term getting dated before their own stint entry -- happens on zero rows.
1,332 of them fail for a different reason: the student's stint itself has an
exit date on or before its entry date. Nothing overlaps a window with no
duration, so no schedule date would fix those. That is a separate defect, it is
Focus-only and new, and it is now tracked on #5024. The other 554 are correct
nulls -- a section closed before the student arrived, or a section in a term
after the student left.

**Both drop flags are now non-null for Miami.** Focus has no equivalent of
PowerSchool's negative-section-id convention. It closes a schedule row by
setting an end date instead. So a row that ends before its term ends, while the
student still has an open schedule row, is a section the student left early.
576 of 19,363 rows qualify, or 2.97%. The three New Jersey regions run 0.23% to
8.55% at the same point in the year, so Miami lands inside that range.

## Reviewer Notes

- **The drop flag is an inference, not a field Focus provides.** Worth reading
  the comment in `focus_conformed` and the column description before approving.
  Both say so plainly, because PowerSchool reaches the same meaning a different
  way and the column name is shared.
- **The `coalesce` around the flag is load-bearing.** An earlier version of this
  change left it out and the flag came back null on 96.8% of rows, which would
  have reinstated the exact bug it replaces. The new rate test caught it.
- **This makes Miami visible in reports that were silently hiding it.** 40
  models filter with `not is_dropped_section`, which removes every row when the
  flag is null. Miami now reaches `int_assessments__course_enrollments` for the
  first time. Whether each report should include Miami is still #4996's call --
  this fixes the null, not the scope. Three stale comments that asserted the
  flags are null are corrected here.

### General

No Dagster changes.

      code location

  No new models. The one new test has a properties entry in
  `src/dbt/kipptaf/tests/properties.yml`.

- [x] Include (or update) an

  No new models, so no new exposures.

      summary above.

  No columns renamed or removed. Two column VALUES change for Miami only, from
  null to a real boolean. That is not contract-breaking, but it does change
  what 40 downstream models return for Miami -- see Reviewer Notes. Rollback is
  a revert; nothing is materialized irreversibly.

      See

  No external sources touched.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Design spec committed at
`docs/superpowers/specs/2026-08-27-focus-schedule-dates-and-section-status-design.md`.

### The rule, and why this shape

```sql
coalesce(
    s.end_date < s.marking_period_end_date
    and countif(s.end_date is null) over (
        partition by s.student_id, s.academic_year
    ) > 0,
    false
) as is_dropped_section
```

It reproduces PowerSchool's derivation in
`base_powerschool__course_enrollments`, which is not a bare `sectionid < 0` but
a three-branch case with two exclusions:

| PowerSchool branch | Focus equivalent | Rows |
| --- | --- | --- |
| `sectionid < 0 and dateleft = s.exitdate` -> false | end_date equals the stint exit (withdrawal sweep) | 48 pre-dedupe |
| `sectionid < 0 and dateleft = cr.max_calendardate` -> false | end_date at or after the term end | 0 |
| `sectionid < 0` -> true | end_date before the term end | 576 |

The surviving-open-row window stands in for the first exclusion: a withdrawal
closes every one of the student's rows at once, so requiring at least one still
open excludes it without joining the stint. A stint-joined derivation of the
same rule agrees on 19,358 of 19,363 rows. The 5 it over-flags hold an open
section and a section ending exactly on their stint exit. A stint join was
rejected because a student with two stints at one school-year would fan out the
schedule row and break the uniqueness test.

`is_dropped_course` needs its own CTE because BigQuery rejects a window
function inside another window function's argument.

### The zero-length question, and why it reversed

An earlier draft excluded rows where `start_date = end_date` (a section
scheduled and removed the same day, 484 rows) because no NJ row carries that
shape -- verified, 0 of 2,492 flagged NJ AY2026 rows. That was wrong. It is a
fact about PowerSchool's data, not a semantic boundary: on "left before the
expected end," a row that starts and ends the same day is the most premature
departure there is. Excluding them read 0.475%, below every NJ region.
Including them reads 2.97%, beside Newark's 2.63%.

### The audit-column lead from #4986 does not transfer

The raw `schedule` table has no `created_by_class` / `created_by_id`. It has
`modified_by`, which holds free-text staff names -- 7 distinct on AY2026, all
individual humans, no cron or system account. So the `SISUser` vs `FocusUser`
split that settled `student_report_card_grades` has no analogue here. What it
does corroborate: all flagged rows trace to 4 staff accounts, with 88%
concentrated in one, which reads as registrar action rather than a job
artifact. Account identifiers are deliberately omitted; they are in the session
notes.

### Which table to measure on

`ref("int_focus__schedule")` resolves to the kipptaf wrapper, not the
focus-package model. The wrapper deduplicates to one row per
`(student_id, course_period_id)`, preferring the row whose `end_date` is null.
Pre-dedupe the focus-package model reads 19,664 rows with 917 end dates
populated; post-dedupe it reads 19,363 with 616. Measure on
`kipptaf_focus.int_focus__schedule`.

### Verification

Built both this branch and a `main` baseline from identical upstream data, using
`--defer --favor-state` against the prod manifest. `--favor-state` is required:
without it, `int_focus__schedule` resolves to a stale empty copy in the personal
schema, the Focus branch produces zero rows, and every Miami assertion passes
vacuously.

| Region | rows, main to branch | flagged | nulls, main to branch |
| --- | --- | --- | --- |
| kippcamden | 146,652 to 146,652 | 28,067 to 28,067 | 0 to 0 |
| kippnewark | 587,567 to 587,567 | 75,781 to 75,781 | 0 to 0 |
| kipppaterson | 14,638 to 14,638 | 667 to 667 | 0 to 0 |
| kippmiami | 118,228 to 118,228 | 26,350 to 26,926 | 19,363 to 0 |

`is_dropped_course` for Miami goes 4,887 to 5,006, the 119 expected.

Tests: `dbt_utils.unique_combination_of_columns` on
`(cc_dcid, _dbt_source_project)` passes, so the new CTE does not fan out.
`test_miami_section_enrollment_orphan_rate` passes unchanged at its 15%
ceiling; the orphan rate does not move, because the join does not change.
`dim_student_section_enrollments__no_multi_stint_overlap` is unchanged -- 147
before and after, computed from one query evaluating both flag regimes over the
same rows, with Miami's documented component of 16 intact. Focus rows
contribute 0 to it either way. `ERROR=0` on
`int_students__course_enrollments dim_student_section_enrollments`. The
relationships warnings in that run are stale-dev deferral drift: the fact
tables resolve to prod while the dim rebuilds fresh.

`trunk check --force --no-fix` over all 9 changed files: no issues, exit 0.

### For a future @claude

- Do not raise the new test's 20% ceiling to silence it. A spike means one half
  of the inference stopped holding -- a bulk close of schedule rows, or a change
  that breaks the withdrawal-sweep exclusion. Investigate the cause.
- The test also fails if the flag returns to null, which is the failure mode
  this PR introduced once and fixed.
- #5024 is the one to work next if the orphan rate matters. It owns 1,332 of
  the 1,886 same-school orphans, more than #5002 and #5003 combined.

</details>

Closes #5002
Closes #4968
Refs #4970
Refs #4996
Refs #5024


cbini:
The red `claude-review` check is not a finding on this PR. The action failed
before reviewing anything:

```text
Action failed with error: Failed to install plugin 'dignified-python@dagster' (exit code: 1)
```

That is the retired plugin #5022 removes. The same failure hit
`cbini/feat/claude-focus-final-grades-kipptaf` ten minutes earlier, and the
branch carrying the fix passes, so it is repo-wide rather than specific to this
branch.

Once #5022 merges, this PR needs the review re-fired by hand. `claude-review`
only runs on `opened` and `ready_for_review`, never on `synchronize`, so a push
will not do it -- toggle draft state to trigger it.

Refs #5022

cbini:
## dbt Cloud passed. Warnings triaged, and the rank criterion verified

### Warnings

17 test warnings on the run. One is in this PR's lineage:

- `dim_student_section_enrollments__no_multi_stint_overlap` -- **147 results**,
  which is the number I measured locally on both `main` and this branch from a
  single query evaluating both flag regimes over the same rows. Unchanged.
  Focus rows contribute 0 to it either way, and Miami's documented component of
  16 is intact.

The two relationships warnings on `dim_student_section_enrollments` (33 and
1,013) are staging drift: the child table reads from `zz_stg` while the parent
rebuilds in the PR schema. The dim emits `is_current_section_enrollment` as a
boolean and does not filter on the rank, so its row set is identical -- neither
key this PR touches can move. The remaining 14 are on unrelated models (survey,
clever, appsheet, tableau gradebook, assessments scaffold).

Both new and existing Miami tests passed and so are absent from the warning
list: `test_miami_section_enrollment_dropped_rate` and
`test_miami_section_enrollment_orphan_rate`.

### The course_enrollment_rank criterion, measured honestly

#4968 asks that the rank order Miami rows on a real drop signal rather than
falling through to the date tiebreakers. It now does -- the flag is the leading
term and is non-null. But the practical effect is smaller than the issue implies,
and that is worth recording rather than glossing.

Replicated the dim's exact partition and computed the rank twice over the same
rows, once with the real flags and once with them forced null as on `main`:

| Measure | Value |
| --- | --- |
| Miami AY2026 rows | 19,363 |
| Rows marked current, main and branch | 16,805 both |
| Rows where which row is current **flipped** | **4** |
| Flagged rows still marked current, main to branch | 16 to 14 |

Only 4 rows change. The reason is that the secondary term,
`coalesce(exit_date, '9999-12-31') desc`, already sorted open rows ahead of
closed ones, so the date tiebreaker was a good proxy for the flag it now sits
behind. The ordering is correct on principle now instead of by coincidence,
which is the durable part -- it stops being a coincidence the moment Focus
starts recording schedule changes, which is exactly what #4968 predicted.

All 14 rows that remain flagged-and-current sit in partitions where **every**
row is flagged, so rank 1 is necessarily flagged. That is correct behavior, not
a residual defect.

### One follow-up committed

`c2d421aa3` corrects the `homeroom_rank` comment. It justified including
`_dbt_source_project` in the partition with a NULL-sorting hazard -- Miami's
flags were null, BigQuery sorts NULL first under `asc`, so a Miami orphan
outranked an NJ orphan on the leading term. That hazard is gone. The partition
key stays, because cross-region competition for one rank is wrong on its own.

cbini:
## Review findings addressed — `fe4e3798c1`

### 1. Cross-region `is_dropped_course` scope mismatch — declined, the premise does not hold

The finding rests on `cc_studyear` being a year, making PowerSchool's
`partition by cc_studyear, cc_course_number` course-wide with "no student key at
all." It is a student-and-year composite key, so the scopes are identical.

Measured on `kippnewark` AY2026:

| Measure | Value |
| --- | --- |
| Rows | 45,265 |
| Distinct `cc_studyear` | 6,864 |
| Distinct `students_student_number` | 6,864 |
| Students per `cc_studyear` value | 1 |

The values decompose accordingly -- `3586436` is studentid `35864` plus yearid
`36`. Across all NJ history (98,556 `cc_studyear` groups) the maximum number of
students sharing one value is 2, so the key is per-student to within a handful
of collisions, not course-wide across a district-year.

Both regions therefore window `is_dropped_section` over one student, one course,
one year. The NJ flag is not the near-degenerate statistic the finding
describes, and the ranking in `dim_student_section_enrollments` does compare
like with like.

The description was still weaker than it should be -- it asserted "the same way"
without naming the mechanism, which is what made the claim hard to check.
Reworded to name `cc_studyear` explicitly so the next reader does not have to
re-derive it.

### 2. SQL comments longer than the expression they annotate — valid, fixed

Correctly cited. `src/dbt/CLAUDE.md`'s YAML conventions say a comment longer
than the expression belongs in the properties file, and that the repo's existing
multi-paragraph comments are not a precedent to extend. Both flagged comments
broke it, and both duplicated descriptions added in the same PR.

- `is_dropped_section`: 27 comment lines on an 8-line expression, now 7.
- `cc_dateenrolled` / `cc_dateleft`: 6 lines on a 2-line projection, now 2.

What stayed is what the line does not show: the window is what excludes a
withdrawal sweep, and the coalesce is required rather than defensive. The
derivation, the measurement and the NJ reconciliation live in the descriptions,
which already carried them. The `focus_course_dropped` CTE comment is unchanged,
as you noted it is proportionate.

### 3. Null `marking_period_end_date` as a second path to false — real, but zero rows

Structurally correct: an unresolved marking period gives a null end date,
`s.end_date < s.marking_period_end_date` is then null, and the coalesce reads it
as not-dropped. Measured on `int_focus__schedule`, all years:

| Measure | Rows |
| --- | --- |
| Total | 19,363 |
| Null `marking_period_end_date` | 0 |
| Null `marking_period_id` | 0 |

#4987 closed that gap, which is why this PR was gated behind it. Not adding a
comment for a path with no rows, particularly under finding 2 -- and the
direction is the safe one, since an unresolved term reads not-dropped rather
than inventing a drop. If the left join ever starts missing, the new rate test
notices it as a rate drop.

### 4. Issue refs in `description:` — keeping them, and the divergence is worth resolving

Your read is right: the written rule says refs belong in inline SQL comments at
the derivation site, and the codebase does the opposite in
`int_students__attendance_daily.yml` and elsewhere. Following the rule for only
these columns would leave them inconsistent with their neighbours, so they stay
as-is. The doc-versus-practice split is worth settling one way or the other
repo-wide, separately from this PR.

### 5 and 6 — agreed

`avg(if(x, 1, 0)) = 1.0` is deliberate mirroring, and switching to
`logical_and()` would diverge the two regions' code shape for nothing.

On lint, `trunk check --force --no-fix` is clean again after this commit, so the
report stands on tooling you could not reach rather than on assertion.

### Unrelated: the earlier dbt Cloud failure

Run `70403222442796` failed on
`Not found: Table zz_stg_kipptaf_powerschool.stg_powerschool__schools`. That
table exists and was created at 15:34:49, inside that run's 15:33:39 to 15:37:32
window, with 82 tables in the schema stamped 15:34:49 to 15:34:52 -- the shared
staging schema was being re-cloned while CI read from it. Not this branch and
not main: PR #5020 carries the same main tip and passed at 15:35:23, right after
the re-clone finished. The run on this push is the clean re-check.

cbini:
## The dbt Cloud failures are a shared-staging race, not this branch

Both failed runs are fully explained, and none of the failures are in this PR's
lineage.

### Run 70403222443979 (`fe4e3798c1`), 15:50:39 to 15:55:04

23 failures. **All 23** are a missing or mid-recreation `zz_stg_kipptaf_*`
table. Zero involve any model or test this PR touches.

Eight distinct tables across six schemas:

```text
zz_stg_kipptaf_deanslist.int_deanslist__incidents
zz_stg_kipptaf_deanslist.int_deanslist__incidents__penalties
zz_stg_kipptaf_deanslist.int_deanslist__students__custom_fields__pivot
zz_stg_kipptaf_google_appsheet.stg_google_appsheet__kfwd_career_conversations__output
zz_stg_kipptaf_google_sheets.stg_google_sheets__kippadb__nsc_crosswalk
zz_stg_kipptaf_marts.dim_dates
zz_stg_kipptaf_powerschool.int_powerschool__gpa_cumulative
zz_stg_kipptaf_students.int_students__calendar_week
```

One error names the mechanism outright rather than leaving it to inference:

```text
Invalid time travel timestamp 1787845955932 for table
teamster-332318:zz_stg_kipptaf_students.int_students__calendar_week.
Cannot read before 1787845955948
```

The query read the table 16 milliseconds before the version it was reading came
into existence. That is only possible against a table being dropped and
recreated under a live query.

### The re-clone window

Roughly 1,000 tables across 43 `zz_stg_kipptaf_*` schemas carry a
`creation_time` between **15:52:12 and 15:53:29** -- a 77-second window sitting
inside that run. Every one is `table_type = CLONE`.

The earlier failure has the same shape. Run 70403222442796 ran 15:33:39 to
15:37:32 and died on
`zz_stg_kipptaf_powerschool.stg_powerschool__schools`, which was created at
15:34:49, with 82 tables in that schema stamped 15:34:49 to 15:34:52.

So this is not intermittent bad luck twice. Something re-clones the whole
kipptaf staging set periodically -- batches are visible at roughly 14:50, 15:34
and 15:52 -- and a CI run overlapping one reads tables that vanish mid-query.

### Why it is not this branch or main

- This exact content passed dbt Cloud twice before, on `12fed227d3` and
  `c2d421aa3f`.
- PR #5020 carries the same `main` tip and passed at 15:35:23, immediately after
  the 15:34 clone batch finished.
- Not one failing node is a model or test this PR adds or edits.

`700f73548f` is an empty commit whose only purpose is to start a run that does
not overlap a clone window. If it goes green, nothing here needs a code change.

Worth someone's attention separately: CI cannot be trusted while a full staging
re-clone can land on top of it. Either the clone should not run during PR CI
hours, or CI should read a snapshot the clone does not mutate in place.

cbini:
## Third run, same cause. I have stopped re-triggering.

`700f73548f` was an empty commit meant to start a run between clone windows. It
landed inside one anyway.

Run 70403222444232 ran 15:56:59 to 16:01:06. Table creations in
`zz_stg_kipptaf_*` during it:

| Minute | Tables created | Schemas touched |
| --- | --- | --- |
| 15:58 | 27 | 3 |
| 15:59 | 607 | 35 |
| 16:00 | 182 | 21 |

The churn is not on a cycle worth timing around -- batches are visible at 15:34,
15:52 to 15:53, and 15:58 to 16:00. Three runs, three overlaps.

### Failures fell from 23 to 2, which is the tell

Both remaining failures are the same class -- a relation that is not there:

- `rpt_powerschool__autocomm_students` --
  `Not found: zz_stg_kipptaf_powerschool.int_powerschool__district_entry_date`
- the unit test on `int_assessments__resolved_section_enrollments` -- could not
  read columns for `stg_renlearn__star` because that relation does not exist

A code defect does not go from 23 failures to 2 while the schema churn subsides
around it. Neither failing node is a model or test this PR authors. The unit
test is pre-existing; this PR reaches it only because
`int_assessments__resolved_section_enrollments` refs
`int_assessments__course_enrollments`, so it enters `state:modified+`.

### What I do not know

I have not established **what** recreates the `zz_stg_kipptaf_*` clones. Two
candidates I could not separate: concurrent PR CI runs each rebuilding their own
defer base, or a Dagster-side process. If it is the former, concurrent kipptaf
PR CI is self-defeating and this will keep hitting other people's PRs at random.

Someone who knows what owns those schemas should look. I am not pushing further
commits at this PR -- three runs is enough to establish the pattern, and more
would just burn shared CI.

For the record, this same content passed dbt Cloud twice before the `main`
merge, on `12fed227d3` and `c2d421aa3f`.

cbini:
## Staging is fixed. CI now fails on #5023 instead, which is also not this branch.

### The staging blocker is resolved

`Clone - Staging` run 70403222447038 succeeded at 16:14:00 to 16:16:05, once
Dagster had materialized the three models #5020 added
(`int_focus__report_card_grades`, `int_focus__student_enrollment`,
`int_students__final_grades`). Before that, `dbt clone` could not copy prod
models that did not exist in prod, and it left `zz_stg_*` incomplete -- which is
what produced the 23 then 2 relation-not-found failures on the earlier runs.

`673eef7b34` merged current `main` and ran in a quiet window: 819 tables cloned,
142 seconds of write silence confirmed before pushing, clone run verified
Success rather than inferred from the silence.

### What fails now

One test, `severity: error`:

```text
int_powerschool__ada_term_pivot__weighted_year_matches_source
Got 167 results, configured to fail if != 0
```

That is **#5023**, whose title is literally "guard test fails CI on unrelated
PRs", with the fix open at **#5029**.

It compares `int_powerschool__ada_term` against
`int_powerschool__ada_term_pivot`. This PR touches neither, nor anything either
one reads -- none of its 9 changed files are under `ada`, `attendance` or
`powerschool`.

The row count corroborates the mechanism #5023 describes. It tracks how long ago
the pivot last built, not anything about the PR: 3,535 on #5017 and #5018, 258 in
prod after a rebuild, 167 here. Dagster was materializing
`kipptaf/powerschool/int_powerschool__ada_term_pivot` at 16:02, shortly before
this run, which is consistent with a freshly reset and then re-accumulating drift
window.

### Where that leaves this PR

Every other check is green: Trunk, Trunk Check Runner, CodeQL, all three Analyze
jobs, and `claude-review`. The code work is complete and reviewed.

This cannot go green on its own. It needs #5029 merged, or the guard test handled
some other way. No change is pending on this branch.

### One thing worth its own issue

`Clone - Staging` breaks for **every** open PR whenever a new kipptaf model
merges, until Dagster materializes that model in prod. It is deterministic, not
flaky. #5020's own `e4f81fefcc` ("re-run CI after reseeding the staging ADA
relations from prod") shows its author hit the same wall on the way in. Today it
cost four CI runs across this PR alone.

cbini:
## Green. Run 70403222452316, zero warnings.

| Check | Result |
| --- | --- |
| dbt Cloud | pass, **0 warnings** |
| Trunk Check / Check Runner | pass |
| CodeQL + Analyze (actions, javascript, python) | pass |
| claude-review | pass |

`mergeable_state` is `blocked` only on the CODEOWNERS review for `src/dbt/`.

### Both blockers were external and are now closed out

- **#5029** deleted the racy `ada_term_pivot` guard outright rather than only
  re-materializing, so #5023's failure mode is gone.
- **#5030** covers the `Clone - Staging` abort. Staging was cloned clean from
  `33937c53c5` at 17:10:04 before this run started.

The zero-warning result is itself a retroactive check on the earlier failures.
The one previously-passing run carried 17 test warnings; this one carries none,
which is what a complete staging clone looks like versus a half-populated one.
The 17 were environment, not this branch.

For #5030's record: three separate `Clone - Staging` runs failed today from the
same cause -- `70403222444276` (#5020's merge), `70403222448763` (#5017's merge),
and the sequence that produced the 23-failure CI run. Two of them needed a manual
retry once prod caught up.

### What this PR delivers

- A Focus schedule `start_date` is the marking period's start date, established
  and documented: it equals `marking_period_start_date` on 90.7% of Miami AY2026
  rows, falls before it on zero, and takes 16 distinct values across 19,363 rows.
- The overlap join is unchanged. The mechanism #4970 proposed occurs on zero
  rows; 1,332 of the 1,886 same-school orphans come from degenerate stint
  windows, now tracked on #5024.
- `is_dropped_section` and `is_dropped_course` are non-null for Miami, at 2.97%
  against a like-for-like NJ band of 0.23% to 8.55%, documented as an inference.
- NJ is byte-identical on rows and both flags across all three regions, verified
  against a `main` baseline built from identical upstream data.

PR #5029, opened 2026-08-27
fix(kipptaf): replace racy ada_term_pivot guard with a unit test

### LINKED ISSUE #5023: fix(dbt): ada_term_pivot weighted-year guard test fails CI on unrelated PRs (3,535 rows)

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will replace
`int_powerschool__ada_term_pivot__weighted_year_matches_source` with a dbt unit
test, so the guard stops failing on unrelated pull requests and stops scanning
1.06 GiB every time it runs.

The singular test compared the pivot table against `int_powerschool__ada_term`,
which is a view. A view is re-read at query time, so the test read attendance
data that had arrived since the pivot last built. It then reported a mismatch
for every student whose current-year attendance moved in between. dbt Cloud CI
reads both sides from the shared staging schema, so the failure had nothing to
do with the diff: it blocked #5018, whose only change is a Python file.

Neither model's SQL is wrong. Running the pivot's own `PIVOT` and `coalesce`
live against the view returns 0 mismatches across 81,753 joined rows.

A unit test guards the same rule with no warehouse scan and no way to see two
instants of the source data at once. It also covers a case the old test could
not: Miami. That test joined on `studentid`, which is null for all 1,662 Miami
rows, and null never joins, so every Miami student was silently skipped.

Refs #5023

## Reviewer Notes

- **The test is verified to catch the bug, not just to pass.** I reintroduced
  the historical defect (`ada_weighted_year` coalescing the unweighted
  `ada_year_q*` columns) and the unit test failed on all 3 fixture students,
  then passed again once reverted.
- **Coverage traded away:** the singular test checked all 83k real rows, so it
  could in principle catch a data-level surprise, not just a code defect. The
  pivot keeps its `unique_combination_of_columns` test and its
  `between 0 and 1` range tests on every rate column, and the guard's own
  description says it exists for the wrong-column `coalesce`, which the unit
  test covers exactly.
- **An earlier commit on this branch took a different approach** --
  materializing `int_powerschool__ada_term` as a table so both sides of the
  comparison came from one build. It worked, and CI passed on it, but it
  measured about +29% slot-seconds per ADA cycle. It is reverted here. The
  branch name still refers to it; the diff no longer does.
- The branch also carries an empty commit that re-triggered dbt Cloud after a
  shared staging-table race, unrelated to either approach.

### General

### dbt

      models.
- [x] Include (or update) an
      consumer or column changes.
- [x] **Breaking change?** No. No model SQL, column, or contract changes. Only
      a test is swapped.
      with **`--target staging`** — no external sources touched.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Investigated with
`superpowers:systematic-debugging`; review feedback processed with
`superpowers:receiving-code-review`.

**Root-cause evidence chain.**

1. `kipptaf_powerschool.__TABLES__` reports `int_powerschool__ada_term` as
   `type = 2` (view) and `int_powerschool__ada_term_pivot` as `type = 1`
   (table), last built 2026-08-27 16:04:22Z. `int_students__attendance_daily`
   is also a view, so the whole chain below `ada_term` re-evaluated on every
   read.
1. Reproducing the singular test against production returned 258 failing rows,
   all `kippnewark` `academic_year = 2026`. Zero failures in any closed year --
   the shape of drift in actively-loading data, not a SQL defect.
1. Two failing students showed the drift directly. Student A: the view carries
   one more absence than the pivot's stored `sum_absences_year` (7 versus 6).
   Student B: identical absences, but the view's
   `sum_attendance_value_weighted_term` is 3.67 against 5 in-session days,
   meaning a tardy was recorded after the pivot built. Both are input changes,
   not mapping errors.
1. Decisive check: run the pivot's exact `PIVOT` plus `coalesce` inline against
   the same view, then compare to the view's own `ada_weighted_year`.
   `countif(pv != sv) = 0` over 81,753 joined rows.
1. #5023's own theory -- that `studentid` no longer maps 1:1 to the models'
   grain -- is disproved. In the pivot,
   `count(distinct student_number) = count(distinct studentid) = count(*)` for
   every `(_dbt_source_project, academic_year)` except `kippmiami` AY2026,
   where `studentid` is null for all 1,662 rows.

**Run-to-PR mapping.** #5023 swapped the two run ids. Run `70403222434767` is
PR #5017 and run `70403222436325` is PR #5018 (each run's
`trigger.github_pull_request_id`). The #5018 run executed
`dbt build --select state:modified+ --full-refresh` for 58 seconds and failed
on this test with "Got 3535 results", on a branch whose only change is
`src/teamster/code_locations/kipppaterson/edplan/assets.py`. Its compiled SQL
names the skewed relations:
`zz_stg_kipptaf_powerschool.int_powerschool__ada_term` (view) against
`zz_stg_kipptaf_powerschool.int_powerschool__ada_term_pivot` (table) -- the
shared staging schema CI defers to, not only production.

**Why a unit test rather than materializing `ada_term`.** The first approach on
this branch set `materialized: table` on `int_powerschool__ada_term`, which is
what `src/dbt/CLAUDE.md` -> "Test config defaults" prescribes for a check that
must refresh reliably. It was correct and CI passed. Measured cost from
`JOBS_BY_PROJECT` for 2026-08-27:

| | bytes | slot-seconds |
| --- | --- | --- |
| Pivot over the view (54 builds) | 1.06 GiB each, 57 GiB total | ~1,198 each, 17.97 slot-hours total |
| `ada_term` as a table | 1.06 GiB | 1,545 |
| Pivot reading that table | 0.03 GiB | 5.6 |

The pivot's own cost collapses, but `ada_term` picks up more than the pivot was
paying for the same scan -- it materializes all 287k rows and every column,
where the view-inlined form let BigQuery prune. Net about +352 slot-seconds per
cycle, roughly +29%, or +5.3 slot-hours per day at 54 cycles. The two
"as a table" figures are one dev-schema measurement each against 54 production
samples; directional, not means.

The unit test costs 0 bytes and 0 slot-seconds and cannot be racy, so it is the
cheaper correct answer.

**Two reasons no upstream materialization substitutes for either fix.** Worth
recording, since materializing `int_students__attendance_daily` instead is the
obvious next idea:

1. `ada_term` filters `calendardate <= current_date('America/New_York')`. The
   view's output shifts at midnight Eastern while the pivot table lags until
   its next build -- a guaranteed daily skew window regardless of what is
   materialized upstream.
1. In CI the test reads deferred SHARED staging relations built by a separate
   process, so the gap between the view side and the table side is arbitrary.
   Materializing upstream would make this failure intermittent rather than
   deterministic, which is harder to debug, not easier.

**Mutation verification.** Reintroduced the historical defect by changing the
pivot's `ada_weighted_year` to
`coalesce(ada_year_q4, ada_year_q3, ada_year_q2, ada_year_q1)`. The unit test
failed on all 3 students: student 1 `0.87 -> 0.91`, student 2 `0.734 -> 0.8`,
student 3 `0.91 -> 0.96`. Reverted the mutation and it passed again.
`git diff` on the model SQL is empty in the shipped branch.

**Test node delta**, per `src/dbt/CLAUDE.md` -> "Verifying a test-removal PR".
`dbt parse --no-partial-parse` on both sides, diffing `resource_type == 'test'`
node names: 868 to 867. Removed:
`int_powerschool__ada_term_pivot__weighted_year_matches_source`. Added: none.
Unit tests 30 to 31, added
`unit_ada_term_pivot_coalesce_reads_weighted_columns`. Nothing unintended
dropped.

**Verification.**
`uv run dbt test --select int_powerschool__ada_term_pivot --target dev --defer`
returns `PASS=12 WARN=0 ERROR=0`, covering the new unit test, the pivot's
uniqueness test, and its 10 range tests. `trunk check --force` on the changed
yml files reports only a prettier reflow, which the pre-commit `fmt` hook
applied.

**Still open, separate work.** `int_students__attendance_daily` is a view with
13 model consumers, so each one re-scans the whole attendance chain.
Materializing it would cut that repeated 1.06 GiB scan network-wide. That is a
cost change with a much wider blast radius and belongs in its own pull request.

</details>


cbini:
## Review response

Verdicts per finding. I verified every convention citation against the files rather than taking them as given.

### Finding 1 — PR body asserts the test ran on PRs with no dbt changes · **declined**

The reasoning is sound but the run record overrides it. Both cited runs exist and belong to the PRs named; #5023 just **swapped the two ids**:

| Run | Actual PR (`trigger.github_pull_request_id`) | #5023 said |
| --- | --- | --- |
| `70403222434767` | 5017 | 5018 |
| `70403222436325` | 5018 | 5017 |

Run `70403222436325` is PR #5018. It executed `dbt build --select state:modified+ --full-refresh` for 58 seconds and failed on `test.kipptaf.int_powerschool__ada_term_pivot__weighted_year_matches_source` with `Got 3535 results, configured to fail if != 0`. That branch's only change is `src/teamster/code_locations/kipppaterson/edplan/assets.py`.

So the "changed no dbt at all" phrasing is accurate as written, and the `state:modified+` inference does not hold here. Why that selection was non-empty on a Python-only diff is a separate open question — worth its own look, and it does not change this fix.

That run's compiled SQL adds a detail worth having: the two skewed relations are `zz_stg_kipptaf_powerschool.int_powerschool__ada_term` (view) and `zz_stg_kipptaf_powerschool.int_powerschool__ada_term_pivot` (table). The mismatch lives in the shared staging schema CI defers to, not only in production. PR body updated to say that.

### Finding 2 — `config: materialized: table` placement · **confirmed, no change**

`src/dbt/CLAUDE.md` line 426, "Materialization overrides go in properties yml", exists and says exactly that. Placement stands.

### Finding 3 — YAML rationale comment precedent · **confirmed, no change**

`int_powerschool__gradescaleitem_lookup.yml` does carry a comparable rationale comment above its `data_tests:` block.

### The "Test config defaults" citation · **confirmed, and stronger than the review states**

`src/dbt/CLAUDE.md` line 780 reads: "A test/asset-check re-runs only when its host model materializes, and the data-change automation condition re-materializes only TABLE models, not views. To make a check refresh regularly, anchor it to a table-materialized model." This is not merely compatible with the fix — it prescribes it.

### Two checks the review's own citations require, which I had not run

Both come from `src/dbt/CLAUDE.md` -> "View->table flips for BigQuery plan depth".

1. **Previously-reverted change?** No. `git log -S 'materialized' -- src/dbt/kipptaf/models/powerschool/intermediate/properties/int_powerschool__ada_term.yml` returns nothing, so this is not a re-attempt of the #4464 / #4587 kind.
1. **Manual post-merge step?** Not needed, read from the adapter source rather than assumed. `dbt/include/bigquery/macros/materializations/table.sql` sets `exists_not_as_table` and calls `adapter.drop_relation` before `create_table_as`, commented "Unlike table -> view, no need for `--full-refresh`: dropping a view is no big deal". The inverse warning in `src/dbt/CLAUDE.md` -> "Table->view materialization conversion needs a drop" does not apply in this direction. Confirmed empirically: the dev relation is now `type = 1` with 286,842 rows.

One caveat that section does raise and the review did not: a properties-yml-only flip does not fire `code_version_changed`, so the deploy itself will not convert the relation. It converts on the first `any_deps_updated` tick, which is roughly every 10 minutes here.

### Refresh cadence, since the flip changes the automation condition

`libraries/dbt/dagster_dbt_translator.py:90-92` moves the model from `dbt_view_automation_condition` to `dbt_table_automation_condition`, adding `any_deps_updated`. Measured on the pivot, which already sits on that signal: 25 materializations over 3h53m, mean gap 9.7 minutes, range 3.0 to 14.5 minutes, 10 to 16 seconds each. Cost is roughly neutral, since that compute was already paid inside every pivot build (a view over another view). Downstream cadence is unchanged; the pivot's consumers already fire on each pivot materialization.

### CI

All checks green on `82bea30`, dbt Cloud included, zero warnings (`warning_only=true` on run `70403222447506`). The earlier `dbt Cloud: failure` on `62696ef` was unrelated: `int_extracts__student_enrollments` hit `Not found: Table zz_stg_kipptaf_powerschool.stg_powerschool__s_nj_stu_x`, whose `creation_time` is `16:15:50.936Z` against an error at `16:15:55` — a concurrent CI run replaced that shared staging table inside the read window.

### Noted, not acted on

The review's environment note is right that it re-ran nothing data-dependent. The figures it took on trust are all reproducible from the tool calls cited in the PR body, and the 3,535 is now read straight from the #5018 run's failure record rather than reconstructed.

cbini:
## Review response, second pass

### Finding 1 — `input: ref("...")` should use single quotes · **fixed, with a correction to the basis**

Changed in `5eb43f1b3c`. The unit test still passes.

The convention claim is right for kipptaf and overstated for the repo. Counted across `src/dbt`, excluding `target/`:

| Form | Count |
| --- | --- |
| `input: ref('...')` / `input: source('...')` | 166 |
| `input: ref("...")` / `input: source("...")` | 43 |

So double quotes are not a one-off — there are 43 of them, in `src/dbt/kippmiami/models/extracts/focus/properties/` and `src/dbt/finalsite/models/api/intermediate/properties/`. What is true is the narrower statement: within the **kipptaf** project every single one is single-quoted, and mine was the only exception. That is the convention the file lives in, so I matched it. Worth flagging because "this repo's established convention" would send the next person looking for a rule that does not hold outside kipptaf.

### Independent arithmetic check · noted

The hand-recomputation of all three fixture students against the model SQL, and of the mutation deltas (`0.87` to `0.91`, `0.734` to `0.8`, `0.91` to `0.96`), matches what the local run produced. Useful given the review had no warehouse access.

### CI, and why the previous run showed 20 warnings

Zero warnings now, on run `70403222450025`. The prior run on this branch reported 20, including two `Breaking change to contracted, unversioned model` entries naming columns this diff never touches (`college_counselor` on `rpt_tableau__kfwd_dashboard`, `college_counselor_id` on `stg_kippadb__contact`).

Cause was main moving, not this branch. PR #5017 merged `feat(dbt): add college counselor to kippadb contact and kfwd dashboard`, so against the updated deferred state an older branch reads as having REMOVED those columns, and the source-hash mismatch fanned `state:modified+` out across marts, extracts, and tableau, which produced the other 18 warnings. Merging `origin/main` (`bbca65f107`) cleared all 20.

Noting it here because the same shape will recur on any branch that sits through a merge to a contracted model: check `git log <merge-base>..origin/main` before treating a breaking-change warning as your own.

PR #5034, opened 2026-08-27
feat(dagster): ingest the Focus student_gpa_calculated table

### LINKED ISSUE #5021: feat(dbt): ingest the Focus student_gpa_calculated table and branch fct_grades_gpa for Miami
## What and why

Miami has no rows in `fct_grades_gpa`, for any year. #5020 gave Miami its term
grades in `fct_grades_term` and left GPA out on purpose. This issue closes that
gap.

Focus computes GPA and class rank itself, in `student_gpa_calculated`. We do not
ingest that table. Sourcing the vendor's own numbers is the right way to fill the
gap. Recomputing them through a branched PowerSchool GPA chain is not, because
the inputs that chain needs mostly do not exist on the Focus side.

Class rank is a bonus. The PowerSchool GPA chain does not produce it at all.

## Proposed approach

Three steps, in order. Each has to land in prod before the next one starts.

1. **Ingest.** Add `student_gpa_calculated` to the Focus dlt config, then add a
   `stg_focus__student_gpa_calculated` staging model and an `int_focus__gpa`
   model to the focus package.
1. **Wait.** Let Dagster materialize both in prod. Not merged — materialized.
   This is #4290, and it does not self-heal.
1. **Wrap and branch.** Declare the source and wrapper in kipptaf, then give
   `fct_grades_gpa` a Focus branch the way #5020 gave one to `fct_grades_term`.

Look at the table's real contents before designing step 3. Everything we know
about it today comes from vendor documentation, not from data.

<details>
<summary>For Claude</summary>

Context, not instructions.

### Verified state, 2026-08-27

- `student_gpa_calculated` is absent from
  `src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml` and absent
  from `dagster_kippmiami_dlt_focus`.
- `kipptaf_marts.fct_grades_gpa` holds Camden, Newark and Paterson only, for
  AY2024 through AY2026. Miami is absent for every year, not just AY2026.
- The Miami archive still holds GPA: `kippmiami_powerschool.int_powerschool__gpa_term`
  16,512 rows, `int_powerschool__gpa_cumulative` 3,053,
  `int_powerschool__gpa_cumulative_year` 5,975, `stg_powerschool__storedgrades`
  239,545. Unlike `base_powerschool__final_grades`, which is 0 rows, the GPA
  year-scope is load-bearing — that history has to stay readable.

### Why recomputing was ruled out

- The chain runs ~1,000 lines across `int_powerschool__gpa_term`,
  `int_powerschool__gpa_cumulative` and `int_powerschool__gpa_cumulative_year`,
  and reads `stg_powerschool__storedgrades` and
- Focus has no `storedgrades` analog. The nearest equivalent,
  `int_focus__report_card_grades where course_history = 'Y'`, is 0 rows for
  AY2026 — so cumulative and prior-year GPA have no input at all for the
  cutover year.
- Only 145 of 1,593 Miami AY2026 report-card rows carry `affects_gpa = 'Y'`.

### Vendor notes

`marking_period_id = -1` marks the course-history GPA. Source: Focus SIS Level 1
Certification, Day 2, Course History and Report Card Grades. Recorded on #4986.

### Constraints

- Year-scope the union off `int_students__sis_cutover`, the boundary #5020 and
  the five `int_students__` attendance conforms all share. Do not exclude Miami
  wholesale.
- `fct_grades_gpa` joins `int_students__student_enrollment_union` on
  `(studentid, schoolid, yearid)`. `studentid` and `yearid` are null on all
  10,075 Miami rows there, so that join cannot match a Focus row. #5020 solved
  the same problem in `fct_grades_term` by moving to
  `(student_number, academic_year)` — the two key pairs are exactly 1:1 inside
  every NJ district, so the swap is row-preserving. Expect to do the same here,
  and measure it the same way rather than assuming it.
- Branch at the altitude the vendor derives the model. Do not recompute at
  network altitude.

### Acceptance criteria

- NJ parity: all 3 NJ regions row-identical to prod, on `count(*)` plus a
  distinct count of `grades_gpa_key`.
- Miami AY2026 rows present in `fct_grades_gpa`.
- Miami historical GPA still readable from the archive for AY2020 through AY2025.
  materialization tag against the upstream's current `data_version` before
  trusting a green build.

Refs #5020
Refs #4986
Refs #4999
Refs #4290

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will load the Focus `student_gpa_calculated` table
into `dagster_kippmiami_dlt_focus`.

Focus computes GPA and class rank itself and stores both in that table. We never
ingested it. That is why `fct_grades_gpa` holds no Miami rows, in any year.
Reading the vendor's own numbers is the right way to fill that gap, because the
PowerSchool GPA chain needs inputs the Focus side mostly does not have.

This is the ingestion half of #5021. The dbt half comes in a second pull request
and cannot start until this table exists in prod with rows in it.

Adding the table to the config was 2 lines. It then failed to load, so this pull
request also adds a per-table numeric widening adapter. See Reviewer Notes.

The table is already loaded and verified in prod — a branch-deployment run writes
to the prod dataset, because dlt has no branch-deployment redirect.

## Reviewer Notes

**The 2-line config change was not enough, and the failure is the interesting
part.** `weighted_gpa` is an unbounded Postgres `numeric`. dlt renders that as
`decimal128(38, 9)`, and pyarrow refuses to rescale a value that needs more
decimal places, so the extract died before writing anything. Measured after the
load succeeded: the real values carry up to **20** decimal places. Focus stores
an unrounded division result.

**Why the widening is opt-in per table rather than source-wide.** Illuminate
applies `unbounded_numeric_adapter` to its whole source. Focus cannot: the
widened type maps to BigQuery BIGNUMERIC, and 45 already-loaded Focus tables
carry 200 NUMERIC columns that the `replace` disposition cannot retype in place.
Making it source-wide would require recreating all 45. So a
`widen_unbounded_numeric` key in `focus.yaml` routes exactly one table.

**`(76, 38)` is a ceiling, not a tuning choice.** dlt maps precision above 38 to
`decimal256`, and the BigQuery destination declares `wei_precision=(76, 38)`.
Nothing wider exists. I went straight to it after `(38, 18)` also overflowed,
rather than guessing a third intermediate scale.

**The adapter returns `Float` untouched, and Illuminate's does not.** `Float`
subclasses `Numeric` and also reflects `precision=None`, so without that guard
every `double precision` column in an opted-in table would land as BIGNUMERIC. I
did not change Illuminate's copy; no float column reaches its path today.

**For the dbt half:** `gpa`, `weighted_gpa`, `cumulative_gpa` and the
`custom_*_gpa` columns land as **BIGNUMERIC**, not NUMERIC.
`libraries/dlt/CLAUDE.md` says to `cast(col as numeric)` at the staging layer to
keep contracts on NUMERIC.

**Not fixed here:** those same 45 tables will hit this identical failure the
moment any of their unbounded `numeric` columns receives a high-scale value —
the same latent shape as the `interval` bug in #4676. That needs a source-wide
adapter plus a recreate migration, which is its own issue.

### General

      feedback; dismiss false positives with a brief reply explaining why. —
      per-finding verdicts posted as a comment; its numeric note was the actual
      blocker

### Dagster

- [ ] Run `uv run dagster definitions validate` for any modified code location —
      cannot run locally, see below
- [ ] Run `uv run pytest tests/test_dagster_definitions.py` for any modified code
      location — same blocker, fails identically on `main`
      with the correct asset key format and IO manager — adds a table and a
      config key to an existing integration

`dagster definitions validate -m teamster.code_locations.kippmiami.definitions`
fails in the codespace with `ConfigFieldMissingException: Missing 1 field(s) ...
drivername`, because the Focus database credentials are not present here. It
fails the same way on unmodified `main`, so this is a codespace limitation, not a
regression. Covered instead by the Dagster Cloud deploy check and by an actual
prod load.

### Docs

      catalog: `uv run scripts/gen-automations-doc.py` — Focus is already absent
      from `docs/reference/automations.md` entirely, and regenerating it needs
      the same credentials the codespace lacks. Tracked separately, not fixed
      here.

## CI checks

- [x] **Trunk** — `trunk check --force` clean on all changed files
- [x] **dbt Cloud** — passed; no dbt changes
- [x] **Dagster Cloud** — all 5 code locations deployed

<details>
<summary>For Claude</summary>

### Verification trail

3 branch-deployment runs of `kippmiami/dlt/focus/student_gpa_calculated`, each
writing to the prod dataset:

| Run | Commit | Result |
| --- | --- | --- |
| `ad883ce0` | `739ff90a` | failed — `inferred_arrow_type=decimal128(38, 9)` |
| `811a7381` | `f974d576` | failed — `inferred_arrow_type=decimal128(38, 18)` |
| `f18848eb` | `a00c7387` | success, 1 materialization |

The second failure is what proves the per-table opt-in routes correctly: the
reported arrow type moved from `(38, 9)` to `(38, 18)` while the other 78 tables
kept the old adapter. The first failure is what proves `cursor_column:
updated_at` exists — the probe runs `SELECT COUNT(*), MAX(updated_at)` before
extraction, and it succeeded.

### Loaded shape

441 rows, 47 columns, `syear = 2026` only.

- Grain is `(student_id, school_id)`: 441 of 441 distinct, and `id` likewise. 122
  of 319 students carry a row at 2 schools.
- **Every row is `marking_period_id = -1`.** The course-history GPA is the whole
  table, not a subset to filter to. `mp` is null throughout.
- `gpa` is null on 167 of 441 rows (38%), concentrated in school 15 (161 of 170)
  and school 68 (6 of 6).
- `class_rank` is populated on **3 rows**, all at school 58.
- Max decimal scale: `weighted_gpa` 20, `gpa` 3, `cumulative_credits` 1. Only
  school 58 carries the high-scale values.

### Two premises on #5021 that the data contradicts

**"Class rank is a bonus."** 3 populated rows out of 441. It is not usable as a
fact column yet.

**`marking_period_id = -1` reads as a filter in the issue body.** It is not a
discriminator here — no other value exists, so there is no term-level GPA in this
table at all. That has consequences for what `fct_grades_gpa`'s term-grained
columns can be filled with on the Focus side.

Both are for #5021's dbt half to resolve; neither changes this pull request.

</details>

Refs #5021

### DIFF
--- a/src/teamster/code_locations/kippmiami/dlt/focus/assets.py
+++ b/src/teamster/code_locations/kippmiami/dlt/focus/assets.py
@@ -14,12 +14,15 @@ sql_database_credentials = resolve_configuration(
     ConnectionStringCredentials(), sections=("FOCUS_DB",)
 )
 
+config_assets = yaml.safe_load(config_file.read_text())["assets"]
+
 tables = [
     # a["cursor_column"], not .get(): a new table added without a declared
     # cursor must fail loudly at module load, not silently become count-only.
     ProbeTable(name=a["table_name"], cursor_column=a["cursor_column"])
-    for a in yaml.safe_load(config_file.read_text())["assets"]
+    for a in config_assets
 ]
+
 """Module-level so sensors.py can import it instead of re-parsing the YAML.
 
 Keeps the credential resolution and config parse to one copy per code-location
@@ -27,10 +30,22 @@ import — both `assets.py` and `sensors.py` load unconditionally via
 `dlt/focus/__init__.py`.
 """
 
+widen_numeric_tables = frozenset(
+    a["table_name"] for a in config_assets if a.get("widen_unbounded_numeric")
+)
+"""Tables whose unbounded Postgres `numeric` columns need an explicit scale.
+
+`.get()`, not `[...]`: this key is absent on all but the tables that need it.
+Opt-in rather than source-wide because widening retypes the column to BigQuery
+BIGNUMERIC, and 45 already-loaded Focus tables carry 200 NUMERIC columns that
+`replace` cannot retype in place.
+"""
+
 assets = [
     build_focus_dlt_assets(
         sql_database_credentials=sql_database_credentials,
         code_location=CODE_LOCATION,
         tables=tables,
+        widen_numeric_tables=widen_numeric_tables,
     )
 ]
--- a/src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml
+++ b/src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml
@@ -79,6 +79,8 @@ assets:
   # Grades
   - table_name: student_report_card_grades
     cursor_column: updated_at
+  - table_name: student_gpa_calculated
+    cursor_column: updated_at
   - table_name: report_card_grades
     cursor_column: updated_at
   - table_name: report_card_grade_scales
--- a/src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml
+++ b/src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml
@@ -81,6 +81,9 @@ assets:
     cursor_column: updated_at
   - table_name: student_gpa_calculated
     cursor_column: updated_at
+    # weighted_gpa is an unbounded Postgres numeric holding more than 9 decimal
+    # places, which the default decimal128(38, 9) cannot represent.
+    widen_unbounded_numeric: true
   - table_name: report_card_grades
     cursor_column: updated_at
   - table_name: report_card_grade_scales
--- a/src/teamster/libraries/dlt/focus/assets.py
+++ b/src/teamster/libraries/dlt/focus/assets.py
@@ -141,8 +141,16 @@ def widen_unbounded_numeric_adapter(col_type: TypeEngine) -> TypeEngine:
     more than 9 decimal places, and the extract dies with
     ``Rescaling Decimal value would cause data loss`` —
     ``student_gpa_calculated.weighted_gpa`` is the first Focus column to hit it.
+    It also overflowed ``decimal128(38, 18)``, so it carries more than 18
+    decimal places: Focus stores an unrounded division result.
 
-    ``Numeric(38, 18)`` maps to BigQuery BIGNUMERIC, not NUMERIC, so a dbt
+    ``(76, 38)`` is the destination ceiling, not a tuning choice. dlt maps
+    precision above 38 to ``decimal256``, and BigQuery declares exactly
+    ``wei_precision=(76, 38)`` — BIGNUMERIC. Nothing wider exists to fall back
+    to. Postgres ``numeric`` scale is unbounded in principle, so a future column
+    could still overflow this; rounding in the query would be the only fix left.
+
+    ``Numeric(76, 38)`` maps to BigQuery BIGNUMERIC, not NUMERIC, so a dbt
     staging model over an opted-in table should ``cast(col as numeric)`` to keep
     contracts on NUMERIC. That retype is also why this is opt-in per table
     rather than applied to the whole source: 200 NUMERIC columns across 45
@@ -159,7 +167,10 @@ def widen_unbounded_numeric_adapter(col_type: TypeEngine) -> TypeEngine:
         return col_type
 
     if isinstance(col_type, Numeric) and col_type.precision is None:
-        return Numeric(precision=38, scale=18)
+        # ponytail: destination maximum, chosen because Focus is unreachable
+        # from CI so the real scale cannot be measured. Narrow it once a loaded
+        # value can be inspected in BigQuery.
+        return Numeric(precision=76, scale=38)
 
     return col_type
 
--- a/src/teamster/libraries/dlt/focus/assets.py
+++ b/src/teamster/libraries/dlt/focus/assets.py
@@ -1,4 +1,4 @@
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 from typing import Any, get_args
 
 import dlt
@@ -15,7 +15,7 @@ from dlt.extract.items import DataItemWithMeta
 from dlt.extract.resource import DltResource
 from dlt.sources.sql_database import remove_nullability_adapter
 from dlt.sources.sql_database.helpers import table_rows
-from sqlalchemy import BigInteger
+from sqlalchemy import BigInteger, Float, Numeric
 from sqlalchemy.sql.sqltypes import _AbstractInterval
 from sqlalchemy.types import TypeEngine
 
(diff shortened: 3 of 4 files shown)


cbini:
## Review verdicts, and an empirical answer to finding 1

Verified every cited fact before responding. All 4 hold — `focus/assets.py:165,168`
passes only the 2 adapters, `illuminate/assets.py:47` has
`unbounded_numeric_adapter` and Focus does not, `libraries/dlt/CLAUDE.md:142-147`
documents the `decimal128(38, 9)` truncation, and `sensors.py:104-109` is an
unguarded dict comprehension.

### Finding 1 — the unbounded `numeric` note. Correct, and it was the blocker.

Rated "low severity, informational, confidence: low." It is the reason the table
would not load at all. The suggested check is exactly the one that failed:

```text
PyToArrowConversionException: Conversion to arrow failed for field `weighted_gpa`
with dlt hint `data_type=decimal` and `inferred_arrow_type=decimal128(38, 9)`
Insufficient decimal precision  Rescaling Decimal value would cause data loss
```

Focus was unreachable from the codespace, so I answered it by loading the table
in the branch deployment and measuring the result. **Max decimal scale on
`weighted_gpa` is 20.** Focus stores an unrounded division result, so 9 was never
going to hold it, and neither did 18.

Fixed rather than dismissed, in `a00c7387`:

- The widening is **opt-in per table**, via a `widen_unbounded_numeric` key in
  `focus.yaml`, not source-wide as Illuminate does it. Source-wide would retype
  200 NUMERIC columns across 45 already-loaded Focus tables to BIGNUMERIC, and
  the `replace` disposition cannot retype a column in place.
- The target is `(76, 38)`, which is a ceiling rather than a tuning choice: dlt
  maps precision above 38 to `decimal256`, and the BigQuery destination declares
  `wei_precision=(76, 38)`.
- The adapter returns `Float` untouched. `Float` subclasses `Numeric` and also
  reflects `precision=None`, so without that guard every `double precision`
  column in an opted-in table would land as BIGNUMERIC. Illuminate's copy omits
  the guard; I left it alone, since no float column reaches that path there
  today.

The systemic half of the note stands and is **not** fixed here: those 45 tables
will hit this same failure the moment one of their unbounded `numeric` columns
receives a high-scale value. Same latent shape as the `interval` bug in #4676.
That needs a source-wide adapter plus a recreate migration, so it belongs in its
own issue rather than in a GPA ingestion change.

### Findings 2, 3, 4 — confirmations. No action, and all still true.

Finding 4 deserves a note: it verified the blast radius of an unverified cursor
column, and that risk resolved cleanly. The first failed run proves
`cursor_column: updated_at` exists, because the probe issues
`SELECT COUNT(*), MAX(updated_at)` before extraction and it succeeded — the
failure was downstream, in arrow conversion.

### Current state

`f18848eb` loaded the table successfully: 441 rows, 47 columns. The PR body now
carries the full shape and 2 findings in the loaded data that contradict
premises on #5021.

PR #5039, opened 2026-08-27
docs(dbt): sweep inline SQL comments against the properties-file rule

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." move model-level explanations out of
inline SQL comments and into the properties files that already own them, and
rewrite the comments that stay so each one reads in a single pass.

A sweep of every `.sql` file in `src/dbt/` found 295 multi-line comment blocks
across 122 model files. Three groups came out of it:

1. 12 blocks explained what or why a model computes, with nothing about the
   line they sat above. Most of that text was already word-for-word in the
   model or column `description:`. Those blocks are gone, and the few facts
   that were missing moved into the description.
2. 46 blocks did explain their own line, but ran 6 to 14 lines. Those are
   rewritten: one idea per sentence, active voice, identifiers in backticks.
3. 30 copies of the same 2-line `grain projection` annotation are now the
   1-line `grain projection, not dup-masking`. The annotation still tells a
   reviewer the `DISTINCT` is deliberate. Only the restated convention text is
   gone, and the SQL guide now asks for the short form.

Two stale comments went with them: one that told CI to rebuild a model for a
pull request that merged long ago, and one that pointed at a plan document.

## Reviewer Notes

- `int_finalsite__contact_id_attributes.yml` said the model was enabled for
  Miami only. The union has read all 4 regions for a while, so the description
  was wrong, not just thin. Fixed here.
- `int_students__course_enrollments` and `int_students__course_sections`
  carried the same 10-line block, character for character. Duplicated prose is
  the clearest sign it belongs in metadata, so it moved.
- No SQL logic changed anywhere in this pull request. Every edit is a comment
  or a description.

### General

### dbt

      models; 4 existing properties files gained text.
- [x] Include (or update) an exposure — no model outputs changed.
- [x] **Breaking change?** No. No column or model was renamed or removed.
- [x] If adding or modifying an external source — none touched.

## CI checks

- [x] **Trunk** — `trunk check --force` over the full branch diff: 49 files
      checked, 2 pre-existing issues, no new issues.
- [x] **dbt Cloud** — run 70403222459612 succeeded with 0 warnings.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The sweep classified comment blocks by whether
they explain the line they sit on or the model as a whole, per the rule at
`src/dbt/CLAUDE.md`.

Verification: `dbt deps` plus `dbt parse` on `kipptaf`, `kippmiami` and
`kippnewark` — the projects that consume the changed `focus` and `finalsite`
packages — all parse clean. `trunk check --force` over the 32 changed files
reports 1 pre-existing issue and no new ones.

Deliberately left alone: 144 single-line comments, which already read as
line-local notes; the 25 long blocks in `tests/` and 8 in `analyses/`, since
`analyses/` has no properties surface and singular tests carry only 36
`description:` entries repo-wide; and 18 `grain projection` comments that name
their actual partition key, which are more informative than the generic form
and stay as they are.

The `DISTINCT` rule in `src/dbt/CLAUDE.md` previously mandated the exact 2-line
wording. It now mandates `grain projection, not dup-masking` and says why the
annotation is required, so a future reviewer does not read the shorter comment
as an omission.
</details>

### DIFF
--- a/src/dbt/CLAUDE.md
+++ b/src/dbt/CLAUDE.md
@@ -1088,13 +1088,15 @@ validation/profiling goes through BigQuery MCP, not `dbt show`.
 - **`DISTINCT` — grain projection only, never dup-masking.** Use `DISTINCT` for
   a `GROUP BY` with no aggregation, and for pure grain projection (every
   projected column is functionally determined by the partition key, so
-  byte-identical tuples coalesce) — annotate the latter with
-  `grain projection: every selected column is functionally determined / by the partition key; not a mask for upstream duplicates`.
-  NEVER `SELECT DISTINCT` or `qualify row_number() over (...) = 1` to mask
-  upstream duplicates, and never `DISTINCT` when a projected column varies
-  within the partition (`min()`, `first_value()`) — use
-  `dbt_utils.deduplicate()` (see _Row picking, dedup & surrogate keys_) with a
-  `-- TODO:` naming the upstream fix.
+  byte-identical tuples coalesce). Annotate the latter with the one-line
+  `grain projection, not dup-masking` — the annotation is what tells a reviewer
+  the `DISTINCT` is deliberate rather than a fan-out mask or a wrong-grain
+  source, so it is required. Name the partition key on the same comment when the
+  `SELECT` list does not make it obvious. NEVER `SELECT DISTINCT` or
+  `qualify row_number() over (...) = 1` to mask upstream duplicates, and never
+  `DISTINCT` when a projected column varies within the partition (`min()`,
+  `first_value()`) — use `dbt_utils.deduplicate()` (see _Row picking, dedup &
+  surrogate keys_) with a `-- TODO:` naming the upstream fix.
 - **No one-sided calculations in join predicates.** Any expression computable
   from a single table's columns is precomputed as a named column upstream — `ON`
   matches plain columns. Expressions that inherently combine columns from both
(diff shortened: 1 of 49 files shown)


cbini:
Per-finding verdicts on the `claude-review` pass.

**1. Dropped clause in `int_students__course_sections.yml` — confirmed, fixed in
a66a21c74.** The edit anchor swallowed "Miami AY2020 through AY2025." My error.
The description now reads "...so the frozen archive keeps serving Miami AY2020
through AY2025. The boundary is derived from the Focus schedule rather than
hardcoded...", matching the sibling `int_students__course_enrollments.yml`.

**2. `state:modified` fan-out across downstream intermediates — accurate, and
the run is clean.** dbt Cloud CI run 70403222459612 finished `success`, and
`get_job_run_error(warning_only=true)` returns `has_warnings: false` — 0 test
warnings, 0 freshness warnings, 0 log warnings. So the wide rebuild happened and
surfaced no noise. The self-review checkbox was stale at the time of the review;
it is checked now.

**3. Trunk-check scope was partial — correct, and now re-run wider.** The "32
files" figure covered only the second commit's diff, so the first commit's 17
other files, `src/dbt/CLAUDE.md` included, were outside that run.
`trunk check --force` over the full `git diff origin/main...HEAD --name-only`
list: **49 files checked, 2 pre-existing issues, no new issues.**

**4-7. No findings / spot-verified claims — no action.** Agreed on all of them,
including that the removed 10-line block was byte-identical between
`int_students__course_enrollments.sql` and `int_students__course_sections.sql`
before removal.

PR #5040, opened 2026-08-27
docs(dbt): record the ratified Miami scope decision in all 40 model descriptions

### LINKED ISSUE #4996: fix(dbt): decide Miami scope per report across the 40 models filtering on the drop flags

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." write down, for all 40 reports #4996
flagged, whether Miami belongs in them and why.

#4996 found 40 kipptaf models that filter on the Focus drop flags. Its point was
not really about the filter. It was that a report which quietly leaves a whole
region out is broken even when the number it prints is correct, because nobody
reading it can tell. So each of the 40 needed a decision, and the decision needed
to be written somewhere a reader would find it.

That is what this does. 23 reports include Miami. 17 do not. Every one now says
which, and why, in its model description. The full table is ratified on #4996.

Nothing else changes. No SQL, no columns, no tests.

Three of the descriptions say something uncomfortable on purpose: the report is
meant to include Miami, and it still shows nothing. An empty result nobody
explained is the same problem as an exclusion nobody explained, so each says what
is actually blocking it:

- `rpt_tableau__dibels_dashboard` — blocked because Focus does not record whether
  a student is in a self-contained classroom. We left that blank rather than
  guess, so Miami stays out for now.
- `rpt_tableau__state_assessments_dashboard` — it carries Florida FAST results and
  a Miami-only Civics rule, so Miami is clearly intended, but it joins test scores
  to students using a New Jersey ID. The Florida side never connects. That is a
  separate bug.
- `int_extracts__course_enrollments_by_term` — blocked by a Focus section-size
  column that is empty for Miami. Fixing it means changing the Focus package, so
  it was deliberately left out of #5038.

29 of the 40 models had no description at all. Those get one short line saying
what the model is, then the scope sentence, so the new text reads like
documentation instead of a lone footnote.

## Reviewer Notes

- **Read the #4996 table first**, not this diff. The diff is 40 near-identical
  paragraphs; the judgement all lives in the ratified table.
- **The three honest-failure descriptions** listed above are the ones worth a
  second look. If you disagree that a report should admit it is empty, say so.
- **No drop-flag `coalesce` here, and none in #5038 either.** #4996 asked for one
  and it turned out to be a no-op. Details in the fold-out.

### General

No Dagster changes.

      models. All 40 properties files already existed.
- [x] Include (or update) an
      consumer changes.
- [x] **Breaking change?** No. Descriptions only.
      with **`--target staging`** — no external source changes.

## CI checks

- [x] **Trunk** — passes. `trunk check --force` clean on all 40 files.
- [ ] **dbt Cloud** — passes.

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. **The ratified decision table is
#4996#issuecomment-5443235788, with the ratification and corrections in
#4996#issuecomment-5443390602.** Read both before reviewing this diff; the diff
carries none of the reasoning.

**Final tally: 23 INCLUDE, 17 EXCLUDE, reconciled by name against the 40 files
the issue's grep returns.** The first table comment listed 20 INCLUDE and 17
EXCLUDE and left `rpt_tableau__student_info_audit` out of both sections through a
transcription error; the ratification comment corrects it to INCLUDE (network-wide
data quality audit, Miami 12,294 rows in prod). The three files marked UNSURE in
the first comment — `rpt_tableau__gradebook_gpa`,
`int_extracts__course_enrollments_by_term`, and the `rpt_gsheets__csgf_hs_*` pair
— were all ratified INCLUDE.

**Why there is no drop-flag `coalesce` in this PR or in #5038.** #4996 asked for
`not coalesce(is_dropped_section, false)` on every include-Miami file. Measured
against prod 2026-08-27: both `is_dropped_section` and `is_dropped_course` are
non-null on all 19,363 Miami AY2026 rows of
`base_powerschool__course_enrollments`, and all 40 files read both flags from that
relation and nowhere else — checked per file, alias by alias. `not
is_dropped_section` already passes 18,787 Miami rows and `not is_dropped_course`
passes 19,244. The coalesce changes zero rows in zero files. It is doubly dead in
18 of the 40, where the filter sits only on a LEFT JOIN and cannot remove a row
at all — including all three Miami anchors the issue named. #4996's own Sequencing
section predicted exactly this if #4968 landed first, which it did.

**The live trap, which #5038 fixes.** Four flags on
`int_students__student_enrollments` were null on 100% of Miami's 1,760 AY2026 rows
and non-null on every NJ row: `is_enrolled_y1`, `is_enrolled_recent`,
`is_out_of_district`, `is_self_contained` (plus `is_504`, whose Miami null is
deliberate and already documented). `rpt_tableau__miami_fast` held **zero rows of
any kind** because its where clause requires `is_enrolled_y1`. #5038 derives the
first two from the roster's dates using PowerSchool's own expressions, sets
`is_out_of_district` false on the grounds that the Focus branch already assigns
every Miami row a real `reporting_schoolid` and an ES/MS/HS `school_level` rather
than the out-of-district values, and leaves `is_self_contained` null.

**Why `is_self_contained` stays null.** Focus's nearest field is
`ese_fefp_code_label`, whose Miami values are 3,890 null, 104 `4-8 Basic ESE
Services [112]`, 57 `PK-3 Basic ESE Services [111]`, and 1 `ESE Support Level 4
[254]`. Support Level 4 is a Florida funding-matrix level, not a placement label,
so mapping it onto PowerSchool's named SPED program is a SPED-semantics call for
Ops rather than a code change. Requester chose "leave null, document it", which is
why `rpt_tableau__dibels_dashboard`'s description now states the exclusion
outright instead of the dashboard silently returning nothing.

**Row counts in the descriptions** are prod as of 2026-08-27, AY2026 unless the
sentence says otherwise. They are a snapshot for the reader's orientation, not an
invariant — they will drift, and that is fine. The scope verdict is the durable
part.

**Mechanics.** Applied by a script with a per-file exact-anchor assertion (abort
unless the `- name: <model>` block matched exactly once), so no file was edited
blind. 11 of the new descriptions needed `trunk fmt` for prettier's YAML block
scalar wrapping; formatting is committed, and `trunk check --force` is clean on
all 40 afterwards. An earlier pass wrapped `PowerSchool-only` across a line break,
which a folded YAML scalar would have rendered as "PowerSchool- only"; fixed by
disabling hyphen breaks before the committed pass. `dbt parse` succeeds on the
project, which is the real YAML validation.

**Stacking.** This PR and #5038 both branch from `main` and share no files, so
they can merge in either order. The descriptions here reference #5038 in the past
tense ("was null for Miami until #5038"), which reads correctly once both land.

Refs #4996
Refs #4973
Refs #4968

</details>

### DIFF
--- a/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__course_enrollments.yml
+++ b/src/dbt/kipptaf/models/assessments/intermediate/properties/int_assessments__course_enrollments.yml
@@ -17,6 +17,10 @@ models:
       should be network-wide or NJ-scoped remains #4996's decision; the
       inclusion is a consequence of fixing the flags, not a scope judgment made
       here.
+
+      Network-wide, Miami included. This model already recorded that #4996 owned
+      the decision; the decision is now made and it is INCLUDE. Miami held
+      61,910 rows in prod as of 2026-08-27.
     config:
       materialized: table
       meta:
--- a/src/dbt/kipptaf/models/extracts/deanslist/properties/rpt_deanslist__final_grades.yml
+++ b/src/dbt/kipptaf/models/extracts/deanslist/properties/rpt_deanslist__final_grades.yml
@@ -1,5 +1,10 @@
 models:
   - name: rpt_deanslist__final_grades
+    description: >-
+      Pushes current-year final grades to DeansList. Network-wide, Miami
+      included: Miami is a DeansList region, with 11,585 incidents in
+      `int_deanslist__incidents` as of 2026-08-27, so it belongs in the grade
+      feed too. Ratified on #4996.
     columns:
       - name: student_number
         data_type: int64
--- a/src/dbt/kipptaf/models/extracts/google/sheets/intermediate/properties/int_gsheets__dds_import.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/intermediate/properties/int_gsheets__dds_import.yml
@@ -1,9 +1,9 @@
 models:
   - name: int_gsheets__dds_import
     description: >-
-      **Excludes Miami by design.** This model is disabled and never compiled,
-      so its Miami scope is moot; if it is ever re-enabled, decide the scope
-      then. Ratified on #4996.
+      **Not evaluated for Miami.** This model is disabled and never compiled, so
+      it has no Miami scope to decide; decide one if it is ever re-enabled.
+      Recorded on #4996.
     config:
       enabled: false
     columns:
--- a/src/dbt/kipptaf/models/extracts/google/sheets/intermediate/properties/int_gsheets__dds_import.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/intermediate/properties/int_gsheets__dds_import.yml
@@ -1,5 +1,9 @@
 models:
   - name: int_gsheets__dds_import
+    description: >-
+      **Excludes Miami by design.** This model is disabled and never compiled,
+      so its Miami scope is moot; if it is ever re-enabled, decide the scope
+      then. Ratified on #4996.
     config:
       enabled: false
     columns:
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__assessment_roster.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__assessment_roster.yml
@@ -10,6 +10,9 @@ models:
       pseudonymous student_token (a stable hash of student_number), so the
       extract is de-identified at the source and shared under the research
       data-sharing agreement.
+
+      Network-wide, Miami included. Miami held 3,125 AY2026 rows in prod as of
+      2026-08-27. Ratified on #4996.
     data_tests:
       - dbt_utils.unique_combination_of_columns:
           arguments:
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__college_assessments_long.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__college_assessments_long.yml
@@ -1,5 +1,10 @@
 models:
   - name: rpt_gsheets__college_assessments_long
+    description: >-
+      **Excludes Miami by design.** SAT and ACT results for students in testing
+      grades. Miami's high school starts at grade 9 and has no college-testing
+      cohort yet, so there is nothing for Miami to report. Revisit when Miami
+      reaches grade 11. Ratified on #4996.
     columns:
       - name: region
         data_type: string
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__college_assessments_wide.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__college_assessments_wide.yml
@@ -1,5 +1,9 @@
 models:
   - name: rpt_gsheets__college_assessments_wide
+    description: >-
+      **Excludes Miami by design.** Same population as
+      `rpt_gsheets__college_assessments_long`: SAT and ACT results for testing
+      grades, which Miami does not have yet. Ratified on #4996.
     config:
       contract:
         enforced: true
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__csgf_hs_ap_offerings.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__csgf_hs_ap_offerings.yml
@@ -1,5 +1,12 @@
 models:
   - name: rpt_gsheets__csgf_hs_ap_offerings
+    description: >-
+      High-school AP course offerings for Charter School Growth Fund reporting.
+      Network-wide, Miami included: there is no region filter, so Miami enters
+      the population as its high school grows. It has no rows yet because the
+      reporting window is the prior academic year, before Miami had a high
+      school. Note the `is_enrolled_recent` filter, which was null for Miami
+      until #5038. Ratified on #4996.
     columns:
       - name: school_name
         data_type: string
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__csgf_hs_enrollment.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__csgf_hs_enrollment.yml
@@ -1,5 +1,10 @@
 models:
   - name: rpt_gsheets__csgf_hs_enrollment
+    description: >-
+      High-school enrollment and Algebra I outcomes for Charter School Growth
+      Fund reporting. Network-wide, Miami included, on the same reasoning as
+      `rpt_gsheets__csgf_hs_ap_offerings`: no region filter, no Miami rows yet
+      only because the window is the prior academic year. Ratified on #4996.
     columns:
       - name: studentid
         data_type: int64
--- a/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__kfwd_flc.yml
+++ b/src/dbt/kipptaf/models/extracts/google/sheets/properties/rpt_gsheets__kfwd_flc.yml
@@ -1,5 +1,9 @@
 models:
   - name: rpt_gsheets__kfwd_flc
+    description: >-
+      **Excludes Miami by design.** KIPP Forward tracks alumni through college,
+      and Miami has no graduating class yet, so it has no alumni to track.
+      Ratified on #4996.
     columns:
       - name: student_number
         data_type: int64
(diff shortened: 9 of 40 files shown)


cbini:
## Review response — per finding

Verified each finding against the code before acting. Fixes in 1d475512a.

| # | Finding | Verdict |
| --- | --- | --- |
| 1 | Tracking-issue refs in `description:` | Confirmed rule conflict — needs a maintainer decision, not my call |
| 2 | ASCII `--` instead of em dash | Fixed |
| 3 | "by design" on disabled models | Fixed |
| 4 | Column descriptions still absent | Declined, out of scope |
| 5 | YAML structural check | No action; adding the parser result the job could not get |

### 1. Tracking-issue refs — the rule really does say this, and I am not overriding it unilaterally

`src/dbt/CLAUDE.md:1262-1264` is explicit: *"Don't put TODOs, history, migration
plumbing, or tracking-issue refs (`#3142`, etc.) in descriptions — those go in
inline SQL comments at the derivation site."* By that rule all 40 descriptions
here are violations.

The precedent is also real: 71 properties files under `src/dbt/kipptaf/models`
already carry `#NNNN` refs, several of them in these same 40 files and untouched
by this PR.

I am leaving the refs in place and flagging it rather than picking a side, for two
reasons:

1. **Stripping them defeats the PR.** The whole point is to record *where a
   decision was made* so a future reader can find the reasoning. A ratification
   note that cannot name the ratification is not much of a record.
2. **The rule's own carve-out points the other way.** `:1047` keeps tracking refs
   *inline at the derivation site* — but a model-level **scope** decision has no
   derivation site. It is not about a line; it is about whether a region belongs
   in the report at all. There is nowhere else for it to live.

So this looks less like 40 mistakes and more like a gap in the rule. The clean
resolution is a carve-out at `:1262` for ratification and provenance refs, which
is a CLAUDE.md change and therefore the maintainer's call, not something I should
slip into a docs PR. **@cbini — your call: add the carve-out, or strip the refs
from all 40.**

Cross-reference: #5038's review pushed the opposite direction, asking for
rationale to move *into* descriptions. Both are right and they are compatible —
rationale to the description, tracking refs to the code. This case is awkward only
because a scope decision has no code site.

### 2. Em dash — fixed

Confirmed the ratio. 6 descriptions switched to `—`.

### 3. "Excludes Miami by design" on disabled models — fixed

A good catch, and the contradiction was in the same sentence: claiming a design
decision and then saying the scope is moot. Those five now read **"Not evaluated
for Miami"** and state what would happen if the model were re-enabled. Two more
disabled models got the same treatment for consistency, so all 7 read alike.

### 4. Column descriptions still absent — declined

Correct, and correctly labelled out of scope. `int_students__student_enrollments`
gained column descriptions in #5038 because that PR modified those columns.
Backfilling descriptions for every column of 40 models is a real piece of work and
belongs in its own issue, not smuggled into a scope-recording PR.

### 5. Structural check — the parser result you could not get

`dbt parse` succeeded on the full kipptaf project against this branch, and
`trunk check --force` is clean on all 40 files (and on the 13 touched by the fix
commit). So the YAML is parser-verified, not just read through.

One thing your static read could not have caught, found while applying these
fixes: my first attempt at the reword normalised each description to a single line
before re-wrapping, which silently collapsed the paragraph breaks in
`int_extracts__course_enrollments_by_term` — including one that predated this
branch. Reverted and redone paragraph-aware. Worth knowing that a folded-scalar
edit can lose structure without changing a single word.

PR #5044, opened 2026-08-27
feat(dbt): branch fct_grades_gpa onto Focus for Miami

### LINKED ISSUE #5021: feat(dbt): ingest the Focus student_gpa_calculated table and branch fct_grades_gpa for Miami

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will put Miami rows into `fct_grades_gpa` for the
first time, in any year.

It adds `stg_focus__student_gpa_calculated` to the focus package and its kipptaf
wrapper, then routes `fct_grades_gpa` through a new SIS-neutral
`int_students__gpa` instead of reading `int_powerschool__gpa_term` and
`int_powerschool__gpa_cumulative` directly. That follows the
`int_students__final_grades` precedent from #5020 rather than branching the
PowerSchool GPA chain in place.

The ingestion half shipped in #5034.

## Reviewer Notes

**Read this before reviewing the SQL: what Focus actually gives us is much less
than #5021 assumed.** Three of that issue's premises are wrong, measured against
the loaded table:

- `marking_period_id = -1` is not a filter to apply. It is the only value in the
  table, so Focus supplies **no term-grained GPA at all**. Term, semester and
  year-to-date columns are null on every Miami row.
- Class rank is not "a bonus". It is populated on **3 of 441 rows**.
- Coverage is 319 students against 1,757 enrolled. Not a join bug: every student
  with imported course history has a row, all 290 of them, plus 29 more. This is
  GPA over grades imported from a prior school.

I raised this before building and was asked to ship the cumulative-only branch
anyway, so that is what this is. The nulls are deliberate and documented rather
than filled from the cumulative value, which would read as a real term measure.

**The key swap surfaces Miami's pre-Focus history, which is a visible change
beyond the issue's ask.** Moving the enrollment join from
`(studentid, schoolid, yearid)` to `(student_number, schoolid, academic_year)`
adds **13,928** Miami PowerSchool rows for AY2020 through AY2025. Those rows
existed all along; `int_students__student_enrollment_union` leaves `studentid`
null on every Miami row, so the old join matched nothing. It is a gain, and it
satisfies the issue's "Miami historical GPA still readable" criterion directly,
but a reviewer should know the fact goes from zero Miami rows to ~14k.

**NJ parity is exact**, measured against prod with both join forms:

| Region | Old key | New key |
| --- | --- | --- |
| kippcamden | 47,018 | 47,018 |
| kippnewark | 165,670 | 165,670 |
| kipppaterson | 2,536 | 2,536 |

**The `unique` test on `grades_gpa_key` warns with 8 duplicates. Those are
pre-existing** — prod has the same 8 keys, verified by set comparison, and this
branch adds none.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — all 3
      new models have one
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application —
      `fct_grades_gpa`'s existing exposures are unchanged; no column was added,
      removed or retyped
- [x] **Breaking change?** No. `fct_grades_gpa` keeps its exact column set and
      types; only its upstream and its enrollment join key change. No
      versioning needed.
      with **`--target staging`** — not applicable, the Focus source is a
      BigQuery-native dlt landing dataset, not an external table

## CI checks

- [ ] **Trunk** — `trunk check --force` clean on all 9 files locally (2 prettier
      reflows, applied by the commit hook)
- [ ] **dbt Cloud** — see the open question below; this will fail until the
      kipptaf source can resolve
- [ ] **Dagster Cloud** — not triggered, no `src/teamster/` change

<details>
<summary>For Claude</summary>

### Local validation

Built against prod upstreams with `--favor-state --defer`:

| Model | Result |
| --- | --- |
| `stg_focus__student_gpa_calculated` (package) | 441 rows, contract enforced, 4 tests pass including the FK to `stg_focus__students` |
| `int_students__gpa` | 240,300 rows, grain test passes |
| `fct_grades_gpa` | builds; `term_key` and `student_enrollment_key` FK tests pass |

`dbt build --empty` over `int_students__gpa+ fct_grades_gpa+
stg_focus__student_gpa_calculated+`: 7 pass, 0 error.

Note the first build ran without `--favor-state` and produced a Focus-only
`int_students__gpa` of 441 rows, plus 267 phantom FK orphans — `--defer` had
resolved the PowerSchool upstreams to stale personal dev tables. Both symptoms
vanished with `--favor-state`. Recorded because both look like catastrophic
findings and neither was real.

### Branch composition, measured

| Region | Branch | Years | Rows | Reaching the fact |
| kippmiami | Focus | 2026 | 441 | 265 |
| kippmiami | PowerSchool archive | 2019–2025 | 16,512 | 13,796 |
| kippcamden | PowerSchool | 2015–2026 | 47,734 | 46,774 |
| kippnewark | PowerSchool | 2002–2026 | 173,062 | 165,178 |
| kipppaterson | PowerSchool | 2023–2026 | 2,520 | 2,520 |

The cutover boundary holds with no overlap: Miami's PowerSchool arm stops at
2025, its Focus arm starts at 2026. 16,512 matches the archive count stated on
#5021 exactly.

441 Focus rows become 265 in the fact because the enrollment join also matches
on school, and a course-history GPA row can sit at a school the student is not
currently enrolled at. NJ behaves the same way.

### Grain-test scope

`int_students__gpa`'s `unique_combination_of_columns` is scoped
`where: student_number is not null`. `int_powerschool__gpa_term` carries 4 Camden
AY2015 rows at `schoolid = 0` whose student fails the `dcid >= 1` placeholder
filter, so the left join leaves `student_number` null and they collide. They are
kept rather than dropped, and `fct_grades_gpa` excludes them on its inner join
anyway.

### Column mapping choice

Focus's `cumulative_weighted_gpa` maps to PowerSchool's `cumulative_y1_gpa` and
`cumulative_gpa` to `cumulative_y1_gpa_unweighted` — across the naming, not by
name, because the unqualified PowerSchool column is the weighted one.
`gpa`/`weighted_gpa` are dropped: they equal the cumulative pair to within
0.0005 on every populated row.

`cumulative_y1_gpa_projected` and `potential_credits_cum` are null on Focus rows;
Focus projects nothing.

</details>

Refs #5021

### DIFF
--- a/src/dbt/focus/models/staging/properties/stg_focus__student_gpa_calculated.yml
+++ b/src/dbt/focus/models/staging/properties/stg_focus__student_gpa_calculated.yml
@@ -1,0 +1,126 @@
+models:
+  - name: stg_focus__student_gpa_calculated
+    description: >-
+      Focus's own GPA and class-rank calculations. One row per student per
+      school. Every row carries `marking_period_id = -1`, which the vendor
+      defines as the course-history GPA — GPA over grades imported into a
+      student's official academic history, not GPA for work done in the current
+      marking period. There is no term-grained row in this table.
+
+
+      Coverage is narrow and that is a property of the source, not of this
+      model. Measured on the first prod load: 441 rows for 319 students, against
+      1,757 Miami students enrolled in AY2026. Every student with imported
+      course history has a row here, all 290 of them with no exceptions, plus 29
+      more. Rows appear at every grade level from K through 9, so this is not a
+      high-school-only table.
+
+
+      `gpa` and `cumulative_gpa` are the same measure at different rounding —
+      they differ by at most 0.0005 across every populated row — as are
+      `weighted_gpa` and `cumulative_weighted_gpa`. Prefer the cumulative pair
+      downstream.
+
+
+      The five decimal columns arrive from dlt as BIGNUMERIC, because
+      `weighted_gpa` is an unbounded Postgres `numeric` carrying up to 20
+      decimal places and needs the `widen_unbounded_numeric` dlt adapter to load
+      at all. They are cast to NUMERIC here, which keeps this package's decimal
+      vocabulary consistent and costs at most 5e-10 of rounding.
+
+
+      The 21 `custom_*_gpa`, `custom_*_rank` and `ncaa_gpa` columns in the
+      source are omitted: all are 100% null, as is `mp`. Add one when Focus
+      starts populating it.
+    columns:
+      - name: id
+        description: Primary key — Focus student GPA calculation id.
+        data_type: int
+        data_tests:
+          - unique:
+              config:
+                severity: error
+          - not_null:
+              config:
+                severity: error
+      - name: student_id
+        description: >-
+          Focus student id. Note this is the network student number prefixed
+          with `8400` (Miami-Dade's FLDOE district number), not the bare student
+          number.
+        data_type: int
+        data_tests:
+          - not_null:
+              config:
+                severity: error
+          - relationships:
+              arguments:
+                to: ref('stg_focus__students')
+                field: student_id
+              config:
+                severity: error
+        config:
+          meta:
+            contains_pii: true
+      - name: school_id
+        description: >-
+          Focus school the calculation belongs to. A student can hold a row at
+          more than one school — 122 of 319 students did on the first load — so
+          the grain is student and school, not student alone.
+        data_type: int
+      - name: syear
+        description: Focus school year (start year).
+        data_type: int
+      - name: marking_period_id
+        description: >-
+          Marking period the calculation covers. `-1` marks the course-history
+          GPA, per Focus SIS Level 1 Certification, Day 2. It is the only value
+          present.
+        data_type: int
+      - name: grade_id
+        description: >-
+          Grade level the student held when the calculation ran. Null on the
+          rows that carry no `grade_total`.
+        data_type: int
+      - name: grade_total
+        description: Number of grades the calculation was computed over.
+        data_type: int
+      - name: class_rank
+        description: >-
+          Class rank Focus computed. Populated on 3 of 441 rows on the first
+          load, so it is not yet usable as a reporting measure.
+        data_type: int
+      - name: weighted_class_rank
+        description: Class rank by weighted GPA. Populated on 3 of 441 rows.
+        data_type: int
+      - name: unweighted_class_rank
+        description: Class rank by unweighted GPA. Populated on 3 of 441 rows.
+        data_type: int
+      - name: rank_last_modified
+        description: Date Focus last recomputed the rank for this row.
+        data_type: date
+      - name: created_at
+        description: Row creation timestamp in Focus.
+        data_type: timestamp
+      - name: updated_at
+        description: Row last-update timestamp in Focus.
+        data_type: timestamp
+      - name: gpa
+        description: >-
+          Unweighted GPA over the student's course history. Equal to
+          `cumulative_gpa` to within 0.0005 on every populated row.
+        data_type: numeric
+      - name: weighted_gpa
+        description: >-
+          Weighted GPA over the student's course history. Equal to
+          `cumulative_weighted_gpa` to within 0.0005 on every populated row.
+        data_type: numeric
+      - name: cumulative_gpa
+        description: Cumulative unweighted GPA.
+        data_type: numeric
+      - name: cumulative_weighted_gpa
+        description: Cumulative weighted GPA.
+        data_type: numeric
+      - name: cumulative_credits
+        description: Credits the cumulative GPA was computed over.
+        data_type: numeric
--- a/src/dbt/focus/models/staging/sources-bigquery.yml
+++ b/src/dbt/focus/models/staging/sources-bigquery.yml
@@ -342,6 +342,15 @@ sources:
                 - dlt
                 - focus
                 - student_report_card_grades
+      - name: student_gpa_calculated
+        config:
+          meta:
+            dagster:
+              asset_key:
+                - "{{ project_name }}"
+                - dlt
+                - focus
+                - student_gpa_calculated
       - name: report_card_grades
         config:
           meta:
--- a/src/dbt/focus/models/staging/stg_focus__student_gpa_calculated.sql
+++ b/src/dbt/focus/models/staging/stg_focus__student_gpa_calculated.sql
@@ -1,0 +1,21 @@
+select
+    id,
+    student_id,
+    school_id,
+    syear,
+    marking_period_id,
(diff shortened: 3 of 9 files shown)


cbini:
## Review verdicts — both findings accepted and fixed in `10a5298f`

Verified each claim against the file before changing anything. Both hold.

### Finding 1 — row-level calculation in `WHERE`. Accepted, fixed differently.

Confirmed: `powerschool_conformed` computed `gt.yearid + 1990 as academic_year`
in the SELECT list and re-computed `(gt.yearid + 1990)` in `WHERE`. That is a
row-level calculation in `WHERE`, which `src/dbt/CLAUDE.md` prohibits.

I did **not** take the suggested `gt.yearid >= sc.focus_start_academic_year -
1990`. That still puts arithmetic in the predicate, just on the other side. The
convention's stated remedy is to precompute as a named column, so `sis_cutover`
now carries `focus_start_yearid` and `WHERE` compares a bare `gt.yearid`:

```sql
sis_cutover as (
    select
        focus_start_academic_year,
        focus_start_academic_year - 1990 as focus_start_yearid,
    from {{ ref("int_students__sis_cutover") }}
),
```

Same effect on the duplicate expression, and it reads as a boundary rather than
an offset applied at the comparison site.

### Finding 2 — misleading `schoolid` description. Accepted, fixed.

Confirmed: the description named `focus_school_id`, which is the join key, while
the value is `loc.powerschool_school_id`. Reworded to name both, in the roles
they actually play. This matters more than it looks — `src/dbt/focus/CLAUDE.md`
documents three distinct school identifier spaces, and joining the wrong one
null-fills every school attribute with no error.

### The change is behavior-preserving, measured

Rebuilt against prod upstreams with `--favor-state`:

| Check | Before | After |
| --- | --- | --- |
| `int_students__gpa` rows | 240,300 | 240,300 |
| Grain test | pass | pass |
| `grades_gpa_key` duplicates | 8 (pre-existing) | 8 (pre-existing) |

Per-region row counts equal prod `int_powerschool__gpa_term` exactly — Camden
47,734, Miami 16,512, Newark 173,058, Paterson 2,520 — which also rules out
fan-out on the `int_powerschool__gpa_cumulative` join.

One thing worth recording: Newark read 173,062 on the earlier build and 173,058
now. That is upstream drift, not this change — prod `int_powerschool__gpa_term`
rematerialized between the two builds, and my filter sits inside
`not (_dbt_source_project = 'kippmiami' and ...)`, so it cannot touch Newark
under any value. Verified by querying the prod table directly.

### On the `is_current` false lead

The review's own note is right, and worth keeping: `int_students__gpa` drops
`is_current`, but `origin/main`'s `fct_grades_gpa` never read `gt.is_current` —
the mart's `is_current` has always come from its local `row_number()`. Dropping
an already-dead column, not an altitude change.

Issue #5041, opened 2026-08-27
fix(dbt): Miami is missing from a network topline metric because is_self_contained is blank


cbini:
## What's happening

Miami students are missing from three reports because one true/false field is
blank for them.

Miami's student system (Focus) does not record whether a student is taught in a
self-contained classroom. New Jersey's system does. So `is_self_contained` is
filled in for every New Jersey student and blank for all 1,760 Miami students.

Reports filter on that field. They ask to keep students who are **not**
self-contained. In SQL a blank does not answer "no" — it answers "unknown", and
unknown rows get dropped. So Miami disappears without any report saying so.

Three reports are affected:

| Report | What Miami loses |
| --- | --- |
| `int_topline__student_metrics` | Miami contributes **zero rows to a network topline metric**. 68,484 AY2026 week rows are available and none are counted. |
| `rpt_tableau__dibels_dashboard` | Miami is absent from the literacy dashboard, even though Miami is a DIBELS region. |
| `rpt_tableau__ops_dashboard` | Correct result today, but reached by accident. This one selects the special-education population, which Miami has none of, so zero is the right answer. |

We chose not to write "no" into the field. Writing "no" would claim someone
checked and found nobody, and nobody checked. That is the same reasoning as
#4968.

The topline metric is the reason to revisit. When that call was made,
`rpt_tableau__dibels_dashboard` looked like the only casualty.

## Steps to reproduce

1. Confirm the field is blank for Miami and filled for everyone else:

   ```sql
   select region, count(*) as n, countif(is_self_contained is null) as blank
   from kipptaf_extracts.int_extracts__student_enrollments
   where academic_year = 2026
   group by 1
   ```

   Expect Miami 1,760 of 1,760 blank, and 0 blank everywhere else.

1. Count Miami rows in `kipptaf_topline.int_topline__student_metrics`. Expect 0.

1. Confirm the rows exist upstream: count Miami AY2026 rows in
   `kipptaf_extracts.int_extracts__student_enrollments_weeks`. Expect 68,484.

## Where

- **Code location / dbt project:** `kipptaf`
- **Environment:** prod
- **Related:** #4996 recorded the decision to leave the field blank. #5038
  populated the three sibling flags and deliberately left this one alone.

<details>
<summary>For Claude</summary>

Context, not instructions.

**Where the field is filtered.** `not is_self_contained` appears in
`rpt_tableau__dibels_dashboard.sql:170`, and the field combines with
[...]

Issue #3887, opened 2026-05-12
fix(data): 62 duplicate FLEIDs across distinct student_numbers in KIPP Miami Focus


cbini:
## Problem

`kippmiami_powerschool.stg_powerschool__u_studentsuserfields` has **14 FLEIDs each associated with 2+ distinct PowerSchool `student_number`s** (28 rows total). Each FLEID should map to exactly one student in PS. Inspection suggests the duplicates are a mix of:

- True duplicates — same student re-enrolled with a new `student_number` rather than reactivated against the prior record.
- Possible data-entry errors — different students assigned the same FLEID.

Filtering out withdrawn enrollments (`enroll_status = 1`) reduces the count to **7 remaining duplicate FLEIDs** among non-withdrawn records, each still tied to ≥2 different `student_number`s.

## Why this matters

The duplicates produce join fan-out in `kippmiami/int_fldoe__all_assessments` when resolving FLEID → `student_number`. Surfaced on [#3886](https://github.com/TEAMSchools/teamster/issues/3886) — the new join now requires `dbt_utils.deduplicate` to keep the model's uniqueness test passing. Without the dedupe, downstream marts inherit ambiguous `student_number` resolution for those FLEIDs and the uniqueness test warns with 175 fan-out rows.

The dedupe is a workaround. The right fix is to clean the source data so each FLEID maps to one canonical `student_number`.

## Identification queries

All-cohort duplicates (14 FLEIDs, 28+ rows):

```sql
with dup_fleids as (
    select suf.fleid
    from `teamster-332318`.kippmiami_powerschool.stg_powerschool__u_studentsuserfields as suf
    where suf.fleid is not null
    group by suf.fleid
    having count(distinct suf.studentsdcid) > 1
)
select
    suf.fleid,
    s.student_number,
    s.dcid,
    s.enroll_status,
    s.entrydate,
    s.exitdate,
    s.first_name,
    s.last_name,
    s.dob,
    s.schoolid,
from `teamster-332318`.kippmiami_powerschool.stg_powerschool__students as s
inner join `teamster-332318`.kippmiami_powerschool.stg_powerschool__u_studentsuserfields as suf
    on s.dcid = suf.studentsdcid
inner join dup_fleids using (fleid)
order by suf.fleid, s.dcid
```

Higher-priority cohort (7 FLEIDs, non-withdrawn only):

```sql
with dup_fleids as (
    select suf.fleid
[...]

PR #5047, opened 2026-08-28
fix(dbt): attach Florida scores on student_number in both state assessment models

### LINKED ISSUE #5042: fix(dbt): the state testing dashboard drops every Florida score at the student join
## What's happening

The state testing dashboard loads Miami's Florida test scores, then throws them
all away when it attaches them to students.

`rpt_tableau__state_assessments_dashboard` is built to cover Miami. It reads
Florida FAST results and it even has a Miami-only rule that relabels a social
studies course as Civics. But it shows **zero Miami rows**.

The dashboard stacks three sources of scores: New Jersey results from Pearson,
Florida results from FLDOE, and a Pearson preliminary file. It then attaches
scores to students with one join, on a column called `localstudentidentifier`.

The Florida source hardcodes that column as blank and puts Florida's identifier
in a different column instead. The join only reads the blank one. A blank never
matches anything, so every Florida score is dropped.

This is not an ID mismatch. `localstudentidentifier` is just the network student
number, and Miami students have one. The Florida rows simply never carry it.

## Steps to reproduce

1. Count Miami rows in `kipptaf_tableau.rpt_tableau__state_assessments_dashboard`.
   Expect 0.

1. Confirm Florida scores exist upstream:

   ```sql
   select academic_year, count(*) as n
   from kipptaf_fldoe.int_fldoe__all_assessments
   where scale_score is not null
   group by 1
   order by 1
   ```

1. Read
   [`rpt_tableau__state_assessments_dashboard.sql:178`](../blob/main/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__state_assessments_dashboard.sql#L178).
   The Florida branch reads `null as localstudentidentifier`.

## Where

- **Environment:** prod
- **Related:** found during the code review on #5038. Recorded on #4996, which
  classified this model as in scope for Miami.

<details>
<summary>For Claude</summary>

Context, not instructions.

**Exact mechanism.** In the `assessment_scores` CTE the three legs are unioned.
The FLDOE leg (line ~174 onward) emits:

```sql
null as localstudentidentifier,
student_id as state_id,
```

The Pearson legs emit a real `localstudentidentifier`. The attaching join at
line 336 is:

```sql
from assessment_scores as a
inner join {{ ref("int_extracts__student_enrollments") }} as e
    on a.academic_year = e.academic_year
    and a.localstudentidentifier = e.pearson_local_student_identifier
```

`null = anything` is null, so the inner join discards every FLDOE row. The same
pattern repeats at lines 365 and 372 for the two `sf` joins, and in
`int_tableau__state_assessments_demographic_comps.sql:114` and `:204`.

**`pearson_local_student_identifier` is the network student number** —
`int_students__student_enrollments.sql:261` reads `ar.student_number as
pearson_local_student_identifier`. It is populated for Miami. The name is
misleading; it is not Pearson-specific and it is not a legacy or state id.

**What FLDOE's `student_id` is.** It is Florida's student identifier, carried
through `int_fldoe__all_assessments`. `int_extracts__student_enrollments` already
exposes Miami's Florida id as `fleid`, and sets `state_studentnumber` to `fleid`
for Miami (see the `if(e.region = 'Miami', e.fleid, e.state_studentnumber)`
expression). So the Florida side has a key on both ends already.

**Two candidate fixes.** Prefer whichever keeps the join single-keyed:

1. Map Florida's id to the network student number inside the FLDOE leg, so
   `localstudentidentifier` is populated for every leg and the existing join
   works unchanged. Cleanest for the consumer, but needs a crosswalk from
   `fleid` to `student_number` at that point in the graph.
1. Add a second join condition — `a.state_id = e.fleid` — alongside the existing
   one, so each leg joins on the key it actually carries. Smaller diff, but the
   join grows an OR-shaped condition, which is easy to get wrong and can
   fan out. Verify row counts per region if you go this way.

**Verification.** NJ counts must not move: Camden, Newark and Paterson currently
hold rows here and the FLDOE leg contributes none of them, so any correct fix is
NJ-neutral by construction — confirm rather than assume. For Miami, sanity-check
the result against `int_fldoe__all_assessments` row counts by year and subject
before declaring it fixed; a fan-out will show up as Miami rows exceeding the
upstream score count.

**Not the same defect as #5041.** That one is a blank boolean upstream. This one
is a hardcoded null inside this model's own union. Fixing either does not fix
the other.

Refs #4996
Refs #5038

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will attach Miami's Florida test scores to students on the network student number, in both state assessment models and in every other model that reads `int_fldoe__all_assessments.student_number`.

Three changes, one key.

**The dashboard union leg carries `student_number`.** `rpt_tableau__state_assessments_dashboard` stacks three score sources and attaches them to enrollments with one join on `localstudentidentifier`. The Florida leg set that column to null, so the join dropped every Florida score. The leg now selects `student_number`, the key the New Jersey legs already use.

**The comps model joins on `student_number`.** `int_tableau__state_assessments_demographic_comps` matched Florida scores on the state id. It now matches on the network student number, like the two New Jersey legs above it.

**`int_fldoe__all_assessments` resolves `student_number` from Focus.** The kippmiami model looked each FLEID up in the frozen pre-Focus PowerSchool archive, which holds the bare 6-digit number. Main made Miami's `student_number` the 8400-prefixed Focus id on 2026-09-04 (b37f575645), so that lookup stopped matching anything. The lookup now reads `int_focus__students`, which is live and already carries the prefix. That also closes the gap the archive could never fill: students who enrolled after the cutover now resolve.

This deletes a second Miami block in the dashboard that joined scores on the state id. With `fleid` now wired from Focus on main, that block is live again and is what puts Miami on the dashboard in prod today. Keeping it alongside the fixed union leg would emit every Miami row twice, and its state-id join fans out on 12 keys. One key path stays.

Outcome, dev build against prod upstreams:

| region   | prod rows | dev rows | prod distinct keys | dev distinct keys |
| -------- | --------- | -------- | ------------------ | ----------------- |
| Camden   | 13,135    | 13,135   | 13,129             | 13,129            |
| Newark   | 35,425    | 35,425   | 35,425             | 35,425            |
| Paterson | 440       | 440      | 440                | 440               |
| Miami    | 21,064    | 20,997   | 21,052             | 20,997            |

Miami lands 67 rows below prod. 24 of those are prod's own duplicates (12 keys emitted twice through the state-id join). The rest are students the Focus lookup does not resolve. The comps model and its report are unchanged on every row and flag count, Miami included.

A new data test fails if either model ever holds zero Miami rows again. It is what caught the renumbering: dbt Cloud built both models with 0 Miami rows on this branch after the main merge.

## Reviewer Notes

- The approval on 2026-09-04 predates the third change. Main's renumbering landed the same day and broke the `student_number` key this PR relies on, so the FLEID lookup moved to Focus. Please re-review the kippmiami diff.
- dbt Cloud CI reads kipptaf's `int_fldoe__all_assessments` from the staging clone `zz_stg_kippmiami_fldoe`, which still holds the archive-resolved numbers. The check stays red until that table rebuilds from this branch. CI builds kipptaf only, so a local `dbt build` is the validation for the kippmiami model.
- 62 FLEIDs sit on more than one Focus student record (#3887). The dedupe keeps the lowest `student_number`, the migrated record: 216 of the 279 affected score rows attach, against 55 for the highest.
- Every other kipptaf consumer keyed on `int_fldoe__all_assessments.student_number` (the enrollment-scoped fact, the assessment roster sheet, the section resolver) picks the fix up on the next build. Not separately verified here.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — no new models. The new data test has a `properties.yml` entry; the dashboard and both `int_fldoe__all_assessments` descriptions are updated.
- [x] Include (or update) an
      for all models consumed by a dashboard — the `state_testing_analysis_tool` exposure already covers both models and is unchanged.
- [x] **Breaking change?** Values, not schema. `int_fldoe__all_assessments.student_number` changes from the bare PowerSchool number to the 8400-prefixed network number, which is what every consumer now joins against. The dashboard's column list is identical: 66 columns before and after, same names and types.
- [x] If adding or modifying an external source, run `stage_external_sources` — no external sources touched. The frozen `kippmiami_powerschool` source is removed from kippmiami; kipptaf keeps its own.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — red until `zz_stg_kippmiami_fldoe.int_fldoe__all_assessments` rebuilds from this branch (see Reviewer Notes).

Closes #5042

Refs #4996

<details>
<summary>For Claude</summary>

**Timeline.** Miami's `fleid` was null on every enrollment row from the Focus cutover (2026-08-07) until cc37022766 wired it from Focus (2026-09-04). In that window both models held 0 Miami rows, which is what #5042 reported. The same day, b37f575645 made `int_focus__students.student_number` the 8400-prefixed Focus id, so `int_extracts__student_enrollments.student_number` and `pearson_local_student_identifier` moved to 10 digits for Miami while `int_fldoe__all_assessments.student_number` stayed 6-digit via the archive lookup. Key overlap between the two, on any join shape, went to 0.

**Change 1, the dashboard union leg.** `rpt_tableau__state_assessments_dashboard.sql:178` emitted `null as localstudentidentifier`; the attach join inner-joins `a.localstudentidentifier = e.pearson_local_student_identifier`. The leg now selects `student_number`.

**Change 2, the dashboard's second Miami block, deleted.** A third final-select block joined `a.state_id = e.state_studentnumber` under `e.region = 'Miami'`. Dead in the cutover window, live again since cc37022766, and the source of prod's 21,064 Miami rows today, 24 of them duplicates on 12 keys in AY2022. Kept beside change 1 it would double every Miami row.

**Change 3, the comps join key.** `int_tableau__state_assessments_demographic_comps.sql:274` joined `e.state_studentnumber = a.student_id`; it now joins `e.pearson_local_student_identifier = a.student_number`.

**Change 4, the FLEID lookup.** kippmiami `int_fldoe__all_assessments` replaced its `fleid_lookup_raw` CTE over `stg_powerschool__students` x `stg_powerschool__u_studentsuserfields` with `int_focus__students` (`florida_education_identifier`, `student_number`). `dbt_utils.deduplicate` order flipped from `student_number desc` to `asc`: measured over the 62 multi-mapped FLEIDs, `asc` attaches 216 of 279 score rows to an enrollment, `desc` 55. The `kippmiami_powerschool` source block is removed from `kippmiami/models/fldoe/sources-bigquery.yml`; kipptaf declares its own `kippmiami_powerschool` source and is untouched.

**Verification, dev build against prod upstreams.** kippmiami model built into `zz_cbini_kippmiami_fldoe` with `--defer --favor-state`; kipptaf `int_fldoe__all_assessments`, both state assessment models, the comps report, and the new test built into `zz_cbini_kipptaf_*` in one selection. Every view definition checked via `INFORMATION_SCHEMA.VIEWS`: FLDOE refs resolve to the dev copies, enrollments to `kipptaf_extracts`. 9 nodes, all pass, including both `unique_combination_of_columns` grain tests and `test_state_assessment_joins_resolve_miami`.

Upstream resolution, dev `int_fldoe__all_assessments`: 22,087 rows, 45 null `student_number` (39 before), every resolved number above 8400000000. Shortfall from upstream to dashboard: 45 null key, 1,039 no `rn_year = 1` enrollment that year, 6 at grade 2 or below, 20,997 attached.

Comps model: `KTAF NJ` 3,814 prod and dev, `KTAF FL` 1,680 prod and dev. `rpt_tableau__state_assessments_dashboard_comps`: identical to prod on every (region, comparison_entity) row count and flag count.

Dashboard column parity: `except distinct` on (column_name, data_type), prod vs dev, 0 rows.

**CI.** Run 70403223961500 at d17f7d30 failed only on the new test (2 results: both models 0 Miami rows). CI's dashboard view resolves `int_fldoe__all_assessments` to `zz_stg_kipptaf_fldoe`, a view over `zz_stg_kippmiami_fldoe`, which is the archive-resolved build. Per `.claude/rules/dbt-models.md`, refreshing a shared `zz_stg_*` table needs direct user authorization, so the staging rebuild is handed to the author.

Not the same defect as #5041, which is a blank boolean upstream.

</details>


cbini:
Verdicts on all three findings. The review ran against `fde0d0b4` (the `opened`
event), so it did not see `c2356b47`, which changes the picture on finding 1.

**Finding 1 — the comps model has the identical bug. Agreed, and now fixed in
this PR.** `c2356b47` repoints
`int_tableau__state_assessments_demographic_comps.sql:274` from
`e.state_studentnumber = a.student_id` to
`e.pearson_local_student_identifier = a.student_number`, matching the two NJ
legs in the same CTE. The model goes from 0 to 1,680 `KTAF FL` rows, and
`rpt_tableau__state_assessments_dashboard_comps` gains 627 `Region` rows.

The consequence was worse than missing rows. Miami's 2,627 existing
city/state/neighborhood comparison rows all read
`region_matched_or_outperformed = false`, because that column is
`if(b.percent_proficient >= a.percent_proficient, true, false)` against a
`comparison_entity = 'Region'` self-join that had no Miami rows to find. A false
there meant "not measured", not "did not outperform". 75 now read true. NJ is
unchanged on every row count and every flag count. No issue needed.

**Finding 2 — the view automation condition limits what the test catches.
Agreed, no change.** Correct as stated, and correctly scoped: it is the same
limitation the cited precedent test carries, and it is not introduced here.
Making the check re-run on upstream data changes means flipping the model to a
table, which is a materialization decision with its own cadence tradeoffs and
does not belong in a join fix.

**Finding 3 — the frozen crosswalk leaves a residual gap. Mechanism agreed, the
trend claim is backwards, and the risk is real but starts later than stated.**

`kippmiami/CLAUDE.md:14-17` confirms the dataset is frozen and retired, and the
`fleid_lookup` CTE holds 3,198 students, so it can never gain a row. That part
is right.

But the prediction that the gap "should widen every year" does not match the
data. Unresolved `student_number` by year, from the suggested query:

| year | rows  | unresolved | rate  |
| ---- | ----- | ---------- | ----- |
| 2020 | 608   | 5          | 0.82% |
| 2021 | 1,168 | 14         | 1.20% |
| 2022 | 3,958 | 6          | 0.15% |
| 2023 | 4,903 | 12         | 0.24% |
| 2024 | 5,779 | 0          | 0     |
| 2025 | 5,671 | 2          | 0.04% |

The worst years are the oldest, and the trend runs toward zero. The snapshot was
taken at cutover, so it holds every student enrolled through AY2025 — which is
why AY2025 resolves at 99.96%.

The real exposure is a cliff, not a slope, and it has not arrived: AY2026 is the
first cohort that can enroll without ever appearing in the frozen snapshot, and
no AY2026 FLDOE scores have landed yet. When the first AY2026 FAST file loads,
those students resolve to a null `student_number` and drop at the same inner
join this PR just fixed.

Out of scope here — it is an upstream crosswalk limitation, and nothing in this
diff introduces or could fix it. Worth its own issue before the AY2026 results
land, rather than after. Happy to file one.

GabyRangelB:
@cbini — flagging you so you know this has been reviewed.

**Accepting this as proposed — no changes requested.** Merging it puts nine years of Miami results on the dashboard, and that shouldn't wait on anything else.

One clarification for the record, because the description attributes the breakage to the hardcoded null and that's only half of it.

**Why the field broke.** The FLDOE leg's `null as localstudentidentifier` was one cause. The other is that `fleid` is null on *every* Miami enrollment row, in every academic year:

| AY | Miami rows | `fleid` populated |
| --- | --- | --- |
| 2018 | 198 | 0 |
| 2021 | 1,356 | 0 |
| 2025 | 1,521 | 0 |
| 2026 | 1,770 | 0 |

`fleid` has one source — `stg_powerschool__u_studentsuserfields` joined on `students_dcid`. When Focus replaced PowerSchool as the source of Miami enrollment history (the spine conformance work on 2026-08-07), Miami's rows stopped carrying `students_dcid`, so that join stopped resolving. Not just for new students — retroactively, for all nine years. `state_studentnumber` for Miami is derived from `fleid`, so it went null network-wide at the same moment.

So the second Miami block wasn't dead by original construction. It worked until the Focus cutover, and the cutover is what killed it.

**Keying FL on the state ID was a deliberate design decision, established in 2023.** It wasn't incidental complexity or an oversight. Florida issues a stable FLEID, is proactive about correcting them, and a student keeps theirs across schools and systems — so FL scores were intentionally attached on the state ID rather than conformed to `localstudentidentifier`. NJ needs the local-identifier path precisely because NJ state IDs on vendor files aren't dependable the same way. Two keys was the point, not an accident.

**And we're accepting the new design anyway.** Given that Focus is now the source of Miami enrollment for all years and the FLEID no longer reaches the enrollment record, keying FL on the network student number — the same key as the NJ legs — is the right call going forward. We are not pursuing alternatives: not restoring `fleid` through the Focus branch, not re-sourcing the FLEID lookup. This supersedes the 2023 decision, and the reasoning above is the record of why.

Two residuals we're knowingly accepting, noted so they're not rediscovered as surprises:

- The `fleid_lookup` in `int_fldoe__all_assessments` reads the frozen pre-Focus PowerSchool dataset, so it can't gain rows. 527 Miami students enrolled in AY2026 have a FLEID in Focus but no entry in it; 328 are in grades 3-9. The first FAST file containing them will drop their scores at the attach join.
- `#3887`'s dedupe (`order by student_number desc`) currently decides the attachment for 195 score rows across 13 FLEIDs.

Neither is introduced by this PR and neither blocks it.

cbini:
@GabyRangelB verdicts on the 2026-09-04 comment, plus a change in the PR since your approval.

**The second Miami block was killed by the cutover, not dead by construction. Agreed, and the PR body now says so.** Confirmed in git: `fleid` on Miami enrollments went null with the Focus spine work on 2026-08-07 and came back when cc37022766 wired it from Focus on 2026-09-04. In that window both models held 0 Miami rows, which is what #5042 reported.

**Keying Florida on the state ID was a 2023 design decision. Noted for the record, no change.** The reasoning stands as written; the PR body records that this supersedes it.

**"Not restoring `fleid` through Focus, not re-sourcing the FLEID lookup." Both happened anyway, and the second one is now in this PR.** Main did the first the same day as your comment (cc37022766). It also renumbered Miami's `student_number` to the 8400-prefixed Focus id (b37f575645). That broke the key this PR relies on: `int_fldoe__all_assessments.student_number` still came from the frozen PowerSchool archive as the bare 6-digit number, so it matched 0 enrollment rows. The new presence test caught it: dbt Cloud run 70403223961500 built both models with 0 Miami rows after the main merge.

Commit 0257668492 moves the FLEID lookup to `int_focus__students`. Focus is live and already carries the prefix. Dev build against prod upstreams: Miami 20,997 dashboard rows, comps `KTAF FL` 1,680, comps report identical to prod on every row and flag count, New Jersey unchanged everywhere. Prod today holds 21,064 Miami rows through the revived state-id block, 24 of them duplicates on 12 keys; the student number path has none.

The same renumbering broke every other kipptaf consumer keyed on `int_fldoe__all_assessments.student_number` (the enrollment-scoped fact and the section resolver join it to Focus-numbered enrollments). Inferred from the join keys, not measured; the lookup fix repairs them on the next build.

**Residual 1, the frozen crosswalk cliff. Resolved by the change above.** The lookup no longer reads the archive, so AY2026 entrants resolve when their first FAST file lands.

**Residual 2, the #3887 dedupe. Still real, in Focus form.** 62 FLEIDs sit on more than one Focus student record (279 score rows). The tiebreak flipped to the lowest `student_number`, which is the migrated record enrollments carry: 216 of the 279 rows attach, against 55 for the highest. The TODO stays on the CTE.

**Asks.** The kippmiami diff is new since your approval, so please look at it. dbt Cloud stays red until `zz_stg_kippmiami_fldoe.int_fldoe__all_assessments` is rebuilt from this branch; that is on me.

GabyRangelB:
@cbini re-reviewed the kippmiami diff. **Still approved, no changes requested.**

Reproduced your numbers independently against prod: the 62 multi-mapped FLEIDs and 279 affected score rows match, and at the real join grain (`rn_year = 1`, `grade_level > 2`, matching academic year) the `asc` tiebreak attaches 216 against 55 for `desc`. Also confirmed `int_focus__students` now emits the 8400-prefixed `student_number`, which is what made the archive lookup stop matching, and that dropping the `kippmiami_powerschool` source is safe — no remaining references in the kippmiami project, kipptaf keeps its own declaration, and the kipptaf copy of `int_fldoe__all_assessments` is a passthrough over the kippmiami source, so it picks the fix up rather than needing its own.

Noted and accepted: the Focus lookup leaves 45 unresolved against the archive's 39, so Miami lands 43 real rows below prod on top of prod's 24 duplicates. Trading a handful of pre-cutover stragglers for a lookup that resolves post-cutover entrants is the right side of that deal, and it retires the frozen-archive dependency entirely — which closes the cliff from my last comment rather than deferring it.

All seven checks are green now that the staging rebuild landed. Good to merge from our end.

GabyRangelB:
@cbini - give me a sec with this one. i asked claude if the rows would be on prod once we merged and he said yeah and now he is saying no

cbini:
@GabyRangelB yes, the rows land in prod after merge. Here is the mechanism, so the two answers you got can be reconciled.

**Prod path.** Prod dbt models are materialized by each code location's automation sensor, not by dbt Cloud. The table condition (`src/teamster/core/automation_conditions.py`) fires on `code_version_changed`, so within minutes of the post-merge kippmiami deploy, `kippmiami/fldoe/int_fldoe__all_assessments` rebuilds with the Focus lookup and 8400-prefixed `student_number`. Its last two materializations were exactly that kind of rebuild. Every kipptaf model in this PR is a view, so the dashboard, the comps model, and the comps report read the new numbers the moment that table lands. There is a window of a few minutes between the kipptaf deploy and the kippmiami rebuild where the dashboard shows 0 Miami rows; if the presence test happens to run inside it, it fails once and passes on the next materialization.

**Where the "no" came from.** dbt Cloud CI builds kipptaf only, and it resolves unmodified kipptaf models to the staging clone `zz_stg_kipptaf_fldoe.int_fldoe__all_assessments`, a view over prod `kipptaf_fldoe`. Every main merge re-clones it from prod, which still has the bare numbers, so CI on this PR went red again at 16:50 after the main merge. That is a CI artifact, not a prod outcome. Commit on the branch just now marks the kipptaf passthrough modified so CI builds it in the PR schema from the staged kippmiami table, the same move 91fcccffa2 used. It stays green across main merges from here.

PR #5049, opened 2026-08-28
fix(dbt): put Miami back in the per-term course report

### LINKED ISSUE #5043: fix(dbt): Miami lost its per-term course rows because Focus records no section enrollment count
## What's happening

A per-term course report lost Miami this school year, because Focus does not
record how many students are in a section.

`int_extracts__course_enrollments_by_term` filters out course sections that
record zero students. That filter exists to remove a PowerSchool quirk: a section
that has enrollment rows but records its student count as zero. It drops 215 of
64,699 New Jersey rows, about a third of a percent.

Miami's `sections_no_of_students` is blank on all 19,363 AY2026 course
enrollments. Comparing a blank to zero gives "unknown", and unknown rows get
dropped. So the filter removes every Miami row.

Miami had 50,155 rows here for AY2025 and has **zero** for AY2026.

There are no zero-enrollment Miami sections to remove. Miami's 19,363 enrollment
rows cover 709 sections and the smallest has 1 student. A section with nobody in
it produces no enrollment rows at all, so it cannot reach this model. The filter
has nothing to do for Miami.

## Steps to reproduce

1. Confirm Miami is gone this year and present last year:

   ```sql
   select academic_year, region, count(*) as n
   from kipptaf_extracts.int_extracts__course_enrollments_by_term
   where academic_year >= 2025
   group by 1, 2
   order by 1, 2
   ```

   Expect Miami 50,155 for AY2025 and absent for AY2026.

1. Confirm the column is blank for Miami:

   ```sql
   select count(*) as n, countif(sections_no_of_students is null) as blank
   where cc_academic_year = 2026
   ```

   Expect 19,363 and 19,363.

## Where

- **Code location / dbt project:** `kipptaf`, and the `focus` package if the
  count is derived there
- **Environment:** prod
- **Related:** #4996 classified this model as in scope for Miami and carved this
  blocker out of #5038, because it is a Focus data gap rather than an enrollment
  flag.

<details>
<summary>For Claude</summary>

Context, not instructions.

**The filter.** `int_extracts__course_enrollments_by_term.sql:115` reads
`where not e.is_dropped_section and e.sections_no_of_students != 0`. The drop
flag half is fine — #4968 made it non-null for Miami. The count half is the
blocker.

**Do NOT carry Focus's own seat field through.** `stg_focus__course_periods`
has `filled_seats`, `total_seats`, `ell_seats` and `sped_seats`. `filled_seats`
reads **0 on all 823 Miami AY2026 course periods** — nobody maintains it. Piping
it into `sections_no_of_students` would swap a blank for a zero and the filter
would still drop every Miami row, while looking like a real fix in the diff.
`total_seats` is populated (only 10 of 823 are zero) but it is room capacity, not
enrollment — wrong number.

**Do NOT hardcode a positive literal.** It asserts a count that was never
measured, and the count is derivable.

**Preferred fix: derive the real count in the Focus branch of
`int_students__course_enrollments`.** `focus_conformed` already sits at the
enrollment grain with `cc_sectionid` and `cc_academic_year`, so a window
(`count(*) over (partition by cc_academic_year, cc_sectionid)`) produces a true
value with no new join. Measured from prod, that yields 709 sections for AY2026
with 1 to 65 students each.

Three other models read this column and would all benefit rather than just the
one report: `int_extracts__course_schedule_by_term`,
`int_tableau__gradebook_audit_teacher_scaffold`, and
`int_tableau__gradebook_audit_student_scaffold` (the last is disabled).

**Cheaper fallback** if the derivation is deferred: make the single filter
null-safe at `:115`. That returns Miami's rows immediately but leaves the column
blank for the other three consumers, so it trades a fixed report for a
still-missing column. Prefer the derivation.

**Note the PowerSchool semantics before matching them.** PowerSchool's
`no_of_students` is a stored section attribute that can disagree with the actual
enrollment count — that disagreement is exactly what the `!= 0` filter exploits.
A derived Focus count cannot reproduce that quirk, and should not try: for Focus
the derived count and the true count are the same thing by construction. Say so
in the column description so a future reader does not assume the two branches
mean the same thing.

**Verification.** NJ counts must not move — the derivation touches only the
Focus branch, so confirm Camden, Newark and Paterson row counts are byte-equal
against prod. For Miami, expect roughly AY2025's order of magnitude once the
filter passes; check against `int_students__course_enrollments` Miami AY2026
row count (19,363) times the quarter fan-out rather than eyeballing it.

Refs #4996
Refs #4968
Refs #5038

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will put Miami back in the per-term course
report. Miami had 50,155 rows there last school year and zero this year.

Two comparisons against blank values kept it out, and this pull request fixes
both.

The first is the section student count. Focus does not record how many students
are in a section, so that count is blank on all 19,398 Miami rows this year. The
report drops sections that record zero students, to remove a PowerSchool quirk
where a section has students in it but records its count as zero. Comparing a
blank to zero gives "unknown", and unknown rows get dropped, so the filter
removed every Miami row. The filter is now blank-safe. It still removes the 202
New Jersey rows that record a real zero.

The second is the enrollment end date. Focus leaves that date blank while a
student is still in the course, which is 95.8% of Miami rows. The report matches
each enrollment to the quarters it overlaps, and that match compares the end
date, so it dropped all but 230 of the rows the first fix rescued. The report now
falls back to the last day of the section's term when the end date is blank.

To make that fallback possible, the Focus half of `int_students__course_enrollments`
now fills in the term start and end dates. Those two fields already exist and are
filled in for every New Jersey row; they were blank for Miami. Focus records the
same fact as the schedule row's marking period, which is filled in on every row.

Miami now lands at 62,952 rows for this year, against 50,155 last year. New
Jersey does not move.

## Reviewer Notes

**Why the fallback is the term end and not a far-future date.** The obvious
fallback is a date far in the future, which is what `dim_student_section_enrollments`
uses for a different purpose. It is wrong here. A far-future date makes a
semester or quarter enrollment look like it ran all year, so it gets matched to
quarters its term never covered. Measured against prod, that adds 3,938 rows
that should not exist. The term end date does not.

**Why the blank end date is left alone.** It would be tempting to fill in
`cc_dateleft` itself. That field is a student's leave date, and it feeds outbound
extracts — `rpt_illuminate__roster` publishes it as "12 Leave Date". Filling it
would claim every Miami student left on their term end date, which is not true.
#5002 already settled this: the field stays blank and each consumer handles it.

### General

Skipped — no Dagster changes.

- [ ] Include (or update) an
      summary above.
      See

No new models, no new exposures, no external sources. No columns renamed or
removed: both changes fill in existing fields that were blank for Miami, and
neither model is contracted.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

**Scope correction against the issue.** #5043's fold-out prefers deriving a real
section count in the Focus branch, on the grounds that 3 other models read
`sections_no_of_students` and would benefit. That premise does not hold:
`int_extracts__course_schedule_by_term` and
`int_tableau__gradebook_audit_teacher_scaffold` read the column from
`base_powerschool__sections`, a different model that a Focus branch on
`int_students__course_enrollments` does not reach, and both gradebook scaffolds
are `enabled: false`. Exactly one live consumer reads the column from
`base_powerschool__course_enrollments` — the model this pull request fixes. So
the derivation was skipped in favor of the blank-safe predicate. Deriving a real
count is still available as future work and is unblocked by this change.

**Predicate form.** `is distinct from 0` rather than `coalesce(x, 1) != 0`. No
sentinel literal, and it states the intent directly.

**Measurements, all against prod on 2026-08-28.**

| Check | Result |
| --- | --- |
| `sections_no_of_students` null, Miami AY2026 | 19,398 of 19,398 |
| `sections_no_of_students` null, every NJ project-year | 0 |
| `sections_no_of_students = 0`, NJ AY2026 | 202 (199 Camden, 3 Newark) |
| `terms_lastday` null, all districts all years | 19,398 — exactly the Focus rows |
| `marking_period_end_date` null in `int_focus__schedule` | 0 of 19,398 |
| `cc_dateleft` null, Miami AY2026 | 18,582 of 19,398 |

**NJ neutrality is a proof, not a sample.** `x != 0` and `x is distinct from 0`
are identical when `x` is not null, and `coalesce(cc_dateleft, terms_lastday)`
returns `cc_dateleft` when `cc_dateleft` is not null. Both columns are non-null
on every NJ row in every year, so NJ output cannot move.

**End-to-end verification.** The compiled model SQL was run against prod
upstreams with the Focus term-window fill simulated by joining
`int_focus__schedule` on `cc_dcid = student_schedule_id`. AY2025 reproduced prod
exactly — Miami 50,155, Newark 156,882, Camden 51,392, Paterson 5,573 — which
establishes the harness is faithful. AY2026 came back Miami 62,952, Newark
167,447, Camden 55,319, Paterson 15,238. The AY2026 NJ numbers differ from the
materialized prod table (165,583 / 54,459 / 15,238) because prod is a snapshot
from its last build and the query reads live views; AY2025 is frozen and matches
to the row.

**Sentinel comparison, at the `schedule_by_terms` grain, Miami AY2026:**

| Focus marking period | Far-future sentinel | Term-end sentinel |
| --- | --- | --- |
| FY | 61,087 | 61,087 |
| SEM | 10,417 | 6,609 |
| QTR | 195 | 65 |

3,938 spurious rows avoided.

**The `not_null` test on `terms_lastday`** guards the assumption the coalesce
rests on. If Focus ever stops populating `marking_period_end_date`, the report
silently regresses to dropping Miami rows; the test makes that loud instead.

**Not fixed here.** `int_assessments__scaffold` matches on
`administered_at between cc_dateenrolled and cc_dateleft`, and
`int_assessments__resolved_section_enrollments` and `fct_grades_assignments`
compare `< cc_dateleft`. All three hit the same blank end date for Miami and are
likely dropping rows for the same reason. Each now has `terms_lastday` available
to coalesce to. Not investigated or measured — out of scope for #5043.

</details>

Closes #5043
Refs #5002
Refs #4996

### DIFF
--- a/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
@@ -112,7 +112,16 @@ with
             and e._dbt_source_project = q._dbt_source_project
             and e.cc_dateenrolled <= q.quarter_end_date_alt
             and e.cc_dateleft >= q.quarter_start_date_alt
-        where not e.is_dropped_section and e.sections_no_of_students != 0
+        where
+            not e.is_dropped_section
+            -- Focus records no section student count, so this column is null on
+            -- every Miami row. Null is not the PowerSchool quirk the filter
+            -- targets -- a section holding enrollment rows while recording its
+            -- count as zero, 215 of 64,699 New Jersey rows. A section with
+            -- nobody in it produces no enrollment rows at all, so it cannot
+            -- reach this model; #5043 measured Miami's smallest section at 1
+            -- student. Keep null, drop only a literal 0.
+            and e.sections_no_of_students is distinct from 0
     ),
 
     days_course_enrolled as (
--- a/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
@@ -62,7 +62,7 @@ with
             e.cc_academic_year,
             e.cc_schoolid,
             e.cc_dateenrolled as dateenrolled,
-            e.cc_dateleft as dateleft,
+            e.scheduled_end_date as dateleft,
             e.cc_sectionid as sectionid,
             e.cc_course_number as course_number,
             e.sections_dcid,
@@ -99,9 +99,9 @@ with
             ) as dateenrolled_alt,
 
             if(
-                e.cc_dateleft > q.last_day_school_year,
+                e.scheduled_end_date > q.last_day_school_year,
                 q.last_day_school_year,
-                e.cc_dateleft
+                e.scheduled_end_date
             ) as dateleft_alt,
 
         from {{ ref("base_powerschool__course_enrollments") }} as e
@@ -111,7 +111,7 @@ with
             and e.cc_schoolid = q.schoolid
             and e._dbt_source_project = q._dbt_source_project
             and e.cc_dateenrolled <= q.quarter_end_date_alt
-            and e.cc_dateleft >= q.quarter_start_date_alt
+            and e.scheduled_end_date >= q.quarter_start_date_alt
         where
             not e.is_dropped_section
             -- Focus records no section student count, so this column is null on
--- a/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
@@ -130,7 +130,7 @@ with
             -- Focus records no section student count, so this column is null on
             -- every Miami row. Null is not the PowerSchool quirk the filter
             -- targets -- a section holding enrollment rows while recording its
-            -- count as zero, 215 of 64,699 New Jersey rows. A section with
+            -- count as zero, 202 New Jersey rows in AY2026. A section with
             -- nobody in it produces no enrollment rows at all, so it cannot
             -- reach this model; #5043 measured Miami's smallest section at 1
             -- student. Keep null, drop only a literal 0.
--- a/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
@@ -62,7 +62,7 @@ with
             e.cc_academic_year,
             e.cc_schoolid,
             e.cc_dateenrolled as dateenrolled,
-            e.scheduled_end_date as dateleft,
+            e.exit_date as dateleft,
             e.cc_sectionid as sectionid,
             e.cc_course_number as course_number,
             e.sections_dcid,
@@ -99,9 +99,9 @@ with
             ) as dateenrolled_alt,
 
             if(
-                e.scheduled_end_date > q.last_day_school_year,
+                e.exit_date > q.last_day_school_year,
                 q.last_day_school_year,
-                e.scheduled_end_date
+                e.exit_date
             ) as dateleft_alt,
 
         from {{ ref("base_powerschool__course_enrollments") }} as e
@@ -111,7 +111,7 @@ with
             and e.cc_schoolid = q.schoolid
             and e._dbt_source_project = q._dbt_source_project
             and e.cc_dateenrolled <= q.quarter_end_date_alt
-            and e.scheduled_end_date >= q.quarter_start_date_alt
+            and e.exit_date >= q.quarter_start_date_alt
         where
             not e.is_dropped_section
             -- Focus records no section student count, so this column is null on
--- a/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
@@ -56,19 +56,6 @@ with
         }}
     ),
 
-    -- Focus leaves cc_dateleft null while an enrollment is open, the normal
-    -- state for 95.8% of Miami AY2026 rows (#5002). Every comparison below
-    -- would then be unknown and the row would drop, so coalesce once here. The
-    -- sentinel is the section's term end, not a far-future date: a far-future
-    -- date fans semester and quarter enrollments into quarters their term
-    -- never covered, 3,938 spurious rows measured on #5043. PowerSchool never
-    -- leaves cc_dateleft null, so this is a no-op for New Jersey.
-    course_enrollments as (
-        select
-            * except (cc_dateleft), coalesce(cc_dateleft, terms_lastday) as cc_dateleft,
-        from {{ ref("base_powerschool__course_enrollments") }}
-    ),
-
     schedule_by_terms as (
         select
             e._dbt_source_project,
@@ -117,7 +104,7 @@ with
                 e.cc_dateleft
             ) as dateleft_alt,
 
-        from course_enrollments as e
+        from {{ ref("base_powerschool__course_enrollments") }} as e
         inner join
             quarters as q
             on e.cc_academic_year = q.academic_year
--- a/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
@@ -56,6 +56,19 @@ with
         }}
     ),
 
+    -- Focus leaves cc_dateleft null while an enrollment is open, the normal
+    -- state for 95.8% of Miami AY2026 rows (#5002). Every comparison below
+    -- would then be unknown and the row would drop, so coalesce once here. The
+    -- sentinel is the section's term end, not a far-future date: a far-future
+    -- date fans semester and quarter enrollments into quarters their term
+    -- never covered, 3,938 spurious rows measured on #5043. PowerSchool never
+    -- leaves cc_dateleft null, so this is a no-op for New Jersey.
+    course_enrollments as (
+        select
+            * except (cc_dateleft), coalesce(cc_dateleft, terms_lastday) as cc_dateleft,
+        from {{ ref("base_powerschool__course_enrollments") }}
+    ),
+
     schedule_by_terms as (
         select
             e._dbt_source_project,
@@ -104,7 +117,7 @@ with
                 e.cc_dateleft
             ) as dateleft_alt,
 
-        from {{ ref("base_powerschool__course_enrollments") }} as e
+        from course_enrollments as e
         inner join
             quarters as q
             on e.cc_academic_year = q.academic_year
--- a/src/dbt/kipptaf/models/students/intermediate/int_students__course_enrollments.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_students__course_enrollments.sql
@@ -37,6 +37,12 @@ with
             csc.is_advanced_math,
             csc.discipline,
 
+            -- PowerSchool has no separate departure date: cc_dateleft is the
+            -- actual leave date when the student left and the term end plus a
+            -- day while the enrollment is open. It is therefore already the
+            -- neutral concept, and is non-null on all 751,359 rows since 2004.
+            a.cc_dateleft as scheduled_end_date,
+
             {{ extract_region("a") }} as region,
 
             case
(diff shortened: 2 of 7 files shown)


cbini:
Verdicts on all 6 findings, verified against the code before replying. 3 fixed in 24db815, 3 declined with reasons.

| # | Finding | Verdict |
| --- | --- | --- |
| 1 | YAML column-test ordering | **Fixed** |
| 2 | `* except (...)` on one line | **Declined** — formatter output |
| 3 | Stale `dateleft` description | **Fixed** |
| 4 | 215 vs 202 New Jersey rows | **Fixed** |
| 5 | YAGNI on `terms_firstday` | **Declined** |
| 6 | Unit test for quarter fan-out | **Declined** |

---

**1. Fixed.** Confirmed against `src/dbt/CLAUDE.md` line 1247: "Columns with per-column `data_tests:` must be sorted to the top of the `columns:` list for visibility." `terms_lastday` carries the `not_null` test, so it belongs there. Moved `terms_firstday` with it rather than splitting the pair.

**2. Declined.** This is sqlfmt's own output, not a missed hook. `trunk check --force` on the file returns "No issues", and the pre-commit `fmt` hook produced exactly this line — my working copy had the two expressions on separate lines and sqlfmt joined them, because together they fit the line budget. The other `* except (...)` sites you cite sit alone because their neighboring expressions did not fit. `src/dbt/CLAUDE.md` states that sqlfmt and sqlfluff enforce formatting; it has no one-expression-per-line rule of its own. Splitting the line by hand would be rejoined on the next commit.

**3. Fixed.** Correct, and the sharpest finding here — this was the one with a real forward-looking risk. `dateleft` now reads: for PowerSchool the raw `cc_dateleft`; for Miami, `cc_dateleft` coalesced to `terms_lastday`, never null, so it cannot be read as a still-enrolled signal. Also corrected `dateleft_alt`, which described clamping the raw value.

**4. Fixed, and it was a plain error, not a grain difference.** The `215 of 64,699` figure was copied from #5043's body, measured when the issue was written. The measurement behind this branch, run 2026-08-28, is 202 New Jersey rows with `sections_no_of_students = 0` in AY2026 — 199 Camden, 3 Newark, 0 Paterson — at the pre-join enrollment grain, which is the grain the filter operates on. The comment now says 202, matching the model description.

**5. Declined, on the judgment you left to the author.** `terms_firstday` and `terms_lastday` are one fact: the section's term window. Both come from the same already-present join at the cost of one `select` line. A half-populated pair is the more expensive outcome — the next reader has to work out whether the missing half is a deliberate decision or a bug, which is the exact class of question #5043 existed to answer.

**6. Declined as a required change; the regression it names is already guarded twice.** A `not_null` test on `terms_lastday` guards the assumption the coalesce rests on — if Focus ever stops populating `marking_period_end_date`, the fallback silently reverts to dropping every Miami row, and the test makes that loud. The existing presence check guards the rows themselves.

Neither catches a wrong-fan-out regression, which is a fair point. But that regression only arises from deliberately swapping the sentinel for a far-future date, and both the filter comment and the column description now state the measured cost of doing so: 3,938 spurious rows, from 10,417 vs 6,609 semester rows and 195 vs 65 quarter rows at the `schedule_by_terms` grain. A unit test for it would need mocks for `base_powerschool__course_enrollments`, `int_students__calendar_week`, `int_students__terms`, `int_extracts__student_enrollments` and `int_people__staff_roster` — a large fixture surface to guard against an edit that two comments already argue against. Worth revisiting if this model gets a unit test for other reasons.

---

On the environment note: the data-dependent claims in the PR body were measured against prod through the BigQuery MCP, and the compiled model was run end-to-end. AY2025 reproduced prod to the row across all four regions, which is what establishes the harness is faithful.

src/dbt/kipptaf/models/students/intermediate/int_extracts__course_enrollments_by_term.sql
```diff
@@ -56,6 +56,19 @@ with
...
+    -- would then be unknown and the row would drop, so coalesce once here. The
+    -- sentinel is the section's term end, not a far-future date: a far-future
+    -- date fans semester and quarter enrollments into quarters their term
+    -- never covered, 3,938 spurious rows measured on #5043. PowerSchool never
+    -- leaves cc_dateleft null, so this is a no-op for New Jersey.
+    course_enrollments as (
+        select
+            * except (cc_dateleft), coalesce(cc_dateleft, terms_lastday) as cc_dateleft,
+        from {{ ref("base_powerschool__course_enrollments") }}
+    ),
```
cbini: i think this is the wrong altitude for this fix. it's a Focus fix. it should be fixed on a Focus model or a Focus leg on a conformed model

cbini: You're right, and the reasoning I used to keep it downstream was wrong. Moved to the Focus leg in 0425761.

I had argued `cc_dateleft` should stay null because it feeds `rpt_illuminate__roster` as "12 Leave Date", so filling it would assert a departure that never happened. I checked that instead of assuming it, and PowerSchool does not mean what I thought: `cc_dateleft` is in the future on ~90% of NJ AY2026 rows and is never null on any of the 751,359 PowerSchool rows since 2004. It is a scheduled end, not a departure. So Focus leaving it null is a conformance gap, exactly as you said, and the fix belongs in `focus_conformed`.

`focus_conformed` now fills it from the schedule row's `marking_period_end_date`. The `course_enrollments` CTE here and the `terms_firstday` / `terms_lastday` carry are both reverted — they only ever existed to serve the downstream coalesce — and the `not_null` test moved onto `cc_dateleft`, which is the column the guarantee is actually about.

One sub-decision worth flagging, since it is a deliberate divergence. PowerSchool's convention for an open enrollment is the day **after** the term ends — `terms_lastday + 1` on 58,772 of the roughly 61,000 open NJ AY2026 rows. I did not match that. Miami's marking periods abut its quarter boundaries, so an exclusive bound lands exactly on the next quarter's start date, and this model compares it inclusively (`cc_dateleft >= quarter_start_date_alt`). Measured against prod, the `+1` form invents 1,712 course-quarter rows with zero instructional days, versus 19 for Camden and 0 for Newark and Paterson. So the Focus leg carries the term's last day, and the two branches differ by one day for an open enrollment. That is documented on the column.

The inclusive comparison is a third defect and I left it alone — 2,203 NJ AY2026 rows already sit past their section's term end because of it, so fixing it moves New Jersey and is not #5043's scope.

Verification: Miami AY2026 lands at 62,952 rows against 50,155 for AY2025, and AY2025 reproduces prod to the row in all four regions. NJ cannot move — `cc_dateleft` is non-null on every PowerSchool row, so the coalesce never fires. The Miami section-enrollment orphan rate goes 10.74% to 10.79%, inside the 15% threshold on `test_miami_section_enrollment_orphan_rate`.

Blast radius, stated plainly: this makes `cc_dateleft` non-null for Miami everywhere, so models that were silently dropping Miami on a null comparison will now include it — `int_assessments__scaffold`, `int_assessments__resolved_section_enrollments`, `fct_grades_assignments`, `int_powerschool__gradebook_assignments_scores`. That is the point of fixing at this altitude rather than in one report, but it is a wider change than the PR title suggests. I also marked the `coalesce(cc_dateleft, '9999-12-31')` in `dim_student_section_enrollments` as now-dead rather than deleting it.

src/dbt/kipptaf/models/students/intermediate/int_students__course_enrollments.sql
cbini, line 1: this makes the same mistake that school enrollments did. we're conforming around powerschool language in what should be a sis-agnostic model.

cbini: Right again, and it's the same reasoning error I made the first time: I asked "what does PowerSchool put in this column?" instead of "what fact does each SIS actually record?" The tell was in my own comment — I had written a paragraph justifying a deliberate one-day divergence from PowerSchool's convention. Conforming to an encoding you then have to diverge from is not conformance.

Fixed in f0bb7e5 and 1ff76d6.

Two facts are tangled in `cc_dateleft`: when the student left, and when the section's term is scheduled to end. PowerSchool has no separate departure date, so it conflates them. Focus keeps them apart — `end_date` only on a real departure, `marking_period_end_date` always. Focus is not missing data; it declines to conflate, and I conflated it back.

`int_students__course_enrollments` now carries `scheduled_end_date`, resolved per branch at the unification layer:

- PowerSchool: `cc_dateleft`, which already is this concept and is non-null on all 751,359 rows since 2004.
- Focus: `coalesce(end_date, marking_period_end_date)`.

`cc_dateleft` goes back to being the departure date, null on an open Focus enrollment, which is what Focus records. The `not_null` test moved onto `scheduled_end_date`, the column that carries the guarantee. `int_extracts__course_enrollments_by_term` bounds its quarter window on the neutral column — no branch-specific logic downstream, and no coalesce that fires for one region only.

Behavior is unchanged from what CI already validated: Miami AY2026 at 63,532 rows against 50,155 for AY2025, New Jersey untouched.

I also reverted the course-overlap test scoping from 21818ce. It only existed to absorb a regression the back-fill caused, and that regression is gone now that `cc_dateleft` is null on Focus again. The underlying defect is real and I left it: the test partitions by `cc_studyear`, which is null on every Focus row, so Focus rows would collapse into one per-course partition if that column ever became non-null. Partitioning on `student_number` and `academic_year` fixes it for both branches but changes what the test measures for New Jersey.

On scope: I added one neutral column where a consumer needed one, and did not rename the PowerSchool-shaped columns around it. Checking first, `docs/superpowers/specs/2026-08-25-focus-course-enrollments-design.md` defers SIS-neutral vocabulary to Phase 5 under #3999, and #3999 names no vocabulary — so if `scheduled_end_date` is the wrong name for where that's heading, say so and I'll change it now while it has one consumer.

cbini: Answering my own open question: the marts already have the name. `dim_student_section_enrollments` calls this value `exit_date` and was doing the aliasing itself (`cc_dateleft as exit_date`), so `scheduled_end_date` was me inventing a term one layer below an established one. Renamed to `exit_date` in 59838a6.

The mart's meaning is the resolved one, not a literal departure — `dim_student_enrollments` documents `exit_date_key` as the day after a student attended class, and NJ's is a future date on ~90% of open AY2026 rows. So `exit_date` is exactly what the new column resolves per SIS branch.

I did NOT repoint the mart off its own alias, and that is a judgment call worth your eyes. Once `int_students__course_enrollments.exit_date` exists, `cc_dateleft as exit_date` in the dim is redundant and wrong for Miami (null). But `exit_date` feeds both dedupe order-bys in that model (`course_enrollment_rank` and `homeroom_rank`), and Miami's rows all tie there today precisely because the value is null. Measured on prod AY2026, Miami has 1,345 partitions holding more than one row, 2,715 rows in total, where populating it can change which row ranks first. That is a dim behavior change, not a rename, so it should not ride along on this PR uninspected. Say the word and I will measure the actual pick changes and do it, here or separately.

The matching upstream rename for the pair, `cc_dateenrolled` to `entry_date`, is Phase 5 / #3999 work — the mart aliases that one too.

Issue #4938, opened 2026-08-20
fix(dbt): int_powerschool__gpa_term has duplicate rows on its natural key and no uniqueness test


GabyRangelB:
## What's happening

`int_powerschool__gpa_term` returns more than one row for the same student, year, and term. There are 20 duplicated keys on `(studentid, yearid, term_name, _dbt_source_project)`, 16 of them in AY2026-27.

For 8 of the 20, the duplicate rows carry **different GPA values** — so any consumer joining on that key gets an arbitrary pick, and any consumer aggregating gets a double count. The other 12 are identical rows, which fan out silently without changing values.

The model has no uniqueness test in either properties yml — not in the `powerschool` package, and not at kipptaf level. Repo convention is that every intermediate model carries one, which is why this went unnoticed.

Found while adding a grain test to `rpt_tableau__hs_early_warning_dashboard`. That extract joins this model per student per term, so 4 duplicated keys belonging to current HS students fan out across each student's course rows and produce 48 duplicate row pairs on the dashboard. All 4 happen to be value-identical, so no number on that dashboard is currently wrong — but the fan-out is real and the extract cannot carry an error-severity uniqueness test until this is fixed.

## Steps to reproduce

1. Run the query below against prod.
1. `dup_keys` returns 20, and `keys_with_differing_gpa` returns 8.

```sql
with dups as (
  select
    studentid,
    yearid,
    term_name,
    _dbt_source_project,
    count(*) as n,
    count(distinct format('%T|%T|%T', gpa_y1, gpa_y1_unweighted, gpa_term)) as distinct_values
  from `teamster-332318.kipptaf_powerschool.int_powerschool__gpa_term`
  group by 1, 2, 3, 4
  having count(*) > 1
)
select
  count(*) as dup_keys,
  countif(distinct_values > 1) as keys_with_differing_gpa,
  countif(yearid = 36) as dup_keys_ay2026
from dups
```

## Where

- **Code location / dbt project:** `kipptaf` and the `powerschool` package
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** surfaced by #4937

<details>
<summary>For Claude</summary>

> Context, not instructions.

Model: `src/dbt/powerschool/models/sis/intermediate/int_powerschool__gpa_term.sql`.

Properties yml with no uniqueness test, both of which need one added:

- `src/dbt/powerschool/models/sis/intermediate/properties/int_powerschool__gpa_term.yml`
[...]

PR #5050, opened 2026-08-28
fix(dbt): dedupe int_students__student_enrollment_union on the enrollment key grain

### LINKED ISSUE #5045: fix(dbt): dim_student_enrollments surrogate key collides for same-day re-enrollments, duplicating fact rows
## What's happening

`dim_student_enrollments` has duplicate `student_enrollment_key` values, and they
fan out into facts: `fct_grades_gpa`'s `unique` test on `grades_gpa_key` warns
with 8 failed keys (16 rows), all AY2026.

The key is a surrogate over `(student_number, _dbt_source_project,
academic_year, entrydate)`. A student with 2 enrollment stints in the same year
starting on the same `entrydate` — a same-day withdraw-and-re-enroll, or a
duplicate stint row upstream — collides. Measured 2026-08-27, the colliding rows
are New Jersey only:

| Region | AY2026 rows | Distinct students |
| --- | --- | --- |
| kippnewark | 7,265 | 7,239 |
| kippcamden | 2,287 | 2,284 |
| kipppaterson | 892 | 891 |

(Counts are the region-year totals where rows exceed distinct 4-part keys; the
collision population is the small excess in each.)

Zero Miami rows are involved — found during the Focus GPA verification on #5021
but unrelated to it.

## Why it matters beyond the warn

`student_enrollment_key` is the dim's primary key. Every fact that joins it
double-counts those students silently, whether or not that fact carries its own
uniqueness test.

## Steps to reproduce

1. `select student_enrollment_key from kipptaf_marts.dim_student_enrollments
   group by 1 having count(*) > 1` — expect 8+ keys.
1. Trace one back to `int_students__student_enrollment_union`: two stints, same
   `student_number`, `academic_year`, `entrydate`.

## Where

- **Environment:** prod

For Claude

Context, not instructions.

Two candidate fixes with different meanings:

- If the same-day double stint is real (withdraw and re-enroll same day), the
  surrogate key needs a disambiguator — the stint id or `exitdate`.
- If it is a duplicate upstream row, dedupe at
  `int_students__student_enrollment_union` and the key is fine.

Decide from the data which one this is before choosing. Check whether the 8 are
one pattern or a mix.

Related: #4938 tracks duplicate rows on `int_powerschool__gpa_term`'s natural
key — possibly the same upstream cause, check before fixing separately.

Acceptance: `dim_student_enrollments.student_enrollment_key` unique; the
`fct_grades_gpa` `unique` warn clears; NJ row counts otherwise unchanged
against prod.

cbini:
# Pull Request

## Summary and Motivation

When merged, this pull request will stop `dim_student_enrollments` from
returning two rows for the same student enrollment, and stop every fact that
joins it from double-counting those students.

`student_enrollment_key` is a surrogate key over `student_number`,
`_dbt_source_project`, `academic_year`, and `entrydate`. PowerSchool keeps a
student's current enrollment in `students` and their earlier stints in
`reenrollments`. When a school re-enters a student and backdates that re-entry
to a stint that is still open, both tables end up holding a record that starts
on the same day. The 4-part key cannot tell those two records apart, so the
model emits 2 rows where the key promises 1.

`int_students__student_enrollment_union` now keeps one row per key. It keeps the
stint the model already ranks first for the year, which is the open one. That
makes the uniqueness test on those 4 columns true, so this pull request also
raises it from a warning to an error.

Fixes #5045.

## Reviewer Notes

**This is containment for reporting, not a cure. Ops still has work to do.**
The 6 records are entered wrong in PowerSchool, and only Ops can correct them.
In 5 of them a school re-entered a student and backdated the re-entry to a
stint that was still open, so the student reads as starting twice on the same
day. In the 6th, one student holds 2 enrollment records covering the identical
window. Correcting them means fixing the entry dates in PowerSchool, or voiding
the duplicate record.

**The upstream test stays a warning on purpose, and it is the alert.** The
matching `dbt_utils.unique_combination_of_columns` on
`int_powerschool__student_enrollment_union` still fires on all 6 records after
this change, and keeps alerting through its Dagster asset check. So the
data-entry problem stays visible to us until Ops clears it, while the marts
stop double-counting in the meantime. Only the downstream test, which this
change actually makes true, goes to error.

**Deduplicating drops rows, never key values.** Both rows in a colliding pair
hash to the same key, so removing one cannot orphan a foreign key anywhere
downstream. That is what made this the safe option.

**Adding a component to the key was the wrong fix, for 2 reasons.** It would
not have worked, and it would have changed every hash. See the fold-out below.

**#4938 is a separate cause, and this does not fix it.** Measured against prod.

### General

### dbt

- [x] Include (or update) an
      exposure changes; no columns added or removed.
- [x] **Breaking change?** No. No column is renamed or removed, and no
      `student_enrollment_key` value changes. Only 6 duplicate rows go away.
      with **`--target staging`** — not applicable, no external sources touched.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

### What the 8 collisions actually are

The issue measured 8 colliding keys on 2026-08-27. Re-measured on 2026-08-28
there are 6, in 134,343 union rows. All are `academic_year` 2026, all have
`entrydate` 2026-07-01, all are New Jersey, and every prior academic year has
zero. Identifiers are held locally in `.claude/scratch/`, not here.

Every pair is 1 row from `stg_powerschool__students` (null
`reenrollments_dcid`) plus 1 row from `stg_powerschool__reenrollments`. They are
real PowerSchool records, not rows dbt invented.

The issue asked whether they are one pattern or a mix. They are a **mix**:

| Count | Region | Pattern |
| --- | --- | --- |
| 3 | Newark | Short stint exited, then a re-entry backdated to 7/1. Distinct `exitdate`. |
| 1 | Newark | School transfer, both sides backdated to 7/1. Distinct `exitdate` and `schoolid`. |
| 1 | Camden | Backdated re-entry plus a grade-level correction. Distinct `exitdate`. |
| 1 | Paterson | Identical `entrydate` **and** `exitdate`, differing only in `entrycode`. A true double-write. |

So neither branch of the issue's fork holds cleanly. `exitdate` as a
disambiguator fixes 5 and misses Paterson.

### Why the key was not changed

Two independent reasons.

**It would not have fixed the reported symptom.** `grades_gpa_key` is
`(student_number, _dbt_source_project, academic_year, entrydate, term_name)`.
Adding a component makes the 8 duplicate keys distinct, so the `unique` test
passes, but `fct_grades_gpa` still carries 2 rows per student-term holding
identical GPA values. The warning clears and the double-count survives. The
fan-out comes from the join to the enrollment model, so the fix has to be at
the enrollment grain.

**The blast radius is real.** Adding a component rehashes all 134,343 rows.

### Consumers of student_enrollment_key that were checked

13 sites derive the key. All 13 recompute it from source columns at build time.
None stores it incrementally — every marts model is `materialized: table` under
`+contract: enforced`, and no model in `marts/` is incremental. So a key change
would have been mechanically safe, but it would have had to land at all 13 at
once, and 3 of them reach `entrydate` through separate lineage
(`int_powerschool__ps_adaadm_daily_ctod` and
`base_powerschool__student_enrollments`) that would each need the new component
plumbed through.

Reads `int_students__student_enrollment_union` directly, so the dedupe covers
them:

- `dim_student_enrollments`
- `dim_student_ell_status`
- `dim_student_iep_status`
- `dim_student_meal_eligibility_status`
- `dim_student_section_enrollments`
- `fct_grades_gpa`
- `fct_behavioral_incidents`
- `fct_family_communications`
- `fct_student_attendance_streaks`
- `fct_student_attendance_interventions`
- `dim_student_enrollment_status` (via the 3 status dims plus
  `dim_student_enrollments`)

Separate lineage, unaffected either way:

- `fct_student_attendance_daily` — key from `int_students__attendance_daily`,
  whose `entrydate` comes from `int_powerschool__ps_adaadm_daily_ctod`. 31 rows
  sit on the colliding keys, but they are 31 distinct calendar days and its
  primary key has 0 duplicates.
- `fct_survey_submissions` and `bridge_survey_expectations` — key from
  `int_extracts__student_enrollments`, which reads
  `base_powerschool__student_enrollments`. 0 and 3 rows on the colliding keys
  respectively, no fan-out.

Marts carrying duplicate primary keys before this change, all of which the
dedupe clears:

| Model | Duplicate primary keys |
| --- | --- |
| `dim_student_enrollments` | 6 |
| `dim_student_enrollment_status` | 6 |
| `fct_grades_gpa` | 8 (16 rows) |
| `dim_student_meal_eligibility_status` | 1 |

### Which row survives

`order by rn_year` inside the `qualify`. `rn_year` is computed upstream as
`row_number() over (partition by studentid, yearid order by yearid desc,
exitdate desc)`, so within a colliding group it is exactly `exitdate desc` and
it is never tied. The survivor is the open stint in all 6 cases, and because
`year_in_school` and `year_in_network` are set only when `rn_year = 1`, keeping
the `rn_year = 1` row keeps them populated. Ordering by `exitdate desc` alone
would tie on the Paterson pair and could leave a survivor with `rn_year = 2`
and null `year_in_school`.

### #4938 ruled a separate cause

Measured against prod on 2026-08-28. `int_powerschool__gpa_term` has 16
duplicate natural keys today, all 16 with differing GPA values. Zero of them
belong to any of the 6 enrollment-collision students. In the other direction,
all 8 duplicate `grades_gpa_key` values in `fct_grades_gpa` are fully explained
by the enrollment collision. So #4938 contributes nothing to the
`fct_grades_gpa` warning and stays open on its own cause.

</details>

### DIFF
--- a/src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
@@ -32,12 +32,32 @@ with
         select *,
         from {{ ref("int_powerschool__student_enrollment_union") }}
         where _dbt_source_project != 'kippmiami'
+    ),
+
+    unioned as (
+        select *,
+        from powerschool_conformed
+
+        full union all corresponding
+
+        select *,
+        from focus_conformed
     )
 
+-- PowerSchool keeps the current enrollment in `students` and prior stints in
+-- `reenrollments`. A re-entry backdated to a still-open stint's entrydate
+-- therefore yields two rows on (student_number, _dbt_source_project,
+-- academic_year, entrydate) -- the grain `student_enrollment_key` hashes -- and
+-- every mart joining that key double-counts the student. Keep the year's
+-- primary stint: rn_year is already ranked exitdate desc upstream, so the
+-- survivor is the open stint and its year_in_school / year_in_network stay
+-- populated. This drops rows, never key values, so no existing hash changes and
+-- no downstream foreign key is orphaned. See #5045.
 select *,
-from powerschool_conformed
-
-full union all corresponding
-
-select *,
-from focus_conformed
+from unioned
+qualify
+    row_number() over (
+        partition by student_number, _dbt_source_project, academic_year, entrydate
+        order by rn_year
+    )
+    = 1
--- a/src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollment_union.sql
@@ -44,20 +44,12 @@ with
         from focus_conformed
     )
 
--- PowerSchool keeps the current enrollment in `students` and prior stints in
--- `reenrollments`. A re-entry backdated to a still-open stint's entrydate
--- therefore yields two rows on (student_number, _dbt_source_project,
--- academic_year, entrydate) -- the grain `student_enrollment_key` hashes -- and
--- every mart joining that key double-counts the student. Keep the year's
--- primary stint: rn_year is already ranked exitdate desc upstream, so the
--- survivor is the open stint and its year_in_school / year_in_network stay
--- populated. This drops rows, never key values, so no existing hash changes and
--- no downstream foreign key is orphaned. See #5045.
-select *,
-from unioned
-qualify
-    row_number() over (
-        partition by student_number, _dbt_source_project, academic_year, entrydate
-        order by rn_year
-    )
-    = 1
+    -- TODO(#5045): remove once Ops corrects the backdated PowerSchool re-entry
+    -- dates that put two stints on one entrydate.
+    {{
+        dbt_utils.deduplicate(
+            relation="unioned",
+            partition_by="student_number, _dbt_source_project, academic_year, entrydate",
+            order_by="rn_year asc",
+        )
+    }}
--- a/src/dbt/kipptaf/models/students/intermediate/properties/int_students__student_enrollment_union.yml
+++ b/src/dbt/kipptaf/models/students/intermediate/properties/int_students__student_enrollment_union.yml
@@ -11,6 +11,22 @@ models:
       Miami rows at all -- including its alumni graduate placeholders. Replaces
       `int_powerschool__student_enrollment_union` as the canonical
       enrollment-grain source for marts.
+
+
+      Deduplicated to one row per (`student_number`, `_dbt_source_project`,
+      `academic_year`, `entrydate`) -- the grain `student_enrollment_key`
+      hashes. PowerSchool holds a student's current enrollment in `students` and
+      their earlier stints in `reenrollments`, so a re-entry backdated onto a
+      still-open stint puts two rows on one entrydate and every mart joining
+      that key double-counts the student. The survivor is the year's primary
+      stint (`rn_year` 1), which is the open one because `rn_year` is ranked
+      exitdate descending upstream, and which is the row carrying
+      `year_in_school` and `year_in_network`. Deduplicating drops rows, never
+      key values, so no existing hash changes and no downstream foreign key is
+      orphaned. The matching test on `int_powerschool__student_enrollment_union`
+      stays `severity: warn` on purpose -- it keeps alerting on the underlying
+      PowerSchool data-entry error, which only Ops can correct. Tracked in
+      #5045.
     config:
       materialized: table
     data_tests:
@@ -21,6 +37,8 @@ models:
               - _dbt_source_project
               - academic_year
               - entrydate
+          config:
+            severity: error
     columns:
       - name: student_number
         description: >-
--- a/src/dbt/kipptaf/models/students/intermediate/properties/int_students__student_enrollment_union.yml
+++ b/src/dbt/kipptaf/models/students/intermediate/properties/int_students__student_enrollment_union.yml
@@ -21,8 +21,6 @@ models:
               - _dbt_source_project
               - academic_year
               - entrydate
-          config:
-            severity: warn
     columns:
       - name: student_number
         description: >-


cbini:
CI is green. dbt Cloud run `70403222572549` succeeded with 3 test warnings.
**All 3 are pre-existing, none is caused by this change.**

**`dim_student_section_enrollments__no_multi_stint_overlap` -- 147 results.**
PR #5049, which does not touch enrollment at all, reports the identical 147 on
its own run (`70403222572300`). The count is unchanged with main's enrollment
data and with the deduped enrollment data, so none of the 6 removed stints
participated in a course-enrollment overlap. Structurally it could only go down:
the test inner-joins stints, and this change only removes stint rows.

**`int_students__calendar_day__zero_enrollment_in_session_days` -- 2 results.**
Scoped to `_dbt_source_project = 'kippmiami'`. This change touches zero Miami
rows -- Miami holds at 10,078 rows before and after -- and the test warns
deliberately, per its own comment, to keep 2 Focus calendar rows visible to Ops.

**`relationships_fct_grades_category_student_section_enrollment_key` -- 2476
results.** Staging-snapshot drift, not a real orphan set. The test compares a
child in `zz_stg_kipptaf_marts` against a parent rebuilt in the PR schema. Three
measurements:

- Prod to prod, the same relationship returns **0** orphans.
- 133 `student_section_enrollment_key` values in prod are missing from the PR
  build, and **all 133** are absent from the staging
  `int_students__course_enrollments` that CI built from, while all 133 are
  present in the prod upstream. The gap is the upstream snapshot, not this
  change.
- This change cannot affect that key at all. It is
  `generate_surrogate_key([cc_dcid, _dbt_source_project])`, derived from the
  course enrollment, and the stint join in
  `dim_student_section_enrollments` is a `left join`, so removing stint rows
  removes no course-enrollment rows. The measured effect on the enrollment
  foreign key is 1 additional null (13,366 vs 13,365).

PR #5049 shows the sibling `relationships_fct_grades_term_...` warning in the
same cross-schema shape, so this family of warnings is a standing CI artifact.

cbini:
Verified all 5 findings against the conventions and the data before acting.
**3 fixed, 2 declined with reasons.** Fixes are in `1740b50`.

---

**1. Test severity not actually raised -- CONFIRMED, fixed.**

Correct, and this was a real defect: the PR claimed a behavior it did not
deliver. `src/dbt/kipptaf/dbt_project.yml:22-23` sets `data_tests: +severity:
warn`, and `src/dbt/CLAUDE.md:800` states removing the override reverts to the
project default rather than `error`. Added `config: severity: error`
explicitly.

**2. `qualify row_number() = 1` masking upstream duplicates -- CONFIRMED, fixed.**

Both cited rules are real: `src/dbt/CLAUDE.md:1077` ("No `QUALIFY`") and
`:1095-1099`, which names this exact pattern and prescribes
`dbt_utils.deduplicate()` with a `-- TODO:`. Rewritten to the macro form with
`TODO(#5045)`.

On the `stg_powerschool__cc` precedent: that rule is scoped to one named model
and to residual fan-out from a date-range join, where a dedupe would hide a
join defect. This is a different situation -- the model violates its own
declared grain, which the general rule at `:1095-1099` handles by prescribing
the macro. #5045 also names deduplicating at this model as one of its two
sanctioned fixes. So the decision to dedupe stands; the form was wrong and is
now corrected.

One note for reviewers: `src/dbt/CLAUDE.md:1001` says a CTE referenced only via
`dbt_utils.deduplicate` needs a `trunk-ignore(sqlfluff/ST03)`. It does not apply
here -- the macro renders `from unioned original`, so sqlfluff sees the
reference, and adding the suppression tripped `trunk/ignore-does-nothing`. Left
it out.

**3. Non-deterministic tie-break for the Paterson pair -- CONFIRMED, not fixed. Reasons below.**

The reviewer flagged this without warehouse access. Measured, and it is real:
that pair ties on the upstream `ORDER BY` in full, so `rn_year` 1 versus 2 is
assigned arbitrarily between the 2 rows. The other 5 collisions are untied.

Not fixed, for 2 reasons.

- **A tiebreaker here cannot fix it.** The ambiguity is in how `rn_year` itself
  is assigned upstream, in `int_powerschool__student_enrollment_union` in the
  `powerschool` package. `rn_year` is a `row_number()`, so it is always distinct
  within this model's partition -- a secondary sort key would never engage. The
  real fix is a deterministic tiebreaker in the upstream ranking, which is in a
  package shared by 4 district projects and outside this PR's kipptaf-only
  scope.
- **The flap does not reach any mart.** Whichever row wins is the `rn_year = 1`
  row, so `year_in_school` and `year_in_network` are always populated. Compared
  both rows as they land in `dim_student_enrollments`: every column is identical
  except `year_in_network`, which is populated on the `rn_year = 1` row and null
  on the other. Ordering by `rn_year asc` always picks the populated one, so the
  dim output is stable either way. The residual flap is confined to `entrycode`
  and `reenrollments_dcid` in this intermediate, on 1 student, in a record pair
  Ops will delete.

Happy to file this against the `powerschool` package as a follow-up if that is
preferred.

**4. Inline comment belongs in `description:` -- CONFIRMED, fixed.**

Rule is at `src/dbt/CLAUDE.md:1040`. `:601` makes it stronger than the finding
states: a doc-only inline comment on a heavily-consumed intermediate marks the
model `state:modified` and fans CI's rebuild across its whole descendant graph.
This model is exactly that. Moved the rationale into the properties
`description:`, which was also stale as the finding notes, and kept only the
`TODO(#5045)` inline per the carve-out.

**5. No unit test for the dedupe -- declined.**

The regression check the invariant needs already exists and now blocks:
`dbt_utils.unique_combination_of_columns` on the 4 key columns at
`severity: error`, running on every build. The survivor choice is explicit in
`order_by="rn_year asc"` and verified against prod (all 6 survivors are the
`rn_year = 1` row). A unit test here needs fixtures covering both branches of a
`full union all corresponding` across 49 columns, which is a lot of scaffolding
for a 4-line macro call. Say the word if you want one anyway.

---

Re-verified the macro form against prod after the rewrite. Identical to the
previous form: 134,336 rows, 0 duplicate keys, 0 keys lost or gained, all
survivors `rn_year = 1`, Miami unchanged at 10,078.

cbini:
CI re-passed on `1740b50` with the `dbt_utils.deduplicate` form. Run
`70403222573355`, no failed steps.

The uniqueness test now runs at `severity: error` and passed in a real build --
it does not appear in the warning list, and an error-severity failure would have
failed the run. That is the acceptance criterion for #5045 met against a full
`--full-refresh` build rather than a simulated query.

The same 3 pre-existing warnings persist at identical counts (147, 2, 2476),
unchanged from the previous run and already triaged above.

Issue #4868, opened 2026-08-14
feat(dbt): Focus branch into the kipptaf enrollment base_ vertical (restore Miami for AY2026)


cbini:
Phase 1 of #4729 restored Miami to the spine
(`int_students__student_enrollment_union`) and the 14 marts downstream of it.
It did not touch the second kipptaf enrollment vertical.

`base_powerschool__student_enrollments` still unions the frozen
`kippmiami_powerschool` archive as its Miami branch, so Miami reads empty for
AY2026 across **15 consumer sites** — and through
`int_extracts__student_enrollments`, roughly 150 more. This is the live
user-facing half of the gap: `rpt_gsheets__student_contact_info` serves 1,114
Miami students frozen at `academic_year = 2025` to four Google Sheets exposures.

Absorbs #4811. Treats #3999 as the follow-on rather than a blocker.

## Approach

Keep `base_powerschool__student_enrollments` for compatibility; build the Focus
branch into a SIS-neutral model underneath it.

1. **New `int_students__student_enrollments`** — today's `base_` body with the
   `kippmiami_powerschool` relation dropped from the union and a
   `focus_conformed` CTE joined by `full union all corresponding`. The 10
   kipptaf-level left joins are unchanged. Mirrors the shape
   `int_students__students` already uses.
2. **New `int_focus__advisory`** in the kipptaf focus layer — `homeroom` rows
   from `int_focus__schedule`, teacher name via `int_focus__users`. Analogue of
   `int_powerschool__advisory`.
3. **`base_powerschool__student_enrollments` becomes a passthrough** —
   `select * from int_students__student_enrollments`. All 15 consumer sites keep
   working untouched. This is the seam #3999 deletes.

Column vocabulary stays PowerSchool for now. The R1-R10 SIS-neutral rename is
Phase 5, where it belongs with the `base_` retirement.

## Derivations

All reproduce the PowerSchool formula against Focus columns:

| Column | Rule |
| --- | --- |
| `cohort_primary` | `(academic_year + 13) + (-1 * grade_level)` |
| `cohort_secondary` | `max(if(year_in_school = 1, cohort_primary, null)) over (student, school)` |
| `cohort` | grade 99 to `cohort_graduated`; 9+ to `cohort_secondary`; else `cohort_primary` |
| `boy_status` | grade-history `lag()` — New / Re-Enrolled / Promoted / Retained / Demoted / Graduated |
[...]

PR #5055, opened 2026-08-28
fix(dbt): derive is_out_of_district from custom_863 for Miami

### LINKED ISSUE #5041: fix(dbt): Miami is missing from a network topline metric because is_self_contained is blank

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will derive `is_out_of_district` for Miami from
`custom_863` — the IDEA educational environment field in Focus — instead of
hardcoding it to `false`. It moves the three columns that travel with that flag
at the same time.

Miami students used to report `false` because Focus had no source for the flag.
`custom_863` is that source. Its age 6-21 options each place a student at a
different institution — correctional facility, center school, public residential
facility, hospital or homebound, private school with public school services —
which is PowerSchool's definition of out-of-district exactly.

The flag cannot move alone. In PowerSchool, one "Out of District" special-program
record drives four columns together, so a student at a hospital gets an
out-of-district reporting school and an `OD` school level to match. Setting only
the boolean would produce a row claiming a student is out of district while still
naming a real KIPP Miami building. So all four now branch on the same
`custom_863` match:

| Column | PowerSchool source | Miami source |
| --- | --- | --- |
| `is_out_of_district` | the "Out of District" special program | `custom_863` code is C, D, F, H or P |
| `reporting_schoolid` | the program's `programid` | the `custom_863` option id |
| `reporting_school_name` | the program's name | the decoded option label |
| `school_level` | `OD` | `OD` |

**Nothing changes today.** Every Miami student holds `custom_863`'s `Z` default
or no value at all, so all four columns resolve to exactly what they resolved to
before. Verified row-for-row: 10078 Miami rows in both the dev build and prod, 0
differences across all four columns. The change matters the day Ops starts
filling the field in, which is the day the old literal would have gone silently
wrong.

`is_self_contained` stays null. Focus has no source for it, and the options that
would express it are absent from `custom_863`'s list — see #4968 and #5041.

## Reviewer Notes

**Why the option id, and not the letter code, in `reporting_schoolid`.** That
column is `int64` in 13 declared places, 9 of them contract-enforced
`rpt_tableau__*` extracts, so a letter cannot go there without a migration
nobody asked for. The `custom_863` option id is an int and it is the direct
analogue of what PowerSchool puts there — the special program's `programid`. A
hospital has no Florida school number either way. The letter still reaches
consumers: `reporting_school_name` ends in `[H]`.

**The decode lives in the pivot that already does the decoding.**
`int_focus__students__pivot` used to decode select fields to labels only,
dropping both the option code and the stored value. It now runs three
aggregates, so every one of its 47 fields carries `value_`, `code_` and
`label_`. BigQuery names multi-aggregate PIVOT outputs `{aggregate}_{field}`, so
the in-list aliases dropped their `_label` suffix and the prefix leads instead.
`int_focus__students` maps them back onto its established `<field>_label` names
so its own consumers are untouched, and adds
`idea_educational_environment_code`.

Two earlier revisions are reverted: one declared
`int_focus__custom_field_options` as a new kipptaf source with its own wrapper,
which put the decode a layer away from the field; the other carried the code
through PIVOT as a grouping column, which worked but decoded only `custom_863`.

**That makes this a column add on a package model kipptaf reads via
`source()`.** Per `src/dbt/CLAUDE.md` that is normally a two-PR split. Shipped
as one PR instead via the documented single-PR path: the kipptaf
`int_focus__students` wrapper's comment is extended so it is `state:modified`
and CI rebuilds it, and `zz_stg_kippmiami_focus.int_focus__students` was seeded
with `dbt build --target staging` so that rebuild sees the two new columns. The
wrapper already carried the same kind of comment from #4868.

**The age 3-5 options are deliberately unclassified.** Florida's 3-5 codes
include real institutional placements (`B` residential facility, `S` separate
school), but Miami is K-9 with no pre-K rows, and whether a 3-5 placement should
read as out-of-district is a question nobody has answered. Rather than guess,
the derivation leaves them in-district and
`test_focus__idea_educational_environment_classified` warns if one ever lands.

**The tripwire test changed purpose, and was renamed.** It used to warn on any
populated `custom_863`, because any value made the literal wrong. Now the
derivation handles the five out-of-district codes correctly, so warning on those
would cry wolf. It warns only on a code the derivation does *not* classify.

### General

- [x] Include (or update) an
      summary above.
      See

## CI checks

## Troubleshooting

Refs #5041 — this does not close it. The two Ops asks on that issue are still
open: does KIPP Miami operate any self-contained classrooms, and should
`custom_863`'s option list gain Florida's age 6-21 regular-class time bands so
`is_self_contained` gets a source at all.

<details>
<summary>For Claude</summary>

**Scope history.** This PR opened as a comment plus a warn-severity tripwire
test, leaving the `false` literal in place with a `TODO(#5041)`. @cbini asked why
the flag was not simply derived from `custom_863`, which was the right question —
the tripwire was guarding a value that could just be correct. My objection was
that the flag cannot move without its three siblings and that no design existed
for what they become. @cbini supplied it: the code as the reporting school id,
the value as the reporting school name. That unblocked the derivation and this PR
was rewritten. The `TODO` and the original tripwire framing are gone.

**Zero-output-change proof.** Dev build via
`--favor-state --defer --state src/dbt/kipptaf/target/prod --target dev`, then
compared to prod on the PK `(student_number, academic_year, schoolid, entrydate)`
for `_dbt_source_project = 'kippmiami'`:

| metric | value |
| --- | --- |
| dev rows | 10078 |
| prod rows | 10078 |
| dev rows with `is_out_of_district` true | 0 |
| dev rows with `school_level = 'OD'` | 0 |
| rows differing on any of the 4 columns | **0** |

Comparison used `format('%T\|%T\|%T\|%T', ...)` rather than `concat` so a NULL in
any column cannot silently collapse the comparison.

**Derivation-fires proof.** The four expressions run against the real options
table with mocked option values, no student data:

| stored value | code | is_out_of_district | reporting_schoolid | reporting_school_name | school_level |
| 759 | Z | false | roster value | roster value | roster value |
| null | — | false | roster value | roster value | roster value |
| 5019 | N/A | false | roster value | roster value | roster value |
| 751 | H | **true** | 751 | `Age 6-21, Hospital/Homebound [H]` | `OD` |
| 749 | D | **true** | 749 | `Age 6-21, Center School [D]` | `OD` |
| 755 | P | **true** | 755 | `Age 6-21,Private Sch with Public Sch Ser [P]` | `OD` |

**Types held.** `INFORMATION_SCHEMA` on the dev build confirms
`reporting_schoolid` INT64, `reporting_school_name` STRING, `school_level`
STRING, `is_out_of_district` BOOL — unchanged, so the 9 contracted downstream
extracts are unaffected.

**Boolean branch, not coalesce.** All four branch on
`stu.is_out_of_district`, a boolean precomputed once in the `focus_students`
CTE, because BigQuery has no lateral column aliases and the four cannot share an
alias computed beside them. `if()` is null-safe on both counts that matter — a
roster row matching no student, and a student with no `custom_863` value, both
fall to the in-district branch, which the 0-null count above confirms.
`coalesce(<label>, enr.school)` would instead have fallen back to the real
school name if a matched option ever had a null label while the flag said true;
`src/dbt/CLAUDE.md` calls out that class of bug.

An earlier revision branched on `ood.option_id is not null` against a CTE of
out-of-district options, and this section described that shape until
`claude-review` caught the drift. No `ood` alias exists in this diff.

**The `focus_students` CTE enumerates its columns rather than `select *`.** That
is what removes the `sqlfluff/AM04` violation in `powerschool_conformed`
downstream, which let the pre-existing `trunk-ignore(sqlfluff/AM04)` there be
deleted rather than left dead. The two are a pair — enumerating while the ignore
stayed made the ignore look pointless, which is why an earlier attempt was
reverted.

**Local validation.**

- `dbt build --select int_focus__students test_focus__idea_educational_environment_classified`
  via `kippmiami` — `PASS=6 WARN=0 ERROR=0`, including the model's own unit test
  and the retargeted singular test.
- `dbt build --select int_focus__students int_students__student_enrollments` via
  `kipptaf` — `PASS=8 WARN=0 ERROR=0`, including the PK
  `unique_combination_of_columns` and `not_null_is_out_of_district`.
- `dbt build --select int_focus__students --target staging` via `kippmiami` —
  `PASS=6`, and `INFORMATION_SCHEMA` confirms
  `zz_stg_kippmiami_focus.int_focus__students` now carries
  `idea_educational_environment` INT64, `idea_educational_environment_code`
  STRING and `idea_educational_environment_label` STRING.
- Decode spot-check on the rebuilt model: 3865 rows at code `Z` with the
  matching label, 190 null. No other code exists in Miami today.
- `trunk check --force` on all 8 changed files — formatting only (prettier,
  sqlfmt), no rule violations beyond the AM04 note above; the commit hook
  applied both and re-checked clean.

**The unit test needed its mocks widened.**
`test_int_focus__students_homeless_and_meal_conform` failed with
`depends on a node named 'int_focus__custom_field_options' which was not found`
— the documented consequence of a column add. Fixed by adding
`idea_educational_environment` to the `stg_focus__students` mock,
`idea_educational_environment_label` to the pivot mock, and an empty
`int_focus__custom_field_options` input. `expect` needed no change: it lists 5
of the model's ~103 columns, and dbt compares only what is listed.

**A gotcha worth recording.** The first two dev builds failed with
`Not found: zz_cbini_kipptaf_focus.int_focus__student_enrollment_roster`, and
`--favor-state` did not help. Cause: the prod manifest at
`src/dbt/kipptaf/target/prod` was 3 days stale (Aug 25) and did not contain that
node at all, so defer had nothing to redirect to. Fixed with
`dbt parse --target prod --project-dir src/dbt/kipptaf --target-path target/prod`.
A missing-relation error under `--favor-state` means a stale manifest, not a
missing dev build.

**CI expectations, and a corrected earlier claim.** An earlier revision of this
PR failed dbt Cloud with `Name exit_date not found inside e` in
`int_extracts__course_enrollments_by_term`, and this section previously blamed a
break on `main`. That was wrong, and the claim is withdrawn.
`base_powerschool__course_enrollments` is a compatibility passthrough --
`select *, from {{ ref("int_students__course_enrollments") }}` per #3999 -- so it
does forward `exit_date`, which `int_students__course_enrollments.sql:44`
produces. The failure was a selection-order artifact of this PR's own earlier
shape: the `sources-kippmiami.yml` addition marked the whole `kippmiami_focus`
source `state:modified`, and in the resulting wide graph the passthrough's
`select *` expanded against the stale staging copy of
`int_students__course_enrollments`, which predates the column. `exit_date`
exists in exactly one schema warehouse-wide -- this PR's branch schema --
confirming it is branch-built rather than missing. With the source addition
reverted the selection includes the producer, and the run is green.

**Warnings: 26, all pre-existing.** None touch `is_out_of_district`,
`custom_863`, or the four changed columns. They are survey keys, gradebook
grain, pearson student numbers, mart `student_key` relationships, and Miami
`not_null` warns on `lunchstatus` / `lep_status` / `spedlep` that predate this
branch. The count rose from 18 because this change legitimately fans out
further: widening `int_focus__students` rebuilds its kipptaf passthrough and
everything downstream.

**One lint note is expected and non-blocking.** `trunk/ignore-does-nothing` on
the pre-existing `trunk-ignore(sqlfluff/AM04)` in `powerschool_conformed`.
Adding a templated `ref()` earlier in the file shifts sqlfluff's templated-slice
map, so trunk's ignore-detector and the AM04 rule disagree about whether it
fires. Removing the ignore produces a real blocking AM04 violation, so it stays;
`trunk check --force` exits 0 in this state.

</details>


cbini:
## Review response, and a scope change

**This PR has been rewritten since the review above.** It opened as a comment
plus a tripwire test, leaving `false as is_out_of_district` in place with a
`TODO(#5041)`. @cbini asked why the flag was not simply derived from
`custom_863`, then supplied the missing design decision — the code as the
reporting school id, the value as the reporting school name — which unblocked
the derivation. The literal and the TODO are gone. The PR body and title are
updated. The review below still applies to the finding it raised, so:

### The one finding — CONFIRMED, and the PR body was wrong

> The PR's "trivially green, not validation" CI expectation is probably
> optimistic.

Right, and I was wrong. Checked against the actual run rather than the
prediction:

| | |
| --- | --- |
| dbt Cloud run | `70403222583277` |
| Step | `dbt build --select state:modified+ --full-refresh` |
| Duration | **4m 31s** (whole run 4m 59s) |
| Result | Success, 18 test warnings |

Not the ~30s no-op I predicted. The inline comment marked
`int_students__student_enrollments` as `state:modified` exactly as you said. The
"For Claude" section now says so.

Your note also under-sold that run: it *was* validation of one thing — the
comment did not break compilation of the ~30 downstream consumers under a full
refresh. It was still not validation of the focus test, which lives in a package
kipptaf CI never runs. That distinction is in the body too.

**The 18 warnings were pre-existing, proven structurally rather than eyeballed.**
Per `src/dbt/CLAUDE.md` § "Verifying a comment-only SQL change": stripping SQL
and Jinja comments from `int_students__student_enrollments.sql` and collapsing
whitespace gave **10654 characters, byte-identical** on both the merge-base
(`b994bc0ec5`) and that commit. The model emitted exactly the SQL it emitted
before, so it could not have moved a value downstream. None of the warnings
touched `is_out_of_district` or `custom_863`.

That proof no longer covers the current commit — the model now has real logic
changes. It is replaced by a row-for-row dev-vs-prod comparison: 10078 Miami
rows on both sides, 0 differences across all four changed columns. Details in
the body.

### On the "checked, no violation found" list

Not taking these on faith either — `src/dbt/CLAUDE.md` says a review agent's
dismissals need verifying as much as its flags. Spot-checked the two that
mattered:

- The inline `cast(...)` in the test's `ON` clause. Your read matches the
  codebase — `int_students__course_enrollments.sql` and
  `int_students__course_sections.sql` both do it. Agreed. Moot now anyway: the
  kipptaf derivation casts once in a CTE and joins int-to-int, so there is no
  cast in an `ON` clause there.
- Project-default `warn` severity via `kippmiami/dbt_project.yml`'s unscoped
  `data_tests: +severity: warn`. Confirmed by reading that file at line 20, and
  by the sibling `focus_unmapped_phone_contact_titles` carrying no override
  either. Agreed, and it is why the explicit `severity: warn` was dropped before
  the first commit.

### Worth a fresh look

`claude-review` fires only on `opened` and `ready_for_review`, never on
`synchronize`, so it has not seen the derivation — which is now the entire PR.
A draft toggle would re-fire it if a reviewer wants that.

src/dbt/focus/models/intermediate/int_focus__students.sql
```diff
@@ -1,4 +1,15 @@
...
+    -- custom_863's option code, which int_focus__students__pivot drops -- it
+    -- decodes to labels only. is_out_of_district in
+    -- int_students__student_enrollments keys on the code, not the label,
+    -- because Focus users can edit the labels while the codes are Florida DOE
+    -- values (#5041).
+    idea_educational_environment_options as (
+        select code, cast(option_id as int64) as option_id,
+        from {{ ref("int_focus__custom_field_options") }}
+        where column_name = 'custom_863' and source_class = 'SISStudent'
+    ),
```
cbini: why this instead of pulling from int_focus__students__pivot?

cbini: No good reason — moved it into the pivot in fa0b93b2ec.

What I was avoiding was real but I picked the wrong way around it. Adding a
second aggregate to the existing pivot switches BigQuery to
`{aggregate}_{value}` output naming, which renames all 47 existing `*_label`
columns. Verified rather than assumed:

```sql
select * from t pivot (any_value(label) as label, any_value(code) as code
                       for k in ('a' as fa, 'b' as fb))
-- columns: label_fa, code_fa, label_fb, code_fb
```

What I missed is that the code does not need to be an aggregate at all. PIVOT
groups by every column it is *not* given, so the code rides through as a
grouping column:

- `decoded` now also selects `options.code`
- `decoded_codes` filters it to `custom_863`, one row per student
- `decoded_labels` left-joins that in, and the existing pivot reads it

The 47-alias block is untouched, and `int_focus__students` goes back to a plain
passthrough of pivot columns with no options join of its own.

Verified on the rebuild: the pivot goes 48 to 49 columns, exactly
`idea_educational_environment_code` added and nothing removed or renamed; 4055
rows and 4055 distinct `student_id` on both dev and prod, so the added join does
not fan out. `int_students__student_enrollments` still matches prod row-for-row
— 10078 Miami rows both sides, 0 differences across all four changed columns.

I also tried a second single-aggregate pivot first. It worked, but it required
wrapping the label pivot in a CTE, and the resulting 4-space reindent pushed two
of the longest aliases past 88 chars (sqlfluff LT05). The grouping-column
version has no reindent and a smaller diff.

src/dbt/focus/models/intermediate/int_focus__students__pivot.sql
```diff
@@ -117,18 +117,46 @@ with
...
     decoded as (
-        select unpivoted.student_id, unpivoted.column_name, `options`.label,
+        select
+            unpivoted.student_id,
+            unpivoted.column_name,
+            `options`.code,
+            `options`.label,
         from unpivoted
         left join
             {{ ref("int_focus__custom_field_options") }} as `options`
```
cbini: don't use reserved keywords as aliases

src/dbt/focus/models/intermediate/int_focus__students__pivot.sql
```diff
@@ -117,18 +117,46 @@ with
...
+    decoded_labels as (
+        select
+            decoded.student_id,
+            decoded.column_name,
+            decoded.label,
+
+            decoded_codes.idea_educational_environment_code,
+        from decoded
+        left join decoded_codes on decoded.student_id = decoded_codes.student_id
     )
```
cbini, line 133: do the second aggregate method. we want value, code, and label

src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollments.sql
cbini: delete

src/dbt/focus/models/intermediate/int_focus__students__pivot.sql
cbini: Renamed to `opt` in fb214fde67.

That alias was pre-existing, but the diff touches the line so it is fair game.

Two reserved words survive as backticks, because in both places the identifier
determines an output column name and renaming it would rename the column:
`any_value(stored_value) as `value`` and `'custom_200000005' as `language``.
Per `src/dbt/CLAUDE.md`, the backtick satisfies RF04 without changing the
produced name — so the outputs are still `value_language`, `code_language`,
`label_language`. sqlfluff flags exactly those two and nothing else after the
alias rename.

src/dbt/focus/models/intermediate/int_focus__students__pivot.sql
cbini: Done in fb214fde67 — three aggregates, value/code/label for all 47 fields.

```sql
decoded pivot (
    any_value(stored_value) as `value`,
    any_value(code) as code,
    any_value(label) as label
    for column_name in (...)
)
```

`decoded` now carries `stored_value` alongside `code` and `label`, and the
`decoded_codes` / `decoded_labels` CTEs from the previous commit are gone.

As expected, the output names invert: BigQuery produces `{aggregate}_{field}`,
so the in-list aliases dropped their `_label` suffix and the prefix leads
instead — `label_race_white`, not `race_white_label`. Verified on the rebuild:

- 142 columns, up from 48 — `student_id` plus 47 x 3
- 4055 rows and 4055 distinct `student_id`, unchanged, so no fan-out
- `custom_863` resolves as `"759"` / `"Z"` / `"Age 0-2, Gifted only, none above
  (deflt)[Z]"` for 3865 students, all three null for the other 190

One decision worth flagging: `int_focus__students` maps them straight back onto
its existing names — `p.label_race_white as race_white_label`. I kept its output
interface rather than propagating the inversion, because 14 of those columns are
consumed inside that model to derive `ethnicity`, `homeless_code`, `lunchstatus`
and `lep_status`, and `idea_educational_environment_label` is read by
`int_students__student_enrollments`. Say the word if you would rather the
`label_` prefix carried all the way down; it is a mechanical follow-up.

The model description also dropped its "join `stg_focus__students` for the
codes" line, since `value_` now makes the pivot self-sufficient.

`int_students__student_enrollments` still matches prod row-for-row: 10078 Miami
rows both sides, 0 differences across all four changed columns.

src/dbt/kipptaf/models/students/intermediate/int_students__student_enrollments.sql
cbini: Deleted in fb214fde67, and the underlying violation is gone rather than moved.

Deleting it alone would have left a blocking `sqlfluff/AM04` at that line, which
is why I had restored it earlier. The reason it went dead in the first place is
that this branch added a `select *` in the `focus_students` CTE above;
enumerating that CTE's 14 columns removes the AM04 violation entirely, so the
ignore is genuinely unnecessary now. `trunk check --force` on the file reports
no issues with the ignore deleted.

My earlier attempt at enumerating looked useless because I tested it while the
ignore was still in place — the pairing is what matters.

cbini:
## Review response

Verified each finding against the code before acting. Two were mine and are
fixed; one is a real gap I am declining here, with a concrete reason.

### 1. Stray `>-` in 16 pivot descriptions — CONFIRMED, fixed in f59823b56

Your read is exactly right, including that it only hits `label_*` fields whose
text wraps. Root cause: I generated the widened yml with a script that harvested
each existing description via

```python
r"description:\s*(.*?)\n        config:"
```

`\s*` consumes the single space after the colon and then stops at `>`, which is
not whitespace, so the fold indicator got captured into the body for every
description written as `description: >-`. Single-line descriptions have no
indicator, which is why `value_*`, `code_*` and the short `label_*` entries are
clean. All 16 stripped; `grep -c '^          >-'` now returns 0 and
`trunk check --force` is clean on the file.

### 2. No unit test on the out-of-district branch — real gap, declining here

Agreed on the substance: nothing repeatable exercises the `true` branch, and
that is the same blind spot this PR exists to close for the old literal. But a
dbt unit test on `int_students__student_enrollments` cannot be written as the
model stands. Line 9 wraps `dbt_utils.union_relations` over three
`source()` relations, and per `src/dbt/CLAUDE.md` that macro "compiles to the
column SUPERSET from source-table `INFORMATION_SCHEMA.COLUMNS` ... it needs
persisted relations, so it cannot union a local CTE" — which is precisely what a
unit test substitutes for every mocked input. The model also refs 15 other
models, so even setting the macro aside the fixture would be large.

The honest fix is to make the Focus branch independently testable — extract
`focus_conformed` into its own model, then unit-test that — which is a
refactor of a model this PR only touches four columns of. I would rather file
that than bolt a fixture onto a model that cannot accept one. Say the word and I
will open the issue.

### 3. PR body named an `ood` alias that no longer exists — CONFIRMED, fixed

Good catch, and `grep` agrees with you: zero `ood.` additions in
`git diff origin/main...HEAD`. That paragraph described the first revision of the
derivation, which joined a CTE of out-of-district options. The shipped code
branches on `stu.is_out_of_district`, precomputed in `focus_students` because
BigQuery has no lateral column aliases. The body now describes the shipped
shape, and also records why the `focus_students` CTE enumerates its columns
instead of `select *` — that is what removes the AM04 violation and let the
pre-existing ignore be deleted rather than left dead.

### On your ST06 note

Your inference was right and I can close the loop with the tooling you lacked:
`trunk check --force` on all changed files reports no issues, so the
logicals-after-case placement is genuinely not flagged by this repo's sqlfluff
config.

PR #5056, opened 2026-08-28
fix(dbt): wire Focus gradebook and category grades into the kipptaf grades facts for Miami

### LINKED ISSUE #5010: fix(dbt): wire Focus gradebook and category grades into the kipptaf grades facts for Miami

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will put Miami's AY2026 gradebook data into
`fct_grades_assignments` and `fct_grades_category` for the first time.

Both facts held zero Miami rows. They read models that union the district
PowerSchool packages, and Miami left that package when it moved to Focus, so no
join-key change could have produced a row. Miami's gradebook now flows from
Focus instead.

Two SIS-neutral models do the work, following the `int_students__final_grades`
pattern rather than branching the `int_powerschool__*` models in place:

- `int_students__gradebook_assignments_scores`
- `int_students__category_grades`

This also removes Miami's frozen PowerSchool gradebook archive from the two
gradebook unions. Focus is Miami's gradebook now, and restoring the pre-cutover
history is an Ops data-migration question rather than something these models
should half-surface.

Finally, it adds one data test and moves another into the package where
source-quality assertions belong.

## Reviewer Notes

**New Jersey row counts move in `fct_grades_category`, on purpose.** That fact
never filtered `is_dropped_section`, unlike `fct_grades_assignments`, which
always has. When a student leaves a section PowerSchool writes a second row with
a negated `sectionid`, and `cc_abs_sectionid` is the absolute value, so the join
matched the dropped stint alongside the live one. Adding the filter removes
188,365 phantom rows. Camden then reconciles exactly 1:1 — 242,070 category rows
in, 242,070 out.

**Miami rows will have no `term_key` until Ops updates a sheet.** The
reporting-terms sheet has no Miami rows for AY2026, so the term join finds
nothing. The data lands correctly but cannot be sliced by term yet. Tracked
separately with Ops; this is not a code fix.

**`fct_grades_assignments` keys are unchanged for New Jersey.** All 23,216,217
existing keys reproduce byte-identically from the new inputs.

**Two tests ship failing on purpose**, at warn severity with documented counts,
matching `int_powerschool__student_enrollment_union_no_shared_stint_boundary`.

### General

### dbt

- [x] Include (or update) an
      external consumers; existing exposures unchanged
- [x] **Breaking change?** No column is renamed or removed. Both facts keep
      their exact output column lists. Row counts change as described above.
      with **`--target staging`** — no external sources changed. The two new
      kipptaf sources point at existing `kippmiami_focus` relations, and
      `zz_stg_kippmiami_focus` was seeded with `dbt clone --target staging`
      plus a staging build of `int_focus__gradebook_grades`.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Design spec:
`docs/superpowers/specs/2026-08-28-focus-gradebook-category-facts-design.md`.
Implementation plan: `docs/superpowers/plans/2026-08-28-focus-gradebook-category-facts.md`.

### Row-count deltas, measured against prod 2026-08-28

`fct_grades_assignments` — New Jersey byte-identical. All 23,216,217 prod keys
reproduce from the new hash inputs (`asg.assignmentsectionid`,
`asg._dbt_source_project`, `asg.students_dcid`); 0 unreproducible. Miami goes
0 to roughly 4,754 rows, PK unique.

`fct_grades_category` — 8,871,760 today to roughly 7,765,102:

| Component | Rows |
| --- | --- |
| Removed by `not is_dropped_section` | -188,365 |
| Removed by dropping Miami from the union | -921,556 |
| Added, Miami AY2026 Focus categories | +3,263 |

`rpt_tableau__gradebook_dashboard` loses 12,650 Miami rows from the union drop.
`int_powerschool__category_grades_pivot` is unaffected — it unions its own
district sources. The three `int_tableau__gradebook_audit_*` models are
`enabled: false`.

### Why `fct_grades_assignments` had zero Miami rows even before AY2026

Not by design. `int_students__student_enrollment_union` carries a NULL
`studentid` on all 10,078 Miami rows across 2018-2026, and the fact joined
`ce.cc_studentid = enr.studentid`. NULL never matches, so every Miami row was
dropped. Swapping to `student_number` (needed for Focus, and measured 1:1 for
all 3 NJ regions) repairs that join and would have admitted roughly 2,388,890
Miami archive rows. Dropping Miami from the unions is the ratified resolution.

### Focus specifics

`gradebook_grades` stores a score against student and assignment with no section
reference — confirmed against Focus's vendor ERD (`Focus DB Diagram.pdf`,
2026-02-23, page 18) and against the raw dlt table. An assignment posted to
several course periods therefore fans to one row per course period the student
is scheduled into. That fan is PowerSchool parity: `assignmentsectionid` is
already per-section, and Focus itself posts a report card grade in every section
the student sits in. `int_focus__gradebook_grades` declared `unique` on
`student_gradebook_grade_id` alone, which fails on 91 rows; the grain test now
declares `(student_gradebook_grade_id, course_period_id)` at `severity: error`.

Focus writes day boundaries in local time as UTC — a due date lands at
`03:59:59Z`, which is `23:59:59` the previous day in `America/New_York`. The
conformed model casts with `{{ var("local_timezone") }}`. The focus package
cannot do this cast; its `local_timezone` var is `UTC`.

`points = -1` is a not-yet-graded or excused sentinel on 192 rows. Left numeric
it yields a score percent of -10 and poisons category averages, so it is nulled.

Focus category titles map onto PowerSchool's `storecode_type` domain:
Formative to F, Homework to H, Work Habits to W, Summative to S. An unmapped
title falls through to itself so a new category surfaces rather than vanishing.
The Focus branch emits H/W/F/S only; the quarter-overall grade reaches the marts
through `fct_grades_term`. The PowerSchool branch still passes Q through
unchanged for the NJ regions.

Marking-period scoping is anchored to `^Q(\d+)$` inside a `type = 'quarter'`
CTE. The wider `short_name` domain (`1NW`, `FY`, `S`, `SS`, `C`, `Q4PR`, `PP11`,
`SS1`) would otherwise produce NULL or nonsense `storecode` values that collide
in the grain test.

### Known source-data observations, not introduced here

One category percent reads 550 — a single score of 55 points on a 10-point
assignment, in "Geosphere Exit Ticket" (grade id 7029913). 36 category rows
exceed 100%, mostly 12/10 extra credit. The aggregation math is correct.

40 Newark category rows match no live enrollment at all and drop out with the
dropped-section filter. Pre-existing, unrelated to this change.

### Tests

`int_focus__schedule__no_overlapping_course_periods` (new, focus package) —
2 failing rows. Resolves each schedule row through its course period's marking
period, because `end_date` is null on 18,412 of 19,699 rows; without that, all
188 semester-paired course periods read as concurrent. Scoped to a shared
`period_id`; the same course in different class periods is a separate 776-group
pattern awaiting an Ops ruling.

`base_powerschool__course_enrollments__no_studyear_course_overlap` — moved from
kipptaf to the powerschool package. 10,807 failing rows: kippnewark 7,901,
kippcamden 2,783, kipppaterson 123. `kippmiami` does not consume the powerschool
package, so Miami is not covered.

### Validation

Full weight: grain uniqueness on both conformed models and both facts; 0-orphan
FK joins; NJ parity by `count(*)` and distinct key counts per region; the
byte-identical key check above; `dbt build --empty` across 14 descendant nodes;
`dbt test` 7/7 on both facts.

Directional until Miami Q1 closes 2026-10-16: every magnitude comparison of
Miami against NJ. Miami Q1 runs 2026-08-12 to 2026-10-16, so the gradebook holds
about 2 weeks of a 9-week term.

Local dbt builds write to a stale dev schema and prove compilation only. Every
value above was verified by querying prod directly.

Refs #5010


cbini:
## Review responses

Verdict per finding, with what I checked.

### 1. `rpt_tableau__student_course_grades` loses Miami history — declined, no effect

The mechanism is real. 198,352 Miami rows do currently pass that CTE's filters
(`yearid >= 35`, `not is_dropped_section`, `storecode_type not in ('Q')`,
`termbin_start_date <= current_date`), so removing Miami from the union does
remove them from the CTE.

It changes nothing downstream, because that extract emits no Miami rows today
and structurally cannot:

- Prod `rpt_tableau__student_course_grades` has exactly 3 `_dbt_source_relation`
  values — kippcamden, kippnewark, kipppaterson. Zero Miami rows, in any year.
- `student_roster` INNER JOINs the `term` CTE on
  `(schoolid, yearid, _dbt_source_project)`. Reproducing that join against prod
  returns **0 Miami rows** — Miami's enrollment `schoolid` values do not match
  `int_students__terms`. Camden 78,612, Newark 252,211, Paterson 5,804, Miami 0.
- The final select is `from student_roster` with the category CTE LEFT JOINed
  onto it, so category rows only surface where a roster row exists. With no
  Miami roster rows there is nothing for Miami category rows to attach to.
- Every window function in the flagged CTE partitions by
  `_dbt_source_relation`, so Miami rows cannot influence New Jersey values
  either.

So there is no AY2025 Miami data in this extract to lose. Repointing it to
`int_students__category_grades` would not restore Miami — the roster join is
what excludes it, and that is a separate pre-existing condition unrelated to
this PR.

Worth noting the contrast you drew with `rpt_tableau__gradebook_gpa` was right
for a different reason than stated: both extracts are Miami-free, but that one
because of its year filter and this one because of the roster join.

### 2. Possible storecode collision on the Focus branch — refuted

Ran your suggested check against prod. **0 collision groups.** No
`(cc_dcid, category_code, quarter_number)` triple resolves to more than one
marking period.

That is expected from the model's shape: the marking-period CTE is scoped to
`type = 'quarter'` and anchored to `^Q(\d+)$`, so two same-quarter-number
periods would have to be two distinct `type = 'quarter'` periods both named
exactly `Q<n>` for one school year. The grain test passing was not the only
evidence — this is the direct check.

### 3. Stale source-specific column prose — fixed

Fixed in df66a044bb. `name`, `is_missing` and `is_late` no longer say "in
PowerSchool". `is_missing` now also records that it is NULL on Focus rows,
which matters because a null there is genuinely unknown rather than false.

You are right that this was not a new deviation — `fct_grades_term.yml` has the
same pattern. I fixed it here anyway because this PR is what made the fact
multi-SIS, so the prose was newly wrong rather than merely inherited. Left
`fct_grades_term.yml` alone as out of scope.

### 4. One-sided calculation in a join predicate — no change

Agreed, and your own precedent settles it: `dim_student_section_enrollments.sql`
does the identical inline `coalesce(cc_dateleft, '9999-12-31')` in an ON clause.
Hoisting it here would make this model inconsistent with its sibling for no
behavioural gain. The coalesce is load-bearing — `cc_dateleft` is null on 18,582
of 19,398 Miami AY2026 course enrollments and on zero NJ rows, so without it
`duedate < null` is NULL and nearly every Miami row drops.

### 5, 6, 8. Ponytail, model altitude, Dagster — no action

Agreed on all three, nothing to add.

### 7. Hash-identity claim — verified

Your architectural reasoning is right, and the data claim checks out. Every one
of the 23,216,217 prod `fct_grades_assignments` keys reproduces from the new
inputs (`asg.assignmentsectionid`, `asg._dbt_source_project`,
`asg.students_dcid`); 0 unreproducible.

### 9. CI status — all green

dbt Cloud passed. 24 test warnings, none on this PR's models — both facts and
both new conformed models are absent from the warning list. The two warnings on
models consuming `int_powerschool__category_grades`
(`rpt_tableau__student_course_grades` 109, `rpt_tableau__gradebook_gpa` 3,687)
reproduce byte-for-byte against prod, so they are pre-existing. Trunk, CodeQL
and all three Analyze jobs pass.

Issue #5015, opened 2026-08-26
feat(dbt): decide whether we need the NJ 45-day chronic absenteeism figure


cristinabaldor:
## What and why

New Jersey excludes students with fewer than 45 days in membership at a school from its state chronic absenteeism calculations. We apply no minimum-days threshold anywhere today.

We are standardizing on 10 days first, in #4994. That is the KIPP Foundation rule, the Florida rule that covers Miami, and what New Jersey itself reports to the federal government. The 45 day threshold is a different number for a different audience: the New Jersey School Performance Report.

Sized on AY2025: 418 enrollments have fewer than 45 membership days, and 266 of those are chronically absent. Moving from the 10 day basis to the 45 day basis takes the rate from 26.2 percent to 25.5 percent.

## Proposed approach

Answer the cheap question first. NJDOE computes the School Performance Report rate from the Cumulative Days in Membership and Cumulative Days Present that we submit through NJ SMART. If the state produces that figure itself, we do not need to reproduce it, and this issue closes without code.

Only if school or regional leaders need to see the state-accountability figure in our own dashboards does this become work.

If it does become work, parameterize the minimum-days threshold rather than hard-coding a second constant. `int_students__attendance_daily` already branches truancy by state, so a state-specific attendance threshold has precedent in that file.

Refs #4994.

<details>
<summary>For Claude</summary>

Source: [NJ Chronic Absenteeism Guidance](https://nj.gov/education/safety/sandp/attendance/docs/ChronicAbsenteeismGuidance.pdf). Exact language: "students with fewer than 45 days in membership at a school are not included as part of any accountability calculations nor are they factored into any attendance items in the New Jersey School Performance Report. The NJDOE is required to report attendance data to the USED through EDFacts reporting which does include students with 10 or more days in membership."

Files in scope if this becomes work:

- `src/dbt/kipptaf/models/students/intermediate/int_students__attendance_daily.sql` — the state-branching truancy logic near line 410 is the pattern to follow. Miami uses the Florida statute of 15 absences in 90 days; the NJ districts use their own rule.
- `src/dbt/kipptaf/models/marts/facts/fct_student_attendance_daily.sql` — where `is_chronically_absent` is stamped.

Both thresholds are per school, not per district. `n_membership_student_year` in the intermediate model partitions by `academic_year` and `student_number` only, so it sums a student's days across schools and cannot serve either threshold without `schoolid` added to the partition.

[...]

Issue #4744, opened 2026-08-05
fix(cube): avg_daily_attendance counts null-attendance days as absences


anthonygwalters:
## Problem

`student_attendance.avg_daily_attendance` divides `SUM(attendance_value)` by `SUM(membership_value)` without excluding rows where `attendance_value IS NULL`. Those rows still carry `membership_value = 1`, so they land in the denominator with nothing in the numerator — silently counted as absences.

`src/cube/model/cubes/student_attendance/student_attendance.yml` lines 221-238:

```yaml
- name: _sum_attendance_value
  sql: attendance_value
  type: sum
  public: false

- name: _sum_membership_value
  sql: membership_value
  type: sum
  public: false

- name: avg_daily_attendance
  sql: "{_sum_attendance_value} / NULLIF({_sum_membership_value}, 0)"
```

`SUM` skips NULLs in the numerator; the denominator counts every membership day regardless.

## Divergence from the Tableau definition

`rpt_tableau__attendance_dashboard` filters `ad.attendancevalue is not null` alongside `ad.membershipvalue = 1`, so the dashboard excludes these days. Cube and the dashboard report different ADA for the same population, and Cube is the one that is wrong.

## Evidence

KIPP Miami, Royalty, AY2023-24:

| metric | value |
| --- | --- |
| total rows | 135,861 |
| rows with null `attendance_value` | 10,708 (7.9%) |
| rows with partial membership | 0 |
| Cube `avg_daily_attendance` | 83.7% |
| dashboard method (nulls excluded) | 90.9% |

A 7.2 point gap, entirely attributable to null handling — partial membership is zero here, so the `membership_value = 1` filter is not involved. Miami region-level for the same year reads 86.4% via Cube against 90.9% dashboard-method.

Surfaced while answering Zendesk 442525, a historical ADA request. The Cube figures produced an apparent 8 point attendance collapse at one school that vanished once null days were excluded — a finding that would have been reported to a school leader as real.

## Proposed fix

Filter the denominator to days with a recorded attendance value. The pattern already exists in this file: `_sum_tardy` filters `{CUBE}.membership_value = 1`, `count_students` filters `{CUBE}.membership_value > 0`.

```yaml
- name: _sum_membership_value
  sql: membership_value
  type: sum
  public: false
  filters:
    - sql: "{CUBE}.attendance_value IS NOT NULL"
```

## Also worth checking

[...]

PR #5057, opened 2026-08-28
refactor(cube): resolve chronic absence once in dbt and retire the anchor injection

### LINKED ISSUE #4994: feat(cube): probe whether Tesseract multi-stage can collapse the attendance snapshot measures
## What and why

Chronic absence logic is spread across 3 layers. Two of them exist only to work around the old Cube query planner.

- `fct_student_attendance_daily` computes a running average daily attendance (ADA) window, then stamps `is_chronically_absent`, `ada_tier`, and 6 period-end anchor flags onto every daily row.
- `student_attendance.yml` defines the same 10 measures 4 times: a base family, plus `_year_end`, `_month_end`, and `_week_end` copies. That is 30 of its 48 measures.
- `cube.js` spends 165 lines choosing the right anchor flag from the requested measures and granularity, then throws when the two do not match.

The duplication exists for one reason. A daily-stamped cumulative flag overcounts under `count_distinct` across a date range, and the old planner could not express "the value at period end."

Cube 1.7.0 made Tesseract the default engine, which adds multi-stage measures. We run 1.7.14. Our Cube representative recommends moving logic out of `cube.js`.

## Proposed approach

Probe before refactoring. Build a throwaway cube next to `student_attendance`, express chronic absence the Tesseract way, and compare it against today's measures on real data. Decide the refactor scope from what the probe finds.

The probe answers 5 questions, ordered so the earliest failure stops the work:

1. Does the academic year work as a grain? The KTAF academic year starts July 1, so a calendar-year window is wrong. This needs a custom granularity with an `origin`.
2. Do school weeks work? `dates_school_week_start_date` follows the PowerSchool calendar, not a fixed 7-day interval. A custom granularity may not express it.
3. Do the numbers match today's output exactly, at both year grain and month grain?
4. How long does each query take against the 12.8M rows in `fct_student_attendance_daily`?
5. Can a pre-aggregation serve a multi-stage `count_distinct`? This decides whether the approach survives in production.

Nothing existing changes during the probe. The throwaway cube gets deleted either way.

<details>
<summary>For Claude</summary>

Files in scope:

- `src/cube/model/cubes/student_attendance/student_attendance.yml` — 48 measures, lines 213 onward. The `_year_end` / `_month_end` / `_week_end` families are 30 of them.
- `src/cube/cube.js` — snapshot constants at lines 178 to 215, `queryRewrite` at lines 448 to 573. The other 440 lines are authentication, row-level security, and emulation. None of that moves.
- `src/dbt/kipptaf/models/marts/facts/fct_student_attendance_daily.sql` — `_running_ada` at lines 77 to 81, the stamped flags at lines 116 to 128.
- `src/cube/model/cubes/conformed/dates.yml` — where a custom academic-year granularity would go.

Two candidate shapes to test, not one:

- Recompute the year-to-date ADA in Cube from the additive `_sum_attendance_value` and `_sum_membership_value` measures, using `grain.include` on the enrollment key plus a `to_date` rolling window. Biggest cleanup, but it moves a window function from dbt into Cube.
- Keep the dbt window and use a multi-stage measure only to select each period's last row, replacing the pre-stamped anchor flags.

Constraints:

- `src/cube/CLAUDE.md` says transformation lives in dbt, and a rolling window inside Cube contradicts that. Flag it as a convention decision for a human. Do not settle it silently.
- Do not reuse the branch referenced from the closed #4214. Its premises did not hold. Verify every mechanic from scratch against real query output.
- The attendance cube reads a prod mart. Never copy a dev table into `kipptaf_marts` to unblock testing. Use the documented dev-schema redirect.

Done when the 5 questions above have answers backed by query output, each labeled as verified or inferred, and the findings recommend a scope.

</details>

### LINKED ISSUE #5060: fix(dbt): fct_student_attendance_daily still publishes the pre-fix chronic-absence definition
## What's happening

`fct_student_attendance_daily` still computes chronic absence and ADA tier using
the pre-2026 definition. `fct_student_attendance_periods` now computes the
network definition. The two disagree for any student at exactly 90.0% ADA.

Three separate divergences, all in
`src/dbt/kipptaf/models/marts/facts/fct_student_attendance_daily.sql:116-128`:

| what | daily fact today | network definition |
| --- | --- | --- |
| `is_chronically_absent` | `_running_ada < 0.90` — exactly 90.0% is **not** chronically absent | at or below 90.0% **is** chronically absent |
| `ada_tier` Tier 2 lower bound | `_running_ada >= 0.90` — **includes** exactly 90.0% | **strictly above** 90.0%; exactly 90.0% is Tier 3 |
| how it is evaluated | on `_running_ada`, an averaged float | on accumulated integer counts, cross-multiplied |

The third row is a stated constraint, not a preference. The design spec requires
the comparison to run on accumulated integer counts and never on an averaged
float, because a float average puts boundary students on the wrong side of the
line. `fct_student_attendance_periods` compares
`n_present_days_ytd * 10 > n_membership_days_ytd * 9` for exactly that reason.

A fourth, smaller point: on the daily fact `is_chronically_absent` and
`ada_tier` are two independent expressions. The network definition derives the
flag from the tier so they cannot drift apart. The daily fact's two happen to
agree with each other today — both place exactly 90.0% on the non-chronic side —
so this is internally consistent, just consistently wrong.

## Why it matters, and why it is not urgent

Nothing reads these two columns any more. Checked after #5057:

- No dbt model refs `fct_student_attendance_daily`.
- Every Cube reference to `is_chronically_absent` and `ada_tier` now points at
  the `student_attendance_periods` cube. #5057 deleted the 30 measures that used
  to read the daily fact's versions.

The table itself is still consumed, through the `cube` exposure — Cube reads its
`attendance_value`, `membership_value`, `is_truant`, and the `is_*_record`
columns. Only the two chronic-absence columns are orphaned.

So the risk is not a wrong number on a dashboard today. The risk is that two
marts in the same layer publish contradictory answers to the same question, and
the wrong one is reachable by anyone querying the warehouse directly.

## What to decide

Pick one. Both are breaking changes to a contracted mart, so both need
coordination and a rollback plan.

1. **Delete both columns** from `fct_student_attendance_daily`. Cleanest —
   nothing reads them, and it removes the second answer entirely. Callers who
   want chronic absence go to `fct_student_attendance_periods`, which is the
   point of #4994.
1. **Fix them in place** — move to integer-count comparison, start Tier 2
   strictly above 90.0%, and derive `is_chronically_absent` from `ada_tier`.
   Keeps a same-grain daily flag available, at the cost of maintaining the
   definition in two models, which is the duplication #4994 set out to remove.

Option 1 is my recommendation, but it is a data-team call, not mine.

Either way, confirm first that no Tableau workbook or ad-hoc query reads
`kipptaf_marts.fct_student_attendance_daily.is_chronically_absent` or
`.ada_tier` directly. The dbt exposure only records that Cube consumes the
table, not which columns.

## Steps to reproduce

1. Read the two expressions:

   ```bash
   sed -n '116,128p' src/dbt/kipptaf/models/marts/facts/fct_student_attendance_daily.sql
   sed -n '130,141p' src/dbt/kipptaf/models/marts/facts/fct_student_attendance_periods.sql
   ```

1. Count the students the boundary disagrees about, once
   `fct_student_attendance_periods` exists in prod:

   ```sql
   select count(*) as n_disagreements
   from `teamster-332318.kipptaf_marts.fct_student_attendance_periods`
   where period_type = 'year'
       and is_ca_eligible
       and n_present_days_ytd * 10 = n_membership_days_ytd * 9
   ```

   Every row this returns is a student the daily fact calls not chronically
   absent and the periods fact calls chronically absent. The AY2025 count on the
   development build was 198.

## Where

- **Code location / dbt project:** `kipptaf`, models
  `fct_student_attendance_daily` and `fct_student_attendance_periods`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** found by the automated review on
  #5057. Not caused by that pull request — #5057 modifies this model's config
  only, from view to table, and does not touch its SQL.

<details>
<summary>For Claude</summary>

The descriptions in
`src/dbt/kipptaf/models/marts/facts/properties/fct_student_attendance_daily.yml`
were corrected on #5057 to describe what the SQL actually does, plus a pointer
here. Do **not** read those descriptions as the network definition — they
deliberately document the divergence rather than hide it. The authoritative
statement lives on `fct_student_attendance_periods`'s `ada_tier` column.

The 198 figure is the AY2025 count of enrollments at exactly 90.0% ADA, measured
on the development build. It is the same 198 that #5057's reviewer notes flag as
the population affected by the KIPP Foundation tier-boundary question — their
criteria item 1 calls Tier 1 and Tier 2 on track at 90% and above, while item 8
defines chronic absence as at or below 90.0%. #5057 resolved that contradiction
in favour of item 8. If KIPP Foundation later rules the other way, option 2
above becomes the wrong fix and option 1 is still correct.

Not in scope here: `rpt_tableau__okrts_referrals.sql:297` computes its own
`is_chronically_absent` as `unweighted_ada <= 0.90`. That operator matches the
network definition, and it is scoped to New Jersey SSDS reporting periods at a
different grain, off `int_extracts__student_enrollments_weeks`. Verified as
legitimately distinct — leave it alone.

Refs #4994.

</details>

### LINKED ISSUE #5103: fix(dbt): truancy carry-forward resets on a school transfer, the flag it carries does not
## What's happening

`fct_student_days` carries `is_truant` forward across days with no recorded
attendance. A break day answers with the value from the last day that recorded
any. The carry runs per student, per school, per academic year.

The flag it carries is not computed per school. `int_students__attendance_daily`
computes `is_truant` per student per academic year, with no school in the
partition, so a student's absences follow them across a transfer.

The two grains disagree. A student who changes schools mid-year can read
`is_truant` as false on their first days at the new school while the upstream
flag says true. The carry has no prior value at the new school, so there is
nothing to carry until that school records an attendance day.

## Why the grains differ

Every other cumulative column on `fct_student_days` is per school:
`n_membership_days_ytd`, `n_present_days_ytd`, `is_ca_eligible`, `ada_tier`,
`is_chronically_absent`. A rate reported at any other grain is a different
number, and a transferring student should start a fresh chronic-absence count at
their new school.

`is_truant` took the same partition so that one row carries one grain. That
choice is consistent within the row and inconsistent with how the flag is
computed upstream.

## How many students this can affect

Student-years with more than one school inside a single source project:

| academic year | student-years | with a transfer | share |
| --- | --: | --: | --: |
| 2022 | 9,750 | 40 | 0.41% |
| 2023 | 10,620 | 62 | 0.58% |
| 2024 | 10,849 | 47 | 0.43% |
| 2025 | 11,277 | 29 | 0.26% |
| 2026 | 12,233 | 19 | 0.16% |

Only the subset of those who are also flagged truant is affected, so the real
population is smaller again. A move between districts changes
`_dbt_source_project`, which already resets every cumulative column, so this is
strictly an in-district transfer.

## What to decide

1. **Leave the carry per school.** A transfer resets it. Consistent with every
   other cumulative column on the row, inconsistent with the upstream flag. This
   is what shipped.
2. **Carry per student and academic year, ignoring school.** Matches the upstream
   computation. Puts a network-grain value on a row whose other cumulative
   columns are per school.
3. **Make the upstream computation per school.** The only option that removes the
   mismatch instead of picking a side. Also the largest: it changes `is_truant`
   for every consumer, not just Cube.

Before option 3, note that `int_topline__truancy_weekly`,
`int_topline__student_metrics` and `rpt_tableau__attendance_dashboard` all read
`is_truant` from `int_students__attendance_daily`. A grain change there is a
network-wide change.

## Steps to reproduce

Once `fct_student_days` is built, list the affected rows:

```sql
select
    student_number,
    academic_year,
    date_key,
    location_key,
    is_truant,
from kipptaf_marts.fct_student_days
qualify
    count(distinct location_key) over (
        partition by student_number, _dbt_source_project, academic_year
    ) > 1
order by student_number, date_key
```

Compare each row's `is_truant` against the same student-day in
`kipptaf_students.int_students__attendance_daily`. Any row where the fact says
false and the intermediate says true is an instance.

## Where

- **Code location / dbt project:** `kipptaf`, models `fct_student_days` and
  `int_students__attendance_daily`
- **Environment:** not in prod yet. Introduced on #5057, which is where the
  carry-forward was added.
- **Run, PR, or dashboard link (if any):** #5057

For Claude

The carry is a `last_value(is_truant ignore nulls)` window in
`fct_student_days.sql`, partitioned by
`student_number, _dbt_source_project, location_key, academic_year` and ordered by
`calendardate` over `rows between unbounded preceding and current row`. Removing
`location_key` from that partition implements option 2 and is a one-line change.

The upstream expression is the `case` at the end of
`int_students__attendance_daily.sql` that reads `n_absent_running_90` for
kippmiami and `pct_absent_running_student_year` times
`n_membership_student_year` for the NJ regions. Both of those running calcs
partition on `student_number, academic_year` only. Option 3 means adding
`schoolid` to those partitions, which changes the flag for every consumer.

Not the same issue as the truancy definition itself. That the NJ branch projects
full-year absences rather than counting cumulative unexcused ones, and that
neither region counts unexcused absences specifically, is separate and larger.
This issue is only about the grain mismatch between the carry and the flag.

Fixed means the carry and the flag agree on whether school matters, and the
reproduce query above returns zero rows.

Refs #5057

cristinabaldor:
# Pull Request

## Summary & Motivation

This pull request separates two kinds of calculation that had been tangled
together in the Cube layer, and moves both out of the query path.

**Day-scoped** questions are additive over days — what share of day-slots were
attended, how many absences, how many tardies. They sum correctly over any date
range. **Student-scoped** questions are not: has this student crossed a
cumulative threshold as of a point in time. Chronic absence, the ADA tier and
truancy are all student-scoped. Answering one means accumulating per student from
the start of the academic year and then picking a single row per period, which is
a multi-stage calculation rather than an aggregate.

The Cube layer was answering student-scoped questions with a 185-line
`queryRewrite` hook that injected a "period end" filter into every snapshot query
at run time, plus 30 duplicated `_year_end` / `_month_end` / `_week_end` measure
variants. We set out to replace that with Cube's Tesseract multi-stage features.

**Tesseract did not hold up for this.** Two of six approaches worked. Multi-stage
`rank` — the one that expresses period-end selection directly — produced correct
SQL and then timed out past 150s, and stayed slow scoped to a single month, which
is what told us the cost was structural rather than volume. Most of the rest
failed silently rather than erroring. The full findings are in #5104, written up
for our Cube reps.

**The performance floor we would have shipped into.** Our own legacy anchor path
measured 38.4s at year grain over three academic years and 51.6s at week grain
over one. The Cube MCP server's poll deadline is 55 seconds. We were one slow
week away from that breaking regardless of Tesseract.

**So both grains are materialized as dbt tables and Cube computes nothing.**
`fct_student_attendance_enrollment_daily` carries the day-scoped record, and on every row the student's
cumulative position. `fct_student_attendance_enrollment_periods` reads that fact at period end and
derives nothing. The daily fact also moved from a view to a monthly-partitioned
table — Cube queries went from 10.1s to 2 to 4s with no figure moving.

**Enrollment and attendance are unified onto one set of cubes.** They were
published as two Cube surfaces over one table and could disagree. One cube and
one view now carry both, so a single query puts a population count next to a
rate from the same rows, at the same grain. Before this, `count_students` lived
on a separate enrollment surface, so showing a denominator beside a rate meant
two queries and a hope that they agreed.

**What comes out.** The entire `queryRewrite` hook: all 175 lines of `cube.js`,
with zero inserted lines, so the authentication path provably did not change. 44
Cube members. The `student_enrollments` cube and its view. Three anchor columns
and the three tests that guarded them. Nothing injects anything at run time any
more — the row already is the period-end value.

## What the methodology review changed

The 2026-09-02 review with @anthonygwalters settled four questions that had been
open in the first draft of this work. All four are shipped. The measured
comparison against Topline lives in
`docs/superpowers/attendance-methodology/2026-09-01-methodology-review-agenda.md`.

**1. No minimum-day eligibility floor.** The draft required 10 cumulative
membership days before reporting a student. That halved the reported population
three weeks into an academic year — 11,241 students down to 5,062 — while moving
the rate a quarter of a point, so every conversation it produced was about the
denominator rather than about attendance. The floor is gone, `is_ca_eligible` is
deleted from both facts, and `count_students_min_10_days` with it. The 1-9 day
band it used to discard is 15,609 students carrying 4,544 chronic-absence cases
at AY2026 month grain, all now counted on both sides.

**2. Chronic absence and truancy accumulate per student, not per school.** The
cumulative windows partitioned by `location_key`, so a student who changed
schools mid-year restarted at zero and read as on track however many days they
had missed at the previous school. `location_key` is out of all three windows.
Rates still aggregate by school: a school is answerable for the days a student
was there, while the student is one student. Affects 11 to 46 student-years a
year, 0.1% to 0.4%, and it **closes #5103** rather than documenting it.

The counting had to follow. Every period measure counted `student_period_key`,
one row per student per school, so a student chronically absent at two schools
counted twice at network grain — 47 phantom students in the AY2024 headcount. All
five now count distinct students, matching the daily cube.

**3. A period headcount anchors on the first membership day.**
`fct_student_attendance_enrollment_periods` gains `period_start_membership_date_key`, the student's own
earliest membership day in the period. The calendar period start is the wrong
date where a year opens mid-week: anchored on the week's Monday, the week of
2026-08-10 reports 0 students when Miami served 1,514, and the week of
2026-08-17 reports the whole New Jersey roster two days before New Jersey's first
student day.

**4. `count_students` counts students with recorded attendance.** It counted
every enrolled row while every numerator required a resolvable tier, so the two
answered different questions and a rate could be deflated by students it
structurally could not reach — 3.7 points on AY2025, 6.3 on AY2023. Filtering it
on `n_membership_days_ytd > 0` makes the population and the denominator the same
number by construction. This is a floor of zero, not ten: one measured day is
enough, and a student with one day still carries a tier and can still be
chronically absent.

Day-weighted measures were never exposed to this. `avg_daily_attendance`,
`pct_tardy` and `pct_ontime` divide day sums by day sums and their helpers
already filter `membership_value = 1 AND attendance_value IS NOT NULL`, so an
unmeasured day contributes zero to both sides (#4744).

**And one defect the review surfaced.** `is_truant` arrived from the attendance
row and survived a null `attendance_value`, a break day, and a pre-AY2021
`membership_value = 0` row, so 246,583 daily rows published a truancy verdict
carrying no ADA tier — 105,679 of them reading truant. Part of the truancy
numerator therefore sat outside any tier-scoped denominator. Both flags are now
gated on the cumulative count, so they resolve over identical rows: zero
divergence in both directions, both null sets at 8,966,989 of 29,590,677 rows. An
error-severity `expression_is_true` on each fact pins it, and it failed at 224,081
rows against the pre-gate relation, so it is not vacuous. Cost: 105,679 daily
rows and 4,367 period rows stop reading truant, 1.72% of the period numerator.

**And the question the review left open is now decided.** Truancy reports the
status on any day of the period — Walters' call, implemented here. `fct_student_attendance_enrollment_periods.is_truant`
aggregates with `logical_or` over the period's membership days instead of
reading the period-end row, which is Topline's reading. Break days are excluded:
the daily fact carries the status across them, so a period opening on a holiday
would otherwise inherit the previous period's verdict. Every other value on the
fact still reads period end, and `test_periods_truancy_reads_any_membership_day`
pins both halves. Section 2.2 of the memo carries the cost: period-end would
have dropped 18.0% of the any-day count over a settled year, and 509 students a
year would have carried a flag under Topline that this fact never showed.

## What is more accurate than before

Each was checked against `origin/main` rather than asserted.

1. **The two sites that computed chronic absence wrongly become one that
   computes it correctly.** On `main`, `fct_student_attendance_daily.sql` and the
   `student_attendance` cube both derived the ADA tier from `_running_ada < 0.90`
   on an averaged float. Both are gone. `rpt_tableau__okrts_referrals.sql`
   already matched the network definition (`unweighted_ada <= 0.90`, at NJ SSDS
   reporting-period grain) and is deliberately left alone.
2. **Exactly 90.0% ADA is now chronically absent.** It was not: `main` reads
   `_running_ada < 0.90` with Tier 2 starting at `>= 0.90`. This follows KIPP
   Foundation criteria item 8 and both state criteria, and moves 198 AY2025
   enrollments. See the caveat below — item 1 of the same criteria reads the
   other way.
3. **Enrollment and attendance cannot disagree.** They were two Cube surfaces
   over one table. They are now one cube and one view, reading one fact whose
   rows come from the enrollment spine.
4. **Row existence is decided by enrollment, not attendance.** The old Cube
   surface was keyed on attendance rows, so a region with no attendance reported
   no enrollment. Rows now exist for every calendar day a student was enrolled.
   A Cube *headcount* still requires a measured day, per review decision 4.
   #5114 restored Miami's pre-AY2026 attendance archive on 2026-09-04, so those
   years carry Miami on these surfaces, and the row-existence gap that used to
   hide it is fixed.
5. **`count_students` counts students.** On `main` the attendance cube counted
   `student_enrollment_key` — the enrollment *stint* key — so a student who
   re-enrolled at the same school counted twice.
6. **A point-in-time count resolves on any calendar date.** Break days carry a
   row, so a weekend or a holiday answers the same as a school day. That retires
   the twelve anchor-column references on `main`, three tests, and the hook that
   read them — pin a date instead.

And three new capabilities:

- **ADA, chronic absence and truancy for one specific date, in a single query on
  a single view.** Before this it took two queries against two views, and
  chronic absence could not be had for an arbitrary date at all.
- **A population count beside any rate, from the same query.** `count_students`
  comes back alongside the rate it scopes rather than from a separate surface.
- **Chronic absence and truancy by subgroup, at period end.** Race, gender,
  gifted status, grade level, ELL, IEP, special education, meal eligibility,
  homeroom teacher and term all cut the period rates now, not only the daily
  ones. That is the equity analysis the dashboard is for.

### `int_students__enrollment_days` — new intermediate

Row existence has to be decided by enrollment, not attendance. Reusing
`int_students__attendance_daily` fails three ways, worst first:

1. **Whole regions and years disappear.** It returns zero rows for Miami AY2025.
   The enrollment spine has 1,346 Miami students on 2025-10-06.
2. **No break days, so no carry-forward.** 180 to 181 distinct dates per
   academic year against ~365 calendar days, and ~168 rows per student against
   180 available dates — the gaps are not only weekends. Carrying enrollment
   across non-membership days is what retires the anchor columns and lets any
   date resolve, and it needs a row on a Saturday.
3. **No school in the grain.** Unique on
   `(student_number, _dbt_source_project, calendardate)`. The period fact keys
   its grain per school, so it needs a school on every day.

It stays a model rather than becoming CTEs because it is a materialization
boundary. Its own comment records that the range join against `dim_dates`
requires the `date_key <= current_date` filter or BigQuery refuses the query on
the CPU-to-bytes ratio. Inlining would put that fan-out — 111,739 stints across
180 to 365 days each — the attendance join, and cumulative windows over 29.6M
rows in one statement.

### `fct_student_attendance_enrollment_daily` — restructured, renamed from `fct_student_attendance_daily`

One fact instead of two Cube surfaces over one table, and it now carries the
cumulative family so the tier ladder lives in one place. The windows partition by
`(student_number, _dbt_source_project, academic_year)` — deliberately not by
location, per review decision 2.

It became a table (Cube queries ran 10.1s, now 2 to 4s, no figure moved) with a
monthly partition. Monthly rather than daily because 7,058 distinct dates are
already in it and it gains ~180 a year, so daily partitioning would pass
BigQuery's 4,000-partition cap inside a decade. Filtering `attendance_date`
prunes it: 63 MiB and 0.7 slot-seconds against 1,257 MiB and 24 to 82
slot-seconds via the `dates` join, identical rows.

### `fct_student_attendance_enrollment_periods` — new fact

Period-end values are pre-computed rather than computed at query time because
query time does not scale. Measured on the real view, cold, with row-level
security and joins, no pre-aggregation:

| approach | time |
| --- | --: |
| multi-stage `rank` | timed out past 150s |
| multi-stage `grain.include`, no window | 14.9s to 30.6s |
| legacy `_year_end`, three academic years | 38.4s |
| legacy `_week_end`, one academic year | 51.6s |
| plain additive aggregate by academic year | 14.3s |
| **pre-computed rows** | **2 to 4s** |

The Cube MCP server's poll deadline is 55s, so the legacy anchor path sat inside
it. Also rejected: `rolling_window` with `trailing: unbounded` emitted zero
window functions and silently discarded the query's lower bound; custom
granularity via `origin` was silently ignored, and via `offset: -6 months` works
but costs 58.7 slot-seconds against the `academic_year` dimension's 31.1 for an
identical answer.

Both anchors on this fact are the student's own membership days inside the
bucket, not the calendar bounds: `period_start_membership_date_key` is the first
and `period_end_date_key` the last. `period_start_date_key` stays rule-derived
(July 1, month truncation, the school week's Monday) because it is a hash input
to `student_period_key` — a data-dependent grain key would move whenever
attendance moved.

### `student_periods` cube and `student_periods_view` — new

A second cube and view rather than more measures on the daily one, because the
split is **weighting, not time grain**. Daily measures are day-weighted — ratios
of summed day counts, additive over any range. Period measures are
student-weighted — counts of students past a cumulative threshold at period end,
non-additive across periods because the counts accumulate from the start of the
academic year.

Three measurements decided it:

- A day-weighted cumulative ADA on the periods cube would **duplicate** the
  daily view at year grain — 0.920679 from both constructions on AY2025 — and
  mislead when summed across periods: 1.65M membership days at year grain
  against 9.43M summing the eleven AY2025 month rows.
- A student-weighted one **diverges** from the daily view's ADA by 0.66 points
  (0.914065 against 0.920679, 9,725 students), so it could not reuse the name.
- Cube has no mandatory-filter mechanism. Asking the wrong view for a measure
  returns "measure not found"; forgetting `period_type` on a merged view sums
  three or four grains into one denominator and looks plausible. The loud
  failure is the better one.

### The two views expose the same cuts

Splitting them is about weighting, not about what you may look at. So anything
you can break a rate down by on one view, you can break it down by on the other.
`student_periods_view` reaches the same chain `student_days_view` does: student
demographics, grade level, graduation year, entry and exit dates, retained-year
flag, ELL and IEP status, special-education fields, meal eligibility, the
homeroom teacher, the reporting term, and location including campus and city.

Two columns carry that: `fct_student_attendance_enrollment_periods` reads `student_enrollment_key` and
`term_key` from the period-end row on `fct_student_attendance_enrollment_daily`, so every attribute
reached through them is the one that applied as of period end, consistent with
every other value on the row.

PII is not a reason to withhold any of it. Student views expose every field and
gate by row-level scope — three policies, one per location scope, each with
`member_level: includes: "*"` — so a view exposing no identifiers would have been
the odd one out rather than the safe one.

### One path to every dimension

`location_key` keys the period fact's grain — one row per student per school per
period — and nothing else. It is a computation input, not a published dimension,
and location is reachable from `student_enrollment_key` through
`dim_student_enrollments`, so declaring a second edge to `dim_locations` would
give both Cube and the generated marts reference two paths to the same dim.

So neither fact declares it as a foreign key. `fct_student_attendance_enrollment_periods` does not
project it at all; `fct_student_attendance_enrollment_daily` keeps the column because the period fact
reads it, as a degenerate key. Cube reaches location by traversing
`student_school_enrollments`, the single path the daily cube already used.
Re-measured same-moment against a freshly built chain: across 4,374,538 period rows the two possible routes agree on all of them, with no disagreements and no null locations on either side.

An earlier measurement here reported five exceptions traced to one enrollment stint missing from `dim_student_enrollments`. That was drift, not a divergence. `int_extracts__student_enrollments` rematerializes, and `entrydate` is a hash input to `student_enrollment_key`, so a stint re-dated between a build and a test run produces keys that no longer resolve — and the tell was that the orphans were absent from the spine as well as the dim, which a real model divergence cannot produce. Any orphan count on these facts has to be read from a build and a spine at the same moment.

## Why not a pre-aggregation

Worth recording, because it is the obvious next question and the obvious answer
is wrong. These facts refresh twice a day and the query cost scales with the
number of readers, which is exactly the shape a cache suits. We measured before
deciding.

**Cube's actual BigQuery spend, last 60 days:**

| surface | queries | billed | avg per query |
| --- | --: | --: | --: |
| **student attendance** | **82** | **$0.48** | 998 MiB |
| other student surfaces | 1,609 | $27.33 | 3,296 MiB |
| everything else | 2,441 | $0.61 | 42 MiB |

Eighty-two queries in two months. Cube query volume is growing fast — 249 in May
against 2,244 in August — so at a few hundred readers this does become real: at
30,000 queries a month, 998 MiB each is roughly $223 a month.

**The monthly partition already captures about 95% of that, at no
architectural cost.** Prod's daily fact is a view today, so nothing prunes
anything and 998 MiB is the unpruned baseline. Filtering `attendance_date`
against the new partitioned table measured 63 MiB against 1,257 MiB. The same
30,000 queries then cost about $11.55 a month. A pre-aggregation would be
competing for what is left, and its own build — scanning 3.0 GiB twice a day —
takes about $1.10 of that back.

**Four things would make one expensive to get right:**

1. `count_distinct` is non-additive as a rollup property, and every measure that
   matters here is a `count_distinct` on a student key. A day-grain rollup cannot
   reaggregate to month or year, so it is one pre-aggregation per grain — or
   `count_distinct_approx`, which is an HLL estimate and wrong for a reported
   headcount.
2. The pre-aggregation has to carry whatever dimensions the access policy filters
   on, including school abbreviation for the school-scoped group, or it cannot
   serve a scoped query at all. High cardinality, and it multiplies the matrix
   again.
3. Cube's own overhead is the latency floor, not BigQuery: 0.9 to 1.5s per query
   in planning and Cube Store transport, against 2.17s of BigQuery on the worst
   29.6M-row query. A perfect pre-aggregation buys about 2.1s of a 55-second
   budget.
4. We have already had the incident. A partitioned pre-aggregation here with no
   explicit build range derived it from a time dimension routed through a
   calendar that runs to 9999, and a production redeploy enumerated roughly
   8,000 empty yearly partitions — #4460, reverted in #4462, bounded in #4463.

**The trigger for revisiting is measurable rather than a hunch.** After this
merges, watch `avg_mib_per_query` for `cube-cloud@teamster-332318` on these
surfaces. If it does not fall toward roughly 60 MiB, readers are filtering
`dates_date_day` instead of `attendance_date` — a routing problem, which a
pre-aggregation would mask rather than fix.

Measuring this also found where Cube's cost actually is: view marts re-expanding
their whole staging chain on every query, at 3.3 to 3.8 GiB a time. That is
roughly 80 times the bytes of a query against a materialized mart, and it is not
on these surfaces. Filed as #5109, where the fix is materialization rather than
caching.

## Verification

Both facts were built and reconciled, not read.

| check | result |
| --- | --- |
| `fct_student_attendance_enrollment_daily` rows | 29.6M |
| `fct_student_attendance_enrollment_periods` rows | 4,374,538 |
| `unique` on `student_day_key` | **PASS** at full scale |
| `unique` on `student_period_key` | **PASS** |
| FK relationships — dates, terms, enrollments, both anchors | PASS |
| `(ada_tier is null) = (is_truant is null)` on both facts | **PASS** — failed at 224,081 rows pre-gate |
| Unit tests, both facts | 6 of 6 PASS |
| Full local build, `int_students__enrollment_days+` | PASS=22, WARN=0, ERROR=0 |
| `fct_student_attendance_enrollment_daily` contract | 24 columns |
| `fct_student_attendance_enrollment_periods` contract | 13 columns; no prod consumer, the fact is new here |

**The period fact reads the right day, proven by an exact identity.** Each
student-school's cumulative count at its year-end row must equal that
student-school's total qualifying days on the daily fact. Summed across AY2025:

| | daily fact | year-grain period rows | delta |
| --- | --: | --: | --: |
| qualifying membership days | 1,650,787 | 1,650,787 | **0** |
| qualifying present days | 1,519,845 | 1,519,845 | **0** |

**AY2025 chronic absence, year grain, against prod attendance:**

| region | students | chronically absent | rate | truants |
| --- | --: | --: | --: | --: |
| Camden | 2,328 | 822 | 35.31% | 90 |
| Newark | 6,848 | 1,737 | 25.37% | 137 |
| Paterson | 548 | 112 | 20.44% | 6 |
| Miami | **0** | 0 | — | 0 |
| **Network** | **9,724** | **2,671** | **27.47%** | 233 |

**It reconciles to Topline on the same population.** Topline publishes 9,724
students and 2,669 chronically absent for AY2025, a rate of 27.45%. Identical
denominator, and the numerator differs by exactly 2 students. No student is
flagged by Topline that this fact does not also flag.

Both of those 2 are mid-year transfers between two Newark schools. Topline
partitions its running ADA by school, so it reports each on the post-transfer
leg alone — 92.7% and 91.8%, both on track — where the whole year is 70.8% and
89.8%, both chronically absent. That is Topline's defect rather than a
reconciliation gap, and it is filed as **#5122** with the Total Enrollment anchor.

**Network AY2025 ADA matches what production serves today:**

| source | AY2025 ADA | qualifying membership days |
| --- | --: | --: |
| `int_students__attendance_daily` (Tableau and Topline upstream) | 0.920678 | 1,650,800 |
| old `fct_student_attendance_daily` | 0.920678 | 1,650,800 |
| **new `fct_student_attendance_enrollment_daily`** | **0.920679** | 1,650,787 |

The whole delta is 13 membership days, 0.0008%, one student across thirteen dates
in November 2025. Those are attendance rows on days the enrollment spine says the
student was not enrolled. The new fact excludes them because enrollment decides
which rows exist, which is the more defensible answer.

Tableau reads `int_students__attendance_daily` directly and is untouched by this
pull request, so no Tableau figure moves.

## Caveats to carry into any report

- **Pre-AY2026 covers four regions again, and the memo's figures predate that.**
  Miami's attendance archive was dropped from `int_students__attendance_daily` on
  `main` (commit 2ed91424a, closing #4803) after Focus re-dated 959 enrollment
  stints. #5114 restored it on 2026-09-04 via #5158, re-keying the archive onto
  Focus enrollment stints — 793,259 rows and 2,582 students, AY2020 through
  AY2025. Every figure in the review memo was measured on a 2026-09-02 build, two
  days before that landed, so those tables read three regions where this branch
  now reads four. They need a re-measure, tracked as chase-down item 1 in the
  memo.
- **Early-year figures move.** With no floor, a September rate is computed over
  students holding a handful of days each, so it shifts by a few points day to
  day and settles as the denominator grows. Accepted cost of review decision 1.
  `count_students` is on both views so the population always shows beside the
  rate.
- **Tier 2 starts strictly above 90.0%, and KIPP Foundation criteria contradict
  themselves here.** Item 1 calls Tier 1 and Tier 2 on track at 90% and above;
  item 8 defines chronic absence as at or below 90.0%. This follows item 8 and
  the state criteria. **Confirm with KIPP Foundation before the figure goes to
  them** — if they want item 1's reading, the change is one comparison operator
  and a rebuild.
- **Chronic absence and truancy on the daily view need a pinned
  `attendance_date`.** The fact re-stamps a cumulative position on every day, so
  over an open range those measures count students who crossed a threshold on
  ANY day in it. Every description says so. That open-range reading is also what
  Topline publishes for truancy, which is the still-open question above.
- **186 Miami AY2026 enrollment stints contribute no rows at all.** Their
  `exitdate` is on or before their `entrydate`, so the spine's clamp puts
  `window_start` after `window_end`. Upstream, tracked in #5024.
- **Paterson presence-derived measures carry a known upstream uncertainty**
  (incomplete PowerSchool attendance-conversion items, #4193). `membership_value`
  is clean. Unlike Miami's, this gap is likely permanent for pre-AY2026 — the
  conversion items were never captured, so there is nothing to restore from.
- **AY2019 and earlier cannot support this rate at all.** 43% to 70% of
  in-session days in Newark and Camden carry no attendance value, so AY2018 rates
  only 2,159 of 6,099 students. Unrelated to Miami and to #5114.
- **The New Jersey truancy rule may not match the statute.**
  `int_students__attendance_daily` computes `is_truant` for the NJ regions as a
  running absence rate times the year's membership days, tested against 50.
  N.J.A.C. 6A:16-7.6 defines truancy as 10 or more cumulative *unexcused*
  absences. That is upstream of this pull request and affects every attendance
  surface; it needs its own issue once someone establishes whether the 50 is
  deliberate.
- **The legacy path excluded students marked "Transferred Out" in the current
  academic year.** This fact includes mid-year leavers, which all three
  authorities specify, and which matches Topline. That predicate never fired for
  AY2025 but will fire in the current year.

## Merging opens a window with no working student-scoped measures

Cube Cloud production redeploys when this lands on `main`. The Dagster deploy,
the code-location reload, and the sensor tick that builds the new facts take
minutes to tens of minutes. Cube wins that race.

For that gap the old measures are gone and the new views exist with no table
behind them, so every query fails at BigQuery — one at a time, with nothing
failing loudly.

Ordering:

1. Switch Cube Cloud to CLI deploy mode, which stops the automatic production
   build.
1. Merge.
1. Wait for the Dagster deploy across all five code locations.
1. Run one ordered run over `int_students__enrollment_days`, then
   `fct_student_attendance_enrollment_daily`, then `fct_student_attendance_enrollment_periods`. Order matters — the periods
   fact reads the daily fact. All three have SQL changes so the sensor will fire
   on its own, but not until the next 06:00 or 15:00 tick.
1. Confirm `kipptaf_marts.fct_student_attendance_enrollment_daily` and
   `kipptaf_marts.fct_student_attendance_enrollment_periods` hold rows.
1. Deploy Cube deliberately, then switch back to Git deploy mode.

### Drop five orphaned relations from the warehouse — @cbini

dbt does not drop a model it no longer manages, and this project has no
`on-run-end` cleanup hook, so renaming `fct_student_attendance_daily` leaves its
prod relations behind. The marts view is the one that matters: it is a view, so
it keeps resolving live and keeps publishing the **pre-2026 chronic-absence
definition** to anyone querying the warehouse directly.

Ninety days of `JOBS_BY_PROJECT` show no consumer outside the model's own SQL, so
this is cleanup rather than a live wrong number. Run after the merge and after
the Dagster build confirms both facts hold rows, so the old view stays available
as a fallback until the new one is proven:

```sql
drop view `teamster-332318.kipptaf_marts.fct_student_attendance_daily`;
drop view `teamster-332318.kipptaf_dbt_test__audit.fct_student_attendance_daily__enrollment_anchor_exists`;
drop view `teamster-332318.kipptaf_dbt_test__audit.fct_student_attendance_daily__enrollment_month_anchor_exists`;
drop view `teamster-332318.kipptaf_dbt_test__audit.fct_student_attendance_daily__enrollment_week_anchor_exists`;
drop view `teamster-332318.kipptaf_dbt_test__audit.not_null_fct_student_attendance_daily_student_enrollment_key`;
```

The per-PR `dbt_cloud_pr_70403104388001_*` copies need no action.

### Branching Minds extracts read the marts layer — #5202

Retiring `fct_student_attendance_daily` left `rpt_branchingminds__daily_attendance` with a dangling `ref()`, so this PR repoints it to `fct_student_attendance_enrollment_daily`. That keeps the outbound feed working and keeps the layering breach: 4 of 184 extract models read marts, and these are the 4.

Nothing the vendor receives changes. `event_id` is identical, because both facts hash `student_number`, `_dbt_source_project` and `calendardate`. `attendance_category` is identical, and the two models' null guards agree on all 7.4M prod rows across AY2023 to AY2026. AY2025 Newark and Camden match exactly. Paterson shows 2,880 extra rows on the new side, about 3.2%, which I have not attributed — the CI schema builds from stale `zz_stg` district copies, so settling it needs a prod build of `fct_student_attendance_enrollment_daily`.

### Still open after this PR

- **The memo's figures predate two changes.** Section 2.4's built-fact column was measured under the period-end truancy reading, and every pre-AY2026 figure was measured before Miami's archive returned. Chase-down item 1 in the memo.
- **63 Miami enrollment keys do not resolve to `dim_student_enrollments`** — 62 in AY2026 and 1 in AY2025, surfacing as a 536-row warn on `fct_student_attendance_enrollment_periods`. The same keys sit on `fct_student_attendance_enrollment_daily`, which carries no `relationships` test. They arrived with the `main` merge rather than with this branch's own changes. Not filed yet.
- Chase-down items 2, 3, 5, 6 and 7 in the memo are review follow-ups. None gates this pull request.

### Topline does not follow the rules this review established — #5122

Topline resets chronic absence at a school transfer, and anchors Total
Enrollment on the week's Monday. Both are wrong in production today and neither
depends on this pull request.

### General

### Dagster

      — **cannot run in this Codespace.** Both `kipptaf` and `kippmiami` fail at
      import while resolving dlt source credentials this machine does not have:
      `kipptaf/dlt/illuminate/assets.py:15` and
      `kippmiami/dlt/focus/assets.py:13`. Neither file is touched here. CI runs
      the real validation.
- [x] Verified the freshness policies directly: both attendance asset keys
      resolve, since a policy on a wrong key silently does nothing and looks like
      coverage. Both return `deadline=0 7,16 * * *`, `lower_bound=1:00:00`, on
      `kipptaf/marts/fct_student_attendance_enrollment_daily` and `kipptaf/marts/fct_student_attendance_enrollment_periods` —
      schema-derived, not the four-part `fqn` that would silently miss.
- [x] New integrations follow the Library + Config pattern — not applicable.

### dbt

      analysis, or application — the `cube` exposure covers both facts
- [x] **Breaking change?** Yes, and it is the point. 30 Cube measures and 5 Cube
      dimensions are deleted, plus the whole `student_enrollments` cube and its
      view, plus 3 more enrollment measures. Both facts are renamed. Nine
      columns come off the daily fact:

      - `student_attendance_daily_key`, renamed `student_day_key`.
      - Six snapshot and anchor flags: `is_current_record`, `is_latest_record`,
        `is_month_end_record`, `is_week_end_record`,
        `is_enrollment_month_end_record`, `is_enrollment_week_end_record`. The
        query-rewrite hook and the retired anchor measures were their only
        readers, and both are deleted here.
      - `ada_tier` and `is_chronically_absent`, which published the pre-2026
        definition on the daily fact. That closes #5060. They return on
        `fct_student_attendance_enrollment_daily` computed the correct way.

      `is_ca_eligible` and `count_students_min_10_days` were built in this
      branch and deleted again by review decision 1; they never reached prod.

      Nothing in this repo referenced any deleted member or column — every name
      was checked. Cube Cloud and the Cube MCP are the consumers. Rollback plan:
      revert this pull request, which restores every deleted member, column and
      model name, then redeploy Cube Cloud.
      not applicable.

### Docs

- [x] No schedule or sensor added. The published marts reference is regenerated
      and includes both facts.
- [x] `src/cube/CLAUDE.md` and `.claude/rules/cube-authoring.md` updated — the
      retired query-rewrite hook was documented as live in eight places, and the
      daily-versus-period split is now stated as a weighting distinction rather
      than a time-grain one.
- [x] The methodology review is written up in
      `docs/superpowers/attendance-methodology/2026-09-01-methodology-review-agenda.md`,
      with every figure re-measured against a local build of both facts.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

Closes #5060
Closes #5103
Refs #4744
Refs #4994
Refs #5015
Refs #5104
Refs #5109
Refs #5114
Refs #5122
Refs #5202


cristinabaldor:
Verdicts on all four findings. Two fixed in `161586f97`, two declined with reasons. I checked each against the code, and where the review said it could not query the warehouse, I ran its query.

### 1. `academic_year` sourced via the dates join — valid, fixed

Confirmed in full. The view pulled `academic_year` through `student_attendance_periods.dates`, that join is keyed on `period_end_date_key`, `dim_dates.academic_year` is a pure July-1 calendar cutover, and the cube defined no native `academic_year` at all. The fact does materialize its own.

You connected this to the surrogate key correctly — `academic_year` is a hash input precisely because it can diverge from the period dates, so a display path derived from those dates can diverge too.

I measured the divergence before fixing: **zero rows today**, across all three period types. So it was latent, not an active mislabel. Fixed anyway, because the view is new and unreleased — free now, expensive once something reads it.

The fix adds native `academic_year` and `academic_year_label` dimensions to the cube, both from the fact's own column, and drops both dates-join members from the view. I considered keeping `academic_year_label` from `dates` and rejected it: two year members that can disagree, one subtly wrong, is the trap this whole branch has been removing. The label expression qualifies with `{CUBE}` per the `dates` cube's own #4546 note, and matches its format — `academic_year` 2025 renders `2025-2026`.

### 2. Redundant `HAVING` — valid, fixed

Both sites now read `having period_end_date_key is not null`. I verified BigQuery accepts the SELECT-list alias in `HAVING` by compiling rather than assuming — `dbt compile` on the model passes.

This had been parked twice as cosmetic. Taking it now only because the file was already open for finding 1.

### 3. INNER JOIN to `int_students__schools` — declining, measured zero

Your reasoning is right and the caveat you attached to it is the reason I am declining: this is data-dependent, and you could not query the warehouse. I ran your query as written.

**0 rows, 0 distinct unmatched schools, 0 source projects.** No attendance row carries a school absent from `int_students__schools`.

A warn-severity test here would have nothing to catch, and you had already established the INNER JOIN is the correct choice per `marts/CLAUDE.md`'s hash-input rule. Recording the measurement here so the next reader does not re-derive it.

### 4. A fourth chronic-absence definition — the flagged file is fine, your parenthetical was the real finding

`rpt_tableau__okrts_referrals.sql:297` is legitimately distinct. It uses `<= 0.90`, which matches the network definition, at a different grain — New Jersey SSDS reporting periods off `int_extracts__student_enrollments_weeks`. No action.

The finding is the aside: **`fct_student_attendance_daily` still publishes the pre-fix definition.** Three divergences, not one:

| what | daily fact | network definition |
| --- | --- | --- |
| `is_chronically_absent` | `_running_ada < 0.90` — exactly 90.0% is not chronically absent | at or below 90.0% is |
| `ada_tier` Tier 2 lower bound | `>= 0.90` — includes exactly 90.0% | strictly above 90.0% |
| how it is evaluated | on `_running_ada`, an averaged float | on accumulated integer counts |

That third row breaks a stated constraint in the design spec, not a preference.

Split deliberately:

- **Descriptions fixed here.** They asserted "below 90%" and "Tier 2 = 90–94%" — the exact strings this branch corrected on the periods cube. They now describe what that SQL actually does, plus a pointer to `fct_student_attendance_periods` as authoritative. I did **not** rewrite them to the authoritative boundary, which would have made them lie in the other direction.
- **Operator filed as #5060.** Changing it is a data change on a contracted mart column, so it needs its own coordination and rollback plan, not a slip-in on a docs-and-deletions PR. That issue lays out the two options — delete the columns, or fix them in place — and recommends deletion, since after this PR nothing reads them: no dbt model refs the daily fact, and every Cube reference to `is_chronically_absent` and `ada_tier` now points at the periods cube.

### Separately, a correction to this PR's own description

Not from the review, but it landed while I was working through it. I had written that Miami's missing AY2025 attendance was a bug and filed #5058. It is not — `main` commit 2ed91424a deliberately dropped Miami's frozen PowerSchool attendance archive, closing #4803, because Focus re-dated 959 AY2025 enrollment stints and 153,577 fact rows then pointed at enrollment records `dim_student_enrollments` no longer holds.

I diagnosed it against my branch's copy of the model while querying prod, and did not run `git log` between the two. #5058 is closed with the correction. The PR body now says AY2025 chronic absence is permanently a three-region figure rather than advising anyone to wait for a fix that will not come.

cristinabaldor:
@claude update your review

cristinabaldor:
@claude update your review

cristinabaldor:
Correcting my own verdict from
[the comment above](https://github.com/TEAMSchools/teamster/pull/5057#issuecomment-5479471654).

I wrote that the `ada_tier` / `is_chronically_absent` operator on the daily fact
was "filed as #5060" and deliberately kept out of this pull request, because
changing a contracted mart column needed its own coordination and rollback plan.
That was accurate when I wrote it at 14:01 and false by 18:42 the same day:
`5dbbd5dff` rewrote the fact around the enrollment spine and both columns came
off it. This pull request already does what #5060 recommended.

Two things I got wrong in that verdict, worth naming because they are the same
mistake in different clothes:

- **The rationale inverted the moment the branch renamed the facts.** Any direct
  reader of `fct_student_attendance_daily` breaks on the rename whether or not
  the columns survive, so the coordination cost was already being paid in full.
  Carving out two columns to a separate pull request bought nothing — it would
  have left the wrong definition on a table no longer reachable under the name
  anyone referenced.
- **The "docs-and-deletions pull request" framing was already stale.**
  `5dbbd5dff` is a fact rewrite, not a docs pass.

**The one prerequisite #5060 named, now run.** #5060 said to confirm first that
no Tableau workbook or ad-hoc query reads these two columns directly. That check
had not been done. Ninety days of `JOBS_BY_PROJECT`:

| reader | queries touching the daily fact | naming either column | verdict |
| --- | --: | --: | --- |
| `dbt-user` service account | 6,580 | 1,677 | dbt executing the model's own SQL |
| Dagster Cloud agent service account | 4,742 | 2,840 | same, via the sensor |
| Codespaces service account | 984 | 161 | same, developer builds |
| `cube-cloud` service account | 154 | 36 | the 30 measures this pull request deletes |
| human accounts | 19 | 6 | `dbt --empty` compile checks, 0 bytes billed |

No consumer outside the model's own SQL. The six human-account hits are one
identical 1,584-character `select * from (select cast(null as string) as ...)`
run six times in July — dbt's empty-model validation, not a data read.

So the deletion is safe and needs no rollback plan beyond the one this pull
request already has.

**Fixed on this pull request:** the body's breaking-change bullet understated the
deletion. It named "the 3 anchor columns"; the documented column set actually
goes from 24 to 17. The bullet now lists all nine, in three groups — the key
rename, six snapshot and anchor flags, and these two — and the body carries
`Closes #5060`.

Also worth flagging on the other side of the ledger: this means #5057 resolves a
filed bug without having said so, which is the same stale-pointer failure mode
the branch spent four commits removing from `src/cube/CLAUDE.md`. The root
`CLAUDE.md` rule added in `b469ff46f` covers the tombstone half of it; it does
not cover a verdict comment going stale under a later commit on the same branch.

cristinabaldor:
Verdicts on the three findings left open by the review above. All three are
closed. Two were valid and are fixed; one was structurally correct and had zero
exposure, and chasing it found a real defect somewhere else.

Also note the branch has moved a long way since that pass — see the bottom.

### Finding 1 — `student_day_key` can collide on dual-enrollment · closed

The reasoning was exactly right and the exposure is zero. I rebuilt the window
logic against prod and looked for any two stints of one student, in one source
project, whose clamped day-windows overlap:

| | |
| --- | --: |
| Stints in `int_extracts__student_enrollments` | 112,825 |
| Overlapping stint pairs, all history, all regions | **0** |
| Cross-school pairs (true dual enrollment) | 0 |
| Same-school pairs (a data-quality overlap) | 0 |

`exitdate` is exclusive, so sequential stints touch but never overlap. Zero pairs
means `int_students__enrollment_days` has never emitted two rows for one
student-day, so the three-column hash cannot collide.

Since then the fact has been built at full scale into a dev schema — 29.6M rows —
and `unique_fct_student_days_student_day_key` **passes**. That is the direct
evidence the review asked for and could not get.

**Declining the `severity: error` promotion.** A bare `- unique` on the PK is the
layer's pattern, not an oversight: `fct_student_periods` does the same, and only
3 of 25 fact properties files in `marts/facts/` use `severity: error` anywhere.
Promoting it here would be a one-off.

**The review did find a real bug, in a different place.** The thing that made the
key look unsafe was `properties/fct_student_days.yml`, which claimed
`membership_value` is "split across schools if dual-enrolled" — describing
behaviour that has never occurred in any year of data. A correct reading of a
false document. That sentence is now replaced with what is true, plus the
measurement, so the finding cannot regenerate.

### Finding 3 — re-verify the INNER JOIN against the larger population · closed

This was the best catch of the pass. The earlier "0 unmatched" really was
measured against the old, smaller `int_students__attendance_daily`, and it really
did not transfer automatically. Re-ran it against
`int_students__enrollment_days`:

| | |
| --- | --: |
| Stints reaching the `int_students__schools` join | 111,737 |
| Rows dropped by the INNER JOIN | **0** |
| Distinct unmatched `(schoolid, _dbt_source_project)` | **0** |

No silent drop on the recovered historical population either. The join stays.

One thing that fell out of the check, not a finding against the join: the
*calendar* inner join upstream in `int_students__enrollment_days` drops exactly
one stint — a Miami AY2026 row with a `NULL` `schoolid`, which cannot join
anywhere. Nothing to fix in the model.

### Finding 5 — three `pct_*` descriptions self-contradict on the denominator · fixed

Confirmed and fixed in `cb055421a`. Two were a clean sentence delete.
`pct_truant` needed a rewrite rather than a delete, because its clause carried
real meaning about the eligibility floor applying to truancy.

Reading them closely turned up something the review did not flag and which
mattered more: **`pct_truant`'s denominator was itself wrong.** It divided a
CA-floored numerator by a CA-floored denominator, so it reported a rate over
CA-eligible students while its name claimed all students — and neither the Miami
nor the New Jersey truancy rule has an eligibility floor. `count_truants` no
longer applies one and `pct_truant` divides by `count_students`. At year grain
the numerator does not move (AY2025 NJ: 233 truant either way); the effect is
concentrated in early months, where 2025-08 goes from 0 to 550.

### The two lower-severity items, and the nits

- **Periods join collision (minor).** The same measurement closes it: 0
  same-school overlapping pairs, which is the only case that could fan out
  there.
- **ST06 on `fct_student_days.sql` (nit).** Dead — Trunk Check is green on that
  file, which the review job could not run.
- **Exposure `depends_on` not alphabetical (nit).** Fixed in `cb055421a`.

### The design has changed since this pass

Worth flagging rather than leaving a reviewer to discover it. The tier ladder,
the eligibility floor and the year-to-date accumulation now live on
`fct_student_days`, one row per student per calendar day, and
`fct_student_periods` reads that fact at period end and derives nothing. Both
facts were rebuilt and reconciled: qualifying membership days agree exactly
between them, 1,650,787 against 1,650,787, and present days 1,519,845 against
1,519,845, delta zero on both.

The Cube layer also gained full cut parity between the two views — chronic
absence and truancy can now be broken down by race, gender, grade level, IEP,
ELL, meal eligibility, homeroom teacher and term at period end, not only on the
daily surface — and location is reached by a single traversal path rather than a
second join, verified against a live Cube compile.

The pull request body is rewritten to describe all of it. A fresh review pass
would be reviewing a different design from the one above.

cristinabaldor:
Verdicts on all eleven findings. Seven fixed — five in `4d0dc38e4`, two more in
`3b59b5442` — two declined with evidence, two accepted-as-noted. I verified each
claim against the code before acting, including the two I am declining.

**Updated:** findings 10 and 11 were originally deferred here. Both are now
resolved, and both resolved to a deletion. See the bottom of the Fixed section.

## Fixed

### 5 — FK diamond to `dim_students` · valid, fixed

Confirmed: `dim_student_enrollments.yml:23-26` declares `student_key` as an FK to
`dim_students`, so declaring both `student_key` and `student_enrollment_key` on
this fact gave two routes to the same dim.

You were right to contrast it with `location_key`, and the contrast decided the
fix. `location_key` had to stay as a degenerate column because it does real
work — it partitions the cumulative windows and is a hash input to
`student_period_key`. `student_key` has neither claim:

| | `location_key` | `student_key` |
| --- | --- | --- |
| Hash input to `student_period_key` | yes | no |
| Partitions the cumulative windows | yes | no |
| Read by another dbt model | yes | no |
| Referenced by Cube | no | no — the view resolves `student_key` through `student_school_enrollments.students`, confirmed against a live Cube compile |

So the column is **dropped**, not made degenerate. That also matches
`fct_student_days`, which has never had a `student_key` at all — as you noted.
Contract goes from 14 columns to 13.

### 3 — Cube folder lists measures · valid, fixed

Confirmed twice over: `src/cube/CLAUDE.md:149-150` says "Folders group dimensions
only. Cube Cloud separates measures natively; don't list measures under
`members:`," and all five members were indeed measures on `student_days.yml`.
Removed; the folder keeps its dimensions. I checked both views programmatically —
`student_periods_view` had no such violation.

### 4 — stale generated reference · valid, fixed

Right conclusion, wrong evidence. The section did **not** list
`location_key → dim_locations` — it contained exactly one FK line,
`fct_student_periods }o--|| dim_students : "student_key"`. But it was stale:
last regenerated at `5dbbd5dff`, three commits before the FK changes.

Regenerated. The diagram now draws `fct_student_periods` to
`dim_student_enrollments` to `dim_students` — one path — and the FK table lists
all four, including the `term_key` you flagged as missing.

Worth recording for whoever regenerates it next: the script emits un-padded
markdown tables while the committed file is `trunk fmt`-padded, so a bare
regeneration shows a 237-line diff. Committing collapses it to the real 15.

### 9 — stale freshness rationale · valid, and it went further than the comment

Confirmed: `fct_student_days` carries zero `is_realized` / `is_*_record` columns —
they were deleted with the query-rewrite hook — so the rationale for
`attendance_daily_policy` was false.

It compounded. `attendance_periods_policy` justified being a *separate constant*
**because** of that now-false comment, while holding identical values. So both
collapse into one `attendance_facts_policy` with the reason true of both: same
materialization cron, both tables, and a failed build silently serves the
last-built rows. Your "collapse both to one shared constant" option, taken.

### 6 — stale `academic_year` provenance · valid, fixed

Confirmed by tracing the SQL. Rewritten to say it is read from
`fct_student_days`, whose `academic_year` comes from the enrollment spine — and
why that matters, since `dim_dates` carries a pure July 1 calendar cutover that
can disagree with the stint's year at a boundary.

### 10 — dead `dates` join · valid, removed in `3b59b5442`

You were right that it was unused, and thinking about *why* gave the reason to
delete rather than wire it up. The join was on `period_end_date_key`, so it could
only ever describe the **last day** of a period — misleading at year grain, where
it would report June, and redundant at month and week, where `period_start_date`
already carries it. Both date dimensions on that cube are native and `type: time`,
so Cube applies its own granularities with no calendar join.

The conformed-dimension rule still holds where the grain is a day: `student_days`
joins `dates` on `date_key` and exposes twelve `dates_*` members. A period simply
has no single day for `dim_dates` to describe.

I verified the removal against a live local Cube compile rather than assuming it,
for the reason I gave when deferring — a join can be reached by a granularity or
filter path that a member scan does not show. Before and after: 30 surfaces,
`student_periods_view` at 49 dimensions and 8 measures, `student_days_view` at 68
and 10. No member lost. A five-dimension query spanning locations, demographics,
enrollment status and terms still generates exactly one join per dim, and now zero
`dim_dates` references.

### 11 — `count_enrollments` · valid, deleted in `3b59b5442`

Went further than the ticket you suggested, because the history settles it. A June
2026 spec created this measure by renaming `student_enrollments.count_students` to
`count_enrollments` specifically "to free the honest name for the new cube," and
recorded that the renamed stint counter "and the entry/exit dates are no longer
exposed."

So it was never meant to be reachable — the rename existed to vacate
`count_students` for a real headcount, and that job is complete and irreversible.
Nothing consumes it, `public: true` on it was wrong metadata inside a
`public: false` cube, and removing the thing beats correcting the lie about it. Its
`measures:` block went too; it was the only member.

Not wired into a view, deliberately: it counts enrollment **stints**, and putting a
stint count beside `count_students` is precisely the confusion this pull request
exists to fix — the old `count_students` counted stints and double-counted a
re-enrollment.

One thing I checked before deciding, since it would have changed the answer: whether
`public: true` on an unreachable measure meant it was directly queryable and so
bypassing view-level row-level security. It was not. All four student cubes are
`public: false` and none carries an `access_policy` — views own access entirely, as
`src/cube/CLAUDE.md` specifies.

## Declined

### 1 and 2 — ST06 column ordering · declining on the stated consequence

Both findings assert these are "likely a Trunk/sqlfluff CI failure." They are
not:

```text
trunk check --force fct_student_days.sql fct_student_periods.sql
Checked 2 files
✔ No issues
```

sqlfluff flags neither file. The finding also reasons from "this PR's CI
checklist still shows Trunk unconfirmed" — Trunk Check is green on the current
head, and I run `--force` on every changed SQL file before pushing precisely
because the pre-commit hook only formats.

Whether the `src/dbt/CLAUDE.md` prose about join-order grouping is *also*
violated is a separate question I am not disputing. But reordering two contract
column lists to satisfy a rule the linter does not enforce is churn on a
1,400-line diff, and the ordering as written groups by table with blank-line
separation, which is the readability the rule exists for. Happy to be overruled
if you want the letter of it.

## Accepted as noted, not acted on

Two remain here — both conventions I agree with, declining the churn.

### 7 — tracking-issue refs in `description:` fields

You flagged this yourself as pre-existing rather than novel, and cited
`int_assessments__course_enrollments.yml:83` on `main` with the same pattern.
Agreed on the convention and declining the churn: on `is_oss` / `is_iss` /
`is_suspended`, "Null for Focus-sourced Miami rows. See #4927" is what tells a
reader whether a null is a bug or a known gap, which is exactly the decision the
description exists to inform. Moving it to the SQL puts it where the person
reading the column list will not see it.

### 8 — `GROUP BY` with no aggregation could be `DISTINCT`

Correct reading of the convention. Declining on the grounds that the convention's
purpose is already met: it asks for a one-line "grain projection, not
dup-masking" annotation so a reviewer can see the choice was deliberate, and the
CTE carries a ten-line comment doing that job, including the measurement (0
overlapping stint pairs across 112,825) and the upstream error-severity test that
would fail the build if it ever changed. Swapping the keyword would not add
information.

## One thing the review got right that I want to underline

The environment note at the top — "no warehouse/BigQuery/dbt/Cube MCP access, so
nothing below asserts a data-dependent claim as fact" — is why findings 3, 4, 5,
6 and 9 all landed: they are static, checkable claims. Every finding that needed
data was labeled as such. That is the right way to review without a warehouse,
and it is why this pass was more useful than the row counts would suggest.

cristinabaldor:
@claude update your review

cristinabaldor:
All four findings verified at the cited lines and fixed in `1ff4a2756`. The sweep in `4e141ffd1` and `b06190ace` was scoped to `src/dbt/` and `src/cube/model/`, so `scripts/` and `tests/` kept the intermediate names. That was my miss.

**1. `tests/scripts/test_generate_marts_reference.py:145` — accepted, fixed.** The assertion now reads `fct_student_attendance_enrollment_daily`. Your reasoning holds: `collect_fact_names` reads the real `properties/*.yml`, and a repo-wide search returns zero files under either intermediate name. I confirmed by inspection and by that search, not by running the test before the fix. After the fix the module passes 13 of 13.

**2. `scripts/generate_marts_reference.py:232` — accepted, fixed.** The `INTRO` string now names the current model.

I did **not** regenerate `docs/reference/marts-data-models.md` in this commit. Running the generator changes 127 lines, and all but a handful are markdown tables losing their padding, which only `trunk fmt` restores. Trunk cannot run in this environment right now — `.trunk/tools` is a dangling symlink in both the main checkout and the worktree, and `~/.cache/trunk` holds no CLI or launcher, so the pre-commit and pre-push hooks are silently no-oping. Committing the raw generator output would fail CI's own Trunk Check. The published doc's content is already correct; what is stale is only its section ordering, which sorts the two renamed facts into their old alphabetical slots. That regen is a follow-up once trunk works.

**3. `scripts/cube_rls_matrix.py:60` — accepted, fixed.** `DEFAULT_QUERY` now selects from `student_attendance_enrollment_daily_view`.

**4. `scripts/gen_column_naming_audit_inventory.py:53-54` — accepted, deleted rather than renamed.** Your first-match-wins analysis is right, so the two rules were unreachable. Renaming them would have produced two more unreachable rules, so both are gone. `tests/test_gen_column_naming_audit_inventory.py:68` now asserts on `fct_student_attendance_enrollment_daily`, which resolves to `Attendance` through the surviving `("student_attendance", "Attendance")` rule — so the test exercises a model that exists.

**A fifth you did not name.** `tests/fixtures/mart_yaml/fct_sample_attendance.yml:2` set its sample model to `fct_student_days`. Nothing reads that fixture — `_FIXTURE` points at `dim_sample.yml` — so it was inert, but the name dangled. It now matches its filename.

**On the checkout note.** Confirmed from this side: my worktree is clean, and the committed content of `CLAUDE.md`, `.claude/rules/cube-authoring.md` and `.claude/skills/cube-ops/SKILL.md` is the current version. The uncommitted reverts were local to the review runner.

**Pre-existing, not fixed here.** Seven tests in `tests/test_gen_column_naming_audit_inventory.py` fail on `_append_edfi_notes`, which `175e39a54` renamed to `_append_standard_notes` without updating them. They fail identically on `origin/main`, and no CI job covers that module, which is how it landed. Out of scope for this PR.

Issue #5064, opened 2026-08-29
fix(dbt): int_topline__suspension_weekly counts Miami AY2026 as zero suspensions, diluting a network topline metric


anthonygwalters:
## What's happening

Miami's DeansList feed stopped on 2026-06-25. Miami students stay in the enrollment spine, so `int_topline__suspension_weekly` left-joins a feed that no longer exists and reads every Miami AY2026 student-week as `is_suspended_y1_all_running = 0`.

Those zeros are structural, not real. They say the feed stopped, not that no one was suspended. The column feeds the network "Student and Family Experience → Suspensions" topline indicator, so Miami currently dilutes a reported metric with an artifact.

#5062 removed this exact artifact from the 3 OKRTS Tableau extracts. It did not touch the topline model, and the spec written for that work implies the problem is bounded to those 3 extracts. It is not.

## Steps to reproduce

1. Query `int_topline__suspension_weekly` for `academic_year = 2026`.
2. Join it to `int_extracts__student_enrollments_weeks` and group by region.
3. Miami returns student-weeks with `suspension_count_all` uniformly 0.
4. Confirm the cause: `int_deanslist__incidents__penalties` has no Miami rows after 2026-06-03.

## Where

- **Code location / dbt project:** `kipptaf`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** `int_topline__suspension_weekly`, feeding `int_topline__student_metrics`

## What fixed looks like

Miami reports as absent rather than as zero for AY2026, and the network suspensions topline stops including Miami student-weeks that have no feed behind them. Miami returns in quarter 2 once its replacement behavior platform is ingested, so whatever lands has to be easy to reverse.

<details>
<summary>For Claude</summary>

Context, not instructions. Found by the final whole-branch review of #5062.

`src/dbt/kipptaf/models/topline/intermediate/int_topline__suspension_weekly.sql:13-21` left-joins `int_deanslist__incidents__penalties` onto `int_extracts__student_enrollments_weeks` filtered `academic_year >= var("current_academic_year") - 1`, with no code-location handling. Consumed at `int_topline__student_metrics.sql:474`.

#5062 already built the mechanism this needs: the `exclude_deanslist_stopped(project_column, year_column)` macro in `src/dbt/kipptaf/macros/utils.sql`, driven by the `deanslist_stopped_code_locations` var in `dbt_project.yml` (currently `{kippmiami: 2026}`). Applying it here is a small change.

[...]

Issue #5063, opened 2026-08-29
chore(tableau): Paterson staff cannot open the OKRTS Dashboard, no AD group in Role Access


anthonygwalters:
## What and why

Paterson staff cannot open the OKRTS Dashboard. The workbook gates access on a `Role Access` calculated field that lists 13 Active Directory groups and 5 named users. None of them covers Paterson.

This blocks people, not data. The data side is fixed: Paterson's 2 location aliases now resolve, and all 23,652 Paterson behavior rows reach the warehouse. A Paterson principal still gets no dashboard.

`Role Access` today, on both OKRTS data sources:

```text
Group Staff NJ Regional          KNJ-SG-Tableau All Staff KTAF
Group Staff TEAM Council         KNJ-SG-Tableau Regional Special Education
KNJ-SG-Tableau All AP            TS-DL-Camden Culture Team
KNJ-SG-Tableau All Data          TS-DL-Newark Culture Team
KNJ-SG-Tableau All HOS           TS-DL-Newark Social Workers
KNJ-SG-Tableau All SL            TS-DL-NJ Culture Leaders
KNJ-SG-Tableau Culture All
```

Camden and Newark each have a culture team group. Paterson has none.

There is a second layer. The `Regional User Filter` set maps AD groups to regions, and no branch returns Paterson. So even a Paterson user who passes `Role Access` sees no Paterson rows on the 15 worksheets that apply the region set.

## Proposed approach

The tech team creates the Paterson group. The dashboard owner initiates the request and adds the group to both `Role Access` and `Regional User Filter` at the next republish.

Open question for the tech team: whether Paterson gets a `TS-DL-Paterson Culture Team` group matching the Camden and Newark pattern, or joins an existing NJ-wide group.

## Owner

Tech team creates the AD group. Dashboard owner initiates and does the workbook change. Tracked here so it does not get lost between the two.

<details>
<summary>For Claude</summary>

Context, not instructions.

Two workbook-side changes, both in `OKRTS Dashboard.twb`, both owned by the dashboard owner, neither in this repo:

1. `Role Access` — an `ISMEMBEROF(...)` chain defined identically on the `rpt_tableau__okrts_referrals` and `rpt_tableau__okrts_behavior` data sources. Gates 13 worksheets.
[...]

Issue #4858, opened 2026-08-13
chore(deanslist): align progress-report behavior category names across schools


anthonygwalters:
## Summary

DeansList report templates resolve behavior categories **by name**, so name drift between schools makes a report section populate at some schools and render empty at others, with no error.

Scope of this audit, deliberately narrow: **active** categories, at schools with data, in the three families the progress report actually reads — Corrective Behaviors, Habits of Excellence Corrections, and Values. Inactive categories retained for historical records are excluded, as are school IDs with no current data.

That leaves 90 categories across 19 schools, and only one item that genuinely needs a config change.

## 1. Habits of Excellence Corrections — two schools off-standard

| Configured name | Schools |
| --- | --- |
| `Tier 1 - Habits of Excellence Corrections` | 10 |
| `Tier 1 - Habits of Excellence Corrections *Only Once Per Class` | BOLD (125) |
| `Tier 1 - Habits of Excellence Correction (Only Once Per Class)` | Cooper Norcross (473) |

Cooper Norcross differs only by a missing `s` on `Correction`.

Both are MS/HS schools, so both are in scope for the redesigned back page. Every other live MS/HS school uses the standard name, and none is missing the category entirely — so this is the whole gap.

Worth deciding separately whether the entry rule belongs in the category name at all. BOLD encodes `Only Once Per Class` there; the rule the report will implement is one behavior per day per roster, which satisfies it either way.

## 2. Hatch has a stale duplicate

Hatch (130) carries two Corrective Behaviors categories:

```text
id 10545   active=false  visible=false    6 behaviors   Corrective Behaviors
id 15051   active=true   visible=true    13 behaviors   Tier 1 - Corrective Behaviors
```

The active one is correctly named. The inactive one is only a hazard for a name-based lookup that ignores the active flag. Low priority — delete it if it is not needed for records.

## 3. Values (5) is active everywhere but unused by the report

All 19 schools carry `Values`, `Values (10 Point Bonus)` and `Values (5)` as three active categories. The current progress-report code matches the first two and ignores `Values (5)`.

This is a question of intent rather than a naming defect: should `Values (5)` count toward BEAT Points on the progress report? If yes, that is a code change on our side, not a config change on yours.

## Checked and clean

Within these three families, at active categories on live schools:

- No leading, trailing or doubled whitespace
- No duplicate names beyond the Hatch pair above
- No live MS/HS school missing a Habits of Excellence category
[...]

Issue #4747, opened 2026-08-06
fix(dbt): okrts_referrals aggregates ignore referral_tier; unmapped categories bucket to Other


anthonygwalters:
## Problem

`rpt_tableau__okrts_referrals` classifies referrals into a `referral_tier` dimension, but the pre-aggregated referral counts bypass that classification entirely, and unsanctioned DeansList categories fall through the mapping into `Other`.

Surfaced by Zendesk [433674](https://teamschools.zendesk.com/agent/tickets/433674) (filed 2025-09-29). The reported symptom — documentation-only and social-work referrals appearing in the OKRTS tracker — is resolved: the workbook filters on `referral_tier`, and `Documentation%` was folded into Non-Behavioral in 8e7db3b6f (2026-03-10). The two issues below are what remains.

## 1. Aggregates ignore `referral_tier`

The `referral_counts` CTE counts `distinct incident_id` with no tier predicate, so `total_referrals`, `referral_rank`, and `referral_percentile` include every category regardless of how the workbook filters rows.

AY2025 network-wide:

| Measure | Incidents |
| --- | --- |
| `total_referrals` as modeled | 57,363 |
| Tier 2 plus Tier 3 only | 47,383 |
| Inflation | 9,980 (17.4%) |

3,631 of 6,062 students with referrals carry an inflated total. Rank and percentile derive from that number, so the ranking is wrong too — and wrong unevenly, since schools log documentation referrals at very different rates.

## 2. Unmapped categories bucket into `Other`

The tier CASE covers the sanctioned prefixes (T1, T2, T3, TX, TB, SW/SS/SSC/SSW). Categories created in DeansList outside that taxonomy have no mapping and land in `Other` — 2,684 incidents in AY2025, 4.7% of the total.

The largest unmapped groups:

| Category | Incidents |
| --- | --- |
| Skipped Reflection | 915 |
| Chronic Absence: 40 | 661 |
| Incident Report/Accident Report | 207 |
| Bus Referral / Bus Incident / Bus referral | 305 |
| T4-prefixed (Phone, Bullying) | 163 |

The T4-prefixed categories are not a real tier — they were added without approval. The bus variants are a straightforward mapping gap: they carry the same meaning as the `TB` prefix the CASE already handles.

Two separable pieces of work: prune or rename the unsanctioned categories at the DeansList end, and extend the mapping for the legitimate variants.

## Current AY2025 distribution

| referral_tier | Incidents | Share |
| --- | --- | --- |
[...]

PR #5065, opened 2026-08-29
fix(dbt): carry both NJ culture taxonomies in the OKRTS behavior model

### LINKED ISSUE #5062: fix(dbt): OKRTS behavior model drops every AY2026 NJ corrective behavior after the culture policy change
## What's happening

The new culture policy renamed and added behavior categories in Newark, Camden, and Paterson. `rpt_tableau__okrts_behavior` filters on a hardcoded list of 11 category names. None of the new names are on that list, so the OKRTS Dashboard shows no corrective behaviors at all for AY2026.

Since 2026-08-01, in the 3 New Jersey regions:

| | Rows |
| --- | --- |
| Passed the filter | 758,025 (all BEAT, from `Values`) |
| Dropped by the filter | 151,429 |
| Reached `category_type = 'Corrective'` | 0 |

`Corrective Behaviors` no longer exists in New Jersey. 3 categories replaced it, and none of them map:

- `Tier 1 - Corrective Behaviors`, about 1,686 events since 2026-08-05
- `Tier 1 - Habits of Excellence Corrections`, about 359 events, and new this year
- `Uniform` and `Dress Code`, which are compliance sweeps rather than corrections

Referrals are mostly fine. The `Tier 1/2/3 -` category prefixes already existed in AY2025, so `referral_tier` still resolves.

## Steps to reproduce

1. Open the OKRTS Dashboard and go to the Landing Page tab.
2. Set the year filter to 2026-27.
3. Read `# Corrective Behaviors per Student`. It returns nothing.
4. Query `kipptaf_deanslist.stg_deanslist__behavior` for `behavior_date >= '2026-08-01'` and confirm that `Tier 1 - Corrective Behaviors` rows exist.

## Where

- **Environment:** prod
- **Run, PR, or dashboard link (if any):** OKRTS Dashboard, exposure `okrts_dashboard`

## Also found

4 defects surfaced while tracing this. The first 3 are in scope here. The 4th is not.

**Paterson is missing from the dashboard entirely.** DeansList sends `Paterson Prep ES` and `Paterson Prep MS`. Neither name is in the location crosswalk, so the inner join in `rpt_tableau__okrts_behavior` drops all 23,652 Paterson behavior rows. Paterson's 843 students are already in the enrollment spine, so 2 crosswalk rows fix it.

**Miami reads as perfect behavior.** Miami's DeansList feed stopped on 2026-06-25. Miami students stay in the AY2026 enrollment spine, so the dashboard shows 5 schools with zero corrections and zero referrals instead of no data. Miami returns in quarter 2 once its replacement behavior platform is ingested, so the exclusion has to be easy to reverse.

**The exposure is missing a model.** `okrts_dashboard` lists 3 of the 4 datasources the workbook uses. `rpt_tableau__home_instruction` drives a whole dashboard tab, and the 4 AM refresh does not wait for it.

**Out of scope, tracked separately:** all 4 Home Instruction worksheets pin `academic_year` to 2025, so they go blank for AY2026 no matter what this issue changes.

## What fixed looks like

- Corrective behaviors appear for AY2026 in all 3 New Jersey regions.
- `Habits of Excellence` reports separately from `Corrective`.
- Filtering the dashboard back to AY2025 still renders every view.
- Paterson appears, with behaviors and incentives both populated.
- Miami shows no AY2026 data instead of empty data, and keeps its AY2025 history.

<details>
<summary>For Claude</summary>

Context, not instructions. Design settled with the dashboard owner over 3 rounds. Spec to follow under `docs/superpowers/specs/`.

**Scope: `kipptaf` only.** Do not touch the district projects or the shared `deanslist` package.

**Primary file:** `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__okrts_behavior.sql`

The Tableau contract is `category_type`, today 2-valued (`BEAT`, `Corrective`). 23 of 74 worksheets depend on it and `referral_tier`. Every worksheet-level `category_type` filter pins to `"BEAT"`; `Corrective` is consumed only by 3 calculated fields on `LP - Tree Data - BEAT Points`. All `behavior` filters are `level-members`, so new members flow through without workbook edits. The dashboard owner owns the republish.

Settled decisions:

1. Extend the category allowlist to cover both taxonomies rather than replace one with the other. Keep `Corrective Behaviors` and Miami's 9 legacy categories for history.
2. `category_type` gains a third value, `Habits of Excellence`. Do not fold it into `Corrective` — only 8 of 19 schools log it, so blending breaks cross-school comparison.
3. Add `Values (5)` and `Values (10 Point Bonus)` to the allowlist, both mapping to `BEAT`.
4. Add a raw `behavior_category` column so the workbook can separate the 3 Values families. This is what makes decision 3 safe.
5. Normalize `TEAMwork` to `Teamwork` in `behavior`. Do not map any renamed or split behavior, for example `Off Task/Not Following Directions` to `NFD/Off Task`.
6. Exclude `Uniform`, `Dress Code`, `System Behaviors`, and `Reflection Period`. Pending stakeholder review.
7. Keep the mapping inline. A sheet-backed crosswalk was considered and rejected for now.

Evidence behind decision 3 and 4:

- `Values (5)` first appears 2026-08-13 and has no prior-year history. All 22,868 events grant all 4 values at once, worth exactly 20 points and 4 rows each.
- `Values (10 Point Bonus)` first appears 2024-12-06 and swings hard: 96% of base `Values` points in Newark AY2024, 9.6% in AY2025.
- Excluded points run 109% of counted points for AY2026 to date. 9,956 of 10,164 students hold at least one `Values (5)` award.
- Adding `Values (10 Point Bonus)` raises the AY2025 BEAT total by about 9.6% in Newark and 6.7% in Camden. Filtering `behavior_category = 'Values'` reproduces the current figure exactly.
- `Values` stopped being 1 point per row in AY2026. `BEAT Exemplary` sits inside it at 5 points, so Newark reads 397,235 rows against 441,927 points. `BEAT Exemplary` stays in `BEAT` and in `total_points`, broken out as its own member.

Sheet edits, owned by the dashboard owner, in spreadsheet `1FCc28XWxFj3gSfItGGJ2tVU0C1fYD1JxKRxuSFqisMo`:

- Tab `src_people__location_crosswalk_v2`: add `Paterson Prep ES` mapped to `Paterson Prep Elementary School`, and `Paterson Prep MS` mapped to `Paterson Prep Middle School`.
- Tab `src_google_sheets__people__locations_v3`: set `Deanslist_School_ID` to 966 for `Paterson Prep Elementary School` and 1070 for `Paterson Prep Middle School`. About 20 models read this column, and 2 of them currently exclude Paterson through `is not null` filters.

Known pre-existing defects, out of scope:

- Miami `Big Reminders` and `Written Reminders` resolve to `behavior = NULL`, 43,263 AY2025 rows. The Miami branch regexes the category, which holds no parenthetical.
- `Weekly Incentive [LOD]` and `Monthly Incentive [LOD]` test `behavior = 'Earned Weekly Incentive'`, which the allowlist makes unreachable. Both always return 0.
- `referral_tier` exists twice and the copies have drifted. See #4747.

Refs #4747, #4858.

</details>

anthonygwalters:
# Pull Request

Closes #5062

## Summary & Motivation

`rpt_tableau__okrts_behavior` filtered on 11 hardcoded category names. A culture policy renamed and added behavior categories in Newark, Camden and Paterson, so every AY2026 corrective behavior fell through the filter and the OKRTS Dashboard showed 0 corrections for New Jersey.

**5 files. One model, its yml, one new test, that test's yml, and a one-line exposure fix.** No shared macro, no project-level var — an earlier revision of this branch carried a Miami stopped-feed exclusion that introduced both, and it has been split into its own PR so those get reviewed on their own terms.

What changes:

- **Two drifting lists become one.** The model carried a `where ... in (...)` allowlist *and* a separate `category_type` CASE, with no guard between them — a category could pass the allowlist, match no branch, and reach the extract as a null `category_type`: invisible to every workbook filter but still fanning out the spine columns. `category_type` is now derived once and filtered on, so the allowlist stops existing as a second list.
- **Both taxonomies map at once.** Old (AY2025 and earlier) and new (AY2026+) names both resolve, so filtering the dashboard back to last year still renders.
- **`Habits of Excellence` becomes a third `category_type`**, not a fold into `Corrective`. Only 8 of 19 schools log it and it is middle- and high-school only, so blending it would inflate the corrective rate at exactly the schools that adopted it.
- **`Values (5)` and `Values (10 Point Bonus)` are admitted as `BEAT`**, with a new raw `behavior_category` column so the workbook can separate the 3 `Values` families.
- **`TEAMwork` normalizes to `Teamwork`.** The split is inside AY2026, not just across years — `Values` logs `TEAMwork` while both bonus categories log `Teamwork`, and the workbook's colour map and manual sort only know `Teamwork`.
- **A detector so this cannot recur silently.** A warn-severity test lists any DeansList category that is neither mapped nor deliberately excluded.
- **`rpt_tableau__home_instruction` added to the `okrts_dashboard` exposure.** The workbook reads 4 datasources; the exposure declared 3.

Paterson's absence was fixed outside this repo, by 2 rows added to the `People` location-crosswalk sheet. This PR's model now carries Paterson — 23,652 behavior rows that previously failed the crosswalk join.

## Reviewer Notes

**The exposure change is lineage, not scheduling.** The extract-refresh `ScheduleDefinition` targets the single exposure asset and its body is one `workbooks.refresh()` REST call. `refs` become `deps`, which are graph edges for lineage and staleness — they never gated execution, before or after.

**A local build may fail on an unselected upstream** with `_dbt_source_project not found inside b`. That is a stale personal dev copy of `stg_deanslist__behavior`, not a defect in this change. Re-run with `--defer --favor-state --state <prod manifest>`.

**One condition was deliberately dropped, not lost.** Diffing against `main`, the Miami branch of the behavior-extraction CASE loses `and behavior_category != 'Earned Incentives'`. That guard became unreachable once `where category_type is not null` became the single filter: `Earned Incentives` maps to no `category_type`, so those rows are dropped before `behavior_extracted` is ever read. Verified as a no-op, removed on review feedback rather than by accident.

**Unfiltered AY2025 BEAT points rise 8.7% in Newark and 6.1% in Camden** because `Values (10 Point Bonus)` is now admitted. Intended — it was excluded by omission, not by design — and reversible in the workbook with a `behavior_category = 'Values'` filter.

**The new test carries a hand-maintained list.** 15 mapped names mirroring the model's CASE, 38 deliberate exclusions. That is a real ongoing sync obligation, documented in the file. #5066 tracks the fact that this does not scale past one consumer.

### General

### dbt

- [x] `behavior_category` added with `data_type: string`. Contracts are enforced on `models/extracts/**` by inheritance, so this was mandatory.
- [x] `category_type` and `behavior` gained `description:` entries, since this change alters what both mean.
- [x] The `okrts_dashboard` exposure now lists all 4 datasources the workbook uses.
- [x] **Breaking change?** No. No column renamed, removed, or retyped. The model gains `behavior_category`; `category_type` gains a third value.

## Workbook republish required

The dashboard owner performs these in Tableau Desktop, not in this repo:

- [ ] Remove the 2 `Exclude Paterson (TEMP)` data-source filters. Until they come off, Paterson stays invisible on every tab regardless of what the warehouse holds.
- [ ] Refresh the embedded `.hyper` extracts in Desktop so `behavior_category` appears in the field list. A new warehouse column does not reach an embedded extract on its own.
- [ ] Add the `Habits of Excellence` member and a parallel measure on `LP - Tree Data - BEAT Points`. Leave the 3 existing corrective calculated fields corrective-only.
- [ ] Add a `behavior_category = 'Values'` filter to `EA - Behaviors - Lines`, `EA - Behaviors - Roster`, `EA - Behaviors - Staff`, `SO - Behaviors - Lines`, `SO - Behaviors - Roster`, and any year-over-year view. `Values (5)` is a bulk award emitting 4 rows per event, so it inflates `SUM(behavior_count)` as well as points — `EA - Behaviors - Staff` ranks staff by entry volume and is the sharpest case.
- [ ] Filter Miami out of AY2026 views. Miami's DeansList feed stopped 2026-06-25 but its students remain in the enrollment spine, so it renders as 5 schools with perfect conduct. Handled in the workbook for now; the dbt-side fix is the follow-up PR.

## Related issues

- #5063 — Paterson staff cannot open the workbook: no Paterson group in `Role Access` or `Regional User Filter`. Not resolved here. Paterson data reaches the warehouse; people still cannot see it.
- #5064 — `int_topline__suspension_weekly` has the same stopped-feed defect and dilutes a network topline metric with structural zeros. Pre-dates this branch.
- #5066 — the unmapped-category detector is per-consumer and does not scale; 5 models hardcode behavior literals.
- Refs #4747, #4858.

## CI checks

- [ ] **Trunk** — `trunk check --force` was run locally on every changed file and returned no issues.
- [ ] **dbt Cloud** — passes.

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Built as 4 implementation tasks plus a validation task, each independently reviewed, then a whole-branch review, a fix wave, a scoped re-review, and finally a split that removed the Miami half.

**The commit history is churny on purpose.** It adds the Miami exclusion and later removes it. Judge the net three-dot diff, not the journey. `rpt_tableau__okrts_referrals.sql` and `rpt_tableau__suspension_over_time.sql` are byte-identical to the branch point.

**AY2025 invariant — the binding requirement.** Filtering `behavior_category = 'Values'` must reproduce production's AY2025 BEAT figures per school. A full-outer-join comparison across Newark, Camden and Paterson returns **0 differing rows**. Miami AY2025 compared separately (its BEAT rows come from 7 parenthetical category names, not from `Values`) is byte-identical. Re-run after every structural change to this model, including the split and the comment cut — not carried forward.

**`category_type` domain.** `BEAT` / `Corrective` / `Habits of Excellence`, with **no NULL bucket** among rows carrying a behavior. Closing that hole is the structural point of the redesign.

**Extract counts against staging (AY2026).** `Tier 1 - Corrective Behaviors` 1,686 raw / 8 dropped. `Tier 1 - Habits of Excellence Corrections` 359 / 2. `Values (10 Point Bonus)` 45,180 / 41,856. `Values (5)` 91,472 / 39,508.

The large `Values` drops are by design: 100% of those rows are dated before the school year started. Schools are asked to delete such entries and do not; the calendar-week inner join removes them. What the dashboard shows is in-session points, which is why this does **not** reproduce the DeansList progress-report total and the spec does not claim it does.

**Paterson (AY2026).** PPES 25,954 rows / 10,860 behaviors; PPMS 14,589 / 7,070. `is_earned_progress_to_quarterly` is 0 for both — investigated and **not** a defect. Paterson's DeansList terms exist and `int_deanslist__behavior_incentive_by_term` holds Paterson rows (966: 1,441 `Weeks`; 1070: 749), which could not exist without the `Deanslist_School_ID` fill. The flag is 0 because the join requires `incentive_type = 'Weeks (Progress to Quarterly Incentive)'` and Paterson has never logged that behavior; 16 other school ids have. An operational practice gap.

**The detector** follows the existing repo idiom at `src/dbt/focus/tests/focus_unmapped_phone_contact_titles.sql`. It is proven in both directions: passes clean on current data, and removing `Values (5)` from its mapped list makes it return exactly that category with 91,472 rows.

It deliberately does **not** trim `behavior_category`, because the model's CASE does not trim either. An earlier draft trimmed, which would have let a padded variant match as mapped and stay silent while the model dropped those rows — reproducing this incident's failure shape inside the guard built to prevent it.

Its exclusion list carries 12 categories found during triage, in 3 commented groups. Six are genuinely corrective (`Courage/Confidence/Voice/Kindness - Negative`, `Negative (5th Grade)`, `Routines and Systems`) but are AY2025-only Newark one-offs totalling 109 rows against 159,609 mapped `Corrective` rows, absent from AY2026 because the new policy retired them, and dropped by the previous allowlist too.

**Known limitation.** The mapped list is a human transcription of the model's CASE, kept in sync by convention rather than construction. See #5066.

</details>


anthonygwalters:
Verdict per finding. Each convention claim was checked against `src/dbt/CLAUDE.md` and each data claim against production before acting.

## 1. `select * except` in the final select of `rpt_` models — **FIXED** (`df581116a`)

Valid. `src/dbt/CLAUDE.md:1121` states the rule verbatim, and the only documented `select * except` exception (line 1137) is scoped to **staging** models dropping `source_file_name`, which does not cover extracts. The "no existing precedent" grep also holds.

All 3 outer selects now enumerate their columns. The CTE wrap stays — it is what lets `exclude_deanslist_stopped` run after the window functions, which is load-bearing for `is_week_ytd`.

Because hand-enumerating about 160 columns is exactly where a silent drop or reorder hides, this was verified against production rather than by eye:

| Model | dev cols | prod cols | Added | Removed | Relative order preserved |
| `rpt_tableau__okrts_behavior` | 38 | 37 | `behavior_category` | none | yes |
| `rpt_tableau__okrts_referrals` | 90 | 90 | none | none | yes |
| `rpt_tableau__suspension_over_time` | 31 | 31 | none | none | yes |

The AY2025 invariant still returns 0 differing rows, all 3 models build, and `trunk check --force` is clean.

One trade-off worth recording, since it argues mildly the other way: with contracts enforced, `select * except` meant a column added to the CTE would flow to the output and fail the contract loudly. An explicit list absorbs it silently as an unused CTE column. The convention still wins — an extract's column list should be a conscious choice — but the tripwire is genuinely weaker now.

## 2. Region-guard removal assumes Miami and NJ categories are disjoint — **VERIFIED, no change**

Ran your suggested query across the table's full history. **0 cross-region overlaps.** `kippmiami` appears only under `Big Reminders` (32,961) and `Written Reminders` (66,758). The 6 NJ categories appear only under `kippcamden`, `kippnewark` and `kipppaterson`.

The design spec records this disjointness as the explicit reason the guards were droppable, and the branch's Task 2 review confirmed it independently. Dropping the guards is also what closes the old NULL hole: a Miami row reading `Values` previously matched no branch and survived with a null `category_type`.

## 3. `exclude_deanslist_stopped` NULL edge case — **VERIFIED, no change**

Correct about three-valued logic: a NULL `year_column` would make the predicate NULL and drop the row. It cannot arise here — `academic_year` is NOT NULL across **3,832,582** rows in `int_extracts__student_enrollments_weeks` and **134,351** in `int_extracts__student_enrollments`, the only two spines the 3 call sites filter on.

Not guarding against it deliberately: a `coalesce` would make a genuinely corrupt spine row silently pass the filter instead of surfacing.

## 4. No uniqueness test or model description on the 3 `rpt_` models — **CONFIRMED pre-existing, out of scope**

The rule at `src/dbt/CLAUDE.md:763` does apply, and the exception at `:767` covers only thin cross-project wrapper `rpt_` models sourcing `kipptaf_extracts` — which these are not. So the gap is real.

It predates this branch on all 3 models, as you confirmed. Not fixing it here: adding uniqueness tests to 3 wide extracts is its own change with its own validation, and folding it into a taxonomy retrofit would obscure both.

## 5–8 — no action needed

Agreed on all four. Worth noting your independent check of `tableau/schedules.py` — an earlier draft of this PR's body claimed the exposure change makes the 4 AM refresh wait for the new model, which is wrong for exactly the reason you identified. It was corrected before the PR opened, and your check confirms the corrected wording.

anthonygwalters:
Verdict per finding.

## 1. Dropped `!= 'Earned Incentives'` guard, unexplained — **DOCUMENTED**

Correct on all counts, including that it is a no-op. It was removed deliberately on feedback from an earlier review round that flagged it as unreachable, not dropped by accident during the CTE split — but you are right that nothing in the shipped artifact said so, and a reader diffing against `main` had no way to tell.

Added to the PR body under Reviewer Notes. Kept out of the SQL: a comment explaining a line that is no longer there describes history rather than the code, and the next reader of the file has no context for it.

Third reviewer to raise this, which is itself the argument for writing it down.

## 2. NULL `behavior_category` making the warn noisy — **VERIFIED, no action**

Zero nulls. Across the test's own window (`behavior_date >= '2025-07-01'`), `stg_deanslist__behavior` holds **29,164,360 rows and 0 with a null `behavior_category`**.

The passing test is independent proof of the same thing. Your reasoning about the join behavior is right, and it cuts the other way: a NULL category misses both `left join`s, so `m.category is null and x.category is null` evaluates TRUE and the row *would* surface. The test returns zero rows, so none exist. The detector already covers this case rather than being blind to it.

Not adding a `not_null` test on the staging column. It would be asserting a property this model does not depend on — a null category is handled correctly today, by being dropped.

## 3. Inline comment overlapping the `behavior` column description — **ACKNOWLEDGED, no action**

Agreed with your read, including which half earns its place inline. The normalization-order rationale is the part that would not survive as a `description:`, and that is the part that stayed.

Worth adding for the record, since it sharpens the case: an intermediate revision of this branch trimmed that comment too aggressively and left it citing `'TEAMwork (Community)'` — a value that does not exist as a behavior anywhere in the source. A local review caught it. The concrete risk was a maintainer grepping the fictional value, finding nothing, concluding the CASE was dead, and deleting a normalization that conforms 234,881 AY2026 rows. `6329b68a7` restored the real fact (`Values` logs `TEAMwork`; `Values (5)` and `Values (10 Point Bonus)` log `Teamwork`).

That is why this landed on the keep-some-inline side. The failure mode was not a reader lacking the description — it was a reader trusting a comment that had drifted from the data.

## 4. Confirmations — noted

Particularly useful: the repo-wide `select * except` grep, and confirming `okrts_referrals.sql` and `suspension_over_time.sql` are untouched. That second one is the load-bearing claim of the split, and an independent check of it is worth more than my own.

Issue #4740, opened 2026-08-05
feat(dagster): create empty BigQuery tables for configured Focus tables with no source rows


cbini:
## Problem

11 of the 77 configured Focus tables are absent from `dagster_kippmiami_dlt_focus` because they hold no rows in Focus yet: `attendance_day`, `attendance_notes`, `attendance_period`, `discipline_incidents`, `discipline_incidents_join_referrals`, `gradebook_grades`, `referral_code_offenses`, `student_groups`, `student_standard_grades`, `students_join_groups`, `students_join_students`.

All 11 are already declared in `src/dbt/focus/models/staging/sources-bigquery.yml`, but none has a staging model — you cannot build a contract-enforced staging model against a table that does not exist. Only 64 of 77 configured tables are staged today, and each newly-populated table currently needs a dbt change before its data is usable.

Cause: dlt's `Extract._handle_empty_tables` writes an empty file only for tables that have previously **seen data** (`schema.data_tables(seen_data_only=True)`). A table that has never loaded produces no file, normalize drops the package, and nothing reaches BigQuery. The asset still succeeds with `jobs: []`, so this is not a failure — just an absence.

## Goal

A configured Focus table with zero source rows gets created in BigQuery as an empty, correctly typed table, so a staging model can be written for it before its data arrives.

## Approach

1. Yield `dlt.mark.materialize_table_schema()` for a never-loaded table so dlt creates the table from the reflected schema. Reaching it means wrapping the exported `table_rows` generator, since `sql_database` yields a plain empty list that dlt discards — the same wrapper pattern `libraries/dlt/powerschool/` already uses.
2. Enable dlt's own row-id columns for the Focus pipeline, set in the op body so it stays inside the step pod:

   ```python
   dlt.config["normalize.parquet_normalizer.add_dlt_id"] = True
   dlt.config["normalize.parquet_normalizer.add_dlt_load_id"] = True
   ```

   This is required, not cosmetic. `materialize_table_schema()` runs through dlt's object path, which always injects `_dlt_id` / `_dlt_load_id` as REQUIRED. Without the knobs, the arrow data path omits them and the first real load fails terminally.

3. One-time migration: a single Dagster run with dlt `refresh` set to drop resources, so dlt drops and recreates all 66 populated Focus tables with the new column set. No hand-written DDL. Cost is one full reload from Focus, which `replace` performs daily anyway.
4. Write staging models for the 11 new tables in the same PR.

[...]

PR #5075, opened 2026-08-31
fix: widen unbounded numeric on the Focus gradebook_grades table

### LINKED ISSUE #5074: fix: Focus sync fails on unbounded numeric in gradebook_grades.points
## What's happening

The KIPP Miami Focus sync has been down since 2026-08-29 12:19 UTC. Every run fails in the dlt extract step.

The failing column is `gradebook_grades.points`. It is an unbounded Postgres `numeric`, so dlt renders it as `decimal128(38, 9)`. A value now carries more than 9 decimal places, and pyarrow refuses to rescale it:

```text
PyToArrowConversionException: Conversion to arrow failed for field `points`
with dlt hint `data_type=decimal` and `inferred_arrow_type=decimal128(38, 9)`
Insufficient decimal precision Rescaling Decimal value would cause data loss
```

All 20 Focus tables share the step `kippmiami__dlt__focus`, so the one bad column blocks the whole sync, not just `gradebook_grades`.

This is the same bug as #5021, which hit `student_gpa_calculated.weighted_gpa`. That fix added an opt-in `widen_unbounded_numeric` key per table. `gradebook_grades` never opted in.

## Steps to reproduce

1. Wait for a `kippmiami__dlt__focus__intraday_sensor` tick, or launch the Focus dlt asset by hand.
2. Watch the step `kippmiami__dlt__focus` fail in extract.
3. Read the bottom of the error chain for the `points` conversion error.

## Where

- **Code location / dbt project:** `kippmiami` / `focus`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** https://kipptaf.dagster.cloud/prod/runs/1a7b7d07-4a91-4ee3-9896-035e2617f5c3

<details>
<summary>For Claude</summary>

Root cause is confirmed from the run's error chain bottom, not inferred.

Files:

- `src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml` — the `gradebook_grades` entry needs `widen_unbounded_numeric: true`.
- `src/dbt/focus/models/staging/stg_focus__gradebook_grades.sql` — widening retypes `points` and `possible_points` to BIGNUMERIC, so both need `cast(... as numeric)`. The contract in the matching `.yml` declares `data_type: numeric`, and `src/dbt/focus/dbt_project.yml` enforces contracts on staging.
- Precedent to copy: `src/dbt/focus/models/staging/stg_focus__student_gpa_calculated.sql`.

Verified against BigQuery: `dagster_kippmiami_dlt_focus.gradebook_grades.points` and `.possible_points` are both `NUMERIC` today, and `student_gpa_calculated`'s widened columns are `BIGNUMERIC`.

Open question, not yet tested: `widen_unbounded_numeric_adapter`'s docstring claims the `replace` disposition cannot retype a column in place. The Focus pipeline runs `autodetect_schema=True` with `loader_file_format="parquet"`, so a `WRITE_TRUNCATE` load may retype fine on its own. Try the plain load first. If BigQuery rejects the type change, launch one manual run selecting only `gradebook_grades` with run config `refresh: drop_resources` — the path `FocusDltConfig.refresh` already supports, added in #4740.

Fixed means the Focus sync runs green and `gradebook_grades` materializes again.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will restart the KIPP Miami Focus sync, which has failed every run since 2026-08-29 12:19 UTC.

`gradebook_grades.points` is an unbounded Postgres `numeric`. dlt renders that as `decimal128(38, 9)`, and a value now carries more than 9 decimal places, so pyarrow refuses to convert it and the extract dies. All 20 Focus tables share the step `kippmiami__dlt__focus`, so the one column took down the whole sync.

This is the same failure as #5021 on `student_gpa_calculated.weighted_gpa`, and it takes the same fix: set `widen_unbounded_numeric: true` on the table in `focus.yaml`.

Widening retypes `points` and `possible_points` to BigQuery BIGNUMERIC, so `stg_focus__gradebook_grades` casts both back to `numeric`. The staging contract declares `data_type: numeric`, and `src/dbt/focus/dbt_project.yml` enforces contracts. This mirrors `stg_focus__student_gpa_calculated`.

Closes #5074

## Reviewer Notes

One thing may need a follow-up run after merge. `points` and `possible_points` are `NUMERIC` in BigQuery today, and the adapter's docstring says the `replace` disposition cannot retype a column in place. If the first post-merge load fails on the type change, someone launches one manual Focus run selecting only `gradebook_grades` with run config `refresh: drop_resources`. That path already exists, from #4740.

### General

### Dagster

      code location

Neither Dagster box is checked, and both need a reason. `kippmiami.definitions`
cannot load in a Codespace: it resolves the Focus database credential at module
import, and that credential is unset here. So `definitions validate` and
`tests/test_dagster_definitions.py` both fail for a reason unrelated to this
change. I checked the config the way the code location does instead — parsed
`focus.yaml` and rebuilt the `widen_numeric_tables` set, which now holds exactly
`gradebook_grades` and `student_gpa_calculated`. The change adds one key to a
YAML file and edits no Python. The third box holds: the asset key and the
Library plus Config pattern are untouched.

### dbt

- [x] Include (or update) an
      summary above.
      See

Every dbt box is checked because nothing triggers it. This PR adds no model, so
there is no new properties file and no new exposure. It renames and removes no
column, so it is not a breaking change — `points` and `possible_points` keep
their names and their declared `numeric` type. It touches no external source.

### Docs

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Diagnosed from run `1a7b7d07-4a91-4ee3-9896-035e2617f5c3` with `superpowers:systematic-debugging`.

Root cause is read from the bottom of the run's error chain, not inferred:

```text
pyarrow.lib.ArrowInvalid: Rescaling Decimal value would cause data loss
```

wrapped by

```text
PyToArrowConversionException: Conversion to arrow failed for field `points`
with dlt hint `data_type=decimal` and `inferred_arrow_type=decimal128(38, 9)`
```

**What I verified**

- Last successful `kippmiami/dlt/focus/gradebook_grades` materialization: 2026-08-29 12:19:31 UTC, 5327 rows.
- The 10 most recent `kippmiami__dlt__focus__intraday_sensor` retries all end `STEP_FAILURE`, roughly 15 minutes apart.
- `dagster_kippmiami_dlt_focus.gradebook_grades.points` and `.possible_points` are `NUMERIC`. `student_gpa_calculated`'s 16 widened columns are `BIGNUMERIC`. Both read from `INFORMATION_SCHEMA.COLUMNS`.
- `tests/libraries/test_dlt_focus_type_adapter.py` — 14 passed.
- Parsing `focus.yaml` yields `widen_unbounded_numeric` for exactly `gradebook_grades` and `student_gpa_calculated`, out of 79 tables.
- `dbt build --select stg_focus__gradebook_grades` in the `kippmiami` project — PASS=5 WARN=2 ERROR=0. The contract holds.

**What I did not verify**

- I could not reach the Focus Postgres database from the Codespace, so I did not read the offending `points` value or measure its real scale. `(76, 38)` stays the destination ceiling, as set in #5021.
- I did not test whether a `replace` load can retype `NUMERIC` to `BIGNUMERIC` in place. The pipeline runs `autodetect_schema=True` with `loader_file_format="parquet"`, so a `WRITE_TRUNCATE` load may handle it. The `refresh: drop_resources` fallback in Reviewer Notes covers the other case.
- `dagster definitions validate -m teamster.code_locations.kippmiami.definitions` cannot run in the Codespace — `kippmiami.definitions` resolves the Focus database credential eagerly at module load, and that credential is unset here. I checked the config the same way the code location does instead, by parsing `focus.yaml` and rebuilding the `widen_numeric_tables` set. The change adds one key to a YAML file and touches no Python.

**Separate, pre-existing, not from this change**

The build warned on 2 relationships tests: 3 orphan `student_id` values and 4289 orphan `assignment_id` values. My change casts 2 numeric columns and cannot affect either column. Worth a look on its own, but it is not this PR.

</details>

### DIFF
--- a/src/dbt/focus/models/staging/stg_focus__gradebook_grades.sql
+++ b/src/dbt/focus/models/staging/stg_focus__gradebook_grades.sql
@@ -3,8 +3,6 @@ select
     student_id,
     assignment_id,
     standard_id,
-    points,
-    possible_points,
     letter_grade,
     exclude_from_average,
     late,
@@ -19,4 +17,7 @@ select
     uuid,
     created_at,
     updated_at,
+
+    cast(points as numeric) as points,
+    cast(possible_points as numeric) as possible_points,
 from {{ source("focus", "gradebook_grades") }}
--- a/src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml
+++ b/src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml
@@ -110,6 +110,9 @@ assets:
     cursor_column: updated_at
   - table_name: gradebook_grades
     cursor_column: updated_at
+    # points is an unbounded Postgres numeric holding more than 9 decimal
+    # places, which the default decimal128(38, 9) cannot represent.
+    widen_unbounded_numeric: true
   - table_name: gradebook_assignments
     cursor_column: updated_at
   - table_name: gradebook_assignment_types


cbini:
Verdicts on the `claude-review` findings above.

**Findings 1, 2, 3, 4, 6 — agreed, no action.** Each confirms a claim the PR already makes, and each traces the real code path rather than the diff alone.

**Finding 5 — confirmed, and it is wider than the review could see.** The review lacked warehouse access. I have it, so here is the answer.

dlt stores its full reflected schema in `dagster_kippmiami_dlt_focus._dlt_version.schema`. Reading the newest row (version 356, written 2026-08-28, before the failure) gives every column's precision and scale hints:

| precision, scale | columns | what it means |
| --- | --- | --- |
| 38, 9 | 191 | dlt's default for an unbounded `numeric` — at risk |
| 76, 38 | 16 | already widened, all in `student_gpa_calculated` |
| real values (3,0), (7,2), (1,0), ... | 9 | declared precision in Postgres — safe |

`gradebook_grades.points` sits in the 191, which is what validates the marker: it is the column that just broke.

So `gradebook_assignments.points` is at risk — the review guessed right. It is 1 of 191 unbounded columns across 45 tables. The full list is in the issue thread.

I am not adding those tables to `widen_unbounded_numeric` in this PR. Opting a table in retypes it to BIGNUMERIC, which needs a `refresh: drop_resources` reload and a `cast(... as numeric)` in its staging model. Doing that for 45 tables is a migration, not a bug fix, and it belongs in its own change.

One more measurement, because it changes how urgent the rest are. Across the 6 highest-risk tables, the widest fractional part in any loaded value is 3 decimal places (`course_weights.credits`). `gradebook_grades.points` normally holds 1. The value that broke the sync is an outlier, not a trend. That points at rare bad writes rather than a column that is chronically dangerous, so the fix worth building is isolation, not a rush to opt every table in.

PR #5078, opened 2026-08-31
feat: scope Grow admin access by region and ADP job function

### LINKED ISSUE #5052: feat: scope Grow admin access by region and ADP job function
## What and why

Grow gives 75 people network-wide Sub Admin today. That role reaches every school in Newark, Camden and Miami. Most of those people lead one region, not the whole network.

Region is not an input anywhere in the pipeline. The extract has no region column. Grow runs as one district covering all 28 active schools, and the region field on Grow schools is empty or null for every school.

The role rule keys on job titles. Three consequences:

- The whole Human Resources department gets network-wide access at any title, including 5 Specialists and 1 Specialist Temporary.
- All 13 Achievement Directors get network-wide access, though each supports schools in one region.
- An admin role replaces the Coach role instead of adding to it, because the role expression wraps the two branches in a `coalesce`. 116 people who manage teachers lost Coach as a result, and none of them can be observed by their own manager.

A separate gap: users who match no rule are dropped at an inner join, so the sync never revokes. 21 people hold admin roles the extract does not manage, 5 of them terminated.

## Proposed approach

Four sub-projects, shipped in order. This issue covers sub-project 1.

**1. Role assignment correctness.** Rewrite the role rule to use ADP `job_function` as the tier and a department list as the gate. Stop assigning Sub Admin entirely. Make Coach additive so it stacks with an admin role.

**2. Region scope delivery.** Write `regionalAdminSchools` and `readonly` from `grow_user_sync`. Verified by spike: the Grow user PUT merges, so fields the payload omits are preserved.

**3. Revoke path.** Emit no-role users instead of dropping them, so the sync can strip roles and archive.

**4. School coverage.** Placeholder locations, the Poinciana name mismatch, multi-school leaders, and per-coach observation groups so a coach-teacher stops seeing their peers.

### Role matrix for sub-project 1

Gate departments: Teaching and Learning, School Support, Teacher Development, New Teacher Development, Special Education, School Leadership, Leadership Development, KIPP Forward, Special Projects, Executive.

Human Resources does not pass the gate. Employee Relations was considered and deliberately excluded: the team rarely uses Grow and can get the same information elsewhere. This also removes the need to identify that team by reporting line, which ADP cannot express.

| job_function | Role | Scope |
| --- | --- | --- |
| Chief Level | Regional Admin plus readonly | all 28 schools |
| EDs, HOSs, MDOs | Regional Admin plus readonly | their region |
| KTAF or Regional Managing Director | Regional Admin plus readonly | their region |
| KTAF or Regional Director | Regional Admin plus readonly | their region |
| School Leader | School Admin | home school |
| Assistant School Leaders | School Assistant Admin | home school |
| Deans | School Assistant Admin | home school |
| Teacher, Teacher in Residence | Teacher | home school |

Coach is additive on top of any of the above.

### Expected effect

- 62 Sub Admin assignments become 47 Regional Admin and 15 removals. No row emits Sub Admin.
- The 15 who lose access are the Executive Assistant and all 14 Human Resources staff.
- About 113 admins regain the Coach role they lost to the `coalesce`.
- 30 people new to Grow gain a role: 17 Assistant Deans in Student Support, 11 regional leaders, 2 network leaders.
- 4 people lose School Assistant Admin, because their job title matched the old substring rule but their job function does not place them in the Deans or Assistant School Leaders tier.

<details>
<summary>For Claude</summary>

> Context, not instructions.

Files:

- `src/dbt/kipptaf/models/extracts/schoolmint/rpt_schoolmint_grow__users.sql` — the role CASE, the `group_type` CASE, and the `instructional_managers` CTE all change here.
- `src/dbt/kipptaf/models/extracts/schoolmint/properties/rpt_schoolmint_grow__users.yml` — column docs and the two model-level tests.
- `src/teamster/code_locations/kipptaf/level_data/grow/assets.py` — read only for sub-project 1. The user payload is built around line 128 and the school payload around line 196.

Decisions already made:

- Treat a job title containing "Associate Director" as the Director tier. ADP records three Special Education Associate Directors as `KTAF or Regional Staff`, which understates them.
- 38 active staff have a null `job_function`. The admin tiers need a job-title fallback, mirroring the one the teacher predicate already has.
- Talent Acquisition and Human Resources do not pass the gate. Student Support does not pass the gate, but its Deans still qualify on tier.

Watch out for:

- The `instructional_managers` CTE has an operator precedence bug. `A and B or C` parses as `(A and B) or C`, so the second branch applies no assignment status filter.
- `contains_substr(job_title, 'Leader')` in the current rule matches "School Leader" and fires before the School Leader branch. Do not carry that pattern forward.

Acceptance criteria:

- No row emits Sub Admin.
- A user who manages teachers and holds an admin role emits both roles.
- `array_length(role_ids)` equals `array_length(role_names)` on every row, and both stay non-empty.
- The counts above hold: 15 lose access, 30 gain access, 47 become Regional Admin.

</details>

### LINKED ISSUE #5054: feat: scope Grow observation groups per coach instead of one group per school
## What and why

A coach who is also a teacher can see every other teacher at their school, including peers they do not coach. That is awkward and it is not what anyone intends.

The cause is that every school has exactly one observation group. `grow_user_sync` builds it like this:

```python
observees = [u["user_id"] for u in school_users if "observees" in u["group_type"]]
observers = {u["user_id"] for u in school_users if "observers" in u["group_type"]}
coaches = {u["coach_id"] for u in school_users if u["coach_id"] is not None}

payload["observationGroups"] = [
    {
        "_id": teachers_observation_group["_id"],
        "name": "Teachers",
        "observees": observees,
        "observers": list(observers | coaches),
    }
]
```

One group, named "Teachers", holding every teacher at the school as an observee and every coach and admin as an observer. Membership is a flat pool, so every observer can see every observee. There is no pairing between a coach and the people they actually coach.

Confirmed in the data: all 28 active schools have exactly one observation group, and it is named "Teachers" at every one of them.

## Proposed approach

Emit one observation group per coach rather than one per school.

`School.observationGroups` is an array and the schema already supports this. An `ObservationGroup` is just a name, an id, an observee list and an observer list. Every extract row already carries `coach_id`, derived from ADP `reports_to_employee_number`, so grouping by it is a regroup of data the sync already sends rather than new data.

Each group would hold one coach as the observer and only that coach's own reports as observees. A coach who is also a teacher then appears as an observee in their own coach's group and as an observer only in their own, so peers stop seeing each other.

**Unverified and needs a probe first.** Grow returns `observationGroups` on a GET, but we have never written more than one element. Two things to establish before designing further:

1. Does a PUT containing a group with no `_id` create that group, or is a separate call needed?
2. Does the Grow UI behave sensibly with many groups per school rather than one?

The same observational approach that settled the user PUT merge question on #5052 may work here, otherwise it needs one deliberate API call against a test school.

## Also in this sub-project

These all live in the same school PUT and are cheaper to fix together than separately.

- **Placeholder locations.** Staff whose home work location is `Room 9`, `Room 10` or `Room 11` never appear in a real school's `school_users`, so they land in no observation group anywhere. Keyna McClinek and Deanna Applewhaite are both in `Room 11`.
- **Name mismatch.** 4 staff have a home work location of `KIPP Miami - Poinciana Campus`, but the Grow school is named `Poinciana Campus`. The join is on exact name equality, so they are dropped entirely.
- **Multi-school leaders.** `school_users` is filtered to a single `school_id` per user, so a School Admin or School Assistant Admin covering two campuses is written to only one. The `coaches` union at line 206 already walks the reporting line to reach other schools, but nothing equivalent exists for the `admins` and `assistantAdmins` lists.

<details>
<summary>For Claude</summary>

> Context, not instructions.

Files:

- `src/teamster/code_locations/kipptaf/level_data/grow/assets.py` — the school payload is built from roughly line 196 to line 232.
- `src/teamster/libraries/level_data/grow/schema.py` — `ObservationGroup` is at line 99, `School` at line 624.
- `src/dbt/kipptaf/models/extracts/schoolmint/rpt_schoolmint_grow__users.sql` — supplies `coach_id`, `school_id` and `group_type`.

Depends on #5052 landing first. That issue makes `group_type` additive and changes who is an observee, so building on the current CASE would mean redoing this.

The dbt models `int_schoolmint_grow__schools__observation_groups`, `__observees` and `__observers` are all disabled in their properties YAML, and `stg_schoolmint_grow__schools` does not select `observation_groups`. The sync reads groups straight from the API instead. Re-enabling that lineage would give a way to verify the result in the warehouse.

</details>

cristinabaldor:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop Grow from handing out network-wide admin access by default, and give regional leaders access scoped to their own region instead.

Today 75 people hold `Sub Admin` in SchoolMint Grow. That role reaches every school in Newark, Camden and Miami. Most of those people lead one region, not the whole network. Region was never an input anywhere in the pipeline: the extract had no region column, and Grow's own school region field is empty on all 28 schools.

Five changes ship together, because none of the first four is safe alone.

**Roles now come from the ADP job function, gated by department.** The old rule matched job titles, which handed network-wide access to the entire Human Resources department at any title, and to all 13 Achievement Directors even though each supports one region. `Sub Admin` is no longer assigned to anyone.

**Regional Admin now carries real school scope.** The sync writes `regionalAdminSchools` and the `readonly` flag, so a regional leader sees their own region and cannot change configuration. Without this, the first change would grant a role over zero schools.

**The sync can now revoke.** A user who matches no role rule was previously dropped at an inner join, so the sync never saw them and never removed anything. 21 people held admin roles the extract did not manage, 5 of them terminated. Those users are now emitted with empty role arrays so their roles get stripped.

**Each coach gets their own observation group.** Every school had a single group holding every teacher as an observee and every coach as an observer, so a coach who also teaches could see peers they do not coach. Each coach now gets a group containing only their own reports.

**New Teacher Network Coordinators become Regional Observers scoped to their own school.** 25 coordinators (18 Newark, 4 Camden, 3 Miami) gain the `Regional Observer` role on top of whatever their job function already earns them. This is the one role not derived from an ADP job function: it reads the ADP membership feed. Scope is the coordinator's own campus rather than their region, because an NTNC coaches the new teachers at one school. The 3 Paterson coordinators fall out on the existing region filter, which is correct -- Grow holds no Paterson schools.

## Reviewer Notes

**The ADP data fix is a hard gate on merging.** Four active staff have a null `job_function` and a job title that clearly implies one. The design deliberately does not paper over a null with a title match, so those four receive no role until ADP is corrected. Employee numbers are in the linked issue. Once the revoke path is live, a null `job_function` means losing Grow roles outright rather than merely going unmaintained.

**Two access changes still need sign-off.** 18 active staff whose job function is `School-based Non-Instructional Staff` hold the `Teacher` role today and lose it, so they stop being observable. About 18 more across several job functions hold `Coach` and lose it. Both are recorded in the spec but neither has been approved by anyone.

**The `regionalAdminSchools` payload shape is now probed, not inferred.** Bare id strings are correct and the surrogate keys converge. Probed 2026-09-02 against the Grow test account (`internalId` 999999, which `stg_schoolmint_grow__users` excludes): the PUT is accepted, and the paginated `GET /external/users` that `grow_users` ingests hydrates the ids into the declared name/id struct, so `regional_admin_school_ids_ws` reads back correctly. Two traps, recorded in the spec because both make a correct implementation look broken -- `GET /external/users/{id}` returns the ids unhydrated, so a single-record readback looks like permanent non-convergence; and the published response examples show bare strings for a shape neither endpoint actually returns. Sending structs instead is rejected outright with a 500.

**Scope replacement, not merge.** 61 users hold a hand-maintained `regionalAdminSchools` today but are not Regional Admin under the new rules, so their scope is emptied. About 7 genuine Regional Admins have their scope narrowed. The narrowing is worth an eyeball before the first sync run.

### General

`dagster definitions validate` cannot run in a Codespace: the `kipptaf` location fails to import on unset Illuminate dlt settings, which is documented behaviour. Verified instead with `py_compile` plus hand-tracing of the grouping logic. Needs a real check in CI.

- [x] **Breaking change?** No column is renamed or removed. Four columns are added to a contracted extract, and two to a contracted staging model, all with `data_type` entries.

## CI checks

- [ ] **Trunk** — `trunk check --force` was run locally on all five changed source files and reports no issues.
- [ ] **dbt Cloud** — `dbt build --select rpt_schoolmint_grow__users` passes locally with 12 tests. CI builds `--full-refresh` against deferred upstreams, which is a harsher path.
- [ ] **Dagster Cloud** — expected to run, since a code location changed.

<details>
<summary>For Claude</summary>

Design: `docs/superpowers/specs/2026-08-28-grow-region-scoped-admin-design.md`. Two implementation plans in `docs/superpowers/plans/`, dated 2026-08-28.

Two API facts were established by probe against a real school, and both are load-bearing:

- The user PUT **merges**. A field the payload omits is left untouched. This is why `readonly` and `regionalAdminSchools` drifted by hand for years, and why sending them now brings them under management.
- The school PUT **replaces** `observationGroups`, `admins` and `assistantAdmins`. A group or member omitted from the payload is deleted. A PUT entry carrying a `name` and no `_id` creates a group and Grow assigns it an id.

Consequences that shaped the design:

- Group identity is keyed on `_id`, never on name, and the payload emits every existing group — those no longer wanted are sent emptied rather than dropped, so nothing recorded against a group is ever orphaned.
- Coach group names are `Coach {employee_number} - {display name}` and are matched on the employee-number prefix, so a rename relabels the group without churning its id. The prefix fallback is gated to names starting with `Coach `, so a hand-made group cannot be hijacked.
- A coach who is inactive, `readonly`, or no longer carries an observer role cannot anchor a group; their reports fall back to the school's `Teachers` group rather than into a group nobody can act in.

Verification against production data, not fixtures: 46 users move from `Sub Admin` to `Regional Admin`, 16 lose access, zero rows emit `Sub Admin`, about 114 admins regain `Coach`, 18 teachers with a null `job_function` drop out. Of the 545 no-role rows the revoke path emits, most hold a Grow role literally named `No Role`, so the substantive revoke is closer to 335 accounts.

One defect was found only by querying the built model against production, after every unit test passed and the model built clean: the revoke path made 333 active staff with no roles and no Grow account eligible for the sync's create branch, which would have opened 333 empty accounts on the first run. Fixed in the extract's final `WHERE`, which now emits a row only when the sync has something to do with it.

A new singular test, `rpt_schoolmint_grow__users__locations_resolve_to_grow_school`, guards the school join. That join is an exact name match and is load-bearing: a school renamed in the Grow UI silently drops all its staff from the extract, and the replacing school PUT then strips them from all three membership lists. The test fails against a stale development roster and passes against production, where zero locations are unresolved.

A related mismatch was fixed in data rather than code during this work: one Miami campus was renamed in Grow and an alias row added to the location crosswalk sheet, so the either-name join an earlier draft of the spec proposed was never implemented. The spec records this.

</details>

---

Closes #5052
Closes #5054


cristinabaldor:
## Review response — 7 accepted, 5 declined

Each finding verified against the code and the warehouse before acting. Fixes in `6eaba4dd9`.

### Accepted and fixed

**1 — terminated manager can be written into a school's admin list.** Correct, and the reasoning about `role_names` not being cleared by termination is right. I had measured this as zero in current data and parked it during development, which was the wrong call: a measurement is not a guard. Added `manager["inactive"] == 0` to the condition, matching the check `school_users` and `_can_anchor_group` already make.

**2 — missing PII tags.** Confirmed against the precedent cited. Added `contains_pii: true` to `user_name`, `user_email`, `user_name_ws` and `user_email_ws`. Left employee numbers and ids untagged: the staging precedent in this repo is deliberately narrower than the full FERPA list, and widening it is a separate decision.

**3 — multi-campus reach not extended to the Teachers group's observers.** Not intentional. The asymmetry was real: an admin reaching a satellite campus through the reporting line was not an observer of that campus's fallback group. Extended `school_observers` the same way, reusing `_can_anchor_group` rather than writing a second capability predicate. Inert today, since the Teachers group currently holds no observees, but correct when one appears.

**4 — no unit coverage on the two pure helpers.** Correct, and the point that it would have caught finding 1 lands. To answer the question the job could not: the module imports cleanly without the dbt manifest, so this was always feasible. Added `tests/assets/test_grow_observation_groups.py`, 9 cases covering exact match, the renamed-coach prefix fallback, the non-Coach name that must NOT fall back, the claimed-id skip, the None case, and each of the three capability checks.

**6 — folded scalar in the multi-line test expression.** The most valuable finding here. `src/dbt/CLAUDE.md` lines 561-566 document this exact failure: trunk-fmt reflows past 80 characters and the folded scalar collapses the inserted newline inside a quoted SQL string, producing `Unclosed string literal` at test runtime. It was one reflow away from a broken test. Changed to a literal block; it was the only multi-line expression in the file.

**10 and 11 — stale placeholders and missing descriptions.** Both done: the two dead comment blocks are gone and the two new columns have descriptions.

### Accepted as convention, not as a lint failure

**5 — ST09 join order.** The convention is real and I have applied it: the predicate now reads `p.region = gs.region`. But the claim that this is a live CI concern does not hold. ST09 is not in `exclude_rules`, yet `trunk check --force` reports no issues on this file both locally and in the passing Trunk check on this PR — sqlfluff does not treat the predicate as a join-order case because it sits inside an `OR` alongside a non-join condition. Worth flagging so the next reader does not go looking for a failure that is not there.

### Declined

**7 — Regional Admin with an empty scope.** Does not reproduce. Ran the suggested query against the built model: zero rows. Every region resolves to at least six schools and `Chief Level` resolves to all 28. The concern was reasonable given `[Training School]` has no region, but that affects only the school-side list, never a person's home location. The new test is a genuine guard rather than a latent failure.

**8 — collapse the repeated array-membership checks in `group_type`.** Declining. Behaviour-neutral churn on an expression that has already been through two rounds of review specifically about its membership semantics. The repetition is legible and the risk of touching it outweighs the tidiness.

**9 — subquery against a CTE.** Pre-existing, carried over unchanged, and out of scope for this PR. Worth its own change if someone wants it.

**12 — a manager whose Grow account is created in the same run.** Correct that their reports fall to the Teachers group for one day and self-correct the next. That is inherent to the extract being built before the sync runs, and pre-dates this PR. Noted, not fixed.

### Verification after the fixes

`dbt build --select rpt_schoolmint_grow__users`: 12 pass, 0 errors. `pytest tests/assets/test_grow_observation_groups.py`: 9 pass. `trunk check --force` on all changed files: no issues.

One thing the review could not see, flagged for whoever merges: four active staff have a null `job_function` with a job title that plainly implies one, and the design deliberately does not paper over that with a title match. Those four receive no role until ADP is corrected, and once the revoke path is live a null means losing Grow roles outright. Employee numbers are in the linked issue rather than here.

cristinabaldor:
@claude update your review

cristinabaldor:
## Review response — 1 accepted in part, and a coverage gap in the review itself

### Finding: coach-group name collision (Medium) — accepted in part, fixed in `d0cda7820`

The code claim is correct. `_match_observation_group`'s fallback is a bare
`group_name.endswith(match_key)` with no requirement that the group look like a
coach group, so the docstring's "a hand-made group can never be claimed by
accident" is true only for the `match_key is None` branch. That sentence was
wrong and is now fixed.

Declined the shape gate, because the exposure measures to zero:

| Check | Result |
| --- | --- |
| Grow observation groups whose name ends in parenthesised digits | 0 of 585 |
| Active-staff employee numbers | all 6 digits, 100001-402082 |
| Employee numbers in the year-like range 1900-2100 | 0 |
| Employee numbers under 5 digits | 0 |

So the example in the finding cannot happen: there is no employee `2024`, and a
group named `New Hires (2024)` has no live employee number to collide with. A
real collision needs a hand-made group whose name ends in a bare six-digit
number that is also a live employee number — which is the coach-group naming
convention written by hand.

What shipped instead of a gate:

- The docstring now states the actual guarantee, names the suffix-match
  behaviour and the membership replacement it implies, records the measurement
  above with its date, and names the condition that would invalidate it
  (employee numbering narrowing to four digits).
- `test_match_observation_group_claims_any_group_with_the_key_suffix` pins the
  claiming behaviour rather than asserting it away, so adding a shape gate later
  has to change that test deliberately.

Happy to add the gate if you'd rather not carry the residual risk at all — it is
a one-line `re.fullmatch` — but it narrows nothing measurable today.

### The updated review did not cover the newest commits

Flagging this so the green review isn't read as broader than it is. The updated
review covers the previous round plus `c108c30dd`, and does not mention the
`Regional Observer` role, its ADP membership source, the reworked
`regional_scope` CTE, or `unit_grow_users_ntnc_regional_observer` — all of which
landed in `83e8659f9`, before that review ran. Re-firing the review via a draft
toggle so it sees them, since `claude-review` does not run on `synchronize`.

### One item from the earlier round is now resolved by probe, not inference

The reviewer notes for this PR flagged `regionalAdminSchools` as "inferred rather
than probed". That is now probed and recorded in the design doc (`a5eaa7ee2`):
bare id strings are correct and the surrogate keys converge. Two traps worth
knowing, because both make a correct implementation look broken —
`GET /external/users/{id}` returns the ids unhydrated, so a single-record
readback looks like permanent non-convergence, while the paginated
`GET /external/users` that the ingest actually reads hydrates them into the
declared name/id struct. Sending structs instead is rejected with a 500.

cristinabaldor:
## Review response — round 3, no code changes needed

### Finding 1 (`is_current` inclusive BETWEEN) — declined, premise doesn't hold

Ran the suggested query plus the boundary dates:

| Check | Result |
| --- | --- |
| Associates with >1 current membership row | 1 (`104547Z5C`) |
| ...and its two rows | different descriptions — `New Teacher Network Coordinator` and `Teacher Leader` — not two NTNC rows |
| Associates with >1 current NTNC row | **0** |
| NTNC cohort windows | prior `2025-07-30` to `2026-07-01`; current `2026-08-01` to `2027-07-01` |

The two NTNC cohorts do not abut — there is a **one-month gap** between the
prior cohort's `expiration_date` (2026-07-01) and the incoming cohort's
`effective_date` (2026-08-01), so no calendar day satisfies the inclusive
`BETWEEN` for both.

The finding reads the commit message's "the prior cohort's expiration date lands
on day one of the next academic year" as describing the two cohorts sharing a
boundary. It describes something else: the fiscal-year bucketing in
`int_adp_workforce_now__employee_memberships_by_year`, whose
`date_to_fiscal_year(start_month=7)` assigns 2026-07-01 to AY2026. That is why
the roster's `memberships` string carries last year's coordinators into this
year's bucket (46 people rather than 29), and it is the reason the new CTE reads
`is_current` off the staging model instead of the roster string. So the
bucketing artifact the finding worries about is the one this commit exists to
avoid.

Independently, `new_teacher_network_coordinators` is
`select distinct associate_id ... where membership_description = '...' and
is_current`, so even a genuine duplicate row could not fan out the join or
double-count anyone.

Agreed the abutting-date convention in `src/dbt/CLAUDE.md` is the right thing to
check here — it just does not fire on this data.

### Finding 5 (confirm the dbt build) — confirmed green

`kipptaf` dbt CI finished after the review ran: run `70403223228261` on
`d0cda7820` is **success with zero warnings**
(`get_job_run_error warning_only=true` returns
`total_warnings: 0`, no test, freshness, or log warnings). That is the third
consecutive clean dbt Cloud run on this branch. Every other check on
`d0cda7820` is also green, including the full `dagster-cloud-deploy` chain.

One artifact worth naming so it isn't misread in the check history: the first
`dagster-cloud-deploy` run on `d0cda7820` shows `cancelled`. That was the draft
toggle used to re-fire this review — the `ready_for_review` event started a
replacement run, which passed. Not a build failure.

### Findings 2, 3, 4 — no action

Confirmations, and the `group by 1` nitpick was correctly self-dismissed against
the sibling singular tests. Nothing to change.

src/dbt/kipptaf/models/extracts/schoolmint/rpt_schoolmint_grow__users.sql
```diff
@@ -1,130 +1,185 @@
...
+                where p.is_regional_admin
+
+                union distinct
+
+                /* Regional Observer: their own school only */
+                select p.user_internal_id, gs.school_id,
+                from people as p
+                inner join grow_schools as gs on p.school_name = gs.school_name
+                where p.is_regional_observer
+            )
```
cbini: unnest subquery

src/dbt/kipptaf/models/extracts/schoolmint/rpt_schoolmint_grow__users.sql
```diff
@@ -1,130 +1,185 @@
...
+            unnest(
+                [
+                    if(p.is_regional_admin, 'Regional Admin', null),
+                    if(p.is_school_admin, 'School Admin', null),
+                    if(p.is_school_assistant_admin, 'School Assistant Admin', null),
+                    if(p.is_coach, 'Coach', null),
+                    if(p.is_regional_observer, 'Regional Observer', null),
+                    if(p.is_teacher, 'Teacher', null)
+                ]
+            ) as rn
```
cbini: unclear what's going on here. business logic should be minimal in joins. could this be extracted to a cte?

src/dbt/kipptaf/models/extracts/schoolmint/rpt_schoolmint_grow__users.sql
cbini: Moved to its own CTE in a7e0b018cf: `regional_scope_schools` holds the two-branch union, and `regional_scope` just aggregates it. Row-identical (1140 rows, 0 diffs either way).

src/dbt/kipptaf/models/extracts/schoolmint/rpt_schoolmint_grow__users.sql
cbini: Extracted in a7e0b018cf. A `people_role_slots` CTE builds the six-slot array from the boolean flags in `people`, and `people_roles` now unnests a plain column. Null slots drop at the `array_agg ... ignore nulls`. Row-identical (1140 rows, 0 diffs either way).

PR #5079, opened 2026-08-31
fix(adp): retry any non-JSON 404 as a gateway error

### LINKED ISSUE #5076: fix: ADP Workforce Now gateway 404 fails without retrying
## What's happening

The `kipptaf/adp/workforce_now/workers` asset failed on partition `07/30/2026`. ADP's API returned a 404. The 404 came from ADP's edge gateway, not from ADP's API, so the request should have been retried. It was not. The step made one request and failed.

The resource decides whether a 404 is retryable by looking for the text `default backend` in the response body. This 404 body was an openresty HTML page instead, so the check did not match. The code treated it as a real client error and raised a non-retryable exception.

The same request worked on the 5 nights before, returning 4749 records. Thirty other partitions ran fine the next night. Nothing about the request changed.

No data was lost. The partition still holds its August 28 write. It has since aged out of the schedule's 45-partition window, so it will not re-run on its own.

## Steps to reproduce

1. Call `hr/v2/workers` while ADP's gateway briefly has no healthy backend.
2. The gateway answers 404 with an openresty HTML page.
3. The step raises `AdpWorkforceNowError` right away instead of retrying.

## Where

- **Code location / dbt project:** `kipptaf` (Dagster)
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** https://kipptaf.dagster.cloud/prod/runs/19266296-c305-47cb-b747-7c9eb6921b42 (auto-retry `1aa79b58-c5b8-4c3a-b79e-de6105d5b9af` failed the same way 62 seconds later)

<details>
<summary>For Claude</summary>

Root cause is `is_transient_gateway_404` in `src/teamster/libraries/adp/workforce_now/api/resources.py`, which gates the retry on the literal substring `default backend`.

The real signal is the body format, not a substring. ADP's API layer answers with JSON; the gateway answers with HTML or plain text. Both known bad bodies are non-JSON: `default backend - 404` and the openresty 404 page.

Fix: in the `HTTPError` handler, when `response.json()` raises `JSONDecodeError` on a 404, log a warning and re-raise the `HTTPError` so tenacity retries. Other non-JSON 4xx keep failing fast with the response text as the detail.

This flips `test_request_non_json_client_error_raises_adp_error` in `tests/resources/test_resource_adp_workforce_now.py`, which asserts a non-JSON 404 fails fast. Retarget it to a non-JSON 400 so it still covers the JSON-parse guard, and add a regression test using the openresty body.

Fixed means: an openresty 404 retries 5 times and exhausts as a `RetryError` wrapping `HTTPError`; a JSON 404 still raises `AdpWorkforceNowError` after 1 request.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make the ADP Workforce Now resource retry a
gateway 404 instead of failing on the first request.

ADP's edge gateway answers with a 404 when it briefly has no healthy backend.
The resource decided whether a 404 was retryable by looking for the text
`default backend` in the body. On [run 19266296](https://kipptaf.dagster.cloud/prod/runs/19266296-c305-47cb-b747-7c9eb6921b42)
the gateway returned an openresty 404 page instead, so the check did not match.
The code treated it as a real client error, raised a non-retryable exception,
and the step died after one request. The auto-retry failed the same way 62
seconds later.

The fix matches on the body format instead of a substring. ADP's API answers
with JSON. The gateway does not. So a 404 whose body is not JSON came from the
gateway, and the resource now re-raises the `HTTPError` for tenacity to retry. A
genuine 404 carries a JSON body and still fails fast after one request.

Closes #5076

## Reviewer Notes

- The retry predicate widened from one substring to "any non-JSON 404". A
  genuine 404 that arrived without a JSON body would now burn 5 attempts before
  failing. It still fails.
- `test_request_non_json_client_error_raises_adp_error` moved from a 404 to a
  400. A non-JSON 404 is now the gateway case, so the old assertion contradicts
  the fix. The test still covers the JSON-parse guard.
- The two existing gateway-404 fakes now set `json_raises=True`. The real
  gateway body is not JSON, and the old fakes let `response.json()` succeed.

### General

      code location

Skipped — no dbt changes.

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Diagnosed with `superpowers:systematic-debugging`.

Evidence the 404 was the gateway and not a real resource 404:

- Partition `07/30/2026` materialized on the 5 nights before, most recently with
  4749 records, from the identical request URL.
- 30 adjacent partitions succeeded the next night against the same endpoint.
- The response body was an openresty HTML page. ADP's API returns a JSON error
  envelope, which is what the `response.json()` path assumes.
- Both attempts made exactly 1 request each, because the non-retryable branch
  bypassed tenacity entirely.

No data was lost. The partition kept its August 28 write. It has since aged out
of the schedule's 45-partition window (`partition_keys[-45:]` against an
`end_offset=14` daily definition), so nothing re-runs it.

One implementation trap worth knowing: a bare `raise` inside the nested
`except JSONDecodeError` re-raises the `JSONDecodeError`, which is not in the
tenacity predicate, so the retry silently does not fire. The final shape sets an
`is_json_body` flag and raises from the outer `except HTTPError` scope, which
also avoids ruff B904.

Verification:

- `uv run pytest tests/resources/test_resource_adp_workforce_now.py` — 13 pass.
  `test_get_talent_associate_memberships` fails on `main` too; it calls the live
  ADP API.
- `uv run pytest tests/test_dagster_definitions.py` and
  `dagster definitions validate -m teamster.code_locations.kipptaf.definitions`
  pass in the main checkout. In the worktree they fail on a missing dbt manifest
  and a missing `ILLUMINATE_DB_DRIVERNAME` configuration value, both unrelated
  to this change. `test_definitions_kippmiami` fails on `main` as well.

</details>

### DIFF
--- a/src/teamster/libraries/adp/workforce_now/api/resources.py
+++ b/src/teamster/libraries/adp/workforce_now/api/resources.py
@@ -90,22 +90,8 @@ class AdpWorkforceNowResource(ConfigurableResource):
             # a 504 HTML page) are not JSON, and JSONDecodeError is not in the
             # retry predicate, so parsing here would convert a retryable error
             # into a non-retryable one.
-            #
-            # ADP's gateway also returns a transient "default backend - 404" when
-            # it briefly has no healthy backend (a load-balancer signature, seen
-            # mid-pagination). That is distinct from a genuine resource 404 —
-            # treat only the gateway variant as retryable so a real 404 still
-            # fails fast.
-            body = response.text
-            is_transient_gateway_404 = (
-                response.status_code == 404 and "default backend" in body.lower()
-            )
-            if (
-                response.status_code == 429
-                or response.status_code >= 500
-                or is_transient_gateway_404
-            ):
-                self._log.warning(msg=body)
+            if response.status_code == 429 or response.status_code >= 500:
+                self._log.warning(msg=response.text)
                 raise
 
             # other 4xx are deterministic client errors — surface a specific
@@ -113,8 +99,20 @@ class AdpWorkforceNowResource(ConfigurableResource):
             # parse: a 4xx body may also be non-JSON.
             try:
                 detail = response.json()
+                is_json_body = True
             except JSONDecodeError:
                 detail = response.text
+                is_json_body = False
+
+            # ADP's API layer always answers with JSON, so a non-JSON 404 did not
+            # come from it: that is the edge gateway briefly without a healthy
+            # backend, seen mid-pagination as "default backend - 404" and as a
+            # plain openresty 404 page. Transient — re-raise the HTTPError so
+            # tenacity retries. A genuine resource 404 carries a JSON body and
+            # still fails fast.
+            if not is_json_body and response.status_code == 404:
+                self._log.warning(msg=detail)
+                raise
 
             self._log.error(msg=detail)
             raise AdpWorkforceNowError(detail) from e
--- a/tests/resources/test_resource_adp_workforce_now.py
+++ b/tests/resources/test_resource_adp_workforce_now.py
@@ -282,7 +282,7 @@ def test_request_non_json_client_error_raises_adp_error(
 
     The deterministic-4xx branch must tolerate a non-JSON body instead of
     crashing on ``response.json()``. Uses a 400 because a non-JSON 404 is the
-    gateway signature and is retried instead (see the openresty test below).
+    gateway signature and is retried instead (see the gateway-404 test below).
     """
     monkeypatch.setattr(AdpWorkforceNowResource._request.retry, "wait", wait_none())  # pyright: ignore[reportFunctionMemberAccess]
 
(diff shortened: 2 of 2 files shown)


cbini:
Verdicts on the review findings, verified against the code before replying.

**1. Test coverage gap — no genuine JSON-bodied 404. Accepted, fixed in 68e40aa.**

Confirmed: `grep -n "_FakeResponse(404" tests/resources/test_resource_adp_workforce_now.py` returned zero hits without `json_raises=True`. The fail-fast branch this fix depends on had no coverage. Added `test_request_json_404_raises_adp_error` — `_FakeResponse(404, {"error": "not found"})`, asserts `AdpWorkforceNowError` and exactly 1 call.

**2. Parametrize the duplicate gateway-404 retry tests. Accepted, fixed in 68e40aa.**

Confirmed the precedent claim: `pytest.mark.parametrize` is already used in `tests/resources/test_resource_ssh_banner_timeout.py` and `tests/resources/test_resource_google_directory.py`. Collapsed both into one parametrized `test_request_retries_on_transient_gateway_404` with ids `default_backend` and `openresty`. Net 22 lines removed, 27 added — the new JSON-404 test accounts for the difference.

**3. Dagster conventions — no issues. Nothing to do.**

**4. dbt conventions — not applicable. Nothing to do.**

**5. Blast radius of the widened predicate. Not a change, already disclosed.**

Independently confirmed the "nothing catches it downstream" claim: `AdpWorkforceNowError` appears in no file under `src/` other than the resource that defines it. The tradeoff stands as written in Reviewer Notes.

**6. Runtime verification not reproduced. Noted, no change.**

Ran locally on 68e40aa: `pytest tests/resources/test_resource_adp_workforce_now.py` gives 14 passed, 1 failed. The failure is `test_get_talent_associate_memberships`, which calls the live ADP API and fails on `main` too. `trunk check --force` on the changed files reports no issues.

Separately, on the previous commit `dagster-cloud-deploy / merge` failed with a `401 Unauthorized` fetching an Artifact Registry OAuth token while pushing the multi-arch manifest. That is CI infrastructure, not the diff. This push re-runs the workflow.

PR #5085, opened 2026-08-31
fix: cast Focus unbounded numeric columns back to numeric

### LINKED ISSUE #5080: feat: widen every unbounded Focus numeric and delete the per-table opt-in
## What and why

An unbounded Postgres `numeric` in Focus takes the whole KIPP Miami sync down the moment one value carries more than 9 decimal places. It has happened twice: `student_gpa_calculated.weighted_gpa` in #5021, and `gradebook_grades.points` in #5074, which cost 2 days of downtime across all 20 tables in the run.

191 columns across 45 tables can still do it. The list is in the #5074 thread.

Today the fix is opt-in. A table gets `widen_unbounded_numeric: true` in `focus.yaml` only after it breaks. That opt-in exists for one reason: widening retypes a column to BigQuery BIGNUMERIC, and the `replace` write disposition cannot retype a column in place, so an already-loaded table has to be dropped and reloaded.

This proposes paying that cost once, for every table, and deleting the opt-in.

The result is less code than today, not more. The YAML key, the `widen_numeric_tables` set, and the branch that routes between two type adapters all exist only to avoid this migration.

## Proposed approach

Two pull requests, then one manual run.

1. **dbt** — add `cast(<col> as numeric)` to 96 columns across 41 `stg_focus__*` models. No behavior change today, because those columns are already NUMERIC. Safe to merge on its own.
2. **Dagster** — apply the widening adapter to every table, and delete the opt-in machinery.
3. **Cutover** — after the Dagster deploy, launch one manual Focus run over all 79 tables with run config `refresh: drop_resources`.

Sensor ticks landing between the deploy and the reload fail harmlessly. A failed load keeps the old dlt baseline, so the table re-selects on the next tick.

<details>
<summary>For Claude</summary>

Design spec: `docs/superpowers/specs/2026-08-31-focus-source-wide-numeric-widening-design.md`.

Sizing, all measured rather than estimated:

- 191 unbounded columns across 45 tables, from `dagster_kippmiami_dlt_focus._dlt_version.schema` version 356. dlt writes `precision 38, scale 9` for an unbounded `numeric`; a declared one carries its real precision.
- 96 of those 191 are projected by a staging model and need a cast. 41 models are affected. 5 casts are on aliased projections, such as `custom_200000002 as experience_length_years` in `stg_focus__users.sql`.
- The 45 tables hold 799,554 rows and 173 MB, biggest table 371,411 rows.
- Focus has no `double precision` columns, so the adapter's `Float` guard is currently dead code. It stays, because source-wide widening makes it cover all 79 tables instead of 2.
- `discipline_incidents` has an unbounded column, is loaded, and has no staging model. Nothing to do there.

Key constraints:

- `dataset_name` is not branch-isolated. A branch deployment writes to the same BigQuery dataset as prod, so there is no rehearsal environment for the reload.
- Contracts are enforced on `focus.staging`, so a missed cast fails CI rather than breaking silently.
- Nothing reads the raw dlt dataset except the `focus` dbt source, and only `stg_focus__*` models read that. Every consumer therefore sees `numeric` at 9 decimal places either way. The migration buys crash-avoidance, not precision. Keeping the raw layer faithful to Postgres is the stated reason for choosing it over rounding in the query.
- The 2 `gradebook_grades` casts are part of the 96 and land in #5075. Subtract them if that merges first.

Out of scope: converging Illuminate's `unbounded_numeric_adapter` at `(38, 18)` with Focus's `(76, 38)`.

Done means every one of the 191 columns reads BIGNUMERIC, `dbt build --select package:focus` is green with contracts on, no `widen_unbounded_numeric` key remains, and one sensor tick runs clean end to end.

</details>

cbini:
# Pull Request

> **Write this whole PR in plain language.** Read `.github/PLAIN_LANGUAGE.md`
> and follow it: every sentence does a job, one idea per sentence, active voice,
> common words. Only the "For Claude" fold-out at the bottom is exempt.

## Summary & Motivation

"When merged, this pull request will..." add 94 `cast(<col> as numeric)`
projections across 40 `stg_focus__*` staging models. These casts are no-ops
today — the underlying columns are already `NUMERIC` in BigQuery. They exist to
protect a follow-up change: a separate PR (tracked by the same issue) retypes
191 KIPP Miami Focus raw columns from `NUMERIC` to `BIGNUMERIC` in the dlt
source. Without this cast in place first, that retype would break the enforced
dbt contracts on these staging models. Merge order matters: this PR must land
before the Dagster retype change goes in.

Refs #5080

## Reviewer Notes

The 94 edits are script-generated from the dlt reflected schema, not hand
written — each cast targets a column the retype will touch. Proof they don't
change behavior: `dbt build --select package:focus` with contracts enforced on
`focus.staging` produced identical results before and after the change
(`PASS=392 WARN=6 ERROR=0 SKIP=0 NO-OP=0 TOTAL=398`, both runs). The 6 warnings
are pre-existing relationship/uniqueness test warnings, unrelated to this
change. `trunk check --force --no-fix` on the touched directory also reported
no issues.

## Self-review

### General

      code location

### dbt

      (no new models; existing properties files unchanged)
- [x] Include (or update) an
      (no exposure changes needed)
      summary above.
      (not breaking — no column renamed or removed, casts are no-ops today)
      See

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. This is task 1 of 3 in the implementation plan
for issue #5080 (source-wide numeric widening for the Focus dlt source).

Exact scope: 94 `cast(<col> as numeric) as <col>` projections added across 40
files under `src/dbt/focus/models/staging/stg_focus__*.sql`. The cast list was
generated from a script reading the dlt reflected schema for the 191 columns
that PR 2 (a separate, not-yet-opened branch) will retype from `NUMERIC` to
`BIGNUMERIC` in the Focus source tables. Casting back to `NUMERIC` in staging
keeps each model's enforced dbt contract (declared column types) satisfied
after that retype lands, since dbt contracts fail the build on a type
mismatch, not just a `NUMERIC`/`BIGNUMERIC` widening at the source.

Verification performed (see task-1 report for full commands/output):

- Baseline `dbt build --select package:focus` (before the change):
  `Done. PASS=392 WARN=6 ERROR=0 SKIP=0 NO-OP=0 TOTAL=398`
- Post-change `dbt build --select package:focus` (contracts enforced on
  `focus.staging`): `Done. PASS=392 WARN=6 ERROR=0 SKIP=0 NO-OP=0 TOTAL=398`
  — identical to baseline.
- `trunk check --force --no-fix` on `src/dbt/focus/models/staging/`:
  `Checked 149 files` / `✔ No issues`.
- 3 diffs eyeballed by hand
  (`stg_focus__users.sql`, `stg_focus__students_join_people.sql`,
  `stg_focus__co_teacher_days.sql`) — cast placement and aliasing correct in
  each.

Sequencing note for whoever picks up PR 2: this PR (PR 1) must merge to `main`
first. PR 2 depends on these casts existing before it retypes the 191 raw
columns; merging in the other order would break the focus staging contracts in
production.

</details>


cbini:
## Review verdicts

Verified each finding against the code before acting. One fix pushed as `d8c4773`; the rest are no-change with reasons.

### Finding 1 — cast block ordering in 6 files (Medium) — ACCEPTED, fixed in `d8c4773`

The convention claim is correct. `src/dbt/CLAUDE.md` lines 1219-1233 put "Column enumerations (plain refs)" in tier 1, and an aliased plain ref like `custom_4 as course_sequence` is a plain ref. `cast()` is tier 3. The design doc's stated shape agrees.

Root cause: the generating script anchored on the last **bare** column reference (`^\s{4}\w+,$`), which does not match an aliased ref, so in 6 files the block landed mid-tier-1.

The fix applies a corrected rule rather than the suggested "move to immediately before `from`": **after the last plain-ref target, before the first non-plain-ref target.** A plain-ref target is bare (`updated_at,`), aliased (`custom_4 as course_sequence,`), or backtick-aliased (`` custom_200000005 as `language`, ``). That distinction matters for `stg_focus__student_report_card_grades.sql`, which carries a genuine `regexp_extract` / `safe_cast` block after its alias group — the cast block sits between the two there, not before `from`.

The 6 files and their post-fix placement:

| file | casts | placement |
| --- | --- | --- |
| `stg_focus__course_periods.sql` | 4 | alias group, blank line, casts, `from` |
| `stg_focus__courses.sql` | 3 | alias group, blank line, casts, `from` |
| `stg_focus__discipline_referrals.sql` | 1 | alias group, blank line, casts, `from` |
| `stg_focus__master_courses.sql` | 5 | split alias groups rejoined, blank line, casts, `from` |
| `stg_focus__student_report_card_grades.sql` | 5 | alias group, casts, expression block, `from` |
| `stg_focus__students.sql` | 2 | alias group (54 refs), blank line, casts, `from` |

`stg_focus__students.sql` deliberately places the casts immediately before `from`: its whole trailing group is plain and backtick-aliased refs, with no expression target anywhere in the select list.

The fix is a pure reordering. Insertions equal deletions in every file (10/10, 8/8, 4/4, 12/12, 12/12, 6/6), and the set of output column names per file is byte-identical to the prior commit — checked programmatically, not by eye.

### Finding 2 — missing blank line in `stg_focus__master_courses.sql` (Low) — ACCEPTED, fixed in `d8c4773`

Subsumed by Finding 1's fix. The two `custom_field_*` groups are now one contiguous group with a single blank line before the cast block.

### Finding 3 — cast count and column-name integrity — CONFIRMED, no action

Independently reconciled: 94 casts across 40 files, entry-for-entry against the generated cast map, including all 5 alias-preserving casts.

### Finding 4 — contract types spot-checked, not exhaustive — RESOLVED, no action

The full check you asked for was run locally. `dbt build --select package:focus --project-dir src/dbt/kippmiami` with contracts enforced on `focus.staging`: `PASS=392 WARN=6 ERROR=0 SKIP=0 TOTAL=398`, identical before the change, after the change, and after the ordering fix. Any of the 94 columns declaring something other than `numeric` would have failed a contract there. dbt Cloud CI on `d8c4773` is green with 0 warnings.

### Findings 5, 6, 7, 8 — model altitude, Dagster N/A, over-engineering, markdown — AGREED, no action

No disagreement on any of these.

### On the review environment note

Understood that `dbt build` and `trunk check` could not run in the review. Both ran locally on the fix commit: `trunk check --force --no-fix` over the 6 changed files reports no issues, with no ST06 and no LT02.

Refs #5080

Issue #4243, opened 2026-06-22
fix(dbt): fct_survey_responses PK uniqueness — survey_response_key duplicated ~18x (46,008 keys / 828,144 rows)


cbini:
## Summary

`unique_fct_survey_responses_survey_response_key` warns: **46,008** distinct `survey_response_key` values are duplicated, spanning **828,144** rows — a near-uniform **~18 rows per key** (max 18). `survey_response_key` is not unique at the fact's current grain.

## Bucketing

| metric | value |
| --- | --- |
| duplicated keys | 46,008 |
| total duplicated rows | 828,144 |
| max rows per key | 18 |
| avg rows per key | ~18 |

The near-uniform 18x multiplicity points to a **join fan-out at question/item grain** (or similar): `survey_response_key` identifies a submission/response, but the fact appears to carry one row per sub-item, so the surrogate alone cannot be the PK.

## Action

1. Determine the true grain of `fct_survey_responses`.
2. Either recompose `survey_response_key` to include the question/item dimension, or rename it and add the correct grain key, so the PK is unique.
3. Promote the test to `severity: error` once it resolves.

## Tracking context

- No open tracker existed for this PK-uniqueness test.
- #3921 (closed as **duplicate**) tracked a *different* test — the `survey_response_key` → `fct_survey_submissions` **relationships** orphan, not PK uniqueness.
- Part of the broader survey-marts grain work: #3890 (umbrella), #3918 (submission-grain intermediate).

## Surfaced by, not caused by, #4182

Unrelated to PR #4182 (staff `job_function`). It appeared in that PR's CI because the stale staging deferred manifest widened `state:modified+` to include the survey marts. Pre-existing.

PR #5086, opened 2026-08-31
perf(dbt): put 6 eager Focus grade tables on a cron

### LINKED ISSUE #5082: perf: 6 eager Focus intermediate tables rebuild 364x/day
## What's happening

6 intermediate models from the Focus GPA and gradebook work are `materialized: table` with no `cron_schedule`, so they inherit the eager `dbt_table_automation_condition()` and rebuild on any upstream data change. They rebuild 364 times a day between them, costing about 309 GiB/day.

Measured over 8/29 to 8/31, a weekend plus Monday, so almost no CI traffic:

| Model | Builds/day | GiB/build | GiB/day |
| --- | --- | --- | --- |
| `int_students__category_grades` | 85.0 | 1.667 | 141.7 |
| `int_students__gradebook_assignments_scores` | 45.3 | 1.458 | 66.1 |
| `int_students__course_enrollments` | 44.3 | 1.209 | 53.6 |
| `int_powerschool__course_enrollments_union` | 24.3 | 1.132 | 27.5 |
| `int_students__gpa` | 85.7 | 0.137 | 11.7 |
| `int_students__final_grades` | 79.7 | 0.107 | 8.6 |

They also drive churn in their PowerSchool ancestors: `snapshot_powerschool__gpa_term` 69.7/day (73.9 GiB/day), `kippnewark_powerschool.stg_powerschool__pgfinalgrades` 44/day (42.6), `snapshot_powerschool__gpa_cumulative` 73.7/day (37.4), `kippnewark_powerschool.base_powerschool__final_grades` 51.3/day (27.2).

Daily BigQuery spend went from about 1.0 TB on a quiet weekend day to about 1.5 TB, and the whole rise is the prod Dagster agent rather than CI.

Same failure mode as #4559, which found `int_assessments__response_rollup` rebuilding 125+ times a day, and as #4821 and #5035. It is the 4th time this pattern has landed.

## Proposed approach

The 6 models are not one group. Split them by how far downstream they reach:

| Group | Models | GiB/day | Downstream models | Exposures | Tightest consumer cron | Action |
| Leaves | `int_students__category_grades`, `int_students__gradebook_assignments_scores`, `int_students__gpa`, `int_students__final_grades` | **228.1** | 1 to 3 each | 1 (`cube_semantic_layer`, no cube file) | none | **cron `0 4 * * *`** |
| Spine | `int_students__course_enrollments`, `int_powerschool__course_enrollments_union` | 81.1 | **184** | **60** | `0 1`, `0 16` Fri, `0 18` (`ddi_suite`) | **cron `0 0,4,15,17 * * *`** |

The leaves feed only `fct_grades_category`, `fct_grades_assignments`, `fct_grades_gpa`, and `fct_grades_term`, which nothing currently reads. A cron there costs about 8.7 GiB/day against 228.1, and delays nothing anyone consumes. `0 4 * * *` matches the `gradebook_and_gpa_dashboard` batch those marts are eventually for.

The spine reaches much further -- `base_powerschool__course_enrollments` is a bare `select *` over `int_students__course_enrollments`, putting those 2 models upstream of 184 models and 60 exposures, including the hourly `ops_dashboard`. A column trace shows the hourly dashboard does not actually depend on them, so a cron is still safe. The binding consumer is `ddi_suite` at 3 ticks a day, so 4 ticks placed an hour ahead of each real consumer batch serve everything.

Expected result: 309 GiB/day down to about 18, a 94% cut, without delaying anything a consumer reads.

<details>
<summary>For Claude</summary>

### Decisions already taken, do not re-open

**Not views, even though all 6 are intermediate and the per-layer requirements permit it.** Views would cost less than a cron — rebuilds go to 0, and prod test runs stop too, since the data-change condition only re-materializes tables. The reason to decline is coverage, not cost.

Each leaf's uniqueness test is the SIS-cutover tripwire. All 4 are `full union all corresponding` of a PowerSchool branch and a Focus branch, split by a year floor from `int_students__sis_cutover`, with predicates that are exact complements:

```sql
-- PowerSchool branch (int_students__final_grades:54-58)
where not (fg._dbt_source_project = 'kippmiami'
           and fg.academic_year >= sc.focus_start_academic_year)
-- Focus branch (int_students__final_grades:131)
where g.academic_year >= sc.focus_start_academic_year
```

A `>=` slipping to `>`, the floor moving, or the `not (...)` guard being dropped makes one Miami academic year emit from both branches. Both carry `_dbt_source_project = 'kippmiami'`, so the uniqueness key catches it. That is a live guard on the one boundary actively changing during the SIS migration, and as views it would fire only on code change. A 2x/day cron keeps it firing for 8.7 GiB/day.

**The tests cannot move upstream to a table.** The invariant is about the seam between the 2 branches, and no upstream model has both branches present. Testing upstream would assert each branch's own grain and lose exactly the failure mode that matters. This is the inverse of #5032, where tests asserted what a `group by` already guaranteed and the real claim belonged upstream.

**Not `+enabled: false`.** Cheaper than either option, and it retires the same tripwire as views.

Caveat on scope: the complement-predicate structure is verified in `int_students__final_grades` and `int_students__gpa`. `int_students__category_grades` and `int_students__gradebook_assignments_scores` carry no `focus_start_academic_year` reference of their own — they inherit the split from their parents, so their tests guard a join fan-out rather than the seam. Slightly lower value, same conclusion.

### Method notes

Group by dataset, not just `table_id`. `base_powerschool__final_grades` exists in every district project, and grouping by table name alone reported 149 builds/day when the real split is kippnewark 51.3, kipptaf 68.3, kippcamden 24.7.

Exposure declarations overstate consumption. All 5 marts downstream of the leaves declare `cube_semantic_layer`, but `grep -rl` over `src/cube/` finds a cube file for `dim_student_section_enrollments` only. `fct_grades_category`, `fct_grades_assignments`, `fct_grades_gpa`, and `fct_grades_term` appear in no cube. 7 days of read traffic on all 5, excluding dbt and Codespaces, is entirely the prod Dagster agent at 0 bytes -- `create or replace view` DDL and test-audit writes, not reads. `cube-cloud@` does not appear. A bare `ref()` in that 60-entry exposure list is not evidence a consumer exists; `fct_assessment_scores_student_scoped` has the same shape and no cube either.

`~any_deps_in_progress` does help here, unlike in #5035. That guard reads direct deps only (`_build_dbt_condition` in `core/automation_conditions.py:90` applies a bare `~any_deps_in_progress()`; the recursive `_build_any_ancestor_*` helpers are trigger-side only). The leaves are direct parent-child with each other, so a shared tick genuinely serializes the pass. Give all 4 the same cron.

### Spine column trace -- the hourly veto does not hold

`ops_dashboard` at `0 * * * *` is reachable from the spine but does not depend on it. Reachability is not dependency, the same finding as the `int_powerschool__ada_term_pivot` work.

The spine reaches the intraday exposures by only 2 routes.

Route 1, through `int_extracts__student_enrollments`. That model joins `base_powerschool__course_enrollments` as `sip` and projects exactly 1 derived column from it, at `int_extracts__student_enrollments.sql:333`:

```sql
if(sip.students_student_number is not null, true, false) as is_sipps
```

An existence flag for course number `SEM01099G1`. That is the model's complete spine contribution, so `is_sipps` is a sufficient tracer for every consumer on this route. `rpt_tableau__ops_dashboard` and `rpt_tableau__attendance_dashboard` both reference `is_sipps` zero times and carry no `select *` or `dbt_utils.star`, so neither depends on the spine. `rpt_gsheets__kfwd_rem_roster`, the terminal model for `kipp_forward_data_suite`, likewise projects nothing spine-derived, though `int_kippadb__roster` upstream of it does carry a `select *` worth re-checking if that model changes.

Route 2, direct. `rpt_tableau__assessment_dashboard` refs `base_powerschool__course_enrollments` itself and projects 6 columns from it: `teachernumber`, `teacher_lastfirst`, `courses_course_name`, `sections_expression`, `sections_section_number`, `is_foundations`, plus `courses_credittype` inside a conditional. This is a real dependency with no table boundary in the path.

Genuine spine consumers, from the models that actually project a spine-derived column:

| Exposure | Cron | Route |
| --- | --- | --- |
| `ddi_suite` | `0 1 * * *`, `0 16 * * 5`, `0 18 * * *` | `rpt_tableau__assessment_dashboard`, `rpt_tableau__ddi_dashboard` |
| `apm_dashboard` | `0 6 * * *` | `rpt_tableau__iready_apm` |
| `literacy_dashboard` | `0 5 * * *` | `rpt_tableau__dibels_dashboard` |

`0 0,4,15,17 * * *` puts a build an hour ahead of each: 00:00 for ddi's 01:00, 04:00 for literacy's 05:00 and apm's 06:00, 15:00 for ddi's Friday 16:00, 17:00 for ddi's 18:00. 4 ticks at 2.341 GiB per full pass is about 9.4 GiB/day against 81.1 today.

Note the table boundary is not a freshness boundary. `int_extracts__student_enrollments` is a table, which makes the hourly dashboard's reads cheap, but it is eager, so its content is only as fresh as its upstreams. It decouples cost, not staleness.

### Acceptance

Builds/day for each of the 4 leaves falls to the tick count, verified in `JOBS_BY_PROJECT` grouped by dataset AND table; `CustomDagsterDbtTranslator.get_automation_condition()` on each parsed node returns a tree containing `0 4`, with a view-materialized control returning no cron; the 2 spine models still return an eager tree with no cron; and `dbt build --select int_students__category_grades+ int_students__gpa+ int_students__final_grades+ int_students__gradebook_assignments_scores+ --target dev --defer --favor-state` passes with prod row counts.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will put 6 Focus grade and enrollment tables on a
cron instead of rebuilding them every time anything upstream changes.

All 6 are `materialized: table` with no `cron_schedule`, so they inherit the
eager table automation condition. They rebuild 364 times a day between them and
cost about 309 GiB/day. This change is expected to bring that down to about 18
GiB/day, a 94% cut, without delaying anything a consumer reads.

The 6 are not one group, so they get 2 different crons.

The 4 leaves get `0 4 * * *`:

- `int_students__category_grades`
- `int_students__gradebook_assignments_scores`
- `int_students__gpa`
- `int_students__final_grades`

Each leaf feeds exactly one mart — `fct_grades_category`,
`fct_grades_assignments`, `fct_grades_gpa`, `fct_grades_term`. None of those 4
marts appears in a cube file or a Tableau extract, so the rebuilds produce
tables nobody queries. `0 4` matches the `gradebook_and_gpa_dashboard` batch
those marts are eventually for. All 4 share the tick, so
`~any_deps_in_progress` serializes the pass.

The 2 spine models get `0 0,4,15,17 * * *`:

- `int_students__course_enrollments`
- `int_powerschool__course_enrollments_union`

These reach much further — 184 models and 60 exposures, including the hourly
`ops_dashboard`. But reaching a model is not the same as feeding it. The hourly
dashboards project no column that comes from the spine. Every consumer that
does is a batch exposure, so the 4 ticks sit an hour ahead of each one:
`ddi_suite` at 01:00, 16:00 Friday and 18:00, `literacy_dashboard` at 05:00,
`apm_dashboard` at 06:00.

Nothing changes in BigQuery when this deploys. A properties-file change does
not alter `code_version`, which hashes the raw SQL, so the first rebuild happens
at the first cron tick.

This is the same fix as #4559, #4821, and #5035.

## Reviewer Notes

**The leaves stay tables rather than becoming views, and that is deliberate.**
Views would cost less, but each leaf's uniqueness test is the tripwire for the
Miami SIS cutover. The data-change condition only re-materializes tables, so as
views those tests would fire on code change alone. A daily tick keeps them
firing for about 8.7 GiB/day.

**The spine cron rests on a column trace, not on the DAG shape.** Worth a second
look if you disagree that `rpt_tableau__ops_dashboard` and
`rpt_tableau__attendance_dashboard` are independent of the spine. The full trace
is in #5082.

**One caveat on scope.** `int_students__category_grades` and
`int_students__gradebook_assignments_scores` inherit the PowerSchool/Focus split
from their parents rather than declaring it, so their tests guard a join
fan-out rather than the cutover seam itself. Lower value than the other 2, same
conclusion.

### General

No Dagster code changed. The automation conditions are read from dbt model
properties by `CustomDagsterDbtTranslator`, which is unchanged.

      models; 6 existing properties files changed.
- [x] Include (or update) an
      exposure changes; consumers are unchanged.
- [x] **Breaking change?** No. No column is renamed or removed, and no model
      changes materialization. Rollback is reverting the 6 properties files,
      which restores the eager condition on the next deploy.
      no external sources changed.

No schedule, sensor, or integration changed. These are per-model automation
conditions, not Dagster schedules, so the automations catalog does not cover
them.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The measurement set and the design come from
#5082 and were not re-derived here.

**Three decisions closed in #5082, not re-opened:** not views, not
`+enabled: false`, and the leaf tests do not move upstream. The leaf tests are
the SIS-cutover tripwire on the `focus_start_academic_year` floor from
`int_students__sis_cutover`; the PowerSchool and Focus branch predicates are
exact complements, so a `>=` slipping to `>`, the floor moving, or the
`not (...)` guard being dropped makes one Miami academic year emit from both
branches, both carrying `_dbt_source_project = 'kippmiami'`. The tests cannot
move upstream because no upstream model has both branches present.

**Discrepancy inside #5082, resolved toward the Proposed approach table.** The
issue's "Acceptance" section says "the 2 spine models still return an eager tree
with no cron," which contradicts its own Proposed approach table, its entire
"Spine column trace" section, and its headline arithmetic. 8.7 GiB/day (leaves)
+ 9.4 GiB/day (spine) = 18.1, which is the issue's stated expected result of
"about 18." The Acceptance line reads as a stale leftover from an earlier draft.
Both spine models therefore carry `0 0,4,15,17 * * *` here.

**Verification, all run against the worktree:**

1. `dbt parse --target staging` — clean, 813 models.
1. `CustomDagsterDbtTranslator.get_automation_condition()` called on each of the
   6 parsed manifest nodes. Each returns a tree containing
   `CronTickPassedCondition(cron_schedule=<intended>, cron_timezone='America/New_York')`.
   The timezone resolves from the kipptaf code location's `LOCAL_TIMEZONE`.
   Controls: `int_assessments__response_rollup` still returns its own
   `0 0,10,13,15,17 * * *`, and the view-materialized control
   `stg_focus__marking_periods` returns no cron at all.
1. `_get_dbt_meta` shadowing checked on the parsed nodes, not assumed. It is an
   or-short-circuit over the whole dict, so a top-level `meta` would be hidden
   by `config.meta`. All 6 resolve `meta_source=config` with no top-level
   `meta` key, so nothing is shadowed.
1. `dbt build --select int_students__category_grades+ int_students__gpa+ int_students__final_grades+ int_students__gradebook_assignments_scores+ --target dev --defer --favor-state --state <prod manifest>`
   — `PASS=29 WARN=0 ERROR=0 SKIP=0`.
1. Row counts identical to prod, diff 0 on all 4:
   `int_students__category_grades` 7,765,626; `int_students__final_grades`
   244,379; `int_students__gpa` 240,283;
   `int_students__gradebook_assignments_scores` 23,352,918.
1. `trunk check --force --no-fix` on the 6 changed files, run from inside the
   worktree — "Checked 6 files, No issues."

**One unrelated failure hit on the way, diagnosed and not worked around.** The
first build run failed with
`Name academic_year not found inside gg` in
`int_students__gradebook_assignments_scores`. Not caused by this branch: the
diff is 6 properties files and the SQL is byte-identical to `origin/main`.
Commit `9ef8227095` added `academic_year` to `int_focus__gradebook_grades` at
2026-08-28 22:42 UTC, and the local prod manifest was generated at
2026-08-28 17:59 UTC, 4h43m earlier. `--favor-state` therefore resolved that
`ref()` to a prod relation predating the column. Adding
`int_focus__gradebook_grades` to the selection so it builds fresh in dev
resolves it. This is the "main moved" shape from root CLAUDE.md. Anyone
re-running the verification locally against a manifest older than that commit
will hit the same thing.

**Do not extend this to the snapshots or `stg_powerschool__pgfinalgrades`.**
Their churn — `snapshot_powerschool__gpa_term` 69.7/day,
`snapshot_powerschool__gpa_cumulative` 73.7/day,
`stg_powerschool__pgfinalgrades` 44/day — is a symptom of these 6. Measure after
this lands before touching them.

**Confirming the result.** Re-run the builds/day query from #5082 24 hours after
deploy, grouping `INFORMATION_SCHEMA.JOBS_BY_PROJECT` by
`destination_table.dataset_id` AND `table_id`. Grouping by table name alone
aggregates across the 5 district projects and inflates the counts.

</details>

### DIFF
--- a/src/dbt/kipptaf/models/powerschool/intermediate/properties/int_powerschool__course_enrollments_union.yml
+++ b/src/dbt/kipptaf/models/powerschool/intermediate/properties/int_powerschool__course_enrollments_union.yml
@@ -21,7 +21,7 @@ models:
           # Same tick as int_students__course_enrollments, which is its only
           # child. That model's genuine consumers are all batch exposures --
           # ddi_suite (0 1, 0 16 Fri, 0 18), literacy_dashboard (0 5),
-          # apm_dashboard (0 6) -- and the 4 ticks sit an hour ahead of each.
+          # apm_dashboard (0 6) -- and each tick precedes its consumer.
           # Sharing the tick means ~any_deps_in_progress serializes the pass:
           # this model builds first, its child follows in the same run. The
           # pair costs about 9.4 GiB/day against 81.1 today.
--- a/src/dbt/kipptaf/models/powerschool/intermediate/properties/int_powerschool__course_enrollments_union.yml
+++ b/src/dbt/kipptaf/models/powerschool/intermediate/properties/int_powerschool__course_enrollments_union.yml
@@ -9,6 +9,29 @@ models:
       inlining `dbt_utils.union_relations`.
     config:
       materialized: table
+      meta:
+        dagster:
+          # 4x/day instead of eager. Rebuilds 24.3 times a day at 1.132 GiB a
+          # build, 27.5 GiB/day, to feed one consumer:
+          # int_students__course_enrollments, which is now on this same cron.
+          # Eager here rebuilds on any change to any of the 4 district
+          # base_powerschool__course_enrollments sources, so it fires far more
+          # often than anything downstream reads.
+          #
+          # Same tick as int_students__course_enrollments, which is its only
+          # child. That model's genuine consumers are all batch exposures --
+          # ddi_suite (0 1, 0 16 Fri, 0 18), literacy_dashboard (0 5),
+          # apm_dashboard (0 6) -- and the 4 ticks sit an hour ahead of each.
+          # Sharing the tick means ~any_deps_in_progress serializes the pass:
+          # this model builds first, its child follows in the same run. The
+          # pair costs about 9.4 GiB/day against 81.1 today.
+          #
+          # Not a view: the uniqueness test on (cc_dcid, _dbt_source_project)
+          # guards the cross-district union grain, and the data-change
+          # condition re-materializes tables only, so as a view the test would
+          # fire on code change alone. Refs #5082
+          automation_condition:
+            cron_schedule: 0 0,4,15,17 * * *
     data_tests:
       - dbt_utils.unique_combination_of_columns:
           arguments:
--- a/src/dbt/kipptaf/models/students/intermediate/properties/int_students__category_grades.yml
+++ b/src/dbt/kipptaf/models/students/intermediate/properties/int_students__category_grades.yml
@@ -48,7 +48,9 @@ models:
           # daily BigQuery spend. Eager buys freshness nothing reads: the only
           # downstream mart is fct_grades_category, which appears in no cube
           # file and no extract, so those rebuilds produce a table nobody
-          # queries.
+          # queries. The cube_semantic_layer exposure does list it, but no file
+          # under src/cube/model/ reads it, so that entry overstates
+          # consumption.
           #
           # Not a view, even though the intermediate layer permits it. The
           # uniqueness test here is a SIS-cutover tripwire: this model unions a
--- a/src/dbt/kipptaf/models/students/intermediate/properties/int_students__category_grades.yml
+++ b/src/dbt/kipptaf/models/students/intermediate/properties/int_students__category_grades.yml
@@ -41,6 +41,33 @@ models:
       row — Focus has no citizenship grade and no year-to-date rollup.
     config:
       materialized: table
+      meta:
+        dagster:
+          # 1x/day instead of eager. Rebuilds 85.0 times a day at 1.667 GiB a
+          # build, 141.7 GiB/day -- the largest single line in the network's
+          # daily BigQuery spend. Eager buys freshness nothing reads: the only
+          # downstream mart is fct_grades_category, which appears in no cube
+          # file and no extract, so those rebuilds produce a table nobody
+          # queries.
+          #
+          # Not a view, even though the intermediate layer permits it. The
+          # uniqueness test here is a SIS-cutover tripwire: this model unions a
+          # PowerSchool branch and a Focus branch split by a year floor from
+          # int_students__sis_cutover, and a floor that drifts makes one Miami
+          # academic year emit from both. This model inherits the split from
+          # its parents rather than declaring it, so the test guards a join
+          # fan-out rather than the seam itself -- still a live guard on the
+          # boundary the SIS migration is actively moving. As a view it would
+          # fire only on code change, because the data-change condition
+          # re-materializes tables only. A daily tick keeps it firing.
+          #
+          # 0 4 matches the gradebook_and_gpa_dashboard batch these marts feed.
+          # All 4 leaves (category_grades, gradebook_assignments_scores, gpa,
+          # final_grades) share the tick, so ~any_deps_in_progress serializes
+          # the pass -- they are direct parent-child. The group costs about
+          # 8.7 GiB/day against 228.1 today. Refs #5082
+          automation_condition:
+            cron_schedule: 0 4 * * *
     data_tests:
       - dbt_utils.unique_combination_of_columns:
           arguments:
--- a/src/dbt/kipptaf/models/students/intermediate/properties/int_students__course_enrollments.yml
+++ b/src/dbt/kipptaf/models/students/intermediate/properties/int_students__course_enrollments.yml
@@ -39,7 +39,27 @@ models:
           # 15:00 for ddi's Friday 16:00, 17:00 for ddi's 18:00. This model
           # and int_powerschool__course_enrollments_union, which shares the
           # tick, cost about 9.4 GiB/day together against 81.1 today.
-          # Refs #5082
+          #
+          # Not a view, and the gap versus a view is enormous here. This model
+          # is read about 853 times a day, and its 3 window functions over a
+          # full union all corresponding block column pruning, so a view read
+          # recomputes the whole model no matter how few columns the reader
+          # selects -- the shape #5035 measured on
+          # int_assessments__all_college_assessments, where 1 column cost
+          # 1.388 GiB against a 1.390 GiB build. At 1.209 GiB a recompute that
+          # is roughly 980 GiB/day, against 53.6 eager today and 9.4 here.
+          # Same call as #4821, which moved
+          # fct_assessment_scores_enrollment_scoped the other way for the same
+          # reason.
+          #
+          # Accepted trade-off: dim_student_section_enrollments is a view over
+          # this model, and Cube reads it live -- cube-cloud@ ran 288 jobs in
+          # 14 days against it, with no refresh_key and no pre-aggregation. So
+          # the Cube class-roster, attendance, and assessment views go from
+          # roughly eager freshness to these 4 ticks, and a mid-morning
+          # schedule change is not visible until 15:00. Weighed and accepted
+          # rather than adding school-day ticks. Revisit if roster staleness
+          # draws a complaint. Refs #5082
           automation_condition:
             cron_schedule: 0 0,4,15,17 * * *
     data_tests:
--- a/src/dbt/kipptaf/models/students/intermediate/properties/int_students__course_enrollments.yml
+++ b/src/dbt/kipptaf/models/students/intermediate/properties/int_students__course_enrollments.yml
@@ -34,7 +34,7 @@ models:
           #
           # So every genuine consumer is a batch exposure, not the hourly one:
           # ddi_suite (0 1, 0 16 Fri, 0 18), literacy_dashboard (0 5), and
-          # apm_dashboard (0 6). The 4 ticks sit an hour ahead of each --
+          # apm_dashboard (0 6). Each tick precedes its consumer --
           # 00:00 for ddi's 01:00, 04:00 for literacy's 05:00 and apm's 06:00,
           # 15:00 for ddi's Friday 16:00, 17:00 for ddi's 18:00. This model
           # and int_powerschool__course_enrollments_union, which shares the
--- a/src/dbt/kipptaf/models/students/intermediate/properties/int_students__course_enrollments.yml
+++ b/src/dbt/kipptaf/models/students/intermediate/properties/int_students__course_enrollments.yml
@@ -13,6 +13,35 @@ models:
       course-enrollment-grain source.
     config:
       materialized: table
+      meta:
+        dagster:
+          # 4x/day instead of eager. Rebuilds 44.3 times a day at 1.209 GiB a
+          # build, 53.6 GiB/day, on top of the churn it drives in its
+          # PowerSchool ancestors.
+          #
+          # Eager looks unavoidable from the DAG shape:
+          # base_powerschool__course_enrollments is a bare select * over this
+          # model, which puts it upstream of 184 models and 60 exposures,
+          # including the hourly ops_dashboard. Reachability is not dependency.
+          # There are only 2 routes to the intraday exposures. Route 1 runs
+          # through int_extracts__student_enrollments, which projects exactly 1
+          # spine-derived column -- is_sipps, an existence flag for course
+          # SEM01099G1 -- and rpt_tableau__ops_dashboard and
+          # rpt_tableau__attendance_dashboard both reference is_sipps zero
+          # times and carry no select * or dbt_utils.star, so neither depends
+          # on this model. Route 2 is rpt_tableau__assessment_dashboard refing
+          # the spine directly, and that one is a real dependency.
+          #
+          # So every genuine consumer is a batch exposure, not the hourly one:
+          # ddi_suite (0 1, 0 16 Fri, 0 18), literacy_dashboard (0 5), and
+          # apm_dashboard (0 6). The 4 ticks sit an hour ahead of each --
+          # 00:00 for ddi's 01:00, 04:00 for literacy's 05:00 and apm's 06:00,
+          # 15:00 for ddi's Friday 16:00, 17:00 for ddi's 18:00. This model
+          # and int_powerschool__course_enrollments_union, which shares the
+          # tick, cost about 9.4 GiB/day together against 81.1 today.
+          # Refs #5082
+          automation_condition:
+            cron_schedule: 0 0,4,15,17 * * *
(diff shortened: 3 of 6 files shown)


cbini:
## dbt Cloud CI warning triage

dbt Cloud passed (run `70403222933122`) with 15 test warnings. All 15 are
pre-existing. None can be caused by this change, which edits 6 properties files
and no SQL — a test warning count moves only when data or SQL moves, and
neither did.

They surface here because CI builds `state:modified+ --full-refresh`. Touching
these 6 models pulls their descendants into the selection, so warnings that
already exist in prod get evaluated on this run.

Two spot-checks rather than an assumption:

- `dim_student_section_enrollments__no_multi_stint_overlap` — 151 rows in CI.
  Running the same test SQL against the prod tables right now returns **147**.
  Same failure, already present in prod; the 4-row gap is drift between prod's
  last build and CI's full refresh against staging upstreams.
- `unique_fct_survey_responses_survey_response_key` — 46,062 rows in CI, against
  the 46,008 recorded in #4243, which is open and tracks exactly this.

The `fct_grades_term` / `dim_student_section_enrollments` FK orphans are the
long-running enrollment coverage gap tracked in #4146.

No new tracker filed, and nothing here blocks this PR.

cbini:
## Review response

Verdict per finding. Verified each claim against the code and the warehouse
before replying.

### 1. Spine feeds a live Cube view — **confirmed, and it is a real gap. Not resolved in this push.**

The finding is right, and it is right for the reason given. Verified:

- `dim_student_section_enrollments.sql:37` does `ref("int_students__course_enrollments")` directly.
- `src/cube/model/cubes/students/student_section_enrollments.yml:4` and
  `student_homeroom_section.yml:12` both set
  `sql_table: kipptaf_marts.dim_student_section_enrollments`, with no
  `refresh_key` and no pre-aggregation.
- `kipptaf_marts.dim_student_section_enrollments` is a **view**
  (`__TABLES__.type = 2`), so a Cube query reads through to the spine table at
  query time. Its freshness is exactly the spine's build cadence.
- Cube traffic is real, not theoretical. Over the last 14 days,
  `cube-cloud@teamster-332318.iam.gserviceaccount.com` ran **288 jobs / 825.6
  GiB billed** on queries referencing `dim_student_section_enrollments`, most
  recently 2026-08-30.

So the practical effect is what the finding says: a roster or section change
made at 09:00 becomes visible in the Cube-backed views at 15:00, not within the
hour. #5082's spine trace enumerated Tableau exposures only and never evaluated
Cube, so this consumer was genuinely outside the analysis.

I am not changing the cron unilaterally, because the cadence is the author's
call and #5082 settled it on an incomplete consumer list. Flagged to the author
for a decision. The obvious middle option, if intraday freshness matters, is
adding school-day ticks — `0 0,4,7,10,13,15,17 * * *` is about 16.4 GiB/day
against 81.1 today, so it keeps roughly 80% of the saving while capping
staleness near 3 hours.

### 2. "No cube file" versus the `cube_semantic_layer` exposure — **fixed in 8f3e157.**

Both halves check out. `models/exposures/cube.yml:66-69` does list all 4
`fct_grades_*` marts in `depends_on`, and no file under `src/cube/model/**`
references any of them. So the comments were accurate but incomplete, and a
future reader who found the exposure would read them as wrong. Each of the 4
leaf comments now says the exposure lists the mart and that the entry overstates
consumption.

### 3. "An hour ahead of each" is wrong for `apm_dashboard` — **fixed in 8f3e157.**

Correct. `tableau.yml:442` puts `apm_dashboard` at `0 6 * * *`, which is 2 hours
after the shared 04:00 tick, not 1. Both spine comments now read "each tick
precedes its consumer"; the enumeration that follows already gave the exact
times.

### 4, 5, 6, 7 — no action, agreed.

Finding 4 restates the scope caveat already in Reviewer Notes. Findings 5, 6,
and 7 report nothing to change.

---

Findings 2 and 3 are comment text only — no cron value, SQL, or behavior
changed. `trunk check --force --no-fix` clean on all 6 files.

cbini:
## Finding 1 resolved — spine cron stays `0 0,4,15,17 * * *`

My earlier reply left this open pending the author's call. Decided: keep the
cron as #5082 specified. The Cube freshness change is accepted, not overlooked.

A view was considered as the alternative and rejected on measurement, not
preference. `int_students__course_enrollments` is read about **853 times a day**
— Dagster agent 411.5, dbt 341, 4 named analysts ~62, Codespaces 27.8, Cube 6.3,
Tableau 1.9 (14-day average). Its 3 window functions over a
`full union all corresponding` block column pruning, so a view read recomputes
the whole model however few columns the reader selects. That is the shape #5035
measured on `int_assessments__all_college_assessments`, where selecting 1 column
cost 1.388 GiB against a 1.390 GiB full build.

| Option | Rebuild | Read | Total |
| --- | --- | --- | --- |
| View | 0 | ~809 recomputes × 1.209 GiB | **~980 GiB/day** |
| Table, eager (today) | 53.6 GiB/day | cheap scans | 53.6 |
| Table, this cron | ~4.8 GiB/day | cheap scans | ~9.4 with the union |

The ~980 is inferred from the read count times the measured per-build cost, not
directly measured. Even if only a quarter of those reads touched the full model
it lands near 250 GiB/day, still worse than today. This is #4821 in reverse:
that PR moved `fct_assessment_scores_enrollment_scoped` from a view to a 5x/day
table for exactly this reason.

Both points are now written into the model's comment in 253dc01, so neither gets
re-derived: why it is not a view, and the accepted Cube trade-off with the
condition to revisit it. School-day ticks
(`0 0,4,7,10,13,15,17`, about 16.4 GiB/day) remain the cheap upgrade if roster
staleness draws a complaint.

Comment text only in that commit — no cron value, SQL, or behavior changed.

PR #5088, opened 2026-08-31
feat: widen every unbounded Focus numeric and delete the opt-in

### LINKED ISSUE #5080: feat: widen every unbounded Focus numeric and delete the per-table opt-in

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will widen every unbounded `numeric` column in
the Focus dlt source, and delete the opt-in mechanism that used to gate
widening per table.

An unbounded Postgres `numeric` column reflects with no precision. dlt renders
that as `decimal128(38, 9)`, and pyarrow rejects any value that needs more than
9 decimal places. When that happens, the whole `kippmiami__dlt__focus` asset
fails, because all 20+ Focus tables load in one Dagster step. This has taken
down the Focus sync twice: `student_gpa_calculated.weighted_gpa` (#5021) and
`gradebook_grades.points` (#5074, 2 days of downtime).

A prior PR (#5085, merged) widened the fix to be safe for every table by
casting 94 columns back to `numeric` in 40 `stg_focus__*` dbt staging models,
so their contracts still declare `numeric` after BigQuery retypes the columns.
This PR builds on that: `_widening_type_adapter` now runs for every table,
and the `widen_numeric` parameter, the `widen_numeric_tables` frozenset, and
the `widen_unbounded_numeric: true` opt-in key in `config/focus.yaml` are all
deleted.

Closes #5080

## Reviewer Notes

**Merging this breaks the Focus sync until someone launches the reload.** The
moment the `kippmiami` code location redeploys, a load writes BIGNUMERIC into
columns BigQuery still holds as NUMERIC, and `replace` fails with:

```text
Provided Schema does not match Table ... Field points has changed type from
NUMERIC to BIGNUMERIC
```

The remedy: launch one manual run over all `kippmiami/dlt/focus/*` assets,
from the Dagster UI, with this run config.

```yaml
ops:
  kippmiami__dlt__focus:
    config:
      refresh: drop_resources
```

This is proven mechanics, not a guess. A run with this config over
`gradebook_grades` alone already cleared the same schema-mismatch error, in 18
seconds, loading 6,855 rows.

The reload has to cover all 79 tables, not only the 44 tables this PR measured
as having an unbounded numeric column today. That 44-table list came from a
schema snapshot dated 2026-08-28 — any column added since would be missed.
Reloading every table is self-correcting and cheap: the whole dataset is
799,554 rows and 173 MB, so a full reload runs in minutes.

Any sensor tick that lands between the deploy and the reload fails harmlessly.
dlt only commits resource state for resources that reached the load package,
so a failed load keeps the old baseline and the same table gets re-selected on
the next tick. The cost is 1 or 2 red runs and their alerts — no sensor or
schedule needs to be paused.

This reload has to be launched by a person from the Dagster UI. It is a
destructive, shared-resource mutation (`drop_resources` drops and reloads
live tables), so it is not something to automate here.

### General

### Dagster

      code location — not run; the test suite exercised was
      `tests/libraries/test_dlt_focus_type_adapter.py` (14/14 passing), which
      covers the actual change
      with the correct asset key format and IO manager — n/a, no new
      integration

### dbt

Skipped — no dbt changes in this PR.

### Docs

      `CLAUDE.md` by running `/claude-md-management:revise-claude-md` — no new
      integration; `src/teamster/libraries/dlt/CLAUDE.md` was corrected by
      hand (table count) instead, since it's an existing-doc fix, not a new
      integration writeup

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. This is PR 2 of 2 for issue #5080. PR 1
(#5085, merged) added `cast(<col> as numeric)` to 94 columns across 40
`stg_focus__*` staging models so their enforced contracts survive the retype;
that ordering is why PR 1 had to land first.

This branch is 3 commits on top of `main`:

- `9ebfbc2219` feat(dagster): widen every unbounded Focus numeric —
  `_widening_type_adapter` now applies to every table; deletes the
  `widen_numeric` parameter, the `widen_numeric_tables` frozenset, and the
  branch between two adapters. Retargets the test suite accordingly.
- `2b3d3cd55e` chore(dagster): drop the Focus `widen_unbounded_numeric` config
  key — removes both dead keys from `config/focus.yaml`. Also corrects the
  Focus table count in `src/teamster/libraries/dlt/CLAUDE.md` (76 to 79, 75 to
  78).
- `b02ea4e78f` docs(dagster): correct the Focus retype column count and tense.

Net diff: 5 files, 21 insertions, 55 deletions.

Verification: `pytest tests/libraries/test_dlt_focus_type_adapter.py` is 14
passed, 0 failed, including a new
`test_every_table_gets_the_widening_adapter` that drives two distinct table
names through the real code path and asserts both reach `table_rows` with
`_widening_type_adapter`. `grep -rn widen_numeric src/teamster/` returns
nothing. `trunk check --force` is clean.

Measured scope: 191 columns across 44 tables reflect as unbounded numeric
today and will retype to BIGNUMERIC on next load. The dataset is 799,554 rows
and 173 MB total, so the drop-and-reload described in Reviewer Notes runs in
minutes, not hours.

Task 7 (tracked separately) is the actual cutover: launch the
`drop_resources` run described above right after this PR merges and the
`kippmiami` code location redeploys. Do not merge this PR without someone
ready to run Task 7 within minutes — that is the whole point of the warning
above.

</details>

### DIFF
--- a/src/dbt/focus/models/staging/properties/stg_focus__student_gpa_calculated.yml
+++ b/src/dbt/focus/models/staging/properties/stg_focus__student_gpa_calculated.yml
@@ -22,11 +22,12 @@ models:
       downstream.
 
 
-      The five decimal columns arrive from dlt as BIGNUMERIC, because
-      `weighted_gpa` is an unbounded Postgres `numeric` carrying up to 20
-      decimal places and needs the `widen_unbounded_numeric` dlt adapter to load
-      at all. They are cast to NUMERIC here, which keeps this package's decimal
-      vocabulary consistent and costs at most 5e-10 of rounding.
+      The five decimal columns arrive from dlt as BIGNUMERIC because the
+      source-wide `widen_unbounded_numeric_adapter` widens every unbounded
+      Postgres `numeric` column — `weighted_gpa`'s up to 20 decimal places is
+      the column that originally forced the widening, since without it the load
+      fails outright. They are cast to NUMERIC here, which keeps this package's
+      decimal vocabulary consistent and costs at most 5e-10 of rounding.
 
 
       The 21 `custom_*_gpa`, `custom_*_rank` and `ncaa_gpa` columns in the
--- a/src/dbt/focus/models/staging/stg_focus__test_history_score_ranges.sql
+++ b/src/dbt/focus/models/staging/stg_focus__test_history_score_ranges.sql
@@ -7,8 +7,9 @@ select
     level,
     gradelevel,
     form,
-    `min`,
-    `max`,
     created_at,
     updated_at,
+
+    cast(`min` as numeric) as `min`,
+    cast(`max` as numeric) as `max`,
 from {{ source("focus", "test_history_score_ranges") }}
--- a/src/teamster/code_locations/kippmiami/CLAUDE.md
+++ b/src/teamster/code_locations/kippmiami/CLAUDE.md
@@ -70,7 +70,7 @@ Constraints to preserve when touching any of these:
   The `0 4 * * *` schedule is the unconditional overnight backstop for only the
   count-only tables (`cursor_column: null` in `config/focus.yaml` —
   `co_teachers` as of writing), which is the set the sensor's count-only probe
-  can't fully see an in-place edit on. The other 75 `updated_at`-tracked tables
+  can't fully see an in-place edit on. The other 78 `updated_at`-tracked tables
   have NO unconditional reload anymore — if one of them is ever suspected of
   drifting (a stuck signature, a bad cursor), fix it with a manual launch of the
   Focus asset job for that table, not by waiting for a nightly pass that no
@@ -79,7 +79,7 @@ Constraints to preserve when touching any of these:
   import.
 - **First diagnostic when midday Focus data looks stale: check whether the
   intraday sensor is running.** It ships with `defaultStatus` STOPPED and must
-  be enabled by hand after a one-off manual load of all 76 tables has seeded
+  be enabled by hand after a one-off manual load of all 79 tables has seeded
   their baselines — the 04:00 tier cannot seed them, since it now targets only
   the count-only tables (see `libraries/dlt/focus/CLAUDE.md`). A stopped sensor
   now silently freezes every `updated_at`-tracked table — the 04:00 tier no
--- a/src/teamster/code_locations/kippmiami/dlt/focus/assets.py
+++ b/src/teamster/code_locations/kippmiami/dlt/focus/assets.py
@@ -30,22 +30,10 @@ import — both `assets.py` and `sensors.py` load unconditionally via
 `dlt/focus/__init__.py`.
 """
 
-widen_numeric_tables = frozenset(
-    a["table_name"] for a in config_assets if a.get("widen_unbounded_numeric")
-)
-"""Tables whose unbounded Postgres `numeric` columns need an explicit scale.
-
-`.get()`, not `[...]`: this key is absent on all but the tables that need it.
-Opt-in rather than source-wide because widening retypes the column to BigQuery
-BIGNUMERIC, and 45 already-loaded Focus tables carry 200 NUMERIC columns that
-`replace` cannot retype in place.
-"""
-
 assets = [
     build_focus_dlt_assets(
         sql_database_credentials=sql_database_credentials,
         code_location=CODE_LOCATION,
         tables=tables,
-        widen_numeric_tables=widen_numeric_tables,
     )
 ]
--- a/src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml
+++ b/src/teamster/code_locations/kippmiami/dlt/focus/config/focus.yaml
@@ -81,9 +81,6 @@ assets:
     cursor_column: updated_at
   - table_name: student_gpa_calculated
     cursor_column: updated_at
-    # weighted_gpa is an unbounded Postgres numeric holding more than 9 decimal
-    # places, which the default decimal128(38, 9) cannot represent.
-    widen_unbounded_numeric: true
   - table_name: report_card_grades
     cursor_column: updated_at
   - table_name: report_card_grade_scales
@@ -110,9 +107,6 @@ assets:
     cursor_column: updated_at
   - table_name: gradebook_grades
     cursor_column: updated_at
-    # points is an unbounded Postgres numeric holding more than 9 decimal
-    # places, which the default decimal128(38, 9) cannot represent.
-    widen_unbounded_numeric: true
   - table_name: gradebook_assignments
     cursor_column: updated_at
   - table_name: gradebook_assignment_types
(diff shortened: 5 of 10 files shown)


cbini:
## Review verdicts

Verified each finding against the code before acting. Fixes pushed as `e761cbf`.

### Finding 1 — stale table counts in sibling docs (Low) — ACCEPTED, fixed

Correct, and the sweep needed to go wider than the two files named. Eight lines across three files carried the stale count, including a code comment in the same directory as the edited `focus.yaml`:

| file | lines |
| --- | --- |
| `src/teamster/libraries/dlt/focus/CLAUDE.md` | 4 |
| `src/teamster/code_locations/kippmiami/CLAUDE.md` | 2 |
| `src/teamster/code_locations/kippmiami/dlt/focus/schedules.py` | 2 (comments only) |

All eight now read 79 total and 78 `updated_at`-tracked, matching `focus.yaml`'s 79 entries with `co_teachers` the only count-only one.

Worth recording for the next sweep: a naive `\b(75|76)\b` grep over these directories returns 9 hits that must NOT be changed — the four `Numeric(76, 38)` BIGNUMERIC precision references in `libraries/dlt/focus/assets.py`, and five "75 min" durations across `kippmiami/CLAUDE.md`, `finalsite/schedules.py`, and `extracts/schedules.py`. Every surviving hit was re-classified after the fix to confirm none was touched.

### Finding 2 — stale opt-in framing in the dbt staging doc (Low) — ACCEPTED, fixed

Correct, with one more error than reported: `widen_unbounded_numeric` was the **config key**, which this PR deletes. The function is `widen_unbounded_numeric_adapter`. The description now names the adapter correctly and describes it as source-wide rather than something this column needs.

### Finding 3 — test redundancy in `test_every_table_gets_the_widening_adapter` (nitpick) — DECLINED

Agreed on the analysis and keeping the test as-is. The two iterations cannot diverge precisely because `_build_focus_resource` no longer branches on table name — that absence is what the test pins. A future change reintroducing per-table routing would make them diverge, which is the regression this guards. No change, consistent with the finding not requesting one.

### Separately — a Critical this PR's own review chain caught

Worth flagging to human reviewers, because CI cannot catch it.

`src/dbt/focus/models/staging/stg_focus__test_history_score_ranges.sql` projected `` `min` `` and `` `max` `` uncast. Both declare `data_type: numeric` under the enforced `focus.staging` contract, and both are unbounded in the source, so the reload would have retyped them to BIGNUMERIC and broken that contract **in production, after the destructive step**, where no CI run could see it — types do not change until the reload runs.

Root cause: the generator that added the other 94 casts in #5085 matched projections with `^\s{4}\w+,$`, and `\w+` cannot match a backtick-quoted identifier. These are the only backtick-quoted numeric columns in the Focus staging layer, `min` and `max` being reserved words, so the same blind spot also misclassified them as unprojected during measurement. The correct reconciliation of the 191 is 94 cast + 2 on `gradebook_grades` + these 2 + 93 unprojected.

Both are now cast, with backtick quoting preserved on the output aliases to match `quote: true`. A full re-sweep of every currently-NUMERIC column against every Focus staging model found no other gap; the remaining uncast candidates all carry declared precision in Postgres and will stay NUMERIC.

### CI note

On the previous commit, 4 of 5 `dagster-cloud-deploy / deploy` jobs passed. The failure was **kippcamden** with `DagsterUserCodeUnreachableError` — gRPC connection refused to the user-code server, a branch-deployment warm-up flake rather than a code defect. kippcamden is redeployed only because this PR touches the shared `libraries/dlt/` package. kippmiami, the location that matters here, deployed successfully. This push re-triggers the workflow.

Refs #5080

PR #5098, opened 2026-09-01
fix(dagster): serialize role assignment inserts per org unit

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop the "Reset Student PW" role assignment
job from reporting a false failure when it grants many staff the same role on the
same student org unit.

Google's Directory API handles one write at a time per org unit. The job sends
role assignments 10 at a time in a single batch. When two of those 10 target the
same org unit, they race, and one comes back with `409 Conflicting requests.
Please try again`. The job does not retry that response, so it records the
assignment as failed and the `zero_api_errors` check fires an alert.

This is what happened in
[run d5145d79](https://kipptaf.dagster.cloud/prod/runs/d5145d79-1b25-4fd1-9971-66862bf10ae7)
on 8/31. The run itself succeeded, but 1 of 135 role assignments failed this way,
and one staff member did not get their password reset role.

The fix changes how batches are built. Each batch now takes at most one
assignment per org unit. Batches run one after another, so two writes to the same
org unit are never in flight at the same time.

## Reviewer Notes

- The new batching makes batches smaller when the work targets few org units.
  The batch count equals the largest single org unit's row count, not the row
  total divided by the batch size. The 8/31 payload of 135 rows spanned 6 org
  units, the largest holding 56, so it goes from 14 batches of 10 to 56 batches
  that taper from 6 down to 1. Each batch is followed by a 1 second pause, so the
  job takes roughly 42 seconds longer. The job has no runtime pressure.
- This fix prevents the conflict rather than retrying it. A 409 from some other
  source would still fail the check. See the fold-out for why we chose prevention.

### General

### Dagster

      code location — `test_definitions_kipptaf` passes. 4 of 6 pass locally;
      `kippmiami` and `all` fail on a missing dlt config value that also fails on
      `main`, so it is pre-existing and unrelated. Details in the fold-out.
      with the correct asset key format and IO manager _(no new integration)_

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Found via `superpowers:systematic-debugging` on
Dagster prod run `d5145d79-1b25-4fd1-9971-66862bf10ae7`.

**Evidence.** The run status was `SUCCESS`; the alert came from the
`zero_api_errors` asset check on
`kipptaf/google/directory/role_assignments_create`, which failed WARN with one
error:

```text
HttpError 409 when requesting .../roleassignments returned
"Conflicting requests. Please try again".
Details: [{'message': 'Conflicting requests. Please try again',
           'domain': 'global', 'reason': 'conflict'}]
```

The 4 prior check executions all recorded `errors: []`.

**Why it surfaced now.** PR #5095 (merge `494847107e`) added the Paterson org
unit mapping to `rpt_google_directory__admin_import`, which took the extract from
roughly 0-3 pending rows a day to 135. Row counts by org unit at the time:

| org_unit_path                                    | rows |
| ------------------------------------------------ | ---- |
| /Students/KIPP Paterson/Paterson Prep Elementary | 56   |
| /Students/KIPP Paterson/Paterson Prep Middle     | 33   |
| /Students/Miami/Legacy Elementary                | 20   |
| /Students/KCNA/KSA                               | 17   |
| /Students/Miami/Legacy Middle                    | 6    |
| /Students/KCNA/KHA                               | 3    |

With `chunk(size=10)` over that input, essentially every batch held multiple
inserts against one org unit.

**Root cause.** `_TRANSIENT_HTTP_CODES` is `{429, 500, 502, 503, 504}`.
`_execute_batch_with_retry` retries a sub-request only when its status is in that
set, so a 409 lands in `failures` on the first attempt and reaches the check.
The reason code was `conflict`, not `duplicate` — Google's concurrency
serialization, not "entity already exists". The extract already anti-joins
`stg_google_directory__role_assignments`, so a true duplicate is unlikely anyway.

**Options considered.** Three were put to the author: add 409 to
`_TRANSIENT_HTTP_CODES`; retry 409 only when `error_details` carries
`reason == 'conflict'`; or serialize per org unit. The author chose
serialization, which removes the contention instead of absorbing it.

**Known limits, stated plainly.** Serialization only helps if the org unit is the
contended resource. That is an inference from the error shape and the row
distribution, not something confirmed against Google's implementation. If 409s
persist, the retry-based options above are still on the table.

**Not changed.** `_TRANSIENT_HTTP_CODES` still omits 409 for the other batch
methods, and `batch_update_users` still carries a hand-rolled 409 retry
(`_retry_update_user`) to compensate. That inconsistency is real but out of scope
here.

**Verification.** 4 new tests in
`tests/resources/test_resource_google_directory.py` cover no-repeated-org-unit
within a batch, every item yielded once, customer-scoped items still packing to
`size`, and `batch_insert_role_assignments` splitting 3 same-org-unit items into 3
batches. Selected run: 12 passed, 46 deselected; 58 collected clean. The live-API
tests in that file were deliberately not run — they mutate the real Google
directory. `trunk check --force` on both changed files: no issues.

**Definitions validation.** `uv run pytest tests/test_dagster_definitions.py` in
the worktree: 4 passed, 2 failed. `test_definitions_kipptaf` — the only location
that imports the changed module — passes. `test_definitions_kippmiami` and
`test_definitions_all` fail on a dlt `ConfigFieldMissingException`, and
`test_definitions_kippmiami` fails identically on `main` in the same Codespace, so
it is a local config gap and not this branch. A fresh worktree also has no
compiled dbt manifest, which fails every location until you copy or regenerate
`src/dbt/*/target/manifest.json` — worth knowing before chasing that error.

</details>

### DIFF
--- a/src/teamster/libraries/google/CLAUDE.md
+++ b/src/teamster/libraries/google/CLAUDE.md
@@ -38,11 +38,21 @@ failed emails to skip group membership for uncreated users (via
 includes the password hash) out of logs and asset-check metadata.
 
 **409 conflict handling**: 409 is deliberately excluded from
-`_TRANSIENT_HTTP_CODES` because its meaning is method-specific: for
-`batch_insert_role_assignments` it means "entity already exists" (not
-retryable), but for `batch_update_users` it means "conflicting request, please
-try again" (retryable). User update 409s are retried individually via
-`_retry_update_user()` after a 1-second cooldown.
+`_TRANSIENT_HTTP_CODES`, so no batch method retries one. Read the `reason` code
+before assuming what a 409 means — `duplicate` ("entity already exists") is
+permanent, `conflict` ("conflicting requests, please try again") is Google
+serializing two writes against one resource and is transient. User update 409s
+are retried individually via `_retry_update_user()` after a 1-second cooldown.
+
+**Role assignment batches are serialized per org unit**: the Directory API
+serializes writes against a single org unit, so two `roleAssignments.insert`
+sub-requests scoped to the same org unit inside one batch race and one returns
+409 `conflict` — which nothing retries, so it fails the `zero_api_errors` check.
+`_batch_by_distinct_org_unit` deals at most one item per org unit into each
+batch, and batches run sequentially, so two writes to one org unit are never in
+flight together. The batch count therefore equals the largest single org unit's
+row count, NOT `len(rows) / size` — a payload of 135 rows across 6 org units
+(largest 56) yields 56 batches, not 14.
 
 **Batch callback signature**: Google's batch executor calls
 `callback(id, response, exception)` with exactly one of `response`/`exception`
--- a/src/teamster/libraries/google/directory/resources.py
+++ b/src/teamster/libraries/google/directory/resources.py
@@ -1,5 +1,6 @@
 import time
-from collections.abc import Callable
+from collections import defaultdict, deque
+from collections.abc import Callable, Iterator
 
 from dagster import ConfigurableResource, DagsterLogManager, InitResourceContext
 from dagster._utils.backoff import backoff, exponential_delay_generator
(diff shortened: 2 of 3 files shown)


cbini:
Verdicts on all 7 findings, verified against the code before responding. Fixed in `857ba5505a`.

| # | Finding | Verdict |
| --- | --- | --- |
| 1 | Reuse `chunk()` + `itertools.zip_longest` | **Accepted, applied** |
| 2 | Batch-count math in the PR body is wrong | **Accepted, PR body corrected** |
| 3 | `libraries/google/CLAUDE.md` 409 section is out of date | **Accepted, applied** |
| 4 | `list(by_org_unit)` re-snapshot is O(units² / size) | **Moot** — finding 1 removes it |
| 5 | Test coverage — no gap | No action |
| 6 | Dagster surface — no issues | No action |
| 7 | Typing conventions — followed | No action |

**1 — applied.** Ran both implementations against the 8/31 payload (56/33/20/17/6/3) and the customer-scoped case: output is identical, 56 batches either way. They diverge only when the distinct org units outnumber `size`, where the current code starves later org units and `zip_longest` interleaves them fairly — so the swap is a small improvement, not just fewer lines. The `deque` import is gone.

**2 — correct, and my error.** I divided 135 rows by 6 org units to get "~23 batches". The batch count is actually set by the largest queue, which stays active every round until drained, so it is 56 batches with sizes tapering 6 → 1, and about 42 added seconds rather than 20. Corrected in the PR body, and the docstring now states the rule so the next reader does not repeat the mistake.

**3 — correct.** [`src/teamster/libraries/google/CLAUDE.md:40-44`](https://github.com/TEAMSchools/teamster/blob/main/src/teamster/libraries/google/CLAUDE.md#L40-L44) said a `batch_insert_role_assignments` 409 means "entity already exists" and is not retryable. The production 409 in run `d5145d79` carried `reason: conflict`, not `duplicate`. The section now distinguishes the two reason codes and documents the serialization strategy, including the batch-count rule from finding 2.

**4 — no separate change.** The rewrite in finding 1 has no per-batch re-snapshot of remaining keys, so the quadratic term is gone rather than merely acceptable.

One thing I did **not** change, to keep this PR to one idea: `_TRANSIENT_HTTP_CODES` still omits 409 for every other batch method, and `batch_update_users` still carries its own `_retry_update_user` to compensate. That inconsistency is real, and it is a separate change.

PR #5099, opened 2026-09-01
fix(dagster): partition renlearn drops by the archive's own school year

### LINKED ISSUE #4388: fix(renlearn): stg_renlearn__star fiscal-year partition overlap duplicates assessment_id rows

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make the renlearn sensor read the school
year out of the file Renaissance drops, and use that to pick the partition the
data lands in.

The sensor used to stamp every drop with the current fiscal year, read off the
clock when the code loaded. Renaissance overwrites one fixed-name archive
instead of publishing a file per year. So a July drop still holding last year's
results landed in the fiscal-year partition that had just rolled over.
`stg_renlearn__star` builds `academic_year` from that partition label, which
means 2,296 SY2025-26 STAR results reported as SY2026-27 results for part of
this summer. The wrong labels cleared on their own once real SY2026-27 data
arrived and overwrote them, but the same thing happens every July 1.

The archive is under 6 KB, and the sensor only reads it when a matched file's
timestamp advances, so reading the school year costs almost nothing.

This pull request also fixes a way to lose data that has nothing to do with the
calendar. Both SFTP asset builders wrote one all-null row whenever they found
nothing to read. For a missing file that is data loss: the asset reported
success and overwrote whatever the partition already held. Renaissance's drops
come and go within hours, so re-running any renlearn partition at the wrong
moment could erase a full school year. A missing file and an unreadable archive
now fail loudly. An archive member that is genuinely empty writes 0 rows
instead of a null row.

The sweep covers all four places that idiom appeared, which includes
`outlier_detection` in `kipptaf/performance_management/assets.py`. That one is
BigQuery-backed rather than SFTP, and the change there is cosmetic — it passes
an empty list to a check call whose result is identical either way.

Two smaller pieces:

- A `unique` test on `stg_renlearn__star.assessment_id`, which issue #4388
  asked for, with `severity: error` so it fails CI rather than warning. It
  passes on today's data.
- A corrected comment in `fct_assessment_scores_enrollment_scoped`. The old
  `TODO(#4388)` said its STAR dedupe could go away once staging was unique.
  That is wrong, and the comment now says why.

Closes #4388

## Scope added after this pull request opened

PR #5102 merged into this branch, so the diff is wider than the summary above
describes. Two things came with it:

- **Newark renlearn STAR is retired** (#5101). Renaissance ships a 0-byte
  `SM.csv` and `SR.csv` for the NJ region, so `kippnewark/renlearn/star` had
  never carried real data -- 2 all-null rows per fiscal year since FY2024. The
  Dagster integration, its wiring, the `kippnj_renlearn` union leg, the renlearn
  package dependency, and the two Newark tests are all gone. Net 172 lines
  deleted.
- **The Newark Accelerated Reader archive stays declared as a source.** That
  schema still holds 34,177 AR rows (18,041 distinct, SY2023-24 and SY2024-25,
  static since 2025-08-01) including identified student and staff fields.
  Deleting the source declaration would have left them in BigQuery with no
  lineage pointing at them, so `sources-kippnj.yml` is restored with the
  `accelerated_reader` table only. Nothing reads it by design -- it keeps the
  archive discoverable without reviving an ingestion that lapsed a year ago.
  Column-level inventory and the retention hand-off are on #5101.

Do **not** delete the `op-renlearn-sftp-kippnj` secret as part of this:
`kippmiami`'s `dagster-cloud.yaml` reads it too, since production uses one
shared renlearn SFTP account named for the NJ region.

## Reviewer Notes

- **The dedupe in `fct_assessment_scores_enrollment_scoped` is permanent.** STAR
  gives each sitting its own `assessment_id`, and students really do retest the
  same subject on the same day. 144 rows collapse at the fact grain today, and
  the fact carries no attempt dimension. Removing that dedupe would break the
  fact's key.
- **Raising on a missing file changes how a failed drop looks.** A renlearn run
  that finds no file now fails instead of quietly writing an empty partition.
  That is the point, but it will show up as run failures where it used to be
  silent. `build_sftp_file_asset` has always behaved this way for a missing
  file, so this makes the two builders agree.
- **`kippnewark/renlearn/star` has never had data.** Renaissance ships a 0-byte
  `SM.csv` and `SR.csv` for the NJ region, which is where its 2 all-null rows
  per fiscal year came from. Those rows stop being created, but existing
  partitions keep theirs until they re-materialize. PR #5102 retires that
  integration.
- **The empty-record change touches `build_sftp_file_asset` too**, which many
  integrations use. Any drop of a 0-byte file now yields 0 rows instead of one
  null row.

### General

      code location

- [ ] Include a `[model_name].yml` properties file for all new models — see
- [ ] Include (or update) an
      summary above.
      See

No new models, no renamed or removed columns, no new sources. The only dbt
changes are one test and one comment.

The catalog needs no update. The sensor keeps its name, its interval, and its
asset selection; only the partition key it picks has changed.

## CI checks

All three were green as of commit `4165385`. The review-fix commit `f7bcb03`
re-runs them.

## Troubleshooting

<details>
<summary>For Claude</summary>

**Verification.** `tests/libraries/test_renlearn_sensor_school_year.py` (6
tests) and `tests/libraries/test_sftp_archive_asset_empty.py` (3 tests) pass.
The empty-member test was mutation-checked: reverting the empty-list assignment
back to a one-element dict makes it fail, so it asserts on the records
themselves rather than the `records` metadata, which read 0 in both the old and
new code. The fallback-branch sensor test pins the configured key to
`2024-07-01` so the wall clock cannot produce the same answer.

Live tests against the renlearn SFTP: 5 of 6 pass in
`tests/assets/test_assets_renlearn_sftp.py` and
`tests/sensors/sftp/test_sensors_sftp_renlearn.py`.
`test_renlearn_fast_star_kippmiami` fails identically on `main` — it pins
`partition_key="2025-07-01|Participation"`, and `fast_star`'s subject dimension
is `["SM", "SR", "SEL", "SEL_Domains"]`, so the key does not exist. Stale test,
not this branch.

`tests/test_dagster_definitions.py`: 4 of 6 pass, `kippnewark` among them.
`test_definitions_kippmiami` and `test_definitions_all` fail identically on
`main` with `ConfigFieldMissingException` for the `FOCUS_DB`
`ConnectionStringCredentials` — a missing local credential, unrelated to
renlearn. Note that in a fresh worktree every code location fails validation,
because `src/dbt/*/target/` is gitignored and therefore absent; symlinking the
main checkout's target directories in is what got 4 of 6 to pass.

**Partition key derivation.** A `SchoolYear` of `"2026-2027"` maps to the
`2026-07-01` key, which the IO manager writes as
`_dagster_partition_fiscal_year=2027`. Verified against existing data: FY2026
holds `school_year` of `"2025-2026"`, and its key is `2025-07-01`. The peek
scans archive members and takes the first populated `SchoolYear`, matched
case-insensitively, so it works for the STAR CSVs, for the Accelerated Reader
CSV, and for archives whose dashboard-standards members carry no `SchoolYear`
column. One archive means one school year, so any member that has the column
answers for all of them.

The sensor advances its cursor before the partition-validity check, so a school
year that falls outside the partitions definition is not re-downloaded every
tick. Both of those branches now have tests.

**Why the fact dedupe stays.** 8,108 source rows reduce to 7,964 distinct
`(_dbt_source_project, student_number, administration_period, module_code,
test_date)` — 144 genuine same-day retests. Verified 2026-09-01 against
`kipptaf_renlearn.stg_renlearn__star`.

**Why the `unique` test passes today.** 0 duplicate `assessment_id` values, and
`n_rows` equals `n_ids` within every fiscal-year partition. The 2,296-row FY2026
/ FY2027 overlap that issue #4388 and its comments describe is gone: FY2027 now
holds 716 real SY2026-27 rows with completed dates from 2026-08-25 onward.
dbt's `unique` test ignores nulls, so the 8 all-null `kippnewark` rows do not
trip it. The open question in the last issue comment about deleting FY2027 GCS
blobs is therefore moot — FY2027 is legitimately populated, and deleting it
would destroy real data.

**Two things found and deliberately left out of scope.**

1. The only populated member of the NJ archive is the Accelerated Reader CSV,
   and no Dagster asset reads it.
   `src/dbt/renlearn/models/staging/stg_renlearn__accelerated_reader.sql`
   exists and its external source points at
   `.../renlearn/accelerated_reader/*`, a path nothing writes to. Orphaned
   model, live vendor data going unread.
2. The assets ignore their fiscal-year dimension when locating the file:
   `remote_dir_regex` is a literal dot and the file regexes carry no named
   groups, so `compose_regex` is a no-op and every partition materializes
   whatever single file is on the SFTP right now. The partition key is a
   write-time label only. With the sensor now deriving that label from the data,
   a sensor-driven run is correct, but a manual backfill of an old partition
   still writes today's file into it. Unchanged by this branch, and worth its
   own issue.

**The 8 existing all-null `kippnewark` rows** persist until each partition
re-materializes. PR #5102 removes the source from the union, which stops them
being read.

**Review response.** Per-finding verdicts are posted as a PR comment. Findings
1, 2, 3, and 5 were fixed; 4, 6, and 7 were declined with reasoning.

</details>


cbini:
Per-finding verdicts. 4 fixed in `f7bcb03`, 3 declined with reasons.

**1. Missing `severity: error` — fixed.** All three claims check out:
`src/dbt/kipptaf/dbt_project.yml:23` is `+severity: warn`, `src/dbt/CLAUDE.md:796-801`
says staging tests MUST set `config: severity: error`, and the per-column shape
appears at `stg_kippadb__kipp_aid.yml:7-9` and `stg_overgrad__universities.yml:11-13`.
The test now sets `severity: error`. This one mattered — as written the test could
not have failed CI on the regression it exists to catch.

**2. Column not sorted to top — fixed.** `src/dbt/CLAUDE.md:1247` confirmed.
`assessment_id` now sits first in `columns:`.

**3. Untested sensor branches — fixed.** Added
`test_sensor_falls_back_when_the_archive_names_no_school_year` and
`test_sensor_skips_a_school_year_outside_the_partitions_definition`. Both drive a
real archive rather than mocking the helper. The fallback test pins the configured
key to `2024-07-01` so the wall clock cannot produce the same answer, and the skip
test also asserts the cursor advanced.

**4. No exception handling in `_school_year_start_date` — declined, deliberate.**
Failing the tick is the intended shape. The cursor is only persisted via the final
`SensorResult`, so a failed tick loses no state and retries at the next interval.
More importantly, catching and continuing would fall back to the wall-clock key,
which reintroduces the exact mislabeling this pull request exists to fix. A silent
fallback is worse than a loud retry here.

**5. Undocumented file in the diff — fixed.** The PR body now mentions the
`performance_management` change and notes it is cosmetic: the empty list goes to a
check call whose result is identical either way.

**6. `build_sftp_folder_asset` keeps quiet empty success — declined, follow-up
welcome.** Verified at `src/teamster/libraries/sftp/assets.py:442-453`. Worth noting
it yields `Output(value=([], avro_schema))`, so the all-null-row bug never existed
there; only the quiet-success-on-no-match shape is shared. It backs 7 assets across
tableau, nsc, collegeboard sat/psat, pearson in 2 locations, and fldoe. `fast` in
`kippmiami/fldoe/assets.py` partitions on school year and term parsed from the
vendor's own directory names, so partitions for terms Florida has not published yet
legitimately have no file — raising there would fail them by design. Sweeping this
safely needs each integration's drop pattern studied first, which is more than this
branch should carry.

**7. Global-uniqueness assumption — no action, agreed.** PR #5102 drops the
`kippnj_renlearn` leg, so `kippmiami_renlearn` becomes the only tenant. The
re-check query is recorded in the PR body for whenever a third tenant appears.

On the "not run" note: `dagster definitions validate`, `pytest`, and `dbt parse`
were all run locally, and the results are in the PR body. The one failing test,
`test_renlearn_fast_star_kippmiami`, fails identically on `main` — it pins a
partition key whose subject does not exist in `fast_star`'s dimension.

PR #5105, opened 2026-09-01
fix: restore FY2027 i-Ready ingestion in all regions

### LINKED ISSUE #4949: fix: i-Ready 2026-2027 export renames have broken FY27 ingestion in all regions
## What's happening

Curriculum Associates renamed the i-Ready SFTP exports for the 2026-2027 school year. None of our asset regexes match the new filenames, so FY27 i-Ready ingestion is partially or fully stopped in every region.

Expected: FY27 reading and assessment data flowing daily, as FY26 did.

Actual, verified against Dagster on 2026-08-21:

| Asset | State |
| --- | --- |
| `kippmiami/iready/personalized_instruction_summary` | `math` materializes daily; `ela` stopped 2026-07-18. No FY27 reading. |
| `kippmiami/iready/diagnostic_results` | Nothing since 2026-07-18. FY27 partition holds 3,933 FY26 rows. |
| `kippmiami/iready/personalized_instruction_by_lesson` | Reading not matched. No FY27 reading. |
| `kippmiami/iready/instruction_by_lesson` (pro) | Reading not matched. No FY27 reading. |

Newark is on the same footing. Miami is roughly five weeks into its school year with no FY27 reading instruction data and no FY27 assessment data at all.

The `diagnostic_results` FY27 partition is the sharper problem. Every July the vendor's `Current_Year` folder lags its own rollover, so the new fiscal year's partition transiently holds the prior year's file, then self-corrects. This year it will not, because the export was renamed. So `_dagster_partition_academic_year=2026` is permanently pinned to FY26 content unless we intervene.

## Steps to reproduce

1. List the i-Ready SFTP export folders for either region and compare against the asset regexes in `src/teamster/code_locations/kippmiami/iready/assets.py`. The `_reading_` and `i-ready_inform_results_` files present in `Current_Year` match nothing.
1. Check materialization history for `kippmiami/iready/diagnostic_results` — last run 2026-07-18.
1. Check materialization history for `kippmiami/iready/personalized_instruction_summary` — every recent run is partition `2026 pipe math`; no `ela` since 2026-07-18.

## Where

- **Code location / dbt project:** `kippmiami` and `kippnewark`; dbt projects `iready`, `kippmiami`, `kippnewark`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** design doc on branch `anthonygwalters/fix/claude-iready-2027-export-renames`

## Change 1: subject token `ela` becomes `reading`

Headers are byte-identical to the old files, so this is a filename change only.

| Previous | New |
| --- | --- |
| `personalized_instruction_summary_ela_CONFIDENTIAL` | `personalized_instruction_summary_reading_CONFIDENTIAL` |
| `iready_instruction_by_lesson_ela_CONFIDENTIAL` | `iready_instruction_by_lesson_reading_CONFIDENTIAL` |
| `iready_pro_instruction_by_lesson_ela_CONFIDENTIAL` | `iready_pro_instruction_by_lesson_reading_CONFIDENTIAL` |
| `standards_results_by_test_ela_CONFIDENTIAL` | `standards_results_by_test_reading_CONFIDENTIAL` |
| `standards_results_ytd_ela_CONFIDENTIAL` | `standards_results_ytd_reading_CONFIDENTIAL` |

## Change 2: Diagnostic Results becomes i-Ready Inform Results

i-Ready Inform is the vendor's new name for the assessment previously called i-Ready Diagnostic. This is a rename of one continuous dataset, not a new report.

| Previous | New |
| --- | --- |
| `diagnostic_results_ela_CONFIDENTIAL` | `i-ready_inform_results_reading_english_CONFIDENTIAL` |
| `diagnostic_results_math_CONFIDENTIAL` | `i-ready_inform_results_math_CONFIDENTIAL` |
| `diagnostic_results_reading_spanish_CONFIDENTIAL` | `i-ready_inform_results_reading_spanish_CONFIDENTIAL` |

Two naming irregularities to design around: the prefix is `i-ready` (hyphenated, unlike every other export), and the reading subject token is `reading_english`, not `reading`.

Field renames in this export:

| Previous | New |
| --- | --- |
| `Baseline Diagnostic (Y/N)` | `Baseline Assessment (Y/N)` |
| `Most Recent Diagnostic YTD (Y/N)` | `Most Recent Assessment YTD (Y/N)` |
| `Diagnostic Gain` | `Assessment Gain` |
| `Diagnostic Language` (math only) | `Assessment Language` (math only) |

One genuinely new field: `Tactile Graphics`, on all subjects.

## Scope

1. Restore FY27 ingestion for the three renamed Personalized Instruction exports.
1. Absorb the Diagnostic to Inform cutover, including the renamed fields and `Tactile Graphics`.
1. Ingest `standards_results_by_test` and `standards_results_ytd`, which have never been ingested. Per-grade files (`_2` through `_8`) whose headers carry grade-specific and state-specific standard columns, with embedded newlines inside quoted column names. Needs its own ingestion shape and its own spec.
1. Retire `instructional_usage_data`. This lineage is dead: no Dagster asset in any code location, staging disabled in both `kippmiami` and `kippnewark`, snapshot `enabled: false`, and no consumer. Delete it rather than building the missing Miami asset.

Reading (Spanish) is in the vendor rename table but no `reading_spanish` file has appeared in either region yet. The partition design should accommodate it without building it now.

<details>
<summary>For Claude</summary>

Vendor source of truth: `faq-what-changes-can-I-expect-with-iReady-exports-for-the-2026_2027-school-year` (Curriculum Associates, 04/26, W3519034). Extracted text at `.claude/scratch/iready_vendor_faq.txt`; SFTP listings and header diffs at `.claude/scratch/iready_probe.json`, `.claude/scratch/iready_miami_listing.txt`, `.claude/scratch/iready_years.json`.

Design doc: `docs/superpowers/specs/2026-08-21-iready-2027-export-renames-design.md` on branch `anthonygwalters/fix/claude-iready-2027-export-renames`.

Key constraints found while scoping:

- `subject` is a `StaticPartitionsDefinition` of `ela` and `math` in `src/teamster/libraries/iready/assets.py`, shared by both code locations.
- `regex_pattern_replace` substitutes the whole `(?P{subject}...)` named group with the literal partition value, so each partition resolves exactly one remote file. Its group scanner also rejects nested parentheses inside a named group, so any optional suffix has to sit outside it.
- `src/dbt/iready/models/staging/stg_iready__diagnostic_results.sql` has 22 hardcoded comparisons against `_dagster_partition_subject = 'ela'` in the high-school growth-measure tables.
- The roughly 40 downstream kipptaf models join on `iready_subject` in `Reading` and `Math` against staging's already-normalized `subject` column, so they need no change provided staging maps the new token to `Reading`.
- The July 2026 work in 9cf84eb97 and 4637ec3d9 already declared every other new column in this vendor release. Only `tactile_graphics` and the four `assessment_*` renames remain undeclared.

The design doc records the chosen approach (boundary translation keeping the partition value as `ela`, plus a hard fiscal-year split between the Diagnostic and Inform assets) and states the trade-offs against the alternative explicitly, because the implementing engineer has final say on that call.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will restore FY2027 i-Ready ingestion in Newark, Camden, and Miami.

Curriculum Associates renamed the i-Ready SFTP exports for the 2026-2027 school year. None of our asset regexes matched the new filenames, so ingestion stopped on 2026-07-18. Miami is six weeks into its school year with no FY27 reading instruction data and no FY27 assessment data at all.

Two things changed at the vendor. The reading subject token `ela` became `reading`. And "Diagnostic Results" became "i-Ready Inform Results" — a rename of one dataset, not a new report. We confirmed that by diffing the file headers: 47 of 50 columns are identical and in the same order for math, 57 of 60 for reading. The whole difference is 3 renamed columns and 1 new one.

There was a second, quieter problem. The FY27 assessment partition had already run once against a leftover FY26 file, so it held 3,933 rows of last year's data labelled as this year's. Anything reading that partition was serving the wrong year.

The fix keeps the Dagster partition subject as `ela` in every school year and translates to the vendor's filename token at the SFTP boundary. Old archive files still resolve to their old names; new files resolve to the new ones. Nothing about the partition layout or the stored data moves.

## Reviewer Notes

**Please do not merge this yet.** The dbt Cloud check will fail on `stg_iready__diagnostic_results`, and that is expected. The 4 new columns do not exist in BigQuery until the Dagster fix deploys and the assets actually pull a new file. We confirmed the columns are absent from every schema today. Merging the dbt half before production catches up would break the whole i-Ready staging model, not just FY27. See the fold-out for the exact ordering.

**One decision needs an analyst's opinion.** `stg_iready__diagnostic_results` now reports `_dagster_partition_subject` as `reading` even though the underlying storage partition is still named `ela`. That was a deliberate choice, so the warehouse speaks the vendor's current language. The catch we found while building it: `stg_iready__personalized_instruction_summary` still reports `ela` for that same partition. So one column name now carries two vocabularies inside one package. Tell us if you would rather keep the raw column honest and add a separate translated column beside it.

**Two things we found that are not this change's fault**, each getting its own issue: Newark has no `iready_pro_instruction_by_lesson_math` file on the SFTP at all, and the `standards_results` exports have never been ingested.

### General

      code location

- [x] Include (or update) an
      summary above.
      See

## CI checks

## Troubleshooting

Refs #4949

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Design spec: `docs/superpowers/specs/2026-09-01-iready-fy27-export-renames-design.md`. Plan: `docs/superpowers/plans/2026-09-01-iready-fy27-export-renames.md`.

**Merge ordering — the load-bearing detail.** `src_iready__diagnostic_results` is an Avro-backed BigQuery external table whose columns are inferred from the GCS files present. `assessment_gain`, `baseline_assessment_y_n`, `most_recent_assessment_ytd_y_n` and `tactile_graphics` return 0 rows from `INFORMATION_SCHEMA.COLUMNS` in every schema, including `zz_stg_*`. The table is also `metadata_cache_mode: MANUAL` with `max_staleness: INTERVAL 7 DAY`, so even after a file lands, `refresh_external_metadata_cache` must run before dbt sees the columns. Order: merge the Dagster half, let a sensor tick materialize FY27, run the refresh macro, then the dbt model builds. Precedent that inference does pick up the superset: `most_recent_diagnostic_y_n` and `most_recent_diagnostic_ytd_y_n` both exist today from a prior vendor rename.

**Why the partition token stays `ela`.** The GCS object path is derived from the partition key, and `MultiPartitionsDefinition` is a cartesian product of two static lists, so one asset cannot use `ela` for old years and `reading` for new ones. Three alternatives were rejected: making the partition `reading` everywhere writes FY24 data to a `_dagster_partition_subject=reading` path; adding `reading` as a third subject value creates an `ela` partition every future year that can never materialize; splitting each dataset into two assets does not help the three Personalized Instruction datasets, whose filename prefix never changed.

**The era boundary is a fixed fiscal year (`RENAME_ACADEMIC_YEAR = 2026`), deliberately not the pre-existing "is this the newest partition" test.** Both branches coexist in `build_sftp_file_asset`. Next July FY27 rolls into a `2026/` archive carrying the new names, so a `Current_Year`-keyed test would wrongly translate it back to `ela`. There is a unit test for exactly that.

**The `_dagster_partition_subject` translation sits in the final select on purpose.** The `hs_goals` CTE holds 22 comparisons against the raw `ela` and `math` tokens for high-school growth measures. Translating earlier silently gives grades 9-12 null growth measures instead of failing.

**Corrections to #4949 and the superseded design doc.** Both claim `Diagnostic Language` was renamed to `Assessment Language`. It was not — neither column exists in either subject in either region. Do not add an alias for it. Miami also retains archives back to 2020, not just 2024 and 2025.

**Verification.** 112 unit tests pass. Live SFTP end to end: 10 of 10 materializations, 8 current-era (`2026|ela`, all 4 assets, both regions) and 2 archive (`2025|ela`), every fetched filename matching the expected era. Miami's FY27 assessment partition moved off the stale 3,933 rows to 1,486 real FY27 rows. The live sensor produces `2026|ela` run requests in both regions from `_reading_` filenames. `tests/test_dagster_definitions.py` fails 2 of 6 on a pre-existing missing Focus database credential at `kippmiami/dlt/focus/assets.py:13`, thrown before i-Ready is reached and unrelated to this branch.

**The regression test guards the shipped regexes**, read from asset metadata for both code locations, not hand-copied literals. That gap was caught in review and the fix was mutation-tested: breaking a regex in either location fails the suite.

</details>


cbini:
Two notes on CI, one of which changes how to read a green check.

**The dbt Cloud check passed, and that does not mean what it looks like.** It is not evidence that the dbt change in this PR works.

The CI job builds the `kipptaf` project. `kipptaf` does not consume the local `iready` package — it reads `stg_iready__diagnostic_results` as a **source**, a table the district projects produce. The changed model lives in `src/dbt/iready`, which only `kippmiami` and `kippnewark` consume. So `state:modified+` selected nothing from our change and the step finished in 36 seconds having built none of it.

The practical consequence: the model in this PR has no CI coverage at all. Its first ever execution will be in production, through the Dagster automation-condition sensor, against the real external table. That is exactly where the missing columns bite.

So the ordering note in the description still stands, and is now sharper. Before this model runs in production, three things must have happened in order: the Dagster half deploys, a sensor tick materializes an FY27 partition so new-schema Avro lands in GCS, and `refresh_external_metadata_cache` runs. The external table is `metadata_cache_mode: MANUAL` with `max_staleness: INTERVAL 7 DAY`, so that last step does not happen on its own.

Given there is no CI safety net here, splitting this into two PRs — Dagster first, dbt after production has caught up — is worth reconsidering. It would end the ingestion outage immediately and let the dbt half land against a warehouse that already has the columns.

**Trunk Check failed on one real issue and is now fixed** in a207069d3: `ruff/B007`, an unused loop variable I introduced in the sensor while making it match archive files. Tests still pass, 109 of them.

cbini:
Verdicts on the four review findings. I checked each against the code before deciding; no code changes from this pass.

**1. Sensor stale-file guard doesn't engage for 3 of 4 assets — accurate observation, declining the fix.**

The mechanism described is right: for the three Personalized Instruction assets the current-era regex keeps `ela` in its alternation, so a legacy-named file matches the primary pattern and never reaches the `is_legacy_year` fallback. That fallback is effectively dead code for those three, since their filename prefix never changed. Correct reading.

Two things make it a non-defect rather than a Medium finding.

The stale files cannot trigger anything. The sensor only builds a run request for files with `st_mtime > last_run`, and it writes the cursor forward to now on every tick that matches anything. The leftovers are frozen at 2026-07-18 and the cursor has been past that for six weeks.

And if they somehow did match — say after a cursor reset — the result is not a duplicate run. The sensor groups run requests by `(job_name, partition_key)` before emitting them. The stale `_ela_` file and the live `_reading_` file both resolve to partition `2026|ela`, land in the same group, and collapse into a single run. That is the behavior `partition_subject`'s docstring describes: the stale file can act as a trigger, never as the payload. The review confirms the fetched file stays correct.

Removing `ela` from the current-era alternation would also push archive matching for those three assets onto the fallback path, which is a behavior change to regexes we mutation-tested, in exchange for no observable difference. Leaving them.

**2. Cross-library import into the shared SFTP factory — declining, with the constraint that forces it.**

The suggested fix is to move the era-translation branch into `build_iready_sftp_asset`. That is not reachable: the regex composition happens inside the asset body that `build_sftp_file_asset` generates, so only the factory can put logic there. `build_iready_sftp_asset` just calls the factory and cannot inject into the body it returns. Passing a callable through instead is possible but adds machinery without removing the coupling, and the sensor separately needs the pattern as a plain string in asset metadata.

Worth noting the `if group_name == "iready"` branch predates this PR — the change deepens an existing seam rather than opening a new one — and `subjects.py` imports nothing from `teamster`, so there is no cycle risk.

**3. `coalesce()` inside a double `cast()` — the convention text is real, the codebase disagrees with it.**

The quoted rule exists at `src/dbt/CLAUDE.md:1050,1056`. But `cast(cast(x as numeric) as int)` appears 99 times across 74 files, and 19 times in this model alone, including on the exact line this PR touched — `cast(cast(diagnostic_gain as numeric) as int)` is what was there on main. The double cast is also load-bearing, not stylistic: the source column is a string like `5.0`, which casts to numeric and then to int but fails a direct int cast.

So this PR added one `coalesce` inside a pattern that is already the house idiom. Satisfying the rule as written would mean splitting the coalesce and the cast across separate CTEs in a 300-line model, which is more churn than the readability gain justifies. Happy to do it if a reviewer disagrees.

**4. Coalesce priority order — already verified, no action.**

The proposed query cannot run yet, since the new columns do not exist in the external table. But the header diff settles it more directly: comparing the old and new exports, the 3 renamed columns appear only in the old file and their counterparts only in the new one, in both subjects and both regions. Any given row comes from exactly one file, so an old-name and new-name pair can never both be populated on the same row. Legacy-first ordering is safe.

cbini:
Probe result, kept short for the record: the GitHub MCP **write** tools do not alter body text.

This comment was posted through `mcp__github__add_issue_comment` carrying bare angle-bracket tokens, `&`, double quotes and apostrophes in prose, in a code span, and inside a fenced block. The stored body came back identical to what was sent — no stripping, no entity encoding.

The mangling is entirely on the **read** side. `pull_request_read` returns that same stored body with the angle-bracket tokens deleted and `&`, quotes, apostrophes, `<=` and `>=` entity-encoded. Verify a stored body with `gh api repos/{owner}/{repo}/pulls/{n} --jq .body`, never through the MCP read tools.

Guidance corrected in `.claude/context/github.md`.

PR #5120, opened 2026-09-03
feat(dbt): list gradebook audit flag reasons in one Tableau column

### LINKED ISSUE #5119: feat: list gradebook audit flag reasons in one Tableau display column
## What and why

The gradebook audit dashboard tells a teacher a category is off track while also
showing they entered enough assignments. Both numbers look fine and the flag
still fires. Nothing on the dashboard explains the gap.

Flagged assignments cause the gap. `not_enough_assignments` compares
`expectation` to the count of assignments carrying no flags. The dashboard shows
the raw entered count instead. A teacher sees 5 entered against 5 expected plus
an off-track flag, and the reconciling number stays hidden.

Add a text column that lists the reasons in plain language. Teachers and school
leaders drop it straight onto a Tableau view. They write no calculated field.

## Proposed approach

Add 2 columns to `rpt_tableau__gradebook_audit`, both on
`row_type = 'category_summary'` rows:

- `flag_reasons` — the checks that tripped for this section, quarter and
  category, joined by a semicolon. NULL when nothing tripped.
- `assignments_entered_count_no_flags` — the count that
  `not_enough_assignments` actually compares to `expectation`.

The 4 reason labels map 1 to 1 onto the 4 terms of `assignment_has_flags` in
`int_powerschool__gradebook_assignment_scores_rollup`:

| Source term                  | Label                   |
| ---------------------------- | ----------------------- |
| `percent_graded_min_not_met` | Under 90% graded        |
| `flags_sum` greater than 0   | Invalid scores entered  |
| `assign_max_score_not_10`    | Not out of 10 points    |
| `overly_exempt_assignment`   | Half the class exempt   |

The labels always print in that order, so Tableau can group and filter on the
string.

<details>
<summary>For Claude</summary>

**Files**

- `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gradebook_audit.sql`
- `src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__gradebook_audit.yml`
- `docs/models/gradebook-audit-data-model.md` — the "No per-rule reason is ever
  surfaced" claim and the `rpt_tableau__gradebook_audit` CTE-chain section both
  become wrong
- `.claude/skills/gradebook-audit/SKILL.md` — the "do NOT add a reason column"
  rule needs narrowing to student-level flags only

**Implementation**

None of the 4 terms reach the report today; only `assignment_has_flags` does.
`category_summary` is a `SELECT DISTINCT` grain projection, so each new column
must be functionally determined by
`(_dbt_source_project, sectionid, quarter, assignment_category_code)`. Project
the 4 terms into `category_join` as window counts over that same partition,
reusing the existing due-date scoping:

```sql
countif(r.duedate <= ge.week_end_sunday and r.percent_graded_min_not_met) over (
    partition by
        s._dbt_source_project,
        s.sectionid,
        s.`quarter`,
        ge.assignment_category_code
) as n_percent_graded_min_not_met,
```

Repeat for `r.flags_sum > 0`, `r.assign_max_score_not_10` and
`r.overly_exempt_assignment`. Same partition as the 3 existing windows, so
BigQuery computes one window.

Build the string in `category_summary` next to `not_enough_assignments`:

```sql
array_to_string(
    [
        if(n_percent_graded_min_not_met > 0, 'Under 90% graded', null),
        if(n_invalid_scores > 0, 'Invalid scores entered', null),
        if(n_max_score_not_10 > 0, 'Not out of 10 points', null),
        if(n_overly_exempt > 0, 'Half the class exempt', null)
    ],
    '; '
) as flag_reasons
```

`array_to_string` drops NULL elements and the delimiter before them when the
call passes no `null_text`, which gives the NULL empty state for free. Confirm
that against the BigQuery docs first. If it does not hold, fall back to
`concat` plus `ltrim` plus `nullif`.

Also project the already-computed `assignments_entered_count_no_flags` window
from `category_join` into `category_summary`.

Thread both new columns through: `cast(null as string) as flag_reasons` and
`cast(null as int64) as assignments_entered_count_no_flags` on the
`assignment_detail` branch of `combined`, matching the existing
null-placeholder pattern. `with_section_flags` picks them up through `c.*`. Add
both to the final `SELECT`.

**Out of scope**

- No change to `health_calc` or either health boolean. `flag_reasons` restates
  checks that `not_enough_assignments` already feeds into `health_calc`, so
  wiring it in would double-count.
- `flag_reasons` populates whenever a check tripped, even when
  `not_enough_assignments` is false. Gating it on the flag is a separate
  decision the requester is taking to stakeholders.

**Acceptance criteria**

- Every section-quarter still has exactly 4 `category_summary` rows.
- `dbt_utils.unique_combination_of_columns` on
  `rpt_tableau__gradebook_audit` passes.
- `flag_reasons` is NULL on every `assignment_detail` row.
- `flag_reasons` is NULL on a `category_summary` row where all 4 counts are 0,
  and non-NULL wherever any count is above 0.
- Rows where `assignments_entered_count` is at least `expectation` and
  `not_enough_assignments` is true always carry a non-NULL `flag_reasons` — the
  case this issue exists to explain.
- No student PII enters the model. Every new column sits at assignment or
  section grain.
- `trunk check --force` passes on the changed SQL.

</details>

anthonygwalters:
## Summary & Motivation

When merged, this pull request will add `flag_reasons` to `rpt_tableau__gradebook_audit`.

The dashboard could tell a teacher a category was off track while showing they entered enough assignments. Both numbers looked right and the flag still fired. Nothing explained the gap.

Flagged assignments cause the gap. `not_enough_assignments` compares `expectation` to `assignments_entered_count_no_flags`, not to the raw entered count. `flag_reasons` names the checks that tripped, in plain language, on the `category_summary` row.

Teachers and school leaders drop the column onto a Tableau view. They write no calculated field.

The 4 labels are `Under 90% graded`, `Invalid scores entered`, `Not out of 10 points` and `Half the class exempt`. They print in a fixed order, so Tableau can group and filter on the string. The column is NULL when no check tripped.

This change adds a column and no rows. It renames nothing and removes nothing, so no downstream exposure breaks.

Refs #5119

## Reviewer Notes

- `category_summary` is a `SELECT DISTINCT`, so a new column that varied within its grain would silently add rows. The 4 labels come from window counts over the exact partition it collapses on. I measured the result: 25,480 partitions produce exactly 25,480 rows, and no partition carries more than 1 label string.
- `flag_reasons` fills in whenever a check tripped, even when `not_enough_assignments` is false. A category can hold flawed assignments and still clear its expectation. That choice goes to stakeholders separately, and reverting it is a 1-line change.
- `Under 90% graded` never prints alone. Grading below 90% of expected students leaves unscored students, which trips the invalid-scores check on the same assignment.
- Neither health boolean changed. `not_enough_assignments` already feeds both, so adding `flag_reasons` would double-count.
- `Invalid scores entered` stands in for 7 separate score checks. A teacher who sees it alone must open the assignment rows to learn which one tripped. The requester chose the 4-label set knowing this.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — the existing `rpt_tableau__gradebook_audit.yml` gains the `flag_reasons` column and an updated model description. The model has an enforced contract, so the build fails without it.
- [x] Include (or update) an [exposure](https://teamschools.github.io/teamster/reference/dbt-conventions/#exposures) for all models consumed by a dashboard, analysis, or application — the `gradebook_audit` exposure already depends on this model and needs no change, because this adds a column rather than a model.
- [x] **Breaking change?** No. This adds 1 column and renames or removes nothing, so contracted downstream consumers keep working.
- [ ] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — not applicable, no external source changed.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

**Implementation**

`category_join` gains 4 window `countif`s over `(_dbt_source_project, sectionid, quarter, assignment_category_code)` — the same partition the 3 pre-existing windows use, so BigQuery computes one window for all 7. They count assignments tripping each term of `assignment_has_flags` in `int_powerschool__gradebook_assignment_scores_rollup`: `n_percent_graded_min_not_met`, `n_invalid_scores` (the `flags_sum > 0` term), `n_max_score_not_10`, `n_overly_exempt`. The due-date scope (`r.duedate <= ge.week_end_sunday`) is copied from `assignments_entered_count_no_flags`, so `flag_reasons` and `not_enough_assignments` can never disagree about which assignments are in scope.

`category_summary` builds the string with `array_to_string` over conditional labels, wrapped in `nullif(..., '')`. The wrapper is required: `array_to_string` drops NULL elements and the delimiter before them, but returns `''`, not NULL, when every element is NULL. Confirmed by direct query, not from memory.

`combined` carries `cast(null as string) as flag_reasons` on the `assignment_detail` branch, matching the existing null-placeholder pattern. `with_section_flags` picks it up through `c.*`.

**Grain safety**

Because no `OVER()` clause carries an `ORDER BY`, each window frame is the whole partition, so every new count is constant across the partition by SQL semantics rather than by luck of current data. `flag_reasons` is a deterministic scalar over those constants, so `SELECT DISTINCT` remains a pure grain projection.

**Verification against the built view**

`zz_anthonygwalters_kipptaf_tableau.rpt_tableau__gradebook_audit`, built with `--defer` against the prod manifest:

- 25,480 `category_summary` rows against 25,480 distinct partitions, so zero fan-out; 0 partitions with a multivalued `flag_reasons`
- 0 section-quarters with other than exactly 4 `category_summary` rows
- 0 `assignment_detail` rows carrying a non-NULL `flag_reasons`
- 0 rows where `flag_reasons is not null` disagrees with `assignments_entered_count > assignments_entered_count_no_flags`, across all 25,480 category rows
- all 4,975 rows in the target case (entered at or above expectation plus `not_enough_assignments`) carry a reason
- 11 distinct label combinations, every one in the fixed order
- `dbt_utils.unique_combination_of_columns` passes; 27,433 `assignment_detail` rows and 52,913 rows total

**The doc claim about `Under 90% graded`**

`is_expected_scored` and `is_expected_null` are complements given `is_expected`, so `n_expected = n_expected_null + n_expected_scored` exactly. `assign_percent_graded < 0.90` therefore forces `n_expected_null > 0`, which makes `flags_sum > 0` for that same assignment. Derived from `int_powerschool__gradebook_assignments_scores.sql` and confirmed empirically: 0 rows carry `Under 90% graded` without `Invalid scores entered`.

**Doc and skill edits**

The reference doc previously claimed no per-rule reason is ever surfaced. That rule came from the July 2026 teacher/student split, whose concern was student flags: they carry PII and fan out per student. Assignment flags sit at section grain, carry no student identifiers, and come from a fixed set of 4 checks. The rule is now scoped to student flags, and the skill gains an "Assignment-level check" procedure covering how to add a 5th label.

**Known gap, deliberately not closed**

No dbt unit test covers the `flag_reasons` derivation. A reviewer suggested one. The model carries 51 contract columns and 4 refs, and dbt requires every `expect` row to list every column, so covering the 4 conditions plus an ordering case runs to roughly 300 lines of fixture. This model gained 2 columns in the past week (`school_name`, `assignments_entered_count_no_flags`), and a column add breaks its own unit test, so the fixture would need editing on each one. The sibling `not_enough_assignments` flag has no unit test either. Left out as a cost call, not an oversight — worth revisiting if the label set stabilizes.

**AI involvement**

Claude-assisted, human-directed. Design settled through the `brainstorming` skill; reviewed by a dispatched reviewer subagent, which found no critical or important issues. Its 1 acted-on suggestion was the guardrail comment in `386c1a8a4`.

</details>

### DIFF
--- a/.claude/skills/gradebook-audit/SKILL.md
+++ b/.claude/skills/gradebook-audit/SKILL.md
@@ -183,8 +183,11 @@ present it as fact.
 `stg_google_sheets__gradebook_flags` is disabled — no sheet step needed. Since
 the July 2026 teacher/student split there are no UNPIVOT lists — every flag is a
 hardcoded boolean column, and which model it lives in depends on its grain. Per
-the split's design goal, do NOT add a "reason" column explaining why a flag
-fired — flags stay aggregated booleans.
+the split's design goal, do NOT add a "reason" column to a **student-level**
+flag — those stay aggregated booleans, because they carry PII and fan out per
+student. **Assignment-level** checks are the exception: `flag_reasons` on the
+`category_summary` row already names them in plain language, so a new
+assignment-level check needs a label added there (see "Add a new flag" below).
 
 **Student-level flag** (per student × section × quarter, like
 `qt_percent_grade_greater_100`/`qt_grade_70_comment_missing`):
@@ -209,6 +212,20 @@ fired — flags stay aggregated booleans.
 4. Build in dependency order — `int_extracts__gradebook_audit_student_flags`
    first, then `rpt_gsheets__gradebook_audit_student_flags` and
    `rpt_tableau__gradebook_audit` (both read the int).
+
+**Assignment-level check** (per assignment, like `percent_graded_min_not_met`):
+
+1. Add the check to `int_powerschool__gradebook_assignment_scores_rollup.sql`
+   and fold it into `assignment_has_flags`.
+2. In `rpt_tableau__gradebook_audit.sql`'s `category_join`, add a window
+   `countif` for it over the existing partition
+   (`_dbt_source_project, sectionid, quarter, assignment_category_code`),
+   matching its 4 siblings.
+3. Add a plain-language label to the `array_to_string` array in
+   `category_summary`, keeping the array order stable — Tableau groups on the
+   string, so reordering changes existing values. Do NOT reword an existing
+   label without telling the requester; the strings are user-facing.
+4. Update the properties YAML for both models.
 
 **Category-level flag** (per section × quarter × category, like
 `not_enough_assignments`):
--- a/docs/models/gradebook-audit-data-model.md
+++ b/docs/models/gradebook-audit-data-model.md
@@ -487,8 +487,8 @@ concept, unlike the pre-July-2026 design below.
    more window `countif`s over the same partition count the assignments tripping
    each term of `assignment_has_flags`: `n_percent_graded_min_not_met`,
    `n_invalid_scores` (the `flags_sum > 0` term), `n_max_score_not_10` and
-   `n_overly_exempt`. They share the existing partition, so BigQuery computes
-   one window for all seven.
+   `n_overly_exempt`. They reuse the existing partition rather than introducing
+   a second one.
 2. `category_summary` — a grain-projection `SELECT DISTINCT` over
    `category_join` that collapses the per-assignment fan-out to one row per
    section × quarter × category, and adds `not_enough_assignments`
--- a/docs/models/gradebook-audit-data-model.md
+++ b/docs/models/gradebook-audit-data-model.md
@@ -38,7 +38,8 @@ The audit operates at several grains:
 - **Assignment** — did this assignment receive valid scores (correct point
   value, enough of the class graded, not over-exempt)? This is the
   `assignment_has_flags` rollup, surfaced on the dashboard as per-assignment
-  detail rows.
+  detail rows and, in plain language, in the `flag_reasons` column on the
+  category row.
 - **Section-category, quarter-to-date** — has this teacher entered the required
   number of assignments in this category so far this quarter? (Quarter-grain as
   of AY 2026-2027 — previously a weekly check.)
@@ -455,9 +456,13 @@ this is an internal ops artifact, unlike the Tableau-facing report below.
 
 `rpt_tableau__gradebook_audit` reads the intermediate above (aggregated, with
 PII dropped) for its section-level "any student flagged" booleans — see below.
-No per-rule "reason" is ever surfaced for either flag; the whole point of the
-July 2026 split was to move to aggregated boolean signals instead of granular
-rule-level detail.
+No per-rule "reason" is ever surfaced for either **student** flag; the whole
+point of the July 2026 split was to move to aggregated boolean signals instead
+of granular rule-level detail. That rule is scoped to the student flags, which
+carry PII and fan out per student. **Assignment** flags do carry a reason:
+`flag_reasons` names them in plain language on the category row (see below).
+Assignment flags sit at section grain, carry no PII, and come from a fixed set
+of 4 checks, so naming them adds no rows and exposes nothing.
 
 ### Final extract: `rpt_tableau__gradebook_audit`
 
@@ -478,7 +483,12 @@ concept, unlike the pre-July-2026 design below.
    rollup data. Computes `expectation`, `assignments_entered_count`, and
    `assignments_entered_count_no_flags` as **window functions** partitioned by
    `_dbt_source_project, sectionid, quarter, assignment_category_code` — not a
-   `GROUP BY` — so the per-assignment row grain survives for step 2 to use.
+   `GROUP BY` — so the per-assignment row grain survives for step 2 to use. Four
+   more window `countif`s over the same partition count the assignments tripping
+   each term of `assignment_has_flags`: `n_percent_graded_min_not_met`,
+   `n_invalid_scores` (the `flags_sum > 0` term), `n_max_score_not_10` and
+   `n_overly_exempt`. They share the existing partition, so BigQuery computes
+   one window for all seven.
 2. `category_summary` — a grain-projection `SELECT DISTINCT` over
    `category_join` that collapses the per-assignment fan-out to one row per
    section × quarter × category, and adds `not_enough_assignments`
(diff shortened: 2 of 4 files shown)


anthonygwalters:
Verdict on each of the 5 findings. Fixed in `fbf23dcbf`.

**1. YAML field order — fixed.** Verified before changing: `flag_reasons` was the only column in the file with `description:` before `data_type:`, and 0 other columns order it that way. `data_type: string` now comes first. Rebuilt after the change — the enforced contract still passes.

**2. No unit test on `flag_reasons` — declining, and you are right that a human should say so out loud rather than let it ride.** The requester has the call. My reasoning, so the decision is on the record: the model carries 51 contract columns and 4 refs, and dbt requires every `expect` row to list every column, so covering the 4 conditions plus a fixed-order case runs to roughly 300 lines of fixture. This model gained 2 columns in the past week (`school_name`, `assignments_entered_count_no_flags`), and a column add breaks the model's own unit test, so that fixture would need editing on each one. The sibling `not_enough_assignments` flag has no unit test either. Worth revisiting once the label set stops moving.

**3. Row-count claims — already verified, and your suggested query targets the wrong relation.** The numbers came from a real `--defer` build, not from reasoning. One correction worth having: the query in your fold-out reads `kipptaf_tableau.rpt_tableau__gradebook_audit`, which is prod — prod does not have `flag_reasons` until this merges, so that query cannot verify a pre-merge claim. The right target is the dev build, `zz_anthonygwalters_kipptaf_tableau.rpt_tableau__gradebook_audit`. Against it: 25,480 `category_summary` rows against 25,480 distinct partitions (zero fan-out), 0 partitions with a multivalued `flag_reasons`, 0 section-quarters with other than exactly 4 `category_summary` rows, 0 `assignment_detail` rows carrying a non-NULL `flag_reasons`, and 0 rows where `flag_reasons is not null` disagrees with `assignments_entered_count > assignments_entered_count_no_flags`. `dbt_utils.unique_combination_of_columns` passes.

**4. Doc performance claim — fixed, good catch.** You are right that nobody checked an execution plan. The claim is gone rather than verified; the doc now says the new counts "reuse the existing partition rather than introducing a second one," which is checkable by reading the SQL. A stronger correctness argument survives without any planner assumption: no `OVER()` clause carries an `ORDER BY`, so each frame is the whole partition and each count is constant per partition by SQL semantics, which is what the `SELECT DISTINCT` grain projection actually depends on.

**5. Uncommitted skill-doc revert — dismissing as a CI-checkout artifact.** `git status --porcelain` on the working tree is empty, and the committed blob carries the skill edit. This repo has seen this phantom "working-tree revert" from this bot before; it is documented in `.claude/CLAUDE.md` as a CI-checkout artifact, with an explicit instruction not to `git checkout` to "fix" a clean tree. No action taken.

One note on the ST06 item you logged under "checked, no issue found": logging the contradicting in-file precedent rather than reporting a violation was the right call, and the precedent you cite is real.

anthonygwalters:
Finding 2 (no unit test on `flag_reasons`): the requester has signed off on the omission. Recording it here because the review specifically asked for an explicit human yes rather than letting the PR description carry the decision.

The tradeoff stands as written: ~300 lines of fixture for 51 contract columns across 4 mocked refs, on a model that gained 2 columns in the past week, where a column add breaks the fixture. Revisit if the label set stabilizes.

Issue #3698, opened 2026-04-22
feat(dbt): add historical school grade levels reference (grade_band versioned by academic_year)


cbini:
## Summary

Promoted from #3631 as a concrete standalone tracker. School grade bands (ES / MS / HS) change year over year as KIPP opens grade levels or shutters programs. A versioned reference is needed so Cube queries over historical periods resolve to the correct grade_band for that period.

## Current state

`dim_locations.grade_band` carries a single current value per school — any historical query yields today's grade_band even for prior years.

## Scope

- Identify the source of truth for school grade bands (currently does not exist; may require ops-owned curation).
- Create a reference table with (school, academic_year, grade_band) grain.
- Update facts that reference grade_band historically to join through the reference instead of the current-value dim column.

Promoted from #3631.

cbini:
## Curation request

Engineering is blocked on this until a source of truth for `(school, academic_year, grade_band)` exists. Today `dim_locations.grade_band` carries one current value per school — historical Cube queries return today's grade_band for prior years.

**Need:** a curated reference covering each school × academic year (back to whichever year analytics needs to query historically). Format flexible — Google Sheet, seed CSV, or whatever's easiest to maintain.

**Suggested columns:** `school`, `academic_year`, `grade_band` (ES / MS / HS), and optionally `grade_low` / `grade_high` so the band rule can be reconstructed if definitions change.

Once curated, engineering will wire it into the dim family and reroute historical fact joins to resolve grade_band by year.

PR #5124, opened 2026-09-03
feat(dbt): add int_students__school_directory

### LINKED ISSUE #5123: feat(dbt): one school and grade level directory by academic year
## What and why

Every dashboard works out which schools and grade levels to count, and each one does it differently. The DIBELS progress monitoring models resolve it one way. `int_tableau__fresh_enrollment_scaffold` resolves it another way. The CSGF annual collection needs it next. I am writing the same logic a third time.

We also cannot answer which schools offered which grade levels in a past year. Schools open, close, merge, and add grades. That history sits only in enrollment records today, so every consumer re-derives it.

One view answers both needs.

## Proposed approach

Add `int_students__school_directory`. It holds one row per academic year, region, school, and grade level.

Derive it from enrollment, not from school metadata. A school and grade then appear in exactly the years students enrolled in them. Nobody maintains an open or close date, and a school that never enrolled anyone never appears.

Three branches, labeled by `school_source`:

- `powerschool` — every region-year that PowerSchool holds enrollment rows for
- `focus` — Miami from SY26-27 onward, where its PowerSchool archive stops
- `finalsite` — next year's schools and grades, before any student information system carries them

Carry school attributes alongside the keys so a consumer does not need its own crosswalk join: `school_name`, `school_short_name`, `school_level`, `focus_school_id`, and `school_level_alt`.

Enrollment beats metadata, and Miami shows why. `int_focus__schools.max_syear` reads 2025 for KIPP Sunrise and KIPP Liberty. The last students enrolled at both schools in 2022. The metadata is wrong by 3 years. Enrollment is right.

Paterson shows the history need. Its grade levels ran 0-2 and 5-6 in SY24-25, then added grade 3 and grade 7, then added grade 4 and grade 8. Only enrollment records that growth.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Shipped in #5124.

**Grain**: `academic_year` / `region` / `ps_schoolid` / `grade_level`. Keyed on `ps_schoolid` rather than `schoolid` on purpose — that also catches two Focus schools mapping to one PowerSchool id, which is what a school merge looks like.

**12 columns**: `_dbt_source_project`, `academic_year`, `region`, `schoolid`, `ps_schoolid`, `focus_school_id`, `grade_level`, `school_source`, `school_name`, `school_short_name`, `school_level`, `school_level_alt`.

**Three ids, deliberately.** `schoolid` is native to whichever system produced the row — the network school number on PowerSchool rows, Focus's internal id on Focus rows. `ps_schoolid` carries the PowerSchool number on every row, so Miami's SY26-27 rows still join PowerSchool-keyed models: KIPP Courage is `schoolid` 30200803 in AY2025 and `schoolid` 15 in AY2026, with `ps_schoolid` 30200803 in both. `focus_school_id` carries Focus's own code, such as `2332B`, which is the key `int_tableau__fresh_enrollment_scaffold` joins `int_focus__schools` on.

**Sources**, one per branch:

- `int_powerschool__student_enrollment_union` — all 4 regions. Needs no Miami year filter: the frozen Miami archive stops at AY2025 on its own, because ingestion stopped and ingestion is gated by the team. Excludes `schoolid` 999999, the graduated-students placeholder.
- `int_focus__student_enrollment_roster` — Miami only. Filtered `academic_year >= 2026`, a fixed boundary rather than `current_academic_year`. Focus reaches back to AY2018 and would double-count against the archive. Also filtered `ps_schoolid is not null`, which drops Focus's non-instructional Applicants school.
- `stg_finalsite__status_report` joined to `int_people__location_crosswalk` on `location_name`, scoped `active_school_year_int = current_academic_year + 1`. Not `int_finalsite__status_report_unpivot` — the unpivot fans each applicant out per status field, and the crosswalk join is the only part of it this model needs.

**School attributes come from `stg_google_sheets__people__locations`, not the crosswalk.** The crosswalk carries one row per alias name — 254 rows across 29 schools — so joining it on a school id fans out. The locations sheet reaches one row per school once the no-school-assigned bucket, Pathways sites, and the single null-abbreviation row are filtered out: 28 rows, 28 distinct `ps_schoolid`. The crosswalk is still correct inside the Finalsite branch, where `location_name` is its unique key (254 of 254).

**`school_level_alt`** reproduces the CASE on `int_extracts__student_enrollments`: from AY2025, Sumner grade 5 and above reports MS while the locations sheet still calls the school ES. Fires on 3 rows.

**Why not `int_students__schools`**: school-grain identity with 5 mart consumers. Adding year and grade would change its grain underneath them. Its Focus branch also carries neither `schoolcity` nor `state_excludefromreporting`.

**Why not `int_extracts__student_enrollments`**: a wide denormalized extract. Carrying it for a handful of columns costs a lot for no gain. `int_powerschool__student_enrollment_union` is a 52-column table; `stg_powerschool__students` is a 156-column view, so reading staging directly would be both wider and deeper.

**Verified in dev**: 1,010 rows, 1,010 distinct grain keys, zero unmatched attributes, all 3 `school_source` values, 7 of 7 nodes pass. Miami hands off cleanly — `powerschool` covers 2018-2025 across 4 schools, `focus` covers 2026 across 5 schools, with no overlapping year.

**Known limitation.** The `finalsite` branch only sees a school once an applicant is routed to it by name. For SY27-28 today, 30 of 31 applicants have no `assigned_school`, so the branch yields 1 row. Empty is a normal state, not a fault.

A workaround was considered and rejected: have SRE enter placeholder students across every planned school and grade, then exclude those ids from the FRESH dashboard. `stg_finalsite__status_report` applies the `stg_google_sheets__finalsite__exclude_ids` filter itself, upstream of the directory, so excluded ids never reach it. Moving that filter downstream would change FRESH's calculations on a live dashboard.

**Nothing consumes the model yet.** It is safe to land before any consumer migrates.
</details>

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will add `int_students__school_directory`. It answers which schools offered which grade levels in which academic years, one row per `academic_year` / `region` / `ps_schoolid` / `grade_level`.

Two problems it solves. Every dashboard works out which schools and grade levels to count, and each one does it differently — the DIBELS progress monitoring models one way, `int_tableau__fresh_enrollment_scaffold` another, and the CSGF collection needs it next. We also cannot answer which schools offered which grades in a past year, because that history sits only in enrollment records.

The model derives everything from enrollment, not from school metadata. A school and grade appear in exactly the years students enrolled in them. Nobody maintains an open or close date.

Two branches, labeled by `school_source`. `sis` reads `int_students__student_enrollment_union`, which already resolves the PowerSchool-to-Focus handoff for Miami, so every historic and current year comes from one place. `finalsite` adds next year's incoming schools before any student information system carries them.

It carries school attributes alongside the keys — `school_name`, `school_short_name`, `school_level`, `focus_school_id`, `school_level_alt` — so a consumer does not need its own crosswalk join.

Nothing consumes the model yet, so it is safe to land ahead of any consumer. `int_tableau__fresh_enrollment_scaffold` is the intended first consumer; that migration is a follow-up, since the scaffold keys on `schoolid` and adds whole-school and region rollup rows this model does not carry.

## Reviewer Notes

**Two id columns.** `ps_schoolid` is the PowerSchool network school number on every row, including Miami's Focus-era rows, because the enrollment union resolves Focus schools to their PowerSchool number. `focus_school_id` carries Focus's own code for the reverse direction. An earlier revision also carried the Focus-native `schoolid`; it went away when the two SIS branches collapsed into one, because it always equaled `ps_schoolid`.

**Attributes come from the locations sheet, not `int_people__location_crosswalk`.** The crosswalk holds one row per alias name, so joining it on a school id fans out. It is still used inside the Finalsite branch, where `location_name` is its unique key.

**The grain test keys on `ps_schoolid`.** That also catches two Focus schools mapping to one PowerSchool id, which is what a school merge looks like.

**The Finalsite branch reads only the file for its own cycle year.** A status report file describes applicants in their grade for that file's cycle, so an older file's rows for next year carry last year's grade. Without the pin, the branch produced a phantom lower grade per school that the grain test cannot see. Measured on prod: 9,119 of 9,346 SY26-27 rows in the `2025_26` file are off by one against the SIS; the `2026_27` file matches on 11,486 of 11,505.

**The Finalsite branch is empty today, and that is correct.** It sees a school only once an applicant is routed to it by name, and only from the `2027_28` file, which lands when SRE opens that cycle. A new school with no routed applicant is not in the directory. That limit is inherent to the source.

**No boundary literal for the Miami handoff.** `int_students__student_enrollment_union` owns that rule (Focus carries Miami's full history back to AY2018; the frozen PowerSchool archive contributes no Miami rows). At this grain the two sources are identical sets for Miami through AY2025, so the earlier two-branch split bought nothing.

**Review history.** `claude-review` found two nullability gaps, fixed in `85ed19c7d`. An independent review then found the cycle-file grade drift and the redundant SIS split, fixed in `25f682a5b` — see the verdict comments on this PR.

### General

### dbt

- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — not applicable, no consumer yet
- [x] **Breaking change?** No. The model is new and nothing reads it.
      with **`--target staging`** — not applicable, no external source changes

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Refs #5123. Refs #4451, whose spec left the multi-year school × grade spine as an open question; this model is that follow-up. Refs #3698, which stays open: `school_level` here is still the locations sheet's single current value per school, not a grade band versioned by year.

**11 columns**: `_dbt_source_project`, `academic_year`, `region`, `ps_schoolid`, `focus_school_id`, `grade_level`, `school_source`, `school_name`, `school_short_name`, `school_level`, `school_level_alt`.

**Sources, one per branch:**

- `int_students__student_enrollment_union` — all 4 regions, all years. Its `schoolid` is already the PowerSchool number. Excludes `schoolid` 999999, the graduated-students placeholder, and null `grade_level`.
- `stg_finalsite__status_report` joined to `int_people__location_crosswalk` on `location_name`, scoped `active_school_year_int = current_academic_year + 1` and `left(_dagster_partition_key, 4) = active_school_year_int`. Excludes campus aliases (`location_powerschool_school_id = 0`) and Pathways aliases, matching the attributes CTE. Deliberately not `int_finalsite__status_report_unpivot`: the unpivot fans each applicant out per status field and the crosswalk join is the only part of it this model needs.

**Why one SIS branch, not PowerSchool plus Focus.** The first revision read `int_powerschool__student_enrollment_union` for all regions and `int_focus__student_enrollment_roster` for Miami from a hardcoded `academic_year >= 2026`. `int_students__student_enrollment_union` (13 mart consumers) already resolves that handoff and its YAML documents it. On prod, Miami PowerSchool ≤2025 and Focus ≤2025 are identical sets at this grain, so the split created a second directory of record with a conflicting rule and no extra rows.

**Why the partition pin.** `stg_finalsite__status_report` unions one file per recruitment cycle. The same `finalsite_enrollment_id` and `active_school_year_int` appear in consecutive files with a different `grade_level`, because each file reports the grade relative to its own cycle. Pinning each row to the file whose start year equals its `active_school_year_int` is the only way to read next year's grade correctly. FRESH does the same through the status crosswalk's `file_year`.

**Attribute join grain, checked before wiring it.** `int_people__location_crosswalk` is 254 rows across 29 `ps_schoolid` — one row per alias name — so joining it on a school id would have fanned the directory out roughly 1.45x and broken the grain test. `location_name` is its unique key (254 of 254), which is why the Finalsite branch still uses it. For the attributes, `stg_google_sheets__people__locations` reaches one row per school with three filters: `powerschool_school_id != 0` (the no-school-assigned bucket), `not is_pathways`, and `abbreviation is not null` — the last one because `ps_schoolid` 179901 carries two otherwise-identical rows differing only by a null abbreviation. Result is 28 rows, 28 distinct `ps_schoolid`, zero fan-out. The join is a `LEFT JOIN` so a school missing from the sheet would surface as null attributes rather than a dropped row; zero rows are unmatched today.

**`school_level_alt`** copies the CASE on `int_extracts__student_enrollments`, including its `TODO`. From AY2025, Sumner grade 5 and above reports MS while the sheet still calls the school ES. `school_level` itself is a school property, so it does not vary by grade or year.

**Why enrollment and not metadata.** `int_focus__schools.max_syear` reads 2025 for KIPP Sunrise and KIPP Liberty. Both schools last enrolled students in 2022 — confirmed in the frozen PowerSchool archive and in Focus. A `max_syear`-based directory would imply 7 Miami schools once operated when only the enrolled set is real. Paterson is the positive case: grade levels ran 0-2 and 5-6 in SY24-25, then gained grade 3 and grade 7, then grade 4 and grade 8. Only enrollment records that.

**Why not `int_students__schools`**: school-grain identity with 5 mart consumers. Adding year and grade would change its grain underneath them.

**Why not `int_extracts__student_enrollments`**: a wide denormalized extract. Carrying it for a handful of columns costs a lot for no gain.

**Nullability.** `grade_level is not null` appears on both branches, and `ps_schoolid` is filtered non-null on both, so neither carries a `not_null` test. `unique_combination_of_columns` does not reject nulls, which is why the filters matter on grain columns.

**Verified in dev at `25f682a5b`**: 1,010 rows, 1,010 distinct grain keys, zero unmatched attributes, 7 of 7 nodes pass. Camden 2014-2026, Miami 2018-2026, Newark 2004-2026, Paterson 2023-2026. The `finalsite` branch is empty; the one row it produced before the partition pin was a `2026_27`-file applicant carrying the wrong grade for SY27-28.

**Rejected workaround, recorded so it is not retried.** Having SRE enter placeholder students across every planned school and grade, then excluding those ids from the FRESH dashboard, does not work. `stg_finalsite__status_report` applies the `stg_google_sheets__finalsite__exclude_ids` filter itself, upstream of this model, so excluded ids never reach the directory. Moving that filter downstream would change FRESH's calculations on a live dashboard.

**Rejected alternative sources for the next-year branch, so they are not retried.** The Finalsite Enrollment API has no schools or programs endpoint; `assigned_school` is a per-contact track attribute there too. `int_finance__enrollment_targets` stops at AY2024 and has no Paterson rows. `stg_google_sheets__finalsite__goals` holds one recruitment year at a time and is empty for SY27-28 until SRE enters targets. `stg_google_sheets__finalsite__school_scaffold` is disabled and unmaintained since #4451. If FRESH's migration needs planned schools before applicants are routed, the goals sheet is the maintained candidate.
</details>

### DIFF
--- a/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
@@ -65,7 +65,6 @@ with
             and sr.grade_level is not null
     ),
 
-    -- trunk-ignore(sqlfluff/ST03): referenced via dbt_utils.deduplicate below
     school_attributes as (
         select
             powerschool_school_id as ps_schoolid,
@@ -75,24 +74,11 @@ with
             focus_school_id,
 
         from {{ ref("stg_google_sheets__people__locations") }}
-        -- schoolid 0 is the no-school-assigned bucket; Pathways sites are not
-        -- schools students enroll into grade levels at.
         where
             powerschool_school_id is not null
+            and abbreviation is not null
             and powerschool_school_id != 0
             and not is_pathways
-    ),
-
-    -- One row per school. ps_schoolid 179901 carries two locations rows,
-    -- identical but for a null abbreviation on one, so prefer the populated one.
-    schools as (
-        {{
-            dbt_utils.deduplicate(
-                relation="school_attributes",
-                partition_by="ps_schoolid",
-                order_by="(school_short_name is null) asc",
-            )
-        }}
     )
 
 select
@@ -120,4 +106,4 @@ select
     end as school_level_alt,
 
 from directory as d
-left join schools as s on d.ps_schoolid = s.ps_schoolid
+left join school_attributes as s on d.ps_schoolid = s.ps_schoolid
--- a/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
@@ -6,40 +6,14 @@ with
             _dbt_source_project,
             academic_year,
             region,
-            schoolid,
             schoolid as ps_schoolid,
             grade_level,
 
-            'powerschool' as school_source,
+            'sis' as school_source,
 
-        from {{ ref("int_powerschool__student_enrollment_union") }}
+        from {{ ref("int_students__student_enrollment_union") }}
         -- 999999 is the graduated-students placeholder
         where schoolid != 999999 and grade_level is not null
-
-        union all
-
-        -- grain projection, not dup-masking:
-        -- academic_year/region/ps_schoolid/grade_level
-        select distinct
-            _dbt_source_project,
-            academic_year,
-            region,
-            schoolid,
-            ps_schoolid,
-            grade_level,
-
-            'focus' as school_source,
-
-        from {{ ref("int_focus__student_enrollment_roster") }}
-        -- A fixed boundary, not the current year -- do not swap for
-        -- current_academic_year. Focus reaches back to AY2018, but Miami's
-        -- PowerSchool archive owns through AY2025. Null ps_schoolid drops Focus's
-        -- non-instructional Applicants school and would break ps_schoolid's job as
-        -- the cross-SIS join key.
-        where
-            academic_year >= 2026
-            and ps_schoolid is not null
-            and grade_level is not null
 
         union all
 
@@ -49,22 +23,24 @@ with
             sr._dbt_source_project,
             sr.active_school_year_int as academic_year,
             sr.region,
-            x.location_powerschool_school_id as schoolid,
             x.location_powerschool_school_id as ps_schoolid,
             sr.grade_level,
 
             'finalsite' as school_source,
 
         from {{ ref("stg_finalsite__status_report") }} as sr
-        -- location_name is the crosswalk's unique key. Joining on
-        -- location_powerschool_school_id instead fans out, because the crosswalk
-        -- carries one row per alias name.
+        -- location_name is the crosswalk's unique key; the school id fans out
         inner join
             {{ ref("int_people__location_crosswalk") }} as x
             on sr.assigned_school = x.location_name
         where
             sr.active_school_year_int = {{ var("current_academic_year") }} + 1
+            -- a file's grade is correct only for its own cycle year
+            and cast(left(sr._dagster_partition_key, 4) as int)
+            = sr.active_school_year_int
             and x.location_powerschool_school_id is not null
+            and x.location_powerschool_school_id != 0
+            and not x.location_is_pathways
             and sr.grade_level is not null
     ),
 
@@ -88,7 +64,6 @@ select
     d._dbt_source_project,
     d.academic_year,
     d.region,
-    d.schoolid,
     d.ps_schoolid,
     d.grade_level,
     d.school_source,
--- a/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
@@ -1,5 +1,5 @@
 with
-    enrolled as (
+    directory as (
         -- grain projection, not dup-masking:
         -- academic_year/region/ps_schoolid/grade_level
         select distinct
@@ -37,22 +37,25 @@ with
         -- non-instructional Applicants school and would break ps_schoolid's job as
         -- the cross-SIS join key.
         where academic_year >= 2026 and ps_schoolid is not null
-    ),
 
-    -- grain projection, not dup-masking: academic_year/region/ps_schoolid/grade_level
-    incoming as (
+        union all
+
+        -- grain projection, not dup-masking:
+        -- academic_year/region/ps_schoolid/grade_level
         select distinct
             sr._dbt_source_project,
             sr.active_school_year_int as academic_year,
             sr.region,
-            sr.grade_level,
-
             x.location_powerschool_school_id as schoolid,
             x.location_powerschool_school_id as ps_schoolid,
+            sr.grade_level,
 
             'finalsite' as school_source,
 
         from {{ ref("stg_finalsite__status_report") }} as sr
+        -- location_name is the crosswalk's unique key. Joining on
+        -- location_powerschool_school_id instead fans out, because the crosswalk
+        -- carries one row per alias name.
         inner join
             {{ ref("int_people__location_crosswalk") }} as x
             on sr.assigned_school = x.location_name
@@ -60,28 +63,61 @@ with
             sr.active_school_year_int = {{ var("current_academic_year") }} + 1
             and x.location_powerschool_school_id is not null
             and sr.grade_level is not null
+    ),
+
+    -- trunk-ignore(sqlfluff/ST03): referenced via dbt_utils.deduplicate below
+    school_attributes as (
+        select
+            powerschool_school_id as ps_schoolid,
+            location_name as school_name,
+            abbreviation as school_short_name,
+            grade_band as school_level,
+            focus_school_id,
+
+        from {{ ref("stg_google_sheets__people__locations") }}
+        -- schoolid 0 is the no-school-assigned bucket; Pathways sites are not
+        -- schools students enroll into grade levels at.
+        where
+            powerschool_school_id is not null
+            and powerschool_school_id != 0
+            and not is_pathways
+    ),
+
+    -- One row per school. ps_schoolid 179901 carries two locations rows,
+    -- identical but for a null abbreviation on one, so prefer the populated one.
+    schools as (
+        {{
+            dbt_utils.deduplicate(
+                relation="school_attributes",
+                partition_by="ps_schoolid",
+                order_by="(school_short_name is null) asc",
+            )
+        }}
     )
 
 select
-    _dbt_source_project,
-    academic_year,
-    region,
-    schoolid,
-    ps_schoolid,
-    grade_level,
-    school_source,
+    d._dbt_source_project,
+    d.academic_year,
+    d.region,
+    d.schoolid,
+    d.ps_schoolid,
+    d.grade_level,
+    d.school_source,
 
-from enrolled
+    s.school_name,
+    s.school_short_name,
+    s.school_level,
+    s.focus_school_id,
 
-union all
+    -- TODO: figure out a better way to track these
+    case
+        when
+            d.academic_year >= 2025
+            and s.school_short_name = 'Sumner'
+            and d.grade_level >= 5
+        then 'MS'
+        else s.school_level
+    end as school_level_alt,
 
-select
-    _dbt_source_project,
-    academic_year,
-    region,
-    schoolid,
-    ps_schoolid,
-    grade_level,
-    school_source,
-
-from incoming
+from directory as d
+left join schools as s on d.ps_schoolid = s.ps_schoolid
--- a/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
@@ -36,7 +36,10 @@ with
         -- PowerSchool archive owns through AY2025. Null ps_schoolid drops Focus's
         -- non-instructional Applicants school and would break ps_schoolid's job as
         -- the cross-SIS join key.
-        where academic_year >= 2026 and ps_schoolid is not null
+        where
+            academic_year >= 2026
+            and ps_schoolid is not null
+            and grade_level is not null
 
         union all
 
--- a/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_students__school_directory.sql
@@ -1,0 +1,87 @@
+with
+    enrolled as (
+        -- grain projection, not dup-masking:
+        -- academic_year/region/ps_schoolid/grade_level
+        select distinct
+            _dbt_source_project,
+            academic_year,
+            region,
+            schoolid,
+            schoolid as ps_schoolid,
+            grade_level,
+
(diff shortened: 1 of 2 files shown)


GabyRangelB:
Verdicts on the `claude-review` findings. I verified each claim against the code and the data before acting. Fixed in `85ed19c7d`.

## Finding 1 — Focus branch missing `grade_level is not null`

**Accepted and fixed.** The finding is correct, and it also caught a false claim in my PR body.

The Focus branch really did omit the filter. I confirmed the causal chain the reviewer traced: `int_focus__student_enrollment_roster` derives `grade_level` from a `CASE` on `g.short_name` fed by a `LEFT JOIN` to `stg_focus__school_gradelevels` (with `and g.short_name != '30'`), so an unmapped `grade_id` leaves `g.short_name` NULL, `regexp_extract` returns NULL, and the cast yields NULL. `unique_combination_of_columns` does not reject nulls, so nothing would have caught it.

Sized the impact as the reviewer asked: **0 rows** with a null `grade_level`, across all 10,084 rows of `int_focus__student_enrollment_roster` and every academic year — not just AY2026+. So the gap was latent, never live. Row count after the fix is unchanged at 1,010, which confirms the filter drops nothing today.

The PR body claimed the filter appeared on all three branches. It did not. That claim is now true rather than reworded.

## Finding 2 — Focus `schoolid` not non-null by construction

**Accepted and fixed.** Correct: nothing constrained the Focus branch's native `schoolid`, so describing it as non-null by construction was not demonstrable. The reviewer's read of the other two branches is also right — `schoolid != 999999` implicitly drops nulls on PowerSchool, and the Finalsite branch filters `location_powerschool_school_id is not null`.

Ran the suggested query: **0 rows** where `schoolid is null and ps_schoolid is not null`.

Added a `not_null` test on `schoolid` rather than softening the wording. The test is non-vacuous precisely because one branch does not filter the column, which is what makes it legal under the "never add `not_null` to a column that cannot be NULL by construction" rule. Moved `schoolid` into the tested block at the top of `columns:` to match the sort-tested-columns-first convention. The PR body's claim about `not_null` is updated.

## Finding 3 — Miami boundary assumption undocumented in the SQL

**Acknowledged, declining the implied change.** The reviewer explicitly is not asking for one, and I agree the assumption is load-bearing.

Declining to add more SQL prose. The reasoning lives in the model `description` and the PR, which is where this repo puts rationale — `.claude/rules/dbt-yaml.md` is explicit that rationale needing no code context belongs in `description:`, not an inline comment, and that the repo's existing multi-paragraph SQL comments are not a precedent to extend. The Focus branch does carry an inline note about why its boundary is fixed rather than `current_academic_year`, since that one is a trap at the line itself.

Worth adding for the record: the assumption is stronger than "the archive is frozen." Miami rows only reach the warehouse when the team ingests them, and ingestion is gated deliberately, so PowerSchool cannot start emitting AY2026 Miami rows on its own. The grain test is also a live tripwire — if both SIS branches ever produced the same year, it fails loudly rather than silently double-counting.

## Convention checks

No action needed. I spot-checked two of the passes rather than taking them on faith, and both hold: `int_people__location_crosswalk` does expose `location_name` and `location_powerschool_school_id`, and the `DISTINCT` annotations do match what the `DISTINCT` is doing.

One caveat on the review's scope: it ran against `48db2eddc`, the first commit. Later commits added the school attributes, `focus_school_id`, and `school_level_alt`, so its `enrolled` / `incoming` CTE names and line numbers no longer match the branch. The two findings survived that drift and both still applied.

After the fix: 1,010 rows, 1,010 distinct grain keys, 0 nulls in any grain column, 8 of 8 nodes pass.

GabyRangelB:
@claude one last time, please

cbini:
Independent review at `85ed19c7d`, run against prod. Two findings the `claude-review` passes missed, fixed in `25f682a5b` along with cleanup. PR body updated to match.

## Fixed: the Finalsite branch read every cycle file, and the files disagree on grade

`stg_finalsite__status_report` unions one file per recruitment cycle. The same applicant and `active_school_year_int` appear in consecutive files with a different `grade_level`, because each file reports the grade relative to its own cycle. Against the SIS grade for the same student and year: the `2025_26` file is off by one on 9,119 of 9,346 SY26-27 rows, and the `2026_27` file matches on 11,486 of 11,505. With no partition filter, the branch would have unioned the `2026_27` file's SY27-28 rows with the `2027_28` file's once it lands and produced a phantom lower grade per school. The grain test cannot see it, because a different `grade_level` is a different key.

Fix: `cast(left(_dagster_partition_key, 4) as int) = active_school_year_int`. The one SY27-28 row the branch produced before was this case, so the branch is now empty until the `2027_28` file lands. That is the correct state.

## Fixed: the PowerSchool/Focus split duplicated a rule the repo already settled

The `>= 2026` literal took Miami AY2018-2025 from the frozen PowerSchool archive. `int_students__student_enrollment_union` (13 mart consumers) already owns the Miami handoff the other way, and at this grain the two sources are identical sets for Miami through AY2025. Replaced both branches with one read of the union: same 1,010 rows, no boundary literal, and `schoolid` goes away because it always equaled `ps_schoolid`. `school_source` is now `sis` / `finalsite`.

## Cleanup in the same commit

- Finalsite crosswalk join excludes campus (`id 0`) and Pathways aliases, matching the attributes CTE.
- The `school_source` description claimed FRESH ranks on it (nothing consumes the model yet) and the `region` description claimed a match to `stg_google_sheets__reporting__terms` (that column holds both `Miami` and `KIPP Miami`). Both reworded.
- Inline comments that restated the description are cut; the Finalsite branch's limits moved into `description`.

## Not changed

- The Sumner `school_level_alt` copy. Third copy of the rule, already noted, out of scope.
- A curated planned-year source. Checked the Finalsite API (no schools endpoint), `int_finance__enrollment_targets` (stops at AY2024, no Paterson), the SRE goals sheet (one cycle at a time, empty for SY27-28 today), and the retired school scaffold sheet (disabled since #4451). None supplies SY27-28 now. Recorded in the PR body so it is not re-investigated.

Dev build at `25f682a5b`: 1,010 rows, 1,010 distinct grain keys, zero unmatched attributes, 7 of 7 nodes pass, trunk clean.

Refs #4451 (this is the multi-year spine that spec left open) and #3698 (still open; `school_level` here is the sheet's current value, not versioned by year).

cbini (approved):

PR #5127, opened 2026-09-03
fix(claude): make the output scanner redact and name the rule in every denial

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." make `check-output.sh` redact tool results that contain secret material, which it has never actually done, and make every `check-sensitive.sh` denial tell Claude which rule fired and what to do instead.

The output hook returned `permissionDecision: deny`. Claude Code honors that field only on PreToolUse. On PostToolUse it is ignored, so every scanned result reached the model unredacted. The regression suite passed the whole time because it checked the hook's stdout, not what the harness did with it. Proof from this session: a Read of a test fixture line returned the fixture in full before the change and `[redacted: secret material]` after.

The hook now returns `updatedToolOutput`, the documented replace-the-result field, with every string in the result redacted and a note to Claude saying why. When the payload has no `tool_response` to redact, it returns `decision: block` and ends the turn rather than emitting an empty replacement the harness would ignore.

The egress gate in `check-sensitive.sh` gains WebSearch queries and the MCP verbs `share`, `forward`, `schedule`, `launch`, `trigger`, which let Drive share, Outlook forward, Slack schedule, and Dagster launch calls carry a 1Password reference outbound unscanned.

Every one of the 24 denials in `check-sensitive.sh` used to read "Cannot access sensitive path". Each now reads `check-sensitive.sh Rule N: <what matched>. <what to do instead>.` so Claude fixes the call instead of searching CLAUDE.md for the workaround.

## Reviewer Notes

- The redaction keeps the keys `type` and `filePath` unredacted. Built-in tools validate the replacement against their output schema and silently fall back to the original output when it fails.
- `decision: block` on PostToolUse is documented in the hooks reference decision-control table; it ends the turn with the reason as a warning line.
- `deny()` builds its JSON with `jq --arg` so reason text with quotes or `$` cannot break the envelope. The default reason is unchanged for a caller that passes nothing.
- A bare `env` command is caught by Rule 1 (all tools), never Rule 3c. Rule 1's message mentions the bare word so the fix reads the same either way.
- `check-sensitive.sh` and `helpers.sh` diffs are mostly tab reindentation from the commit hook's shfmt. Review with `git diff -w`.
- Docs updated in `.claude/CLAUDE.md` and `tests/CLAUDE.md`: redaction instead of denial, the new egress verbs, rule-named denials, and a note that Auto mode does not replace either hook because the classifier never sees tool results.

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Address valid feedback; dismiss false positives with a brief reply explaining why. (Claude is advisory — use your judgement, but don't ignore it.) Per-finding verdicts posted as a PR comment; finding 1 fixed in the second commit.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes trivially (no dbt changes).
- [ ] **Dagster Cloud** — not triggered (no code location changes).

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Claude reviewed both hooks against the hooks reference (PostToolUse decision control table: `decision`, `reason`, `additionalContext`, `updatedToolOutput`; `permissionDecision` listed for PreToolUse only), probed each hook with scratch harnesses, and drafted the edits in `.claude/scratch/`. The user transcribed or copied the protected hook files; Claude wrote the tests and CLAUDE.md edits. `bash tests/hooks/run_all.sh`: 10 suites, 0 failures on the committed files. New suites and helpers: `expect_redacted` (redaction literal present, secret gone), `expect_block` (drift payload ends the turn), `tests/hooks/test_deny_reasons.sh` (each rule names itself and offers an alternative). Probe results that motivated the egress change: `Google_Drive__share_file`, `outlook_forward_mail`, `slack_schedule_message`, and a `WebSearch` query each passed the old gate carrying a 1Password reference. Known gaps left as-is because only Auto mode covers them: glob-obfuscated dotfile reads (`cat .en?`), `curl` exfiltration from Bash, and Python `getattr`/`chr` construction of the process variable mapping. Reading `check-sensitive.sh` with the Read tool now gets redacted by `check-output.sh` because the file holds the secret regexes; snapshot it into `.claude/scratch/` and patch by script (`.claude/scratch/patch-deny.py` pattern) instead.

</details>

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -113,10 +113,10 @@ evidence in `.claude/scratch/` and reference it. For non-Bash tools only Section
 `content`/`new_string` is content-exempt, so editing docs is unaffected.)
 
 **Non-Bash tool inputs are path-scanned too:** `TodoWrite` / `AskUserQuestion`
-text containing a bare `env` (or other sensitive-path token) trips "Cannot
-access sensitive path". Reword (`environment variable`; avoid cred-suffix tokens
-like `_KIPPMIAMI`). Also fires on `mcp__github__*` PR / issue bodies — prose
-like "staging env" / "dev env" is denied; write "environment".
+text containing a bare `env` (or other sensitive-path token) trips Rule 1 or 3c.
+Reword (`environment variable`; avoid cred-suffix tokens like `_KIPPMIAMI`).
+Also fires on `mcp__github__*` PR / issue bodies — prose like "staging env" /
+"dev env" is denied; write "environment".
 
 **Your own ad-hoc Bash self-blocks on `$UPPER_CASE`:** Rule 7 denies any Bash
 command expanding a non-allowlisted uppercase var — including one you define in
@@ -126,8 +126,11 @@ that same command (`sc=$(...); echo "${SC}"`). Use lowercase names
 **BigQuery MCP** — queries must start with SELECT/SHOW/DESCRIBE/WITH; embedded
 DML/DDL (INSERT, UPDATE, DELETE, CREATE, DROP, etc.) is blocked. The block
 matches the keyword as a substring — including inside a string literal
-(`where type = 'Drop'`), which is denied with the misleading "Cannot access
-sensitive path" message. Reword to avoid the literal (`like 'Dr%'`).
+(`where type = 'Drop'`). Reword to avoid the literal (`like 'Dr%'`).
+
+**Deny messages name the rule.** Every `check-sensitive.sh` denial reads
+`❌ check-sensitive.sh Rule N: <what matched>. <what to do instead>.` Follow the
+instruction in the message before consulting this file; the two agree.
 
 **Output scanning** (PostToolUse) — redacts tool results containing secret
 material (keys, tokens, connection strings, high-entropy strings): every string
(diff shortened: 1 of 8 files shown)


cbini:
Per-finding verdicts on the claude-review comment:

**1. Drift fallback redacts nothing — accepted, fixing.** Reproduced: a payload with the secret under `.result` and no `.tool_response` produced `updatedToolOutput: {}`. For built-in tools the harness rejects a replacement that fails the output schema and shows the original. Fix: `emit_redacted` now checks `has("tool_response")` and otherwise calls `deny_output`, which returns `decision: block` and ends the turn. Nothing can be redacted in the tool's shape when the key is missing, so failing closed is the only safe move. Tests: new `expect_redacted` asserts the redaction literal is present and the secret string is gone; new `expect_block` asserts the drift cases end the turn. Against the current commit the 2 drift tests fail, against the patched hook all 9 suites pass. The hook edit lands in the next push once transcribed (protected file).

**2. `filePath` exempt from redaction — declined.** The path in a Read result is the path Claude itself passed in, so it is already in context. Redacting it makes the replacement fail schema validation and shows the whole original. The risk it names (a secret-shaped file name) is real but the exemption is the smaller exposure.

**3. `decision: block` on PostToolUse — confirmed.** The hooks reference decision-control table lists `PostToolUse` under "Top-level `decision`: `decision: "block"`, `reason`", and the `continue` section says a PostToolUse block ends the turn and shows the reason as a warning line. Verified against https://code.claude.com/docs/en/hooks.md during the change.

**4. Reindent noise — acknowledged, not reverting.** Measured: `check-sensitive.sh` 367 lines changed, 7 without whitespace; `helpers.sh` 277 and 9. shfmt in the commit hook switched the files to tabs and reports them clean now, so reverting to spaces would fail `trunk fmt`. Review with `git diff -w`.

**5. Egress verb list incomplete — acknowledged, no change.** The list grows from probes, as the comment says. `publish`/`invite`/`grant`/`assign`/`sync` match no tool currently exposed to this repo's MCP servers, so adding them now would be untested regex.

**6, 7.** No action.

PR #5132, opened 2026-09-03
fix(zendesk): migrate ZendeskResource to OAuth client_credentials

### LINKED ISSUE #4499: fix(zendesk): migrate zendesk_user_sync auth to OAuth client_credentials
## Context

Zendesk is removing **API tokens** as an authentication method for the Support API and requiring migration to OAuth. Our integration currently authenticates with email + API token basic auth, which is in scope for removal.

Timeline:

- **2026-07-28** — tokens unused for 30+ consecutive days auto-deactivate (ours runs daily, so not at immediate risk)
- **2026-10-27** — no new API tokens can be created
- **2027-04-30** — all remaining API tokens permanently deactivated (hard cutoff)

OAuth access tokens and API tokens run in parallel until the cutoff, so we can migrate and validate without breaking the current sync.

Ref: [Zendesk API token deprecation](https://support.zendesk.com/hc/en-us/articles/10851263566234-Announcing-the-removal-of-API-tokens-as-an-authentication-method-for-API-requests), [OAuth migration guide](https://developer.zendesk.com/documentation/api-basics/authentication/oauth-migration/)

## What is affected

`zendesk_user_sync` (active, scheduled daily at 03:00) pushes `rpt_zendesk__users` to the Zendesk Support API via `users/create_or_update_many` and polls `job_statuses/show_many`. It uses `ZendeskResource`, which sets `email/token` basic auth — the deprecated method.

- Resource: `src/teamster/libraries/zendesk/resources.py`
- Wiring: `src/teamster/core/resources.py` (`ZENDESK_RESOURCE`)
- Asset: `src/teamster/code_locations/kipptaf/zendesk/assets.py`

Out of scope:

- **Airbyte Zendesk Support connection** — auth is configured in Airbyte Cloud, handled separately by an admin.
- **dlt Zendesk pipeline** — currently disabled (`assets = []`); already has a `ZendeskCredentialsOAuth` class for whenever it is revived.

## The fix — client_credentials grant

Use Zendesk's `client_credentials` grant (headless, server-to-server). Mint a short-lived bearer token from the OAuth client secret, no browser step:

```text
POST https://{subdomain}.zendesk.com/oauth/tokens
Content-Type: application/x-www-form-urlencoded

grant_type=client_credentials
client_id={client_id}
client_secret={client_secret}
scope=read users:write

-> { "access_token": "...", "expires_in": 3600, "token_type": "Bearer" }
```

Then authenticate API calls with `Authorization: Bearer {access_token}`.

## Tasks

- [ ] Admin: register a **confidential** OAuth client in Zendesk Admin Center (Apps and integrations -> APIs -> OAuth clients); capture client ID + secret
- [ ] Add Dagster secrets `ZENDESK_CLIENT_ID` and `ZENDESK_CLIENT_SECRET`
- [ ] Update `ZendeskResource` to mint a token via client_credentials in `setup_for_execution` and set the Bearer header (replace `email`/`token` fields with `client_id`/`client_secret`/`scope`)
- [ ] Update `ZENDESK_RESOURCE` wiring in `core/resources.py`
- [ ] Validate on a branch deployment (old token still covers prod)
- [ ] After cutover, retire `ZENDESK_TOKEN` (and `ZENDESK_EMAIL` if unused elsewhere)

## Notes

- **Scopes:** `read users:write` is least-privilege (`users:write` for the POST, `read` for `job_statuses`). Widen to `read write` only if `job_statuses` returns 403.
- **Token lifetime:** ~1 hr; the sync finishes in minutes, so mint-once-per-run is fine. If a run ever exceeds the token lifetime, add re-mint-on-401.

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." switch the nightly Zendesk user sync from API-token auth to OAuth. Zendesk is retiring API tokens: no new tokens after 2026-10-27, and every token stops working on 2027-04-30.

`ZendeskResource` now mints a short-lived bearer token at the start of each run using the `client_credentials` grant, then sends it on every request. The old `email` and `token` fields are gone; `client_id` and `client_secret` replace them, wired from a new 1Password item.

Closes #4499

## Reviewer Notes

- The OAuth token acts as the Zendesk user who owns the OAuth client (currently Charlie). Zendesk audit entries for the sync will show that user instead of the API token's owner.
- `ZENDESK_EMAIL` and `ZENDESK_TOKEN` stay in `dagster-cloud.yaml` for now. The disabled dlt Zendesk asset still reads them at import time, so removing them would break the `kipptaf` code location load. Retiring them is a follow-up.
- The new `OnePasswordItem` in `.k8s/1password/items.yaml` is already applied to the cluster, and the synced secret keys (`username`, `credential`) were verified with `kubectl`. The `kipptaf` branch deployment loaded with the new variables.
- The token mint is wrapped in the repo's standard tenacity retry, same as the Coupa and ADP resources.

### General

### Dagster

      code location

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Verified before opening: a throwaway pytest built `ZendeskResource` with the `Zendesk OAuth - Dagster` credentials, minted a token (scope `read users:write`, `expires_in` 1800), and got 200 from `users/me` (role `admin`) and `job_statuses`. An over-scope mint with `scope=write` returned 400, so the client's scope restriction holds. Two OAuth clients exist on purpose: `airbyte` (scope `read`) for the warehouse sync and `dagster` (scope `read users:write`) for this asset, so rotation and audit attribution stay separate.

1Password API Credential items sync to k8s with field keys `username` (client id) and `credential` (client secret), matching the `secretKeyRef` keys here. Verified on the cluster after applying the item; `op-airbyte-api` and `op-coupa-api` use the same mapping.

Token lifetime is 30 minutes, not the hour the issue assumed. The sync runs in minutes, so the resource mints once per run and does not re-mint on 401.

</details>


cbini:
Per-finding verdicts on the Claude review (fixes in a5c8254d7e):

1. **Retry on token mint — accepted.** The mint now lives in `_fetch_access_token`, wrapped with the same tenacity decorator as `CoupaResource._fetch_access_token` (`stop_after_attempt(5)`, `wait_exponential_jitter(initial=10, max=60)`, retry on `RequestsConnectionError`, `Timeout`, `HTTPError`).
2. **Log the response body on a failed mint — accepted.** `_fetch_access_token` logs `response.text` via `self._log.error` before re-raising, matching `_request`.
3. **Inline the scope literal — accepted.** The `scope` field is gone; `"read users:write"` is inline in the token payload. One call site, no override.
4. **Switch to `requests_oauthlib` — declined.** Zendesk's `/oauth/tokens` takes a JSON body and oauthlib sends form-encoded, so the swap needs a body override and ends up as more code than the 10-line manual call, which is verified live against the real client. Revisit if a third client-credentials resource makes the shared pattern worth it.
5. **1Password key mapping — confirmed.** Charlie also verified the synced secret keys directly with `kubectl get secret`, so `username`/`credential` are checked, not inferred.
6. **No other `email`/`token` callers — confirmed.**
7. **Local validation — done.** `uv run pytest tests/test_dagster_definitions.py -k kipptaf` passes on this branch. The standalone `dagster definitions validate` needs the pytest fixture for secrets and a built dbt manifest, which is why CI cannot run it.

Also: the first `kippnewark` branch deploy failed on a gRPC `UNAVAILABLE` while the agent synced, then loaded on its automatic retry (`get_location_load_history` shows `LOADED` at the same commit). Transient, and the push above re-runs the deploy checks.

PR #5138, opened 2026-09-03
feat(dbt): give the Cumulative GPA Monitor one data source

### LINKED ISSUE #5097: feat(kipptaf): carry the GPA goal rate onto rpt_tableau__gpa_cumulative_year
## What and why

The Cumulative GPA Monitor reads 2 data sources. Student rows come from `rpt_tableau__gpa_cumulative_year`. Goal rates come from `rpt_tableau__gpa_goals`. Tableau cannot drive both with one filter, so the dashboard works around the split in 4 places. The dashboard ships with those workarounds. This issue removes them later.

Grade level is the worst of the 4. The student model calls it `grade_level`, an integer. The goals model calls it `grade_band`, a string. A quick filter on either one moves half the dashboard and leaves the rest on a different cohort, with nothing on screen to say so. The build uses a parameter and 2 matching calculated fields instead of a filter.

School has no workaround at all. `rpt_tableau__gpa_goals` carries `schoolid` but no `school`, so the workbook hard-codes 3 school ids in a CASE statement. The same CASE already sits in the workbook, commented `// MUST BE DEPRECATED`.

Network-level goal rows carry a null `region`. A Tableau user filter on `region` drops every one of them, so both goal tiles go blank for every non-admin user. Tableau Desktop cannot show that failure, because in Desktop you are always yourself.

7 of the 50 calculated fields in the build exist only to bridge the 2 sources.

## Proposed approach

Make the goal that applies to a student available on the student's own row. Two ways to do it, and the data team should pick:

1. Add the goal proportion and the goal rate for the row's own grade to `rpt_tableau__gpa_cumulative_year`.
2. Build a new model at student-by-academic-year grain carrying both the student's cumulative GPA and the goal that applies to them.

Either way the dashboard reads 1 data source. Grade becomes a real filter. The headline band cannot contradict itself.

The 3 goal rungs are not a fan-out risk. A student row belongs to exactly 1 school in 1 region, so the org, region and school goals for that student all resolve to a single row.

## For Claude

Context, not instructions.

### Models involved

- `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gpa_cumulative_year.sql` — student by academic year, Newark and Camden, grades K-12
- `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__gpa_goals.sql` — thin passthrough over `int_gpa__goal_aggregations`
- `src/dbt/kipptaf/models/gpa/intermediate/int_gpa__goal_aggregations.sql` — builds the org, region and school rungs as separate row sets

### Where this came from

The Tableau build runbook for the Cumulative GPA Monitor. The dashboard is being built against the 2-source design now, so nothing here blocks it. Predecessor issues: #4619 (the dashboard), #4581 and #4598 (the goals model).

The 7 bridging calculated fields, all on the goals data source: `Year filter`, `Grade`, `Grade sort`, `Grade filter`, `Network or region row`, `School row`, `School (goals)`. The 5 that would survive a merge are `Goal rate`, `Actual rate`, `Gap to goal (pts)`, `Goal status` and `Students still needed`.

### Grain facts, measured 2026-09-01

- AY2026 `cumulative_gpa_unweighted` goals: 4 org rows, 8 region rows, 12 school rows. Mixing the rungs triples every count. Grade 11 reads 1260 students measured instead of 420, and the rate still looks plausible.
- `goal_proportion` varies by grade: 0.69 for grade 9, 0.64 for grade 10, 0.60 for grade 11, 0.56 for grade 12. There is no single network goal.
- `grade_band` is built as `grade_low` when the 2 bounds match and `grade_low-grade_high` when they differ. Every goal today is a single grade. A banded goal would produce `9-12`, which no grade selector can match.
- School ids are 73253 (NCA), 732511 (NLH), 179904 (KHS).

### Two things any merged model has to decide

1. Miami Tech. Miami opened a high school in AY2026 with about 114 grade 9 students. `rpt_tableau__gpa_cumulative_year` excludes them by region filter. `int_gpa__goal_aggregations` does not — `joined_org` matches on grade alone, so grade 9 network `n_students_in_grain` reads 606 against 509 measured. They have no posted or projected GPA yet, so `metric_rate` is unaffected today. That stops being true when their grades project. Note that #4619 recorded Paterson and Miami as having zero HS students in any year on record, and treated that as structural. It is no longer accurate.
2. Which metric. The goals table holds `cumulative_gpa_unweighted` and `y1_gpa_weighted`. The second covers grades 9 to 11 only and exists at org and school level but not region. Carrying both onto a student row needs a column per metric, not one goal column.

### Acceptance criteria

- A single Tableau data source answers 3 questions: what is this student's cumulative GPA, what is the goal for their grade, and what is the rate against it at their school, region and the network.
- Row count of `rpt_tableau__gpa_cumulative_year` is unchanged if option 1 is taken. The model holds 22 years of history at student-year grain, so a join that fans out is the main risk.
- `AVG` of the goal column across a grade returns that grade's goal, not an average of 3 rungs.

### Related, smaller, not blocking

- `cumulative_y1_gpa_unweighted_as_of_today` reads as including today's work but means from Y1 grades already posted. A grade 9 student has no completed year at their high school, so 489 of 511 are null. The name drives people to the wrong column.
- `rpt_tableau__gpa_cumulative_year` carries `school_leader_tableau_username` but no regional equivalent, so region-level row access still needs a hand-maintained Tableau group filter.
- `progress_to_goal` is capped at 1.0, so it cannot distinguish just meeting the goal from beating it comfortably. Grade 9 is above its goal and reads exactly 1.0.

anthonygwalters:
## Summary & Motivation

When merged, this pull request will let the Cumulative GPA Monitor dashboard read
one Tableau data source instead of two.

Today the dashboard reads student rows from `rpt_tableau__gpa_cumulative_year`
and goal rates from `rpt_tableau__gpa_goals`. A Tableau filter belongs to one
data source, so nothing on the dashboard can filter both at once. Pick a grade
and half the screen moves; the other half keeps describing a different group of
students, with nothing on screen to say so. The build works around this in four
places, and seven of its fifty calculated fields exist only to bridge the two
sources.

This adds two models:

- `int_gpa__student_goal_definitions` puts each student's GPA goal on their own
  row, at the network, region and school level. It carries the goal
  *definition* — the GPA a student must reach, and the share of students
  expected to reach it — never a rate someone else already calculated.
- `rpt_tableau__gpa_goal_progress` passes every column of
  `rpt_tableau__gpa_cumulative_year` through unchanged and adds four goal
  columns. This is what the dashboard will read.

**Nothing published changes.** `rpt_tableau__gpa_cumulative_year`,
`rpt_tableau__gpa_goals` and `rpt_tableau__student_course_grades` are untouched,
so Academic Health Home and Academic Health Schools cannot be affected by this
work.

Closes #5097

## Reviewer Notes

**Why a wrapper instead of adding the columns to the extract.** Three facts
together made changing the extract unsafe for now. It is a view, so new logic is
live the moment it deploys and there is no earlier copy to fall back on. Its
uniqueness test only warns, because that is the project default. And the
published workbook tells Tableau that side of its join cannot duplicate, so
Tableau skips the check it would otherwise do. A bad join key would therefore
swap a view in place, warn instead of failing, and quietly multiply student
counts on two published dashboards — and a Tableau extract refresh in that
window would keep the bad rows after a revert. The wrapper avoids all of it. The
spec has the full reasoning and the steps to fold it back in later.

**A report reading another report.** `rpt_tableau__gpa_goal_progress` reads
`rpt_tableau__gpa_cumulative_year`. Ten models in this project already do this,
three of them inside `extracts/tableau/`. It is temporary by design.

**The passthrough list is written out by hand.** The repo forbids `select *` in
a final `rpt_` select, so a column added to the extract will not arrive here on
its own. That is now stated in the model's description and in a comment at the
top of the SQL, because the contract will not catch it — the column simply never
reaches the dashboard.

**One test projects `student_number`.**
`rpt_tableau__gpa_goal_progress__goals_land_on_students` selects it so a failure
names the affected students. Test failures are stored, so that would write
student numbers into the audit dataset. The same column already exists on the
model being tested and the audit dataset carries the same access controls, so
this is not new exposure — but it is wider than the check needs. Say the word and
it can project a grade-level count instead.

**Verification.** Row counts match the extract exactly in every academic year.
All 52 passthrough column names match the source exactly. The share of students
at or above the goal, computed from these student rows, reproduces the published
rate at every grade to within 0.0005 — the check that proves the dashboard will
agree with the goals model.

### General

### dbt

- [x] Include (or update) an exposure for all models consumed by a dashboard
- [x] **Breaking change?** No. No column on any contracted model is renamed or
      removed. The two new models are additive, and the three models read by
      published dashboards are untouched — `git diff origin/main...HEAD` over
      those paths is empty.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Built with `superpowers:subagent-driven-development`
against `docs/superpowers/plans/2026-09-03-student-goal-definitions.md`, which
argues from `docs/superpowers/specs/2026-09-01-student-goal-definitions-design.md`.
Both are on this branch.

**Grain and why the join cannot fan out.** `int_gpa__student_goal_definitions`
is `(academic_year, student_number, metric)`, with a
`unique_combination_of_columns` test at `severity: error` — the project default
is `warn`, so the override is required, not stylistic. Filtered to
`metric = 'cumulative_gpa_unweighted'` it has zero duplicate
`(academic_year, student_number)` pairs, asserted directly against prod rather
than inferred from a row count: 7,718 rows total, 4,273 for that metric, 0
violations at either grain. That is what makes the wrapper's left join
one-to-at-most-one.

**Population scoping.** The inner join to the network-rung goal scopes the
population, so a grade with no goal produces no row rather than a row of nulls.
That is how K-8 drops out without a `school_level` filter and without naming a
region. The region and school rungs are left joins, which is why
`y1_gpa_weighted` lands a null region proportion — it has no region rung in
either 2025 or 2026.

**Overlapping grade bands are meant to fail the build.** Nothing in
`int_google_sheets__gpa_goals` prevents a `9-12` band alongside an `11-11` band.
The between-join would then match two rows for one student and the uniqueness
test fails. A most-specific-band-wins tie-break would absorb a Google Sheet data
entry error and produce a number nobody could trace.

**Two singular tests guard the silent-drop paths.**
`int_gpa__student_goal_definitions__org_rung_covers_all_rungs` asserts every
region- and school-rung goal sits inside a network-rung band at the same year
and metric — containment, not equality, so a school serving a narrower band than
the network does not read as an omission. `rpt_tableau__gpa_goal_progress__goals_land_on_students`
asserts the mirror: in any year the goals sheet covers, no grade 9-12 row of the
wrapper has a null network proportion. Without it, a drift in
`int_extracts__student_enrollments`'s `rn_year` semantics or a `student_number`
key would silently blank every goal on the dashboard with no test failing.

**Verification numbers, AY2026 org rung.** Wrapper-computed rate against the
published `metric_rate`: grade 9 350/510 = 0.686 vs 0.686; grade 10 283/493 =
0.574 vs 0.574; grade 11 202/419 = 0.482 vs 0.482; grade 12 188/410 = 0.459 vs
0.459. Largest gap 0.0005 against a 0.005 tolerance. Row parity with the extract
returns zero differing rows across every academic year, not only the current one.

**A trap for anyone verifying this locally.** `--defer` does not redirect a
`ref` when a relation already exists in your dev schema, so a stale personal
clone silently shadows prod. That produced a wrong verification result during
this build — eight rows where four were expected — traced to clones five weeks
old. Every dbt command for this work carries `--favor-state`, and the plan's
Global Constraints now say so.

**Known follow-ups, deliberately not in this PR.**

- `not_null` on `stg_google_sheets__gpa_goals.goal`. `goal_proportion_org` has
  no `not_null` here, justified as non-nullable by construction — but an inner
  join guarantees a matching row, not a non-null value, so a blank cell in the
  sheet would pass through. The right fix is on the staging model, where it
  closes the hole for every consumer and scans a table rather than re-expanding
  a view chain. Related and pre-existing: that column's `expression_is_true`
  does not catch nulls, and `org_level`, `metric` and `direction` carry
  `accepted_values` with no paired `not_null`.
- `direction` / `higher_is_better` are not carried onto the wrapper, so the
  dashboard hardcodes `>=`. Stated in the `gpa_goal_threshold` description.
- The exposure now names only the wrapper, while the workbook still reads two
  sources until it is rebuilt. Low risk: the exposure has no Dagster id, its URL
  is `TBD`, the dashboard is unpublished, and `rpt_tableau__gpa_goals` keeps its
  other exposure.
- Adding `school` to `rpt_tableau__gpa_goals` and `gpa_gap_to_3_0` to the
  extract. Both are out of scope here and belong together in their own change.


anthonygwalters:
@claude review

anthonygwalters:
Verdicts on all three suggestions, with what I checked.

### 1. Unit test for the join logic — accepted, added in `af9b9ae3c`

Verified the convention claim before acting on it. Both other models in
`models/gpa/intermediate/` do carry a `unit_tests:` block, so this model was the
only one of three without. Worth noting for anyone reading later: it is a local
convention, not a repo-wide one — only 17 of 971 properties files in `kipptaf`
have unit tests.

The suggestion turned out to be worth more than "matches the neighbours." Four
cases now covered: all three rungs resolving, rung-level nulls (the live Miami
case, where a student carries a network goal and nothing below it), population
scoping via the network inner join, and **duplicate rows from overlapping
network bands**.

That last one is the reason this was worth doing now. The spec relies on an
overlapping band in the Ops sheet failing the uniqueness test at build time
rather than being silently absorbed by a tie-break — but production has no
overlapping bands, so nothing exercised that path at all. The unit test is the
only thing that can prove it.

The test also immediately earned its keep: the first fixture draft put the
overlap student at grade 11, which also fell inside the 9-12 band and
contaminated the grade-10 expectations. The test caught it rather than a human
noticing later.

### 2. School-rung join drops the region predicate — accepted, but the reasoning here was wrong

Comment added in `af9b9ae3c`, then corrected in `e2117898c`.

The conclusion is right — this is deliberate and safe — but not for the reason
given, and the stated reason would have been enshrined in a code comment.

**`schoolid` is not globally unique across regions.** I checked: of 29 distinct
schoolids in `int_extracts__student_enrollments`, one appears in more than one
region — `999999`, in both Newark and Camden.

It is safe for a different reason. `999999` is the placeholder for graduated
students, and every one of its 21,563 rows carries `grade_level = 99`. Grade 99
never falls between a goal band's `grade_low` and `grade_high`, so those rows are
dropped by the network-rung inner join before the school-rung join is reached.
That is a stronger guarantee than uniqueness would have been, and it is what the
comment now says.

Flagging the correction explicitly because the original claim is the kind that
reads as obviously true and would not get re-checked.

### 3. Inline `TODO` for the `direction` / `higher_is_better` gap — agreed, but not done yet

Confirmed the risk is live rather than theoretical:
`stg_google_sheets__gpa_goals.direction` carries
`accepted_values: [">=", "<="]`, so a `<=` GPA goal is a schema-legal sheet entry
today, and the dashboard would keep computing `>=` against it silently.

The repo convention is `TODO(#issue)` with a real issue number, so this needs a
tracking issue first, and opening one is the author's call rather than something
to do unilaterally. Holding until @anthonygwalters says whether to open it — at
which point the inline pointer goes at the derivation site, not just in the PR
body.

---

Also worth recording here: CI is green on all seven checks including dbt Cloud,
which the review noted it could not reach from its session.

Issue #4835, opened 2026-08-11
fix(data): kippcamden duplicate AY2026 enrollment stints sharing an entrydate (2 students)


cbini:
## Problem

Two kippcamden students have **two active enrollment stints for AY2026 that share an `entrydate` of 2026-07-01**. One is a short 2-3 day stint, the other runs the full year to 2027-06-30. Both carry `enroll_status = 0`.

A student cannot occupy two stints simultaneously, so one row in each pair is a data-entry artifact that should be removed or corrected in PowerSchool.

Student identifiers are deliberately omitted here (PII). The affected rows are reproducible with the query below, which returns exactly the two groups.

```sql
select
  student_number, _dbt_source_project, academic_year, entrydate,
  count(*) as n_rows,
  string_agg(cast(exitdate as string), ' | ' order by exitdate) as exitdates
from `teamster-332318.kipptaf_students.int_students__student_enrollment_union`
where _dbt_source_project = 'kippcamden' and academic_year = 2026
group by student_number, _dbt_source_project, academic_year, entrydate
having count(*) > 1
```

## Why it surfaced now

This is the same class as #3919 (kippmiami overlapping `(studentid, yearid)` enrollments), which was resolved as a source-data cleanup.

It went unnoticed because dbt Cloud CI builds `state:modified+` only, and nothing had modified `fct_student_attendance_streaks` or `int_students__student_enrollment_union` since the rows appeared. #4822 touches every mart, which pulled both into the CI selection and failed the fact's PK uniqueness test with 5 results. Prod carries the same 5 duplicates today.

## Downstream effect

`fct_student_attendance_streaks` joins the enrollment union by half-open date range purely to derive `student_enrollment_key`. When two stints share an `entrydate` and both cover a streak's start date, the join matches twice and emits a duplicate fact row.

Both matches hash to the **same** `student_enrollment_key` — `entrydate` is the only enrollment input to that key — so the duplicate rows are byte-identical across every column. 5 duplicated streak rows in a 6.2M-row fact.

## Interim mitigation

#4822 collapses the join fan-out in `fct_student_attendance_streaks` with a `TODO` pointing here. Because both matches produce an identical key, that collapse is information-preserving rather than duplicate-masking, and a genuine overlap (two stints with *different* entrydates covering one streak) would still fail the PK test as it should.

Remove the mitigation once PowerSchool is cleaned.

## Acceptance

- [ ] The reproduce query above returns zero rows
[...]

Issue #3902, opened 2026-05-12
fix(dbt): residual student-email-no-enrollment orphans in fct_survey_submissions student leg


cbini:
Refs #3794 / #3894 (parent PR). Surfaced as the irreducible-by-date-range portion of the student-leg residual after #3894's enrollment-window fix.

## Problem

After #3894 changed the `fct_survey_submissions` student leg to attribute submissions via half-open `[entrydate, exitdate)` membership instead of `(academic_year, enroll_status = 0)`, ~134 distinct student SCD survey submissions still appear in `fct_survey_responses` without a matching row in `fct_survey_submissions`. Of those:

- **106** fail the upstream terms join (`date_submitted` falls outside any AY2025 SCD term window) → tracked at #3895.
- **28** make it through the terms join but find no enrollment whose `[entrydate, exitdate)` window covers `date_submitted`.

This issue covers the 28.

## Sub-classes (prod, 2026-05-12)

| Subclass | Count | Description |
| --- | --: | --- |
| Email never in `int_extracts__student_enrollments` | 19 | `respondent_email` does not appear in any enrollment row. Plausible causes: typos, external/personal addresses, alumni submitting by mistake, sibling using a different student's account. |
| Email in enrollments but no window covers `date_submitted` | 9 | `respondent_email` matches one or more enrollment rows, but none has `entrydate <= date_submitted < exitdate`. Plausible causes: submission in an enrollment-history gap (between exit and re-entry), or `entrydate` / `exitdate` data quality issues at the source. |

## Acceptance

Two options, in rough order of preference:

1. **Investigate and remediate at source**: characterize the 19 unknown emails (are they really unknown students, or upstream identity-resolution gaps?) and the 9 enrollment-gap cases (are the gaps real or `entrydate`/`exitdate` data issues?). Where possible, fix at the source.
2. **Accept and document**: flip the student-leg INNER JOIN to LEFT JOIN with nullable `student_enrollment_key`, so unresolvable students still produce a row in `fct_survey_submissions` (null FK). Wraps `student_enrollment_key` with the nullable-FK `if(... is not null, generate_surrogate_key(...), cast(null as string))` pattern. Closes the orphan count to 0 for this subclass without recovering the identity.

The parent PR opted for the strict INNER JOIN per the original spec; this issue revisits the decision.

## Out of scope

- `#3895` (AY2025 SCD term-window config gap, 106 orphans)
[...]

PR #5146, opened 2026-09-04
fix(powerschool): give student_enrollments dedupe a real tiebreaker

### LINKED ISSUE #5141: fix(powerschool): base_powerschool__student_enrollments dedupe tiebreaker is a no-op and arbitrarily drops live enrollments
## Problem

`base_powerschool__student_enrollments` dedupes enrollment stints with an `order_by` that cannot break any tie. When a student has two stints sharing an `entrydate`, which row survives is arbitrary. When the wrong one survives, the student's live full-year enrollment is silently replaced by a short transfer-out stub, and the student drops out of every downstream model that filters `rn_year = 1`.

`src/dbt/powerschool/models/sis/base/base_powerschool__student_enrollments.sql` lines 2-10:

```sql
dbt_utils.deduplicate(
    relation=ref("int_powerschool__student_enrollment_union"),
    partition_by="student_number, academic_year, entrydate",
    order_by="student_number desc",
)
```

`student_number` is one of the `partition_by` keys, so it is constant within every group. The `order_by` is a no-op.

On BigQuery, `dbt_utils` dispatches to `bigquery__deduplicate`, which compiles to:

```sql
array_agg(original order by student_number desc limit 1)[offset(0)]
```

With a constant sort key this returns an arbitrary row from the group. Nothing pins the result, so it can change between builds with no code or data change.

## Why the wrong row is harmful

`rn_year` is assigned upstream in `int_powerschool__student_enrollment_union`:

```sql
row_number() over (
    partition by studentid, yearid order by yearid desc, exitdate desc
) as rn_year
```

So `rn_year = 1` means the latest-exiting stint of the year, and every downstream `rn_year = 1` filter assumes that row is the one that survived dedupe. The dedupe does not honor that ordering, so the assumption can silently fail. `base_powerschool__student_enrollments` passes `rn_year` straight through, as does `int_extracts__student_enrollments`.

## Observed impact

A Newark kindergarten student has two AY2026 stints, both with `entrydate` 2026-07-01:

| rn_year | exitdate | exitcode | meaning |
| --- | --- | --- | --- |
| 1 | 2027-06-30 | null | live full-year enrollment |
| 2 | 2026-07-02 | T8 | one-day transfer-out stub |

Dedupe kept the stub. The student now carries `rn_year = 2`, so `rpt_clever__students` excludes them via its `rn_year = 1` filter. `rpt_clever__enrollments` reads `stg_powerschool__students` instead, which is student-grain with no `rn_year` filter and still shows `enroll_status = 0`, so it continues to ship 5 section rows. Clever receives enrollments for a student it has no student record for.

Current production state, all in AY2026:

| Region | Surviving rn_year | Students |
| --- | --- | --- |
| kippnewark | 1 (correct) | 1 |
| kippnewark | 2 (wrong) | 2 |
| kippcamden | 1 (correct) | 1 |
| kipppaterson | 1 (correct) | 1 |

Two of the five collisions kept the wrong row. One of those two surfaces as a Clever orphan; the other is `enroll_status = 2` at the student level and so falls out of both feeds.

`rpt_clever__enrollments__student_id_resolves_to_students_feed` is failing in production against this today.

Student identifiers are omitted here. The reproduce query below returns them.

## This likely explains a previously unexplained observation

The "Not in scope" section of #4745 records a production orphan at one school that "resolved on its own within the hour" with no intervention. An arbitrary `array_agg` pick flipping between builds produces exactly that signature: an orphan that appears and disappears with no code or data change. Worth confirming as the same root cause when this is fixed.

## Reproduce

```sql
with collisions as (
  select student_number, _dbt_source_project, academic_year, entrydate
  from `teamster-332318.kipptaf_powerschool.int_powerschool__student_enrollment_union`
  group by student_number, _dbt_source_project, academic_year, entrydate
  having count(*) > 1
)
select
  b._dbt_source_project,
  b.academic_year,
  b.rn_year as surviving_rn_year,
  count(*) as students
from `teamster-332318.kipptaf_powerschool.base_powerschool__student_enrollments` b
inner join collisions c
  using (student_number, _dbt_source_project, academic_year, entrydate)
group by 1, 2, 3
order by 2 desc, 1, 3
```

Any row with `surviving_rn_year` not equal to 1 is a student whose live enrollment was discarded.

## Suggested fix

Give the dedupe a real tiebreaker aligned with the `rn_year` ordering it feeds:

```sql
order_by="exitdate desc"
```

This makes the surviving row deterministic and guarantees it is the same row `rn_year = 1` designates.

Blast radius needs checking before merge. `base_powerschool__student_enrollments` is upstream of a large share of the project, so this changes row selection anywhere else a same-`entrydate` collision exists. Today that is 5 students in AY2026 and none in prior years, so the change should be observable and small, but confirm rather than assume. See also #3999, which proposes migrating consumers off this model entirely.

## Relationship to other issues

- #4835 is the source-data side of the same condition: duplicate AY2026 stints sharing an `entrydate` in kippcamden, tracked with Ops for cleanup in PowerSchool. That issue removes the trigger. This issue makes the model behave predictably when the trigger is present. Both are worth doing, and neither closes the other.
- #4745 covers CI flakiness of the affected test caused by `state:modified+` splitting the two Clever feeds. Distinct cause, same test.

## Acceptance

- [ ] Dedupe uses a tiebreaker that is deterministic within the partition
- [ ] The reproduce query returns no row with `surviving_rn_year` not equal to 1
- [ ] `rpt_clever__enrollments__student_id_resolves_to_students_feed` passes in production
- [ ] Row counts on `base_powerschool__student_enrollments` and its immediate downstream models are compared before and after, and any change is attributable to the 5 known collisions

Refs #4835, #4745, #3999

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make the PowerSchool enrollment models keep the right row when a student has two enrollment stints with the same `entrydate`.

The dedupe in `base_powerschool__student_enrollments` ordered by `student_number`, which is also a partition key, so the sort was a no-op and BigQuery picked an arbitrary row. When it picked the wrong one, the student's live enrollment was replaced by a short transfer-out stub, and every downstream `rn_year = 1` filter dropped the student.

The dedupe now lives one step upstream in `int_powerschool__student_enrollment_union`, ordered by `rn_year asc`, so every consumer of the union gets one row per `(student_number, academic_year, entrydate)` and the surviving row is always the one `rn_year = 1` filters expect. `base_` reads the union directly. The union gains the uniqueness test that grain now earns.

Closes #5141

## Reviewer Notes

- The issue suggested `order_by="exitdate desc"`. I used `rn_year asc` instead. One of the 3 live collisions (Paterson) has both stints sharing `entrydate` and `exitdate`, so `exitdate` alone would still be arbitrary there. `rn_year` is a `row_number`, so it is unique within the group.
- The 4 `rn_*` windows in the union now end in `reenrollments_dcid asc`. Without it, two stints sharing `yearid` and `exitdate` (the Paterson pair) left `rn_year` itself unstable across builds. The students-table row has a null `reenrollments_dcid` and nulls sort first under `asc`, so the current record wins the tie.
- Moving the dedupe upstream changes the union's row count by the number of collisions (3 network-wide today, all AY2026). `base_` row counts do not change.
- kipptaf's `int_students__student_enrollment_union` reads the district union tables via `source()` and carries its own identical dedupe. It stays for now: dropping it in this PR would run kipptaf CI against stale `zz_stg_*` copies that still hold the duplicates. Removing it is a follow-up after the districts rebuild in prod.

### General

### dbt

- [x] Include (or update) an exposure for all models consumed by a dashboard — no new consumers
- [x] **Breaking change?** No. No columns added, removed, or renamed.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Diagnostic re-run on 2026-09-04 before the fix. The issue's reproduce query returned 3 collision groups, all AY2026: kippcamden (surviving `rn_year` 1, correct), kippnewark (2, wrong), kipppaterson (2, wrong). The issue reported Paterson correct and Newark 2 wrong the same morning, which confirms the pick flips between builds with no code or data change.

Blast radius query (min `rn_year` per collision group vs the current surviving row): 3 groups, 2 flip to `rn_year = 1`, 1 already correct. Paterson's group has `count(distinct exitdate) = 1`, which is why `exitdate desc` was rejected as the tiebreaker.

Related coverage already in place: `int_powerschool__ps_enrollment_all` has a warn-severity `(studentid, entrydate)` uniqueness test that fires on 63 groups today (61 Newark, 1 Camden, 1 Paterson). 60 of the Newark groups are a real stint plus a zero-length stub (`exitdate <= entrydate`) that the union filters out, so the union only sees 3. That test is the Ops cleanup list; #4835 is the tracker.

Paterson pair differs in `entrycode`, `next_school`, `sched_nextyeargrade`, and `reenrollments_dcid`. After the window tiebreaker, a kipppaterson dev build of the union passed and the surviving row is `rn_year = 1`, the students-table row, with `next_school` populated.

Local validation, kippnewark dev target with `--defer --favor-state`: `dbt build` of `int_powerschool__student_enrollment_union` and `base_powerschool__student_enrollments` passed (PASS=3 WARN=1). The warn is the pre-existing `no_shared_stint_boundary` test (60 rows, #3902). The new uniqueness test passed. Dev union rows 98,282 vs prod 98,283 (the 1 collision); dev `base_` rows 98,282 vs prod 98,282.

dbt Cloud CI is trivial for this PR: the change is confined to the powerschool package and no kipptaf column set changes, so `state:modified+` selects nothing in kipptaf.

Acceptance items still open after merge: confirm `rpt_clever__enrollments__student_id_resolves_to_students_feed` passes in production, and re-run the issue's reproduce query to confirm no `surviving_rn_year` other than 1. Follow-up: remove the dedupe from kipptaf `int_students__student_enrollment_union` once the 3 districts rebuild.

</details>

### DIFF
--- a/src/dbt/powerschool/models/sis/base/base_powerschool__student_enrollments.sql
+++ b/src/dbt/powerschool/models/sis/base/base_powerschool__student_enrollments.sql
@@ -1,5 +1,7 @@
 with
     deduplicate as (
+        -- TODO(#4835): remove once Ops cleans up the PowerSchool stints that put
+        -- two enrollments on one entrydate.
         {{
             dbt_utils.deduplicate(
                 relation=ref("int_powerschool__student_enrollment_union"),
--- a/src/dbt/powerschool/models/sis/base/base_powerschool__student_enrollments.sql
+++ b/src/dbt/powerschool/models/sis/base/base_powerschool__student_enrollments.sql
@@ -1,16 +1,4 @@
 with
-    deduplicate as (
-        -- TODO(#4835): remove once Ops cleans up the PowerSchool stints that put
-        -- two enrollments on one entrydate.
-        {{
-            dbt_utils.deduplicate(
-                relation=ref("int_powerschool__student_enrollment_union"),
-                partition_by="student_number, academic_year, entrydate",
-                order_by="rn_year asc",
-            )
-        }}
-    ),
-
     enr_bools as (
         select
             enr.*,
@@ -26,7 +14,7 @@ with
                 then true
                 else false
             end as is_enrolled_recent,
-        from deduplicate as enr
+        from {{ ref("int_powerschool__student_enrollment_union") }} as enr
         left join
             {{ ref("int_powerschool__calendar_rollup") }} as cr
             on enr.schoolid = cr.schoolid
--- a/src/dbt/powerschool/models/sis/base/base_powerschool__student_enrollments.sql
+++ b/src/dbt/powerschool/models/sis/base/base_powerschool__student_enrollments.sql
@@ -4,7 +4,7 @@ with
             dbt_utils.deduplicate(
                 relation=ref("int_powerschool__student_enrollment_union"),
                 partition_by="student_number, academic_year, entrydate",
-                order_by="student_number desc",
+                order_by="rn_year asc",
             )
         }}
     ),
--- a/src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_enrollment_union.sql
+++ b/src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_enrollment_union.sql
@@ -279,8 +279,8 @@ with
         from window_calcs_2
     )
 
-    -- TODO(#4835): remove once Ops cleans up the PowerSchool stints that put two
-    -- enrollments on one entrydate.
+    -- keep one stint per entrydate; PowerSchool data entry occasionally creates two
+    -- and downstream rn_year = 1 filters assume this row survived
     {{
         dbt_utils.deduplicate(
             relation="final",
--- a/src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_enrollment_union.sql
+++ b/src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_enrollment_union.sql
@@ -206,20 +206,23 @@ with
             ) as grade_level_prev,
 
             row_number() over (
-                partition by studentid order by yearid desc, exitdate desc
+                partition by studentid
+                order by yearid desc, exitdate desc, reenrollments_dcid asc
             ) as rn_all,
 
             row_number() over (
-                partition by studentid, yearid order by yearid desc, exitdate desc
+                partition by studentid, yearid
+                order by yearid desc, exitdate desc, reenrollments_dcid asc
             ) as rn_year,
 
             row_number() over (
-                partition by studentid, schoolid order by yearid desc, exitdate desc
+                partition by studentid, schoolid
+                order by yearid desc, exitdate desc, reenrollments_dcid asc
             ) as rn_school,
 
             row_number() over (
                 partition by studentid, if(grade_level = 99, true, false)
-                order by yearid desc, exitdate desc
+                order by yearid desc, exitdate desc, reenrollments_dcid asc
             ) as rn_undergrad,
         from enr_union
     ),
--- a/src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_enrollment_union.sql
+++ b/src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_enrollment_union.sql
@@ -242,32 +242,46 @@ with
                 partition by studentid, rn_year order by yearid asc, exitdate asc
             ) as year_in_network,
         from window_calcs
+    ),
+
+    final as (
+        select
+            * except (rn_undergrad, year_in_school, year_in_network),
+
+            if(grade_level != 99, rn_undergrad, null) as rn_undergrad,
+            if(rn_year = 1, year_in_school, null) as year_in_school,
+            if(rn_year = 1, year_in_network, null) as year_in_network,
+            if(exitcode = 'G1', cohort_primary, null) as cohort_graduated,
+            if(exitdate is not null, true, false) as is_enrolled_y1,
+
+            if(
+                date(academic_year, 10, 1) between entrydate and exitdate, true, false
+            ) as is_enrolled_oct01,
+            if(
+                date(academic_year, 10, 15) between entrydate and exitdate, true, false
+            ) as is_enrolled_oct15,
+            if(
+                date(academic_year + 1, 3, 15) between entrydate and exitdate,
+                true,
+                false
+            ) as is_enrolled_mar15,
+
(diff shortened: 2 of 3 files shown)


cbini:
Review verdicts, per finding:

1. **Cross-link #5045** — Agree on the shared root cause; disagree on the pointer. #5045 is closed (fixed by #5050). The live Ops tracker for the source data is #4835, so that is what the new TODO cites. #5141 already Refs #4835.
2. **Missing `-- TODO:` on the dedupe** — Fixed in 2b7f79f. `TODO(#4835)` added above the `deduplicate` call, same shape as the kipptaf sibling.
3. **Needs warehouse verification** — Done locally. `dbt build` of the model against `kippnewark` (dev target, `--defer --favor-state`) passed, and the dev relation keeps `rn_year = 1` for the Newark collision group that prod currently gets wrong. Prod confirmation stays on the issue checklist for after merge.
4. **Nulls / ORDER BY rules** — Agree, no change.
5. **ponytail-review** — no change.

cbini:
Second-pass verdicts, per finding:

1. **ST03 suppression on `final`** — Declined, verified locally. I first added the `trunk-ignore(sqlfluff/ST03)` line and `trunk check --force` reported `trunk-ignore(sqlfluff/ST03) is not suppressing a lint issue` (`trunk/ignore-does-nothing`), so I removed it. ST03 does not fire on the terminal, unwrapped `deduplicate` shape; the 20 precedents are all wrapped mid-chain CTEs. The kipptaf sibling is the same shape and is correct as is.
2. **Paterson tie inside `rn_year`** — Confirmed and fixed in 9307591. The Paterson pair differs in `entrycode`, `next_school`, `sched_nextyeargrade`, and `reenrollments_dcid`, so the pick matters. All 4 `rn_*` windows now end in `reenrollments_dcid asc`; the students-table row (null dcid, nulls first) wins the tie. Dev build for `kipppaterson` passed; the surviving row is `rn_year = 1`, the students-table row, with `next_school` populated.
3. through 6. **No-violation checks and the unverifiable-statically note** — Agree, no change. Local validation for Newark and Paterson is in the PR body.
7. and 8. — no change.

PR #5162, opened 2026-09-05
fix: score NJGPA by assessment version, then simplify the graduation pathway model

### LINKED ISSUE #3994: fix(dbt): unexpected grad_eligibility value in int_students__graduation_path_codes (1 row)
## Context

`accepted_values_int_students__graduation_path_codes_grad_eligibility__*` test warns on 1 row in CI: a `grad_eligibility` value falls outside the accepted-values list (`Grad Eligible`, `ELA Only`, `Not Grad Eligible`, `Math Only`, `No FAFSA`, `FAFSA Only`, `ELA Only / No FAFSA`, `Math Only / No FAFSA`).

## Acceptance

- Identify the offending value.
- Either extend the accepted-values list or fix upstream classification logic.

## Surfaced in

PR #3985 CI run.

### LINKED ISSUE #4937: docs: HS Early Warning dashboard data model reference plus AY2026-27 rollover
## What and why

The High School Early Warning dashboard needs its annual rollover, and unlike FRESH, CARAT, and the gradebook audit it has no reference doc. Nobody has a written answer to "what has to be refreshed each August, and who owns each input" — so the rollover gets rediscovered from scratch every year.

Two deliverables: close the rollover gaps for AY2026-27, and write the missing data model doc.

### The dbt layer already rolled over

All three models behind the `high_school_early_warning_dashboard` exposure filter on `var("current_academic_year")`, and that var is `2026` in all five dbt projects. No SQL edits are needed for the year to advance:

- `rpt_tableau__hs_early_warning_dashboard`
- `rpt_tableau__graduation_requirements`
- `rpt_tableau__community_service`

Reporting terms are in place too — all three NJ high schools (KIPP Newark Collegiate Academy, KIPP Newark Lab High School, KIPP Cooper Norcross High) have AY2026-27 `RT` rows in `stg_google_sheets__reporting__terms`.

### Confirmed gap: graduation path cutoffs stop at cohort 2027

`stg_google_sheets__student_graduation_path_cutoffs` holds 16 rows each for cohorts 2023 through 2027 and nothing after. `int_students__graduation_path_codes` joins it on `cohort`, so a cohort with no rows gets no `cutoff`, no `met_pathway_cutoff`, and a null `grad_eligibility`.

Current NJ high school enrollment against that coverage:

| Grade | Cohort | Students | Cutoff rows present |
| ----- | ------ | -------- | ------------------- |
| 12    | 2027   | 395      | yes                 |
| 11    | 2028   | 405      | no                  |
| 10    | 2029   | 476      | no                  |
| 9     | 2030   | 568      | no                  |

Grade 11 is the NJGPA year, so cohort 2028 is the urgent one: graduation eligibility reads blank for roughly 405 juniors right now. The staging table last rebuilt 2026-08-13, so this is a real gap in the sheet, not stale-table lag.

### Still to trace

Two inputs whose owner and refresh cadence are not documented anywhere:

- Community service hour requirements per grade, read from DeansList student custom fields via `int_deanslist__students__custom_fields__pivot`. Unclear whether these carry forward automatically or need an annual reset.
- `stg_google_sheets__student_graduation_path_combos`, which has no `cohort` column at all — so whether it needs annual maintenance is an open question.

## Proposed approach

1. Trace every annual input behind the three models and record the owner and refresh trigger for each.
2. Get cohort 2028 rows into the cutoffs tab, and 2029 and 2030 while we are there. The values are a policy question for the HS team, not a data team call — we supply the shape, they supply the thresholds.
3. Write `docs/models/hs-early-warning-dashboard-data-model.md` and add it to the `mkdocs.yml` nav under Models, matching the existing three.

## Notes

Related: #3994 tracks an unexpected `grad_eligibility` value in `int_students__graduation_path_codes`.

A companion Claude skill, matching the `fresh-dashboard` and `gradebook-audit` ones, is a sensible follow-up once the doc exists. Out of scope here.

<details>
<summary>For Claude</summary>

> Context, not instructions.

Exposure: `high_school_early_warning_dashboard` in `src/dbt/kipptaf/models/exposures/tableau.yml`, LSID `6333e047-e7a9-4d8f-a740-3df30f179d11`, refreshes `0 6 * * *`. Workbook link is in `src/launch/links.yml` under id `high_school_early_warning`.

Models in scope:

- `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__hs_early_warning_dashboard.sql`
- `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__graduation_requirements.sql`
- `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__community_service.sql`
- `src/dbt/kipptaf/models/students/intermediate/int_students__graduation_path_codes.sql`

Sheet-backed inputs: `src_google_sheets__student_graduation_path_cutoffs` and `src_google_sheets__student_graduation_path_combos`, both on spreadsheet `1CbSOG-gs0dIPtf7wneJNhSqudRFKWpfcNoefyAEi5AM`, declared in `src/dbt/kipptaf/models/google/sheets/sources-external.yml`.

Follow the structure of `docs/models/fresh-dashboard-data-model.md` and `docs/models/gradebook-audit-data-model.md`. Both carry an explicit rollover procedure section — FRESH under "Rolling the dashboard over to a new cycle", the gradebook audit under "Start-of-year procedure". This doc needs the equivalent, since the rollover procedure is the reason it is being written.

Two oddities worth resolving in the doc rather than papering over:

- `rpt_tableau__hs_early_warning_dashboard` aliases `gr.need_60 as need_65`. Confirm which threshold is actually in force before documenting either number.
- `rpt_tableau__graduation_requirements` excludes Miami and filters cohort to a window of `current_academic_year - 1` through `current_academic_year + 5`, while `rpt_tableau__hs_early_warning_dashboard` has neither filter. Worth stating why they differ.

Acceptance: the doc names every annual input, its owner, and its refresh trigger, and a reader who has never touched this dashboard can complete next year's rollover from it alone. No student-level values in the doc — aggregate counts only.

</details>

### LINKED ISSUE #5153: fix: score NJGPA by assessment version, then simplify the graduation pathway model
## What's happening

New Jersey replaced the NJGPA with the adaptive NJGPA-A at the spring 2026 administration. The new test uses a different score scale and a graduation-ready cut score of 450 rather than 725. Both versions arrive with the same `assessment_name` and the same `testcode`, so a cut score keyed only on cohort compared every adaptive score against the retired scale. Every spring 2026 score read as not met. For the class of 2027 that is 221 English language arts passes and 147 mathematics passes reported as failures.

That is not only a dashboard problem. `final_grad_path_code` is written back to PowerSchool through `rpt_powerschool__autocomm_students` into the state reporting fields `s_nj_stu_x__graduation_pathway_ela` and `s_nj_stu_x__graduation_pathway_math`, on a daily extract. Wrong codes became wrong state submissions.

Fixing it exposed how fragile `int_students__graduation_path_codes` is. The same class of silent wrong answer can happen again in at least three other ways, so this issue now covers the fix and the cleanup that stops it recurring.

## What we are changing

**1. Key cut scores on the assessment version.** Each vendor's staging model sets `assessment_version` as a literal, the union carries it, and the cut score join uses it. In flight as #5162.

**2. Retire the graduation eligibility combos sheet.** `stg_google_sheets__student_graduation_path_combos` is a hand-maintained sheet of 83 rows joined on eight boolean columns. Any combination it does not cover falls through to a placeholder label, which is 68 rows today. The sheet turns out to be a pure function of its own inputs, so it can be replaced with a `case` expression and deleted. That also fixes 4 rows where the sheet has drifted.

**3. Make the pathway list data driven.** The cut score sheet is meant to be the source of truth for which pathways exist, but the list is hardcoded in three places. Adding a pathway to the sheet does nothing unless all three are edited, and missing one is silent.

**4. Add a grain test.** The model has none. Its grain silently doubled when the second NJGPA cut score row was added, and nothing would have caught it.

**5. Decode the pathway code upstream of Tableau.** The workbook mapped `final_grad_path_code` to a display label in a calculated field. That mapping also existed in dbt as `pathway_option`, and the two disagreed on code `O`. Move it into the model as `final_grad_path_name` and add a test asserting the two label lists stay in step.

**6. Rename `need_65` to `need_60` on `rpt_tableau__hs_early_warning_dashboard`.** The extract exposed `gr.need_60 as need_65`. Both grading scales the high schools use pass at 60, so the exposed name matched no scale in use. Values are unchanged; the workbook field needs repointing after merge.

## Steps to reproduce

1. Select `cutoff` from `kipptaf_google_sheets.stg_google_sheets__student_graduation_path_cutoffs` where `cohort` is 2027 and `score_type` is `ELAGP`. It returns 725, the cut score for the retired scale.
2. Select `testscalescore` and `administration_period` from `kipptaf_pearson.int_pearson__all_assessments` where `academic_year` is 2025 and `testcode` is `ELAGP`. The spring rows range from 300 to 562 and the fall rows from 650 to 786.
3. Select distinct `final_grad_path_code` from `kipptaf_students.int_students__graduation_path_codes` where `cohort` is 2027 and `grade_level` is 12. The result omits `S`, the state assessment pathway code.
4. Count rows in that model where `grad_eligibility` is the placeholder label. It returns 68.

## Where

- **Code location / dbt project:** `src/dbt/kipptaf`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** #5162

<details>
<summary>For Claude</summary>

Source document for the cut score: the NJDOE broadcast of August 5, 2026, "Graduation Ready Cut Scores Established for the New Jersey Graduation Proficiency Assessment-Adaptive (NJGPA-A)". Requirements index at <https://www.nj.gov/education/assessment/requirements/>, which links a page per graduating class. NJDOE publishes the cut score but not the scale range.

Both cut scores were confirmed from our own `testperformancelevel` rather than taken on faith. Fall administration: level 1 tops at 724, level 2 starts at 727, so the cut is 725. Spring administration: level 1 tops at 449, level 2 starts at 450.

The vendors separate cleanly with no scale overlap. `stg_pearson__njgpa` holds scores from 650 to 850 across spring 2021 through fall 2025. `stg_cambium__njgpa` holds 300 to 562 and covers spring 2026 only.

### The combos sheet is a pure function

Verified against all 83 rows of the sheet, zero mismatches:

```
eff_ela  = met_ela  and attempted_njgpa_ela
eff_math = met_math and attempted_njgpa_math
fafsa_counts = grade_level = 12 and fafsa_season_12th

case
  when eff_ela and eff_math and fafsa_counts and not has_fafsa then 'No FAFSA'
  when eff_ela and eff_math                                    then 'Grad Eligible'
  when eff_ela  and fafsa_counts and not has_fafsa             then 'ELA Only / No FAFSA'
  when eff_ela                                                 then 'ELA Only'
  when eff_math and fafsa_counts and not has_fafsa             then 'Math Only / No FAFSA'
  when eff_math                                                then 'Math Only'
  when has_fafsa and fafsa_season_12th and grade_level = 12    then 'FAFSA Only'
  when grade_level = 11 and not njgpa_season_11th
       and (not fafsa_season_12th or attempted_nothing)        then 'Grad Eligible'
  else 'Not Grad Eligible'
end
```

Three rules were encoded in the sheet and written down nowhere:

1. A subject only counts if the student sat the NJGPA in it. `met_ela` with `attempted_njgpa_ela` false contributes nothing, duplicating the `njgpa_attempt` rule already in the model.
2. FAFSA is required for graduation, but a student is not held to it until January of their senior year. Before that it is tracked and not counted against them.
3. An 11th grader holding NJGPA records is treated as a 12th grader, minus FAFSA, because testing ahead of their peers usually means they are behind on credits. The grace period belongs only to 11th graders with no records yet, and it ends once results land in late June, which is what `njgpa_season_11th` gates. The retired sheet gave that grace to any 11th grader before January, including those who sat the assessment and missed.

The published rule above drops the `and not njgpa_season_11th` term the sheet implies, which is drift rather than intent: it currently lets a senior who met both subjects with no FAFSA read "Grad Eligible" after June 30. Correcting it changes 4 sheet rows, all post-June seniors without FAFSA. No live effect today, because both season flags are false until January 2027.

### Fragility to address in step 3

The pathway list is hardcoded in the `unpivot_calcs` pivot, the `met_subject` or-chain (twice), and the `final_grad_path_code` cascade. That cascade also expresses precedence by restating every prior negative, so adding a pathway means editing every later branch correctly. The cut score sheet already carries `pathway_code` per option, so precedence could be driven from data.

Other fragility found in review, all three now addressed in #5162: the transfer score join used `studentid` with no `_dbt_source_project`, and `studentid` is unique only within a PowerSchool instance, so it now joins on both; the M/N/O/P branch duplicated a 16 column select list that BigQuery matches by position, and now uses `union all corresponding`, which binds by name; `met_pathway_cutoff` used `if(x >= y, true, false)` so a null score became false rather than null, and is now the bare comparison, which yields null and no longer disagrees with `points_short` directly below it.

</details>

GabyRangelB:
## Summary & Motivation

When merged, this pull request will score NJGPA results against the cut score for the test the student actually sat.

New Jersey replaced the NJGPA with the adaptive NJGPA-A at the spring 2026 administration. The new test uses a different score scale and a graduation-ready cut score of 450 instead of 725. Both versions arrive with the same `assessment_name` and the same `testcode`, so a cut score keyed only on cohort compared every adaptive score against the retired scale. Every spring 2026 score read as not met. For the class of 2027 that is 221 English language arts passes and 147 mathematics passes reported as failures, and `final_grad_path_code` is written back to PowerSchool for state reporting, so those became wrong state submissions rather than a dashboard problem.

Each vendor's staging model now sets `assessment_version` as a literal, `int_pearson__all_assessments` carries it through the union, and the cut score join uses it. `pathway_option` stays `NJGPA` for both versions, which leaves the roll-up and the pathway code logic untouched.

The cut score sheet moved to a new tab and named range, gained an `assessment_version` column, and renamed its headers to lowercase. Transfer scores are entered by hand in PowerSchool, and both versions now share one test holder, so that model reads the version off the score field name instead.

Fixing that exposed how fragile the pathway model was, so this pull request also simplifies it. `int_students__graduation_pathway_scores` now owns matching scores against cut scores, and `int_students__graduation_path_codes` keeps only the roll-up to a per-student standing. The hand-maintained graduation eligibility sheet is retired, because its label turned out to be a pure function of its own inputs. Both models gain a uniqueness test, which neither had, plus `one_cutoff_per_score_type`, which is the one that can actually fail on a fanned-out cut score join.

Refs #5153
Closes #3994
Closes #4937

## Reviewer Notes

**The cut score sheet must be saved before this merges.** The contract expects 8 lowercase columns including `assessment_version` and `pathway_code`. I verified the live sheet matches and built the model against it — 82 rows, all 14 tests pass.

**Student-visible changes, checked deliberately.** Comparing this branch's build against prod over the 430 students who carry a pathway code in both: 241 change code and **every one of them moves to `S`** — zero move anywhere else, so no student loses a pathway. Eligibility moves up for 33 and down for 4. All four of those are 11th graders who already hold NJGPA records and have met neither subject, which is the corrected rule rather than a regression: the pre-June grace now applies only to 11th graders with no records yet.

A further 107 students in prod are absent from this build, which is not this change. Every one is withdrawn, inactive, or gone from the source roster; the `where enroll_status = 0` filter is byte-identical on both branches. Prod's table has simply not rebuilt since they left.

**`rpt_tableau__graduation_requirements` gains exactly one column and loses none.** A full column diff against prod shows `final_grad_path_name` added and nothing else changed. Rows whose scores fall short are still carried, which the dashboard shows on purpose; rows carrying a matched score rise from 2,271 to 2,335, which is the fix landing.

**Two tests fail or warn on purpose today.** `cutoff_within_score_scale` returns 2 results against current prod data, which is the bug this fixes; it passes once the version-keyed cut scores land. `scores_have_cutoffs` warns for the classes of 2026 and 2028, where a retained student sat a version their own cohort has no row for and New Jersey has published no cut score yet.

**`int_pearson__all_assessments` gains its first per-column tests, both at error severity.** That halts everything downstream of a hub model when it trips. I think that is right for a value that decides state reporting, but it is a real change in that model's failure behavior.

**Accuplacer is deliberately absent, permanently.** The broadcast lists four Accuplacer options. KTAF does not use Accuplacer and does not intend to, so no cut score rows were added and the decision is recorded in the new skill. We also hold no New Jersey pathway code letter for them, and that value goes to the state.

**A contract column is renamed on a second extract, and Tableau needs repointing.** `rpt_tableau__hs_early_warning_dashboard` exposed `gr.need_60 as need_65`. The underlying calculation targets 60, which is the lowest passing grade on both grading scales the high schools actually use, so the exposed name was wrong and no scale in use passes at 65. It is now `need_60`. A column diff against prod confirms that is the only change to that model, and the values are untouched: of the rows that differ, every single one also has a different `y1_grade_percent_adjusted`, and none differ on the need value alone — that is upstream grade movement, not this change. The workbook field must be repointed after merge or it breaks.

**Documentation.** `docs/models/hs-early-warning-data-model.md` is new and covers all five dashboard tabs, the three feeds, the early warning flags and their verified thresholds, and the pathway rules. The `graduation-pathways` skill carries the runbooks for new cut scores and portfolio appeals. Both were written for whoever picks this up next.

### Before this leaves draft

- [x] Add unit tests to `int_students__graduation_path_codes` locking the cut
      score boundaries — a retired-scale score at 724 and 725, an adaptive score
      at 449 and 450, and one student holding both versions. Done, and
      mutation-checked: removing the `assessment_version` predicate from the cut
      score join makes the test fail.

### General

### dbt

- [x] Include (or update) an
      summary above.
      See

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Source document — the NJDOE broadcast of August 5, 2026, "Graduation Ready Cut Scores Established for the New Jersey Graduation Proficiency Assessment-Adaptive (NJGPA-A)". Sets NJGPA-A English language arts and mathematics at 450 for the class of 2027, keeps the alternative assessment menu unchanged from the classes of 2024-2026, and keeps Portfolio Appeals available.

Both cut scores confirmed from our own `testperformancelevel` values rather than taken on faith. Fall administration, level 1 tops at 724 and level 2 starts at 727, so the cut is 725. Spring administration, level 1 tops at 449 and level 2 starts at 450.

The vendors separate cleanly with no scale overlap. `stg_pearson__njgpa` runs 650-850 across spring 2021 through fall 2025. `stg_cambium__njgpa` runs 300-562 and holds spring 2026 only.

Why the version is a literal per staging model rather than derived. The first attempt matched `regexp_contains(_dbt_source_relation, 'stg_cambium__njgpa')` in the union view with an `else 'NJGPA'` fallback. That fallback silently mislabels a future third vendor as the retired form and reintroduces this exact bug. Setting the literal where the vendor already asserts `'NJGPA' as assessment_name` removes the inference. `_dbt_source_project` cannot substitute — it resolves to region, and Newark and Camden each carry both vendors.

Because `union_relations` null-fills a column absent from a relation, a sixth relation added without the literal would arrive null, so `assessment_version` carries `not_null` alongside `accepted_values`. `accepted_values` alone passes nulls.

Verified impact, enrolled students with complete scores. 371 student-discipline records move to `S`. Of those, 75 records across 67 students currently read `R`, meaning no pathway met — 37 at KIPP Newark Collegiate Academy, 22 at KIPP Newark Lab High School, 16 at KIPP Cooper Norcross High. The remaining 296 already showed as met via SAT or PSAT/NMSQT and only get a corrected pathway code. Zero students lose a pathway; that was checked explicitly rather than assumed.

Why the class of 2027 keeps both a 725 row and a 450 row. Three enrolled students in that cohort hold retired-scale scores, none of which cleared 725. A flat move to 450 would mark them as having met the pathway.

Why `cohort` stays in the key even though it is the wrong shape for this. `cohort` freezes at first year in the school for grade 9 and up, which is correct for New Jersey's 4-year adjusted cohort graduation rate. The consequence is that a retained or accelerated student sits a version their cohort has no row for; `scores_have_cutoffs` names them rather than hiding them.

Transfer scores. The PowerSchool holder was renamed to `NJGPA/NJGPA-A` and carries four score fields, `ELAGP`, `MATGP`, `ELAGP-A`, `MATGP-A`. The model reads the version off the `-A` suffix and strips it so `score_type` still joins to the sheet, keeps the raw value in `score_name` for tracing, and matches both the old and new holder names because the two PowerSchool instances are renamed independently. Paterson has no holder configured and therefore no rows, which is expected — that is recorded in the new reference doc and skill.

Local verification. `stage_external_sources` plus `dbt build` into a personal dev schema against the live sheet, contract enforced, 15 nodes pass. The version join was checked against real scores — 748 adaptive rows for the class of 2027 with 368 met at 450, and 4 retired-scale rows with 0 met at 725.

Still outstanding. Unit tests locking the 724/725 and 449/450 boundaries, and a student holding both versions. `int_students__graduation_path_codes` has six inputs and `assessment_version` is a same-pull-request column add, so the fixtures need `format: sql` rather than dict format.

</details>

---
  - https://app.asana.com/0/0/1218202955524334


GabyRangelB:
Verdicts on all seven findings, verified against the code and the warehouse before responding. Fixed in `51f7474`.

**1. Grain tests are tautological — split verdict.**

Correct for `int_students__graduation_pathway_scores`. `rn_highest` is `row_number()` partitioned by `(student_number, score_type)`, a strict subset of the tested tuple, so the combination is unique by construction and the test cannot fail.

Not correct for `int_students__graduation_path_codes`. There `rn_highest` is passed through, not regenerated, and the `roster` join to `calcs` could fan out — so a duplicate is reachable and that test does real work. It caught something, see finding 8 below.

The suggested fix does not work either: dropping `rn_highest` leaves 3,490 duplicate keys across 75,908 rows, because students legitimately sit the same test more than once. That is the grain, not a defect. And every intermediate is required to carry a uniqueness test, so it cannot simply be removed.

What was missing is a test that can fail, so there is now `one_cutoff_per_score_type`. It asserts a student faces exactly one cut score per subject, score type and version. The sheet's own uniqueness test does not cover this: the sheet is keyed on cohort, discipline, score type and version, while the join also matches `pathway_option` — two sheet rows differing only there pass the sheet test and fan a student out here, which is exactly how the grain doubled when the second NJGPA row was first added. It scans 58,221 groups carrying up to 8 rows each.

The PR body claim has been corrected to match.

**2. Grace-period boundary unchecked — fixed.** The test now selects `njgpa_season_11th` and its exemption reads `grade_level = 11 and attempted_nothing and not njgpa_season_11th`, matching the model.

**3. `final_grad_path_name` not_null reachable — mechanism confirmed, no change, deliberately.**

The unmapped codes are real: `W` on 5,155 rows, plus `B`, `A`, `C` and `L`. But all five sit only on grade 11-12 rows with `enroll_status != 0`, and the model filters `enroll_status = 0`; at grade 10 and below the only codes present are `R` and `M`, both mapped. So it cannot fire today.

Leaving it as-is is the intent rather than an oversight. If a code with no branch in the decode ever does reach an active student, `final_grad_path_code` still carries it into the daily state submission. Failing the build is the better outcome there than emitting a code we cannot label.

**4. Stale `grad_eligibility` description — fixed.** It was the same grade-level text, and it now describes what the column holds.

**5. Missing comment on the strict union — fixed.** Worth more than style: this is 2 of 27 uses in the repo, every other one being the `FULL` form, so the hard failure on a one-sided column add is genuinely surprising without the note.

**6. `scores_have_cutoffs` dagster ref — deliberate, confirmed.** The failure is always a missing sheet row and the fix is always a sheet edit, so the check belongs on the sheet's asset. Agreed it is a new pattern rather than precedent.

**7. Pre-existing YAML typos — fixed, all three.** `studentsid` and `rn_discipline_distict` named no real column. `has_fafsa` now has a `data_type` and description. Predating the PR is fair, but the model was rewritten here, so they are mine now.

**8. Found while verifying finding 1, not in the review.** `calcs` grouped by `_dbt_source_relation` while `roster` joined it on `student_number` and `discipline` only — the cross-region join rule in the kipptaf guide. A student number under two source relations would have fanned out. It now groups and joins on `_dbt_source_project`. No student spans two projects today, so this is a verified no-op that closes the latent fan-out.

GabyRangelB:
@claude please re-review. Commit `51f7474` responds to all seven findings from the previous pass, and the per-finding verdicts are in the comment above.

Worth a close look on three of them, since I did not simply implement what was suggested:

- Finding 1 was accepted for `int_students__graduation_pathway_scores` but declined for `int_students__graduation_path_codes`, where `rn_highest` is passed through rather than regenerated and the test does real work. The suggested fix of dropping `rn_highest` was rejected on data: 3,490 duplicate keys across 75,908 rows, which is the real grain. A new `one_cutoff_per_score_type` test covers the gap instead.
- Finding 3 was declined deliberately, not overlooked. Please check the reasoning about unmapped codes reaching the state submission.
- Finding 8 is new, found while verifying finding 1: a cross-region join that omitted `_dbt_source_project`.

GabyRangelB:
Both points taken, and one of them is a correction I should own.

**On the path_codes uniqueness test, your phrasing is more accurate than mine.** I said it "does real work"; the honest statement is that it *could* fail if `calcs` ever yields more than one row per source project, student and discipline, which needs `ps_grad_path_code` or the attempted flags to disagree across a student's own rows. Today's upstream grain makes those constant per student and discipline, so it is a guard against the joins drifting, not a demonstrated risk. The distinction from the `pathway_scores` version is real — that one is unique by construction and cannot fail under any data — but I overstated the other side of it.

**Description added to both blocks** in `dbef251`, since you are right that blame leads here and the nuance would otherwise have to be re-derived. Each now says whether it can fail, why it is kept, and which test covers the gap. Precedent for a `description:` on a generic test is `rpt_tableau__ap_assessment_dashboard.yml`.

On the two data claims you flagged as unverified rather than doubted, for the record, both were measured against the warehouse before I replied:

- `one_cutoff_per_score_type` scans 72,576 groups: 58,221 carry exactly one distinct cutoff with up to 8 rows each, and 14,355 carry none, those being the PowerSchool-decided rows where `cutoff` is null by design. So the predicate constrains groups that really do hold multiple rows.
- The unmapped codes are `W` on 5,155 rows, `B` on 329, `A` on 267, `C` on 90 and `L` on 3. Every one sits on a grade 11-12 row with `enroll_status != 0`, and at grade 10 and below the only codes present are `R` and `M`, both mapped.

The full quality comparison against prod is in the PR body: 241 students change pathway code and all 241 move to `S`, with no code regressions.

GabyRangelB:
@claude one more pass please, this time for a merge-readiness verdict on `dbef251` rather than a fresh hunt.

Since your last pass on `51f7474`, the only change is `dbef251`, which adds a `description:` to the two `unique_combination_of_columns` blocks — the non-blocking suggestion you closed on, plus the correction to my "does real work" wording. No SQL changed.

What I am asking for specifically:

1. Do you consider this mergeable as it stands, or is there anything you would want fixed first?
2. Are the two test descriptions accurate? I wrote them to state whether each test can fail and why it is kept, and I would rather they be right than flattering.
3. Anything in the PR body or the reference doc that overstates what was verified. Two passes have now flagged that you cannot check data claims from this job, so I would rather you challenge the wording than take it on trust.

For context on the data claims you could not verify, the measurements are in my reply above and the prod comparison is in the PR body. Headline: 241 students change pathway code, all 241 to `S`, zero regressions; column diffs limited to one added column, one rename, and no change at all to the model that feeds the state submission.

GabyRangelB:
Good catch on the checkbox — it was an unticked box, not a skipped step, and it is worth having the evidence on the record rather than the inference. Ticked now.

`stage_external_sources --target staging` was run. Confirmed against `INFORMATION_SCHEMA.TABLE_OPTIONS` rather than reasoning from CI passing:

| Schema | `sheet_range` |
| --- | --- |
| `zz_stg_kipptaf_google_sheets` (what dbt Cloud CI reads) | `src_students__graduation_path_cutoffs_v2` |
| `kipptaf_google_sheets` (prod) | `src_students__graduation_path_cutoffs` |

So CI is genuinely validating against the new range, and prod still points at the old one, which is correct pre-merge.

Worth noting for whoever merges: prod needs no manual re-stage. Sheets externals are re-staged by the `build_dbt_assets` stage-refresh-build flow on the next tick, unlike a new Avro or GCS source, which does need launching in prod immediately after merge. That distinction is in `.claude/rules/dbt-yaml.md`, and the rule there says to verify with `INFORMATION_SCHEMA` before filing a manual re-stage as outstanding work — which is the query above.

The live sheet save remains the one real pre-merge gate, as all three passes have said.

cbini:
Verdicts on the 4 findings from the pass on the merged package refactor. Fixes are in `160f95f`.

**1. Six passthroughs deleted instead of disabled — declined, and the rule is amended.** The deletes were a deliberate call made on #5194: these were `select *` union wrappers over district sources with no logic and no remaining `ref()`, and the district relations they read stay in place, so disabling preserves nothing a forgotten consumer could still use. The rule in `.claude/rules/dbt-models.md` did not carve that case out, which is why the finding is correct as written. It now does, citing this PR, so the next reviewer reads the decision instead of the contradiction. The `stg_google_sheets__student_graduation_path_combos` retirement is different: that model holds logic and a prod relation consumers may still read, so disable is right there.

**2. `src/dbt/pearson/CLAUDE.md` said "Staging-only" — fixed.** The opening line now reads staging plus two intermediates.

**3. Two-level function nesting in `int_pearson__student_list_report` — fixed.** The `Grade (\d+)` regexp is hoisted into a `base` CTE as `grade_number`, so both `concat(... lpad(...))` branches nest one level. Rebuilt in kippnewark dev: 21,454 rows, uniqueness test passes, and the `aligned_test_code` distribution is identical to the pre-change staging copy.

**4. Race, IEP and ML mappings duplicated between the pearson and cambium packages — acknowledged, no change.** The tradeoff and its reason are in both models and in #5194's body. A shared macros package for about 40 lines used in two places is not worth its weight today.

cbini (approved):

PR #5168, opened 2026-09-06
feat(fresh): flag SIS-enrolled students missing from Finalsite on the QC worklist

### LINKED ISSUE #5167: feat: flag SIS-enrolled students missing from Finalsite on the FRESH QC worklist
## What and why

The FRESH dashboard Progress to Goals tab showed 55 students for Paterson Prep Middle School grade 5 on 2026-09-06. PowerSchool showed 56 actively enrolled students for the same school, grade and year.

One student caused the whole gap. That student is enrolled in PowerSchool and carries a Finalsite ID on the PowerSchool record. No Finalsite record with that ID exists in `stg_finalsite__status_report`, either in the network-level model or in the `kipppaterson_finalsite` source it unions.

`rpt_tableau__fresh_dashboard_qc` cannot show this student. The worklist reads `int_tableau__finalsite_student_scaffold`, and that model starts from Finalsite. A student with no Finalsite record produces no row at all, so no flag can fire on them.

The existing `is_missing_sis_record` flag covers only the opposite direction. It finds a Finalsite record that matched no SIS enrollment record.

So the enrollment team has no way to find these students. Someone has to notice that 2 counts disagree, then ask the data team to trace a single student by hand. That is how this one surfaced.

5 students are in this state across New Jersey today. Newark has 2, Camden has 2, and Paterson has 1. I checked New Jersey only, so Miami needs its own count.

Add a flag so SRE and the enrollment team can find these students themselves.

## Proposed approach

Add a flag named `is_missing_finalsite_record` to `rpt_tableau__fresh_dashboard_qc`.

Build the new rows from the SIS side, then union them onto the existing Finalsite-grain roster. A student with no Finalsite record has no `finalsite_id`, so these rows cannot come from the current source.

Decide 2 things before writing SQL. The worklist grain has to change, because the uniqueness test keys on `finalsite_id` and the new rows have none. Miami scope also needs a decision, because Miami reaches the SIS through Focus rather than PowerSchool.

<details>
<summary>For Claude</summary>

> Context, not instructions.

**Where the gap is.** `rpt_tableau__fresh_dashboard_qc` selects only from `int_tableau__finalsite_student_scaffold`. That model's first CTE, `latest_status_calc`, reads `int_finalsite__status_report_unpivot` and inner-joins the status crosswalk. Every downstream row therefore requires a Finalsite record. `deduplicate_enrollments` left-joins the SIS on `finalsite_id = infosnap_id`, so the SIS is reachable only for students Finalsite already knows about. The reverse direction (SIS row, no Finalsite row) is structurally unrepresentable in the current model.

**Constraint 1 — the uniqueness test.** `properties/rpt_tableau__fresh_dashboard_qc.yml` runs `dbt_utils.unique_combination_of_columns` on `(enrollment_academic_year, finalsite_id, flag_name)` at `severity: error`. SIS-only rows have a null `finalsite_id`, and BigQuery groups nulls together, so 5 such rows in one year collide on one key and fail the test. The grain needs a real key for these rows — likely `powerschool_student_number` plus a source discriminator, or a coalesced student key.

**Constraint 2 — enforced contract.** `src/dbt/kipptaf/dbt_project.yml` sets `+contract: enforced: true` on the `extracts` folder (line 120-121 region), so the model is contract-enforced. Any new or retyped column needs a matching `data_type` in the YAML or the build fails.

**Constraint 3 — Miami.** The SIS side is two sources, not one. `int_extracts__student_enrollments` covers Newark, Camden and Paterson and carries `infosnap_id`. Miami comes from `int_focus__student_enrollment_roster`, and its Finalsite link runs through `int_finalsite__contact_id_attributes` — see the `focus_enrollments_with_finalsite` CTE in `int_tableau__finalsite_student_scaffold` for the existing join. Confirm with the user whether Miami is in scope for the first pass.

**Constraint 4 — PII.** `finalsite_id`, `powerschool_student_number`, `first_name` and `last_name` all carry `config.meta.contains_pii: true` on this model. New identifier columns need the same tag.

**Constraint 5 — the exposure.** `models/exposures/tableau.yml` already lists `ref("rpt_tableau__fresh_dashboard_qc")` under `fresh_dashboard`, so no exposure change is needed, but a grain change is visible to Tableau.

**Verification evidence, gathered 2026-09-06 against prod.** PPMS is `schoolid = 2`, region Paterson. `int_extracts__student_enrollments` at `academic_year = 2026`, `rn_year = 1`, `enroll_status = 0`, `grade_level = 5` returns 56 distinct student numbers. `rpt_tableau__fresh_dashboard_progress_to_goals` at the same school and grade, `row_type = 'Student'`, `enrollment_type = 'All'` returns 55 distinct students, all with `latest_status = 'Enrolled'`. Anti-joining the 2 sets leaves exactly 1 student, and 0 dashboard rows have a null student number.

The missing student is not on `stg_google_sheets__finalsite__exclude_ids`, is absent from `int_finalsite__status_report_unpivot`, and no Finalsite record matches their last name under a different ID. Paterson's `2026_27` partition holds 1635 records, so ingestion is healthy. The student's only enrollment record ever is AY2026 at PPMS grade 5 with a 2026-07-01 entry date, which is the standard New Jersey bulk-rollover stamp.

Network-wide count, which is the query the new flag should reproduce:

```sql
with ps as (
  select region, student_number, infosnap_id
  from `teamster-332318.kipptaf_extracts.int_extracts__student_enrollments`
  where academic_year = 2026 and rn_year = 1 and enroll_status = 0
    and region in ('Newark', 'Camden', 'Paterson')
),
fs as (
  select distinct finalsite_enrollment_id
  from `teamster-332318.kipptaf_finalsite.stg_finalsite__status_report`
)
select
  ps.region,
  count(*) as ps_enrolled,
  countif(ps.infosnap_id is null) as no_infosnap_id,
  countif(ps.infosnap_id is not null and fs.finalsite_enrollment_id is null) as infosnap_not_in_finalsite
from ps
left join fs on ps.infosnap_id = fs.finalsite_enrollment_id
group by 1
```

It returns Camden 2, Newark 2, Paterson 1, with `no_infosnap_id` at 0 for all 3 regions.

**Acceptance criteria.** `rpt_tableau__fresh_dashboard_qc` returns 1 row for the Paterson student described above with `flag_name = 'is_missing_finalsite_record'`, and 5 such rows across the 3 New Jersey regions for AY2026. The existing 4 flags return the same rows they do today. The uniqueness test passes. `uv run dbt build --select rpt_tableau__fresh_dashboard_qc+` succeeds.

**Docs.** `docs/models/fresh-dashboard-data-model.md` states the worklist has 4 flags in its QC section, and `.claude/skills/fresh-dashboard/SKILL.md` repeats that count. Both need updating, along with the model description in the YAML.

</details>

GabyRangelB:
## Summary & Motivation

When merged, this pull request will add a fifth flag to the FRESH QC worklist, `is_missing_finalsite_record`, so the enrollment team can find students the dashboard cannot count.

The Progress to Goals tab counts Finalsite records, not SIS enrollments. A student who is enrolled in the SIS but has no Finalsite record is therefore missing from the count, with nothing to explain the gap. Paterson Prep Middle School grade 5 showed 55 students against PowerSchool's 56 for exactly this reason.

The QC worklist could not show that student either. It reads `int_tableau__finalsite_student_scaffold`, which starts from Finalsite, so a student with no Finalsite record produces no row and no flag can fire. The existing `is_missing_sis_record` flag covers only the opposite direction.

Today someone has to notice that 2 counts disagree and then ask the data team to trace a single student by hand. After this change they filter the worklist by the new flag instead.

9 students fire the new flag right now. Newark has 2, Camden has 2, Paterson has 1, and Miami has 4.

## Reviewer Notes

**The new rows are unioned on, not unpivoted.** Every other flag describes a Finalsite record, so it can be computed on a Finalsite-sourced roster. This one describes a student Finalsite has never heard of, who cannot appear in that roster at all. The model reads `int_extracts__student_enrollments` directly and anti-joins Finalsite.

**The anti-join ignores the recruitment year on purpose.** I measured the year-scoped version and rejected it. Details in the fold-out.

**Miami needs a different id path.** Every Miami row carries a null `infosnap_id`, so the Finalsite identity comes from `int_finalsite__contact_id_attributes` instead.

**`finalsite_id` is null on the new rows, and the uniqueness test changed because of it.** The test now also keys on `powerschool_student_number`.

**`sis_school` is inconsistent upstream, and this change does not hide that.** Worth a look if you disagree with the call. Details in the fold-out.

### General

No Dagster changes.

- [x] Include (or update) an
      summary above.
      See

No column was added, renamed or removed, so the contract is unchanged and no exposure edit was needed. `rpt_tableau__fresh_dashboard_qc` is already listed under the `fresh_dashboard` exposure. Tableau does see new rows, and on those rows `finalsite_id` is null for the first time. No external source changed.

No schedule, sensor or integration changes. `docs/models/fresh-dashboard-data-model.md` and the `fresh-dashboard` skill both said the worklist has 4 flags, so this updates both.

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Started from a count discrepancy the user reported (PPMS grade 5, 56 in PowerSchool vs 55 on Progress to Goals) and traced it to a single student whose PowerSchool record carries a Finalsite ID that exists nowhere in `stg_finalsite__status_report`, at either the kipptaf union or the `kipppaterson_finalsite` source. That student is not on `stg_google_sheets__finalsite__exclude_ids`, and no Finalsite record matches their last name under a different id, so it is a genuinely absent record rather than a stale link. Paterson's `2026_27` partition holds 1635 records, so ingestion was healthy.

**Why the anti-join is unscoped by year.** Scoping to `var("finalsite_recruitment_year")` via `int_finalsite__status_report_unpivot` was measured against prod. It took Newark from 2 flagged students to 8. All 6 additions were records that exist but sit in the 2025 or 2027 cycle, or that have no status dates for the `UNPIVOT` to produce rows from (BigQuery `UNPIVOT` drops all-null rows). None of those is "Finalsite has no record for this student", and each needs a different conversation with SRE, so folding them into this flag would make it noisy and ambiguous. Camden and Paterson were unchanged between the two definitions. If SRE later wants "enrolled but not in the current cycle", that is a separate flag.

**Why one branch instead of two.** The approved design called for a second branch reading `int_focus__student_enrollment_roster` for Miami, mirroring `int_tableau__finalsite_student_scaffold`. That turned out to be unnecessary. `int_extracts__student_enrollments` already carries all 1507 currently-enrolled Miami students at `academic_year = 2026`, `rn_year = 1`, `enroll_status = 0`, and its `student_number` domain matches the Focus roster exactly (1507 of 1507 join). The single branch plus a `coalesce` onto the contact-id crosswalk reproduces the identical 9 students, so the Focus ref and the `UNION ALL` were dropped. Verified by running both shapes.

**Fan-out was checked, not assumed.** `int_finalsite__contact_id_attributes` has 4608 rows and 4608 distinct `(focus_student_id, _dbt_source_project)` pairs, and `int_extracts__student_enrollments` at those filters has 11024 rows and 11024 distinct `(student_number, academic_year)` pairs. So the crosswalk left join cannot duplicate a student, and `sis_only` emits at most one row per student. That is what makes the widened uniqueness test safe.

**Why `powerschool_student_number` joined the uniqueness test.** The test was `(enrollment_academic_year, finalsite_id, flag_name)` at `severity: error`. New rows have a null `finalsite_id`, and BigQuery groups nulls together in `GROUP BY`, so all 9 would collide on one key and fail. Adding `powerschool_student_number` can only refine the existing key, so the 4 original flags keep passing — confirmed by the dev build, where all 4 return byte-identical counts to prod.

**The `sis_school` wart.** `sis_school` is already inconsistent in prod. On the 4 unpivoted flags it comes from `int_tableau__finalsite_student_scaffold`, which reads the PowerSchool branch for New Jersey (an abbreviation, e.g. `PPMS`) and the Focus branch for Miami (a full name, e.g. `KIPP Courage Academy`). The new rows read `int_extracts__student_enrollments.school`, which is the abbreviation in every region including Miami (`Courage`, `Royalty`). I chose not to reproduce the full-name form for Miami, because the abbreviation matches the `school` column's documented meaning and matches every New Jersey row. The result is that Miami `sis_school` differs by flag. This is documented on the column, and the underlying inconsistency is arguably the thing to fix, separately.

**Verification.** Built with `uv run dbt build --select rpt_tableau__fresh_dashboard_qc --target dev --defer --favor-state --state <main repo>/src/dbt/kipptaf/target/prod` from the worktree. `PASS=2 WARN=0 ERROR=0`, both before and after `trunk fmt` reformatted the SQL. Against the built dev view, the new flag returns 9 rows (Newark 2, Camden 2, Paterson 1, Miami 4), including the PPMS grade 5 student, whose row carries `schoolid = 2`, `school = 'PPMS'`, `grade_level = 5`, `sis_enroll_status = 0` and a null `finalsite_id`. The 4 existing flags return 19, 18, 31 and 7 rows in both dev and prod. `trunk check --force --no-fix` over all 4 changed files reports no issues.

One of the 9 rows (Miami, grade 7) has a null `schoolid` and `school`. That is the single Focus enrollment with no school assigned, not a join failure.

**Not covered.** A student whose Finalsite record exists but is dropped by the status crosswalk inner join still produces no worklist row. That is a pre-existing gap in a different place and is out of scope here.

Closes #5167

</details>

---
  - https://app.asana.com/0/0/1218210363748280

### DIFF
--- a/.claude/skills/fresh-dashboard/SKILL.md
+++ b/.claude/skills/fresh-dashboard/SKILL.md
@@ -994,11 +994,27 @@ regression, not a cleanup — the same trap that produced a wrong doc claim abou
 `is_grade_level_mismatch` / `is_school_mismatch`, which DO collapse NULL to
 `false` because they are wrapped.
 
-**The worklist has four flags, not five.** `is_same_day_status_tie` was deleted
-at the AY2026 review and replaced by the pending-status set inside
-`is_enroll_status_mismatch`. The same-day tie still happens in the data and the
-Reset Protocol is still the fix — it just no longer gets its own worklist row,
-so don't re-add the flag when someone reports a wrong `latest_status`.
+**The worklist has five flags, and `is_same_day_status_tie` is not one of
+them.** It was deleted at the AY2026 review and replaced by the pending-status
+set inside `is_enroll_status_mismatch`. The same-day tie still happens in the
+data and the Reset Protocol is still the fix — it just no longer gets its own
+worklist row, so don't re-add the flag when someone reports a wrong
+`latest_status`. The fifth flag is `is_missing_finalsite_record`, added in
+#5167, which is a different thing entirely.
+
+**`is_missing_finalsite_record` is the only flag sourced from the SIS side.** It
+fires when a student is currently enrolled in
+`int_extracts__student_enrollments` but no Finalsite record exists for them
+under any cycle, and it is `UNION ALL`ed onto the worklist rather than unpivoted
+— a student Finalsite never knew about cannot appear in a Finalsite-sourced
+roster. Miami reaches it through `int_finalsite__contact_id_attributes`, because
+Miami rows carry no `infosnap_id`. Do NOT "fix" the unscoped anti-join by adding
+a `finalsite_recruitment_year` filter: that was measured and rejected, because
+it also flags students whose Finalsite record merely sits in an adjacent cycle
+or has no status dates to unpivot. Those rows are also the students who are
+missing from the Progress to Goals count, so this flag is what a "PowerSchool
+says N, the dashboard says N-1" question should be answered with now, instead of
+tracing one student by hand.
 
 **`is_enroll_status_mismatch` has TWO directions in the docs, not three.** The
 "left" and "not finished enrolling" statuses were presented separately until the
--- a/.claude/skills/fresh-dashboard/SKILL.md
+++ b/.claude/skills/fresh-dashboard/SKILL.md
@@ -938,6 +938,22 @@ it with:
 uv run dbt parse --target prod --project-dir src/dbt/kipptaf --target-path target/prod
 ```
 
+**Deferred cleanup — rename the shadowed `focus_student_id` alias.** Do this at
+the AY2027-2028 rollover, not before; it was raised on #5168 and explicitly
+deferred as not worth its own PR. Both
+`int_tableau__finalsite_student_scaffold.sql` and
+`rpt_tableau__fresh_dashboard_qc.sql` open a `finalsite_contact_ids` CTE with
+`cast(focus_student_id_prefixed as int) as focus_student_id`. That alias shadows
+a real, differently-valued column on `int_finalsite__contact_id_attributes` —
+the genuinely unprefixed `focus_student_id` — so a reader diffing either CTE
+against its upstream model can take the prefixed value for the bare one. The
+prefixed value IS the student number (`concat('8400', focus_student_id)`), and
+the better-named precedent is `stg_people__student_logins.sql`, which calls it
+`student_number` and reserves `student_number_bare` for the unprefixed id.
+**Rename in BOTH models in one change** — doing one leaves the pair divergent,
+which is worse for the reader than the shared bad name. It is a CTE-internal
+alias in both places, so nothing outside those two models sees it.
+
 **When to make the change:** whenever SRE says the recruitment cycle has rolled
 over — not on a fixed schedule. There is no "revert" step the way
 gradebook-audit's summer toggle has; this is a one-directional bump forward each
--- a/docs/models/fresh-dashboard-data-model.md
+++ b/docs/models/fresh-dashboard-data-model.md
@@ -904,6 +904,15 @@ whole pipeline at the new cycle, and several models `inner join` against sheets
 scoped to that year. Flipping the var before those sheets carry the new year's
 rows does not error — it silently returns zero rows.
 
+One cleanup is parked against this rollover rather than done on its own.
+`int_tableau__finalsite_student_scaffold` and `rpt_tableau__fresh_dashboard_qc`
+both alias `cast(focus_student_id_prefixed as int)` to `focus_student_id`, which
+is the name of a different, genuinely unprefixed column on
+`int_finalsite__contact_id_attributes`. The value is correct and nothing is
+broken; the name misleads. Rename it in both models together at the AY2027-2028
+rollover — renaming one alone leaves the two CTEs divergent, which reads worse
+than the shared bad name. Raised on #5168 and deferred there.
+
 ### Steps, in order
 
 | #   | Step                                                              | Owner           |
--- a/docs/models/fresh-dashboard-data-model.md
+++ b/docs/models/fresh-dashboard-data-model.md
@@ -744,17 +744,33 @@ through untouched.
 
 ### The QC worklist: `rpt_tableau__fresh_dashboard_qc`
 
-`is_enroll_status_mismatch` is one of the worklist's four flags. Three of them —
+`is_enroll_status_mismatch` is one of the worklist's five flags. Three of them —
 this one, `is_grade_level_mismatch` and `is_school_mismatch` — are computed in
-`int_tableau__finalsite_student_scaffold`; `is_missing_sis_record` is derived in
-`rpt_tableau__fresh_dashboard_qc` itself. `rpt_tableau__fresh_dashboard_qc` is
-the SRE-facing surface for all four: it takes the roster at
-`grouped_status_timeframe = 'Current'`, `UNPIVOT`s the flags into
-`(flag_name, flag_value)`, and keeps only the rows where a flag actually fired
-(`where flag_value`). So it is a **worklist, not a report** — one row per
-student per problem, and an empty result is the good outcome.
+`int_tableau__finalsite_student_scaffold`; `is_missing_sis_record` and
+`is_missing_finalsite_record` are derived in `rpt_tableau__fresh_dashboard_qc`
+itself. `rpt_tableau__fresh_dashboard_qc` is the SRE-facing surface for all
+five: it takes the roster at `grouped_status_timeframe = 'Current'`, `UNPIVOT`s
+the first four flags into `(flag_name, flag_value)`, keeps only the rows where a
+flag actually fired (`where flag_value`), then `UNION ALL`s the
+`is_missing_finalsite_record` rows on. So it is a **worklist, not a report** —
+one row per student per problem, and an empty result is the good outcome.
 
-#### The four flags, in plain language
+**`is_missing_finalsite_record` is unioned on, not unpivoted, and that is
+structural.** Every other flag describes a Finalsite record, so it can be
+computed on a Finalsite-sourced roster. This one describes a student Finalsite
+has never heard of, and no such student can appear in that roster at all — the
+model therefore reads `int_extracts__student_enrollments` directly, resolves
+each student's Finalsite identity (`infosnap_id`, falling back to
+`int_finalsite__contact_id_attributes` for Miami, whose rows carry no
+`infosnap_id`), and anti-joins against `stg_finalsite__status_report`. The
+anti-join is deliberately unscoped by year: a record filed under an adjacent
+cycle, or one with no status dates for `int_finalsite__status_report_unpivot` to
+unpivot, still means Finalsite knows the student. Scoping it to
+`finalsite_recruitment_year` instead was measured and rejected — it took
+Newark's count from 2 to 8, and all 6 additions were records that exist but sit
+in the 2025 or 2027 cycle, which is a different and murkier problem.
+
+#### The five flags, in plain language
 
 For explaining the worklist to a non-technical audience. Each row is one student
 with one problem; a student with several problems appears once per problem, and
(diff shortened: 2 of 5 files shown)


GabyRangelB:
Verdicts on the three findings. One accepted and fixed, two declined with reasons.

### 1. Naming of the Miami id-crosswalk alias — declined for this PR

The claim checks out. `int_finalsite__contact_id_attributes` does expose both `focus_student_id` and `focus_student_id_prefixed`, so aliasing the prefixed value to `focus_student_id` shadows a real column with a different meaning. The cited precedent is accurate too — `stg_people__student_logins.sql:12-13` casts the prefixed value to `student_number` and the bare one to `student_number_bare`, keeping the distinction visible.

Declining anyway, because the finding stops one model short. `int_tableau__finalsite_student_scaffold.sql:250-256` contains the byte-identical CTE under the same name, and this PR's version was copied from it deliberately so the two read the same. Renaming only the copy in `rpt_tableau__fresh_dashboard_qc` would replace a consistent-but-poorly-named pair with a divergent pair, which is worse for the reader the finding is trying to protect.

The right fix is renaming both, which touches a table-materialized model feeding three dashboard views for a readability gain. That does not belong in this PR. Happy to open a follow-up issue for the lineage-wide rename if you want it tracked.

### 2. Crosswalk join has no width guarantee — accepted, fixed in 321529827

Valid, and the failure mode is worse than described. The finding predicts a fan-out producing more than one worklist row. That case is already caught: both rows would carry a null `finalsite_id`, the same `powerschool_student_number`, `flag_name` and year, so the model's own `unique_combination_of_columns` test fails at `severity: error`.

The uncaught case is the asymmetric one. If a student had two contact records and only one appeared in `stg_finalsite__status_report`, the anti-join drops the matching row and keeps the other. Exactly one row survives, no uniqueness test fires, and the worklist reports a student Finalsite does know as missing. A false positive on a data-quality worklist is the one outcome worth spending a test on.

Added `dbt_utils.unique_combination_of_columns` on `(focus_student_id_prefixed, _dbt_source_project)` in `int_finalsite__contact_id_attributes`, scoped `where focus_student_id_prefixed is not null` — it is null on 27319 of 31927 rows, all outside Miami. Verified against prod-deferred data: `PASS=3 WARN=0 ERROR=0`.

Severity is `warn`, not `error`, on purpose. A second contact record for one student is a plausible Finalsite state rather than corruption, and this model has many consumers. It should surface for a human, not break every build downstream. This also guards `int_tableau__finalsite_student_scaffold`, which makes the same join.

The current numbers, for the record: 4608 non-null rows, 4608 distinct pairs.

### 3. No dbt unit test for the new branch — declined

Agreed on the fact, declining the suggestion. No model under `models/extracts/tableau/` carries a `unit_tests:` block, so adding one here sets a precedent this PR should not set on its own. The branch is covered by the uniqueness test plus the dev-build verification in the PR body, and the test added for finding 2 now covers the crosswalk assumption that a unit test would most usefully have pinned.

### On `sis_school`

Noted, no action. The column stays documented rather than normalized, and the upstream inconsistency is tracked as the separate question it is.

GabyRangelB:
@claude the three findings from your review are answered in the comment above. Finding 2 is fixed in 321529827; findings 1 and 3 are declined with reasoning.

`claude-review` only fires on `opened` / `ready_for_review`, so it has not seen that commit. Please review it, and specifically check these three things:

1. The new `dbt_utils.unique_combination_of_columns` test on `int_finalsite__contact_id_attributes` is scoped `where focus_student_id_prefixed is not null`. Confirm that scope is right — the column is null on 27319 of 31927 rows, all outside Miami, so an unscoped test would compare nulls across four districts rather than assert anything useful.

2. `severity: warn` rather than `error`, argued in the comment above on the grounds that a second Finalsite contact record for one student is plausible rather than corrupt, and this model has many consumers. Push back if you think an invariant a downstream anti-join depends on should hard-fail instead.

3. My claim in finding 2 that the asymmetric case (a student with two contact records, only one of which is in `stg_finalsite__status_report`) escapes `rpt_tableau__fresh_dashboard_qc`'s own uniqueness test because exactly one row survives the anti-join. That reasoning is what justifies the new test, so it is the part worth attacking.

Also worth a second opinion on my decline of finding 1. I declined because `int_tableau__finalsite_student_scaffold.sql:250-256` holds the byte-identical CTE under the same name, so renaming only the copy here trades a consistent bad name for an inconsistent one. Say so if you think the rename is worth doing in both models now rather than as a follow-up.

anthonygwalters (approved):

Issue #5174, opened 2026-09-07
feat(tableau): support the roles and meetings named in the HS GPA monitoring protocol


anthonygwalters:
## What and why

The `HS GPA Monitoring Strategy 2026-27` protocol gives each high school role a weekly or monthly job and names the data behind it. Ten roles and five recurring meetings each have a "what gets monitored" list. We walked that document against the two dashboards it points at — the Academic & Gradebook Health Suite and the gradebook fidelity draft — and asked one question per line: can the person named open something and find it?

Most of it lands. Someone can find a grade level's distribution, a school's progress against its goal, a cusp list, and the per-student "what grade does this student need next" columns the protocol's teacher-commitment example calls for. Cumulative unweighted progress for the Monthly Match Meeting has a purpose-built tab. On the gradebook side, the school rollup drills to a teacher panel and on to a per-teacher view with assignment counts against expectation, which covers the gradebook-health responsibilities.

The gaps cluster in a few places, and they are the reason for this issue. Some are a missing control or an unplaced sheet. Some need a field the models already carry to be put on screen. A few are goals or figures that need a decision from the protocol owner before anything on a dashboard can change. This issue lists them so they can be prioritised together rather than found one at a time by the people trying to run their meetings.

A wayfinding map for the roles and meetings that are already served is local at `.claude/scratch/gpa-protocol-wayfinding.md`. It stays local.

## Proposed approach

Grouped by what each would take. Nothing here is a decision I should make alone.

### 1. Getting a roster off the screen

This is the largest open item and the one worth settling first.

A lot of the protocol ends in a list leaving the dashboard. Grade Level Chairs pull cusp data into a grade level meeting. GLC O3 generates Office Hours lists. Office Hours is meant to produce a Deans List attendance roster. Campus Ops distributes progress reports every two weeks. Teachers need the students in their own sections.

The lists themselves exist and read well — the Grades Roster, the Office Hours roster, and the teacher panel on the gradebook rollup. What is not established is whether the people named in those responsibilities can get any of them off the screen, and in what shape. That question cannot be answered from the workbooks; it needs checking in a browser as each role. Everything downstream of "pull the cusp list" depends on the answer, so it is worth resolving before the smaller items below.

### 2. Attendance beside GPA

[...]

Issue #5169, opened 2026-09-07
feat(tableau): settle the remaining open decisions on the Academic Health tabs


anthonygwalters:
## What and why

We are tuning the `Academic Health Home` and `Academic Health Schools` tabs of the Academic & Gradebook Health Suite ahead of wider use. The work so far sits in a review copy and has not gone to production.

Four items are left. Three are decisions rather than mechanical changes, so each wants a call from the dashboard owner. The fourth is a hand fix the owner is taking. This issue tracks them.

Already settled and in the review copy, for context:

- Year, marking period and region now run on the same parameters across both tabs, so one control moves every sheet it is able to move.
- The headline tiles compare like-for-like student cohorts, and hold a comparison back when there is not yet enough gradebook history behind it. A one-week comparison was added alongside the existing two-week line.
- Goal status colours now match the Cumulative GPA Monitor, and the school bars moved to a goal-neutral brand colour now that school-level goals are not being set this year.
- Every sheet on the Home tab has a purpose-written tooltip, including a grade-level breakdown embedded in the course-failures heatmap.
- The `Academic Health Home_backup` dashboard is removed.

### 1. What the Home-to-Schools click should carry

Clicking the course-failures heatmap on the Home tab opens the Schools tab filtered to both the school and the subject that was clicked. The subject half is not shown on any card, so someone arriving that way sees a narrower roster than they might expect, with nothing on screen explaining why.

Options:

1. Add a `Subject` filter card to the Schools tab, so the carried subject is visible and can be cleared.
2. Narrow the action to carry region and school only.
3. Leave it as it is.

My recommendation is 1 if the subject context is useful to the person clicking through, and 2 if it is not. 3 is reasonable only if click-through from the heatmap is rare.

### 2. Which letter grade the Pop out action should carry

The `Pop out` action on the teacher grade distribution carries the marking-period letter grade into `Y1 Schools - Grades Roster`, which displays the year-to-date letter. Those two are the same value while Q1 is the whole year so far, and they separate once Q2 opens.

Options:

1. Carry the year-to-date letter, so the action matches what the roster shows.
2. Show the marking-period letter on the roster, so the roster matches the action.
3. Leave it as it is.

My recommendation is 1, because the roster is a year-to-date view and its other columns already read that way.

### 3. Whether to drop an unused embedded data source

[...]

PR #5175, opened 2026-09-07
feat(dbt): add a per-student course priority rank for the office hours roster

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will add one column, `office_hours_priority_rank`, to `rpt_tableau__student_course_grades`. It ranks each of a student's courses within a term, where 1 is their lowest percent grade.

The office hours strategy asks for a student list that can be handed to DeansList, and that list has to cap how many teachers one student is sent to in a week. Filtering to `rank <= N` does that in one step.

The dashboard cannot do this ranking itself. Tableau can only rank a student's courses when the course is on the sheet, and anything on the sheet also lands in the export — so a clean two-column export and a Tableau-side ranking cannot both exist. A plain column can be filtered on any sheet no matter how that sheet is laid out.

Nothing reads the column yet. The dashboard work that will use it is a separate change, so merging this on its own changes no number anyone currently sees.

## Reviewer Notes

Three judgement calls worth a look:

- **It hands out unique ranks, where the dashboard's own calculation lets ties share one.** That is deliberate. With shared ranks, a "top 4" filter can return five or six courses when grades tie, which would quietly change how many sessions a student is sent to. Ties break on course number so the choice is the same every rebuild.
- **A course with no grade posted gets no rank at all**, rather than being ranked last. This matches what the roster already does — it drops those rows today — and it stops a course with nothing graded from looking like a student's most urgent one.
- **It ranks on the marking-period grade, not the year-to-date grade.** People are already pulling their lists off the marking-period grade, and while the first quarter is the whole year so far the two are identical anyway. They separate when the second quarter opens. That decision is tracked on #5169.

### General

### dbt

- [x] Properties file updated — the new column is documented in
      `properties/rpt_tableau__student_course_grades.yml`.
- [x] Exposure — `academic_gradebook_health_suite` already exists and already
      depends on this model, so adding a column needs no exposure change.
- [x] **Not a breaking change** — this adds a column, removes and renames
      nothing, and this model does not enforce a contract. No rollback plan
      needed beyond reverting the commit.
      is required.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The column:

```sql
row_number() over (
    partition by _dbt_source_project, studentid, yearid, `quarter`
    order by quarter_course_percent_grade asc, course_number asc
)
```

computed in a `course_priority` CTE over `quarter_grades` filtered to
`quarter_course_percent_grade is not null`, then LEFT JOINed on the same keys as
the existing `qg` join, so an ungraded course-section receives a null rank from
the join rather than a rank of its own.

`_dbt_source_project` is in the partition because ids collide across districts —
the same reason commit `7407d9088` gives for putting it in the grade-scale join.
The ordering follows the `category_ranked` idiom already in this model, including
its stated reason for a final tiebreaker ("so the pick is reproducible across
rebuilds"); the `(x is null) asc` guard that idiom carries is unnecessary here
because the WHERE removes nulls before the window runs.

On why this is not a copy of the dashboard calculation. The workbook holds three
fields captioned `Teacher Priority`, all with the formula
`RANK(MIN([quarter_course_percent_grade]), 'asc')`, and they do not agree:

- `Teacher Priority (copy)` — its `<column>` says `ordering-type='Rows'`, but its
  `<column-instance>` on `Y1 - Office Hours Roster` overrides that with
  `ordering-type='Field'` over course_name / section_or_period / teacher_name.
  That makes it per-student, and it sits on Rows just above `course_name`. This
  is the one this column reproduces.
- `Teacher Priority` — `ordering-type='Rows'`, so it ranks down the whole table.
  It reaches both Office Hours sheets only as a member of the `:Measure Names`
  filter, on no shelf and no encoding, so nothing displays it.
- `Teacher Priority (label)` — the "see first" / "backup" strings. Reads the
  whole-table field and appears on no sheet at all.

Also note the sheet's visible order comes from two `<computed-sort>` entries on
`course_name` and `student_number`, both ascending by
`MIN(quarter_course_percent_grade)` — the rank dimension groups, the computed
sorts order. So this column reproduces the intent rather than any single
mechanism, and `RANK` being competition-ranked is what the `row_number` choice
above departs from.

The term partition matters because `quarter_course_percent_grade` is term-scoped
rather than fixed: the current-year quarter branch aliases
`term_percent_grade_adjusted` into it, the current-year Y1 branch aliases
`y1_percent_grade_adjusted`, and the prior-year branch takes the stored percent
for whichever storecode the row carries. So on a Y1 row this ranks year-to-date
grades and on a quarter row it ranks that quarter's, which is the behaviour the
marking-period control expects.

Verification run before committing:

- Rank shape against prod for `academic_year = 2026`: 10,576 student-term
  partitions, 0 partitions with a duplicate rank, 0 not starting at 1, 0
  non-contiguous, max 12 courses in a partition.
- `dbt build --empty --defer --favor-state` on the model: `PASS=2 WARN=0
  ERROR=0`, so BigQuery accepted the CTE, the window function and the join, and
  the existing `unique_combination_of_columns` test still runs.
- `trunk check --force --no-fix` on both changed files: no issues.

Not verified, and not verifiable from here: whether the export that will consume
this column can actually be downloaded by the roles that need it. That is the
open question in #5174 and it gates the dashboard-side work, not this column.

</details>

### DIFF
--- a/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
+++ b/src/dbt/kipptaf/models/extracts/tableau/properties/rpt_tableau__student_course_grades.yml
@@ -548,6 +548,30 @@ models:
           no letter ladder. The second case does NOT require need_60 to be null
           — a student at 95 percent has a perfectly good need_60 and still gets
           a null need_next, because there is nothing above an A to need.
+      - name: office_hours_priority_rank
+        data_type: int64
+        description: >-
+          Rank of this course among the student's courses for the same term, 1
+          being the lowest percent grade. Intended for an intervention or
+          office-hours list that caps how many teachers one student is sent to:
+          filter to rank <= N.
+
+          Null whenever quarter_course_percent_grade is null. Ungraded courses
+          are excluded from the ranking rather than sorted last, so a course
+          with nothing posted yet can never look like a student's most urgent
+          one.
+
+          Unique within the partition rather than tie-preserving. Two courses on
+          the same percent take consecutive ranks, broken by course_number so
+          the pick is reproducible across rebuilds. A tie-preserving rank would
+          let a "top 4" filter return five or more courses, which would change
+          how many sessions a student is sent to for a reason invisible to
+          whoever reads the list.
+
+          Partitioned on the term, so a Y1 row ranks the student's year-to-date
+          course grades and a quarter row ranks that quarter's. That follows
+          quarter_course_percent_grade, whose own meaning already changes with
+          the row's term.
       - name: category_name_code
         data_type: string
         description: Gradebook category code (e.g. F/S/W; NULL on Y1 rows).
--- a/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__student_course_grades.sql
+++ b/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__student_course_grades.sql
@@ -863,9 +863,10 @@ with
     ),
 
     course_priority as (
-        /* Ungraded courses are filtered out rather than sorted last so they
-           receive no rank at all; the left join below then leaves the rank
-           null. */
+        /* No (x is null) asc guard, unlike category_ranked above — the
+           filter removes nulls before the window runs, so an ungraded course
+           takes no rank and the left join at the foot of the model is what
+           nulls the column. */
         select
             _dbt_source_project,
             studentid,
--- a/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__student_course_grades.sql
+++ b/src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__student_course_grades.sql
@@ -860,6 +860,25 @@ with
         from category_ranked
         where rn_latest_term = 1
         group by _dbt_source_project, studentid, yearid, sectionid
+    ),
+
+    course_priority as (
+        /* Ungraded courses are filtered out rather than sorted last so they
+           receive no rank at all; the left join below then leaves the rank
+           null. */
+        select
+            _dbt_source_project,
+            studentid,
+            yearid,
+            `quarter`,
+            course_number,
+
+            row_number() over (
+                partition by _dbt_source_project, studentid, yearid, `quarter`
+                order by quarter_course_percent_grade asc, course_number asc
+            ) as office_hours_priority_rank,
+        from quarter_grades
+        where quarter_course_percent_grade is not null
     )
 
 select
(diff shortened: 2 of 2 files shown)


anthonygwalters:
Verdicts on the review, one per finding. One fix pushed in `2c2bb8b42`; the rest are no-change with reasons.

**1 — inline comment duplicates the column description. Accepted, fixed in `2c2bb8b42`.**

The counter-argument in the finding was the right half of this. Both `.claude/rules/dbt-sql.md` and `.claude/rules/dbt-yaml.md:398-401` explicitly permit an inline comment for "why a filter exists," so the comment was not a violation wholesale. What was duplicative was its first clause, which restated the `office_hours_priority_rank` description. The replacement keeps only what a reader of that `WHERE` cannot see: that the `(x is null) asc` guard `category_ranked` uses sixty lines above is deliberately absent here because the filter removes nulls before the window runs, and that the null in the column arrives from the left join at the foot of the model. Comment-only, no change to the compiled SQL.

**2 — rank repeats across gradebook-category rows. Accepted as a note, with a correction to the grain, and already handled downstream.**

The repetition is real, but it is marking-period dependent rather than a property of the model. Verified against current prod for `academic_year = 2026`: at `quarter = 'Q1'` there are 174,970 rows over 57,255 distinct student-courses, so 3.06 rows per course and the rank does repeat per category. At `quarter = 'Y1'` there are 57,255 rows over 57,255 student-courses — exactly one row each, because `category_name_code` is null on every Y1 row.

That matters because the office-hours roster runs at Y1 by default, so the grain it reads is already one row per course and there is nothing to de-duplicate. The planned export also places only two dimensions on its sheet, which collapses to distinct combinations regardless of source grain. Keeping the note on the follow-up anyway, since a quarter-scoped export would need it.

**3, 4 and 6 — confirmed, no action.**

I re-checked the two claims in finding 4 that would have been expensive to get wrong: the ST09 predicate order against the existing `qg` join, and that omitting the `(x is null) asc` guard is correct here because the `WHERE` precedes the window.

**5 — conclusion agreed, but the cited rule is misquoted.**

`.claude/rules/dbt-yaml.md:281` reads "Never add `not_null` to a column that cannot be NULL by construction" — that is, do not test a column that can never be null, because the test can never fail and still costs a full scan. The finding renders it as "cannot be non-null by construction," which inverts the meaning. The conclusion is still correct, for the simpler reason that a `not_null` test on a legitimately nullable column would fail outright. Flagging it so the inverted form does not get picked up from this thread later.

**7 — data-dependent verification. Done, re-run against current prod.**

5,288 partitions per marking period (10,576 across Q1 and Y1), 0 with a duplicate rank, 0 not starting at 1, 0 non-contiguous, and at most 12 courses in a partition. Matches the pre-commit figures. I also pulled the dbt Cloud run for this PR's head commit with `warning_only=true`: zero warnings of any kind, including test and freshness.

One note for whoever picks this up: `claude-code-review.yaml` triggers on `opened` and `ready_for_review` only, so the fix push above will not draw a second review pass. That is expected rather than an oversight, and the pushed change does not alter the compiled SQL.

PR #5187, opened 2026-09-08
fix(k8s): drop the dagster-run PriorityClass so run pods stop preempting code servers

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." remove the `dagster-run` PriorityClass so run and step pods sit at the same priority as code servers and can no longer preempt them.

On Friday 2026-09-04 a run pod preempted the `kippcamden` code server 2 minutes after it loaded. The agent re-uploaded the location's metadata while the replacement pod was still pending. That one gRPC `UNAVAILABLE` wrote `ERROR` to the Dagster+ control plane. The agent never retries a location the control plane already marks as errored, so `kippcamden` stayed down until a manual redeploy on Tuesday.

With both pod types at priority 0, a run pod that finds no room waits for a new node instead of killing a code server. The original reason for priority 1000 was kubelet eviction order under memory pressure. That never applied: the `required` anti-affinity already keeps run pods and code servers off the same node.

`helm upgrade` is manual. The change takes effect only after someone runs it.

## Reviewer Notes

- The `dagster-agent` PriorityClass stays at 1000. The agent is pinned to amd64 nodes and everything else to arm64, so it never competes with run pods. Its description text is updated to say so.
- The test in `tests/test_k8s_config.py` now asserts that neither `serverK8sConfig` nor `runK8sConfig` sets a `priorityClassName`, and that `dagster-agent` is the only PriorityClass manifest.
- Skill and CLAUDE.md text that described run-pod preemption of code servers as by design is rewritten. A code server `Preempted` event now means a GKE system-critical pod, or a reintroduced priority gap.

### General

No code location changes. `uv run pytest tests/test_k8s_config.py` passes (11 tests).

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Timeline from agent container logs and `k8s_pod` events, all 2026-09-04 UTC:

- 17:10:49 `kippcamden-prod-18ff21` loaded, "Updated successfully".
- 17:12:43 pod `kippcamden-prod-18ff21-8575cbccc9-bbpkt` `Preempted by pod 4a1636f9-...` on node `...-x242`. Replacement pod `...-9dmr2` hit `FailedScheduling` (Insufficient cpu/memory) until 17:13:56.
- 17:13:50 agent reconcile had `kippcamden` in its "To upload" set, called `list_repositories`, got `DagsterUserCodeUnreachableError: gRPC Error code: UNAVAILABLE`, and wrote error data to the control plane. `_update_location_data` in `dagster_cloud/workspace/user_code_launcher/user_code_launcher.py` has no retry around that call.
- 17:13:53 health check warning: "If it continues failing for more than 300 seconds, a replacement code server will be deployed." Pod was back at 17:14:03, so the replacement path never fired.
- `_should_trigger_recovery` returns `False` when the location is in `control_plane_error_locations`, so the agent never re-uploaded. Recovered only at the 2026-09-08 14:22 UTC redeploy.

Not changed: `podFailurePolicy`, `safe-to-evict` annotations, anti-affinity, the agent PriorityClass. Specs under `docs/superpowers/specs/` that describe the old priority scheme are left as history. The Phase 3 watchdog in `2026-08-18-code-server-scheduling-resilience-design.md` is still the durable fix for the terminal `ERROR` state; its error-signature filter needs widening to include `DagsterUserCodeUnreachableError` when built.

</details>

### DIFF
--- a/.claude/skills/dagster-day2/SKILL.md
+++ b/.claude/skills/dagster-day2/SKILL.md
@@ -165,13 +165,13 @@ Step 8 churn: at replicas=1 with maxSurge=200%, normal Helm upgrade = 2 creates.
 3-4 creates = a rollout retry; >>4 = sustained storm.
 
 In this codebase, agent pods run at priority 1000 (`dagster-agent`
-PriorityClass) — same tier as run/step pods, so they CANNOT be preempted by
-them. If you see "no agents have recently heartbeated" AND the agent pod shows
-Preempted events, the preemption would have to come from a pod at priority >1000
-(system-cluster-critical, etc.) — which is rare and worth investigating. If
-agents are RUNNING with heartbeats but there's agent-level ReadTimeout to
-`*.agent.dagster.cloud`, that's a control-plane connectivity issue, not
-pod-level preemption.
+PriorityClass) — above run/step and code server pods (0), so they CANNOT be
+preempted by them. If you see "no agents have recently heartbeated" AND the
+agent pod shows Preempted events, the preemption would have to come from a pod
+at priority >1000 (system-cluster-critical, etc.) — which is rare and worth
+investigating. If agents are RUNNING with heartbeats but there's agent-level
+ReadTimeout to `*.agent.dagster.cloud`, that's a control-plane connectivity
+issue, not pod-level preemption.
 
 **Timeline table** (ET): run failures, tick failures, terminal schedule tick
 failures, location load failures, agent errors, code server failures, unhealthy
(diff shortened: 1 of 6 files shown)


cbini:
Verdicts on the claude-review findings, applied in 89a2ba1176:

1. **Double `_helm_values()` call in the priority test** — valid. The test now parses the values once and reads `workspace` and `extraManifests` from the same dict.
2. **Tense in the `#4921` measurement bullet of `dagster-k8s-ops`** — valid. "preempt" is now "preempted" so the sentence reads as history.
3. through 6. Informational, no change.

PR #5199, opened 2026-09-08
fix(grow): retry a body truncated mid-stream on idempotent requests

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will retry a Grow API request whose response body is cut off mid-stream, the same way it already retries a reset connection or a timeout.

Dagster run `2f03caff-dd90-4ea8-8d21-e4d8b25ee40b` failed on the `observations` step because Grow dropped the connection 1771 bytes before the end of a page. `requests` raised `ChunkedEncodingError`. That class is not a `ConnectionError` or `Timeout`, so `GrowResource._request` did not mark it transient and the tenacity retry never fired. Dagster's run-level auto-retry recovered the partition 1 minute later, so no data was lost.

The fix adds `ChunkedEncodingError` to the transient tuple in `_request`. POST stays excluded by the existing `transient=method != "POST"` guard. One regression test asserts 3 attempts on a truncated body.

## Reviewer Notes

The other `requests`-based resources (coupa, tableau, finalsite, zendesk, adp, knowbe4) have the same gap. Left for a separate change.

### General

- [x] Ran `uv run pytest tests/resources/test_resource_grow_retry.py` — 9 passed

## CI checks

- [ ] **Trunk** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Diagnosed via `get_run_logs` error chain bottom: `urllib3.exceptions.IncompleteRead: IncompleteRead(545846 bytes read, 1771 more expected)` raised from `requests/models.py` `content` → `ChunkedEncodingError`, caught nowhere in `_request`. Auto-retry run `2658a290-e6ab-43cb-ad07-ca1cf827ce20` succeeded. `ChunkedEncodingError` inherits directly from `RequestException`, not from `ConnectionError`.

</details>

### DIFF
--- a/src/teamster/libraries/level_data/grow/resources.py
+++ b/src/teamster/libraries/level_data/grow/resources.py
@@ -6,6 +6,7 @@ from dagster_shared import check
 from oauthlib.oauth2 import BackendApplicationClient
 from pydantic import PrivateAttr
 from requests import Response, Session
+from requests.exceptions import ChunkedEncodingError
 from requests.exceptions import ConnectionError as RequestsConnectionError
 from requests.exceptions import HTTPError, Timeout
 from requests_oauthlib import OAuth2Session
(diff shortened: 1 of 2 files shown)


cbini:

1. **Docstring line width** — already fixed. The review ran against the first push; the amended head `16dfcea8d7` rewraps that paragraph and no line in the file exceeds 88 chars.

2. **Symmetric POST test for `ChunkedEncodingError`** — declined. The three exception types share one `transient=method != "POST"` expression in `_request`, and `test_connection_error_on_post_does_not_retry` already locks that guard in. A second test on the same line would not catch a regression the first one misses.

3. **`_get_access_token()` lacks retry** — agreed, pre-existing and out of scope. Same bucket as the other six `requests`-based resources noted in Reviewer Notes.

4–5. No action.

PR #5200, opened 2026-09-08
fix(tableau): probe the fresh PAT session and re-sign-in on 401002

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." stop Tableau workbook refresh runs from failing when Tableau rejects a session it issued half a second earlier.

Three prod runs this week (`afa747ef`, `1de7236e`, `dd451efd`) signed in with the PAT successfully, then had the very next request rejected with `401002 Unauthorized Access` 0.2 to 0.4 seconds later. The Dagster auto-retry, which is a full new sign-in, recovered every time. This change does that recovery inside the resource: after sign-in, `_sign_in` probes the session with a `users.get_by_id` self-lookup. If the probe is rejected, it raises `StaleSessionError`, which the existing tenacity loop retries with a fresh sign-in. The op never sees the stale token, so no failed run and no alert.

## Reviewer Notes

- The fix lives in the resource, not the op, so all 3 Tableau assets (workbook refresh, workbook stats, gradebook group sync) get it. Do not add a per-op 401 retry on top.
- A real bad PAT still fails fast. `FailedSignInError` from `auth.sign_in` itself stays outside the retry predicate; only the post-sign-in probe maps to `StaleSessionError`.
- Root cause on the Tableau side is unconfirmed. The pattern (first request only, sub-second, ~3% of runs, retry always works, no other Dagster Tableau step running) points at a server-side session race. Whether it is that or another client sharing the PAT, a re-sign-in is the same recovery the auto-retry already performs.

### General

- [ ] Run `uv run pytest tests/test_dagster_definitions.py` for any modified code location
- [x] New integrations follow the [Library + Config pattern](https://teamschools.github.io/teamster/reference/adding-an-integration/) with the correct asset key format and IO manager (n/a, no new integration)

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Diagnosed via `superpowers:systematic-debugging` from run `afa747ef-551d-4a30-8a2a-255fbee7f2e4`.

Evidence: `ResourceInitSuccessEvent` at 17:05:55.153 UTC (sign-in ok), `ExecutionStepStartEvent` at 17:05:55.359, `ExecutionStepFailureEvent` at 17:05:55.601 with `FailedSignInError 401002` raised from `workbooks.get_by_id` in `libraries/tableau/assets.py:41`. Same shape for `1de7236e` (2026-09-06 20:07:09 UTC) and `dd451efd` (2026-09-07 10:14:56 UTC). In all 3 ticks the `tableau_pat_session_limit` pool serialized Tableau steps; the nearest other Tableau step ended 30 s or more before and started 25 s or more after. Tableau docs state a second sign-in with the same PAT terminates the earlier session, but an external sign-in landing inside a 0.5 s window 3 times in 3 days is implausible; a multi-node session propagation race fits better. Tableau Server logs were not available to confirm.

Verification: 6 offline unit tests in `tests/resources/test_resource_tableau.py` pass (5 existing plus `test_sign_in_re_signs_in_when_fresh_session_is_rejected`). A throwaway live pytest ran `setup_for_execution` against real Tableau with the probe in place and passed (`is_signed_in()` true), then was deleted. `trunk check --force` clean on all 3 files. `dagster definitions validate` and `tests/test_dagster_definitions.py` were not run; the change is a 5-line addition inside `_sign_in` plus one import, exercised by the unit tests.

Probe choice: `users.get_by_id(server.user_id)` needs no admin role (any user may query their own record) and is cheap. `sites.get_by_id` was rejected because Query Site needs site admin.

</details>

### DIFF
--- a/src/teamster/libraries/tableau/CLAUDE.md
+++ b/src/teamster/libraries/tableau/CLAUDE.md
@@ -26,3 +26,9 @@ Asset metadata, kinds, and label come from the workbook's dbt
 
 Authenticates to Tableau Server via Personal Access Token (PAT). Exposes
 `_server` (a `tableauserverclient.Server` instance) for workbook operations.
+
+`_sign_in` probes the new session with a `users.get_by_id` self-lookup and
+re-signs-in (inside the existing tenacity loop) when Tableau rejects it with
+`401002`. Tableau intermittently rejects a token it issued under a second
+earlier; a fresh sign-in always recovers. Do not add a per-op 401 retry: this is
+the single fix point for all three Tableau assets.
--- a/src/teamster/libraries/tableau/resources.py
+++ b/src/teamster/libraries/tableau/resources.py
@@ -3,7 +3,10 @@ from pydantic import PrivateAttr
 from requests.exceptions import ConnectionError as RequestsConnectionError
 from requests.exceptions import Timeout
 from tableauserverclient.models.tableau_auth import PersonalAccessTokenAuth
-from tableauserverclient.server.endpoint.exceptions import InternalServerError
+from tableauserverclient.server.endpoint.exceptions import (
+    FailedSignInError,
+    InternalServerError,
+)
 from tableauserverclient.server.server import Server
 from tenacity import (
     retry,
@@ -11,6 +14,10 @@ from tenacity import (
     stop_after_attempt,
     wait_exponential_jitter,
 )
+
+
+class StaleSessionError(Exception):
+    """Tableau rejected the session it issued a moment earlier."""
 
 
 class TableauServerResource(ConfigurableResource):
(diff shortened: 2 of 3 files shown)


cbini:

1. **Missing exhaustion test for the stale-session path — accepted.** Added `test_sign_in_exhausts_on_persistent_stale_session` in `8721eb02cd`. A probe that stays rejected raises `StaleSessionError` after 5 sign-ins, mirroring the persistent-connection-error test.

2. **No test for a transient error raised by the probe — declined.** tenacity's `retry_if_exception_type` matches on the exception class wherever it is raised inside `_sign_in`, so a `ConnectionError` from the probe takes the same path as one from `auth.sign_in`, and `test_sign_in_retries_on_network_faults` already proves that predicate. A second test would pin an implementation detail of one code path, not a distinct behavior.

PR #5204, opened 2026-09-09
fix(grow): keep coaches off the Teachers fallback observation group

cristinabaldor:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop every coach from also observing their school's `Teachers` fallback group, and stop school leaders from being observed at all.

PR #5078 gave each coach their own observation group. It left the `Teachers` group in place as the fallback for teachers whose own manager cannot anchor a group, and it left every observer-capable person at the school on that group as an observer. So a coach saw two groups: their own reports, plus a wider group holding teachers they do not coach.

Three changes ship together.

**Only a teacher is an observee.** A school admin or assistant admin holds a leadership role, not a coached teaching assignment. 171 of them stop being observed.

**Only a school's admins and assistant admins observe the fallback.** That drops the fallback's observers from 235 to 165. Every other coach reaches their own reports through their own group.

**The fallback's observers are emptied when it holds no observees.** An empty group with observers still shows up as a group in each of their pickers. 24 of the 28 schools now have an empty fallback.

The fallback itself survives. It holds the 9 teachers whose own manager cannot anchor a group, down from 26.

## Reviewer Notes

**Leaders who coach only other leaders now anchor no group.** Coach groups drop from 174 to 150, because 24 coaches had no reports left once admins and assistant admins stopped being observees. A principal who coaches only their assistant principals can no longer observe them in Grow. That follows directly from the decision that leaders are not observees, but it is worth an eyeball before the first sync run.

**165 people still observe a fallback, and 122 of them also anchor their own coach group.** Restricting the fallback to admins and assistant admins does not remove every duplicate group — most coaches hold `School Assistant Admin`. Where their school's fallback is empty the group carries nobody, which is the case at 24 of 28 schools. The four exceptions are Room 9, KIPP Newark Collegiate Academy, KIPP Cooper Norcross High and KIPP Newark Lab High School.

**Nobody observes an empty fallback, so a newly-uncoached teacher is invisible until the next sync.** The sync runs daily, and the group is re-populated with its observers the moment it has an observee again.

### General

### Dagster

`dagster definitions validate` cannot run in this Codespace — the `kipptaf` location fails to import on unset Illuminate dlt settings, which is documented behaviour. Verified instead with `py_compile` plus the pure-helper unit tests. Needs a real check in CI.

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — no new models; `group_type`'s description and two unit tests are updated in place.
- [x] **Breaking change?** No. No column is renamed, removed, or retyped. `group_type` changes values, not its shape.

## CI checks

- [ ] **Trunk** — `trunk check --force` passes locally on all four changed files.
- [ ] **dbt Cloud** — `dbt build --select rpt_schoolmint_grow__users` passes locally: 14 nodes, 0 errors, 0 warnings. All 7 unit tests pass.
- [ ] **Dagster Cloud** — expected to run, since a code location changed.

<details>
<summary>For Claude</summary>

Reported by a school assistant admin at KIPP Royalty Academy who saw two groups in her picker: her own coach group, and the school's `Teachers` group with her listed as a `Coach` on it.

The original report was that the second group let her see everyone. It did not: measured against the extract, KIPP Royalty Academy's fallback held 1 observee, not 42. The duplicate group was real; its width was not. The live group membership could not be read back to confirm — the 1Password token in the working session was expired, and the `schools` ingest snapshot in BigQuery predates the last sync by an hour.

The fallback observer rule was chosen over two alternatives, both rejected by the requester: `School Admin` only (27 observers network-wide, about one per school), and dropping only the coaches who anchor a group. The chosen rule leaves 122 anchoring coaches on a fallback that is empty at 24 of 28 schools, which is why the third change — emptying the observers with the observees — carries most of the fix.

`is_teacher` and the two admin flags are mutually exclusive by construction: `is_teacher` reads `job_function in ('Teacher', 'Teacher in Residence')` while both admin flags read `tier`, which is derived from the same `job_function`. Confirmed against production: zero active staff hold both the `Teacher` role and an admin role, so no teacher loses observee status from this change.

Before-and-after counts, both measured over active users with a resolved Grow account and school, production versus a local dev build of the model:

| measure                | before | after |
| ---------------------- | -----: | ----: |
| observees              |    873 |   702 |
| fallback observees     |     26 |     9 |
| fallback observers     |    235 |   165 |
| coach groups           |    174 |   150 |

</details>

### DIFF
--- a/src/dbt/kipptaf/models/extracts/schoolmint/properties/rpt_schoolmint_grow__users.yml
+++ b/src/dbt/kipptaf/models/extracts/schoolmint/properties/rpt_schoolmint_grow__users.yml
@@ -159,12 +159,13 @@ models:
         data_type: string
         description: >
           Observation-group membership directive consumed by `grow_user_sync`.
-          Observee and observer are independent: `observees` covers `Teacher`,
-          `School Admin` and `School Assistant Admin`, `observers` covers any
-          admin role plus `Regional Observer` and `Coach`, and a user who
-          matches both gets `observees;observers`. A New Teacher Network
-          Coordinator is usually a teacher too, so they get both. Empty string
-          when the user matches neither.
+          Observee and observer are independent: `observees` covers `Teacher`
+          only, `observers` covers any admin role plus `Regional Observer` and
+          `Coach`, and a user who matches both gets `observees;observers`. A New
+          Teacher Network Coordinator is usually a teacher too, so they get
+          both. A school admin or assistant admin observes without being
+          observed -- they hold a leadership role rather than a coached teaching
+          assignment. Empty string when the user matches neither.
       - name: surrogate_key_source
         data_type: string
         description: >
@@ -711,10 +712,9 @@ unit_tests:
 
   - name: unit_grow_users_group_type_additive
     description:
-      Pins observee and observer membership as two independent predicates. A
-      School Assistant Admin who coaches is now both, where the old first-match
-      CASE made every admin an observer only. A Regional Admin is an observer
-      but never an observee. A plain teacher stays an observee.
+      Pins observee and observer membership as two independent predicates. Only
+      a teacher is an observee -- a School Assistant Admin who coaches observes
+      and is never observed, and neither is a Regional Admin.
     model: rpt_schoolmint_grow__users
     given:
       - input: ref('int_people__staff_roster')
@@ -820,7 +820,7 @@ unit_tests:
     expect:
       rows:
         - { user_internal_id: 200, group_type: observers }
-        - { user_internal_id: 201, group_type: observees;observers }
+        - { user_internal_id: 201, group_type: observers }
         - { user_internal_id: 202, group_type: observees }
 
   - name: unit_grow_users_region_scope
@@ -1089,7 +1089,7 @@ unit_tests:
             role_names: [Regional Observer, School Assistant Admin],
             regional_admin_school_ids: [sch-courage],
             readonly: 0,
-            group_type: observees;observers,
+            group_type: observers,
           }
         - {
             user_internal_id: 402,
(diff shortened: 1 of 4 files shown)


cristinabaldor:
## Review verdicts

**1. Test coverage gap — fallback-observer emptying is untested. Accepted, fixed in `329bbe1e0`.**

Verified: no test reached that ternary. My `_observes_fallback` tests cover who observes the fallback, and the two dbt unit tests cover who is an observee, but nothing pinned the emptying rule — which is the change that clears the duplicate group at 24 of the 28 schools.

Extracted it as `_fallback_group(uncoached, observers)` and pinned both branches. That matches the module's three other pure helpers (`_match_observation_group`, `_can_anchor_group`, `_observes_fallback`), all of which are module-level functions tested from `tests/assets/test_grow_observation_groups.py` — so this is the existing pattern, not a new abstraction. 18 tests pass; `trunk check --force` is clean on both files.

**2. Model altitude — agreed, nothing to flag.** Confirmed independently: `grep -l 'local: ../schoolmint' src/dbt/*/packages.yml` returns nothing, and the staging models sit under `src/dbt/kipptaf/models/schoolmint/grow/staging/`. There is no source-system package to push the hash down into.

**3. dbt SQL/YAML conventions — agreed, nothing to flag.**

**4. Ponytail — agreed on the original diff.** Note that finding 1's fix adds a helper, which nets +4 lines of source. The trade is a tested rule for an untested one-liner, and it keeps the group-assembly loop reading as three named rules rather than two names plus an inline conditional.

**5. Environment limitations — noted, and here is what was actually run locally.** `dbt build --select rpt_schoolmint_grow__users`: 14 nodes, 0 errors, 0 warnings. All 7 dbt unit tests pass. dbt Cloud CI now passes with 0 warnings, after the first run failed on a stale staging clone unrelated to this change — the clone was last seeded 2026-09-02 and lacked the `read_only` column that #5078 added on 09-08, so `Clone - Staging (On-Demand)` was re-run before the retry.

One thing neither of us can check from CI: the live Grow group membership. The 1Password token in the working session was expired, so no probe could run against the API.

GabyRangelB (dismissed):

PR #5201, opened 2026-09-08
refactor(dbt): move PowerSchool-only intermediates into the package and rebuild the Miami archive once

cbini:
# Pull Request

> **Write this whole PR in plain language.** Read `.github/PLAIN_LANGUAGE.md`
> and follow it: every sentence does a job, one idea per sentence, active voice,
> common words. Only the "For Claude" fold-out at the bottom is exempt.

## Summary & Motivation

> What changed and why. A reviewer skimming just this section should understand
> the change. Save tradeoffs, edge cases, and verification detail for "Reviewer
> Notes" below.

When merged, this pull request will move 8 PowerSchool-only intermediate models from kipptaf into the shared `powerschool` package, and re-include that package in `kippmiami` with 15 post-hooks so one Dagster build rebuilds Miami's frozen PowerSchool archive with the 8400 student-number prefix and the AY2025 bound applied.

kipptaf keeps its copies of the 8 models. PR 2 (#5012) deletes kipptaf's Miami-only steps and swaps the 8 models for union wrappers. The only kipptaf change here is a 1-line guard in the `focus_student_number` macro so the rebuilt archive is not prefixed twice before PR 2 lands.

Closes #5195. Refs #5012.

## Reviewer Notes

> Optional. Name what's worth a second look and why, in a line or two each. Full
> technical detail, exact values, and edge cases belong in "For Claude" below.
> Delete if there's nothing here.

- The post-hooks in `src/dbt/kippmiami/dbt_project.yml` are `update` and `delete` statements. They run inside `dbt build` on the 15 kippmiami tables being rebuilt in the same run, and on nothing else.
- The 8 moved models dropped their `_dbt_source_relation` and `_dbt_source_project` columns. The PR 2 wrappers restore them.
- 2 moved models derive a column their kipptaf inputs used to carry: `int_powerschool__log` computes `academic_year` inline, and `int_powerschool__gpprogress_grades` derives `is_transfer_grade` from `stg_powerschool__schools` (a grade earned outside the district).
- 23 of the package's ODBC staging models have no frozen file in GCS. Those 23, their 3 consumers, and 2 models with no properties file are disabled in kippmiami. No archive relation reads them.
- `main` already carried a package `int_powerschool__state_assessments_transfer_scores` (#5194) and had removed the `fldoe` source for `kippmiami_powerschool` (#5078). The merge takes main's versions of both.
[...]

PR #5206, opened 2026-09-09
fix(k8s): pin code servers against the autoscaler now that run pods cannot preempt them

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." add `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` to code server pods so the cluster autoscaler stops killing them during a `ScaleDown`.

On 2026-09-09 at 14:40 UTC the autoscaler deleted the `kippmiami` code server pod for a node scale down. The agent was re-uploading the location's metadata at that moment, got one gRPC `UNAVAILABLE`, and wrote `ERROR` to the control plane. The agent never retries that state. It is the same mechanism as the kippcamden outage in #5187, with the autoscaler as the trigger instead of a preemption.

This annotation was tried in #4921 and reverted the same day. Back then run pods sat at priority 1000, so a pinned code server could still be preempted, and that happened 118 times more often than baseline. #5187 put run pods at priority 0, so preemption between the two is impossible now and the annotation only does what it was always meant to do.

`helm upgrade` is manual. The change takes effect only after someone runs it.

## Reviewer Notes

- **Keep-or-revert rule** is written into the values file comment. Judge it at matched run-pod load, 25 or more `Scheduled` events per 15 minutes. Revert if code-server `Preempted` or agent gRPC errors rise above the 0/0 baseline from #4921. Keep if `ScaleDown` kills of code servers reach 0.
- **Cost:** a node holding a code server cannot consolidate, so a run pod that fits nowhere waits for Node Auto Provisioning instead. This is the same wait run pods already take when no node has room.
- The test that asserted the annotation's absence now asserts its presence on all 3 pod types, and its docstring carries the #4921 history so the priority dependency stays visible.

### General

No code location changes. `uv run pytest tests/test_k8s_config.py` passes (11 tests).

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Timeline from agent container logs and `k8s_pod` events, 2026-09-09 UTC:

- 14:36:53 control plane update trigger for kippmiami (merge of #5201). 14:38:49 `Updated successfully`.
- 14:40:06 pod `kippmiami-prod-63995e-6fc7dcfc65-jgjqt` event `ScaleDown: deleting pod for node scale down`. Replacement `...-ssjwt` scheduled the same second, container started 14:40:10, definitions import takes about 90 seconds.
- 14:40:17 agent `Fetching metadata for prod:kippmiami`, then `Unable to update prod:kippmiami ... gRPC Error code: UNAVAILABLE`, then `Updated with error`.
- 14:43:04 recovered by a `reloadRepositoryLocation` from a Dagster MCP session (audit log `authorAgentTokenId` 2951). The agent itself never re-uploaded.

Why the #4921 measurement no longer applies: `_should_trigger_recovery` and the health-check path are unchanged, but the preemption that #4921 measured needed a priority gap. Audit log confirms the `dagster-run` PriorityClass was deleted 2026-09-08 16:16 UTC and run pods created since carry no `priorityClassName`.

Alternative considered: PDB `minAvailable: 1` per location. Same effect, since the autoscaler honors PDBs, but it also blocks GKE node maintenance drains until the 1 hour override. The annotation is narrower. Both are ruled out only under a priority gap, which the test above guards.

Not in scope: the Phase 3 watchdog in `docs/superpowers/specs/2026-08-18-code-server-scheduling-resilience-design.md`, parked by the user. Upstream issue on the agent's missing retry is drafted for dagster-io/dagster.

</details>

### DIFF
--- a/.claude/skills/dagster-k8s-ops/SKILL.md
+++ b/.claude/skills/dagster-k8s-ops/SKILL.md
@@ -46,20 +46,22 @@ description:
 - `safe-to-evict: "false"` only blocks cluster autoscaler evictions — kubelet
   node-pressure evictions (exit 137, OOM) are unaffected. Scale-Out density
   makes these occasional; Dagster retries automatically.
-- **Do NOT add `safe-to-evict: "false"` to code server pods.** Tried 2026-08-18
-  to stop the 38-59 weekly autoscaler relocations, reverted the same day
-  (#4921). Because it blocks only autoscaler eviction and NOT scheduler
-  preemption, pinning a priority-0 pod removes the graceful way to free its node
-  and leaves only the violent one: capacity fragments, then run pods (then
-  priority 1000) preempted code servers to obtain it — and every preemption
-  recreates the Service with a fresh ClusterIP, which is the churn the
-  annotation was meant to reduce. Measured at matched load (~32 run/step pods
-  per 15 min): with it, 18-20 agent gRPC errors and 8-9 `Preempted` per 15 min;
-  without it, 0 and 0 across 12 hours including the nightly wave. **General
-  rule: pinning a low-priority pod against the autoscaler converts graceful
-  relocation into preemption.** The agent and run pods keep the annotation —
-  nothing outranks them the way run pods used to outrank code servers. Absence
-  is asserted in `tests/test_k8s_config.py`.
+- **`safe-to-evict: "false"` on code server pods is safe ONLY at equal priority
+  with run pods.** First tried 2026-08-18 to stop the 38-59 weekly autoscaler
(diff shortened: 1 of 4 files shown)


cbini:

1. **Stale framing in the 2026-08-18 scheduling-resilience spec** — agreed it is stale, no change in this PR. Specs under `docs/superpowers/specs/` are history by repo convention. The spec's Phase 3 section also needs its error filter widened and its gating question marked answered, so it gets one edit when that work is picked up rather than a partial touch now.
2. through 4. Verification and not-applicable notes, no change.

Issue #5213, opened 2026-09-09
perf(powerschool): remove the storedgrades join fan-out in int_powerschool__category_grades


cbini:
## What and why

`int_powerschool__category_grades` is the single most expensive dbt node in prod: 221 slot hours over 7 days, 16 slot minutes per run, 118 runs a day.

The cost is a join fan-out. In the latest prod run, stage `S05: Join+` read 142M rows and wrote 6.9M. Two aggregate stages then collapsed those to 6.1M. The input tables total 0.13 GB. The model does 23 times more work than its output needs.

## Proposed approach

Pre-aggregate `stg_powerschool__storedgrades` to one row per `(studentid, course_number, yearid, storecode)` before the join, or join on `sectionid` where it is available. Then confirm the `deduplicate` step is still needed.

<details>
<summary>For Claude</summary>

> File: `src/dbt/powerschool/models/sis/intermediate/int_powerschool__category_grades.sql`, lines 47 to 58. The `storedgrades` left join keys on `studentid`, `course_number`, `yearid`, `storecode` but not `sectionid`, so a student in several sections of one course multiplies. `dbt_utils.deduplicate` on line 61 hides the fan-out.
>
> Verify with the stage plan before and after: `records_read` on the join stage should drop to roughly the size of `base_powerschool__course_enrollments` times termbins.
>
> Behavior check: row count and `count(distinct format("%T|%T|%T|%T", studentid, yearid, course_number, storecode))` must match prod before and after.
>
> Parent: #5212

</details>

cbini:
Re-ran the stage plan before starting. The join does not fan out: in the latest Newark run the storedgrades join took 6.89M rows in and put 6.89M out. The 142M-row read on that stage is BigQuery broadcasting the 1.3M-row storedgrades table to 100 workers, and the dedupe collapses the enrollment-by-termbin expansion by 13%.

What drives the 221 slot hours is cadence plus the dedupe shape. Newark rebuilt the table 478 times in 7 days (68 a day) on the eager table condition at 26 slot-minutes each. Every consumer is a daily batch (DeansList extract 01:25, Tableau gradebook batch and `int_students__category_grades` 04:00). The `dbt_utils.deduplicate` call compiles to 16 `array_agg` columns; a `row_number` pick over the same keys measured 3.2 slot-minutes against 12.1 with identical output.

[...]

Issue #5212, opened 2026-09-09
perf(dbt): cut warehouse slot time on the 6 nodes that dominate prod dbt cost


cbini:
## What and why

Six dbt nodes account for most of the BigQuery slot time our prod dbt runs spend each week. Three of them do work that grows faster than their data (join fan-out, repeated CTE evaluation, view fan-in). Two rebuild more than they need to. One runs far more often than its data changes.

Fixing them lowers slot contention for every other model in the same runs, and lowers what we pay for BigQuery.

Measured over the 7 days ending 2026-09-09, prod target only, parent script jobs excluded:

| Node | 7-day slot hours | Runs | Rows read per run | Verdict |
|---|---|---|---|---|
| `model.powerschool.int_powerschool__category_grades` | 221 | 828 | 104M read, 4.6M peak | join fan-out |
| `model.kipptaf.int_extracts__student_enrollments` | 152 | 1012 | 90M read, 134k out | view fan-in, 195 stages |
| `model.kipptaf.stg_people__student_logins` | 151 | 600 | 3M read, 31k out | plan bloat, ~420 stages |
| `model.deanslist.stg_deanslist__behavior` | 67 | 54 | 34M read | full-history dedupe each run |
| `relationships` tests on `dim_staff` (5 tests) | 58 | 128 each | 104M read each | recompute a view per test |
| `snapshot_powerschool__gpa_term` + `gpa_cumulative` | 40 | 7220 | small | 143 merges per day each |

## Proposed approach

One child issue per row. Fix the top 3 first. They are the ones where the SQL shape, not the data size, drives the cost.

<details>
<summary>For Claude</summary>

> Ranking query: `region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT` on `teamster-332318`, filtered to `"app": "dbt"` and `"target_name": "prod"` in the query comment, `statement_type not like 'SC%'`, grouped by `node_id`. Per-stage evidence comes from `unnest(job_stages)` on the same table: `records_read`, `records_written`, `slot_ms` per stage.
>
> Re-run the ranking before starting any child. Numbers drift week to week.
>
> Done when each child is closed and the 7-day ranking no longer has any node above 60 slot hours.

</details>

cbini:
**Child 1 of 6 (#5213, `int_powerschool__category_grades`) shipped in #5219, merged 2026-09-09.**

[...]

PR #5238, opened 2026-09-10
perf(people): collapse the 5 self-joins in stg_people__student_logins into one pass

### LINKED ISSUE #5214: perf(people): collapse the 5 self-joins in stg_people__student_logins into one pass
## What and why

`stg_people__student_logins` costs 151 slot hours over 7 days to produce about 31k rows. Each run is 15 slot minutes for 3M rows read. The query plan has about 420 stages, and most are 1.3 slot minute `Coalesce` stages over the same 31,423 rows.

The cost is repeated evaluation. The model left-joins the `username_filter` CTE 5 times, once per username priority. BigQuery evaluates the CTE and its 2 window functions 5 times. Four `not in (select ... from {{ this }})` subqueries add 4 more full reads of the target table.

## Proposed approach

Replace the 5 joins with one `group by student_number` over `username_filter`, picking each priority with `max(if(priority_order = N and rn_username = 1, username, null))`. Replace the `not in` subqueries with one left anti-join against `{{ this }}`.

<details>
<summary>For Claude</summary>

> File: `src/dbt/kipptaf/models/people/staging/stg_people__student_logins.sql`. The 5 joins are lines 143 to 190. The `not in` subqueries are lines 26, 37 to 39, and 128.
>
> The incremental branch is the only one that costs anything. Output must be identical: same `student_number`, `username`, `default_password`, `google_email` for every new row.
>
> Verify: stage count in `job_stages` drops from about 420 to under 50.
>
> Parent: #5212

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make `stg_people__student_logins` build in one pass instead of re-reading its 3 upstream views 5 times.

The incremental branch left-joined the `username_filter` CTE 5 times, once per username priority, and ran 4 `not in (select ... from this)` subqueries. BigQuery inlines a CTE per reference, so it re-evaluated the PowerSchool union view, the Finalsite views, and both window functions for every join. Each run cost about 15 slot-minutes and 100 stages to add almost no rows. That is about 145 slot hours a week.

Now the model reads the target table as a 2-column projection through 3 plain anti-joins instead of 4 correlated subqueries, generates the 5 candidate usernames with one `unnest` over a single scan of the upstream views, and picks each student's username with one `group by`. Output is unchanged.

Closes #5214. Refs #5212.

## Reviewer Notes

- The dev and staging branch of the model reads the archived source, so dbt Cloud CI does not exercise the rewritten SQL. Verification ran against prod as a SELECT (details in the fold-out).
- `username_pick` uses `coalesce(max(if(priority_order = N ...)))` rather than a second `row_number`, per the repo's priority-pick convention. It is equivalent to the old `coalesce(u1..u5)`.

### General

### dbt

- [x] Include (or update) an exposure — no consumer change
- [x] **Breaking change?** No. Column set and values are unchanged.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes on retry. First run hit a transient BigQuery `Error encountered during execution. Retrying may solve the problem.` on 4 downstream Tableau views.

<details>
<summary>For Claude</summary>

Verification: compiled both versions with `--target prod` and ran them as one BigQuery SELECT against a simulated target (`stg_people__student_logins` minus students whose `student_number` ends in 0), so both sides saw the same snapshot and had real work to do. Result: 1,099 rows each, 1,099 distinct `student_number` and `username`, 0 rows only in old, 0 rows only in new, on a full join over all 4 output columns. Against the real target both versions return 0 rows today.

Cost: the new SQL alone ran in 14 stages at 0.5 slot-minutes (job `xsde2O27V9epxNriE6xaCBjXkz0`). Prod runs on the old SQL over the prior 7 days averaged 99 to 100 stages and 9.5 to 21 slot-minutes.

Equivalence notes: the old `username_filter` renumbered `priority_order` with `row_number` after the `not in` filter; the new one keeps the original priority. Both pick the lowest surviving priority with `rn_username = 1`, so the result is the same. `rn_username` in the old code ordered by the input `priority_order` (BigQuery resolves window `order by` against input columns, not select aliases), which is what the new code does too.

`existing_logins` is referenced 3 times (2 anti-joins on `student_number`, 1 on `username`). BigQuery may evaluate it per reference, but it is a 2-column projection of a 31k-row table and the whole query measured 0.5 slot-minutes, so collapsing further is not worth a non-equi `or` join.

Out of scope, noted for the parent: the model runs 76 to 109 times a day on prod. Its data changes daily at most. A cron automation condition like #5213 used would cut the remaining cost further.

</details>

### DIFF
--- a/src/dbt/kipptaf/models/people/staging/stg_people__student_logins.sql
+++ b/src/dbt/kipptaf/models/people/staging/stg_people__student_logins.sql
@@ -3,6 +3,8 @@
     from {{ source("people", "src_people__student_logins") }}
 {% elif is_incremental() %}
     with
+        existing_logins as (select student_number, username, from {{ this }}),
+
         miami_candidates as (
             select
                 c.first_name,
@@ -20,11 +22,16 @@
         ),
 
         union_source as (
-            select student_number, dob, first_name, last_name,
+            select
+                student_number,
+                dob,
+                first_name,
+                last_name,
+
+                cast(null as int64) as student_number_bare,
             from {{ ref("stg_powerschool__students") }}
             where
-                student_number not in (select t.student_number, from {{ this }} as t)
-                and _dbt_source_project != 'kippmiami'
+                _dbt_source_project != 'kippmiami'
                 and dob is not null
                 and first_name is not null
                 and last_name is not null
@@ -32,15 +39,25 @@
 
             union all
 
-            select student_number, birth_date as dob, first_name, last_name,
+            select
+                student_number,
+                birth_date as dob,
+                first_name,
+                last_name,
+                student_number_bare,
             from miami_candidates
             where
-                student_number not in (select t.student_number, from {{ this }} as t)
-                and student_number_bare
-                not in (select t.student_number, from {{ this }} as t)
-                and birth_date is not null
+                birth_date is not null
                 and first_name is not null
                 and last_name is not null
+        ),
+
+        new_students as (
+            select u.student_number, u.dob, u.first_name, u.last_name,
+            from union_source as u
+            left join existing_logins as e1 on u.student_number = e1.student_number
+            left join existing_logins as e2 on u.student_number_bare = e2.student_number
+            where e1.student_number is null and e2.student_number is null
         ),
 
         components as (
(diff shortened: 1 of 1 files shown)


cbini:

1. **`existing_logins` referenced 3x** — count is correct. Fixed the PR body: it now says the target is read as a 2-column projection through 3 anti-joins, not "once". Declining the `or`-join collapse: BigQuery treats `left join ... on a = x or b = x` as a non-equi join (nested loop or rejected outright), and the whole new query measured 0.5 slot-minutes in 14 stages, so there is nothing left to win there.
2. **Function nesting in `username_candidates`** — pre-existing expressions moved verbatim, as the review confirmed. No change in this PR.
3. **No `contains_pii` tag** — pre-existing, and this PR does not touch the properties yml. Leaving for a separate change.
4. to 8. — no action needed.

dbt Cloud: the first run failed on 4 downstream Tableau views with a transient BigQuery `Error encountered during execution. Retrying may solve the problem.` at 15:46 UTC, nothing in this model. Retried as run 70403224232805.

PR #5240, opened 2026-09-10
fix(dagster): give Directory API retries a 63s budget for 429s

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will let the Google Directory API pulls ride out a rate-limit error instead of failing the run.

Run `76ab2239-6d3b-467e-9d6a-66af92ac9116` (the noon `google/directory/users` schedule) failed with `HttpError 429 rateLimitExceeded` on `users.list`. The retry layer already treats 429 as transient, but it used the dagster `backoff` defaults: 4 retries with delays of 0.1, 0.2, 0.4 and 0.8 seconds. That is 1.5 seconds total, all inside the same per-minute window that produced the 429. The Dagster auto-retry succeeded two minutes later, which is exactly what the in-step retry now covers.

The fix is one shared `_backoff` helper with delays of 1, 2, 4, 8, 16 and 32 seconds (63 seconds total). Both retry sites use it: each `_list` page request and the batch envelope in `_execute_batch_with_retry`.

## Reviewer Notes

- The batch sub-request retry loop already used `exponential_delay_generator()` from 0.1s. I left it alone; it is a different failure path and this PR fixes only the one that failed.
- Worst case a truly dead page request now takes 63 seconds longer to fail. The schedule has no `max_runtime` tag, so this does not collide with a runtime cap.

### General

### Dagster

- [x] `uv run pytest tests/resources/test_resource_google_directory.py`: 47 passed. The 12 failures are the live-credential tests and fail identically on `main` in this Codespace.
- [ ] Run `uv run dagster definitions validate` for any modified code location (no definitions changed)

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — not applicable.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Root cause read from the run pod's `RUN_FAILURE` log line in GKE (`dagster-run-76ab2239-*`), because the Dagster event log for this run trips the repo's output-redaction hook and the step pod writes no stdout to k8s logs. Bottom of the error chain: `teamster.libraries.google.directory.resources._TransientHttpError: <HttpError 429 when requesting None returned "Resource has been exhausted (e.g. check quota).". Details: "[{'message': 'Resource has been exhausted (e.g. check quota).', 'domain': 'global', 'reason': 'rateLimitExceeded'}]">`. The same schedule has failed with an auto-retry recovering about every 1 to 3 weeks since February 2026, so this is chronic, not a one-off.

New test `test_list_outlasts_a_sustained_429` fails on `main` (retries exhausted after 5 calls) and passes here (7 calls, total sleep >= 60s).

</details>

### DIFF
--- a/src/teamster/libraries/google/CLAUDE.md
+++ b/src/teamster/libraries/google/CLAUDE.md
@@ -19,7 +19,9 @@ account with domain-wide delegation. Provides batch methods:
 
 **Retry pattern**: All `.execute()` calls (both `_list` and batch methods) are
 wrapped via the module-level `_retryable_execute(request)` factory +
-`backoff(fn=..., retry_on=(_TransientHttpError,))`. Never use bare
+`_backoff(fn)`, which retries `_TransientHttpError` 6 times with delays of 1, 2,
+4, 8, 16, 32s (63s total, enough to outlast a per-minute 429). Do not call
+`dagster.backoff` directly: its defaults give up after 1.5s. Never use bare
 `errors.HttpError` — 4xx client errors must not be retried. Transient codes:
 `{429, 500, 502, 503, 504}`.
 
--- a/src/teamster/libraries/google/directory/resources.py
+++ b/src/teamster/libraries/google/directory/resources.py
@@ -37,6 +37,15 @@ def _backoff(fn: Callable[[], dict]) -> dict:
     ``dagster.backoff``'s defaults (4 retries, delays from 0.1s) give up after
     1.5s, inside the same rate-limit window that produced the 429. Delays of
     1, 2, 4, 8, 16, 32s total 63s, past a per-minute ``rateLimitExceeded``.
+
+    Args:
+        fn: Zero-arg callable from :func:`_retryable_execute`.
+
+    Returns:
+        The response dict from the first successful call.
+
+    Raises:
+        _TransientHttpError: If all 7 attempts fail.
     """
     return backoff(
         fn=fn,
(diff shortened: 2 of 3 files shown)


cbini:
Per-finding verdicts on the claude-review comment (62f1571 addresses the accepted ones):

1. **Batch-envelope call site has no test** — accepted. Added `test_batch_envelope_outlasts_a_sustained_429`: the mocked batch's `execute()` raises 429 six times, then succeeds. 7 calls, total sleep >= 60s.

2. **Possible jitter in `exponential_delay_generator`** — declined. Read the source in the installed dagster: `i = base_delay; while True: yield i; i = i * backoff_factor`. No jitter parameter, no randomness, so the 63s figure is exact and the tests are deterministic.

3. **`libraries/google/CLAUDE.md` retry pattern stale** — accepted. It now names `_backoff`, the 1..32s delays, and says not to call `dagster.backoff` directly.

4. **Batch worst case compounds to ~5×63s** — correct as stated, no change. The batch write paths run on the 1am and 3am schedules with no `max_runtime` tag, and a batch that returns 429 on the envelope five times in a row is the case where waiting is the right behavior.

5. **`_backoff` docstring missing Args/Returns** — accepted. Added Args, Returns, and Raises.

6. Ponytail pass, nothing to cut — no action.

Tests: 48 passed locally, the 12 live-credential failures are unchanged from `main`.

PR #5241, opened 2026-09-10
perf(extracts): read the enrollment view once in int_extracts__student_enrollments

### LINKED ISSUE #5215: perf(extracts): materialize the heavy upstream views of int_extracts__student_enrollments
## What and why

`int_extracts__student_enrollments` costs 152 slot hours over 7 days, 9 slot minutes per run, 145 runs a day. Its output is 134k rows. In the latest prod run, stage `SC3: Join+` read 80M rows to write those 134k, and the plan has 195 stages.

The SQL itself is a long chain of left joins. The cost is that several of the refs are views, so each run recomputes their upstream logic. This is a materialization problem, not a join-shape problem. Inferred from the stage count and read ratio, not yet confirmed per ref.

## Proposed approach

Read the stage plan, find the 2 or 3 refs that contribute the 80M rows, and materialize those as tables. Leave the model SQL alone.

<details>
<summary>For Claude</summary>

> File: `src/dbt/kipptaf/models/extracts/int_extracts__student_enrollments.sql`. Stage plan: `unnest(job_stages)` on `JOBS_BY_PROJECT` filtered to this `node_id`, sort by `records_read`. Match stage input names to refs.
>
> Check `__TABLES__.type` on each ref's dataset to tell views (2) from tables (1) before changing anything.
>
> Verify: `records_read` on the top join stage drops below 10M and the run count and output rows stay the same.
>
> Parent: #5212

</details>

cbini:
## Summary & Motivation

When merged, this pull request will make `int_extracts__student_enrollments` read the enrollment view once per run instead of 6 times.

The model read `base_powerschool__student_enrollments` in its main select and in 5 CTEs. That ref is a passthrough view over `int_students__student_enrollments`, a 735-line view, so BigQuery inlined the whole enrollment SQL 6 times per run. Each run had 195 stages, and 45 `Coalesce` stages carried 14 of its 17 slot-minutes. At about 145 runs a day that was 153 slot hours a week, the second most expensive node in prod.

The 5 lookups are now window functions over one scan of the view. The dev build has 81 stages and takes 3.4 slot-minutes. Output is unchanged: same 134,501 rows, and all 6 affected columns match prod on every row.

The model also moves from the eager table condition to an hourly cron. Its fastest consumers, the Clever extract job and the ops dashboard, refresh hourly; everything else is nightly or a Connected Sheet that pulls on open. About 145 rebuilds a day become 24.

Refs #5215

## Reviewer Notes

- `es_graduated` and the two attended columns use `first_value(... ignore nulls)` with a full-partition frame, ordered by `rn_all asc`. The frame clause is what lets a row of any school level see the latest MS or ES stint. `rn_all` replaces the old `exitdate desc` because it carries a deterministic tiebreak (#5146); prod picks are identical on every row.
- `next_year_school` partitions the `lead()` by `rn_year` instead of filtering to `rn_year = 1`, then broadcasts the primary stint's value to the other rows of that student-year with `max()`.
- The Finalsite flag moved from a `group by` CTE to a `sum()` window plus a `case`. Grade 99 rows still get null, and a student-year with no `exitdate` still reads `NTK`.

### General

### dbt

- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — no consumer change
- [x] **Breaking change?** No. Column names and types are unchanged; the contract build passed.
- [x] External sources unchanged.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The issue's stated cause (view fan-in on the 80M-row join stage) was wrong: that stage cost 2.4 of 17 slot-minutes. The cost was the 6-fold inlining of `int_students__student_enrollments`. Run alone, that view costs 1.4 slot-minutes and 41 stages; its Focus-branch `Coalesce` stages cost 0.03 slot-minutes each standalone and 0.7 each when inlined into the extract's 195-stage plan.

Probe evidence (prod SQL wrapped in `count(*)` plus a row fingerprint): baseline 19.3 slot-min, 194 stages, 44 `Coalesce`; the same query with the 5 side CTEs pointed at a table 5.6 slot-min, 82 stages. Dev build of this branch (job `c6ff4578-8d13-49da-9cc9-a40db05b84a0`): 3.4 slot-min, 81 stages, 21 `Coalesce` at 1.1 slot-min. The last 2 prod runs were 7.3 and 9.3 slot-min, 195 stages each.

Parity: `zz_cbini_kipptaf_extracts.int_extracts__student_enrollments` vs `kipptaf_extracts.int_extracts__student_enrollments`, keyed on (`_dbt_source_project`, `student_number`, `academic_year`, `entrydate`) for 112,938 rows and on a 9-column null-safe key for the 21,563 rows with null `entrydate`. Zero differences on `ms_attended`, `es_attended`, `es_graduated`, `next_year_school`, `next_year_schoolid`, `next_year_enrollment_history`. Zero students appear in more than one `_dbt_source_project` and no student has 2 `rn_year = 1` rows in a year, so partitioning the attended windows by project and broadcasting the lead with `max()` cannot change results. Ties on `exitdate` were arbitrary before (`row_number` with no tiebreaker); ordering by `rn_all asc` makes them deterministic, and prod has 0 such ties today so no value changes. All three windows partition by `_dbt_source_project, student_number`.

Cadence map behind the hourly cron, all 111 direct consumers: Clever extract job `@hourly`; `rpt_tableau__ops_dashboard` `0 * * * *`; attendance dashboards 06:00 and 15:00; assessment dashboard 01:00, 18:00 and Fri 16:00; ParentSquare (Newark job) 03:00 and 18:00; DeansList, Illuminate, autocomm and the remaining Tableau exposures nightly between 00:00 and 08:00; 69 Google Sheets exposures are Connected Sheets (pull); 2 Cube marts are staff surveys. The hourly consumers fire at :00 like the rebuild, so they may read the previous hour's table; `50 * * * *` is the alternative if that matters.

Not done here: `int_students__student_enrollments` stays a view. Materializing it (the config is commented out in its YAML) is a separate decision.

</details>

### DIFF
--- a/src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments.sql
@@ -50,19 +50,19 @@ with
 
             first_value(ms_school_abbreviation ignore nulls) over (
                 partition by _dbt_source_project, student_number
-                order by exitdate desc
+                order by rn_all asc
                 rows between unbounded preceding and unbounded following
             ) as ms_attended,
 
             first_value(es_school_abbreviation ignore nulls) over (
                 partition by _dbt_source_project, student_number
-                order by exitdate desc
+                order by rn_all asc
                 rows between unbounded preceding and unbounded following
             ) as es_attended,
 
             first_value(es_grad_school_abbreviation ignore nulls) over (
                 partition by _dbt_source_project, student_number
-                order by exitdate desc
+                order by rn_all asc
                 rows between unbounded preceding and unbounded following
             ) as es_graduated,
 
--- a/src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments.sql
@@ -61,7 +61,7 @@ with
             ) as es_attended,
 
             first_value(es_grad_school_abbreviation ignore nulls) over (
-                partition by student_number
+                partition by _dbt_source_project, student_number
                 order by exitdate desc
                 rows between unbounded preceding and unbounded following
             ) as es_graduated,
--- a/src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments.sql
+++ b/src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments.sql
@@ -5,50 +5,79 @@ with
         from {{ ref("int_students__calendar_week") }}
     ),
 
-    esms_attend as (
+    enrollments as (
         select
-            _dbt_source_project,
-            student_number,
-            school_level,
-            school_abbreviation,
+            *,
 
-            row_number() over (
-                partition by student_number, school_level order by exitdate desc
-            ) as rn,
-        from {{ ref("base_powerschool__student_enrollments") }}
-    ),
+            if(
+                school_level = 'MS', school_abbreviation, null
+            ) as ms_school_abbreviation,
+            if(
+                school_level = 'ES', school_abbreviation, null
+            ) as es_school_abbreviation,
+            if(
+                grade_level != 99, date_diff(exitdate, entrydate, day), null
+            ) as days_enrolled,
 
-    es_grad as (
-        select
-            student_number,
-            school_abbreviation,
-
-            row_number() over (
-                partition by student_number order by exitdate desc
-            ) as rn,
-
-        from {{ ref("base_powerschool__student_enrollments") }}
-        where
-            grade_level = 4
-            and extract(month from exitdate) = 6
-            and exitdate < current_date('{{ var("local_timezone") }}')
-    ),
-
-    next_year_school as (
-        select
-            student_number,
-            academic_year,
+            if(
+                grade_level = 4
+                and extract(month from exitdate) = 6
+                and exitdate < current_date('{{ var("local_timezone") }}'),
+                school_abbreviation,
+                null
+            ) as es_grad_school_abbreviation,
 
             lead(school_abbreviation, 1) over (
-                partition by student_number order by academic_year asc
-            ) as next_year_school,
+                partition by student_number, rn_year order by academic_year asc
+            ) as next_year_school_lead,
 
             lead(schoolid, 1) over (
-                partition by student_number order by academic_year asc
+                partition by student_number, rn_year order by academic_year asc
+            ) as next_year_schoolid_lead,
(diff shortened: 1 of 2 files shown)


cbini:
Per-finding verdicts on the claude-review comment. Citations checked against the file at f801ff7.

1. **`es_graduated` window omits `_dbt_source_project`** — Declined. Deliberate: it mirrors the old `es_grad` CTE and join, which also keyed on `student_number` alone. The collision check the reviewer asked for is already run and recorded in the PR body: 0 of 28,988 students appear in more than one `_dbt_source_project`, so the two partitionings produce the same rows today.

2. **Arbitrary tie-break on `exitdate`** — Declined for this PR. Adding a tiebreaker changes which row wins on ties, which is a behavior change outside a perf refactor. The old `row_number()` had the same arbitrary order. Parity against prod came back identical on every row, so no tie is currently resolving differently.

3. **No unit test** — Declined. Verification for this change is the full-row comparison against the prod table (134,501 rows, 0 differences on the 6 columns). A dbt unit test would need mocked inputs for a 130-column view and would not add coverage beyond that comparison.

4. **Lint not independently re-run** — Done. `trunk check --force --no-fix` on the file reports no issues, and the Trunk check run on this PR passed.

5. **Named `WINDOW` clause** — Declined. No model in the repo uses it, and sqlfmt support is unverified. Two 4-line duplicates are not worth being the first.

6. **Model altitude** — No issue, agreed.

cbini:
Update on finding 1: applied in 46dbef0. `es_graduated` now partitions by `_dbt_source_project, student_number` like the other two windows. Every cross-region match keys on the project per the students conventions, so the old omission was a gap, not a design. Output is unchanged because no student spans projects.

On finding 2, checked prod: 0 students have 2 stints sharing the top `exitdate` at the same school level or for the grade-4 June exit, so the tie-break has no effect on current data. Left as is.

The earlier dbt Cloud failure was the staging-schema clone race (3 "Not found: Table" errors in `zz_stg_kipptaf_powerschool` while the staging build recreated those tables). This push re-runs CI after that build finished.

cbini:
Update on finding 2: applied in 310b1df. The three `first_value` windows now order by `rn_all asc` instead of `exitdate desc`. `rn_all` already ends in the deterministic keys #5146 added upstream (`reenrollments_dcid asc` in PowerSchool, `startdate desc` in Focus), so ties no longer depend on scan order. Compared both orderings over the prod view: identical picks on all 134,501 rows, and `rn_all` is never null.

PR #5239, opened 2026-09-10
refactor(kipptaf): read PowerSchool identity from the package facts and drop Miami from students and 14 more unions (PR B)

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make kipptaf read PowerSchool student identity and school attributes from the package facts instead of joining `stg_powerschool__students`, `stg_powerschool__schools`, and `stg_powerschool__terms`.

It will drop the `kippmiami` relation from `stg_powerschool__students` and 14 more unions that no reader keeps Miami rows from.

It will delete 6 dead Miami literals.

It will remove the `powerschool` package from kippmiami again, after the second archive rebuild.

This closes the PR B half of #5228.

## Reviewer Notes

- Before merging, materialize `kipptaf/powerschool/int_powerschool__gpa_term` in the Dagster UI. Its prod view was compiled at 22:35 UTC on 2026-09-09, before Newark, Miami, and Paterson carried the package's `students_student_number` column, so it currently reads null identity for those 3 regions. This PR does not change that model's SQL, so the deploy will not recompile it. Until someone recompiles it, `int_students__gpa` reads null identity for those 3 regions.
- Every touched model is row-identical to prod, by region and year. See the count tables in "For Claude".
- We verified all 15 union drops reader by reader against pre-edit baselines.
- The KIPP Forward roster now picks a student's `gpa_cumulative` row through their AY2025 primary school instead of the archive students table. `previous_year_gpa` counts stay identical.
- The DIBELS expectations model dropped the `state_excludefromreporting` filter. It is row-identical.
- The scaffold's schools-side Miami literal stays, because `schools` still keeps Miami.
- dbt Cloud CI builds `state:modified+`. That will rebuild every reader of the 15 dropped unions, so expect a long CI run.
- Accepted Miami loss: `rpt_deanslist__transcript_gpas` loses 5,975 Miami rows, and `rpt_deanslist__transcript_grades` loses 48,486 Miami rows from AY2019 to AY2025. kippmiami no longer uses DeansList, so we accepted this loss instead of restoring the join.

Effect on #5193: the exit-marking DML is no longer needed, the 2 `exclude_frozen` calls on `rpt_clever__enrollments` and the 1 on `rpt_clever__students` are dead, and `schools` and the 5 staff gates remain.

### General

- [ ] Update **due date** and **assignee** on the
      [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
[...]

Issue #5228, opened 2026-09-09
refactor(dbt): carry student identity and school attributes on the PowerSchool package facts so kipptaf drops Miami from students and 14 more unions


cbini:
## What and why

Every Miami PowerSchool archive fact keys students by `studentid`, the PowerSchool surrogate. Only `stg_powerschool__students` maps that to `student_number`. So kipptaf models that want Miami grade or GPA history join the 4-region students union to get a Miami `student_number`, and the students union has to keep Miami for them. Miami is otherwise served by Focus, and Ops will not import stored grades or course enrollments into Focus, so the archive is permanent and kipptaf must keep reading it.

`base_powerschool__course_enrollments` already solves this for one fact: the shared `powerschool` package joins students and emits `students_student_number` next to `cc_studentid`. This issue extends that pattern to GPA and final grades, adds school and term attributes to the facts that kipptaf joins dimensions onto, and rebuilds the Miami archive once so the new columns exist for Miami. After that, no kipptaf model joins archive students, `stg_powerschool__students` drops Miami, and 14 more unions whose readers never keep Miami rows drop it too.

## Proposed approach

Package (`src/dbt/powerschool`):

- `base_powerschool__final_grades` passes `students_dcid`, `students_student_number`, `school_abbreviation`, and `school_level` through from course enrollments.
- `int_powerschool__gpa_term`, `gpa_cumulative`, and `gpa_cumulative_year` join students and schools and emit `students_student_number`, `school_name`, `school_abbreviation`, `school_level`.
- New `int_powerschool__calendar_day`: staging calendar days left joined to terms and schools, emitting `yearid`, `academic_year`, and the school columns plus `schoolcity`.
- `base_powerschool__student_enrollments` adds `entry_school_abbreviation`.

Miami archive: re-run the #5201 recipe (re-include the package in kippmiami with the 16 hooks, one prod build, remove the package again).

Kipptaf:

- One new wrapper for `int_powerschool__calendar_day`. The other wrappers pick up the new columns on their own.
- 10 readers stop joining students, schools, or terms and read the column from the fact. 4 of those joins are already redundant today because course enrollments carries the school columns.
- `stg_powerschool__students` and 14 more unions drop the `kippmiami` relation. 6 dead Miami literals on students readers go with them.

Delivered as 2 PRs around the prod rebuild, the same shape as #5195 and #5196.

<details>
<summary>For Claude</summary>

Context, not instructions.

[...]

PR #5245, opened 2026-09-10
refactor(students): enumerate union branches in the 3 kipptaf student views

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make `int_students__students`, `int_students__terms`, and `int_students__calendar_day` list their columns on both sides of a positional ANSI `union all` instead of `select *` over `full union all corresponding`, and will document why in `.claude/rules/dbt-sql.md`. It also adds the rule: prefer ANSI SQL or a dbt macro over BigQuery-only syntax unless the BigQuery form adds capability, and restructures `dbt-sql.md` so the rules every model hits come first and the one-construct traps come last. It also denies the `Agent` tool in the `claude-code-review` workflow.

BigQuery fixes a view's column list when the view is created. Dagster rebuilds a view only when its raw SQL changes. A `select *` view therefore never picks up a column an upstream adds. Enumerating makes the column add an edit to the view's own SQL, so the view recompiles on deploy. On 2026-09-09 `kipptaf_powerschool.int_powerschool__gpa_term` compiled `students_student_number` as a null cast for 3 regions and stayed that way until a manual materialization.

Follow-up to #5239. The 15 union wrappers from that PR are untouched.

Refs #5228

## Reviewer Notes

- This branch was cut 1 minute before #5239 merged, so the first push still carried the `_dbt_source_project != 'kippmiami'` filters that #5239 deleted. `origin/main` is merged in now and both filters are gone. The PowerSchool staging unions no longer contain a Miami relation, so the filters were dead either way.
- `full union all corresponding` is BigQuery-only. Standard `CORRESPONDING` keeps only shared columns, and the `FULL` null-fill is an extension. The narrower Focus branch now lists every column in the PowerSchool branch's order, with `cast(null as <type>)` where Focus has no equivalent, and the union is positional.
- The first dbt Cloud run failed the uniqueness test on `int_students__attendance_daily` with 793,259 duplicates, all Miami. Cause: the PR schema kept a `stg_powerschool__students` view from the pre-merge run that still unioned the Miami archive, and the post-merge run deferred to it instead of the staging clone. The PR was closed and reopened to drop that schema. Not a logic change.
- `int_students__calendar_day` moves `yearid` to the end of the Focus branch's select list. sqlfluff ST06 flagged the old order once the outer select was enumerated. Output order is set by the PowerSchool branch, so nothing downstream changes.
- `int_students__students` drops the `focus_conformed` wrapper CTE and puts the Focus join directly in the union branch. `current_stint` stays.
- `claude-code-review.yaml` adds `Agent` to `--disallowedTools`. Verified against the pinned action source: `base-action/src/run-claude-sdk.ts` breaks the SDK loop on the first `result` message, so a background subagent is orphaned and the run reports success with nothing posted (anthropics/claude-code-action#1499). The first 2 review runs on this PR posted nothing, but their job summaries show the cause was the account's session usage limit ("You've hit your session limit · resets 8:20pm (UTC)"), not this bug. The deny is hardening against a verified failure mode, not a fix for those runs. A PR that edits this workflow runs its own copy, so this change reviews itself.
- `dbt-sql.md` restructure: column order and structure rules open the file, then `select *` and UNION rules, date-range joins, dedup and surrogate keys, BigQuery syntax traps, sqlfluff traps. Every bullet keeps its text. A word-bag diff of old vs new shows only the new intro sentence, section leads, and cross-reference wording changed.

### General

### dbt

- [x] Include (or update) an exposure for all models consumed by a dashboard — none affected
- [x] **Breaking change?** No. Column names and types match prod exactly.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Verification after the `origin/main` merge, dev build with `--defer --favor-state` against a prod manifest regenerated from `77977d0430`, schema `zz_cbini_kipptaf_students`:

- Build: PASS=9 WARN=2 ERROR=0. Both warns are pre-existing: `not_null_int_students__students_lunchstatus` warns 4068 rows, and prod has 4068 null `lunchstatus` rows for kippmiami. `int_students__calendar_day__zero_enrollment_in_session_days` warns 2.
- Columns, `INFORMATION_SCHEMA.COLUMNS` dev vs prod, name, type, and ordinal position: 0 differences. Counts: students 159, terms 35, calendar_day 13. Prod columns re-read after the #5239 deploy at 17:10 UTC and unchanged.
- Row counts per `_dbt_source_project`, dev vs prod, all deltas 0:

| model | kippcamden | kippmiami | kippnewark | kipppaterson |
| students | 6003 | 4068 | 20010 | 1107 |
| terms | 743 | 247 | 1211 | 57 |
| calendar_day | 32598 | 14467 | 76287 | 4839 |

The NJ counts matched exactly, so the stale personal `zz_cbini_<district>_powerschool` source-copy caveat did not come into play.

Prod baseline saved at `.claude/scratch/5228-views/prod_cols.txt` and `rows_prod.txt` (local only).

</details>


cbini:
Verdicts on the review findings, one per item.

1. **ST06 on interleaved `cast(null as <type>)` — no change.** `trunk check --force --no-fix` on `int_students__students.sql`, `int_students__terms.sql`, and `int_students__calendar_day.sql` passed after the positional rewrite, so sqlfluff does not fire on ref/cast interleaving. That matches the rule text: sqlfluff buckets `cast()` as a simple target. One correction to the count: both `int_students__terms` branches list 35 columns, not 34, matching prod's 35.

2. **"drops the 2 wrapper CTEs" — fixed in the PR body.** Relative to `origin/main` only `focus_conformed` was removed. `current_stint` stays.

3. **CI checkboxes — checked.** Trunk, dbt Cloud, and CodeQL are green on the tip; Dagster Cloud did not trigger.

4. **`qualify` / `group by all` in the ANSI-first bullet — fixed in `e36c9a5f0c`.** The bullet now says both never pass the capability test and stay banned outright below.

PR #5248, opened 2026-09-10
feat(cube): exclude not_taken from proficiency, expose strand breakdowns (2 of 2)

cristinabaldor:
# Pull Request

## Summary & Motivation

When merged, this pull request will exclude `not_taken` rows from Cube's proficiency measures, expose the strand-level breakdowns that #4710 adds to the fact, and rebuild the proficiency pre-aggregation under a new name.

**This is the second of two merges, and it is based on #4710, not on `main`.** Do not merge it until #4710 has merged **and** `fct_assessment_scores_enrollment_scoped` has rebuilt in production. See _Reviewer Notes_ for what happens if that order slips.

`response_type` used to be null for every non-Illuminate source and for every Illuminate assessment a student was assigned but never took. #4710 splits those into `overall` and `not_taken`. This PR is what makes that split matter: the roughly 940,000 `not_taken` rows leave every proficiency denominator. They have no recorded response, so they could never be proficient, and they were dragging every rate down.

Measured 2026-09-10 on a local Cube dev server, running this branch over a dev-schema build of #4710's fact:

| Measure | Before | After |
| --- | ---: | ---: |
| Global `pct_proficient` | 45.66% | 48.24% |
| Illuminate-only `pct_proficient` | 46.17% | 49.73% |
| Global `count_scores` | 13,485,050 | 14,021,353 |

**What is in this PR:** `student_assessment_scores.yml`, `student_assessment_scores_view.yml`, and the three `src/cube/mcp/project_knowledge/` files.

Refs #4708.

**Rollback plan.** Revert this PR before reverting #4710 — the order is not a preference, and the reason is in `docs/superpowers/plans/2026-08-28-strand-scores-rollback.md` on #4710's branch. Reverting this PR also restores the pre-aggregation's old name, so there is no manual rename step.

## Reviewer Notes

**This redefines two existing metrics.** `pct_proficient` and `count_scores` change value for every source, not just the two that gained breakdown rows. Please look at the measure filters first.

**The measure filters use `IS DISTINCT FROM`, not `!=`, on purpose.** Against a fact whose `response_type` is still null on non-Illuminate rows, `null != 'not_taken'` evaluates to null rather than true, so the row is dropped. Measured 2026-09-10 with this branch's cube model against the then-current production fact: DIBELS, i-Ready, STAR and all five NJ/FL state sources each returned **exactly 0** — 1.32M rows, silently, with no error. The null-safe form counts them instead, which is their pre-change behaviour. That is what makes the merge order recoverable rather than merely correct, and it is what protects the rollback path, where the old fact comes back while this model may still be deployed. Do not simplify it back to `!=`.

**The pre-aggregation is renamed, deliberately.** `proficiency_rollup` becomes `proficiency_rollup_v2`. Without the rename its twelve yearly partitions would rebuild independently against one name, and a multi-year query could mix old and new vocabulary mid-sweep and return partial history with no error. Merging after the fact has rebuilt means `_v2` gets built against the correct data on its very first build.

**Every measure read goes through that pre-aggregation, not the fact.** Confirmed through Cube's `/sql` endpoint on 2026-09-10. A correct fact table is therefore not evidence that consumers see correct numbers — the rollup has to have rebuilt too. Its `refresh_key` is `every: 1 day` across twelve yearly partitions, with no `incremental` and no `update_window`.

**The claude.ai Project update is a merge gate with a named owner.** **@anthonygwalters owns both halves** — uploading `assessment-cube-orchestrator.md` and `assessment-cube-reference.md` as project knowledge, and pasting the `README.md` content into the Project's custom instructions. These are two separate mechanisms in the claude.ai interface, not one. The repository is the source of truth; the Project is a deployment target that does not update itself. Reassign if someone else should carry it.

**One consequence of the split worth knowing:** the `response_type` vocabulary flips when **#4710** merges, but these knowledge files ship here. Between the two merges the Project's knowledge is stale in one direction — it still says `response_type` is null for vendor and state rows. That affects an internal agent's answers, not production numbers, and it closes when this merges. Uploading right after #4710 merges rather than waiting for this one is fine and slightly better.

**Consumers outside the repository were already audited for #4710.** The null-`response_type` filter breaks at that merge, not this one. Nothing extra is needed here.

### General

- [x] No Asana update needed — tracked on #4708
- [ ] Review the **Claude Code Review** comment posted on this PR. Note that `claude-review` does not run on a PR whose base is not `main`, so it may not appear; re-target or review by hand.

Not applicable — no Dagster changes.

Not applicable — no dbt changes. The fact model this reads is #4710.

Not applicable — no schedule, sensor, or integration changes.

## CI checks

- [x] **Trunk** — passes on the current head; all five files checked with `--force --no-fix`.
- [ ] **dbt Cloud** — expected to be a no-op; this PR selects no models.

<details>
<summary>For Claude</summary>

## AI assistance

Claude-authored, human-directed. Split out of #4710 on 2026-09-10 after a before-and-after validation on the local Cube dev server found the deploy-ordering hazard. The requester chose the two-PR split, chose to keep #4710 as the dbt half rather than open two fresh PRs, and approved revising the rollback runbook for the new shape.

## Detail behind the Reviewer Notes flags

**`assessment_type` added to the rollup.** Without it, no source-scoped query — the documented way to select i-Ready or DIBELS — can hit the rollup, and every such query falls back to the 14.6M-row fact. It is near-functionally-determined by `module_code`, so it costs close to zero added rollup rows.

**Functional determination in the rollup is mixed, not uniform.** For the vendor breakdown rows, `response_type_code` is populated and does determine `response_type` and `response_type_description`. For Illuminate `group` rows the code is null across all 2,867,069 rows while the description varies, so the rollup does gain rows there. The comment in the file says so rather than claiming zero added rows.

**`avg_scale_score` needs a narrower grain than the old note claimed.** Holding source, subject and grade constant is no longer enough: a DIBELS Composite score pools with an ORF words-per-minute rate. Holding `response_type = 'group'` is also not enough, because all eight subtests share it. `response_type_code` is the safe single-breakdown boundary.

**Measurement method, for anyone re-running it.** Local Cube server (`npm --prefix src/cube run dev`) started with a locally-chosen API key, queried over REST with a minted HS256 token carrying an `email` claim, with the scores cube's `sql_table` temporarily pointed at `zz_cristinabaldor_kipptaf_marts` for the after-state. Four states were captured: `main`'s model over the production fact, `main`'s model over the branch fact, this model over the branch fact, and this model over the production fact. The last of those is the deploy-ordering scenario, and it is where the zeros appeared. A fifth run after the `IS DISTINCT FROM` fix reproduced the before-state global `count_scores` of 13,485,050 and `pct_proficient` of 45.66% exactly.

**Known measurement caveat.** Because every read is served by the pre-aggregation, per-source counts and the global count can disagree within a single capture when partitions have different build times. Two such gaps were observed, 670 rows and 84,237 rows, and were not isolated further. They point at the old rollup name, which this PR retires. Treat sub-0.1% deltas from that layer as noise; the order-of-magnitude findings above are unaffected.

</details>

### DIFF
--- a/src/cube/mcp/project_knowledge/README.md
+++ b/src/cube/mcp/project_knowledge/README.md
@@ -52,7 +52,10 @@ Non-negotiables, every session:
    answer any substantive question — no matter how urgent or simple the request
    looks. Force-refresh `meta` at the start so you're never trusting a stale
    catalog.
-2. Filter `response_type` explicitly on every assessment query.
+2. Filter `response_type` explicitly on every assessment query. It is
+   non-nullable with four values (`overall`, `group`, `standard`,
+   `not_taken`) — never filter for a null/not-set `response_type` to reach
+   vendor or state rows; use `response_type = 'overall'` instead.
 3. Log every query as you go — never skip logging, never batch it to the end.
    Flag trips, inferences, and low-confidence answers in real time.
 4. State your confidence (High / Medium / Low) and name every assumption you
--- a/src/cube/mcp/project_knowledge/assessment-cube-orchestrator.md
+++ b/src/cube/mcp/project_knowledge/assessment-cube-orchestrator.md
@@ -68,9 +68,12 @@ Before step 1, do two things:
    stale catalog has already produced a confident-but-wrong "unanswerable" in a
    prior session. If a field or view you need appears to be missing, refresh
    `meta` before concluding it is unavailable.
-3. **Filter `response_type` explicitly** on every assessment query. Never rely
-   on the silent default blend — see `assessment-cube-reference.md` (Shared
-   conventions) for the accepted values and the default.
+3. **Filter `response_type` explicitly** on every assessment query. The column
+   is non-nullable with four values (`overall`, `group`, `standard`,
+   `not_taken`) — never rely on the silent default blend, and never filter for a
+   null/not-set `response_type` to reach vendor or state rows; that idiom no
+   longer returns anything. See `assessment-cube-reference.md` (Shared
+   conventions) for the full vocabulary and the default.
 4. **State confidence and flag every inference.** Give each answer a High /
    Medium / Low confidence rating, and explicitly list every interpretation or
    default you chose on the participant's behalf. Surface these for human
--- a/src/cube/mcp/project_knowledge/assessment-cube-reference.md
+++ b/src/cube/mcp/project_knowledge/assessment-cube-reference.md
@@ -20,15 +20,46 @@ Apply to every assessment source unless a source section overrides them.
   - Treat this list as current, not closed: other categories exist upstream
     (`college`, `ap`, `state_nj_parcc`, `state_fl_fsa`, plus `_unknown`
     fallbacks) but carry no scores on this view today.
-- **`response_type` — always filter it explicitly.** Values: `overall`,
-  `standard`, `group`, `null` (singular `standard` / `group`, not the older
-  `standards` / `groups`). Not additive across types. Default to `overall`
-  unless a standard- or group-level breakdown is explicitly requested. Only
-  Illuminate populates `standard` / `group`; every other source is
-  `response_type = null` (overall only). To isolate those null rows, filter with
-  operator `notSet` (or `set` for present) — `equals "null"` matches the literal
-  string, not SQL NULL, and silently returns zero rows. This holds for any NULL
-  filter.
+- **`response_type` — always filter it explicitly. The column is non-nullable**,
+  with exactly four values (singular `standard` / `group`, not the older
+  `standards` / `groups`): `overall`, `group`, `standard`, `not_taken`. Not
+  additive across types. Default to `overall` unless a standard- or group-level
+  breakdown is explicitly requested.
+  - **`overall`** — the summary row for every source (Illuminate, i-Ready,
+    DIBELS, STAR, NJ state, FL state). Filter `response_type = 'overall'` to
+    isolate it — this is the only cross-source filter that yields one row per
+    sitting. (Older guidance said to isolate the non-Illuminate sources with
+    operator `notSet`, because `response_type` used to be `null` for every one
+    of them. That idiom now returns zero rows — the column is populated on every
+    row — so use `response_type = 'overall'` wherever `notSet` used to be the
+    instruction.)
+  - **`group`** — named breakdowns: Illuminate reporting groups, and now also
+    i-Ready domains and DIBELS subtests.
+  - **`standard`** — Illuminate standards only.
+  - **`not_taken`** — an Illuminate assessment a student was expected to take
+    but has no recorded response for.
+  - `response_type_code` is populated on Illuminate `standard` rows and on
+    vendor `group` rows (i-Ready domain codes, DIBELS subtest codes) — it is
+    **NULL on Illuminate `group` rows.** Do not assume `group` always carries a
+    code.
+- **Querying the i-Ready/DIBELS breakdowns:** filter `response_type = 'group'`
+  and combine with `assessment_type` to pick `iready` vs `dibels` specifically,
+  since `group` also covers Illuminate reporting groups. For i-Ready, break out
+  by `response_type_description` (i-Ready domain names don't collide). **For
+  DIBELS, break out by `response_type_code` (`measure_standard`) instead —
+  `response_type_description` collides across DIBELS subtests:
+  `Decoding (NWF-WRC)` and `Letter Sounds (NWF-CLS)` both render as
+  `Nonsense Word Fluency`, and `Reading Fluency (ORF)` and
+  `Reading Accuracy (ORF-Accu)` both render as `Oral Reading Fluency`. Grouping
+  DIBELS by description silently merges those distinct subtests into one
+  figure.** `assessment_type` is now carried in the pre-aggregation, so these
+  queries hit the rollup rather than falling through to a live query.
+- **Proficiency measures exclude `not_taken`.** `count_scores`, the
+  `_sum_proficient`-derived measures, and the formative/CRQ pairs
+  (`pct_proficient_formative`, `pct_proficient_crq`) all exclude `not_taken`
+  rows now — a student who was not tested no longer silently drags down the
+  denominator. This moved the topline Illuminate `pct_proficient` from 45.80% to
+  49.54%. The backing pre-aggregation is renamed `proficiency_rollup_v2`.
 - **Headline metric: `pct_proficient`.** It is the one score measure comparable
   across the incompatible scales of all sources (proficient scores / total).
   `is_mastery` is the underlying per-score proficient flag. `scale_score`,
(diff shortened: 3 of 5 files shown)


cristinabaldor:
## Validation for this PR's measure changes lives on #4710

Posted the row and measure reconciliation there rather than here, because it compares fact tables and the fact change is that PR: https://github.com/TEAMSchools/teamster/pull/4710#issuecomment-5626084245

The parts that matter for reviewing **this** PR:

- **`pct_proficient` and `count_scores` are safe to redefine.** Illuminate's scored population is identical before and after — 12,173,456 rows and 6,053,158 mastery. The rate moves 46.16% to 49.72% only because the 939,982 `not_taken` rows leave the denominator; the numerator does not move. Excluding them is a fix, not a redefinition.
- **Every vendor and state source is bit-identical at `response_type = 'overall'`** — row count, mastery count, and rate to four decimals. So a consumer who scopes to the summary grain sees no change from either PR.
- **`avg_scale_score` now has 1,476,429 new `group` rows to pool with**, which is what the narrowed grain warning on that measure is for.

One finding relevant to how you verify this PR: reading these measures through Cube gave a 670-row and a 146-row Illuminate discrepancy that vanished against the tables. That was `proficiency_rollup` serving partitions built at different times. Validate measure values against the fact, not the semantic layer, until `proficiency_rollup_v2` has built once cleanly — which is the whole reason this PR merges after the fact rebuilds.

cristinabaldor:
## Two open questions for you before this comes out of draft

@anthonygwalters — both are semantics calls rather than bugs, so I would rather you decide them than guess.

### 1. `count_students` did not get the `not_taken` exclusion — deliberate, or an oversight?

Six measures got the `not_taken` filter: `count_scores`, `_sum_proficient`, `_sum_proficient_formative`, `_count_scores_formative`, `_sum_proficient_crq`, `_count_scores_crq`. The `percent_correct` and `scale_score` primitives need no filter — they already guard on `IS NOT NULL`, and `not_taken` rows are null in both.

`count_students` is the one left over. Its description reads "Distinct students (per student-year) with a score in the filtered slice," but a student whose only row in a slice is `not_taken` has no recorded response and still counts. So after this PR `count_students` and `count_scores` disagree about what a row is — the ~940k rows that leave every proficiency denominator still contribute students.

**Option 1 — add the filter.** One line: `{CUBE}.response_type IS DISTINCT FROM 'not_taken'` on `count_students`. It then means "students with a score", matches its own description, and stays consistent with `count_scores`. Cost: anyone currently reading it as an assigned-roster count loses that.

**Option 2 — leave it and reword the description** to say it counts students *assigned* the assessment, scored or not. That makes it a deliberate roster denominator sitting next to `count_scores`, so `count_scores / count_students` reads as a completion rate. Cost: two meanings for "student population" in one cube, and it wants a sentence in `assessment-cube-reference.md` so the agent does not treat the two counts as the same population.

### 2. Is `test_date_key` corrupt for some internal rows, or just null?

The cube and the dbt YAML disagree about what is wrong with this column, and I cannot tell from the descriptions which one is right.

The cube's `date_taken` dimension (which reads `test_date_key`) says it "is corrupt for a small share of internal rows," and that is the stated reason calendar and academic-year rollups route through `assessment_date_key` instead. The dbt column says only that it is "nullable for internal rows that lack a recorded sitting date upstream."

Corrupt and null are different failures with different consequences. If the values are merely absent, a consumer filtering on `date_taken IS NOT NULL` is safe. If some present values are wrong, that same consumer is silently reading bad dates and the dbt YAML gives them no warning. Whichever is true, one of the two files is currently misleading — tell me which and I will align both.

### Housekeeping

Both questions came out of an audit of all 14 cube dimensions against `fct_assessment_scores_enrollment_scoped.yml`. The rest matched, apart from three small gaps I have already fixed in `07f5fc786`: `term_key` was missing the dbt column's grain and null case, `response_type` called `not_taken` an "Illuminate" assessment where dbt says "internal," and two prose mentions spelled the vendor `iReady` instead of `i-Ready`. The literal `iready` `assessment_type` values are untouched.

Nothing else is outstanding: all six checks pass, and the merge gate is still #4710, which needs a second approval.

anthonygwalters:
> ## Two open questions for you before this comes out of draft
> @anthonygwalters — both are semantics calls rather than bugs, so I would rather you decide them than guess.
> 
> ### 1. `count_students` did not get the `not_taken` exclusion — deliberate, or an oversight?
> Six measures got the `not_taken` filter: `count_scores`, `_sum_proficient`, `_sum_proficient_formative`, `_count_scores_formative`, `_sum_proficient_crq`, `_count_scores_crq`. The `percent_correct` and `scale_score` primitives need no filter — they already guard on `IS NOT NULL`, and `not_taken` rows are null in both.
> 
> `count_students` is the one left over. Its description reads "Distinct students (per student-year) with a score in the filtered slice," but a student whose only row in a slice is `not_taken` has no recorded response and still counts. So after this PR `count_students` and `count_scores` disagree about what a row is — the ~940k rows that leave every proficiency denominator still contribute students.
> 
> **Option 1 — add the filter.** One line: `{CUBE}.response_type IS DISTINCT FROM 'not_taken'` on `count_students`. It then means "students with a score", matches its own description, and stays consistent with `count_scores`. Cost: anyone currently reading it as an assigned-roster count loses that.
> 
> **Option 2 — leave it and reword the description** to say it counts students _assigned_ the assessment, scored or not. That makes it a deliberate roster denominator sitting next to `count_scores`, so `count_scores / count_students` reads as a completion rate. Cost: two meanings for "student population" in one cube, and it wants a sentence in `assessment-cube-reference.md` so the agent does not treat the two counts as the same population.
> 
> ### 2. Is `test_date_key` corrupt for some internal rows, or just null?
> The cube and the dbt YAML disagree about what is wrong with this column, and I cannot tell from the descriptions which one is right.
> 
> The cube's `date_taken` dimension (which reads `test_date_key`) says it "is corrupt for a small share of internal rows," and that is the stated reason calendar and academic-year rollups route through `assessment_date_key` instead. The dbt column says only that it is "nullable for internal rows that lack a recorded sitting date upstream."
> 
> Corrupt and null are different failures with different consequences. If the values are merely absent, a consumer filtering on `date_taken IS NOT NULL` is safe. If some present values are wrong, that same consumer is silently reading bad dates and the dbt YAML gives them no warning. Whichever is true, one of the two files is currently misleading — tell me which and I will align both.
> 
> ### Housekeeping
> Both questions came out of an audit of all 14 cube dimensions against `fct_assessment_scores_enrollment_scoped.yml`. The rest matched, apart from three small gaps I have already fixed in `07f5fc786`: `term_key` was missing the dbt column's grain and null case, `response_type` called `not_taken` an "Illuminate" assessment where dbt says "internal," and two prose mentions spelled the vendor `iReady` instead of `i-Ready`. The literal `iready` `assessment_type` values are untouched.
> 
> Nothing else is outstanding: all six checks pass, and the merge gate is still #4710, which needs a second approval.

@cristinabaldor 

1. Being extra difficult - can we break into two metrics? Count assessed and count assigned?
2. I'm not sure where this comes from - can you ask the agent to investigate? I'd call from here but this one won't have access to the warehouse

PR #5249, opened 2026-09-10
perf(deanslist): dedupe stg_deanslist__behavior with a row_number window

### LINKED ISSUE #5216: perf(deanslist): stop re-deduplicating the full behavior history on every run of stg_deanslist__behavior
## What and why

`stg_deanslist__behavior` costs 67 slot hours over 7 days at 74 slot minutes per run. Every run sorts the whole raw GCS history by `dlsaid` to keep the newest `_file_name`. That sort is O(n log n) over all rows ever loaded, and it grows every day as new files land.

## Proposed approach

Make the model incremental. Only rows from files newer than the last run need the dedupe, merged on `dl_said`. Or partition the external source by file date so the dedupe reads a bounded window.

<details>
<summary>For Claude</summary>

> File: `src/dbt/deanslist/models/staging/stg_deanslist__behavior.sql`. The dedupe is `dbt_utils.deduplicate` on `src_deanslist__behavior` partitioned by `dlsaid`, ordered by `_file_name desc`.
>
> Check how `_file_name` is structured and whether Dagster re-pulls old partitions. A re-pull of an old file must still win over the older copy.
>
> Verify: slot minutes per run drop by an order of magnitude and `count(*)` plus `count(distinct dl_said)` match the full rebuild.
>
> Parent: #5212

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will cut the warehouse cost of `stg_deanslist__behavior` by about 5 times, with no change to its output.

The model drops duplicate rows that arrive when the same DeansList records are pulled into more than one file. It did that with `dbt_utils.deduplicate`. On BigQuery that macro turns into `array_agg(original order by _file_name desc limit 1)[offset(0)]`, which packs all 39 columns of every row into a struct and sorts those structs. Newark's last run shuffled 352 GB and spent 248.7 slot-minutes on that one step, to drop 331,705 of 44,805,382 rows — 0.74%.

This swaps the macro for `qualify row_number()`. It picks the same row and costs about 81 slot-minutes instead of 461.

Newark is where the money is: 53.8 of the node's 68.4 slot hours over the last 7 days, against 14.6 for Camden and near zero for Paterson.

Refs #5216
Refs #5212

## Reviewer Notes

The `where not is_deleted or is_deleted is null` filter still runs **after** the dedupe, in the next CTE. That ordering matters: if the newest file marks a record deleted, the record must disappear, not fall back to an older copy. Folding the filter into the `qualify` step would have broken that.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — no new models; the existing properties file is unchanged.
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — no lineage change.
- [x] **Breaking change?** No. The output column set, order, and types are untouched. The diff sits entirely inside the `deduplicate` CTE.
- [x] If adding or modifying an external source, run `stage_external_sources` — no source change.

## CI checks

- [ ] **Trunk** — passes. `trunk check --force --no-fix` on the changed file is clean locally.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Child 4 of 6 under #5212.

**Diagnosis.** Ranking re-run before starting, per #5212. Node total 68.4 slot hours over the 7 days ending 2026-09-10, split `kippnewark_deanslist` 53.8 (7 runs, 461 slot-min avg), `kippcamden_deanslist` 14.6 (4 runs, 218.8 avg), `kipppaterson_deanslist` 0.0 (7 runs, 0.3 avg). Only 1.2 GB processed per Newark run, so the cost is shuffle and sort, not scan.

Stage plan of the latest Newark prod run (`unnest(job_stages)`):

| stage | records_read | records_written | slot_min | shuffle_output_bytes |
|---|---|---|---|---|
| `S00: Input` | 44,805,382 | 44,805,316 | 169.6 | 352,211,176,042 |
| `S05: Aggregate+` | 44,805,316 | 44,473,677 | 248.7 | 11,851,000,351 |
| `S07: Output` | 44,473,677 | 44,473,677 | 10.7 | 0 |

The issue framed this as an O(n log n) sort over a growing history. That is real but secondary. The dominant term is the struct: `bigquery__deduplicate` in dbt_utils 1.4.1 aggregates the full row, so `S00` emits 352 GB of shuffle for 1.2 GB of AVRO. The same swap on #5213 moved that node from 20.3 to 2.8 slot-minutes.

**Why not incremental**, which the issue proposed. The window function gets the node under the 60-slot-hour bar on its own (68.4 → roughly 12 projected) with a 7-line diff and no state. Incremental would add a merge strategy, a full-refresh path, and re-pull correctness handling on a 44M-row table. Not worth it yet.

**Why not narrow the projection.** The model uses 32 of the source's 39 columns. Dropping the other 7 (`ExpireTime`, `TimeIn`, `TimeOut`, and the 5 hive partition columns) would shrink the window shuffle a little more. Skipped: it means listing 32 column names twice in the file, and the measured win without it already clears the bar.

**Verification.** Value-level, against the prod snapshot. Old and new ran in one query, full-outer-joined on `dl_said`, compared with `to_json_string` over all 32 output columns in contract order:

```
only_in_old: 0
only_in_new: 0
differing_rows: 0
new_rows: 44,473,677
old_rows: 44,473,677
```

`old` was the deployed `kippnewark_deanslist.stg_deanslist__behavior` table, built 2026-09-10 04:08 UTC.

Cost, measured from that same job's stage plan: the new dedupe path is `S00: Input` 68.2 slot-min (24.3 GB shuffle, down from 352 GB), 4 repartition stages at 4.1 combined, and `S06: Sort+` 8.7 — about 81 slot-min against the 461 baseline. The other stages in that job (`S0D: Join+` at 26.6 and `S05`, `S07`–`S0A`) are the comparison scaffolding, not the model.

The two shapes are equivalent by construction, ties included: `array_agg(x order by f desc limit 1)[offset(0)]` grouped by `k` and `row_number() over (partition by k order by f desc) = 1` both pick an arbitrary winner among rows tied on `f`, and both put NULL `k` in a single group.

Compile checked with `dbt compile --select stg_deanslist__behavior --target prod --project-dir src/dbt/kippnewark`. No full `dbt build` was run: the diff changes no output column, and the prod SELECT above executed the exact compiled SQL, which is stronger evidence for values than a dry run. A reviewer who wants the contract assertion itself exercised should lean on dbt Cloud CI.

**Duplicate structure**, for context. Duplicates are concentrated in the fiscal years that were pulled repeatedly:

| fiscal_year | raw rows | distinct DLSAID | file generations |
|---|---|---|---|
| 2017–2023 | 188,113 | 188,047 | 9–11 each |
| 2024 | 7,646,523 | 7,588,408 | 11 |
| 2025 | 16,831,753 | 16,831,753 | 1 |
| 2026 | 18,812,301 | 18,812,301 | 1 |
| 2027 | 1,326,692 | 1,326,692 | 1 |

Scoping the dedupe to the years that actually have duplicates was rejected as brittle — one re-pull of a recent year would silently break it.

**Follow-ups, not in this PR.**

1. `stg_deanslist__behavior` still rebuilds all 44M rows on every run, 7 times a week in Newark. A cron automation condition, or the incremental build the issue proposed, would take the remaining ~12 slot hours lower.
2. 117 models across 9 dbt projects call `dbt_utils.deduplicate` (88 of them in kipptaf). Every one pays the struct-shuffle tax in proportion to its width and row count. Worth a sweep issue driven off the cost ranking rather than a blanket rewrite.

</details>

### DIFF
--- a/src/dbt/deanslist/models/staging/stg_deanslist__behavior.sql
+++ b/src/dbt/deanslist/models/staging/stg_deanslist__behavior.sql
@@ -1,12 +1,9 @@
 with
-    -- `dbt_utils.deduplicate` compiles on BigQuery to
-    -- `array_agg(original order by ... limit 1)[offset(0)]`, which pushes the whole
-    -- 39-column row through the shuffle as a struct. `qualify row_number()` picks
-    -- the same row for a fraction of the slot time.
-    deduplicate as (
-        select *,
+    -- dbt_utils.deduplicate array_aggs the whole row and costs 5x here (#5216)
+    row_numbered as (
+        select
+            *, row_number() over (partition by dlsaid order by _file_name desc) as rn,
         from {{ source("deanslist", "src_deanslist__behavior") }}
-        qualify row_number() over (partition by dlsaid order by _file_name desc) = 1
     ),
 
     transformations as (
@@ -42,8 +39,8 @@ with
             nullif(studentmiddlename, '') as student_middle_name,
             nullif(studentlastname, '') as student_last_name,
             nullif(`weight`, '') as `weight`,
-        from deduplicate
-        where not is_deleted or is_deleted is null
+        from row_numbered
+        where rn = 1 and (not is_deleted or is_deleted is null)
     )
 
 select
--- a/src/dbt/deanslist/models/staging/stg_deanslist__behavior.sql
+++ b/src/dbt/deanslist/models/staging/stg_deanslist__behavior.sql
@@ -1,12 +1,12 @@
 with
+    -- `dbt_utils.deduplicate` compiles on BigQuery to
+    -- `array_agg(original order by ... limit 1)[offset(0)]`, which pushes the whole
+    -- 39-column row through the shuffle as a struct. `qualify row_number()` picks
+    -- the same row for a fraction of the slot time.
     deduplicate as (
-        {{
-            dbt_utils.deduplicate(
-                relation=source("deanslist", "src_deanslist__behavior"),
-                partition_by="dlsaid",
-                order_by="_file_name desc",
-            )
-        }}
+        select *,
+        from {{ source("deanslist", "src_deanslist__behavior") }}
+        qualify row_number() over (partition by dlsaid order by _file_name desc) = 1
     ),
 
     transformations as (


cbini:
Verdicts on the `claude-review` findings, one per finding. Fixed in `4ba64f25a7`.

## Finding 1 — `QUALIFY` is banned. Accepted, fixed.

Confirmed against the rules file, not just the bot's quote. `.claude/rules/dbt-sql.md:179-180` reads "**No `QUALIFY`.** Compute the window function as a named column in a CTE and filter it with `WHERE` in the next CTE." That is unambiguous and I violated it.

Cause, for the record: I read the model with `cat` through Bash. The `paths:` frontmatter on `.claude/rules/dbt-sql.md` triggers on the Read tool, not on a Bash command string, so the rules file never loaded and I never went looking for it. Not a judgement call — I did not know the rule existed.

The fix differs from the suggested rewrite in two ways, both to satisfy other rules in the same file:

```sql
with
    row_numbered as (
        select
            *, row_number() over (partition by dlsaid order by _file_name desc) as rn,
        from {{ source("deanslist", "src_deanslist__behavior") }}
    ),

    transformations as (
        select
            ...
        from row_numbered
        where rn = 1 and (not is_deleted or is_deleted is null)
    )
```

1. No `select * except (rn)` and no third CTE. The ANSI-first rule in the same file lists `select * except` among the BigQuery-only forms to reach for only when the standard form cannot do the job. Here it can: `transformations` enumerates its columns, so `rn` passes through and costs nothing.
2. The soft-delete predicate joins `rn = 1` in one `WHERE`. This preserves the ordering my reviewer note called for. `rn` is evaluated inside `row_numbered`, before either predicate applies, so filtering both together picks exactly the rows that dedupe-then-filter picks. A record the newest file marks deleted still disappears rather than falling back to an older copy.

Re-verified rather than assumed, since the shape changed. Old prod table versus new SQL on the same snapshot, full-outer-joined on `dl_said`, compared with `to_json_string` across all 32 columns:

```
only_in_old: 0
only_in_new: 0
differing_rows: 0
new_rows: 44,473,677
old_rows: 44,473,677
```

The perf win survives, as the finding predicted it would. Stage plan of the CTE form against the `QUALIFY` form: `S00: Input` 66.9 slot-min versus 68.2, `S06: Sort+` 9.4 versus 8.7, 24.3 GB shuffle in both. Still about 81 slot-minutes against the 461 baseline.

## Finding 2 — the inline comment. Partly accepted.

Trimmed from 4 lines to 1, and re-pointed at the tracking issue:

```sql
-- dbt_utils.deduplicate array_aggs the whole row and costs 5x here (#5216)
```

The 4-line version was rationale, which `dbt-sql.md:137-145` sends to the properties file. A one-line tracking-issue reference at the derivation site is the carve-out that same rule grants. Keeping it because the failure mode is someone restoring the macro during a future edit; the issue number is where the full reasoning lives.

## Finding 3 — not a defect. Agreed, no change.

`row_numbered` derives `rn`, so it is real work, not a pass-through import CTE.

## One correction to the record

Not a review finding, but the PR's checklist should not be read as saying CI validated this. The dbt Cloud CI job builds the `kipptaf` project only. This PR is confined to the `deanslist` package and changes no `source()` column set, so `state:modified+` selected nothing and the green check is a no-op, not validation. The prod SELECT above and `dbt compile --target prod` are the actual evidence.

`claude-review` fires on `opened` and `ready_for_review` only, so it will not re-run on this fix push.

PR #5254, opened 2026-09-10
chore(devcontainer): hardlink the uv cache to stop per-worktree venv duplication

anthonygwalters:
## Summary & Motivation

When merged, this pull request will stop every git worktree from costing about 1.1GB of disk.

Right now `uv` copies each Python package into each worktree's `.venv`. It copies because the package cache sits on a different filesystem from `/workspaces`, and hardlinks cannot cross filesystems. With 40+ worktrees, this fills the 31GB Codespace volume over and over.

This change puts the cache on the same filesystem as the venvs and turns hardlinking on. Packages are then shared between venvs instead of copied.

Two settings change in `.devcontainer/devcontainer.json`:

- `UV_CACHE_DIR` is new. It points at `/workspaces/.uv-cache`.
- `UV_LINK_MODE` changes from `copy` to `hardlink`.

**You need to rebuild your container for this to take effect.** After you rebuild, the old cache at `~/.cache/uv` is no longer used and you can delete it. On my Codespace that cache had grown to 9.9GB.

## Reviewer Notes

- The cache lives outside the repo, at `/workspaces/.uv-cache`. That keeps a multi-gigabyte directory out of file scans and off `.gitignore`. It is still on the same filesystem, which is the part that matters.
- Hardlinking means venvs share package files on disk. Editing a file inside `site-packages` by hand would now change it for every venv and for the cache. Normal installs are safe. Hand-patching a package is not.
- I measured this instead of assuming it. Numbers are in the fold-out.

### General

Dagster, dbt, and Docs sections are skipped. This change touches no Dagster code, no dbt models, and no schedules or sensors.

## CI checks

- [x] **Trunk** — passes. Ran `trunk check --force --no-fix` on the changed file in the worktree: no issues.
- [ ] **dbt Cloud** — expected to pass or not run. No dbt files changed.
- [ ] **Dagster Cloud** — expected not to trigger. No Dagster paths changed.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The change came out of repeated disk-exhaustion cleanups on a Codespace.

**Root cause.** `/home/vscode/.cache` reports `dev=45`. `/workspaces/teamster/.worktrees` reports `dev=1796`. They are the same 31GB storage pool but different devices, so `link()` fails across them and uv falls back to copying. `UV_LINK_MODE: "copy"` has been set since the original `chore: switch to uv` commit (2024-11-25); that commit records no rationale, and it reads as boilerplate added to silence uv's cross-filesystem warning rather than a deliberate choice.

**Measurement.** Built two identical `pandas` venvs against a shared cache at `/workspaces/.uvtest/cache` with hardlink mode, measuring `df` before and after the second build:

- first venv: package files show link count 2 (cache + venv)
- second venv: apparent size 93MB, actual new disk consumed **1MB**, link count 2 -> 3

Before the change, sampled `.so` files in worktree venvs all showed link count 1, confirming full physical copies. The test directory was removed afterward.

**Why the cache is outside the repo.** `/workspaces` and `/workspaces/teamster` are both `dev=1796`, so either location enables hardlinking. Outside the repo avoids adding a `.gitignore` entry and keeps ripgrep, trunk, and dbt file scans from walking a 9GB tree.

**Follow-up not in this PR.** After everyone rebuilds, `~/.cache/uv` is orphaned. Deleting it is a one-time reclaim (9.9GB on my Codespace). It could not be cleared in place beforehand: `uv cache clean` times out on `~/.cache/uv/.lock`, which is held by long-running `uv run` MCP server processes.

</details>

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -23,7 +23,8 @@
     "GOOGLE_CLOUD_PROJECT": "teamster-332318",
     "PYTHONDONTWRITEBYTECODE": "1",
     "TRUNK_TELEMETRY": "off",
-    "UV_LINK_MODE": "copy",
+    "UV_CACHE_DIR": "/workspaces/.uv-cache",
+    "UV_LINK_MODE": "hardlink",
     "UV_RESOLUTION": "highest"
   },
   "customizations": {


anthonygwalters:
@claude review

anthonygwalters:
## Review response

Per-finding verdicts below. One finding is adopted, two are not — reasoning and evidence for each.

### 1. Explicit `hardlink` vs `auto` — not adopting

Two problems with this one, both verified on `uv 0.12.15` in the Codespace.

**`auto` is not a valid value.** The accepted set is `clone, copy, hardlink, symlink`. Setting `UV_LINK_MODE=auto` makes uv exit with a usage error, so `uv venv` and `uv sync` fail outright:

```
$ UV_LINK_MODE=auto uv venv /workspaces/.uvt2/v
  [possible values: clone, copy, hardlink, symlink]
For more information, try '--help'.
```

**The feared failure mode does not exist.** I forced the cross-filesystem case to find out what a forced hardlink actually does when it cannot be satisfied — cache on one device, target venv on another:

```
cache device: 2049   venv device: 1796
$ UV_CACHE_DIR=/tmp/uvt2cache UV_LINK_MODE=hardlink uv pip install ... packaging
Prepared 1 package in 9ms
warning: Failed to hardlink files; falling back to full copy. This may lead to
         degraded performance. If the cache and target directories are on
         different filesystems, hardlinking may not be supported.
Installed 1 package in 6ms
 + packaging==26.3
installed file link count = 1
```

uv warns and falls back to a full copy. The install succeeds. So `hardlink` is a preference with graceful degradation, not a directive — a future machine-type or storage change costs performance, not a broken `uv sync`. The trade the finding was worried about isn't on the table, and the suggested alternative is the option that would actually break things.

### 2. Automatic stale-cache cleanup in `postStart.sh` — not adopting

The discoverability concern is fair, but `postStart.sh` runs on **every** container start, so this puts a recursive delete on a hot path to solve a one-time, per-person migration. A guard bug there deletes a live cache on every resume rather than once. The blast radius is worse than the problem: the manual version is a single command, run once, by someone who has just read why.

I'd rather solve discoverability through documentation, which is finding 3 — now done. Happy to revisit if maintainers would prefer the automation; note that `.devcontainer/scripts/` is a protected path for me, so it would be a drafted hand-off either way.

### 3. Docs entry — done

Added in 36abb9f4b as a **Container Quirks** bullet in `docs/guides/codespaces.md`, which already exists and is the right audience for this. It covers where the cache lives, why it sits outside the repo, that a change to either setting needs a container rebuild, and that the old cache is safe to delete afterwards. It also records the fallback behavior from finding 1, so the next person doesn't have to re-derive it.

One correction to the finding's framing: `docs/CLAUDE.md`'s "update docs for engineering-level changes" list enumerates specific triggers — new integration, new or changed schedule/sensor, new core pattern, new or reworked data model — and a devcontainer settings change is not among them. So this wasn't strictly required by that guidance. I added it anyway because the Codespaces guide is exactly where someone hits this, not because the rule compelled it. I also confirmed nothing under `docs/` referenced either setting beforehand, so the finding's "nothing goes stale" note holds.

### Also in this push

Merged `origin/main` (the branch was 674 commits behind, no conflicts; `docs/guides/codespaces.md` was untouched on main since the merge base). The three-dot diff is still just the two-line `devcontainer.json` change plus the new docs bullet.

cbini (approved):
noice

PR #5250, opened 2026-09-10
feat(powerschool): carry calendar week fields on the attendance fact and rebuild the Miami archive (PR 1)

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will add 3 calendar-week columns to the package attendance fact `int_powerschool__ps_adaadm_daily_ctod` (`week_start_monday`, `week_end_sunday`, `week_number_academic_year`) and re-include the `powerschool` package in `kippmiami` so the Miami archive can be rebuilt once with those columns.

This is PR 1 of 3 for #5193. Today kipptaf's `int_students__attendance_daily` inner-joins `int_students__calendar_week` to get these 3 fields. That join is the only thing that still needs the Miami calendar unions. Once every region's attendance fact carries the week fields, PR 2 drops the join and 9 Miami union relations with it.

The join is a left join on `schoolid`, `yearid`, and `calendardate between week_start_monday and week_end_sunday`. Rows with no matching calendar week keep null week fields instead of disappearing.

## Reviewer Notes

- Newark dev build: 13,915,561 rows, identical to prod. 2,908,468 of them (20.9%) have null week fields. Those are pre-2019 days with no `calendar_week` coverage (a `bell_schedule` and `cycle_day` gap). Today's kipptaf inner join drops exactly those rows, so PR 2 will keep a `week_start_monday is not null` guard to stay row-identical.
- The kippmiami re-include is byte-identical to #5231's (commit `ed848b22e9`): 16 archive hooks, same disables, same lockfile.

## After merge

Charlie: materialize the `powerschool` asset group of `kippmiami_dbt_assets` once from the Dagster UI, as for #5231. Pre-build baseline for `kippmiami_powerschool.int_powerschool__ps_adaadm_daily_ctod`: 1,285,741 rows, max `calendardate` 2026-06-04. After the build the row count must match and `countif(week_start_monday is null)` should be small. PR 1b then removes the package again.

### General

- [ ] Update **due date** and **assignee** on the [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
- [ ] Review the **Claude Code Review** comment posted on this PR. Address valid feedback; dismiss false positives with a brief reply explaining why. (Claude is advisory — use your judgement, but don't ignore it.)

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — the 3 columns are declared in the package YAML and the kipptaf union YAML
- [ ] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — none affected
[...]

Issue #5012, opened 2026-08-26
chore(dbt): retire kippmiami from the powerschool unions and delete the Miami exclusion filters once Focus fully serves Miami


cbini:
## What and why

Miami is migrating from PowerSchool to Focus. When that migration finishes, `kippmiami` should come out of the powerschool `union_relations` models, and every filter that exists only to keep Miami out of a current-state feed should be deleted.

**This is not ready to do yet.** It is filed now because the hard part is an inventory that is cheap to capture today and expensive to reconstruct later.

The hard part is not the deletion. It is deciding, for each model that reads PowerSchool, whether it wanted Miami's history or only current-state Miami. Removing `kippmiami` at the union is **silent**: the views simply return fewer rows and nothing errors. So a model that quietly needed 9 years of Miami history just starts under-reporting, and no test catches it.

Today's counts, measured on `main` at `bce5a512e5`:

| what | count |
| --- | --: |
| models reading powerschool staging, base, or intermediate | 144 |
| of those, models that never mention Miami — each needs a verdict | 89 |
| Miami exclusion filters ("keep Miami out") | 61 sites in 44 models |
| Miami-required branches ("this only applies to Miami") | 77 sites in 64 models |
| `source("kippmiami_powerschool")` calls to remove | 33, one per union model |

Miami's PowerSchool history is 9 academic years, AY2018 through AY2025 — 198 enrollment rows in 2018 rising to 1,521 in 2025, and 98,865 rows in `stg_powerschool__cc`. Around 25 models exist specifically to report on Miami, including Florida state reporting (`rpt_tableau__miami_fast`, `int_students__fldoe_fte`, `int_reporting__promotional_status`) and four mart dimensions.

## Outcome, 2026-09-10

Closed completed. The decision this issue existed to force is made and shipped: **11 unions keep the `kippmiami` relation permanently**, and the archive dataset stays.

Step 4 is revised, not executed as written. Removing all 33 `source("kippmiami_powerschool")` calls would have orphaned Miami's grades and attendance from their sections, because course enrollments were never ported to Focus and the 135 course numbers behind 3,433 Miami history sections exist only in the archive. `sources-kippmiami.yml` now lists exactly the 11 tables it keeps and says they are permanent.

[...]

PR #5259, opened 2026-09-10
refactor(kipptaf): keep 11 Miami history unions, drop 9, delete exclude_frozen (PR 2)

### LINKED ISSUE #4670: refactor(dbt): replace hand-copied frozen-region exclusion filters in current-state extracts with a shared var

### LINKED ISSUE #5193: chore(dbt): decide the fate of exclude_frozen and the Miami exclusion literals after the archive rebuild
## What and why

Split out of #5012. Once the Miami PowerSchool archive is rebuilt with the 8400 prefix and the AY2025 bound (the #5012 spec), most Miami exclusion filters in kipptaf become redundant. Not all of them do, and the ones that remain are not about the archive. This issue decides what to delete, what to keep, and what to rename.

Measured on `main` at `cd6c6700d4`: 59 Miami exclusion literals across 43 models, and 12 `exclude_frozen` calls across the 6 `rpt_clever__*` models.

The 12 `exclude_frozen` calls fall into 3 groups:

| Calls | What they gate | After the archive rebuild |
| --- | --- | --- |
| 4 | `stg_powerschool__students`, `cc`, or `sections`, alongside a year or `enroll_status` filter | Redundant once archived students are marked exited (see below). The archive has no AY2026 rows. |
| 3 | `stg_powerschool__schools` in `rpt_clever__schools`, `rpt_clever__sections`, `rpt_clever__staff` | Still needed while `schools` stays in the archive. 3 history consumers join it for Miami school names: `rpt_deanslist__transcript_grades`, `int_powerschool__gradebook_assignments_scores`, `int_extracts__course_schedule_by_term`. |
| 5 | The staff roster's `dagster_code_location` in `rpt_clever__staff`, `rpt_clever__teachers`, `rpt_clever__sections` | Not about PowerSchool tables. They keep Miami staff out of Clever because Clever serves the PowerSchool regions. `rpt_clever__staff` cross joins the roster to schools, so without the gate Miami staff get NJ school rows. |

So `frozen_powerschool_code_locations` really means "regions Clever does not serve", and 5 sites need that list whatever happens to the archive.

**Archived students still look enrolled.** `kippmiami_powerschool.stg_powerschool__students` carries 1,114 rows at `enroll_status = 0` (plus 38 at `1`) with `exitdate = 2026-06-30`, because PowerSchool was retired before the year rolled. `rpt_clever__enrollments` filters on `enroll_status in (0, -1)`, not `exitdate`, so it would emit them without the gate. Marking them exited is truthful (every one of them left the SIS on 2026-07-01) and is one `update` on the archive table; it does not need the #5012 rebuild. It is part of this issue, not #5012.

## Proposed approach

Options, in order of how much they delete:

1. Mark archived students exited. Delete the 4 redundant calls. Keep the var and macro for the other 8. Reword the macro comment from "frozen archive" to "regions Clever does not serve". No consumer edits.
1. Also drop `stg_powerschool__schools`' Miami relation from the kipptaf union and give the 3 history consumers a Focus school-name join. 5 calls remain.
1. Also replace the 5 staff gates with a positive filter on a "uses PowerSchool" region attribute, and delete the macro and var. This restates the same list under another name.

The 59 literals need the same triage: dead (reader of a relation that dropped Miami), redundant (reader with a year predicate reaching archive rows), NJ-only business rule (`rpt_gsheets__nj_state_test_roster`, `rpt_gsheets__njsmart_transfer_unverified`, `rpt_tableau__nj_school_register`, `dim_student_ell_status`, `dim_students`), or still load-bearing. `rpt_tableau__crdc_roster` lines 181 and 247 filter Miami with a regexp on `_dbt_source_relation` inside `WHERE` and should become `exclude_frozen("_dbt_source_project")` or be deleted, whichever the triage says.

<details>
<summary>For Claude</summary>

Context, not instructions.

- Blocked on the #5012 archive rebuild landing in prod (PR 1 and its materialization in the #5012 spec), so the archive has no AY2026 rows before the year-predicate triage runs.
- Regenerate the inventory from `src/dbt/kipptaf/models/`:

  ```bash
  grep -rn "not like '%kippmiami%'\|!= 'kippmiami'\|region != 'Miami'" --include='*.sql' .
  grep -rn 'exclude_frozen' --include='*.sql' .
  ```

- `exclude_frozen` and `frozen_powerschool_code_locations` live in `src/dbt/kipptaf/macros/utils.sql` and `src/dbt/kipptaf/dbt_project.yml`. Added by #5007, closing #4670.
- The `enroll_status` fix is warehouse DML on `kippmiami_powerschool.stg_powerschool__students`; it runs from a terminal, not from Claude. Before running it, grep archive consumers for `enroll_status` reads on Miami rows; on 2026-09-08 the only readers without a Miami filter were `rpt_powerschool__autocomm_students`, `rpt_deanslist__family_contacts`, and `rpt_clever__enrollments`, all current-state.
- Triage rule for a literal on an archive-relation reader: it is redundant only if the model's year predicate reaches the archive rows on every path. A `current_academic_year` filter on one CTE does not cover a second CTE that reads the union unfiltered.
- Acceptance: every one of the 59 literals and 12 calls has a recorded verdict (deleted, kept with reason, or converted), and the Clever feeds are row-identical to prod for NJ after the change.

Refs #5012
Refs #4670
Refs #5007

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will settle which Miami PowerSchool history kipptaf keeps. 11 union models keep the `kippmiami` archive relation for good: stored grades (5), attendance (3), and course enrollments (3, including `courses`). The other 9 drop it: `schools`, `assignmentscore`, `sectionteacher`, `roledef`, `teachers`, and the 4 calendar models. `sources-kippmiami.yml` shrinks to those 11 tables and says they are permanent.

Course enrollments stay because Focus never receives them: the 135 course numbers behind 3,433 Miami history sections exist only in the archive, and dropping them would orphan every Miami grade and attendance row from its section.

`exclude_frozen` and its var are gone. The 7 call sites that still do work are inline `!= 'kippmiami'` literals, the same form as the other 43 in kipptaf. The 5 dead call sites and 8 dead literals are deleted. Every remaining Miami filter is a business rule for a vendor Miami does not use, or an NJ-only report.

`int_students__attendance_daily` now reads its 3 week columns from the SIS facts (PR 1, #5250) instead of joining `int_students__calendar_week`, with a `week_start_monday is not null` guard so its row set is unchanged. 5 readers that need Miami school attributes moved from `stg_powerschool__schools` to `int_students__schools`, which carries Focus Miami schools.

Closes #5193. Closes #4670. Refs #5012.

## Reviewer Notes

- Every buildable changed model is row-identical to prod: the 6 Clever feeds byte-for-byte, `fct_student_attendance_daily` at 12,592,332 rows, `dim_course_sections` at 33,815 rows with 0 orphan course keys, `int_students__attendance_daily` with exactly 793,259 Miami history rows.
- Accepted loss: Miami PowerSchool calendar history (13,015 days before AY2026) leaves `int_students__calendar_day` and `dim_school_calendars`. Current Miami days are unchanged at 1,334.
- The 9 dropped unions return 0 Miami rows in dev. The enrollment-targets relationships test moved to `int_students__schools`; it drops from 61 orphans to 2, both a sheet row with an unknown school id that was already orphaned.
- `rpt_tableau__crdc_roster` and `rpt_clever__sections` switch off the `base_powerschool__*` passthrough wrappers while touched. The full wrapper sweep stays on #3999.
- `int_students__schools` now emits the 7 columns its properties yml documents, down from 53. `select *` over the PowerSchool union carried 46 columns, and `full union all corresponding` hid which ones the union produced. Both branches enumerate now, so a positional `union all` replaces the BigQuery-only operator. Rows and values are unchanged at 34 rows, identical to prod on every column but `_dbt_source_relation`, which differs only by the dev dataset prefix. None of the 10 readers, and no exposure, Cube model, analysis, or singular test, references a dropped column.
- The 3 `int_students__calendar_*` models said the frozen Miami archive still feeds their PowerSchool branch. It does not after this PR, so those comments and 2 matching yml descriptions are corrected.

### General

### dbt

- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — no exposure changes needed; no model output columns change
- [x] **Breaking change?** No columns renamed or removed. Row-level change is the Miami calendar history loss named above.
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — not applicable; the `kippmiami_powerschool` source is BQ-native

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Spec: `docs/superpowers/specs/2026-09-10-miami-history-unions-design.md`. Plan: `docs/superpowers/plans/2026-09-10-miami-history-unions.md`. Both in this PR.

Verification detail beyond the notes above. Targeted dev build of the 36 changed models plus the 4 history marts: PASS=99 WARN=5 ERROR=0. The `fct_student_attendance_daily` relationships warning (2,287 orphans) matches prod exactly and is pre-existing. `dbt compile --select state:modified+` (692 models) succeeded, which proves no `exclude_frozen` call survives. A `--empty` gate over the same 692 hit only environment errors (unstaged personal externals, Drive permissions) and 3 pre-existing `--empty`-only failures, one of them `extract_region("stg_powerschool__schools")` in `int_tableau__fresh_enrollment_scaffold` qualifying by table name, identical on main.

Kept as follow-ups, not in scope: 6 kipptaf readers still perform a PowerSchool-internal join, now filed as #5260; the 2 cutover-seam predicates in `int_students__course_sections` and `int_students__course_enrollments` are redundant after the archive bound but left alone; the sibling SIS-conform models (`int_students__courses`, `int_students__calendar_week`, `int_students__attendance_daily`) still use `full union all corresponding` with `select *`, which their asymmetric column sets make a larger job than the `int_students__schools` fix here.

Both column-order fixes from review are in: the `fcw` group sits with `fs` in `focus_conformed`, and the `mem.week_*` columns with the rest of `mem` in `calcs`. Neither failed sqlfluff ST06 in CI; they are the documented convention.

The 20-union disposition, the 50 remaining literals, and the per-site verdicts are in the spec's "Keep, 11", "Drop, 9", and "Filters that stay, 50" sections; the closing comment on #5193 copies them.

</details>


cbini:
Verdicts on the claude-review findings. All 3 fixed in `7dc5cfbe38` and `30b93bb484`.

1. ST06 column order in `focus_conformed`: fixed. The 3 `fcw.*` plain refs moved up beside `fs.schoolid`, before the casts and logicals. One correction to the finding: this does not fail CI. Trunk and sqlfluff passed on this PR with the columns in their old position, so ST06 did not fire. It is the documented convention in `.claude/rules/dbt-sql.md`, which is reason enough, but severity is style, not a lint break.

2. Pass-through CTE in `int_students__schools`: fixed. The model reads `stg_powerschool__schools` directly in the union's first branch. A separate final review raised the same point.

3. ST06 interleaving in `calcs`: fixed rather than deferred. The 3 `mem.week_*` columns moved up with the rest of the `mem` group, ahead of `t.academic_year`. The PR body listed this as a follow-up; since finding 1 opened the same file, it was cheaper to close both.

Rebuilt `int_students__attendance_daily` after the reorder and re-checked row identity against prod. All 8 project and history groups have delta 0, no null week fields, and Miami history holds at 793,259 rows.

A separate whole-branch review also flagged stale comments in the 3 `int_students__calendar_*` models, which said the frozen Miami archive still serves their PowerSchool branch. That branch has no Miami rows after this PR. Fixed in `30b93bb484`, along with the 2 matching YAML descriptions, a comment explaining the attendance week guard, one-line reasons on the remaining Clever gates, and the `int_powerschool__gradebook_assignments_scores` description.

cbini:
dbt Cloud CI is green on `b3b7fb3232`, run `70403224272784`, with 22 test warnings. All 22 are pre-existing prod noise, not regressions. Checked against prod:

| Test | CI warned | Prod |
| --- | --- | --- |
| `unique_fct_survey_responses_survey_response_key` | 46,062 keys | 1,363,717 rows, 580,663 distinct |
| `unique_fct_survey_submissions_survey_submission_key` | 2,559 keys | 92,459 rows, 48,956 distinct |
| `rpt_tableau__gradebook_gpa` combination | 169 keys | 2,434,480 rows, 2,434,311 distinct |
| `rpt_tableau__student_course_grades` combination | 109 keys | 964,286 rows, 963,759 distinct |
| `unique_dim_terms_term_key` | 95 keys | 4,430 rows, 4,219 distinct |
| `int_assessments__scaffold` expression | 738 | 735, live drift |
| `relationships_fct_student_attendance_daily` | 2,287 | threshold is warn above 2,286, a tracked baseline |

The earlier run on `8c1cd3632e` reported 0 warnings, so the jump needs an explanation. It is the fan-out this repo's CLAUDE.md documents: an inline SQL comment on a heavily-consumed intermediate still marks it `state:modified`, so `state:modified+` rebuilds its whole descendant graph and surfaces unrelated warn-tests. Correcting the false archive comments in the 3 `int_students__calendar_*` models did exactly that, pulling in the survey, assessment, KIPP Forward, Pearson, and AppSheet subgraphs that the narrower earlier selection never built.

Those comments were already inline on `main` and had become false, so any correction marks the models modified once. The rule's preferred end state is to keep such notes in the properties `description` instead, which does not mark a model modified. I left the notes inline to keep this PR's diff to the correction itself; say the word and I will relocate all 3 into their yml descriptions.

Nothing in the 22 is downstream evidence against the `int_students__schools` narrowing. Its own primary key test passed, and every mart primary key test in the run matches prod.

Issue #5261, opened 2026-09-10
perf(assessments): cut the non-deduplicate cost on the two assessment nodes above 10 slot hours


cbini:
## What and why

Two assessment models still cost more than 10 slot hours a day after #5252 rewrote their `dbt_utils.deduplicate` calls: `int_assessments__resolved_section_enrollments` at 16.74 slot hours over 7 days in prod, and `fct_assessment_scores_enrollment_scoped` at 12.91. #5252 deliberately scoped itself to the cost of the `deduplicate` macro alone, so this leftover cost was never chased.

For `int_assessments__resolved_section_enrollments`, both dedup sites (`internal_anchored` at 2,709,411 rows and `all_candidates` at 2,678,052 rows) were already rewritten to a `row_number()` window under #5252, so whatever is keeping it above 10 slot hours is not the macro.

For `fct_assessment_scores_enrollment_scoped`, both dedup sites (`iready_scores_raw` at 273,791 rows and `star_scores_raw` at 8,128 rows) fell below #5252's threshold gate (roughly 1M rows and at least 1 slot hour of model cost) and were left untouched, so the macro was never a meaningful part of its cost in the first place. This is the more striking case: it is the #2 node in the whole cost ranking, and the macro explains almost none of it.

We don't yet know where either model actually spends its slots. This issue asks for that measurement, not a fix: find the real cost driver for each model, then decide what to cut.

## Proposed approach

Start from `unnest(job_stages)` in `region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT`, filtered to each model's `node_id`, comparing `slot_ms`, `records_read`, and `shuffle_output_bytes` per stage. This is the same method #5252 used to isolate the macro's cost, and it should attribute the remaining cost the same way. Do not assume the fix ahead of the measurement — the point of this issue is that the cause is unknown and has to be found first.

<details>
<summary>For Claude</summary>

> Context, not instructions.

Ranking query shape: `region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT` on `teamster-332318`, filtered to `"app": "dbt"` and `"target_name": "prod"` in the query comment, `statement_type not like 'SC%'`, and `node_id` extracted via `regexp_extract(query, r'"node_id": "([^"]+)"')`.

[...]

PR #5262, opened 2026-09-10
perf(dbt): threshold the dbt_utils.deduplicate rewrite instead of sweeping every caller

### LINKED ISSUE #5252: perf(dbt): measure the dbt_utils.deduplicate struct-shuffle cost before replacing it anywhere else
## What and why

`dbt_utils.deduplicate` compiles on BigQuery to `array_agg(original order by <expr> limit 1)[offset(0)]` grouped by the partition key. That packs the whole row into a struct and pushes it through the shuffle.

On `stg_deanslist__behavior` that was expensive: 352 GB of shuffle for 1.2 GB of source, 248.7 slot-minutes in one aggregate stage, 461 slot-minutes per run. Replacing it with a ranked column filtered by `WHERE` took the run to about 81 slot-minutes, byte-identical output. Same swap on `int_powerschool__category_grades` in #5213 went from 20.3 to 2.8 slot-minutes.

115 models across 9 dbt projects call the macro. The obvious conclusion is "sweep them all." The measurements say otherwise, and that is what this issue is for: **find out where the tax is actually worth paying down, and stop there.**

Cost of all 115 callers over the 7 days ending 2026-09-10, prod only:

| | |
|---|---|
| Callers that ran in prod at all | 51 of 115 |
| Total slot hours across those 51 | 175 |
| `stg_deanslist__behavior` alone | 68.4 (39% of the total) |
| Callers above 1.5 slot hours | 17 |
| The other 34 callers, combined | 11.1 |

So 64 callers cost nothing because they never ran, and 34 more are rounding errors. Rewriting those is churn on contract-enforced models for no gain, and every rewrite is a chance to change behavior.

## Measured outcome

Step 2 measured the ranked-column form against the macro form at four sizes, and the ranked-column form never lost, at any size tested. Ratios, macro over ranked slot time: 44.5M rows 6.6x, 4.16M rows 4.6x, 125k rows 2.2x, 18k rows 2.5x.

Row count drives the penalty; row width does not. The widest table tested (3629 bytes/row, 71 columns) showed the smallest penalty; the narrowest (56 bytes/row, 7 columns) showed one of the largest.

The gate applied: dedup input above ~1M rows AND the model at 1 slot hour or more. Sites measured and their input row counts:

| Model | Dedup site | Input rows | Outcome |
|---|---|---|---|
| `int_assessments__resolved_section_enrollments` | `internal_anchored` | 2,709,411 | rewritten |
| `int_assessments__resolved_section_enrollments` | `all_candidates` | 2,678,052 | rewritten |
| `int_assessments__scaffold` | `internal_assessments` | 2,931,688 | rewritten |
| `stg_illuminate__dna_assessments__students_assessments` | — | 4.16M | rewritten |
| `fct_assessment_scores_enrollment_scoped` | `iready_scores_raw` | 273,791 | skipped |
| `fct_assessment_scores_enrollment_scoped` | `star_scores_raw` | 8,128 | skipped |
| `stg_schoolmint_grow__observations` | — | 125k | skipped |

Notable: `fct_assessment_scores_enrollment_scoped` is the #2 node in the cost ranking at 12.91 slot hours but fails the gate on both its dedup sites, so the macro is not its cost driver.

Every rewrite proved value-identical against prod: 0 only-in-old, 0 only-in-new, 0 differing.

`stg_deanslist__behavior` was rewritten separately by #5249, which has merged.

## Proposed approach

Measure first, rewrite second, and only where the measurement says so.

1. Re-run the cost ranking. The numbers above drift week to week.
2. **Test whether the struct-shuffle penalty scales, or whether it only bites above some size.** Both data points so far are huge tables (44.8M and 6.9M rows). Pick 3 callers spanning the range — a large one, a mid one around 100k rows, and a small one under 10k — and measure the macro form against the ranked-column form on each. If the penalty is roughly proportional at every size, the rule is "always prefer the window function" and it belongs in `.claude/rules/dbt-sql.md`. If it only shows up past some row count or row width, the rule is a threshold, and everything below it stays on the macro.
3. Rewrite only the callers the measurement justifies. Each one gets the same value-level proof used on #5216: old and new against the same prod snapshot, full-outer-joined on the key, compared across every output column.
4. Update `.claude/rules/dbt-sql.md` with whatever step 2 concludes. Today the file tells you to use `dbt_utils.deduplicate()` and says nothing about its cost.

Top candidates by cost, for step 3:

| Node | 7-day slot hours | Runs |
|---|---|---|
| `model.kipptaf.int_assessments__resolved_section_enrollments` | 16.7 | 35 |
| `model.kipptaf.fct_assessment_scores_enrollment_scoped` | 12.9 | 105 |
| `model.kipptaf.stg_schoolmint_grow__observations` | 10.0 | 84 |
| `model.kipptaf.int_assessments__scaffold` | 7.1 | 36 |

Two flaws in this ranking surfaced on re-run. `stg_powerschool__pgfinalgrades` is not a caller and has been removed from the table above: its `dbt_utils.deduplicate` call lives in `src/dbt/powerschool/models/sis/staging/odbc/`, which `dbt_project.yml` sets `+enabled: false` (archived under #4442), and its 12.9 slot hours actually belong to the live `dlt/` variant of the model, which contains no dedup at all. Six more nodes totalling 14.3 slot hours are misattributed the same way: the ranking regex pulls `node_id` out of the query comment, and that id carries only the model name, so it cannot distinguish the three ingestion variants (`dlt/`, `odbc/`, `sftp/`) of the same PowerSchool staging model.

Follows #5216. Related: #5212, #5213.

<details>
<summary>For Claude</summary>

> Caller list: `grep -rl -E "^[[:space:]]*(\{\{[[:space:]]*)?dbt_utils\.deduplicate\(" src/dbt/*/models --include='*.sql'`. The earlier form counted comment mentions, not just invocations — a model rewritten away from the macro that kept an inline comment naming it still matched. Real totals on current main: 116 callers, of which 108 are live.
>
> Ranking query: `region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT` on `teamster-332318`, filtered to `"app": "dbt"` and `"target_name": "prod"` in the query comment, `statement_type not like 'SC%'`, `node_id` pulled out of the comment with `regexp_extract(query, r'"node_id": "([^"]+)"')`, then intersected with the caller list. Per-stage evidence from `unnest(job_stages)`: `records_read`, `records_written`, `slot_ms`, `shuffle_output_bytes`.
>
> The macro is `bigquery__deduplicate` in `dbt_utils` 1.4.1, at `dbt_packages/dbt_utils/macros/sql/deduplicate.sql`. Its own comment cites dbt-utils issue #335 claiming `array_agg` is the more performant BigQuery form. That claim is what step 2 is testing. It may well hold for narrow tables — `stg_deanslist__behavior` is 39 columns, and the shuffle cost tracks row width, so width may matter more than row count. Measure both dimensions if the 3 test cases do not separate them.
>
> Replacement shape, per `.claude/rules/dbt-sql.md`. `QUALIFY` is banned in this repo, so it is a named column plus a `WHERE`, never `qualify ... = 1`:
>
> ```sql
> with
>     row_numbered as (
>         select
>             *, row_number() over (partition by <key> order by <expr> desc) as rn,
>         from {{ source(...) }}
>     ),
>
>     transformations as (
>         select ...
>         from row_numbered
>         where rn = 1
>     )
> ```
>
> Two traps worth carrying forward from #5216:
>
> 1. A filter that ran AFTER the macro (a soft-delete predicate, typically) cannot move ahead of the window. It can share the `WHERE` with `rn = 1`, because the window is evaluated in the previous CTE before either predicate applies — but it must not sit in the CTE that computes `rn`.
> 2. Do not reach for `select * except (rn)` to drop the helper column. The ANSI-first rule discourages it, and a downstream CTE that enumerates its columns does not need it.
>
> Verification per rewrite, since these are contract-enforced staging models and dbt Cloud CI builds kipptaf only: run old and new compiled SQL against the same prod snapshot in one query, full-outer-join on the model's key, and compare `to_json_string` across every output column in contract order. Expect 0 only-in-old, 0 only-in-new, 0 differing.
>
> Done when: the rule is recorded in `.claude/rules/dbt-sql.md`; every site clearing the gate is rewritten with a value-level proof; every site failing the gate is recorded as measured-and-skipped; the issue premise is corrected; and the follow-up issue for the non-macro cost is open.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will cut the BigQuery slot cost of four `dbt_utils.deduplicate` calls, and record a rule saying when to use the cheaper form and when not to bother.

The macro compiles on BigQuery to `array_agg(original order by <expr> limit 1)[offset(0)]`. That packs the whole row into a struct and pushes it through the shuffle. A `row_number()` window does the same job without the struct.

I measured both forms on four prod tables. The window form never lost — it ran 2.2x to 6.6x cheaper on slot time, and the output was byte-identical every time.

That does not mean rewriting all 116 callers. Below about 1M rows the saving is a fraction of a slot-minute per run, which does not repay the extra CTE the window form costs. So this PR applies a gate — the dedup input must exceed about 1M rows AND the model must cost at least 1 slot hour a week — and rewrites only the 4 sites that clear it. The other sites stay on the macro, and the ones I measured are recorded as measured-and-skipped so nobody re-measures them.

Every rewrite is proved value-identical against prod: 0 rows only in the old version, 0 only in the new, 0 differing.

Refs #5252. Follow-up: #5261.

## Reviewer Notes

**Two of the issue's three headline candidates turned out not to be macro problems.** `stg_powerschool__pgfinalgrades` is not a caller at all — its macro call sits in an archived, disabled directory, and its 12.9 slot hours belong to a live variant that has no dedup in it. `fct_assessment_scores_enrollment_scoped` is the #2 node in the cost ranking but fails the gate on both its dedup sites, so it is untouched here. I corrected the issue body and opened #5261 for that leftover cost.

**One rewrite is not shaped like the other two.** In `int_assessments__resolved_section_enrollments`, the `all_candidates` input is a two-branch UNION. Ranking inside each branch would rank them independently and break the tie-break, so that site ranks in a separate CTE reading the whole union. The non-union sites inline the window instead.

**The rule text is the part most worth reading.** It is the durable artifact — the SQL changes are a one-time cleanup, but the rule is what future model authors will follow.

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Address valid feedback; dismiss false positives with a brief reply explaining why. (Claude is advisory — use your judgement, but don't ignore it.)

### dbt

- [x] **Breaking change?** No. All three models keep their output column names, types and order, so the contracts are unchanged. Only the internal dedup mechanism changed.

## CI checks

- [ ] **Dagster Cloud** — passes or not triggered (only runs when relevant paths change and the PR is not a draft).

<details>
<summary>For Claude</summary>

**Measurement method.** Four prod tables, two arms each. Arm A is the macro's compiled form; arm B is `row_number()` in one CTE filtered by `where rn = 1` in the next. Both arms wrapped in `sum(length(to_json_string(t)))` to force full-row materialization so neither could win by column pruning. Slot time and per-stage shuffle from `region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT`. No cache hits; every pair returned byte-identical output.

| Table | Rows | Bytes/row | Cols | Macro slot-min | Ranked slot-min | Ratio |
|---|---:|---:|---:|---:|---:|---:|
| `stg_deanslist__behavior` | 44,473,677 | 241 | 32 | 117.11 | 17.75 | 6.6x |
| `students_assessments` (illuminate) | 4,155,044 | 56 | 7 | 3.48 | 0.76 | 4.6x |
| `stg_schoolmint_grow__observations` | 125,167 | 3629 | 71 | 1.27 | 0.58 | 2.2x |
| `stg_finalsite__status_report` | 18,167 | 195 | 65 | 0.11 | 0.05 | 2.5x |

**Mechanism**, from `unnest(job_stages)`: the macro form makes BigQuery add `Repartition` stages the window form never emits. Deanslist arm A ran 8 repartition stages moving about 22 GiB on top of a 43.05 GiB Input shuffle; arm B was `Input -> Sort+ -> Output`, 12.20 GiB total.

**Row width does not amplify the penalty.** The widest table tested (3629 bytes/row, 71 columns) showed the smallest ratio; the narrowest (56 bytes/row, 7 columns) showed one of the largest. The issue's stated suspicion was the opposite, so the rule says this explicitly to stop it being re-derived.

**The ~1M threshold is interpolated, not measured.** There is no data point between 125k and 4.16M. The defensible claim is that the ranked form never lost at any size tested, and that below ~1M the absolute saving does not repay the extra CTE. The rule text says so rather than claiming a measured inflection point.

**Absolute per-run savings from the benchmark are upper bounds, not estimates.** The `to_json_string` wrapper inflates both arms and does not fully cancel — in arm A the projection runs on struct-unpacked rows. On illuminate the implied saving exceeds the model's entire per-run cost, which cannot be right. Trust the ratios and the shuffle bytes (measured before the final projection). The ratios are corroborated by two real prod rewrites: 5.7x on #5249 against 6.6x measured here on the same table, and 7.2x on #5213.

**Gate results, all five sites sized from compiled SQL:**

| Model | Input CTE | Rows | Disposition |
|---|---|---:|---|
| `int_assessments__resolved_section_enrollments` | `internal_anchored` | 2,709,411 | rewritten |
| `int_assessments__resolved_section_enrollments` | `all_candidates` | 2,678,052 | rewritten |
| `int_assessments__scaffold` | `internal_assessments` | 2,931,688 | rewritten |
| `stg_illuminate__dna_assessments__students_assessments` | whole source | 4,155,044 | rewritten |
| `fct_assessment_scores_enrollment_scoped` | `iready_scores_raw` | 273,791 | skipped |
| `fct_assessment_scores_enrollment_scoped` | `star_scores_raw` | 8,128 | skipped |
| `stg_schoolmint_grow__observations` | whole source | 125,167 | skipped |

**Proof per rewrite.** Old and new compiled SQL run against the same prod snapshot in one query, full-outer-joined on the model's key, `to_json_string` compared across every output column. All returned 0 only-in-old, 0 only-in-new, 0 differing. Both `int_assessments__*` proofs compiled with `--target prod` so refs resolved to prod schemas rather than dev copies; a first attempt on `resolved_section_enrollments` used dev relations and was re-run against prod, which confirmed prod is the wider slice.

**Two corrections to the issue's own diagnostic**, both now fixed in the issue body:
1. `src/dbt/powerschool/models/sis/staging/odbc/` is `+enabled: false` (archived under #4442). Eight callers live there. The ranking regex pulls `node_id` from the query comment, which carries only the model name and cannot distinguish the `dlt/`, `odbc/` and `sftp/` variants — so 14.3 slot hours across 6 nodes were attributed to disabled code.
2. `grep -rl "dbt_utils.deduplicate"` matches comment mentions. #5249 rewrote `stg_deanslist__behavior` and left a comment naming the macro, so it still matched. Counting invocations gives 116 callers, 108 live.

**Known pre-existing issue, not introduced here.** In `int_assessments__resolved_section_enrollments`, `subject_area` is projected but sits in neither the dedup partition nor the order key, so the pick is arbitrary on ties and it feeds `score_grain_key` downstream. This was equally true under the macro — the prod proof returned 0 differing — so it is not a regression, but it is latently run-dependent and worth a look someday.

**A dev-schema build error appears in the task logs** for `int_assessments__college_assessment_practice`: a missing Google Sheets source table in the personal dev schema. Unrelated to this work and not touched.

Claude-assisted, human-directed. Spec at `docs/superpowers/specs/2026-09-10-dbt-utils-deduplicate-threshold-design.md`, plan at `docs/superpowers/plans/2026-09-10-dbt-utils-deduplicate-threshold.md`.

</details>


cbini:
Both findings accepted and fixed in 0c785664b4. Also merged `origin/main` (the branch was 27 commits behind, so the earlier green CI ran on a stale base).

**Finding 1 — "measured, skipped" not recorded at the call sites: accepted.**

The PR body claimed the skipped sites were recorded so nobody re-measures them, and that record lived only in the design spec. The root `CLAUDE.md` Docs section says a stale spec is expected, so the claim did not hold where it mattered. Added a one-line disposition comment at each of the three sites:

- `fct_assessment_scores_enrollment_scoped.sql` `iready_scores` — 273,791 input rows
- `fct_assessment_scores_enrollment_scoped.sql` `star_scores` — 8,128 input rows
- `stg_schoolmint_grow__observations.sql` — 125,167 input rows

Each names the measured row count, the threshold it failed, and `#5252`. That fits the inline-comment carve-out in `.claude/rules/dbt-sql.md` for tracking-issue refs at the derivation site.

**Finding 2 — duplicate column enumeration in `int_assessments__scaffold`: accepted.**

`internal_assessments_ranked` hand-listed the same 20 columns as `internal_assessments` purely to attach `rn`, while `all_candidates_ranked` in the sibling rewrite does the identical job with `select *,`. Collapsed to match the sibling. Net -20 lines.

Output is unchanged by construction, not just by test: `internal_assessments_deduped` still enumerates the 20 columns explicitly, so the ranked CTE's list was a second hand-maintained copy that never reached the result. Verified `internal_assessments` has no column named `rn` (no collision from the star), `dbt compile` succeeds, and `trunk check --force` is clean on all three files. The value-level proof from the original rewrite still holds, so it was not re-run.

**On the review's own caveat:** the note that the row-count, slot-time and byte-identical claims are author-reported is fair. Those came from `region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT` and from full-outer-join proofs run against prod-resolved compiled SQL; the queries and their results are in the "For Claude" fold-out of the PR body and in the design spec.

PR #5274, opened 2026-09-11
fix(kipptaf): key Miami teacher grade levels on employee_number

### LINKED ISSUE #5273: fix: the PowerSchool crosswalk overrides Miami's teacher number on the staff roster
## What's happening

I expected every Miami teacher who has a grade level in `int_students__teacher_grade_levels` to show that grade level on the staff roster. Some Miami staff show no grade level instead.

`int_people__staff_roster_history` builds `powerschool_teacher_number` as `coalesce(idps.powerschool_teacher_number, cast(w.employee_number as string))`. The PowerSchool crosswalk wins whenever it holds a row. That is correct for Newark, Camden, and Paterson, because PowerSchool issues its own teacher numbers in those regions. It is wrong for Miami. Miami moved to Focus, and every Miami producer keys on `employee_number` instead.

`stg_google_sheets__people__powerschool_crosswalk` still holds 12 active rows that map Miami staff to a retired PowerSchool number. 8 of those rows belong to active primary staff. Those 8 staff match nothing in `int_students__teacher_grade_levels`, because its Focus arm emits `employee_number` as `teachernumber`.

1 of the 8 staff loses a real grade level today. The other 7 do not teach right now. The number grows as soon as any of the 8 picks up a section.

The same key mismatch also drops those staff from `bridge_course_section_teachers` and `dim_student_section_enrollments`, which both join a Focus-era section `teachernumber` against `powerschool_teacher_number`.

## Steps to reproduce

Run this against prod. It returns 8 rows, and the `focus_grade_level` column shows the grade level each staff member should carry on the roster.

1. Query `int_people__staff_roster` for active primary Miami staff whose `powerschool_teacher_number` differs from their `employee_number`.
2. Left join `int_students__teacher_grade_levels` on `employee_number` for `_dbt_source_project = 'kippmiami'`, `academic_year = 2026`, and `grade_level_rank = 1`.
3. Compare `primary_grade_level_taught` against `focus_grade_level`.

```sql
with mia as (
  select employee_number, powerschool_teacher_number, primary_grade_level_taught
  from `teamster-332318.kipptaf_people.int_people__staff_roster`
  where primary_indicator and worker_status_code = 'Active'
    and home_work_location_region = 'KIPP Miami'
    and powerschool_teacher_number != cast(employee_number as string)
)
select m.primary_grade_level_taught, t.grade_level as focus_grade_level
from mia as m
left join `teamster-332318.kipptaf_students.int_students__teacher_grade_levels` as t
  on cast(m.employee_number as string) = t.teachernumber
  and t._dbt_source_project = 'kippmiami'
  and t.academic_year = 2026
  and t.grade_level_rank = 1
```

## Where

- **Environment:** prod
- **Run, PR, or dashboard link (if any):** none

<details>
<summary>For Claude</summary>

Root cause is the coalesce at `src/dbt/kipptaf/models/people/intermediate/int_people__staff_roster_history.sql:278-280`.

Agreed fix shape is additive, so nothing that works today changes behavior:

1. Add a new `sis_teacher_number` column to `int_people__staff_roster_history`. It resolves to `cast(employee_number as string)` for `kippmiami` and to the existing `powerschool_teacher_number` expression everywhere else.
2. Leave `powerschool_teacher_number` exactly as it is.
3. Repoint the Focus-era consumers at `sis_teacher_number`: the `int_students__teacher_grade_levels` joins in `int_people__staff_roster`, `int_people__renewal_status`, and `int_people__next_year_salary_projections`; the same joins in the 9 `rpt_*` models that read `int_students__teacher_grade_levels`; and the `focus_teachers` arm of `bridge_course_section_teachers`.

Do NOT gate `powerschool_teacher_number` itself. 2 Miami staff resolve only through the crosswalk across 687 rows of `int_students__course_enrollments`, in AY2021, AY2022, AY2024 and AY2025. `dim_student_section_enrollments` joins them through an unscoped LEFT JOIN, so a blanket gate would silently orphan that history. Keeping `powerschool_teacher_number` untouched preserves it and lets each consumer move one at a time.

```sql
select ce.teachernumber, count(*) as archive_enrollment_rows
from `teamster-332318.kipptaf_students.int_students__course_enrollments` as ce
where ce._dbt_source_project = 'kippmiami'
  and ce.teachernumber in (
    select powerschool_teacher_number
    from `teamster-332318.kipptaf_people.int_people__staff_roster`
    where home_work_location_region = 'KIPP Miami'
      and powerschool_teacher_number != cast(employee_number as string)
  )
group by ce.teachernumber
```

Checked and ruled out during investigation:

- Miami has 0 rows in `int_powerschool__teachers`, so the PowerSchool arm of `bridge_course_section_teachers` never resolved a Miami teacher. No archive links live there.
- Miami is excluded by an explicit region filter from `rpt_clever__staff`, `rpt_illuminate__users`, and `rpt_littlesis__enrollments`, and it resolves no rows in `rpt_powerschool__autocomm_teachers`. Changing a Miami teacher number causes no outbound identity churn.
- `int_topline__people_spine` and `int_topline__staff_metrics` carry `powerschool_teacher_number` as a pass-through attribute, never as a join key.
- The 9-character alphanumeric crosswalk values are real PowerSchool teacher numbers, not corrupt data. They dominate network-wide at 763 rows in Newark and 204 in Camden. The crosswalk stays load-bearing for the 3 NJ regions.
- Miami's AY2026 sections carry 753 employee-number-style `teachernumber` values and 0 legacy-style values, so `employee_number` is the correct forward key.

Fixed means the reproduce query returns `primary_grade_level_taught` equal to `focus_grade_level` for all 8 rows, the NJ regions' grade-level coverage counts stay byte-identical, and the 687 archive rows above still resolve a staff member in `dim_student_section_enrollments`.

Ops follow-up, separate from this fix: the 12 active Miami rows in the crosswalk sheet are dead data and should be deactivated. The fix makes them harmless, so this is cleanup rather than a blocker.

</details>

kevinverhoff:
# Pull Request

## Summary & Motivation

When merged, this pull request will give Miami teachers the grade level the data model already knows they teach.

`int_people__staff_roster_history` builds `powerschool_teacher_number` by preferring the PowerSchool crosswalk sheet over `employee_number`. That is right for Camden, Newark and Paterson, because PowerSchool issues its own teacher numbers there. It is wrong for Miami. Miami moved to Focus, and Focus keys teachers on `employee_number`. 8 active Miami staff still have a crosswalk row, so they carried a retired PowerSchool number and matched nothing in `int_students__teacher_grade_levels`.

This pull request adds `sis_teacher_number` to the roster. It resolves to `employee_number` for `kippmiami` and equals `powerschool_teacher_number` everywhere else. The 12 teacher-grade-level join sites and the Focus arm of `bridge_course_section_teachers` now read the new column.

`powerschool_teacher_number` does not change. The change is additive, so the rollback is to point each consumer back at the old column, one at a time.

Closes #5273

## Reviewer Notes

The gate hardcodes `kippmiami`. `rpt_clever__staff` and `rpt_illuminate__users` already hardcode the same region test, so this matches what the repo does today.

I left `powerschool_teacher_number` alone on purpose. 2 Miami staff resolve only through the crosswalk, across 687 pre-Focus rows of `int_students__course_enrollments` that `dim_student_section_enrollments` reads. Gating the old column would have silently orphaned that history.

`int_people__staff_attrition_details` carried `powerschool_teacher_number` only to serve this one join, and never exposed it. I renamed the column there rather than adding a second one. `int_topline__staff_retention` is the only other reader of that model and does not use the column.

`dim_student_section_enrollments` still joins on `powerschool_teacher_number`, which keeps the archive working. An earlier draft of this section claimed that left the 8 Miami staff unresolved there for current-year sections. I measured it, and that is wrong: those 8 staff teach 0 course sections in Miami AY2026, so no section-enrollment row depends on their key. Miami's `lead_teacher_staff_key` is null on 427 of 20257 AY2026 rows (2.1%) and 0 of 17065 AY2025 rows, and every one of the 427 is a section where Focus supplies a null `teachernumber` — no teacher of record at all, which no join key can resolve. That residual is a Focus data-entry gap, unchanged by this pull request and unrelated to it.

5 of the touched files use CRLF line endings already, which made `git-diff-check` flag the lines I edited. I stripped the carriage return on those lines only. 55 of 1435 kipptaf SQL files have the same condition, so normalizing them is a separate job.

### General

Skipped. No Dagster changes.

- [x] Include (or update) an
      summary above.
      See

No new models, so no new properties file. I added column entries for `sis_teacher_number` to both roster properties files. No new external consumer, so no exposure change. No external sources changed.

The one column rename is `int_people__staff_attrition_details.powerschool_teacher_number`, which becomes `sis_teacher_number`. That model is an intermediate with no contract and no exposure, and its only 2 readers are both in this pull request or verified clean.

Skipped. No schedule, sensor or integration changes.

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Kevin chose the additive-column shape over gating `powerschool_teacher_number`, specifically to keep the archive links reversible.

Root cause is the coalesce at `int_people__staff_roster_history.sql:278`. The new expression is the repo-blessed priority-pick form:

```sql
coalesce(
    if(
        lc.location_dagster_code_location = 'kippmiami',
        null,
        idps.powerschool_teacher_number
    ),
    cast(w.employee_number as string)
) as sis_teacher_number,
```

Verification, all against prod via a dev build of `int_people__staff_roster_history` and `int_people__staff_roster` with `--defer --favor-state`:

- `dbt build` of both roster models: PASS=15, WARN=0, ERROR=0, including the new `unique_int_people__staff_roster_sis_teacher_number` test.
- Miami active primary staff carrying a grade level: 91 in prod, 92 in the dev build. Camden 153, Newark 487, Paterson 50, all identical on both sides.
- `sis_teacher_number != powerschool_teacher_number` on 8 rows, all Miami, 0 in every other region. The gate fires only where intended.
- Joined dev to prod on `employee_number` across all 4781 rows: `powerschool_teacher_number` changed on 0 rows, 1 row gained a grade level, 0 lost one, 0 changed value.
- `dbt build --empty` over all 17 changed models plus `int_topline__staff_retention`: PASS=40 after re-running one model with `--full-refresh`. The single error was a stale personal dev relation existing as a table where the model is a view, not a code problem. The single warning was `int_people__staff_roster__tableau_location_set_expected`, which is an `--empty` artifact; the same test passed in the real build.
- `sis_teacher_number` is unique and non-null across all 4781 roster rows, which is why it carries the same `unique` test as its sibling.

Investigated and ruled out:

- Miami has 0 rows in `int_powerschool__teachers`, so the PowerSchool arm of `bridge_course_section_teachers` never resolved a Miami teacher. No archive links live there.
- The 8 stale-crosswalk Miami staff teach 0 course sections in AY2026, so `dim_student_section_enrollments` and the Cube views over it carry no current-year dependency on their key. Measured against prod rather than the dev build, since the join in question is unchanged by this pull request.
- Miami is excluded by an explicit region filter from `rpt_clever__staff`, `rpt_illuminate__users` and `rpt_littlesis__enrollments`, and resolves no rows in `rpt_powerschool__autocomm_teachers`. No outbound identity churn.
- `int_topline__people_spine` and `int_topline__staff_metrics` carry `powerschool_teacher_number` as a pass-through attribute, never a join key.
- The 9-character alphanumeric crosswalk values are real PowerSchool teacher numbers, dominant network-wide at 763 rows in Newark and 204 in Camden. The crosswalk stays load-bearing for the 3 NJ regions.
- Miami AY2026 sections carry 753 employee-number-style `teachernumber` values and 0 legacy-style, so `employee_number` is the correct forward key.
- No `src/cube/` reference to `powerschool_teacher_number`, so no Cube change is needed.
- The 3 contract-enforced `rpt_*` models with a final `select *` (`rpt_tableau__seat_tracker`, `_seat_tracker_snapshot`, `rpt_tableau__operations_pm`) enumerate roster columns in their `final` CTE, so the added column does not reach their contracted output.

Follow-ups deliberately not in this pull request:

- Ops should deactivate the 12 Miami rows in the PowerSchool crosswalk sheet — 8 active and 4 terminated, all 12 carrying a PowerSchool number that differs from `employee_number`. This fix makes them harmless, so it is cleanup, not a blocker.
- Focus leaves `teachernumber` null on the sections behind 427 of Miami's 20257 AY2026 course-enrollment rows, so those rows carry no lead teacher in `dim_student_section_enrollments` or in the two Cube views that read it (`student_section_enrollments_view`, `student_assessment_scores_view`). Separate from this fix and worth its own issue if it is not already tracked.
- Grade level taught is still absent from `dim_staff` and the `staff_directory` Cube view, so Cube cannot answer the question at all.

</details>


kevinverhoff:
## Review verdicts

One verdict per finding. Finding 1 is fixed in eb4b41096. Findings 2, 3 and 4 need no code change, and 3 I disagree with.

### 1. Duplicate inline cast — accepted, fixed

Correct, and the rule citation is right. `cast(w.employee_number as string)` ran twice in one SELECT.

Fixed in eb4b41096 by deriving `employee_number_string` as a named column in the `adp_df_union` CTE and referencing the plain column in both coalesces. `employee_number_string` is excluded from the `w.*` passthrough, so the model's output columns do not change.

Worth recording for the next person: my first attempt put the cast at the end of the Dayforce branch and after the nulls in the ADP branch. BigQuery binds UNION ALL branches by position, so the build failed with `Column 78 in UNION ALL has incompatible types: STRING, DATE`. The cast now sits at the same ordinal position in both branches. This is the hazard `.claude/rules/dbt-sql.md` warns about, and here it failed loudly only because the two columns had different types.

I rebuilt and re-ran the full comparison after the change. It is value-preserving: across all 4781 roster rows `powerschool_teacher_number` is still identical to prod on every row, neither key is ever null, and the grade-level delta is still 1 gained, 0 lost, 0 changed.

### 2. No properties.yml for `int_people__staff_attrition_details` — agreed, out of scope

Confirmed, and the gap is wider than the finding says. That model has no properties file and therefore no uniqueness test, which the per-layer requirements in `.claude/rules/dbt-models.md` require of every intermediate.

Both conditions predate this pull request and the renamed column was equally undocumented under its old name, so I am not fixing it here. Documenting a 400-line model is its own change and belongs in its own pull request.

### 3. Composite uniqueness test not extended — declined

Not deliberate-but-unexamined; deliberate. Adding `sis_teacher_number` to the composite would weaken the test rather than strengthen it.

`employee_number` already carries a standalone `unique` test on this model. Any composite that contains a column which is itself unique cannot fail unless that column's own `unique` test fails first. So `unique_combination_of_columns([employee_number, powerschool_teacher_number])` is already vacuous today, and adding a third column would keep it vacuous.

`sis_teacher_number` instead carries its own column-level `unique` test, which is the test that can actually fail. It passes: verified unique and non-null across all 4781 rows.

The existing composite test looks removable for the same reason, but that is unrelated to this fix and I left it alone.

### 4. Verification claims need local re-confirmation — already done

Ran in this session, before and again after the finding 1 fix. Recorded here because the reviewer correctly could not check it.

- `uv run dbt build --select int_people__staff_roster_history int_people__staff_roster --target dev --defer --favor-state --state target/prod`: PASS=15, WARN=0, ERROR=0, including the new `unique_int_people__staff_roster_sis_teacher_number` test.
- Miami active primary staff carrying a grade level: 91 in prod, 92 in the dev build. Camden 153, Newark 487, Paterson 50, identical on both sides.
- `sis_teacher_number != powerschool_teacher_number` on 8 rows, all Miami, 0 in every other region.
- Dev joined to prod on `employee_number` across all 4781 rows: `powerschool_teacher_number` changed on 0 rows, 1 row gained a grade level, 0 lost one, 0 changed value.
- `dbt build --empty` over all 17 changed models plus `int_topline__staff_retention`: PASS=40. The one error was a stale personal dev relation existing as a table where the model is a view, cleared with `--full-refresh`. The one warning was `int_people__staff_roster__tableau_location_set_expected`, an `--empty` artifact; that test passed in the real build.

dbt Cloud CI also passed on the first commit, which independently exercised the modified selection across the roster's descendant graph.

PR #5278, opened 2026-09-11
refactor(powerschool): move the section-teacher join into the package

### LINKED ISSUE #5260: refactor(dbt): move 6 PowerSchool-internal joins from kipptaf into the powerschool package
## What and why

#5228 set the rule: joins between PowerSchool tables happen in the shared `powerschool` package, and kipptaf unions only the finished, denormalized package model. Six kipptaf models still rebuild a PowerSchool-internal join themselves. They produce correct output today (all run NJ-only after #5259), so this is a placement fix, not a bug fix. Moving the joins down means each region computes them once and kipptaf stops carrying PowerSchool table knowledge in extract and mart code.

Found while triaging #5193.

## Proposed approach

One model at a time, each its own small PR, verified row-identical to prod for NJ.

| kipptaf model | Join it rebuilds | Package model that already has it or should |
| --- | --- | --- |
| `rpt_tableau__student_course_grades` | `base_powerschool__course_enrollments` to `base_powerschool__final_grades` | `base_powerschool__final_grades` already carries enrollment attributes |
| `rpt_deanslist__transcript_grades` | `stg_powerschool__storedgrades` to `stg_powerschool__courses` and `stg_powerschool__schools` | storedgrades carries `course_name` and `schoolname` natively |
| `bridge_course_section_teachers` | `base_powerschool__sections` to `stg_powerschool__sectionteacher`, `stg_powerschool__roledef`, `int_powerschool__teachers` | `base_powerschool__sections` already joins teachers; add role if needed |
| `rpt_clever__sections` | same section-teacher join as above | same |
| `int_powerschool__gradebook_assignments_scores` | `int_powerschool__gradebook_assignments` to `stg_powerschool__assignmentscore` and `base_powerschool__course_enrollments` | new package model |
| `rpt_tableau__gradebook_assignments` | same assignment-score join | same new package model |

<details>
<summary>For Claude</summary>

Context, not instructions.

- Inventory measured on `main` after #5259. Regenerate with `grep -rl 'ref("stg_powerschool__\|ref("int_powerschool__\|ref("base_powerschool__' --include='*.sql' src/dbt/kipptaf/models` and inspect each reader for a second PowerSchool ref joined on `_dbt_source_project`.
- A model moves into the package when all its inputs are PowerSchool models the package already has (the #5012 rule). `int_powerschool__gradebook_assignments_scores` qualifies as-is.
- Package changes materialize in every region and in the Miami archive; the archive needs a rebuild only if a kept history union (`storedgrades`, `final_grades`, `gpa_*`, `ada`, `attendance_streak`, `ps_adaadm_daily_ctod`, `course_enrollments`, `sections`, `courses`) gains a column kipptaf reads for Miami. Recipe in `src/dbt/kippmiami/CLAUDE.md`.
- Acceptance: each moved model is row-identical to prod for NJ, and the kipptaf reader is a `union_relations` wrapper or a consumer of one.

Refs #5012
Refs #5193
Refs #5228

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will move the PowerSchool section-teacher join out of two kipptaf models and into one shared package model, `int_powerschool__section_teachers`.

PowerSchool records a section's teachers two ways. The `sections` table names a single primary teacher, which `base_powerschool__sections` already resolves. The `sectionteacher` table holds many rows per section, each with a role and a date range. Nothing in the package covered the second one, so `bridge_course_section_teachers` and `rpt_clever__sections` each rebuilt the same 4-table join themselves.

Now each region computes that join once, and kipptaf reads the result. The bridge drops from 4 joins to 1. The Clever feed drops from 3 to 1.

Both models produce exactly the same rows as before. This is a placement fix, not a behavior change.

Moving the join also left two kipptaf passthrough models with no readers at all, so this pull request deletes them: `stg_powerschool__sectionteacher` and `stg_powerschool__roledef`, along with their six source entries across the three New Jersey source files. Nothing else in the repo referenced them.

This pull request also re-includes the `powerschool` package in `kippmiami` so the frozen archive can build the new model. A follow-up pull request removes the package again and adds Miami to the union.

## Reviewer Notes

- **Sections come from `base_powerschool__sections`, not `stg_powerschool__sections`.** The narrower source is the correct one, and the difference is 595 Newark rows. Detail below.
- **The bridge's `course_section_key` values do not change.** The hash inputs are untouched, so nothing downstream re-keys.
- **The Focus branch of the bridge is untouched.** Only the PowerSchool branch changed.
- **Verified against production, not a dev build.** Both join forms return identical row sets. Numbers below.
- **`kippmiami` gets the `powerschool` package back.** This is the documented archive-rebuild recipe, restored from the third rebuild's commit. It comes back out in the follow-up.
- **Two kipptaf models are deleted, not disabled.** `.claude/rules/dbt-models.md` carves out exactly this case — a union passthrough with no remaining readers gets deleted outright. The district relations they read are untouched. Their orphaned production views stay in BigQuery; dbt does not drop removed models.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — both new models have one
- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — no exposure changes; no model output columns change
- [x] **Breaking change?** No. No column renamed or removed, and no surrogate key value changes.
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — not applicable; no external source changed. The new model was seeded into the 3 NJ `zz_stg_*` datasets with `dbt build --target staging` so CI can resolve the new source.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Spec: `docs/superpowers/specs/2026-09-11-powerschool-section-teachers-design.md`. Plan: `docs/superpowers/plans/2026-09-11-powerschool-section-teachers.md`. Both in this PR.

Refs #5260. Supersedes rows 3 and 4 of that issue's table, which were wrong on two points.

**Why "add role to `base_powerschool__sections`" was not possible.** The issue proposed widening that model. `sections.teacher` and `sectionteacher` are different tables, not a superset relationship. Adding role would change the model's grain from one row per section to one row per section by teacher by role, fanning out its 3 readers: `dim_course_sections`, `rpt_clever__sections`, and `rpt_branchingminds__course_info`.

**Why the kipptaf reader is not a bare `union_relations` wrapper.** The issue's acceptance criterion asked for that. `int_people__staff_roster` is a kipptaf model, and the bridge's `staff_key` hashes `sr.employee_number`, which does not exist in the package. So the package model stops at `teachernumber` and kipptaf keeps the roster join. Everything genuinely internal to PowerSchool moved down; nothing else did.

**Section source.** Sourcing from `stg_powerschool__sections` yields 52,540 Newark rows against 51,945 from `base_powerschool__sections` — 595 extra on 506 sections. All 506 drop for a missing course, none for a term or school, and all fall in AY2004 through AY2015 with retired mixed-case course numbers (`Span300`, `Sci200`, `Tec101`, `AGRI`, `NUTRI`). `dim_course_sections` excludes them for the same reason, so sourcing from staging would emit 595 orphan `course_section_key` values while adding no usable history.

**The dropped `_dbt_source_project` join predicates.** The old joins ran at kipptaf level over `union_relations` models, where that guard prevented cross-district id collisions. The new joins run pre-union inside one district's package build, where all rows already belong to one project, so the guard is a structural no-op there.

**Verification method, and why not a dev build.** The plan originally called for `dbt build --target dev --defer`. That is unusable here: kipptaf `sources-kipp*` resolve to personal `zz_<user>_*` copies under `target=dev`, and `--favor-state` governs `ref()` but not `source()`, so the new model would have been missing from the personal copies. Verified instead by running both join forms against production relations and diffing the tuple sets with `EXCEPT DISTINCT` in both directions.

| Check | Old | New | Only in old | Only in new |
| Bridge PowerSchool branch | 69,092 | 69,092 | 0 | 0 |
| Clever feed first branch | 70,397 | 70,397 | 0 | 0 |

The bridge reconciles exactly: production holds 69,845 rows, of which 753 are Focus-branch rows carrying a null `effective_start_date`. 69,092 plus 753 is 69,845, so the proof covers the whole mart. Production orphan `course_section_key` count is 0 and stays 0, since the row set is unchanged.

**Staging seed.** `dbt build --select int_powerschool__section_teachers --target staging` was run per NJ district, serially: Newark 49.0k rows, Camden 16.6k, Paterson 1.2k, `unique` test passing in each. Camden and Paterson needed `dbt deps` first — fresh worktree.

**Known non-blocking lint note.** The wrapper's `trunk-ignore(sqlfluff/AM04)` reports `trunk/ignore-does-nothing`. Its sibling `stg_powerschool__sectionteacher.sql` has the identical inert suppression, and matching that sibling was the stated requirement. `src/dbt/kipptaf/CLAUDE.md` prefers inline `regexp_extract` over `extract_source_project()` for `select *` union views for exactly this reason — worth a sweep across both files later, out of scope here.

**Miami is deliberately absent.** PR #5259 dropped `kippmiami` from `stg_powerschool__sectionteacher`, `stg_powerschool__roledef` and `int_powerschool__teachers`. The bridge was not among that PR's 43 changed files and not in its verification set, so it silently lost 19,517 Miami archive rows across 3,433 sections and 4 roles, AY2018 through AY2025 — and that PR's "Accepted loss" section names only the calendar history. Those are the same 3,433 sections #5259 deliberately kept, and `dim_course_sections` still carries them, so Miami history sections currently have zero teacher rows against them. The follow-up PR restores them once the archive has built the new model.

**Follow-up sequence.** After this merges: build `int_powerschool__section_teachers` into `kippmiami_powerschool` with `--target prod` (expect 19,529 rows), then open the follow-up to remove the `powerschool` package from `kippmiami`, add Miami as the 4th union relation, and add its `sources-kippmiami.yml` entry.

</details>


cbini:
Verdicts on the three findings.

**1. Staging seed required before merge — already done, and CI proves it.**

`dbt build --select int_powerschool__section_teachers --target staging` ran per NJ district with explicit authorization, serially: Newark 49,022 rows, Camden 16,633, Paterson 1,247, `unique` test passing in each. Camden and Paterson needed `dbt deps` first — fresh worktree. The dbt Cloud check on this PR is green, which is the end-to-end confirmation the finding asks for.

**2. Row-count and grain verification — done, by a stronger method than suggested.**

The suggested `--target staging` build and row-count diff would not have settled this. Counts matching is weaker than set identity, and a dev build was unusable here for a separate reason: kipptaf `sources-kipp*` resolve to personal `zz_<user>_*` copies under `target=dev`, and `--favor-state` governs `ref()` but not `source()`, so the new model would simply have been absent from those copies.

Ran both join forms against production relations instead and diffed the tuple sets with `EXCEPT DISTINCT` in both directions:

| Check | Old | New | Only in old | Only in new |
| Bridge PowerSchool branch | 69,092 | 69,092 | 0 | 0 |
| Clever feed first branch | 70,397 | 70,397 | 0 | 0 |

The bridge reconciles exactly. Production holds 69,845 rows, of which 753 are Focus-branch rows with a null `effective_start_date`, and 69,092 + 753 = 69,845 — so the proof covers the whole mart rather than one branch. Orphan `course_section_key` count is 0 in production and stays 0, since the row set is unchanged.

**3. `src/dbt/powerschool/CLAUDE.md` transiently false — fixed, and made permanent.**

Correct, and worth more than the "self-corrects at PR 2" disposition: that line is false during *every* archive rebuild, not just this window, and there have now been four. `f1267b1b77` states the exception permanently and points at the recipe in `src/dbt/kippmiami/CLAUDE.md`, so the doc is accurate in both states and a reader who finds the package in `packages.yml` knows why.

---

Two things that landed after the static review ran, so they are not in its diff:

- `d3a6b9c0cb` deletes `stg_powerschool__sectionteacher` and `stg_powerschool__roledef` at the kipptaf level, plus their six source entries. Moving the join left both with zero `ref()` consumers anywhere in the repo. `.claude/rules/dbt-models.md` carves this case out for deletion rather than the usual disable, since the district relations they read are untouched.
- `da715a6f44` and `9a475d4443` fix three items a separate whole-branch review caught: stale `source_column` pointers on the bridge properties yml, an undeclared `_dbt_source_relation` on the wrapper, and an inert `trunk-ignore(sqlfluff/AM04)` whose correct template was the sibling one directory over.

PR #5281, opened 2026-09-11
refactor(powerschool): move the gradebook assignment-score join into the package

### LINKED ISSUE #5260: refactor(dbt): move 6 PowerSchool-internal joins from kipptaf into the powerschool package

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will move `int_powerschool__gradebook_assignments_scores` out of kipptaf and into the shared `powerschool` package, and push two of its calculations further upstream still.

The model joins three PowerSchool tables to build one row per assignment per student. Those joins belong in the package, where each region computes them once, rather than in kipptaf. This is row 5 of #5260.

Two of the model's derived columns turned out not to need the join at all, so they moved to where their inputs live: `school_level_alt` to `base_powerschool__sections` and `half_total_point_value` to `int_powerschool__gradebook_assignments`.

Nothing downstream changes. All three models that read the gradebook-scores model keep reading the same name and are untouched. Its output is identical: 23,504,022 rows, every value matching.

Two columns are removed, `region` and `credit_type`. Neither had a single enabled reader.

Refs #5260

## Reviewer Notes

**The removed columns are the first thing to check.** `region` had to go, because the macro that built it does not exist inside the package and there is no region to extract when one project is one region. `credit_type` did not have to go, but keeping it would have been worse. kipptaf rewrites four credit types on the way in (`ELA` to `ENG`, `Math` to `MATH`, `Science` to `SCI`, `Homeroom` to `HR`), and the package does not. That rewrite fires on 4,894 Paterson rows. Keeping the column would have meant either copying that rewrite into the wrapper, duplicating logic that already lives in `int_students__course_enrollments`, or shipping a silent value change. Both columns are read only by three disabled models in the dead pre-AY2627 audit cluster.

**`school_level_alt` had five hand-copied definitions and they did not agree.** One business rule — KIPP Sumner Academy, grade 5 and above, from academic year 2025, reads MS — written out five times, matching the school three different ways and the year two different ways. Two of the five carry a literal `TODO: figure out a better way to track these`. This defines it once on `base_powerschool__sections`, where all four inputs live, and it reaches `base_powerschool__course_enrollments` for free through the existing `star()`. Two of the five sites now read it. The other three cannot yet: two are student-grained, and `int_extracts__course_schedule_by_term` sits behind a cross-SIS passthrough that needs a district prod rebuild first.

**The equivalence of those definitions is measured, not assumed.** Sumner is school number 179905, and the academic-year columns agree:

| Check | Rows | Disagreements |
| --- | --- | --- |
| `school_level_alt`, term year vs enrollment year, all 3 NJ districts | 753,825 | **0** |
| `half_total_point_value` vs `totalpointvalue / 2` | 837,441 | **0** |

**The `coalesce(islate, 0)` trio deliberately stayed put.** They look join-free, but that zero absorbs a LEFT JOIN miss rather than a null inside an existing row. Coalescing upstream would fix only the second case and leave the same logic in two places.

**The enrollment input changes model, not values.** In kipptaf, `base_powerschool__course_enrollments` is a passthrough over a model spanning both PowerSchool and Focus. In the package it is the package's own model. Every column this model reads exists on both, and the row counts match exactly.

**The move itself is row-for-row identical to prod**, per district and per academic year:

| District | New rows | Prod rows | Years | Every year's fingerprint matched |
| Newark | 18,170,523 | 18,170,523 | 11 | yes |
| Camden | 5,047,790 | 5,047,790 | 11 | yes |
| Paterson | 285,709 | 285,709 | 4 | yes |
| **Total** | **23,504,022** | **23,504,022** | | |

**Staging is already seeded.** A new package model has no staged copy, and neither do package models that gain a column, so dbt Cloud CI could not otherwise resolve them. `dbt build --target staging` ran for all four affected models in all three NJ districts. The one warning in each is `no_studyear_course_overlap`, the documented #3900 `stg_powerschool__cc` double-write, which a column add cannot cause.

**No Miami and no archive rebuild.** `kippmiami` does not consume the `powerschool` package, and this model is NJ-only by design per #4996.

### General

### dbt

- [ ] Include (or update) an
      summary above.
      See

Notes on the three unchecked boxes. No exposure applies — these are intermediate and base models, and the reporting views above them are unchanged. No external source changes, so `stage_external_sources` does not apply; the staging seeds this pull request needed were plain `dbt build --target staging` runs, already done. The breaking-change box is checked because two columns are removed and the coordination was done, though none of the affected models is contract-enforced.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Spec and plan are on this branch under `docs/superpowers/`. The upstream-calcs commit came after the spec and is not described there.

**The five deletions in the move.** The package model is the kipptaf original minus exactly five things, verified by diff — nine hunks, all deletions, zero additions, no reordering: `a._dbt_source_project` from the `scores` select list, `_dbt_source_project` from the `assignment_coding` select list, the two `and a._dbt_source_project = <alias>._dbt_source_project` join predicates, and the `region` and `credit_type` pairs in both CTEs. Column order is untouched so sqlfluff ST06 still passes.

**`if(condition, true, false)` is load-bearing, not redundant.** Do not simplify any of these to the bare condition. `score_entered` is null on 1,834,662 prod rows, and `if(null, true, false)` is `false` while the bare expression is `null`. Verified: on every one of those rows `assign_score_above_max` reads `false`, never null.

**The six `assign_*` category flags are not duplication.** Six distinct business rules producing six separately-named output columns, and `int_powerschool__gradebook_assignment_scores_rollup` sums all six by name. No consolidation preserves the output contract.

**`credit_type` detail.** The rewrite fires on 4,894 score rows, all Paterson: `ELA` 1,640, `Math` 1,579, `Homeroom` 988, `Science` 687. `Language` and `ELASkill` survive unrewritten in prod, so the conform is not exhaustive today either. `int_tableau__gradebook_audit_categories_teacher` joins `s.credit_type = e1.credit_type` against a scaffold carrying the conformed value, so leaving a raw `credit_type` on the package model would have made that join under-match if the cluster were re-enabled.

**The five `school_level_alt` definitions, before this PR.** `int_students__school_directory:77` matched `school_short_name = 'Sumner'`; `int_extracts__student_enrollments:421` matched `school_abbreviation = 'Sumner'`; `int_extracts__course_schedule_by_term:99` matched `sections_schoolid = 179905` on `terms_academic_year`; the gradebook-scores model matched `cc_schoolid = 179905` on `cc_academic_year`; and `int_tableau__gradebook_audit_teacher_scaffold:26` matched `school_abbreviation = 'Sumner'` with **no academic-year guard at all** and returned `null` rather than `school_level`, recovered downstream by a `coalesce`. That last one is disabled, so rewriting it to read the shared column is hygiene against a wrong re-enable, not a live fix.

**Why `int_extracts__course_schedule_by_term` was left alone.** It reads kipptaf's `base_powerschool__sections`, which is a compatibility passthrough over `int_students__course_sections` — a cross-SIS union. The new column does propagate there through that model's `sec.*`, null-filled on the Focus branch, but only after the districts rebuild prod. Collapsing that fifth site is a follow-up, not something to land blind here.

**Making all five agree permanently** wants the override expressed as data — a keyed `(school, grade, year)` lookup both student-grained and section-grained models can read. That is what the two `TODO: figure out a better way to track these` comments are asking for, and it is out of scope here.

**Verification method for the move.** No district checkout has a `target/prod` manifest, so `--defer --favor-state` was unavailable, and a dev build would have been wrong anyway because kipptaf `source()` resolves to personal copies under `target=dev` and `--favor-state` governs `ref()` but not `source()`. Instead the package model was compiled at `--target prod` per district, its compiled SQL run directly as a CTE, and both sides fingerprinted per academic year with `sum(cast(farm_fingerprint(to_json_string(struct(...))) as bignumeric))` over the 38 non-grouping columns. The `bignumeric` cast is required: a plain `sum(farm_fingerprint(...))` overflows at these row counts. Because every year matched, the `EXCEPT DISTINCT` drill-down never ran and no rows were sampled, so no PII was handled.

**A trap worth recording for the upstream-calcs commit.** Comparing the freshly seeded `zz_stg` copy against prod is NOT a valid equivalence check — it measures staging staleness, not the change, and it produced convincing-looking row-count gaps plus a missing AY2026 on the first attempt. The two moved derivations were instead compared directly against prod relations, which is unconfounded: 0 disagreements on 753,825 rows for `school_level_alt` and 0 on 837,441 for `half_total_point_value`. The row-producing joins are untouched by that commit, so row counts cannot move from it.

**A caveat for anyone re-running the move's equivalence check.** The clean Camden and Paterson result depended on timing. The prod kipptaf model reads through to `kipptaf_students.int_students__course_enrollments`, which can lag the district relations. It had rebuilt after all three, so both sides saw the same population. Re-running inside an open staleness window could show a small delta that is not a regression — compare `last_modified_time` before concluding otherwise.

**PII tags** carry over exactly as they were: `contains_pii: true` on `student_number` and `teacher_name`, nothing else. Widening to the full tier-3 set in `.claude/rules/ferpa-pii.md` is a separate decision, deliberately out of scope.

**Known deferred minors.** Five column descriptions carry migration history ("Renamed from ...") where `.claude/rules/dbt-yaml.md` wants an inline SQL comment; the identical strings already ship on `main`, and a `description:` edit marks nothing `state:modified`, so they cannot land here. Eighteen wrapper column descriptions diverge from the package copy (eleven empty, seven terse), also all pre-existing; rather than duplicate 39 descriptions across two projects, the wrapper's model description names the package model as the authoritative reference.

</details>


cbini:
## Review record

A whole-branch review ran before this was handed over, plus a per-task review on each of the two code tasks. Both task reviews came back clean. The whole-branch review returned 1 Important and 5 Minor findings and a verdict of safe to merge. Two were fixed in `4d75437f8b`; four are parked with reasons. A scoped re-review confirmed the fixes and found no new defects.

### Fixed

**`_dbt_source_relation` was emitted but undeclared.** `union_relations` adds that column and the old hand-written model never produced it, so the wrapper's properties yml documented 40 of its 41 output columns. Now declared, matching `int_powerschool__section_teachers`.

**The rule-file edit went further than my evidence.** I had rewritten `.claude/rules/dbt-yaml.md` to say a `--target prod` build is not classifier-blocked, citing #5278. My only first-hand evidence this session was a `--target staging` build. That edit loosened a production-write guardrail on secondhand information, so it is walked back: the instruction is again to hand prod runs to the user, and the #5278 report is marked secondhand and explicitly not permission.

### Parked, and why

**Wrapper column descriptions diverge from the package copy — 18 of them, not the 3 first reported.** Eleven wrapper columns carry no description and seven carry terse text. Every one of those strings already ships on `main`, so none is a regression here. Rather than copy 39 descriptions into a second file that will drift out of sync, the wrapper's model description now names the package model as the authoritative column reference. That is what `src/dbt/kipptaf/CLAUDE.md` already prescribes for a passthrough.

**Five column descriptions carry migration history** ("Renamed from ...") where `.claude/rules/dbt-yaml.md` wants an inline SQL comment. The identical strings are on `main` today, and a `description:` edit does not mark a model `state:modified`, so this cannot land on this branch.

**The spec asked for the three consumers to be row-compared against prod; they were compiled, not row-compared.** Each consumer is a deterministic function of the wrapper, and the wrapper is identical to prod on all 39 shared columns across all 23,504,022 rows, so consumer identity follows. Recording it rather than claiming the check ran.

**Verification used per-year row counts plus fingerprint equality instead of a two-way `EXCEPT DISTINCT`.** Deliberate. Equal counts and equal per-year fingerprints prove equality at a fraction of the cost on a 23.5M-row model, and the `EXCEPT DISTINCT` drill-down was kept as the path to take on any mismatch. No year mismatched, so it never ran.

### One note for whoever re-runs the equivalence check

`sum(farm_fingerprint(...))` overflows at this row count. Accumulate as `bignumeric`. And the clean Camden and Paterson result depended on timing: the prod kipptaf model reads through to `kipptaf_students.int_students__course_enrollments`, which can lag the district relations. It had rebuilt at 21:04 UTC, after all three, so both sides saw the same population. Re-running inside an open staleness window could show a small delta that is not a regression — compare `last_modified_time` before concluding otherwise.

PR #5284, opened 2026-09-13
fix(fldoe): match the 2026-27 FAST files and ingest grade 9 ELA

### LINKED ISSUE #5283: fix(fldoe): Florida changed 3 things in the 2026-27 FAST file spec, so no PM1 file lands
## What's happening

I placed the 2026-27 Progress Monitoring 1 (PM1) files for FAST (Florida Assessment of Student Thinking) on Couchdrop at `/data-team/kippmiami/fldoe/fast/SY27/PM1`. I expected `kippmiami__couchdrop__sftp_asset_job_sensor` to pick them up and materialize `kippmiami/fldoe/fast`. The sensor skipped every tick instead. No file landed.

The Florida Department of Education (FLDOE) changed 3 separate things for 2026-27. Each one blocks a different stage of the load. I verified all 3 against the 13 files that are on Couchdrop right now.

**1. The file name prefix lost its hyphen.** The prefix went from `KIPPMIAMI-LIBERTYCITY` to `KIPPSOUTH`. The asset matches file names with `\w+-\w+_`, which requires exactly one hyphen. `KIPPSOUTH` has none, so the pattern matches 0 of the 13 files. This is the only reason the sensor is quiet today.

**2. FLDOE moved the ordinal to the front of the standards columns.** Last year a column read `Category`. This year the same column reads `1. Category`. The same change hit `Benchmark`, `Points Earned`, and `Points Possible`. That is 160 columns in each English Language Arts (ELA) file and 140 to 144 columns in each mathematics file. The column counts per file did not change at all, so FLDOE renamed these columns and restructured nothing.

**3. Grade 9 FAST ELA Reading is new.** FLDOE sent `KIPPSOUTH_Grade9FASTELAReading_StudentData_164859 PM.csv`. Grade 9 does not exist in the asset partitions, the Avro schema, or `stg_fldoe__fast`. FLDOE sent no grade 9 mathematics file.

Change 1 and change 2 are renames. Change 3 adds data we have never loaded.

The same `\w+-\w+` pattern sits in the `science` and `eoc` assets. Those assets will break the same way on their next drop from the renamed entity.

## Steps to reproduce

1. List `/data-team/kippmiami/fldoe/fast/SY27/PM1` over SFTP with the `ssh_couchdrop` credentials. 13 files are present, modified 2026-09-13 20:48 UTC.
2. Match each file path against the `remote_dir_regex` and `remote_file_regex` metadata on the `kippmiami/fldoe/fast` asset. 0 of 13 match.
3. Read the header row of `KIPPSOUTH_Grade3FASTELAReading_StudentData_164859 PM.csv`. Slugify each column the way `dict_reader_to_records` does. 160 of the 184 columns are absent from `FAST_SCHEMA`.
4. Repeat step 3 against `/data-team/kippmiami/fldoe/fast/SY26/PM1`. 0 columns are absent from `FAST_SCHEMA`. The 2025-26 files are the control.

## Where

- **Code location / dbt project:** `kippmiami` / `kippmiami`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** `kippmiami__couchdrop__sftp_asset_job_sensor` has skipped every tick since the files landed. The skip reason reads `Sensor function returned an empty result`. The sensor reports no error.

<details>
<summary>For Claude</summary>

> Context, not instructions.

**Files**

- `src/teamster/code_locations/kippmiami/fldoe/assets.py` — the `fast`, `science` and `eoc` assets, their partitions, and `slugify_replacements`.
- `src/teamster/libraries/fldoe/schema.py` — `FAST`, `FLDOECore`, `FLDOECategories`.
- `src/dbt/kippmiami/models/fldoe/staging/stg_fldoe__fast.sql` — the `coalesce` lists over per-grade score columns.
- `src/dbt/kippmiami/models/fldoe/intermediate/int_fldoe__fast_standard_performance_unpivot.sql`.

**Evidence, measured 2026-09-13**

| File | Columns | Columns absent from `FAST_SCHEMA` |
| --- | --- | --- |
| `KIPPSOUTH_Grade3FASTELAReading` | 184 | 160 |
| `KIPPSOUTH_Grade3FASTMathematics` | 165 | 140 |
| `KIPPSOUTH_Grade6FASTMathematics` | 168 | 144 |
| `KIPPSOUTH_Grade9FASTELAReading` | 184 | 163 |
| `KIPPMIAMI-LIBERTYCITY_Grade3FASTELAReading` (SY26 control) | 184 | 0 |

Every SY26/PM1 file reports 0 absent columns. Column counts match SY27 exactly, grade for grade and subject for subject.

The only columns new in SY27 outside the renamed standards block are these 3:

```
grade_9_fast_ela_reading_achievement_level
grade_9_fast_ela_reading_percentile_rank
grade_9_fast_ela_reading_scale_score
```

The only columns that SY26 had and SY27 lacks are these 4: `category`, `benchmark`, `points_earned`, `points_possible`.

**Suspected fix for change 1**

The prefix is not a capture group, so nothing reads it. Replace `\w+-\w+_` with `.+?_` in the `fast`, `science` and `eoc` `remote_file_regex` values. That survives the next rename without another edit.

**Suspected fix for change 2**

`FLDOECategories` already declares `category_1` through `category_44`, `benchmark_1` through `benchmark_44`, `points_earned_1` through `points_earned_44`, and `points_possible_1` through `points_possible_44`. The new files reach ordinal 40 at most. So `slugify_replacements` entries that map `"1. Category"` to `"category_1"` land the new columns on schema fields that already exist. No schema change is needed for this part. The `fast` asset already uses `slugify_replacements` for the performance columns, so this follows the pattern that is there.

**Open question that changes the size of change 2**

SY26 named all 40 standards blocks identically: every block used the bare words `Category`, `Benchmark`, `Points Earned`, `Points Possible`. `csv.DictReader` keeps only the last value when keys repeat. So SY23 through SY26 may have loaded 1 block per file and left `category_1` through `category_44` empty for 4 years. Confirm against the warehouse before designing the fix. If that is true, then fixing change 2 does not restore parity. It loads 40 times more standards data than we have ever held, and `int_fldoe__fast_standard_performance_unpivot` has been reading 1 block.

**Suspected fix for change 3**

Grade 9 needs 3 Avro schema fields, a partition key, and 3 more entries in each `coalesce` list in `stg_fldoe__fast`. Note that the `grade_level_subject` partition is a cross product of grades and subjects. Adding 9 to the grade list also mints `Grade9FASTMathematics`, which FLDOE does not send. Build the list explicitly instead of extending the range.

**What fixed looks like**

All 13 SY27/PM1 files materialize. `check_avro_schema_valid` passes on each. `stg_fldoe__fast` returns rows for `academic_year = 2026`, `administration_window = 'PM1'`, including grade 9 ELA.

</details>

anthonygwalters:
## Summary & Motivation

When merged, this pull request will let the couchdrop sensor pick up the 2026-27 FAST files that are sitting on Couchdrop right now. The sensor matches 0 of the 13 files today, so no Progress Monitoring 1 (PM1) score has landed.

The Florida Department of Education (FLDOE) changed 3 things for 2026-27. This pull request handles all 3.

The file name prefix lost its hyphen. `KIPPMIAMI-LIBERTYCITY` became `KIPPSOUTH`, and the asset matched file names with `\w+-\w+_`, which requires exactly one hyphen. The prefix is not a capture group, so nothing reads it. I widened it to `[^/]+?_`, which does not care how the entity is named. The `science` asset carried the same pattern and would break on its next drop, so it is widened too.

FLDOE moved the ordinal to the front of the standards columns. `Category` became `1. Category`, and the same change hit `Benchmark`, `Points Earned` and `Points Possible`. `FLDOECategories` already declares `category_1` through `category_44`, so these map straight onto fields that exist. The new files reach ordinal 40 at most.

Grade 9 FAST ELA Reading is new. FLDOE sends no grade 9 mathematics file, so grade 9 is listed as its own partition key rather than widened into the grade and subject cross product. `stg_fldoe__fast` holds grade 9 back from every downstream model for now.

Closes #5283

## Reviewer Notes

**Grade 9 stops at staging on purpose.** `stg_fldoe__fast` feeds `int_reporting__promotional_status`, `rpt_tableau__miami_fast`, `rpt_deanslist__state_test_scores`, `rpt_gsheets__kippfwd_miami_roster`, `dim_assessments` and `dim_assessment_administrations`. None of them expect a grade 9 row. Dagster now ingests the raw grade 9 file, and one line in `stg_fldoe__fast` keeps it out of all 6. Lifting that filter is a separate decision with its own dashboard work.

**The standards replacements must run highest ordinal first.** slugify applies `replacements` as plain string substitutions in list order. Ascending order rewrites the `1. ` inside `11. Category` and yields `1category_1`. The column then lands somewhere nothing reads and no error fires. `tests/test_fldoe_slugify.py` pins the ordering for all 44 ordinals.

**Mapping the standards columns is not strictly required, and I did it anyway.** `check_avro_schema_valid` carries WARN severity, and the Avro writer drops unknown keys, so those 160 columns would have cost only a warning. Leaving it would have made `avro_schema_valid` warn on every FAST materialization from now on. That check exists to catch FLDOE changing the spec, so a permanently yellow check is worse than the 8 lines.

**One risk I could not verify.** `src_fldoe__fast` is an Avro external table, and this pull request adds 3 fields to the Avro schema. The 2026-27 files are not in the prod bucket yet, so I could not test how BigQuery reads a bucket holding both 294-field and 297-field files. The schema has grown this way before, which is why `fast_grade_3_*` and `grade_3_fast_*` both exist, so there is precedent. Worth a look at the first prod materialization.

**No backfill needed.** The sensor advances its cursor only on matched files, so the cursor never moved past these 13. The next tick after the code location deploys picks them all up.

### General

### Dagster

- [ ] `uv run dagster definitions validate -m teamster.code_locations.kippmiami.definitions` — **not ticked on purpose.** It fails in my codespace on `main` as well as on this branch, because `dlt/focus/assets.py` resolves `FOCUS_DB` credentials at import and this shell does not hold them. The failure happens before any FLDOE code loads. I verified the asset another way, below.
- [ ] `uv run pytest tests/test_dagster_definitions.py` — fails for the same pre-existing reason, on `main` too.
- [x] New integrations follow the Library + Config pattern — no new integration, existing asset builders unchanged.

### dbt

- [x] Properties file for all new models — no new models.
- [x] Exposure for models consumed by a dashboard — no new models, and grade 9 does not reach any dashboard.
- [x] **Breaking change?** No. The change to `stg_fldoe__fast` adds a `WHERE` condition and nothing else, so the contracted column set is untouched. I built the model to confirm the contract still passes.
- [x] External source — `sources-external.yml` is unchanged, so no re-stage is needed for the source definition.

## CI checks

- [x] **Trunk** — `trunk check --force --no-fix` over all 5 changed files reports no issues.
- [ ] **dbt Cloud** — note that dbt Cloud CI builds kipptaf only, so it will not build this kippmiami model. I built it locally instead, as described below.
- [ ] **Dagster Cloud** — expected to run, since `src/teamster/**` changed.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The investigation, the design and the verification below were run by Claude; the scope decision to hold grade 9 at staging was the author's.

**Verification actually run**

Materialized 3 real SY27/PM1 partitions against the files on Couchdrop, through `materialize()` with the live `ssh_couchdrop` resource and a test GCS IO manager:

| partition | records | `avro_schema_valid` extras |
| --- | --- | --- |
| `Grade3FASTELAReading` | 186 | `''` |
| `Grade6FASTMathematics` | 171 | `''` |
| `Grade9FASTELAReading` | 92 | `''` |

An empty `extras` means 0 columns fell outside `FAST_SCHEMA`, which is the end-to-end proof that the regex, the standards mapping and the grade 9 partition key all work together.

Partition space after the change: 195 keys, up from 180. `grade_level_subject` holds 13 keys. `Grade9FASTELAReading` is present and `Grade9FASTMathematics` is absent, which is the point of listing grade 9 explicitly instead of widening the cross product.

dbt: `dbt build --select stg_fldoe__fast --target dev` created the table with 19,006 rows and the contract passed. Prod holds 19,006 rows. The counts match exactly, so the new filter is provably a no-op on every row that exists today. Prod `max(assessment_grade)` is 8, so there was never a grade 9 row for the filter to remove.

**Why the standards columns were never load-bearing**

`category_1` through `category_44`, and the matching `benchmark_*`, `points_earned_*` and `points_possible_*` families, hold 0 rows for SY23 through SY26. SY23 through SY26 named all 40 standards blocks identically, and `csv.DictReader` keeps only the last value when keys repeat, so 1 block per file ever landed. No dbt model reads any of those columns. `int_fldoe__fast_standard_performance_unpivot` reads only the 18 `*_performance` columns. So this pull request populates fields that were previously empty, and changes no downstream number.

**Known pre-existing failures, unrelated to this branch**

`tests/test_dagster_utils.py::test_ghseet_asset_key_rename` fails on `main`. In a fresh worktree it fails earlier, on a missing `src/dbt/kipptaf/target/manifest.json`.

</details>


anthonygwalters:
Verdicts on the review, one per finding. I verified each claim against the code rather than taking it as given.

**1. Duplicated `STANDARDS_LABELS` in the test — declined.**

The duplication is deliberate, and I added a comment saying so. If the test imports the label tuple from `assets.py`, then the test and the code under test share one definition, and renaming a label in production silently redefines what the test asserts. The test would still pass. Spelling the labels out in the test is what makes a production rename fail. The forcing function the finding asks for already exists in the other direction: adding a 5th label to production without touching the test leaves that label untested, which is the weaker of the two failure modes.

**2. Redundant parametrized test — accepted, fixed in 1c838ea.**

Correct, and I confirmed it. `test_every_ordinal_maps_onto_an_existing_schema_field` loops ordinals 1 through 44 across all 4 labels and asserts the same equality plus schema membership, so all 6 parametrized cases sat inside it. I deleted the parametrized test and kept the module docstring, which is where the explanation of the ordering trap belongs.

**3. The filter is a no-op today — agreed, and that is the intent.**

The trace is right. `coalesce()` lists only grades 3 through 8, so a grade 9 row's `achievement_level` is NULL, and `NULL not in (...)` already drops it. I measured the same thing from the other end: prod holds 19,006 rows with `max(assessment_grade) = 8`, and a dev build of the new SQL returned 19,006 rows. The line exists so that whoever widens the `coalesce` lists has to decide about it.

**4. Regex change — agreed, and independently checked.**

I confirmed `re.search` at `libraries/sftp/assets.py:177`. `[^/]` is the reason this is safe, and it is why I did not use `.+?`: `.` matches `/` in Python, so a non-matching file name could let the prefix swallow a directory boundary. `[^/]` cannot.

**5. Descending ordinal order — agreed.**

**6. Model altitude — agreed.**

**7. dbt SQL conventions — agreed.**

**8. `dagster-expert` plugin — not reproduced on my side.**

The skill loaded and returned its reference index in my session. This looks like a difference in the review job's plugin set rather than something this pull request should change.

---

**Correction to my own risk note, prompted by finding #1's pointer.**

My "one risk I could not verify" section is vaguer than it needed to be. `.claude/rules/dbt-yaml.md` documents this exact trap, so here is the precise version.

Avro external tables autodetect their schema from the last alphabetical file. `SY27` sorts last among `SY23` through `SY27`, so the declared schema will pick up the 3 new `grade_9_*` fields once a 2026-27 file lands. That part is fine.

The real behavior is schema heterogeneity, and it is deterministic rather than a maybe. A query scanning both old 294-field files and new 297-field files resolves one Avro reader schema and drops the new field for the whole scan, including the new files that hold it. `stg_fldoe__fast` scans full partition history, so the 3 `grade_9_*` columns will read NULL everywhere until every partition is re-encoded with `scripts/reencode_avro_partitions.py`.

This does not affect this pull request, for the reason the review gives: `stg_fldoe__fast` never selects those columns. It does affect the follow-up that surfaces grade 9, which now has a named prerequisite rather than a vague warning.

The renamed standards columns are not exposed to this at all. `category_1` through `category_44` and their siblings already exist in the old files' schema, which is why a query for `category_1` across SY23 through SY26 returns rows rather than an error. Only the 3 grade 9 fields are new.

cbini (approved):

PR #5286, opened 2026-09-14
refactor(kipptaf): read the shared school_level_alt in course schedule by term

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will delete the last collapsible copy of the Sumner grade-5 school-level override, so two more models read the single definition #5281 put in the `powerschool` package.

One business rule — KIPP Sumner Academy, grade 5 and above, from academic year 2025, reads MS — was written out by hand in 5 places, matching the school 3 different ways and the year 2 different ways. #5281 defined it once on `base_powerschool__sections` and collapsed 2 of the 5. This collapses a third.

Both models also stop reading the `base_powerschool__sections` compatibility passthrough and read `int_students__course_sections` directly. That drops the passthrough from 8 consumers to 6, which is the direction #3999 wants.

Refs #5012
Refs #3999

## Reviewer Notes

**The passthrough swap is required, not tidying.** `base_powerschool__sections` in kipptaf is a `select *` view over `int_students__course_sections`. BigQuery fixes a view's column list when the view is created, and Dagster rebuilds a view only when its own SQL changes — so that view can never pick up a column an upstream adds. This is not a prediction. After #5281 merged and everything upstream rebuilt, the column reached the districts, `int_powerschool__sections_union`, and `int_students__course_sections`, and the view still does not have it:

| Relation | Kind | Has `school_level_alt` |
| --- | --- | --- |
| `kippnewark` / `kippcamden` / `kipppaterson` `base_powerschool__sections` | table | yes |
| `kipptaf_powerschool.int_powerschool__sections_union` | table | yes |
| `kipptaf_students.int_students__course_sections` | table | yes |
| `kipptaf_powerschool.base_powerschool__sections` | **view** | **no** |

So reading the shared column required moving off the view. Their column sets were verified identical before the swap, in both directions, so nothing else changes.

**This also fixes a latent bug from #5281.** That PR pointed `int_tableau__gradebook_audit_teacher_scaffold` at `school_level_alt` on the frozen view — a column it could never have had. The model is disabled so nothing broke, but the reference was wrong.

**The definition it replaces was the divergent one.** `int_tableau__gradebook_audit_teacher_scaffold` matched on `school_abbreviation` with **no academic-year guard at all** and returned `null` rather than `school_level`, recovered downstream by a `coalesce`. `int_extracts__course_schedule_by_term` matched `sections_schoolid = 179905` on `terms_academic_year`. Both now read the same column.

**Staging was refreshed for this.** The deferred relations were stale after the #5281 merge, so `Clone - Staging (On-Demand)` ran (run 70403224737407) and `school_level_alt` is now on `zz_stg_kipptaf_students.int_students__course_sections`. Without that, CI would have failed on `Name school_level_alt not found`.

### General

### dbt

- [ ] Include a `[model_name].yml` properties file for all new models — see
- [ ] Include (or update) an
      summary above.
      See

No new models, so no properties file and no exposure. No columns renamed or removed — `school_level_alt` keeps its name and its values on both models. No external source changes; the staging refresh this needed was the `Clone - Staging` job, already run.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Follow-up to #5281, which is where the package-side definition landed.

**The 5 definitions and where they stand.** Before #5281: `int_students__school_directory:77` on `school_short_name = 'Sumner'`; `int_extracts__student_enrollments:421` on `school_abbreviation = 'Sumner'`; `int_extracts__course_schedule_by_term:99` on `sections_schoolid = 179905` and `terms_academic_year`; `int_powerschool__gradebook_assignments_scores` on `cc_schoolid = 179905` and `cc_academic_year`; `int_tableau__gradebook_audit_teacher_scaffold:26` on `school_abbreviation = 'Sumner'` with no year guard and a `null` else-branch. #5281 collapsed the gradebook-scores model. This collapses course-schedule-by-term and the teacher scaffold. The remaining 2 are student-grained and cannot read a section column.

**Making the last 2 agree** wants the override expressed as data — a keyed `(school, grade, year)` lookup both student-grained and section-grained models can read. That is what the `TODO: figure out a better way to track these` comments on those 2 models are asking for. Out of scope here.

**Equivalence was measured on #5281, not assumed.** Sumner is school number 179905, confirmed against `stg_powerschool__schools` (one matching row in `kippcamden_powerschool`, zero in Newark and Paterson). The academic-year columns agree: `terms_academic_year`/`sections_schoolid` against `cc_academic_year`/`cc_schoolid` produced 0 disagreements across 753,825 enrollment rows in all 3 NJ districts.

**Why the swap targets `int_students__course_sections` and not the passthrough.** The passthrough's own comment says "Delete it once they have", referring to the #3999 migration. Its column set and the target's were compared in both directions and are identical, so the swap is content-neutral today. The frozen-view behavior is documented in `.claude/rules/dbt-sql.md` under `select *` and UNION branches, and this is a live instance of it.

**Ordering constraint for anyone reproducing this.** The change could not be pushed until `school_level_alt` reached both prod and the Staging environment CI defers to. Prod propagated on its own after the #5281 merge; Staging did not, because `Clone - Staging (On-Demand)` has `schedule: false` and `Build - Staging (Modified)` only builds modified models. Merging before Staging caught up would have failed CI, and merging before prod caught up would have stranded the prod view, since BigQuery validates view SQL at read time.

</details>


cbini:
## Verdicts on the review findings

Both items marked "needs local verification" are now verified against the warehouse, and the precedent the reviewer cited on finding 1 checks out — which resolves that finding the other way.

### 1. Stale `source_model` pointers — not a defect, no change

The reviewer flagged 21 `source_model: base_powerschool__sections` pointers as stale now that the `from` reads `int_students__course_sections`, at low-medium confidence, and named a precedent that argues the other way. That precedent is real. Verified in the same file:

`quarter` and `semester` (lines 229 to 244) declare `source_model: int_powerschool__terms`, while the SQL has joined `int_students__terms` since before this PR (line 41).

So `config.meta.source_model` in this file tracks the **origin model in the source system**, paired with `source_system: PowerSchool`, not the literal immediate `ref()`. Read that way the pointers are correct — and for `school_level_alt` specifically the pointer is now *more* accurate than before, because that column genuinely originates in the package `base_powerschool__sections`, which is where #5281 defined it. Note there are two models with that name; the annotation names the package one.

Rewriting 21 pointers to `int_students__course_sections` would break the file's own convention and lose the origin-system information. Leaving them.

### 2. Convention compliance — agreed, no action

### 3. Focus and Miami rows — confirmed, no behavior change

The reviewer traced this statically and asked for warehouse confirmation. Run against prod `kipptaf_students.int_students__course_sections`:

| Region | Rows | Non-null `school_level_alt` |
| --- | --- | --- |
| kippmiami | 4,276 | **0** |
| kippnewark | 22,697 | 22,695 |
| kippcamden | 6,318 | 6,316 |
| kipppaterson | 532 | 532 |

Miami null-fills exactly as predicted, so the Focus branch is unaffected. The 2 nulls each in Newark and Camden are rows whose underlying `school_level` is null, which predates this change.

### 4. Contract type — confirmed

`school_level_alt` is `STRING` on `kipptaf_students.int_students__course_sections`, matching the `data_type: string` both affected models declare.

The build itself is the stronger evidence: CI built `int_extracts__course_schedule_by_term` into the PR schema and the built relation carries `school_level_alt`, with 0 rows across 4 test-audit tables.

### 5. Ponytail pass — agreed, no action

No code changes result from this review.

PR #5304, opened 2026-09-14
feat(extracts): grant Miami School Leaders in Residence region-wide observer scope

cristinabaldor:
## Summary & Motivation

When merged, this pull request will let Miami School Leaders in Residence enter observations for staff at any school in their region.

They cannot today. Grow roles and school scope come from ADP through `rpt_schoolmint_grow__users`, and a School Leader in Residence is homed at Room 11, the Miami regional office, rather than at a campus. Their `School Admin` role is scoped to their home location, so their observation picker holds no teachers. Nobody can fix this in the Grow interface either: the nightly `grow_user_sync` writes the ADP-derived state back over any manual change.

This grants them `Regional Observer` and scopes it to every active Grow school in their region. `Regional Observer` is the one role that carries multi-school scope and stays writable, so they can create observations rather than only read them.

The job-title match is deliberate and temporary. It goes away once People Operations populates the ADP `Principal in Residence` membership for these staff, which is the same membership feed the New Teacher Network Coordinator rule already reads. That work is tracked in Asana.

Newark School Leaders in Residence are unchanged. They sit in the same position, but this change is scoped to Miami by request.

## Reviewer Notes

The model bans job-title matches by design, with one documented exception for Associate Director. This adds a second exception, on purpose and with a comment saying when it is removed. The alternative was gating on `job_function = 'School Leader'`, which would hand every principal region-wide scope.

`regional_admin_school_ids` now means two different things for `Regional Observer`: the user's own school for a coordinator, every school in the region for a Miami School Leader in Residence. The `observes_whole_region` flag is what picks between them, and both branches stay under one `where p.is_regional_observer`.

Worth confirming with SchoolMint: no writable user has ever held more than one school in `regionalAdminSchools`. Regional Admins hold a whole region but are `readonly`, and Regional Observers have held exactly one school. The Grow behavior we are relying on is documented in the model but not yet observed in production.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — the existing properties file is updated, no new model
- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — no exposure change, the consumer is the `grow_user_sync` asset
- [x] **Breaking change?** No column is added or removed, so `grow_user_sync` needs no change
- [x] If adding or modifying an external source, run `stage_external_sources` — no source change

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

<details>
<summary>For Claude</summary>

Three changes in `rpt_schoolmint_grow__users.sql`:

1. `staff` gains `is_school_leader_in_residence`, a `coalesce`-wrapped match on `job_title = 'School Leader in Residence'` and `home_work_location_dagster_code_location = 'kippmiami'`. The `coalesce` matches the `is_teacher` idiom four lines up. It changes no output: every consumer of `is_regional_observer` treats null and false identically, so a null `job_title` produces the same extract either way.
2. `people` gains `observes_whole_region`, and `is_regional_observer` becomes the coordinator flag `or` the new flag.
3. The `Regional Observer` branch of `regional_scope_schools` joins on `if(p.observes_whole_region, p.region = gs.region, p.school_name = gs.school_name)` instead of the school-name equality alone.

Verification, all run before the commit:

- `dbt test --select "rpt_schoolmint_grow__users,test_type:unit"` — 8 pass, 0 fail. That is the 7 pre-existing tests plus the new `unit_grow_users_miami_slr_region_scope`, which pins three cases: a Miami School Leader in Residence gets `Regional Observer` plus region-wide scope, a Newark one gets neither, and a Miami school leader at a campus gets neither.
- The compiled model was run against production data and filtered to the job title. The 3 active Miami holders come out with `Regional Observer` added, 8 schools in `regional_admin_school_ids`, `readonly` 0, `group_type` `observers`. The 3 Newark holders present in the extract are byte-identical to their current rows.
- `trunk check --force --no-fix` on both changed files: clean after the commit hook reformatted one block scalar in the YAML.

Two things found while investigating, neither fixed here:

- One of the 4 active Newark School Leaders in Residence has no `user_principal_name` in ADP, so `people` filters them out and they have no Grow account at all. That is a People Operations data gap, flagged in the Asana task.
- Newark School Leaders in Residence also cover Paterson per `rpt_clever__staff`, but Paterson is excluded from this model entirely, so region-wide Newark scope would still reach nothing there. Relevant if the Miami-only restriction is lifted later.

</details>


cristinabaldor:
Verdict per finding. Fixes are in 5c42bcdab.

**1. Job-title match doesn't confirm "homed at the regional office" — no change, by design.**

Checked prod: all 3 active Miami holders resolve to Room 11, so the case does not occur today. More to the point, the requirement is title-based, not location-based: a School Leader in Residence should be able to observe anyone in the region. The regional-office homing explains why they cannot today; it is not a condition on the grant. A campus-homed holder getting region-wide scope stacked on their School Admin scope is the intended outcome, not an overreach. The inline comment is trimmed so it no longer reads as if Room 11 were part of the predicate.

**2. Migration plumbing in the YAML `description` — fixed.**

`.claude/rules/dbt-yaml.md:395` says it directly. The "temporary, replaced by the ADP membership" sentence is out of the description and stays only in the SQL comment.

**3. `observes_whole_region` breaks ST06 plain-ref ordering — fixed.**

Moved up next to `sr.is_teacher`, ahead of the `or` expression that computes `is_regional_observer`.

**4. Possible ST06 issue in the `staff` CTE — no change.**

`trunk check --force --no-fix` passes on both files, before and after these fixes. sqlfluff buckets `coalesce(...)` and `x in (...)` together as calculations, so the order there is not a violation.

**5. The `coalesce(..., false)` wrapper changes no output — correct, and my PR rationale was wrong.**

Traced every consumer: the `or` in `people`, the role slot `if(...)`, `where p.is_regional_observer`, and the `group_type` `if(...)` all treat null and false identically, and `array_to_string` drops null elements with no `null_text`. A null `job_title` produces the same extract either way. Keeping the wrapper anyway: it matches the `coalesce(sr.job_function in (...), false) as is_teacher` idiom four lines up, and intent-clarity is the honest reason rather than behavior. Correcting the PR body to say that.

**6. Inline comment duplicates the description — fixed.**

Cut to the two things a reader of the predicate cannot see: that the title match is temporary, and what replaces it.

**7. Region-wide scope includes the regional office as a school — no change.**

Confirmed that is real prod data: Room 11 is one of the 8 active Miami schools in `stg_schoolmint_grow__schools`. It is inherited from the Regional Admin region-wide branch and not touched here, so any fix belongs in its own change.

Verification after the fixes: 8 unit tests pass, `trunk check --force --no-fix` clean on both files.

kevinverhoff (approved):

PR #5081, opened 2026-08-31
refactor(dbt): align new diverging assessments strategies for DIBELS PM

GabyRangelB:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..."

...add a new dbt model, `int_amplify__mclass__pm_student_summary_aimline`,
that combines DIBELS progress-monitoring data for grades 3-8, and fix a
`term_key` bug in `dim_terms` that the DIBELS PM grade-band scaffold work
surfaced.

**Part 1 — new aimline model.** Amplify sends two files for progress
monitoring: a base file, and a file carrying aimline goals and status.
Amplify has moved the aimline fields between these two files before,
without warning, and made no promise it will not happen again. The new
model reads both files and keeps whichever one has a value for each row,
so neither file is a single point of failure. Grades K-2 are not affected
— they keep using `int_amplify__mclass__pm_student_summary`, unchanged.
Scoped to Newark, Paterson, and Camden; wiring it into
`int_amplify__all_assessments` is a separate step, not included here.

**Part 2 — `term_key` fix.** `reporting__terms` can now carry a
`Grade Band`-tagged variant of the same `LIT` code (needed for the DIBELS
PM scaffold to give grades K-2, 3-4, and 5-8 their own rows — `PLIT` stays
K-2-only, see Part 3). `dim_terms.term_key` never included `grade_band`,
so those variants collided on the same key. Fixed by adding `grade_band`
to the hash — on `dim_terms` itself and on every other model that
independently re-derives the same hash rather than joining `dim_terms`
(12 files total; see the fold-out below for the full list). Approved by
Cristina, who owns these models.

**Part 3 — docs, skill, and canonical process notes.** Re-landed a stale
DIBELS reference doc, corrected it and the `dibels-dashboard` skill on a
real mistake (`PLIT` is K-2-only, not something to extend to grades 3-8 —
I generated it for 3-8 once before this was caught), and documented the
canonical annual rollover process (Benchmark + PM, K-2 vs 3-8) in both
places plus on #3834 itself, with two small generator scripts, so none of
this needs re-discovering next year.

**Part 4 — Expected Assessments cohort split, sheet-authored
`assessment_type`, and two data-quality fixes.** One real code change:
`round_number` in `stg_google_sheets__dibels_expected_assessments.sql` now
derives from `regexp_extract(test_code, r'LIT(\d+)')` instead of
`right(test_code, 1)`, so it no longer silently mis-parses a double-digit
[...]

PR #5309, opened 2026-09-14
fix(dibels): restore the grade 6-8 benchmark goals, and stop the generator losing rows

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will fix three things in the DIBELS goals chain and write down what nobody had written down. It grew past its original title while we walked the SY26-27 goals rollover; the title change is deliberate.

**1. Grades 6-8 get their benchmark goal back.** Academics set goals for grades 6-8 at end of year only. Grades K-5 get both mid-year and end-of-year. A beginning-of-year row is measured against the mid-year goal, so grades 6-8 matched nothing, and a filter added earlier this year then dropped those rows entirely. The dashboard has 31,076 Benchmark rows for grades 6-8 this year with no goal at all, against 29,702 populated rows for K-5. The model now uses a grade's end-of-year goal wherever that is the only goal it has.

**2. The row generator stops losing rows quietly.** The academics sheet writes grades K-5 as percentages ("57 - 61%") and grades 6-8 as decimals ("0.51"). The script divided everything by 100, so a goal of 0.51 came out as 0.0051. Separately, any cell it could not read at all was skipped with no warning, exactly like a deliberately blank cell. Both are fixed, and every run now prints a coverage table so a missing grade is visible instead of being counted by hand.

**3. `participation_group` reaches the dashboard.** The roster model works out which rows make one clean participation rate -- every Benchmark row, the internal method for K-2, aimline for grades 3-8 -- but the column stopped at the intermediate model, so Tableau could not filter on it.

Plus a missing uniqueness test on the hand-pasted foundation goals table, and the skill notes behind all of it.

## Reviewer Notes

The grades 6-8 change is the one to look at. It adds a mid-year goal row carrying the end-of-year figure wherever a grade has end-of-year and no mid-year. K-5 gains nothing, since it already has real mid-year rows. The join predicates are untouched and the staging model still copies the sheet faithfully.

This is what a person did by hand last year. In SY25-26 the At/Above filter did not exist yet, so the unmatched grade 6-8 rows survived with real counts and empty goals, and someone typed the goals in. That is how Paterson grades 6 and 7 ended up holding Newark's numbers. Expect this year's output to differ from last year's tab for Paterson -- the tab is wrong, not the model.

`participation_group` is empty on 16,971 rows, all SY23-24 progress monitoring. Those are the same rows where `completed_test_round` is already empty, because the roster holds no progress monitoring rows before SY24-25.

**After merge, BM Goals needs re-pasting** -- 74 rows now, replacing the 49 there today, because grades 6-8 are new rows rather than edits.

### General

### dbt

- [x] Properties files updated -- new contract column, new test, and a model description for the calculations model.
- [x] No breaking change. Columns are added, none renamed or removed.

## CI checks

- [x] **Trunk** -- passes.
- [x] **dbt Cloud** -- passes, zero warnings.
- [x] **Dagster Cloud** -- not triggered (no Dagster paths).

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The user caught every one of the three substantive errors below; none were found by review.

**The grades 6-8 fallback.** New `foundation_goals_by_season` CTE, unioning the staging model to an anti-joined copy that emits `'MOY' as period` wherever a `(academic_year, region, grade_level, population, grade_goal_type)` group has an EOY row and no MOY row. The three foundation joins in `roster` read the CTE; `and a.benchmark_goal_season = f.period` is unchanged. Put in the calc model rather than in `benchmark_goal_season` (which is a plain next-season map used beyond this model, and should not encode a goal-coverage quirk) or in the staging model (which must stay a faithful copy of the sheet).

Verified with `--defer --favor-state` against a fresh prod manifest: AY2026 goes 49 to 74 rows, grades 0-5 to 0-8, 25 new rows. Each new row carries exactly its region and grade's foundation EOY goal -- min equals max equals the foundation figure in all nine region-grade groups. K-5 is byte-identical: 49 rows, `except distinct` empty in both directions. Grain holds at 74 distinct of 74. Contract enforced and passing.

**How the history actually reads**, since I got it wrong twice before landing on it. The join has been `benchmark_goal_season = f.period` since the model's first commit (`3a3f32f13`, 2025-08-05), and `benchmark_goal_season` has always been the plain next-season map. So the model never derived a grade 6-8 BOY goal. What changed is `aac3e5a86`, which added `where c.grade_goal_type = 'At/Above'` -- correct in itself, since grades with both goal types were emitting duplicates. Before it, the unmatched grade 6-8 rows came through with null `grade_goal_type` and null goals but real window-function counts, which is the evidence that the model produced them: AY2025's tab has `n_admin_season_school_gl_all` and `n_admin_season_school_gl_at_above_expected` populated on all 26 of its grade 6-8 BOY rows. Nobody hand-types those. The goal column was hand-filled on top.

**The generator's scaling bug.** `parse_range` divided every parsed value by 100. The academics tab mixes formats in the same columns -- K-5 as percentage ranges, grades 6-8 as bare decimals with no percent sign -- so grade 6-8 goals came out at 0.0051 rather than 0.51. Now only real percentages are scaled: a cell carrying `%`, or one whose upper bound exceeds 1 and so cannot be a fraction. A bare `1` stays 1.0, which is 100% either way. Verified by running the generator over the live academics tab: all 90 AY2026 foundation rows reproduce prod exactly, same keys, zero value mismatches.

**The generator's silent skip.** `if parsed is None: continue` treated an unreadable cell the same as a blank one. Blank and `n/a` still skip silently -- that is what grades 6-8 hold in the MOY columns, legitimately. A non-empty cell that yields no digits now warns and quotes the cell. Also added a per-run coverage grid of rows by academic_year, grade, period and goal_type, driven off the grades present in the INPUT so a grade whose every cell failed still appears rather than vanishing. The header-row warning was suppressed because a warning that fires every run is a warning nobody reads, which matters now that warnings are the mechanism.

**Reading the academics sheet at all.** Both MCP routes are dead ends: the BigQuery service account has no Drive scope, so any Sheets-backed external returns "Permission denied while getting Drive credentials" regardless of sharing, and every Drive MCP response for this file is redacted wholesale by `check-output.sh` as containing a high-entropy string. What works is ADC from Python with `spreadsheets.readonly` and `drive.readonly` requested explicitly, writing the fetch to a file and printing only parsed numbers so the output scanner has nothing to catch. Worth knowing before anyone else tries.

**`participation_group`.** Three projections, one per union branch: `r.` in the Benchmark branch, `rs.` in the internal and aimline branches, placed after `completed_test_round_int` so each branch keeps its roster columns contiguous and the branches stay positionally aligned for `union all`. The Benchmark branch's roster join carries no `model_type` predicate, unlike the other two -- safe, because roster Benchmark rows are only ever `BM`, verified across 2023 to 2026, and always `Combo`. Dashboard row counts per `model_type` match prod exactly: BM 373,193, Internal 97,064, Aimline 75,679.

**The uniqueness test** on `stg_google_sheets__dibels_foundation_goals` keys on academic_year, region, grade_level, period, population, grade_goal_type with explicit `severity: error`, since the project default is `warn`. 263 rows, 263 distinct. The table is pasted by hand once a year, so a doubled or overlapping paste was previously uncatchable.

</details>


GabyRangelB:
@claude Thanks for the review. Verdicts on both findings below, plus one unrelated CI note.

**Finding 1 — colon-space in the `participation_group` description: declined, false positive.**

The rule you cited is real, but it scopes to *plain* (unquoted) multi-line scalars, the `description:` followed by bare wrapped text form. Its own example, `"types: parent"`, is that shape. The new description is a `>-` folded **block** scalar, and inside a block scalar there are no structural characters at all — a colon is literal text, so `: ` cannot break parsing there.

Your supporting grep is what to discard. The claim was that zero other multi-line `description: >-` blocks in `src/dbt` contain colon-space. Scanning every non-target, non-package `.yml` under `src/dbt` and reading each `description: >-` block to its dedent, **217 of them contain colon-space**, starting with `src/dbt/kipptaf/tests/properties.yml:147`. So it is not a consistently-avoided pattern; it is the norm.

The empirical side agrees, and all three ran after the YAML edit, not before:

- `.trunk/tools/trunk check --force --no-fix` on that exact file — `✔ No issues`.
- `dbt parse --target prod` — clean.
- `dbt build --select rpt_tableau__dibels_dashboard` — `PASS=1`, with `contract: enforced` validating the full column list including `participation_group`. A YAML parse failure would have stopped this before the contract check.

No change made. Worth noting your own environment note called this out — no `trunk` binary and no live parse — so the confidence label on this one was higher than the available evidence supported.

**Finding 2 — no `contains_pii` tag: declined, no action, and I agree with your reasoning.**

`Combo` / `Not Combo` records which goal method a row is evaluated under. It is methodology metadata, not an attribute of the student, and it is identical for every student in a grade band. It also matches the untagged treatment of its siblings (`model_type`, `assessment_type`, `completed_test_round`) and of the column on the roster model itself. Nothing to revisit.

**Accepted as correct:** every item in your "checks performed" list, including the ST06 placement, the positional alignment across the three `UNION ALL` branches, the `data_type: string` match against the upstream case expression, and the additive-only contract read.

**Unrelated CI note.** The `dbt Cloud` failure on this PR is not from this change. The post-merge `Clone - Staging` run (`70403224754849`) fired about a minute after #5081 merged, before Dagster had materialized the new models into prod, so it failed on 8 nodes — 5 that could not be cloned and 3 views whose contracts then failed over stale upstreams. This PR is just the first thing since to rebuild the dashboard, so it hit the missing `zz_stg_kipptaf_amplify.int_amplify__benchmark_student_summary`. All the prod relations exist now, so a fresh `Clone - Staging` clears it; that is being run separately. It currently blocks any kipptaf PR touching those 8 models, not only this one.

GabyRangelB:
@claude Both findings closed, agreed. Three corrections on the corpus count, then the CI resolution you couldn't reach.

**Your in-file evidence is better than mine, and I'm adopting it.** `completed_test_round` at `:439` and `completed_test_round_int` at `:443` use the bare plain-scalar `description:` and contain no colon; `participation_group` at `:447` uses `>-` and carries one safely. Two forms side by side in one file, behaving exactly as the rule predicts. That is the argument to keep.

**The 217 caveat does not apply to the scan that produced it.** The artifact you describe — a window-based grep catching the next sibling key (`data_type: string`, `config:`) a line or two below the block — is real, and it is why a naive multiline grep is untrustworthy here. Mine was not window-based. It read each `description: >-` line, took its indentation, and consumed continuation lines only while indentation was strictly greater, breaking at the first line whose indentation was less than or equal. A sibling key sits at exactly the description's own indent, so it terminates the block rather than joining it. Trailing keys are structurally excluded, not filtered after the fact.

**Two of your four spot-checks do contain in-prose colons.** `src/dbt/kipptaf/tests/properties.yml:147`, the example you named specifically, has two inside the folded text: `TODO(#4801): downgraded to warn for one bad Focus record` and `Restore severity: error once Ops corrects it`. And `int_pearson__all_assessments.yml` has four across its kipptaf and pearson copies, including `computed upstream: int_pearson__all_assessments in the pearson package` and `NJSLA only: levels 1-2`.

You were right about the other two. `base_powerschool__sections.yml` and `int_surveys__manager_submission_subjects.yml` have no folded block with an in-prose colon — and neither was ever in my 217. That is the tell: the files you spot-checked were flagged by your reconstruction, not by mine, which is why two of them looked like false positives.

**One thing I stated imprecisely.** 217 is the number of *blocks*, not files. Recounted: **217 blocks across 139 distinct files**. You read it as files, and the distinction matters if anyone cites the figure later.

**CI resolution.** Green. `dbt Cloud: success` on `4bac9ead` (run `70403224764486`, 1m08s), with zero test, freshness or log warnings. No code changed between the failing runs and this one — a fresh `Clone - Staging` reseeded the stale relations and the same commits passed. Both earlier failures were that staleness: the missing `int_amplify__benchmark_student_summary`, and my new uniqueness test erroring `Unrecognized name: population` because CI ran it against a staging copy sitting at 8 columns against prod's 12. My local run passed only because `--favor-state` pointed it at prod.

Worth a separate issue, not this PR: the on-merge `Clone - Staging` fires about a minute after a merge, before Dagster materializes newly added models, so any PR that adds models leaves shared staging broken until someone re-runs the clone by hand. Every kipptaf PR touching those relations fails in the meantime. This is the second time today it bit — two unrelated PRs failed during the same window.

cbini (approved):

PR #5312, opened 2026-09-14
perf(assessments): split the resolver score union into int_assessments__score_anchors

### LINKED ISSUE #5261: perf(assessments): cut the non-deduplicate cost on the two assessment nodes above 10 slot hours

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." move the six-way score union out of `int_assessments__resolved_section_enrollments` into a new cron table, `int_assessments__score_anchors`, so the resolver reads it once instead of inlining it three times.

The resolver costs 15.15 slot hours a week in prod (36 builds, 25 slot-minutes each) while scanning 0.3 GiB per build. The #5261 measurement traced that to two things. The `scores` CTE is referenced three times and BigQuery inlines it each time, and two of its inputs are deep views (`int_amplify__all_assessments` is 65 stages on its own). That makes a 154-stage plan with 424 parallel inputs, and the two candidate joins then broadcast the 706k-row course inventory to every one of those inputs. The new table is built once per tick on the same `0 0,10,13,15,17 * * *` schedule as the rest of the chain.

The resolver's output columns, types, and grain are unchanged. `fct_assessment_scores_enrollment_scoped` is untouched.

Refs #5261

## Reviewer Notes

- The new model ends in `select distinct` over its nine columns. Every column is part of the grain, so only byte-identical rows collapse (re-pulled iReady and STAR sittings). The value-level proof below shows it changes nothing downstream.
- The resolver's five unit tests now mock `int_assessments__score_anchors` with `format: sql` fixtures instead of the eight upstream refs. Expected rows are unchanged.
- The internal anchoring rule moved with the union, so the new model carries its own unit test for it, plus one for the duplicate collapse.

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Address valid feedback; dismiss false positives with a brief reply explaining why. (Claude is advisory — use your judgement, but don't ignore it.)

### dbt

- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application (none; the new model feeds the resolver only)
- [x] **Breaking change?** No. No column renamed or removed on any contracted model.
- [x] Not applicable: no external source added or modified.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Spec: `docs/superpowers/specs/2026-09-14-resolver-score-anchors-split-design.md`. Plan: `docs/superpowers/plans/2026-09-14-resolver-score-anchors-split.md`.

Value-level proof, 2026-09-14: the compiled resolver from `main` and the new chain (new model SQL as a CTE feeding the new resolver SQL) ran against the same prod snapshot in one query, full outer join on the seven-column grain, `to_json_string` over every output column. Result: `only_in_old 0, only_in_new 0, differing 0, total_keys 2,655,272`.

Local dev build of both models: 15 of 15 pass (7 unit tests, 6 generic tests, the `int_assessments__resolved_section_enrollments__unique_per_score` singular test). Per-source row counts of the new table match the spec's measurement exactly (internal 2,262,873; state_nj 71,996; state_fl 20,285; iready 264,787; star 7,965; dibels 62,485).

Baseline: resolver 15.15 slot hours / 7 days, 154 stages, 424 parallel inputs on the tier-1 join. Done when the resolver plus the new model combined are under 10 slot hours 7 days after merge (plan Task 5). If they are not, the follow-up is materializing `int_amplify__all_assessments` and `int_assessments__assessments_canonical`, filed under #5212.

</details>


cbini:

1. **PII tag: fixed in c5db52d5a5.** `powerschool_student_number` now carries `config.meta.contains_pii: true` at the column, matching the one sibling in this directory that tags (`int_assessments__fast_previous_year`). The other columns describe which test was sat, not a score, and the resolver that already exposes the same columns is untagged; a wider tier-3 tag on this family is a separate decision.

2. **Uniqueness test cost: measured, no change.** From the dev build's BigQuery job stats: `int_assessments__score_anchors` CTAS 3.6 slot-min (101 stages), the resolver CTAS 2.0 slot-min (11 stages, down from 154), the new uniqueness test 0.05 slot-min. Combined about 5.6 slot-min per build against the 25.3 baseline. At 5 builds a day that projects to roughly 3.3 slot hours a week for the pair, under the 10-hour bar with room. Task 5 of the plan confirms it in prod 7 days after merge.

3. **Literal-after-cast ordering: no action.** Copied unchanged from the resolver; sqlfluff ST06 ran in Trunk on this PR and passed, so the rule as configured accepts it.

PR #5314, opened 2026-09-14
fix(dibels): aimline model cleanup -- labelled twins, missing-data split, and the sheet read

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will make the DIBELS aimline method reportable: the two testing-state fields Tableau had nothing to bind to, the four reporting buckets academics asked for, a corrected category, a regression test for the benchmark-goal counts, and the reference document that explains all of it.

The aimline branch of the dashboard extract was half-wired. Its verdict columns were hardcoded to `null`, so half the PM rows carried no goal status at all, and its category put the words "Not Tested" on 1,842 rows that display a score. Academics then asked five questions about aimline reporting, and answering them turned out to need no new modelling — only the labels and one convenience column.

## Reviewer Notes

Five things in one PR, in the order they matter:

**1. The aimline model's labelled twins.** `int_amplify__pm_met_criteria_aimline` gains `admin_benchmark_goal_status` and `measure_standard_goal_status`, matching the internal sibling's vocabulary so one BI field reads across both methods. `measure_standard_goal_status` is three-valued here — Met, Not Met, No Aimline Data — because this method has a state the internal one does not. The extract now carries them through instead of `null`, and `met_measure_standard_goal` carries `met_aimline_goal` on the Aimline branch.

**2. `aimline_category` rewritten, six values.** `Round Incomplete` replaces `Not Tested` for the incomplete case, which is the actual fix: the old label read "Not Tested" on rows with scores. The two `No Aimline Data` values split what used to be one undifferentiated bucket plus 1,533 rows that read `Meeting Aimline, On-Track` with no aimline verdict behind the claim.

T&L's benchmark-wins rule is preserved deliberately: `Meeting Aimline, On-Track` still fires on the benchmark ahead of the aimline check, so 696 AY2025 rows in it are below the aimline. Their written rule, recorded in the yml so nobody "corrects" it. The `accepted_values` test still listed the five pre-rewrite values and is updated.

**3. Two testing-state fields.** `round_test_status` on the participation roster (Not Tested / Round Incomplete / Fully Tested) and `measure_test_status` on the extract (Tested / Not Tested, row grain). Academics are specific that an untested required measure is a miss and a partly-tested round is incomplete, and those are two different grains, so they take two columns. The measure-grain one can only live on the extract: both PM criteria models carry scored rows only, so an untested measure has no row in them. Verified to agree with zero leakage — Fully Tested rounds are 33,640 tested / 0 untested, Not Tested rounds 0 / 7,927, and only Round Incomplete mixes.

**4. `aimline_trajectory_category` on the extract.** Academics' four reporting buckets plus the untested row, for the aimline views. It crosses `met_aimline_goal` with `met_admin_benchmark_goal` directly rather than reading `aimline_category`, because the category applies the benchmark-wins rule and would report 713 below-aimline rows as meeting one. It also ignores round completeness, where the category puts Round Incomplete first — academics' legend has no such slice. The two columns therefore disagree on 2,554 AY2025 rows by design, and the yml says so and says not to union them.

**5. A regression test for the benchmark-goal counts.** `rpt_gsheets__dibels_bm_goals_calculations__region_counts_school_invariant` asserts one distinct region `bl_wb` value per academic year / region / grade / period, and zero nulls, for All / IEP / MLL. It fails on 18 of 27 groups against prod before #5315 and on 0 after — verified in both directions, which is what makes it a real test rather than a passing one.

Plus the docs: the reference document gains the six-value category table, the testing-state section, the per-measure-per-round reporting rule, the three grains of "met aimline" with their AY2025 rates, the `bl_wb` bug write-up, and a section on a whole region going missing upstream of dbt. The skill gains the sheet-reading technique, the additive-per-region paste rule, and a trigger on any DIBELS question rather than only code changes.

Row counts, AY2025 aimline, all measured against dev:

| Column | Distribution |
| --- | --- |
| `aimline_category` | Below Aimline 16,813 / Meeting Off-Track 7,071 / Meeting On-Track 6,844 / Round Incomplete 2,554 / No Aimline Data Off-Track 1,688 / On-Track 1,534 |
| `aimline_trajectory_category` | Below Aimline 18,284 / Not Tested 9,383 / `On Aimline, Below Benchmark` 7,431 / On Track to Benchmark 6,456 / No Aimline Data 3,313 |

Both sum to the branch's expected rows, so a percent-of-total stack needs no filtering.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — no new models; every new column is documented in the existing properties files
- [x] Include (or update) an exposure for all models consumed by a dashboard — the `literacy_dashboard` exposure already covers the extract; no new consumer
- [x] **Breaking change?** No. Every change is additive except `aimline_category`'s value set, which no view binds to yet
- [x] If adding or modifying an external source, run `stage_external_sources` with `--target staging` — no external source change

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Follow-up to #5309 and #5315.

**Reading the academics goals sheet.** Both obvious paths fail and neither says so. `mcp__bigquery__execute_sql` against a Sheets-backed source returns `403 Access Denied: Permission denied while getting Drive credentials` — the MCP service account has no Drive scope, so this is scope and not ACL, and sharing the file cannot fix it. The Drive MCP reads the file and then `check-output.sh` blanks the entire response, because a spreadsheet nearly always contains something that looks like a random string; `download_file_content` is worse, since base64 is high-entropy by construction. `scripts/read_sheet_tabs.py` requests `spreadsheets.readonly` + `drive.readonly` via `google.auth.default(scopes=...)`, writes each tab to a TSV, and prints only tab name and dimensions. Two footguns documented inline: echoing cell values re-redacts the run, and an output path containing the session UUID does too.

**Verification.** `measure_test_status` and `round_test_status` cross-tabulate with zero leakage. Percent tested belongs at `expected_measure_standard`, not `expected_measure_name_code` — paired standards are never split once both are expected (0 of 7,399 ORF, 0 of 7,524 NWF), but Reading Accuracy is a BOY→MOY measure with zero expected rows in rounds 5 to 8, so the name code blends two populations in one season and means one of them in the other. The three teacher-level asks resolve on 99.4% of AY2025 rows and 99.9% of AY2026 across 180+ teachers, with `scheduled = 1` exactly where the teacher resolves.

**The regression test was verified in both directions**, which is the part worth checking: 18 of 27 failing groups against prod pre-fix, 0 post-fix, and the residual gap proven closed on AY2024's four zero-Well-Below groups.

</details>


GabyRangelB:
## Review verdicts

Thanks — three of the five were real and are fixed in `0b1161277`. All doc-only, as you said; nothing reached the warehouse. Verdicts per finding, with the numbers re-derived against the branch build rather than taken from the docs.

**1. Category counts vs the stated total — VALID, fixed.** You were right that they cannot both be true, and right about the cause. Re-derived: the six buckets are 16,813 / 7,071 / 6,844 / 2,554 / 1,688 / 1,534, summing to **36,504**, which is exactly what the model produces. The stale part was the `36,486` total *and* two bucket counts transcribed a row low. `all_assessments` holds 36,514 aimline rows for AY2025 over **36,507 distinct keys**, so the 36,507 you saw was a distinct-key count and was never wrong — only the total beside it was.

The root cause is worth recording: some figures were measured against prod and some against the branch build, and the two differ by a handful of rows precisely because this PR changes the category. Every count in these sections now names its basis.

**2. PR body vs shipped docs — VALID, and the PR body was the correct side.** 7,071 / 1,534 / 7,431 / 3,313 are the branch-build numbers; the ymls and reference doc carried the prod-measured set. Fixed by moving the docs to the branch numbers, not the other way round.

**3. `On Aimline, Below Benchmark` comma — VALID, fixed, and it was hiding a second problem.** Adding the comma to the description turned a comma-separated list of five values into an ambiguous one, since a value now contains a comma. The description quotes each value individually instead and says to quote it when filtering.

**4. `admin_benchmark_goal_status` and a NULL condition — CORRECT REASONING, ASSUMPTION HOLDS, no change.** Good catch to flag it as unverifiable from your side; the query you suggested is the right one. Run against the branch build:

```text
countif(met_admin_benchmark_goal is null) = 0   -- AY2025
countif(met_admin_benchmark_goal is null) = 0   -- all years
```

Zero in both, so `IF()` never sees a NULL condition here and the two-valued status is safe. The yml's assertion that it "does not occur in the rows this model keeps" is accurate. Worth noting your read of the mechanism was exactly right — it is the same shape as the bug this PR fixes in `aimline_category`; it just does not fire, because `met_admin_benchmark_goal` is non-null wherever `benchmark_goal` resolves, and the model keeps only those rows.

**5. ST06 ordering — AGREED, no action.** Your own dismissal is correct, and thank you for surfacing it rather than filing it. `.claude/rules/dbt-sql.md` does put `case` after `if()`, and the existing files you cite already mix them this way.

One correction to the framing, not to a finding: while re-deriving for #1 I found the population-loss note also claimed "21 rows, five Newark students" with a per-cause breakdown. The anti-join now returns 3 rows and 2 students, so the breakdown went with it rather than being carried forward unverified — that one you could not have caught without warehouse access, and it was adjacent to the arithmetic you did catch.

GabyRangelB:
@claude the five findings from your review are answered in the comment above, and the three valid ones are fixed in `0b1161277`. Please check the fixes rather than re-reviewing the whole PR.

Specifically worth your attention:

1. **Finding 1** — I re-derived the counts against the branch build rather than reconciling to either number in the docs. The six `aimline_category` buckets now sum to 36,504, which is what the model produces; `all_assessments` holds 36,514 aimline rows over 36,507 distinct keys. Please confirm the arithmetic now closes, and that I have not introduced a new mismatch anywhere the old numbers were also cited — the skill and the reference doc both carried copies.

2. **Finding 3** — adding the comma to `On Aimline, Below Benchmark` made the description's comma-separated list of five values ambiguous, since one value now contains a comma. I quoted each value instead. Check the list reads unambiguously, and that the SQL literal and the description now agree character for character.

3. **Finding 4** — you flagged this as needing local verification and you were right to. `countif(met_admin_benchmark_goal is null)` returns 0 on the branch build, for AY2025 and for all years, so `IF()` never sees a NULL condition and the two-valued status holds. If you think zero-today is too weak a guarantee for a column whose own yml asserts the condition never occurs, say so and I will add the test rather than rely on the assertion.

One thing you could not have caught without warehouse access, now fixed: the population-loss note claimed 21 rows and five Newark students with a per-cause breakdown. The anti-join returns 3 rows and 2 students, so I dropped the breakdown rather than carry it forward unverified.

GabyRangelB:
Following up on finding 4, since I acted on it differently than suggested rather than silently.

**Declined the exact suggestion, implemented the stronger version.** A `not_null` on `met_admin_benchmark_goal` would never fire. The column is

```sql
case when measure_standard_score >= benchmark_goal then 1 else 0 end
```

so a null `benchmark_goal` makes the comparison null and falls through to `else 0` — the column is 0 or 1 by construction and never null. `.claude/rules/dbt-yaml.md` forbids a `not_null` on exactly that shape: it can never fail and still costs a full scan per CI run.

**The exposure you identified is real, one step upstream, and your number was exact.** `benchmark_goal` can be null — the by-levels gate carries **30 null standards across its 938 kept AY2025 rounds**, which matches your figure. And because the flag falls through to 0, such a row would report "not at grade level" rather than "no standard published". That is the same silent mislabelling `aimline_category` was rewritten to fix, so your instinct to hold this to the same bar was right; it just applies to the input, not the output.

So `a398baf32` adds `not_null` with `severity: error` to **`benchmark_goal`** on `int_amplify__pm_met_criteria_aimline`, and moves the column up to sit with the other tested ones per the properties convention. Measured 0 of 36,504 rows on AY2025, which is what makes it meaningful rather than vacuous — it passes today and fails the day a null-standard round becomes score-eligible. Your framing of the value was right and I kept it: this converts "we checked once" into something CI will tell us about.

**One change of state since your run, relevant to anyone reading the numbers above.** #5322 has merged, and `2ee0113a3` merges `main` into this branch. It touches this exact lineage — including `90a20ce61 fix(amplify): widen the aimline staging key to device_date and assessment_grade`. Prod has not rebuilt with it yet (its aimline model still carries the `surrogate_key` that #5322 drops), so every count in this thread was measured against pre-#5322 upstreams. A widened staging key can change how many rows reach this model, so the documented counts get re-derived once prod rebuilds rather than assumed to have held. Flagging it here so a reviewer does not treat the current figures as final.

GabyRangelB:
@claude two things for you since your verification run, both in the comment above and in commits you have not seen.

**Finding 4 — I took your advice but moved the test one column upstream.** `not_null` on `met_admin_benchmark_goal` would never fire: that column is `case when measure_standard_score >= benchmark_goal then 1 else 0 end`, so a null goal makes the comparison null and falls to `else 0`. It is 0 or 1 by construction, and `.claude/rules/dbt-yaml.md` forbids a `not_null` on that shape.

Your underlying concern was right and your number was exact — the by-levels gate does carry 30 null standards across its 938 kept AY2025 rounds. Because the flag falls through to 0, such a row reports "not at grade level" rather than "no standard published", which is the mislabelling this PR exists to fix. So `a398baf32` puts `not_null` with `severity: error` on **`benchmark_goal`** instead, measured 0 of 36,504 rows today. Please sanity-check that placement: is `int_amplify__pm_met_criteria_aimline` the right layer for it, or does it belong on the by-levels gate with a `config.where` scoping it to the rows that can reach a scored row?

**The lineage moved under both of us.** #5322 merged and `2ee0113a3` merges `main` into this branch — it touches this exact chain, including `90a20ce61 fix(amplify): widen the aimline staging key to device_date and assessment_grade`, and it drops `surrogate_key` from the mclass staging and intermediates. Prod has not rebuilt with it yet, so every count you verified was measured against pre-#5322 upstreams. Two questions worth your read of the merged diff: does anything in #5322 change which rows reach `int_amplify__pm_met_criteria_aimline`, and does its `surrogate_key` removal touch anything this branch relies on? I am re-deriving the documented counts once prod rebuilds rather than assuming they held, but if you can see from the diff that the row set cannot change, that is worth knowing now.

GabyRangelB:
Closing the loop I left open above: prod has now rebuilt with #5322, and **every documented count held unchanged**. Re-derived against the post-#5322 lineage rather than assumed.

| Figure | Pre-#5322 | Post-#5322 |
| --- | --- | --- |
| `aimline_category` six buckets | 16,813 / 7,071 / 6,844 / 2,554 / 1,688 / 1,534 | identical |
| Model total, AY2025 | 36,504 | 36,504 |
| `all_assessments` aimline rows / distinct keys | 36,514 / 36,507 | 36,514 / 36,507 |
| Population loss | 3 rows, 2 Newark students | 3 rows, 2 Newark students |
| `aimline_trajectory_category` | 18,284 / 9,383 / 7,431 / 6,456 / 3,313 | identical |

So the widened aimline staging key changed which columns the staging model carries, but not which rows reach `int_amplify__pm_met_criteria_aimline` — the join keys this model uses were already at or below that grain. The figures in the ymls and the reference doc stand as written; no further doc edits.

The new `not_null` on `benchmark_goal` also passes against the rebuilt lineage, along with the model's other seven nodes, so the 0-of-36,504 measurement behind it is not an artifact of the pre-merge upstreams.

GabyRangelB:
@claude your two answers hold up, and I verified them a different way. Also chased the Maze finding you flagged but could not query.

**Test placement — agreed, keeping it.** Your reasoning is the part that settles it, not the conclusion: `config.where` only sees the gate's own columns, so "rows that can reach a scored row" would mean re-deriving this model's inner joins inside a generic column test. And the gate's null is documented as legitimate and permanent with three consumers sharing it, so a test there would fire on the 30 known-good rows. No change.

**#5322 row admission — your SQL trace and my row counts agree.** Prod rebuilt after your run, so I re-derived rather than assuming. Every figure is byte-identical: six `aimline_category` buckets 16,813 / 7,071 / 6,844 / 2,554 / 1,688 / 1,534, model total 36,504, source 36,514 rows over 36,507 distinct keys, population loss 3 rows / 2 Newark students, and the five `aimline_trajectory_category` buckets 18,284 / 9,383 / 7,431 / 6,456 / 3,313. Two independent methods, same answer — your read that the joins were already at or below the widened grain was right.

**Maze — partially checkable, and here is what the data says.** Post-rebuild prod, Benchmark rows only:

| academic_year | Maze rows | scored |
| --- | --- | --- |
| 2026 | 5,877 | 5,877 (100%) |
| 2025 | 19,880 | 19,880 (100%) |
| 2024 | 18,492 | 18,492 (100%) |

The 100% is not the interesting number — `int_amplify__all_assessments` keeps scored rows only, so that percentage is true by construction and proves nothing. **The row count is the signal**: 5,877 live-year Maze benchmark rows exist, so Maze is flowing under the SFTP column names now. What I cannot do is see the pre-rebuild state, since prod has already overwritten it — so your claim that live-year Maze read NULL before the merge is consistent with what I see but not retroactively provable from here.

Two things follow that I would rather state than leave implied. It does not touch this PR: the Maze work in this branch is PM/aimline-scoped, where Maze comes from the PM file rather than the benchmark unpivot, and the aimline population's Maze behaviour — 30 eligible rows for the consecutive-miss flag, because each region schedules Maze once per season — reproduces identically post-rebuild. But if you are right, then Maze-specific **Benchmark** reporting for the live year was understated before today, which is a Literacy Dashboard question rather than a PR question, and worth someone checking against Amplify's own reporting rather than against our warehouse, since our pre-fix state is gone.

GabyRangelB:
@claude the Maze finding does not hold — I was able to check the pre-merge state after all, and it refutes it. Worth correcting on the record since we both flagged it as a live-year data-quality problem.

**BigQuery time travel reads up to 7 days back**, so "prod overwrote it" was wrong of me. `int_amplify__mclass__benchmark_student_summary_unpivot` rebuilt at **21:03:51Z**, so a snapshot at 20:30Z is genuinely pre-rebuild. Maze in the unpivot, both sides:

| Snapshot | Maze rows | scored |
| --- | --- | --- |
| `for system_time as of 20:30:00+00` (pre-#5322) | 43,693 | 43,693 |
| now (post-#5322) | 43,693 | 43,693 |

Byte-identical. Live-year Maze benchmark scores were never reading NULL.

**Why the inference missed**: the column rename is real, but #5322 removed a `coalesce` in the same commit that was already covering both spellings. Pre-merge, `int_amplify__mclass__benchmark_student_summary.sql` carried

```sql
basic_comprehension_maze_local_percentile as reading_comprehension_maze_local_percentile,
coalesce(basic_comprehension_maze_score, reading_comprehension_maze_score) as reading_comprehension_maze_score,
```

and the same pair for semester growth, year growth, national-norm percentile, level, tested-out and discontinued — eight pairs in all, each resolving the SFTP name first and falling back to the API name. So both spellings already landed on the same output column. #5322 deletes those pairs precisely because the new archive wrappers alias the API names to the SFTP names upstream, which makes the fallback redundant rather than newly necessary. Same values, different plumbing.

The general shape is worth keeping: a rename that looks like it must have been silently dropping data can be load-bearing-free if a `coalesce` upstream was absorbing it, and the commit that renames is often the same commit that removes the absorber. Reading the rename without the deletion inverts the conclusion.

No change to this PR either way, and no Literacy Dashboard follow-up needed — I had told the team there might be one, and that is now withdrawn.

PR #5315, opened 2026-09-15
fix(dibels): compute the benchmark goals bl_wb counts on the row that reports them

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will fix the blank "Below / Well Below" counts on the benchmark goals report — **35 of 74 rows were blank, and now none are.** Academics are looking at those blanks now, which is why this is split out on its own rather than riding with the larger aimline work in #5314.

**One SQL file, and it removes more code than it adds** — 18 lines in, 51 out. The only review path is `src/dbt/`.

Six columns were affected: the Below/Well Below headcounts at school and region level, for all students and for the IEP and MLL breakouts.

Each of those six counts was being calculated in a way that made it come out as zero on the very row that has to report it, so the model was fetching the real value from a second, related row instead. That fetch is what broke. It could land on a student whose row cannot supply the number, and when it did, all six columns came out blank for that school — including the region-wide figures, which are the same for every school in a region by definition.

The fix calculates the counts on the row that reports them. The second lookup is then unnecessary and has been deleted, which is why the file gets shorter.

Three things this explains, all of which were observed:

- **Newark looking correct was luck.** The chance of a good outcome depended on the mix of students at each school. Newark's mix won every time; Camden and Paterson did not.
- **The answer changed between runs.** The numbers pasted into the goals sheet are blank for Camden K, 1 and 4, while the report today is blank for 1, 6 and 8 — same code, different run.
- **The region figure disagreed with itself.** Camden grade 1 read 82 for one school and blank for the other.

## Reviewer Notes

Checked against the source data rather than against itself: all 74 rows for 2026-2027 match a Below/Well Below headcount computed directly from the assessment and enrollment tables — no blanks, no mismatches, and at-or-above plus below/well-below adding to the total on every row. Row count unchanged.

A second, rarer version of the same problem is also closed, and demonstrably so. A school with below-benchmark students but none in the lowest band had no second row to fetch from at all. The 2023-2024 data contains four such cases; the old approach returns blank for all four, the new one returns the correct counts.

**Once this merges, the goals sheet needs re-pasting** — the blanks are baked into what is there now.

The reference-document write-up rides on #5314, because a `docs/` path pulls in a second reviewing team and this needs to clear review today.

### General

- [x] Review the **Claude Code Review** comment posted on this PR — findings 1, 2 and 3 are addressed in this revision; verdicts posted per finding.

### dbt

- [x] No breaking change. No columns added, removed or renamed.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.
- [ ] **Dagster Cloud** — not triggered.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. This revision supersedes the first one, which swapped the `rn` partition to the three-way `foundation_measure_standard_level` and kept the self-join. That fixed the random draw but left the structural gap; `claude-review` caught it (findings 1 and 2) and was right. Rewritten to the root-cause form and squashed to one commit.

Mechanism. The six `n_admin_season_*_bl_wb[_iep|_mll]` windows each partitioned by `a.aggregated_measure_standard_level` — the same column their own `if()` filters on — so the count evaluated to zero on an `At/Above` row. The final `SELECT` anchors on `At/Above` (`where c.grade_goal_type = 'At/Above'`), hence the `left join needed_count_calcs as b ... and b.grade_goal_type = 'Well Below'` to fetch all six from the sibling. The `at_above` columns never needed it: their `if()` is true on the partition their anchor row lives in.

Two failure modes, both from that join. `grade_goal_type` derives from the three-valued `foundation_measure_standard_level` while the partition was two-valued, so the `Below/Well Below` partition mixed `Well Below` students (goal type present) with `Below` students (goal type NULL), and `rn` had no `ORDER BY` — a non-deterministic pick between a row that can serve `b` and one that cannot. Separately, a school with `Below` students but zero `Well Below` had no qualifying sibling at all.

Fix drops `a.aggregated_measure_standard_level` from those six `partition by` clauses only. The combined count then lands on the anchor row and the six columns read `c.` directly; the self-join is deleted. `rn` keeps `aggregated_measure_standard_level` — it must, or the surviving row may not be the `At/Above` one the `WHERE` needs — and gains `order by a.student_number` for determinism.

Verification, AY2026, against independently computed source truth (`int_amplify__all_assessments` joined to `int_extracts__student_enrollments`, filters `assessment_type = 'Benchmark'`, `measure_standard = 'Composite'`, `period != 'EOY'`): 74 rows compared, 0 null school `bl_wb`, 0 null region `bl_wb`, 0 school-level mismatches against source, 0 rows where `at_above + bl_wb != all`. Grain 74, unchanged. Camden region values by grade: 118, 82, 100, 92, 94, 131, 106, 124, 160.

Finding 1 closed with evidence rather than argument. Grouping the roster join by `(region, school, assessment_grade, period)` across every year the model has emitted gives four groups with zero `Well Below` students, all in AY2024: grade 8 MOY with one `Below` student, and three grade 4 groups with none. Old self-join returns NULL for all four. New count returns 1 and 0 respectively — both correct, and NULL never was.

Finding 3 (no regression test) is still open and belongs in a follow-up, not here. The invariant worth asserting is `count(distinct n_admin_season_region_gl_bl_wb) = 1` per `(academic_year, region, assessment_grade, period)` — a region figure differing across schools in one region is the signature of both failure modes. Finding 4 (uniqueness test on the `rpt_` layer) declined as a directory-wide convention: none of the 33 `rpt_gsheets__*` properties files declare `data_tests`.

Adjacent, untouched: `grade_goal_iep` and `grade_goal_mll` are NULL on all 74 AY2026 rows because this year's foundation goals carry `population = 'All'` only. AY2025 had all three. Deliberate pause on those targets, not a regression. The IEP and MLL headcount breakouts are populated and were fixed alongside the All pair.

</details>


GabyRangelB:
@claude Good review. Verdicts per finding, and finding 1 is confirmed rather than plausible — I could run the check you specified.

**Finding 1 — residual null gap when a school has zero Well Below students: CONFIRMED, and it has occurred. Accepting as a follow-up, not fixing here.**

Your reasoning is right and the mechanism is live. Running your exact check across every year in the model's history, grouping the roster join by `(region, school, assessment_grade, period)` under the same three filters:

| Academic year | School-grade-periods | `Below > 0` and `Well Below = 0` | Any with no Well Below |
| ------------- | -------------------: | -------------------------------: | ---------------------: |
| 2023          |                   48 |                                0 |                      0 |
| 2024          |                  147 |                            **1** |                  **4** |
| 2025          |                  160 |                                0 |                      0 |
| 2026          |                   74 |                                0 |                      0 |

So it is not structurally ruled out and it is not hypothetical — AY2024 has one group that would hit it exactly as you describe, plus three more with no Well Below students at all. **AY2026, the year this model emits, is clean at zero**, which is why this ships as-is: the reported bug is fixed for the data in scope, and the residual gap cannot fire on the current year.

Worth noting it is strictly narrower than what was fixed. Before, a bad draw could null a school that HAD Well Below students — which is most of them. Now only a school with none can, which is rare enough to have happened once in 429 group-years.

**Finding 2 — drop `aggregated_measure_standard_level` from the six `bl_wb` windows and delete the self-join: agreed, and it is the better fix. Follow-up.**

You are right about the root cause. Those six windows partition by the column their own `if()` filters on, which is what forces the value to zero on the `At/Above` anchor row and creates the need for a sibling at all — and the `at_above` columns do not have the problem for exactly the reason you give. Widening the partition puts the combined Below+Well Below count directly on the anchor row, the self-join disappears, and finding 1 disappears with it because nothing then depends on a Well Below row existing.

Not doing it in this PR, deliberately. This is a same-day fix for blank cells academics are looking at right now, it is green, and it is one file. Swapping to the wider design touches six window definitions plus six column references plus the join block, and re-opens review. Taking it as a follow-up with finding 3.

**Finding 3 — no test guarding the regression: agreed, and it belongs with finding 2.**

A test here would need to assert the invariant rather than the value — `count(distinct n_admin_season_region_gl_bl_wb) = 1` per `(academic_year, region, assessment_grade, period)`, since a region figure differing across schools in the same region is the signature of both this bug and finding 1. That is a real test worth having and it would have caught this. It goes in the follow-up, where the design it guards is settled.

**Finding 4 — missing uniqueness test on the `rpt_` layer: declining, pre-existing convention.**

Your own grep is the argument: none of the 33 `rpt_gsheets__*` properties files in this directory declare `data_tests`. These are Sheets-paste extracts whose grain is enforced by what a human pastes, not by the model, so the layer rule reads as not applied here by convention rather than overlooked on this one model. Not introduced by this change and not something to fix inside a hotfix.

**Finding 5 — inline comment versus properties description: keeping it inline.**

You identified the carve-out correctly. The comment is a guardrail against reverting to the two-way partition, sitting at the derivation site — that is the "a defect belongs in the code, not the metadata" case, not background prose.

**Findings 6, 7 — accepted.** No Dagster, no surrogate key, nothing to cut.

**Finding 8 — moot now, and worth flagging so it does not confuse a later reader.** The docs section you reviewed is no longer in this PR. I amended the commit after your run to drop `docs/models/dibels-dashboard-data-model.md`, because a `docs/` path adds a second CODEOWNERS team and this needs to clear review today. The write-up rides on #5314 instead. Your line references in findings 1 and 8 point at content that has moved, and the "region value identical across every school" sentence you flagged is on that PR — I will qualify it there with the finding-1 caveat, since as written it overstates what the fix guarantees.

PR #5318, opened 2026-09-15
perf(surveys): build survey submissions from response grain

### LINKED ISSUE #5276: fix(surveys): 68 submissions carry two academic years, and a tiebreaker silently picks one
## What's happening

`int_surveys__survey_responses` is question-grain. For a given submission, every question row should agree on that submission's academic year. For 68 submissions they do not — the same `(survey_id, survey_response_id)` carries more than one `academic_year`.

Nothing errors. `int_surveys__survey_submissions` collapses each submission to one row, so whichever question sorts first decides the academic year for that submission. That year then drives the reporting-terms join in the marts, which decides which survey administration the submission attaches to.

So 68 submissions are attached to an administration chosen by a sort order rather than by a rule.

This is not a regression. It predates the collapse and was found while extracting the submission-grain intermediate in #5270.

## Steps to reproduce

1. Run the query below against `kipptaf_surveys`.
2. `varying_academic_year` comes back 68, out of 126,808 submissions.
3. Every other projected column comes back 0, so academic year is the only one affected.

```sql
with per_partition as (
  select
    count(distinct survey_title) as n_title,
    count(distinct respondent_email) as n_email,
    count(distinct respondent_employee_number) as n_empnum,
    count(distinct date_submitted) as n_date,
    count(distinct academic_year) as n_ay,
    count(distinct term_code) as n_term,
  from `teamster-332318.kipptaf_surveys.int_surveys__survey_responses`
  group by survey_id, survey_response_id
)
select
  count(*) as n_partitions,
  countif(n_title > 1) as varying_title,
  countif(n_email > 1) as varying_email,
  countif(n_empnum > 1) as varying_employee_number,
  countif(n_date > 1) as varying_date_submitted,
  countif(n_ay > 1) as varying_academic_year,
  countif(n_term > 1) as varying_term_code,
from per_partition
```

## Where

- **Environment:** prod
- **Run, PR, or dashboard link (if any):** found during #5270

<details>
<summary>For Claude</summary>

> ### This is an investigation issue, not a ready fix
>
> The cause is NOT established. Two candidate mechanisms, neither tested:
>
> 1. **Overlapping reporting-term windows.** `int_surveys__survey_responses` resolves the year with `left join stg_google_sheets__reporting__terms as rt on fr.info_title = rt.name and date(fr.last_submitted_time) between rt.start_date and rt.end_date and rt.type = 'SURVEY'`. If 2 SURVEY rows share a `name` and their date windows overlap across academic years, one submission matches both and fans out.
> 2. **The Alchemer arm disagreeing with itself.** That arm computes `coalesce(sr.campaign_fiscal_year - 1, rt.academic_year)`, so a campaign-derived year and a terms-derived year can differ for the same submission.
>
> Start by splitting the 68 by arm (Google Forms `form_id` vs stringified Alchemer `survey_id`) and by survey title. That alone should decide between the two.
>
> ### Why it matters downstream
>
> `int_surveys__survey_submissions` picks one row per `(survey_id, survey_response_id)` ordered by `survey_question_id` ascending. The model documents this at the `live_ranked` CTE. `academic_year` flows into the terms join in `fct_survey_submissions` and `dim_survey_administrations`, and `survey_administration_key` hashes term columns — so the pick can change which administration a submission belongs to.
>
> The current order key is deliberately stable, so today's behavior is reproducible. Do not change it as a side effect of something else; that would silently move those 68 submissions between administrations.
>
> ### What fixed looks like
>
> The reproduce query returns 0 for `varying_academic_year`, resolved at the source of the ambiguity rather than by a tiebreaker. Confirm `survey_administration_key` on `fct_survey_submissions` either does not move, or moves in a way you can explain submission by submission.
>
> Related: #3918, #5217.

</details>

### LINKED ISSUE #5277: perf(surveys): build survey submissions from response grain instead of collapsing 4.1M question rows
## What and why

`int_surveys__survey_submissions` produces 126,808 submissions by collapsing `int_surveys__survey_responses`, which holds 4,132,485 question-grain rows. The submissions already exist at response grain further upstream, so the collapse is undoing an expansion rather than deriving anything.

| Model | Rows | Grain |
| --- | ---: | --- |
| `stg_google_forms__responses` | 80,804 | 1 row per response |
| `stg_alchemer__survey_response` | 112,318 | 1 row per response |
| `int_google_forms__form_responses` | 1,631,066 | form item x response |
| `stg_alchemer__survey_response__survey_data` | 2,555,620 | answer |
| `int_surveys__survey_responses` | 4,132,485 | question |

The expansion is structural. `int_google_forms__form_responses` joins form **items** to responses on `form_id` alone, so every item pairs with every response of that form and 80,804 responses become 1.6M rows.

Reading the 2 response-grain models instead would take the input from 4.1M rows to about 193,000.

## Proposed approach

Source `int_surveys__survey_submissions` from `stg_google_forms__responses` and `stg_alchemer__survey_response`, so the model has no grain collapse at all.

This is a proposal, not a plan. It depends on extracting logic that currently lives inside `int_surveys__survey_responses`, which has 7 other consumers. See the fold-out for what is verified and what is not.

Worth about 12 slot hours a week, mostly in dbt Cloud CI. Real, but far smaller than the 735 a week #5217 already recovered, so this is not urgent.

<details>
<summary>For Claude</summary>

> ### This is an investigation issue, not a ready fix
>
> **Verified:**
>
> - The row counts above, from `__TABLES__` in `kipptaf_google_forms` and `kipptaf_alchemer`.
> - The cross product at `int_google_forms__form_responses.sql:118-119`.
> - All 8 columns `int_surveys__survey_submissions` projects resolve from response-grain inputs. In particular `respondent_employee_number` joins `int_people__staff_roster_history` on `respondent_email` and `last_submitted_time`, NOT on any answer value, so it does not force question grain. That was the main thing that could have killed the idea, and it does not.
> - The current cost: the model builds in about 5.87 slot minutes scanning 914.9 MB.
>
> **Not verified, and each of these can sink it:**
>
> - Whether a response-grain build reproduces the same 126,808 submissions and the same `survey_submission_key` set. This is the only check that matters and it has not been run.
> - The Alchemer arm's real grain. It reads the `base_alchemer__survey_results` and `int_surveys__response_identifiers` sources, neither of which was inspected.
> - Where `survey_title` comes from. It arrives today as `info_title` through the per-item join, so a response-grain build needs another path to it.
> - Whether the roster resolution (including the `gdir_alias_map` alias fallback and the `srh` / `srh_alias` coalesce) and the reporting-terms resolution can be extracted into a shared model without disturbing the 7 consumers of `int_surveys__survey_responses`: `int_topline__school_community_diagnostic`, `rpt_tableau__school_community_diagnostic`, `rpt_tableau__survey_responses`, `int_surveys__renewal_responses_feed`, `fct_survey_responses`, plus `int_surveys__survey_submissions` itself.
>
> Duplicating that resolution rather than sharing it is the failure mode to avoid — it is the identity logic the repo's "Reuse existing entity identity" rule exists for.
>
> ### Verification method
>
> #5270 established a workable one. Fingerprint the key sets before and after with `bit_xor(farm_fingerprint(key))` over the DISTINCT set — order-independent and overflow-free, unlike `sum()`. A behavior-preserving change returns identical row counts, distinct counts and fingerprints for `fct_survey_submissions` (submission and administration keys), `dim_survey_administrations`, and `fct_survey_responses` (response and submission keys).
>
> ### Watch for
>
> `int_surveys__survey_submissions` is `materialized: table` on `0 3 * * *`, the cadence its 3 consumers share. Keep that, or the collapse goes back to running once per consumer.
>
> Related: #3918 (the extraction that created this model), #5217, #5276 (68 submissions whose `academic_year` is ambiguous in the same upstream).

</details>

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." build `int_surveys__survey_submissions` from the response-grain staging tables instead of collapsing the 4.1M-row question-grain view, make `int_surveys__survey_responses` read its per-submission columns from that table, and union the decommissioned Alchemer feed in from build-once tables instead of recomputing it.

Closes #5277. Closes #5276.

Today `int_surveys__survey_submissions` reads `int_surveys__survey_responses` (4.1M rows) and collapses it to about 129,000 submissions. Those submissions already exist 2 models upstream at 1 row per response. Reading them directly cuts the model's input from 4.1M rows to about 193,000. The roster and reporting-term resolution moves with them, so it now lives once, in the submissions model, and the question-grain view joins submissions for those columns. Its 7 consumers see the same columns and need no edits.

The nightly cron on the submissions table is removed. Google Forms responses land every 15 minutes and the survey dashboard refreshes hourly through the question-grain view, so the table now rebuilds when its inputs change. That is affordable at the new input size.

`fct_survey_responses` used to join `int_surveys__survey_submissions` on top of a view that had already joined it. The view now exposes `survey_submission_key` and the fact reads it; only the manager-survey arm, which does not pass through the view, still joins submissions.

Alchemer is decommissioned, so every Alchemer join and calculation moves into 2 build-once tables: `int_surveys__alchemer_submissions` at 1 row per response, which also carries the historic Alchemer Manager archive, and `int_surveys__alchemer_responses` at question grain with every per-submission column already attached. The live submissions and responses models union them in as-is. Both tables were built into prod `kipptaf_surveys` on 2026-09-15 and now follow the archive pattern: the models are disabled, the tables are declared under the `surveys` source, and the live models read `source()`.

Calculations that were living in the survey models move to the furthest upstream model: `stg_google_forms__responses` derives the UTC submit date, the response link, and the lowercased email local part; `int_google_forms__form__items` and `int_google_forms__form_responses` carry the lowercased question abbreviation, and `form_responses` derives `answer`, `answer_value`, and `is_open_ended`; a new `int_google_directory__users__addresses` replaces the alias-map CTE that both submission models had copied; and the question-department crosswalk lowers `abbreviation` at staging.

This also closes #5276. The 68 submissions with 2 academic years came from 2 overlapping `KTAF Support Survey` windows in the reporting-terms sheet. The sheet row was corrected on 2026-09-14, and a `mutually_exclusive_ranges` test on `stg_google_sheets__reporting__terms` now fails, for every term type, when 2 windows of the same type, code, name, region, school, and grade band overlap, or when a window's end date precedes its start date.

## Reviewer Notes

- **The Alchemer tables are frozen.** `int_surveys__alchemer_submissions`, `int_surveys__alchemer_responses`, and the new `int_google_directory__users__addresses` view they read were built into prod from commit 1d36ed07e1 on 2026-09-15. Commit d37bbafec5 then disabled the 2 Alchemer models and their tests, declared both tables under the `surveys` source, and switched the 2 `ref()` calls to `source()`. To change the frozen rows, re-enable and rebuild.
- **Behavior is preserved, verified against prod after every round.** `fct_survey_submissions` (48,956 rows) and `fct_survey_responses` (580,663 rows) are identical to prod, full-row `except distinct` in both directions returning 0. `dim_survey_administrations` differs on 1 of 56 rows: a 2019 response deadline that the 2026-09-15 reporting-terms sheet fix moved from January to December, which prod's submissions table (built 07:06 UTC that day) predates. The #5276 reproduce query returns 0.
- **The terms test needed the staging copy of the sheet.** Against the prod table as built at 2026-09-15 14:49 the widened test failed on 5 rows, all since corrected in the sheet: `SURVEY` `Self and Others` 2019 `SO1` (ended 2020-01-31, overlapping `SO2` from 2020-01-01, which doubled 4,561 January 2020 Alchemer submissions); `MAP` `Fall` 2016 `MAP1` (ended 2017-11-30, overlapping the 2017 `MAP1` window); and 3 KIPP Miami rows named `04/27 - 05/08` (`MG`, `O3`, `WT`, academic year 2025) whose start and end were inverted. The 53 rows with a null start or end date (52 `AR`, 1 `ACT`) are excluded from the test and are a separate defect. CI went green once `Clone - Staging` refreshed the copy.
- **4 phantom submissions disappear.** The old path left-joined responses from the form-items side, so 4 forms with items and zero responses produced a submission row with a null response id. Those rows never reached the marts, whose survey titles are modeled explicitly. The new population rule also runs the other way: a form with responses but zero items, which cannot occur today, would now produce submissions where it produced none before.
- **Key uniqueness now rests on the sheet, and the tests that guard it are error severity.** Nothing collapses in the submissions model any more, so an overlapping `SURVEY` window fails the `unique` test on `survey_submission_key` instead of being silently picked through. Intermediate tests default to warn in this project, so the uniqueness tests on `int_surveys__survey_submissions`, `int_surveys__alchemer_submissions`, and `int_surveys__alchemer_responses` carry `severity: error` explicitly. A duplicate there doubles every question row downstream. The YAML says so.
- **The 26 CI warnings on the final run are pre-existing.** The forms staging edit fans `state:modified+` out to 289 models, so CI surfaced warn tests it does not normally run. The 3 on survey models were re-run against prod: `bridge_survey_questions` orphans 18 in both, KIPP Forward career-launch duplicates 365 in both, and the single `fct_survey_submissions` enrollment-key orphan is 0 in prod (CI's staging copy of `dim_student_enrollments` lags). The rest are on unrelated models and appear in other PRs' CI runs.
- **The crosswalk lowering is a value no-op today.** 0 of 309 prod rows have a mixed-case `abbreviation`; lowering at staging just moves the existing `lower()` out of the join.

### General

- [x] Review the **Claude Code Review** comment posted on this PR. Address valid feedback; dismiss false positives with a brief reply explaining why. (Claude is advisory — use your judgement, but don't ignore it.) Per-finding verdicts posted as a PR comment.

Skipped: no Dagster changes. The cron removal is a dbt properties change; the model falls back to the default eager table automation condition.

- [x] Include a `[model_name].yml` properties file for all new models — `int_surveys__alchemer_submissions.yml`, `int_surveys__alchemer_responses.yml`, and `int_google_directory__users__addresses.yml` added; every edited model's properties file carries the new columns with descriptions
- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — no exposure changes; consumers and their columns are unchanged
- [x] **Breaking change?** No column renamed or removed on any contracted model. `stg_google_forms__responses` gains 3 columns, `int_google_forms__form__items` 1, `int_google_forms__form_responses` 4, `int_surveys__survey_submissions` 8, `int_surveys__survey_responses` `survey_submission_key`; `fct_survey_responses` keeps its 5.
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — not applicable; the 2 new source declarations point at frozen native BigQuery tables in `kipptaf_alchemer`

Skipped.

## CI checks

- [x] **Trunk** — passes. If it fails, run `trunk check` and `trunk fmt` locally, fix any issues, and push again.
- [x] **dbt Cloud** — passes.

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Spec: `docs/superpowers/specs/2026-09-14-survey-submissions-response-grain-design.md`. Plan: `docs/superpowers/plans/2026-09-14-survey-submissions-response-grain.md`. The `fct_survey_responses` join change was added after the plan from a final-review recommendation. The Alchemer precompute and the upstream calc moves came from the author's review on 2026-09-15; the spec predates them.

**Verification numbers (prod versus `zz_cbini_kipptaf_*` dev build, 2026-09-15, after the review round).**

| Model | Prod rows | Dev rows | Prod-only | Dev-only |
| --- | ---: | ---: | ---: | ---: |
| `fct_survey_submissions` | 48,956 | 48,956 | 0 | 0 |
| `fct_survey_responses` | 580,663 | 580,663 | 0 | 0 |
| `dim_survey_administrations` | 56 | 56 | 1 | 1 |

Prod-only and dev-only are full-row `except distinct` counts. The `dim_survey_administrations` pair is one `survey_administration_key` whose `response_deadline_date` moved 2019-01-31 to 2019-12-31 with the sheet fix. Dev `int_surveys__alchemer_submissions`: 50,665 API rows and 2,559 archive rows, keys unique; `int_surveys__alchemer_responses`: 2,763,720 rows over the same 50,665 submissions. Dev build: 60 pass, 1 warn (the pre-existing `expression_is_true` on `int_surveys__survey_responses`, 480 rows), 0 errors. The crosswalk staging model was excluded from the dev build and deferred to prod because its Sheets external has no dev copy.

**Design decisions with reasons.**

- Google Forms terms join keys on `last_submitted_date`, the UTC date now derived in staging, because that is what `date()` of the raw string resolved to before. `last_submitted_date_local` differs on 3,552 of 80,817 rows. Switching to local is a separate behavior change and is not made here.
- Alchemer reads `stg_alchemer__survey_response`, `stg_alchemer__survey`, `stg_alchemer__survey_campaign` through `source()` because the Alchemer models sit under a project-level `+enabled: false`. The campaign join (`date_started between link_open_date and link_close_date`) is the one the disabled `int_alchemer__survey_results` used.
- The Alchemer terms join carries `rt.type = 'SURVEY'`, which the Google Forms arm already had, and is an inner join: a response outside every window is not a submission, as before.
- The Alchemer tables compute `respondent_identifier` and `survey_submission_key` themselves, with the same expressions as `int_surveys__survey_submissions`, so the responses table can carry the key without a live join. Two call sites, one composition; both YAMLs say so.
- `zero_length_range_allowed: true` on the terms test, since a single-day survey window is a legitimate sheet entry that a `between` join handles.
- `fct_survey_responses` keeps its inner join to `fct_survey_submissions`: that join is the filter to the submissions the fact models, not a key lookup.
- The widened terms test partitions by `type, code, name, region, school_id, grade_band` because the code is part of the sheet's unique key (`LIT3` and `PLIT4` may overlap under one name) and same-name term rows per school and band are legitimate; `name` alone would flag every `Q1`. It excludes null-bound rows because the macro fails them and the `not_null` question belongs to a separate test.

</details>


cbini:
Per-finding verdicts on the claude-review pass (comment 5682187526). The review ran against the pre-split branch; the Alchemer arm has since moved out of `int_surveys__survey_submissions.sql` into `int_surveys__alchemer_submissions.sql`, so line citations below map to the new file.

1. **PII tags missing on `int_surveys__survey_responses.yml`** — Accepted. The six `respondent_*` columns are selected through from the submissions model, which tags them. Added `config.meta.contains_pii: true` to each in 977756f20e.

2. **ST06 join-order in the Alchemer select** — Accepted. `rt.name` now precedes `ri.respondent_mail` in `int_surveys__alchemer_submissions.sql`. Same commit.

3. **`1 as round_rn` after the casts** — Accepted. The literal now sits before the two `cast()` targets. Same commit; trunk passes on the file.

4. **Comment on the `archive_submissions` union arm** — Declined. All three arms sit in one `all_submissions` CTE with identical 15-column lists, so the positional dependency is visible at the edit site. sqlfluff does not flag the order, and the repo's SQL-comment rule keeps inline comments to what the line cannot show.

5. **Cast timing on Alchemer `survey_id` / `response_id`** — Declined. Both ids join to INT64 columns on `stg_alchemer__survey_campaign` and `int_surveys__response_identifiers` before the final select. Casting at first read would force matching casts on the join partners; casting at the output is the earliest point the STRING value exists.

6. **Shared respondent-identity CTE across the two arms** — Declined. The Alchemer model is a static table over a decommissioned source and is headed for the archive pattern (disable, source, `ref` to `source`). Coupling the live Google Forms arm to it through a shared spine would have to be undone at archive time.

7. **Unit test on the positional UNION ALL** — Declined. The `unique` test on `survey_submission_key` (severity error) and the contract on `fct_survey_submissions` cover a swapped column in practice, and a three-arm fixture would cost more than it guards.

One note on the review's "checked and clear" list: the `mutually_exclusive_ranges` test on `stg_google_sheets__reporting__terms` no longer scopes to `type = 'SURVEY'`. It now covers every type, partitioned by `type, code, name, region, school_id, grade_band`, after 81f68ce8c0 and 41c43713e3.

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
cbini, line 1: alchemer models are still being computed against live models. they should be fully precomputed and only incorporated as a source that is union-ed

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__alchemer_submissions.sql
```diff
@@ -0,0 +1,111 @@
...
+
+    concat(
+        sr.survey_link_default, '?snc=', sr.session_id, '&sg_navigate=start'
+    ) as survey_response_link,
+from alchemer_responses as sr
+left join
+    {{ source("alchemer", "stg_alchemer__survey_campaign") }} as sc
+    on sr.survey_id = sc.survey_id
+    and sr.date_started between sc.link_open_date and sc.link_close_date
+inner join
```
cbini: inner join out of order

cbini (commented):
refactor all calcs to the correct altitude: the furthest upstream model possible.

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__survey_responses.sql
cbini: Done in 1d36ed07e1. `int_surveys__alchemer_submissions` now also carries the historic Manager archive arm plus `respondent_identifier` and `survey_submission_key`, and a new build-once `int_surveys__alchemer_responses` carries every per-submission column at question grain. Both live models union them in as plain columns; no Alchemer row is joined to a live model any more. They still build from `ref()` in this commit so CI can build them. One step remains: build both to prod from this branch (`uv run dbt build --select int_surveys__alchemer_submissions int_surveys__alchemer_responses --target prod --project-dir src/dbt/kipptaf`), then the flip commit disables the two models and their tests, declares both tables under the `surveys` source, and switches the two `ref()` calls to `source()`.

src/dbt/kipptaf/models/surveys/intermediate/int_surveys__alchemer_submissions.sql
cbini: Fixed in 1d36ed07e1. The inner join to `stg_google_sheets__reporting__terms` now precedes the left joins to the campaign, identifiers, roster, and directory tables.

cbini:
Re "refactor all calcs to the correct altitude": done in 1d36ed07e1. Each calc now sits on the furthest upstream model that has its inputs.

- `stg_google_forms__responses` derives `last_submitted_date` (UTC), `response_link`, and `respondent_local_part`. The submissions model reads them as plain columns; its `gforms_responses` CTE is gone.
- `int_google_forms__form__items` derives `item_abbreviation_lower`; `int_google_forms__form_responses` passes it through and derives `answer`, `answer_value`, and `is_open_ended`. The responses model selects them; its inline `safe_cast` / `coalesce` / `if` are gone, and `fct_survey_responses` reads `answer_value` instead of re-casting `answer`.
- `int_google_directory__users__addresses` is a new model, one row per (primary email, known address), replacing the alias-map CTE that both submission models had copied.
- `stg_google_sheets__google_forms__question_department_crosswalk` lowers `abbreviation` at staging, so the crosswalk join in the responses model compares plain columns. 0 of 309 prod rows were mixed-case, so values do not change.
- The Alchemer-side calcs (id casts, submit date, local part, answer columns) live in the two build-once Alchemer tables, which is as far upstream as a frozen source allows.

Stayed put, with reason: `round_rn` and the four respondent-identity `coalesce` columns depend on the terms and roster joins in the submissions model, and `respondent_identifier` / `survey_submission_key` are computed where the arms meet.

Verified against prod after the round: `fct_survey_submissions` (48,956 rows) and `fct_survey_responses` (580,663 rows) identical both ways; `dim_survey_administrations` differs on 1 of 56 rows, a 2019 deadline the 2026-09-15 sheet fix moved, which prod's submissions table predates.

PR #5306, opened 2026-09-14
refactor(dbt): read course and school names off base_powerschool__final_grades in rpt_deanslist__transcript_grades

cbini:
## Summary & Motivation

When merged, this pull request will drop 2 joins from `rpt_deanslist__transcript_grades` that rebuild PowerSchool-internal lookups inside kipptaf. The in-progress grades branch joined `stg_powerschool__courses` for the course name and `stg_powerschool__schools` for the school name. `base_powerschool__final_grades` has carried both columns since 2f1b4a6ea6 (2025-08-19), so the model now reads them directly.

This closes row 2 of #5285. Row 1 (`rpt_tableau__student_course_grades`) is a separate change and stays open.

## Reviewer Notes

- The school name now comes from the section's school instead of the student's enrollment school. Against prod today, that changes 7 rows, all one Camden student. Details in the fold-out. The storedgrades branch already reported the section's school, so the 2 branches now agree.
- The `_dbt_source_project` join predicates on the dropped joins go with them. No other join changes.

### General

- [x] Update **due date** and **assignee** on the
      [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
- [ ] Review the **Claude Code Review** comment posted on this PR. Address valid
      feedback; dismiss false positives with a brief reply explaining why.
      (Claude is advisory — use your judgement, but don't ignore it.)

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — no new models
- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — no exposure change; the model's output columns are unchanged
- [x] **Breaking change?** No. Contract columns and types are unchanged.
- [x] If adding or modifying an external source, run `stage_external_sources` — no external source change

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.
- [ ] **Dagster Cloud** — passes or not triggered.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed.

**Why the joins existed.** The model landed in one commit on 2023-08-11 (45fb83c91a, PR #243). At that date the package `base_powerschool__final_grades` projected neither `course_name` nor `school_name`. Commit 2f1b4a6ea6 added `sch.name as school_name` to `base_powerschool__sections` and carried `courses_course_name` and `school_name` through `base_powerschool__final_grades`. Nobody revisited this model.

[...]

PR #5321, opened 2026-09-15
refactor(dbt): move the student course grade joins into the powerschool package

### LINKED ISSUE #4687: feat(dbt): resolve category driver at Y1 and backfill prior-year running grades

### LINKED ISSUE #5285: refactor(dbt): move the last 2 PowerSchool-internal joins from kipptaf into the package
## What and why

#5260 set out to move 6 PowerSchool-internal joins out of kipptaf and into the shared `powerschool` package. It closed with 4 of them done. Rows 1 and 2 of its table were never shipped, and they are the two this issue carries.

Both models still rebuild a PowerSchool-to-PowerSchool join in kipptaf. They produce correct output today, so this is a placement fix, not a bug fix — the same framing #5260 used.

| Model | Join it rebuilds |
| --- | --- |
| `rpt_tableau__student_course_grades` | `base_powerschool__course_enrollments` to `base_powerschool__final_grades` |
| `rpt_deanslist__transcript_grades` | `stg_powerschool__storedgrades` to `stg_powerschool__courses` and `stg_powerschool__schools` |

The two are not equally hard, and #5260's table described only one of them correctly.

## Proposed approach

Two separate PRs. Do the second one first — it is the easy one.

### `rpt_deanslist__transcript_grades` is the swap the table described

`stg_powerschool__storedgrades` really does carry `course_name` and `schoolname` natively, alongside `course_number` and `schoolid`. Confirmed against the Newark relation. So the joins to `stg_powerschool__courses` and `stg_powerschool__schools` at lines 61 and 65 can be dropped and the columns read straight off storedgrades.

One thing to check before starting: this model also inner-joins `stg_powerschool__students` at lines 23 and 52, and #5228 dropped Miami from that union. The model lost its 48,486 Miami history rows as an accepted consequence, recorded on #5012. Do not try to restore them here.

### `rpt_tableau__student_course_grades` is not the swap the table described

#5260 said `base_powerschool__final_grades` "already carries enrollment attributes". It does carry some, but not the ones this model needs, and it does not carry the years it needs.

Measured against `main` and the Newark relation:

- **4 columns are missing.** `base_powerschool__final_grades` has no `sections_dcid`, no `section_number`, no `external_expression`, and no `rn_course_number_year`. None of the 4 appear in its output, and none exist on the built relation.
- **It is current-year only.** Line 56 filters `enr.cc_academic_year = {{ var("current_academic_year") }}`.
- **The consumer needs 2 years.** `rpt_tableau__student_course_grades` reads the prior year in several places — lines 104, 306 and 362 use `current_academic_year - 1`, and lines 362 and 363 together span prior year through current year.

So the move needs a real decision first: widen the package model, add a second package model at a different grain, or leave this one in kipptaf and record why. Budget design time rather than treating it as a swap.

<details>
<summary>For Claude</summary>

Context, not instructions.

- Parent is #5012. #5260 is closed and stays closed; this issue carries only its unshipped rows 1 and 2.
- Prior art for the pattern: `docs/superpowers/specs/2026-09-11-powerschool-section-teachers-design.md` and `docs/superpowers/specs/2026-09-11-powerschool-gradebook-assignment-scores-design.md`. Both follow the same shape — one package model carrying every PowerSchool-internal join, a kipptaf `union_relations` wrapper above it, and the kipptaf-only joins left in kipptaf.
- Re-run the diagnostic before designing. #5260's table was wrong about 3 of its 6 rows, and each error would have shipped a defect. The two claims above were re-measured on 2026-09-14 and held; re-measure anyway, since `base_powerschool__final_grades` may gain columns.
- Verification method that worked twice: compile the package model at `--target prod` per district and run its compiled SQL as a CTE against production relations, fingerprinting per academic year. Do NOT compare a `zz_stg` staging copy against prod — that measures staging staleness, not the change, and it produced a convincing false alarm on #5281. Accumulate the fingerprint as `sum(cast(farm_fingerprint(...) as bignumeric))`; a plain `sum` overflows above roughly 20M rows.
- A new package model needs a `dbt build --target staging` seed per NJ district before kipptaf CI can resolve its source, and so does an existing package model that gains a column. That writes shared tables and needs direct user authorization.
- No Miami and no archive rebuild, unless a kept history union gains a column kipptaf reads for Miami. `kippmiami` does not consume the `powerschool` package. Recipe in `src/dbt/kippmiami/CLAUDE.md`.
- Acceptance: each moved model is row-identical to prod for NJ, proved against production relations, and the kipptaf reader is a `union_relations` wrapper or a consumer of one.

Refs #5012
Refs #5228
Refs #5260

</details>

cbini:
## Summary & Motivation

When merged, this pull request will move every PowerSchool-to-PowerSchool join behind `rpt_tableau__student_course_grades` into a new `powerschool` package model, `int_powerschool__student_course_grades_spine`, read through a kipptaf union wrapper. Closes row 1 of #5285; row 2 shipped in #5306. Refs #5285. Closes #4687. Follow-up for the closed-year running grade in #5330.

The package model runs once per district (Newark, Camden, Paterson) at the consumer's grain: one row per student, year, course, and quarter. The kipptaf wrapper unions the 3 district copies. The consumer now joins the roster to the wrapper plus 2 enrichment models, so its select list, column order, and contract do not change. Verified row-identical to prod for AY2025 in all 3 districts; every AY2026 residual traced to an enrollment-snapshot build lag, not to logic, except the Paterson category-grade rows described below, which are an intended change.

## Reviewer Notes

- **Term spine.** The package fans each enrollment over `int_powerschool__terms` keyed on the section's school. Prod used `int_students__terms` keyed on the roster's school. Both yield Q1 to Q4 plus Y1 per NJ school year, and the consumer's left join on `quarter` drops any extra package row.
- **`teacher_number` and `teacher_name`** are aliased in the package select, so the consumer reads them under their existing names.
- **`section_or_period` stays in kipptaf.** It depends on the roster's grade level for the year, which the enrollment row does not carry. The package exports `section_number` and `external_expression` only.
- **2 fixes from the prod comparison, plus 1 intended change.** `need_next` returned to its original ordinal. The package normalizes Paterson's spelled-out credit types (`ELA`, `Math`, `Science`, `Homeroom`) to the PowerSchool abbreviations inline, the same CASE `int_students__course_enrollments` applies. Paterson now carries category grades: the kipptaf `int_powerschool__category_grades` wrapper excludes Paterson on a stale #3908 note, so Paterson AY2026 gains about 5,000 category-grade rows relative to prod.
- **Uniqueness test** on the package model is `severity: warn` with the `TODO(#3915)` comment. The prior-year storedgrades double-write duplicates move with the data.

### General

### dbt

- [x] Include (or update) an [exposure](https://teamschools.github.io/teamster/reference/dbt-conventions/#exposures) for all models consumed by a dashboard, analysis, or application (consumer unchanged; existing exposure still applies)
- [x] **Breaking change?** No. The consumer's contract is unchanged.
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** so the dbt Cloud CI job can find the table. The 3 new `zz_stg_<district>_powerschool` sources were seeded with `dbt build --target staging` per district.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed, via subagent-driven development.

Spec: `docs/superpowers/specs/2026-09-14-powerschool-student-course-grades-design.md` Plan: `docs/superpowers/plans/2026-09-14-powerschool-student-course-grades.md`

**Verification method.** Compiled package SQL per district run as CTEs against prod; compiled consumer with the wrapper relation swapped for those CTEs; `except distinct` both ways against `kipptaf_tableau.rpt_tableau__student_course_grades`; column order asserted against `INFORMATION_SCHEMA.COLUMNS` (131 columns, identical).

**Final round (after the 3 fixes), prod_only / new_only rows:**

| District | AY2025 | AY2026    |
| -------- | ------ | --------- |
| Camden   | 0 / 0  | 15 / 66   |
| Paterson | 0 / 0  | 6 / 6     |
| Newark   | 0 / 0  | 158 / 170 |

Row and duplicate-key counts: Paterson 27,709 rows / 443 dup keys both sides. Camden 251,508 vs 251,457 rows (net +51 = the except-distinct net), 84 dup keys both sides. Newark 686,249 vs 686,237 (net +12), 0 dups both sides. Zero value differences on matched keys in every district.

**Lag tie-back.** Every AY2026 residual key (39 Camden, 118 Newark) exists in exactly one of the two enrollment snapshots, or differs between them on `cc_sectionid`, `sections_section_number`, `sections_external_expression`, `cc_dateenrolled`, `teachernumber`, or `teacher_lastfirst`. The package reads the live district `base_powerschool__course_enrollments` (Newark 14:34Z, Camden 13:45Z on 2026-09-15); prod read the kipptaf `int_students__course_enrollments` copy (08:05Z). The 28 `section_or_period` and 10 enrichment-on-ungraded rows are a subset of the same set. No residual differs on `credit_type`.

**Earlier rounds.** Paterson +5,001 AY2026 rows: category grades the kipptaf wrapper excluded on a stale #3908 note; kept as intended after review. Symmetric current-year diffs in every region: `need_next` had moved from ordinal 127 to 125, fixed. Paterson AY2025 1,020 rows on `credit_type` only: raw vs normalized, fixed by normalizing in the package. Camden 384 `quarter_course_percent_grade` diffs in a matched-key join: artifact of 12 non-unique AY2025 keys (#3915), except-distinct is 0.

**Slot time.** Round 3: 12 jobs, 2.65 slot hours, 9.99 GiB billed.

**Staging seeds.** `zz_stg_kippnewark_powerschool`, `zz_stg_kippcamden_powerschool`, `zz_stg_kipppaterson_powerschool` `.int_powerschool__student_course_grades_spine`, built with `dbt build --target staging` one district per call.

**Review round (b7fcc4dd93).** Dropped the `powerschool_has_category_grades` var, inlined the credit-type CASE in the package and deleted the macro, reverted `int_students__course_enrollments` to main, term spine reads `stg_powerschool__terms` for Y1 and drops both `distinct`s (grain verified unique in all 3 districts). The AY2025 0/0 result predates this round; the changes are value-preserving by construction except the Paterson category rows.

**Review round (4243f6f47e).** Year window moved to reporting altitude: the package model now carries every academic year and the consumer's existing `student_roster` filter (current-1..current) is the only window. The #4687 backfill is removed on both sides per its runbook (five `backfill_*` CTEs and `bfc` join in the package, `backfill_running_gpa`, `bfg` join, and the `gpa_y1` coalesce in the consumer); `y1_course_in_progress_*` return to null on stored-grade rows, `gpa_y1` to the stored value, and `y1_course_letter_grade_adjusted` to null on prior-year Q1-Q4 rows. Category-driver CTEs and `courses_gradescaleid` untouched. Prod comparison for AY2025 will now differ on those columns by design. Dry run of the compiled Newark SQL: 579 MB, down from 609 MB.

**Review round (d8ed033fa8).** Package model renamed to `int_powerschool__student_course_grades_spine`: it consumes `base_powerschool__final_grades` rather than sitting beside it, and the suffix names that role. Package model, kipptaf wrapper, the three NJ source entries, the consumer ref, and the spec and plan move together. The three `zz_stg_*` seeds were rebuilt under the new name (Newark 4.2M rows, Camden 1.1M, Paterson 50k, all years); the old-name seeds remain as orphans.

**CI note (d8ed033fa8).** The dbt Cloud failure on that commit was `int_people__staff_roster_history` reading a `zz_stg_kipptaf_adp_workforce_now` clone that Clone - Staging had not finished writing (the race #5329 documents); the model is untouched here and the same run failed on a sibling PR at the same minute. Merged main (7d5403b8ab) and re-ran.

**Review round (b8f39676bb).** `int_powerschool__category_grades` (kipptaf) now unions `kipppaterson_powerschool`; the #3908 exclusion note predates the plugin install and the `_pivot` sibling already carried Paterson. Compiled `--target staging` against all three `zz_stg_*` clones with no null-filled columns.

**#4687 runbook check (PR schema at 7d5403b8ab vs prod, student-term grain, students present in both).** AY2026 Q1 and Y1: `gpa_y1`, `gpa_for_quarter`, `gpa_n_failing_y1` identical on all 9,504 student-terms each. AY2025 Q1-Q4: `gpa_y1` differs on roughly half the student-terms (running value reverts to the stored year-end value, by design); AY2025 Y1 identical on all 9,094; `gpa_for_quarter` and `gpa_n_failing_y1` identical everywhere. `need_next_*` null on every AY2025 row. `lowest_category_y1_name` non-null on 36,787 AY2025 Y1 rows (prod 33,217; the delta is Paterson). AY2026 row-count and category-driver deltas in the PR schema trace to stale `zz_stg_*` clones, not this branch.

**Deferred minors.** Consumer select list groups g, r, f while joins are g, f, r (plan-mandated order).

</details>


cbini:
Per-finding verdicts on the claude-review comment, each checked against the checkout or prod.

**1. `term_spine` inner join keyed on the section's school: no change.** The old consumer also inner-joined its term spine (`int_students__terms` on the roster's `schoolid` and `yearid`, old file line 282), so the join shape did not change, only the school key did. Ran the suggested query: every `cc_schoolid`/`cc_yearid` pair in `base_powerschool__course_enrollments` with `cc_sectionid > 0` for yearids 35 and 36 (the model's 2-year window) has rows in `int_powerschool__terms` in all 3 districts. The only gap is Paterson yearid 34 (2 school-years, 3,394 enrollments), outside the window. Consistent with the AY2025 0/0 reconciliation in the PR body.

**2. No unit tests on the new model: declined for this PR.** The CTEs moved verbatim from `rpt_tableau__student_course_grades`, which never carried unit tests, and the PR's acceptance test was row-level parity against prod. Adding unit tests for the backfill reconstruction and grade-band ladder is a reasonable follow-up, but it is a separate change from the move.

**3. Consumer select-group order vs join order: deferred, as disclosed.** Plan-mandated order; sqlfluff passes.

**4. Runtime `{{ var() }}` boolean vs compile-time `{% if %}`: deferred, as disclosed.** The CTE must exist for the downstream joins, so the compile-time form would need an empty-CTE branch. Zero-row scan for Paterson.

**5. `y1_final_grades` cast placement: no change.** Carried over verbatim; trunk `sqlfluff` is clean on the file.

**6. Test severity: confirmed, matches prod.** The old consumer's test also had no explicit `severity:`; `kipptaf`, `kippnewark`, `kippcamden`, and `kipppaterson` all set `data_tests: +severity: warn` at the project level, so the inherited behavior is identical. The staging seeds showed Camden and Paterson WARN, Newark PASS.

Findings 7 to 13: checked-and-clear, agreed.

src/dbt/kipppaterson/dbt_project.yml
```diff
@@ -36,6 +36,10 @@ vars:
   pearson_state_assessment_relations:
     - int_pearson__njsla
     - int_pearson__njsla_science
+  # No GradeBook plugin deployed, so category grades are unusable here. The
+  # kipptaf int_powerschool__category_grades wrapper already omits Paterson
+  # for the same reason (#3908).
```
cbini: i don't believe this is correct. all powerschool regions should have category grades. remove this var.

src/dbt/kipptaf/macros/utils.sql
```diff
@@ -20,3 +20,20 @@
         0
     )
 {%- endmacro %}
+
+{# PowerSchool credit types are conformed to the Focus vocabulary so the two SIS
```
cbini: PowerSchool credit types are not conformed to the Focus vocabulary. If true, that is purely incidental. Is this macro really necessary, or is there an upstream model that can implement the case statement inline?

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
```diff
@@ -0,0 +1,577 @@
...
+            sections_external_expression as external_expression,
+            courses_credittype as credit_type,
+            courses_course_name as course_name,
+            courses_excludefromgpa as exclude_from_gpa,
+            teachernumber as teacher_number,
+            teacher_lastfirst as teacher_name,
+        from {{ ref("base_powerschool__course_enrollments") }}
+        where
+            cc_academic_year >= {{ var("current_academic_year") - 1 }}
+            and cc_academic_year <= {{ var("current_academic_year") }}
```
cbini: is there a significant cost difference if we drop this filter? it would be more appropriate at the reporting altitude.

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
```diff
@@ -0,0 +1,577 @@
...
+                'LOG300',  -- Study Hall
+                'LOG9',  -- Lunch
+                'SEM22106G1',  -- Advisory
+                'SEM22106S1'  -- Not in SY24-25 yet
+            )
+    ),
+
+    term_spine as (
+        /* grain projection, not dup-masking: (schoolid, yearid, term) */
+        select distinct schoolid, yearid, term as `quarter`,
```
cbini: is this grain not unique?

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
```diff
@@ -0,0 +1,577 @@
...
+        from {{ ref("int_powerschool__terms") }}
+        where term like 'Q%'
+
+        union all
+
+        /* grain projection, not dup-masking: (schoolid, yearid) */
+        select distinct schoolid, yearid, 'Y1' as `quarter`,
```
cbini: i believe you can get the correct grain by using staging

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
```diff
@@ -0,0 +1,577 @@
...
+            earnedcrhrs as y1_course_final_earned_credits,
+            potentialcrhrs as y1_course_final_potential_credit_hours,
+            gpa_points as y1_course_final_grade_points,
+        from {{ ref("stg_powerschool__storedgrades") }}
+        where
+            storecode = 'Y1' and academic_year >= {{ var("current_academic_year") - 1 }}
+    ),
+
+    backfill_quarter_running as (
+        /* TODO(#4687): TEMPORARY. Delete this CTE, backfill_course_anchored,
```
cbini: verify this whole chain is still necessary. patching a view like this is bad design regardless. would dropping year filters altogether resolve what it's trying to fix?

src/dbt/kipppaterson/dbt_project.yml
cbini: Removed in b7fcc4dd93 (var, both dbt_project.yml entries, and the WHERE gate). Verified first: kipppaterson_powerschool.int_powerschool__category_grades has rows in every year (yearid 34 to 36), so the exclusion was wrong. Effect: Paterson AY2026 gains about 5,000 rows carrying category grades relative to prod, which is the +5,001 the round-1 comparison flagged. Separately, the kipptaf int_powerschool__category_grades wrapper still omits Paterson on the same stale #3908 note; that is a gradebook-audit change, so I left it for a follow-up.

src/dbt/kipptaf/macros/utils.sql
cbini: Agreed, the Focus framing was wrong; the CASE normalizes Paterson's spelled-out PowerSchool values (ELA, Math, Science, Homeroom) to the PowerSchool abbreviations. In b7fcc4dd93 the CASE now lives inline in the package model's course_enrollments CTE, the macro is deleted, int_students__course_enrollments is back to its main version byte for byte, and the consumer reads the wrapper directly (no course_grades CTE).

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
cbini: Measured: a dry run of the compiled Newark package SQL processes 609 MB with the filter and 609 MB without it. None of the upstream tables are partitioned by year, so the filter buys no scan pruning; it only trims the joined and materialized rows. Newark base_powerschool__course_enrollments holds about 436k rank-1 enrollments across 2004 to 2026 versus 83k in the 2-year window, so the package table would be roughly 5x the rows. Moving it to the consumer also changes the storedgrades branch, which is hard-wired to the prior year for the #4687 backfill, so older years would carry Y1 finals but no quarter grades unless that branch widens too. Not changed in this PR. Say the word and I will do it here or as a follow-up.

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
cbini: It is unique. int_powerschool__terms has one row per (schoolid, yearid, term) in all 3 districts (Newark 528 of 528, Camden 288 of 288, Paterson 16 of 16). The distinct is gone in b7fcc4dd93.

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
cbini: Done in b7fcc4dd93. The Y1 branch now reads stg_powerschool__terms where isyearrec = 1 and schoolid != 0, which is unique on (schoolid, yearid) in all 3 districts, so no distinct on either branch.

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
cbini: Not removed in this PR; the chain moved verbatim. It reconstructs a prior-year running Y1 percent that PowerSchool never stored (storedgrades holds only the final Y1), so dropping year filters would not replace it; the data does not exist upstream. #4687 is the removal runbook: delete once the dashboard runs on current-year data, and it warns that y1_course_letter_grade_adjusted reverts to null at removal with no backfill_ marker. Whether the prior-year view is still needed is a T&L call. If it is not, I would do the removal as its own PR per that runbook.

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
cbini: Done in 4243f6f47e. The package model now carries every academic year: the window is gone from `course_enrollments`, the `academic_year` floor from `y1_final_grades`, and the `yearid` floor from `category_grades`. The 2-year window already lived in the consumer (`student_roster` filters `enr.academic_year` to current-1..current, and the package join is a left join off that roster), so nothing had to be added there. Dry run of the compiled Newark SQL scans 579 MB, down from 609 MB with the window, because the backfill's extra `storedgrades` passes went with it.

src/dbt/powerschool/models/sis/intermediate/int_powerschool__student_course_grades.sql
cbini: Removed in 4243f6f47e, both halves, following the #4687 runbook: the five `backfill_*` CTEs and `bfc` join in the package model, and `backfill_running_gpa` with its `bfg` join and the `gpa_y1` coalesce in the consumer. Branch 3 of `quarter_grades` is back to `cast(null ...)` for the two in-progress columns and now covers every completed year. `courses_gradescaleid` and the category-driver CTEs are untouched. `y1_course_letter_grade_adjusted` reverts to null on prior-year Q1-Q4 rows, as the runbook says it will; its description and the three backfill descriptions are updated. PR body carries `Closes #4687`.

PR #5322, opened 2026-09-15
refactor(amplify): move the Amplify-only value logic into the package and alias the frozen archive at kipptaf

### LINKED ISSUE #5305: refactor(amplify): move the Amplify-only value logic into the package and alias the frozen archive at kipptaf
## What and why

The original framing of this issue was to move most of `src/dbt/kipptaf/models/amplify/` into per-district `int_` models in the `src/dbt/amplify` package. That framing assumed each district's SFTP file carried that district's rows. It does not: Amplify exports one network account, and the file lands in Newark's bucket. Newark's file holds Camden and Miami rows for AY2025 and Camden and Paterson rows for AY2026, while Paterson's own file stopped at AY2025. Region can only be resolved in kipptaf through the location crosswalk, the base-plus-aimline combine only works after the kipptaf union, and the frozen API archive is a kipptaf-level table that needs the same unpivot.

So the scope is narrowed to what is actually Amplify-only and district-independent, plus a guard against the double-count that appears the day Paterson's feed resumes.

## Proposed approach

- Package `stg_amplify__mclass__sftp__pm_student_summary`: `device_date` falls back to `sync_date` (via `select * replace`), and the surrogate key widens to student, school year, period, measure, probe, device date, assessment grade so it is unique. Both are value-only, so no contract change and no `zz_stg` seeding. `unique` tests at severity error on the PM and benchmark package staging models.
- kipptaf: two new `stg_amplify__mclass__api__*` wrappers over the frozen archive that alias its API column names to the live SFTP names. The two mClass intermediates become a plain union of SFTP wrapper plus API wrapper, crosswalk join, and Focus offset, with every `coalesce` pair removed.
- kipptaf: severity-error natural-key uniqueness tests on both mClass intermediates, so a resumed Paterson feed fails CI instead of doubling silently.
- Nothing outside `models/amplify/` changes. The unpivot, the aimline combine, DDS, and the gate joins stay where they are.

Design: `docs/superpowers/specs/2026-09-14-amplify-package-intermediates-design.md` on the linked branch.

<details>
<summary>For Claude</summary>

> Live SFTP names win; the archive is aliased once at kipptaf. `select * replace` cannot rename (BigQuery: `Column device_date in SELECT * REPLACE list does not exist`), so renames are `except` plus `<old> as <new>`.
>
> Consumer edits, all inside `models/amplify/`: the unpivot's 5 `reading_comprehension_maze_*` refs become `basic_comprehension_maze_*`; `int_amplify__benchmark_student_summary` reads `bss.device_date as client_date`; `int_amplify__all_assessments` reads `p.device_date` as `client_date` in the Internal branch's select and window join. `int_amplify__benchmark_student_summary`, `int_amplify__all_assessments`, and the `pm_met_criteria` pair keep their output columns, so the 17 kipptaf consumers are untouched.
>
> Verification: row count and distinct natural key on both mClass intermediates, `int_amplify__all_assessments`, and `rpt_tableau__dibels_dashboard` must match prod; column-by-column diff on the intermediates with renamed columns paired. The one expected value change is the PM `surrogate_key` hash.
>
> Docs: rewrite `src/dbt/amplify/CLAUDE.md` (kipptaf does not keep native amplify models; one network account lands in Newark's bucket); update the DIBELS skill paragraph about the PM surrogate key collision.
>
> Open: PII tag scope on the 5 tagged models (default tiers 1 to 3); whether Paterson Preparatory is permanently inside the network account, which would let the Paterson package models and `kipppaterson_amplify` sources be disabled in a follow-up.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will move the Amplify-only value logic into the `amplify` package, alias the frozen AY2025 archive at kipptaf, and stop computing surrogate keys anywhere below the kipptaf marts.

Amplify exports one network account, and that export lands in Newark's bucket. So the per-district framing in the mClass intermediates was wrong and is dropped here.

The package edits are value-only: a `device_date` fallback, natural-key uniqueness tests at `error`, and PII tags. The `surrogate_key` column is gone from every amplify staging model and kipptaf amplify intermediate. The one join that used it, benchmark rows to their unpivot, now joins on the four natural-key columns.

kipptaf gets 2 archive wrappers that alias the API column names to the SFTP names. The 2 mClass intermediates lose every coalesce pair and gain natural-key tests that fail loudly if Paterson rows ever arrive through both files.

No model outside `models/amplify/` changes. The only other edits are test-severity config in every district and package (Task 8, see below) and docs.

Closes #5305

## Reviewer Notes

The kipptaf PM wrapper keeps a `coalesce(device_date, sync_date)` that duplicates the package's fallback on purpose. It covers the deploy window between code locations.

During that same window the district tables still carry `surrogate_key`, so the kipptaf wrappers pass it through until each district rebuilds. The wrappers are not contract-enforced and nothing reads the column, so it is harmless and disappears on its own.

Task 8 is a network-wide test-severity change riding in this PR by the author's choice. Package tests now default to `error`. 17 package tests that fail in prod today are pinned to `warn` with a `TODO(#5305)`, and 3 more that documented `warn` as their intent now declare it. Details in the "For Claude" section.

Verification, prod vs dev:

| Model | Prod | Dev |
| --- | --- | --- |
| `int_amplify__mclass__benchmark_student_summary` | 59,947 | 59,947, natural keys identical |
| `int_amplify__mclass__pm_student_summary` | 132,107 | 132,107, natural keys identical |
| `int_amplify__benchmark_student_summary` | 337,512 | 337,512 |
| `int_amplify__mclass__pm_student_summary_aimline` | 68,111 | 68,111 |
| `int_amplify__all_assessments` | 410,540 | 410,540, 48 rows differ |
| `rpt_tableau__dibels_dashboard` | 546,066 | 545,894 |

The 48 changed rows and the dashboard delta are time drift, not code. Prod `int_amplify__all_assessments` was last built 2026-09-14 18:47 UTC. Its prod PM upstreams rebuilt 2026-09-15 07:54 and the DIBELS goal sheets rebuilt 2026-09-15 14:49, and dev read those. The changed columns are goal, aimline status, and model type, which come from the goal sheets. Before the review changes, at commit 5803e355e7, the dashboard was byte-identical in both directions.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — present for the 2 new kipptaf archive wrappers
- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — no exposure change
- [x] **Breaking change?** No contracted column rename. `surrogate_key` is dropped from non-contracted staging and intermediates only; no mart, exposure, or Cube view read it.
- [x] If adding or modifying an external source, run `stage_external_sources` with `--target staging` — no external source change

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Spec: `docs/superpowers/specs/2026-09-14-amplify-package-intermediates-design.md` (see its review amendment). Plan: `docs/superpowers/plans/2026-09-14-amplify-package-intermediates.md`.

**Task 8: test severity.** Every district `dbt_project.yml` had an unscoped `data_tests: +severity: warn`. dbt lets that root-project config override a package's resource-level `severity: error`, so every package test resolved to `warn` when built through a district. Now each district scopes the block under its own project name, and the 12 source packages declare no project-level severity, so a package test resolves dbt's default, `error`, unless the test says otherwise. Verified by `dbt ls` on kippcamden and kippnewark: an unset package test resolves `ERROR`, a declared `error` stays `error`, and the pinned tests resolve `warn`.

Measured against the `<district>_dbt_test__audit` stored-failure views: 17 distinct package tests that would newly resolve `error` have failing rows today (Newark 11, Camden 8, Miami 8, Paterson 6; the largest is `base_powerschool__course_enrollments__no_studyear_course_overlap` at 8,441 rows in Newark). Each is pinned to `warn` in its package with `TODO(#5305): failing today; resolve then drop the warn`. 4 of them already declared `warn` with their own rationale and were left alone. 3 further tests documented `warn` as intended behavior without declaring it (`int_finalsite__enrollment_lifecycle__status_report_not_stale`, `test_focus__idea_educational_environment_classified`, `test_int_powerschool__gpa_cumulative_year__reconciles_stored`) and now declare it. Caveat: point-in-time snapshot.

Two things noticed and left alone: no district-owned test lacks a declared severity today, so the scoped district `warn` default governs zero tests; and kipptaf still carries an unscoped `data_tests: +severity: warn`, harmless while kipptaf installs no source packages.

**Surrogate keys.** Removed from 5 package staging models (sftp and api, benchmark, PM, aimline), the kipptaf sftp aimline wrapper, both kipptaf api archive wrappers (added to their `except` list; the BQ-native archive always carries the column), `int_amplify__mclass__pm_student_summary_aimline`, `int_amplify__mclass__benchmark_student_summary_unpivot`, `int_amplify__benchmark_student_summary`, and `int_amplify__all_assessments`. The disabled `int_amplify__dds__data_farming_unpivot` still computes one and is out of scope. Staging uniqueness moved to `dbt_utils.unique_combination_of_columns` on the natural key.

**PII scope.** Tiers 1 to 3 confirmed by the author.

**Paterson.** Paterson Preparatory is inside the network account. Its own AY2025 file is kept as the archive of those rows, so the Paterson package models stay enabled.

</details>


cbini:

1. **Missing PII tag on the two mClass intermediates.** Confirmed. Both properties files had no `contains_pii` tag and both carry name, birth-date, and program-flag columns. Fixed in f816c4c207 with a model-level `config.meta.contains_pii: true` on each, matching the tag this PR added to the package staging models.

2. **Aimline surrogate key recomputed at kipptaf altitude.** Agree it is not a defect. The aimline hash runs on `student_primary_id` after the Miami offset, so it cannot be computed in the package. No change.

3. **Reviewer could not run trunk or dbt.** Whole-branch `trunk check --force --no-fix` ran clean before the PR opened, and again on the two files in the fix commit. The row counts in the description came from prod-vs-dev queries run during Task 5.

cbini:
you fixed the test severity in the wrong direction. error severity from the package should persist into the installed projects

cbini:
surrogate_keys should be computed only in the kipptaf marts

cbini:
Both comments addressed. Four commits, PR body rewritten to match.

**Test severity** (30bd79b508, d9a4596497, 9f913a67a2). The 12 source packages no longer declare a project-level severity, so a package test resolves dbt's default, `error`, unless the test itself says otherwise. Districts keep their scoped `warn` default. Verified with `dbt ls` on kippcamden and kippnewark: unset package test resolves `ERROR`, declared `error` stays `error`.

Before flipping, I measured the stored-failure views. 17 package tests that would newly go red have failing rows in prod today (Newark 11, Camden 8, Miami 8, Paterson 6; largest is `base_powerschool__course_enrollments__no_studyear_course_overlap` at 8,441 rows in Newark). Per your call, each is pinned to `warn` in its package with `TODO(#5305): failing today; resolve then drop the warn`. 3 more tests documented `warn` as the intended behavior but never declared it (`int_finalsite__enrollment_lifecycle__status_report_not_stale`, `test_focus__idea_educational_environment_classified`, `test_int_powerschool__gpa_cumulative_year__reconciles_stored`); they now declare it. Full list is in the PR body's For Claude section.

Two things left alone for you to decide separately: no district-owned test lacks a declared severity today, so the scoped district `warn` default governs zero tests; and kipptaf still has an unscoped `data_tests: +severity: warn`, harmless while it installs no packages.

**Surrogate keys** (2623f6b34c). `surrogate_key` is gone from all 5 amplify package staging models, the kipptaf sftp aimline wrapper, both api archive wrappers, and the 4 kipptaf amplify intermediates that carried it. The benchmark-to-unpivot join now uses the 4 natural-key columns. Staging uniqueness moved to `unique_combination_of_columns` on the natural key. Nothing in the marts, exposures, or Cube read the column. During the deploy window the district tables still carry it and the non-contracted kipptaf wrappers pass it through, which is harmless.

Prod vs dev after the change: the 4 amplify intermediates match prod exactly on rows and natural keys. `int_amplify__all_assessments` and the DIBELS dashboard differ slightly because prod `all_assessments` last built 2026-09-14 18:47 UTC, before this morning's PM and goal-sheet rebuilds that dev read. Timestamps in the PR body.

PR #5324, opened 2026-09-15
feat(gradebook-audit): add an all-weeks calendar grid gsheet view

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will add `rpt_gsheets__gradebook_audit_all_weeks`, a new view that lists every school week of the current academic year.

The existing `rpt_gsheets__gradebook_audit_template` only shows weeks that have already finished. It works that way because it joins `stg_powerschool__u_expectations` to attach the W, H, F, and S counts, and that join drops any week the counts do not cover yet.

The new view drops that join. It carries no counts, just the calendar: academic year, region, school level, quarter, week number in the quarter, the Monday the week starts, and the last in-session day of the week. That means it runs through the end of June instead of stopping at last week.

Coverage is the same set the gradebook audit already uses — Newark and Camden at MS and HS, Paterson at MS. Miami and elementary schools are excluded.

It publishes as a third tab on the same Expectations Upload Template spreadsheet, so it was added to that exposure rather than given its own. The tab itself still needs to be created in the spreadsheet as a Connected Sheet pointing at the new view. Dagster does not write these tabs.

## Reviewer Notes

- `week_end_friday` keeps the template's `max(school_week_end_date)` aggregation, so it is the region's latest in-session day. Without it the view would return one row per school.
- The two exclusions are copied from `rpt_tableau__gradebook_audit`, so this view and the dashboard agree on scope. See "For Claude" for the exact predicates.

### General

- [ ] Review the **Claude Code Review** comment posted on this PR. The bot could not run — the API credential is out of credit. See "For Claude" below.

### dbt

- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application
- [x] **Breaking change?** No. This adds a model and one line to an exposure. Nothing is renamed or removed.
- [x] If adding or modifying an external source, run `stage_external_sources` with `--target staging`. Not applicable — no external source changes.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

**Scope predicates.** `_dbt_source_project != 'kippmiami'` and `school_level != 'ES'`, both in the model's `WHERE`. These mirror `category_join`'s `WHERE` in `rpt_tableau__gradebook_audit` and the matching filters in `int_extracts__gradebook_audit_student_flags`, except that those use `school_level_alt` (from `int_extracts__course_schedule_by_term`) while this model uses `school_level`, which is what `int_students__calendar_week` carries. Same intent, different upstream column.

Note that `school_level != 'ES'` also drops rows with a null `school_level`, since a null comparison is never true. After the Miami exclusion there are none — see the upstream gap below — but it is worth knowing the predicate is doing double duty rather than assuming nulls are handled somewhere.

**Verification.** Built and tested in dev from the worktree, after the scope change:

`dbt build --select rpt_gsheets__gradebook_audit_all_weeks --target dev --defer --favor-state --state <main repo prod manifest>` → PASS=2, including the `dbt_utils.unique_combination_of_columns` test on `academic_year, region, school_level, quarter, week_number_quarter`. That test passing is the meaningful result: it proves no region has two different `week_start_monday` values for the same quarter week, which is the divergence the `max()` aggregation could otherwise hide. `trunk check --force --no-fix` clean on all files, both commits.

Row counts off the dev view, all four quarters present in every group, zero null school levels:

| region | school level | weeks | first Monday | last Friday |
| Camden | MS | 41 | 2026-08-17 | 2027-06-17 |
| Camden | HS | 41 | 2026-08-17 | 2027-06-17 |
| Newark | MS | 40 | 2026-08-24 | 2027-06-17 |
| Newark | HS | 40 | 2026-08-24 | 2027-06-17 |
| Paterson | MS | 40 | 2026-08-24 | 2027-06-17 |

**A pre-existing upstream gap, now out of scope for this model but still real.** Before the Miami exclusion, this view surfaced two Miami schools with `school_level` null in `int_students__calendar_week` — PowerSchool school ids 30200801 (KIPP Sunrise Academy) and 30200802 (KIPP Liberty Academy), across all 43 of their AY2026 week rows. Both are real schools with `grade_band` populated on the locations sheet, ES and MS respectively, so the gap is in the Focus branch of the calendar model rather than in the locations data. Five of Miami's seven schools resolve correctly. The template model never surfaced this because its `u_expectations` inner join excluded Miami entirely. Excluding Miami here removes it from this view's output but does not fix it, and it will affect anything else that groups Miami calendar weeks by school level. Probably deserves its own issue.

**Summer toggle.** The `academic_year` filter carries the `/* summer toggle: see skill */` marker so the `gradebook-audit` skill's rollover procedure catches it. Unlike the template, this model projects the real `academic_year` column from `int_students__calendar_week` rather than stamping `{{ var("current_academic_year") }}` as a literal, so there is one toggle point here instead of two. The template needs the literal only because `U_EXPECTATIONS` carries no year of its own.

**`claude-review` could not run on this PR. The API credential behind the workflow is out of credit — confirmed out of band, not a defect in this diff.** Two runs failed identically: 35002645256 on commit 6e351db3 (`is_error: true`, 434ms, `num_turns: 1`, `total_cost_usd: 0`, empty `modelUsage`) and 35003685658 on commit 8c63608a (same, 456ms). Both initialized fine, reporting `model: claude-sonnet-5`, then errored before any model call. Zero tokens billed either time, so nothing was ever reviewed. The action surfaces no error text in the step log for this failure mode, which is worth knowing next time it happens — the signature to look for is a sub-second `is_error: true` with an empty `modelUsage`.

Re-firing will keep failing until the credential is restored, so it is not worth another draft toggle. **This PR therefore has no automated review coverage.** A human reviewer should not read the red check as a finding, and should not assume the usual bot pass happened.

For whenever it is back: `claude-review` fires only on `opened` / `ready_for_review`, never on `synchronize`, so pushing to this branch will not re-run it. Re-fire with a draft toggle (`convertPullRequestToDraft` then `markPullRequestReadyForReview` via GraphQL; the REST `draft=true` PATCH silently no-ops).

**Process.** No issue was opened; the requester chose to skip both the issue and a brainstorming session. Scope was confirmed with them before building: drop the `u_expectations` join rather than making it a left join, current academic year only, new tab on the existing spreadsheet. The Miami and ES exclusions were added on their follow-up request, after the first commit had already shipped the unscoped version.

</details>


GabyRangelB:
@claude are you back?

cbini (approved):

PR #5328, opened 2026-09-15
feat(gradebook-audit): resolve week-grid school level via school_level_alt

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will make both week-grid models on the gradebook-audit upload spreadsheet treat Sumner as middle school, and make the all-weeks grain test fail a build instead of only warning.

Sumner is an elementary school, but its grade 5 and up is audited as middle school. Both models grouped on `int_students__calendar_week`'s per-school `school_level`, which classified Sumner as ES — so the ES filter dropped it and Sumner's calendar never reached the Camden MS week grid.

Both now resolve school level the way `rpt_tableau__gradebook_audit` already does: `school_level_alt` from `int_students__school_directory`, falling back to the per-school `school_level`. That reuses the Sumner reclassification the directory already computes rather than restating it, so if the rule ever covers another school these models follow automatically.

**Neither model's output changes today.** `rpt_gsheets__gradebook_audit_all_weeks` is 202 rows before and after; `rpt_gsheets__gradebook_audit_template` is 17 rows before and after. No row differs in either direction in either model. Sumner's calendar already matches the other Camden MS schools week for week, and the template drops every ES row at its `u_expectations` join anyway. The point is that a future divergence will be reflected instead of silently dropped, and that the two tabs on the spreadsheet now agree on how school level is derived.

Separately, the all-weeks grain test now has `severity: error`. It was inheriting the kipptaf project default of `warn`, which means the invariant the model depends on — one row per region, school level, and quarter week — would not have failed a build if it broke. For a sheet T&L reads to set assignment expectations, a silently wrong week grid is worse than a failed build.

## Reviewer Notes

- The `gradebook-audit` skill's summer-toggle rollover procedure is updated because this change moves and adds toggle points. It named them by CTE and count, and both counts changed. A stale rollover procedure breaks the toggle silently, which is the entire reason those markers exist.
- `school_level_alt` depends on grade level, so one school can map to two levels. The join deliberately returns both an ES and an MS row for Sumner and the ES one is filtered out. Sumner is the only school in the network where the two columns disagree.
- The all-weeks `severity: error` deliberately differs from the template's equivalent test, which stays at the project default. Not proposing we change the template's test here.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models. No new models.
- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application. Both models were already wired; the all-weeks one in #5324.
- [x] **Breaking change?** No. No column added, removed, or renamed, and both models' row sets are byte-identical to production.
- [x] If adding or modifying an external source, run `stage_external_sources` with `--target staging`. Not applicable.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

**Why a join rather than an inline CASE.** Writing `case when school_short_name = 'Sumner' and grade_level >= 5 then 'MS' else school_level end` directly in these models was considered and rejected: it would be the fourth and fifth copy of that rule (`int_students__school_directory` derives it, `int_tableau__gradebook_audit_teacher_scaffold` consumes it), and `int_students__calendar_week` carries no `grade_level`, so each model would have to join enrollments anyway to evaluate it. Joining the already-derived `school_level_alt` is the same logic with one definition site.

**The coalesce is precedent, not invention.** `int_tableau__gradebook_audit_teacher_scaffold.sql:155-158` uses `coalesce(sec.school_level_alt, tw.school_level) as school_level` for exactly this pairing. The fallback matters because `int_students__school_directory` is enrollment-driven: a school with no enrolled grade rows would otherwise resolve to NULL, and `NULL != 'ES'` is not true, so that school would silently vanish from the grid rather than fail loudly.

**Verification, both models.**

`dbt build --target dev --defer --favor-state --state <main repo prod manifest>` on each: PASS for model and grain test, no warnings. `trunk check --force --no-fix` clean on all four files after the fmt hook reflowed them.

Before/after against the production views, which still carry the old logic, using `except distinct` in both directions:

| model | prod rows | dev rows | only in prod | only in dev |
| `rpt_gsheets__gradebook_audit_all_weeks` | 202 | 202 | 0 | 0 |
| `rpt_gsheets__gradebook_audit_template` | 17 | 17 | 0 | 0 |

Which schools feed each region and level in the all-weeks model after the change, confirming Sumner moved and nothing else did:

| region | school_level | raw school_level | schoolids |
| --- | --- | --- | --- |
| Camden | HS | HS | 179904 |
| Camden | MS | **ES** | **179905 (Sumner)** |
| Camden | MS | MS | 179902, 179903 |
| Newark | HS | HS | 73253, 732511 |
| Newark | MS | MS | 73252, 73258, 732513, 732514, 133570965 |
| Paterson | MS | MS | 2 |

Sumner is the single row where the resolved level differs from the raw one.

Why the template's output is unchanged despite the same edit: its `u_expectations` inner join drops every ES row, because T&L never loads ES expectations. Confirmed against the prod view — the template has Camden HS/MS, Newark HS/MS and Paterson MS rows and no ES rows at all. So Sumner never surfaced there under either derivation, and Camden MS rows come from 179902 and 179903 regardless. The edit is alignment, not a fix, for that model.

**Skill changes.** `.claude/skills/gradebook-audit/SKILL.md` rollover procedure:

- Template went from 2 toggle occurrences to 3, and its calendar-year filter moved out of `term_weeks` into the new `week_school_levels` CTE. The new `school_levels` CTE carries the third. Called out explicitly because leaving `school_levels` un-toggled resolves prior-year weeks against current-year school levels, which is silent when wrong.
- The all-weeks model was absent from the procedure entirely — it postdates the skill — and is now listed with its 2 occurrences, plus added to the verification build command. It projects `academic_year` from the calendar rather than stamping a literal, so it has no third occurrence.
- Corrected a pre-existing reference to `int_powerschool__calendar_week` in the template step; the model reads `int_students__calendar_week`.
- Added the two other models on the upload-template exposure to the gsheets list, with a note that all of these publish as Connected Sheets tabs and that Dagster's exposure asset is a marker that writes nothing.

**Test severity.** Confirmed the project default at `src/dbt/kipptaf/dbt_project.yml:22-23` (`data_tests: +severity: warn`). `.claude/rules/dbt-yaml.md` permits `rpt_` tests to omit the override "where a warning is acceptable," so the original was not a convention violation. Verified the override resolves via `dbt ls --output-keys "name config"`, which reports `severity: error`. Sibling precedent in the same directory: `rpt_gsheets__nsc_enrollment_new.yml` and `rpt_gsheets__people__location_crosswalk.yml`.

**Earlier CI red on this branch, resolved.** An earlier revision touched only the properties yml, and per `.claude/rules/dbt-models.md` a yml `description` change does not mark a model `state:modified` — so `state:modified+` selected the test but deferred the model to a `zz_stg_*` relation that did not exist, because `Clone - Staging` had run on #5324's merge commit at 19:03 before Dagster materialized the new view at 19:13. Repaired with a `Clone - Staging (On-Demand)` run (70403224899793). Now that the `.sql` files are edited, CI builds the models in the PR schema and that failure mode does not apply. The underlying timing gap is discussed in the review thread; the recommendation there was a `pr-ci-review` skill note rather than a pipeline change, which is not included here.

</details>


GabyRangelB:
## Verdicts on the review

**Finding 1 — severity bump could newly fail CI: addressed, the test passes on real data.**

Ran it before pushing, from the worktree:

```bash
uv run dbt build --select rpt_gsheets__gradebook_audit_all_weeks \
  --project-dir <worktree>/src/dbt/kipptaf \
  --target dev --defer --favor-state \
  --state /workspaces/teamster/src/dbt/kipptaf/target/prod
```

`PASS=2, WARN=0, ERROR=0` — the model and the now-`error` test. `--favor-state` resolves `int_students__calendar_week` to prod, so the test ran against production calendar data, which is the population that matters. Also confirmed the override actually resolves rather than being silently ignored: `dbt ls --select rpt_gsheets__gradebook_audit_all_weeks,resource_type:test --output-keys "name config"` reports `severity: error`.

Supporting evidence that the grain holds: grouping the built view by region and school level returns exactly five groups (Camden MS/HS, Newark MS/HS, Paterson MS), each with all four quarters and no null school level. A `week_start_monday` divergence within a region would have shown up as a duplicate there.

**The `dbt Cloud` check on this PR is red, and it is not a grain violation.** The error is:

```
Not found: Table teamster-332318:zz_stg_kipptaf_extracts.rpt_gsheets__gradebook_audit_all_weeks
```

Sequence of events, from the run history:

1. #5324 merged at 19:02, introducing this model.
1. `Clone - Staging` ran at 19:03 (run 70403224897763) and **failed** on this very model: `Not found: Table teamster-332318:kipptaf_extracts.rpt_gsheets__gradebook_audit_all_weeks`. Dagster had not materialized the new view in prod yet, so there was nothing to clone.
1. Dagster created the prod view at **19:13:41** (confirmed via `INFORMATION_SCHEMA.TABLES.creation_time`).
1. This PR's CI ran at 19:18 (run 70403224899207) and failed, because `zz_stg_kipptaf_extracts` still had no copy.

The reason CI defers instead of building the model itself: this PR changes only the properties yml, and per `.claude/rules/dbt-models.md` a properties-yml `description` change does not mark a model `state:modified`. So `state:modified+` selected the **test** but not the model, and the test resolved to the missing staging relation.

The prod view now exists, so a fresh `Clone - Staging` will create the staging copy and this goes green. No code change needed — flagging it rather than silently re-running, since that job recreates shared `zz_stg_*` relations.

**Findings 2, 3 and 4 — no action needed, and thanks for the independent trace.** You re-derived the Sumner reclassification, the dashboard's use of `school_level_alt`, and the sibling template's plain `school_level` grouping from the source rather than taking the PR body's word for it, which is the right instinct. I also verified your sibling-precedent claim for finding 1: `rpt_gsheets__nsc_enrollment_new.yml` and `rpt_gsheets__people__location_crosswalk.yml` both do carry `severity: error`, so this is an existing pattern in the directory rather than a new one.

One correction to the framing in finding 2: the two data claims are not merely "stated as checked" — here is what was run. Joining `int_students__calendar_week` to `int_extracts__course_schedule_by_term` on schoolid and source project for academic year 2026, exactly one school disagrees between `school_level` and `school_level_alt`: 179905, Sumner, ES versus MS. And Camden's ES and MS week grids are identical at 41 weeks spanning 2026-08-17 to 2027-06-17, which is why the divergence changes no output today.

GabyRangelB:
@claude — verdicts on your review are in [the comment above](https://github.com/TEAMSchools/teamster/pull/5328#issuecomment-5686819411). Short version:

- **Finding 1 (severity bump risk): addressed.** The test passes against production calendar data — `PASS=2, WARN=0, ERROR=0` via a dev build with `--favor-state`, and `dbt ls` confirms the override resolves to `severity: error` rather than being silently ignored.
- **Findings 2, 3, 4:** no action needed. I verified your sibling-precedent claim and it holds.

Two things worth your read, since you flagged the CI risk and the red check is real but unrelated:

**The `dbt Cloud` failure is a missing staging relation, not a grain violation.** `Clone - Staging` ran at 19:03 on the merge commit for #5324 and failed on this model (`Not found: kipptaf_extracts.rpt_gsheets__gradebook_audit_all_weeks`) because Dagster had not yet materialized the new view in prod. Dagster created it at 19:13:41. This PR's CI then ran at 19:18 and failed resolving the test against the never-created `zz_stg_kipptaf_extracts` copy. CI deferred rather than building the model because a properties-yml change does not mark a model `state:modified`, so `state:modified+` selected the test but not its model. I am triggering `Clone - Staging (On-Demand)` and re-running CI.

**A question I would value your opinion on.** This looks like a repo-level trap rather than a one-off: any brand-new model hits it, because `Clone - Staging` fires on the merge commit before Dagster has materialized that model in prod, and the failure only becomes visible when some later PR touches the new model without marking it `state:modified`. #5324 happened to have an immediate follow-up, which is the only reason it surfaced. Do you agree that is the general shape, and is it worth an issue? If so, is the right fix ordering (clone after the Dagster materialization), tolerance in the clone job for a not-yet-existing prod relation, or something else? Push back if you think this is narrower than I am reading it.

GabyRangelB:
@claude — asking for a second look before merge, because your earlier pass does not cover what is now in this PR.

**Why re-review.** Your review passed on commit `2ba15a578`, which was yml-only: a `severity: error` bump and a description. The PR has since gained two SQL model changes, a skill rewrite, and a `main` merge — head is now `4750966b1`, four files. The green check next to the old review could easily read as coverage it does not have.

**What is new since you looked:**

1. `rpt_gsheets__gradebook_audit_all_weeks.sql` — resolves school level via `coalesce(school_level_alt, school_level)`, joining `int_students__school_directory`, so Sumner grade 5+ lands in Camden MS instead of being dropped by the ES filter.
1. `rpt_gsheets__gradebook_audit_template.sql` — the same change, for consistency. **This is the model that generates T&L's live upload CSV**, so it is the one I would most want a second pair of eyes on.
1. `.claude/skills/gradebook-audit/SKILL.md` — the summer-toggle rollover procedure, because both models' toggle-point counts and CTE names changed.

**What I verified, so you can attack it rather than repeat it.** Both models are byte-identical to production: `except distinct` in both directions gives 202/202/0/0 for all-weeks and 17/17/0/0 for the template. Grain tests pass at `severity: error`. CI run `70403224901829` on the current head is green with zero warnings.

**Specific things I would like challenged:**

- The `school_levels` CTE uses `select distinct` and deliberately returns two rows for Sumner (ES and MS), relying on the downstream `where school_level != 'ES'` to drop one. Is that a fan-out risk for any school other than Sumner, or if a school's grade range changes?
- The `coalesce` fallback to the per-school `school_level` exists because `int_students__school_directory` is enrollment-driven, so a school with no enrolled grade rows would resolve NULL and — since `NULL != 'ES'` is not true — vanish from the grid silently rather than fail. Is that the right failure mode, or should it be loud?
- I moved the template's year filter out of `term_weeks` into a new `week_school_levels` CTE. Check the skill's rollover procedure actually matches the code now; that procedure going stale breaks the summer toggle silently.

**One thing only you can check.** Run `70403224901080` on commit `dd8e12cc9` — the `main`-merge commit pushed to this branch before my last commit — **failed** at 19:46 UTC. I tried twice to read its error and this repo's output scanner redacted the response both times, so I genuinely do not know what it was. It is on a superseded commit and the current head is green, but nobody has established what it was, and I would rather not merge on the assumption it was unrelated. Your workflow has `actions: read` and the review workflow does not, so you are the one who can open it. Please say what failed and whether it bears on the current diff.

GabyRangelB:
## Verdicts — both review rounds, head `705be3782`

The re-review earned its keep. Three real defects, all mine, all fixed.

### Fixed

**Missing summer-toggle marker** (flagged independently by both rounds — review finding 1, re-review finding 3). Confirmed: `rpt_gsheets__gradebook_audit_all_weeks.sql` had one marker where the SKILL.md this branch ships claims two. Marker added. Counts now match the procedure exactly — verified by grep: `all_weeks` has 2 markers for 2 `current_academic_year` occurrences, `template` has 3 for 3. This was the right catch; it would have sat invisible until July and then silently reverted the exact bug this PR fixes.

**Stale yml claim** (re-review finding 4). `rpt_gsheets__gradebook_audit_all_weeks.yml` said the template "still groups on the plain per-school school_level" — true when written, false one commit later when the template was brought in line. Your read of the cause is exactly right: leftover from earlier sequencing in this branch that I did not revisit. Now states they agree.

**Template yml never documented the alt logic** (re-review finding 5). Correct, and it matters more than the severity suggests, since that is the model T&L builds the upload CSV from. Its `school_level` column now carries the same `school_level_alt`/Sumner description.

**DISTINCT annotation imprecise** (review finding 3, marked "no action needed"). Took it anyway — it was two lines and the annotation exists to tell a reviewer the `DISTINCT` is deliberate, so naming the whole select list instead of the determining key defeated the purpose. Both models now say the key is `(_dbt_source_project, academic_year, ps_schoolid)` and that `school_level_alt` varies within it for Sumner only.

### Declined, with reasons

**Shared `int_` model to remove the duplicated CTEs** (review finding 2). Agreed in principle, and your point that the marker drift is a *symptom* of the duplication is the strongest argument for it — but not in this PR. It means a new intermediate plus rewiring both reports, one of which feeds T&L's live CSV, in a change whose entire current value is being provably inert. Worth its own issue.

**A note next to the `TODO` in `int_students__school_directory.sql:76`** (re-review finding 1). Good instinct, declining on a repo-specific ground: `.claude/rules/dbt-models.md` warns that a doc-only inline SQL comment on a heavily-consumed intermediate still marks it `state:modified`, fanning CI's `state:modified+` rebuild across its whole descendant graph. That model is consumed widely. The convention is to put such a note in the properties `description` instead, which does not mark it modified — better as a separate change than smuggled into this one.

### One clarification, not a concession

On the NULL-fallback (re-review finding 2) — I think we agree and the phrasing crossed wires. You are right that a row never vanishes, and the reason is precisely the `coalesce`: `cw.school_level` comes from `int_students__calendar_week`, which is not enrollment-derived, so it is populated regardless. My note described what would happen *without* the fallback, as the justification for having it — not current behavior. So: fallback stays, silent-with-a-safe-fallback is the right call, and we concur on the mechanism.

### The failed run — my error, and still unresolved

You were right and I was wrong: `70403224901080` is a 14-digit **dbt Cloud** run ID, not an 11-digit GitHub Actions one. `actions: read` was never the key, so I sent you after it on a false premise. Apologies for the wasted cycles.

I could not read it either — `get_job_run_error` (twice) and `get_job_run_details` all came back redacted by this repo's output scanner. What I could establish indirectly:

- **No BigQuery job errored in its window.** `JOBS_BY_PROJECT` between 19:43 and 19:47 with `error_result is not null` returns zero rows. So it was not a SQL failure — most likely a data test returning rows, which is a *successful* query.
- **Its selection was wide.** That run and mine both followed the `main` merge, and the per-PR audit dataset shows my run alone materialising tests across `graduation_path_codes`, `clever`, `parentsquare`, `schoolmint_grow` and `kippfwd` — models this PR does not touch. That is the documented `state:modified+` fan-out after a branch merges `main`, and `pr-ci-review` explicitly warns to expect latent `severity: error` failures unrelated to the change in exactly that situation.
- **The same wide selection passed on the current head** eight minutes later, warning-free.

So: probably an unrelated latent test, self-resolved. But "probably" is the honest word — nobody has read it, and I would rather say so than assert it was unrelated. The run page is the place to settle it: https://rn998.us1.dbt.com/deploy/116510/projects/211862/runs/70403224901080/

### Re-verified after the fixes

`dbt build` on both models, `--target dev --defer --favor-state`: PASS=4, zero warnings, both grain tests green. Output still byte-identical to production — `all_weeks` 202/202 with 0 rows differing either way, `template` 17/17 with 0. `trunk check --force` clean.

Note `claude-review` fires only on `opened`/`ready_for_review`, so it will not re-run on this push; these fixes are unreviewed by the bot unless someone toggles draft again.

GabyRangelB:
@claude — verdicts on your re-review are in [the comment above](https://github.com/TEAMSchools/teamster/pull/5328#issuecomment-5687445917), and the fixes are pushed as `705be3782`. Since `claude-review` only fires on `opened`/`ready_for_review`, it has not seen that commit — so this is the ask: **confirm the three fixes actually landed correctly**, rather than taking my word for it.

**Fixed, all three yours:**

1. **The missing summer-toggle marker** — added to the `school_levels` filter in `rpt_gsheets__gradebook_audit_all_weeks.sql`. Please re-run your own grep. It should now show 2 markers against 2 `current_academic_year` occurrences in `all_weeks`, and 3 against 3 in `template`, matching what the SKILL.md in this branch claims. This was the catch of the day — it would have sat invisible until July and then silently undone the fix.
1. **The stale yml claim** — `all_weeks.yml` no longer says the template "still groups on the plain per-school school_level". Your diagnosis of the cause was exactly right: leftover from earlier sequencing in this branch.
1. **The template's undocumented `school_level`** — now carries the alt-logic description.

I also took your finding 3 from the first round (the `DISTINCT` annotation naming the whole select list rather than the determining key), which you had marked "no action needed". The annotation exists to tell a reviewer the `DISTINCT` is deliberate, so an imprecise one defeats its purpose.

**Declined two, with reasons** — push back if you disagree:

- **The shared `int_` model.** Agreed in principle, and your "the marker drift is a symptom of the duplication" argument is the strongest case for it. Not in a PR whose entire current value is being provably inert, and which touches T&L's live CSV model. Its own issue.
- **A note by the `TODO` in `int_students__school_directory.sql:76`.** Declining on a repo-specific ground you may not have had loaded: `.claude/rules/dbt-models.md` warns that a doc-only inline SQL comment on a heavily-consumed intermediate still marks it `state:modified` and fans CI's `state:modified+` rebuild across its whole descendant graph. The convention is a properties `description` instead, which does not mark it modified.

**One clarification, not a concession.** On the NULL-fallback: I think we agree and the phrasing crossed wires. You are right that a row never vanishes, and the `coalesce` is exactly why — `cw.school_level` is not enrollment-derived. My note described what would happen *without* the fallback, as the justification for having it, not current behavior.

**And a correction of mine.** You were right that `70403224901080` is a dbt Cloud run ID, not GitHub Actions — I sent you after it with `actions: read`, which was never the key. I could not read it either; all three dbt MCP tools came back redacted by this repo's output scanner. Indirect evidence: zero BigQuery jobs errored in its window, so it was a test failure rather than a SQL error; its selection was the wide post-`main`-merge fan-out that `pr-ci-review` warns produces unrelated latent failures; and the same selection passes on the current head. Probably unrelated — but unread, and I would rather label it that way than assert it.

Current state: all 7 checks green on `705be3782`, CI run `70403224904616` with zero warnings, both models still byte-identical to production (202/202 and 17/17, zero rows differing either way).

GabyRangelB:
Closing in favour of shipping this on the PowerSchool plugin migration branch, `GabyRangelB/feat/claude-ps-plugin-migration` (issue #5433).

The whole diff is ported, verbatim. Commit [`b2508629a`](https://github.com/TEAMSchools/teamster/commit/b2508629a). Same 5 files, same 153 insertions and 29 deletions. The 4 dbt files are byte-identical to this branch's versions; `.claude/skills/gradebook-audit/SKILL.md` merged cleanly with the 6 commits `main` landed on it since this branch forked, and the only lines it replaces are the rollover text this change rewrites — five changes to six, the all-weeks toggle points, and the `int_powerschool__calendar_week` correction. `trunk check --force` clean on all 5.

Two reasons to move it rather than land it here.

This branch forked at `8a41570c6` and `main` has moved 435 commits since, so it needed a merge and a fresh CI run regardless. Doing that on the migration branch costs one run instead of two.

More usefully, the migration puts the plugin source, both skills, the dbt models and the CI contract checks in one place. `rpt_gsheets__gradebook_audit_all_weeks` is about to matter more than it did when this PR was opened: the end-user skill cannot see the school calendar today, only the rows already loaded into `U_EXPECTATIONS`, and this model is what closes that. Reviewing the week grid next to the skill that will read it beats reviewing it alone.

Nothing is lost by closing. Reopen this if the migration stalls and the school-level fix needs to ship on its own.

PR #5331, opened 2026-09-15
fix(amplify): name the frozen mClass API sources after their real tables

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will bring the `kipptaf` code location back up in prod.

The location is down right now. #5322 added two new kipptaf models, `stg_amplify__mclass__api__pm_student_summary` and `stg_amplify__mclass__api__benchmark_student_summary`, which rename the frozen mClass API archive columns to match the SFTP feed. The archive tables were already exposed as dbt sources under those same two names, so the new models and the sources now claim the same Dagster asset keys. Dagster refuses to build a code location with duplicate asset keys, so the whole location fails to load.

This change names each source after the BigQuery table it actually reads: `stg_amplify__pm_student_summary` and `stg_amplify__benchmark_student_summary`. The names no longer collide, so the location loads again.

## Reviewer Notes

The sources moved, not the models. Downstream code already reads the new model names, and the archive is frozen, so its asset key has no materializations to protect.

Each source entry also loses its `identifier` line and its pinned `config.meta.dagster.asset_key`. Both were only needed while the source name differed from the real table name. The default key Dagster derives is `[kipptaf, amplify, <table name>]`, which is exactly what the pins used to spell out.

### General

### Dagster

- [x] Run `uv run pytest tests/test_dagster_definitions.py` for any modified code location
- [x] New integrations follow the [Library + Config pattern](https://teamschools.github.io/teamster/reference/adding-an-integration/) with the correct asset key format and IO manager

### dbt

- [x] Include (or update) an [exposure](https://teamschools.github.io/teamster/reference/dbt-conventions/#exposures) for all models consumed by a dashboard, analysis, or application
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** so the dbt Cloud CI job can find the table. See [Staging external sources](https://teamschools.github.io/teamster/guides/dbt-development/#staging-external-sources)

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed, from a Dagster "Code location down" alert in Slack.

Exact failure, reproduced locally before the fix and gone after:

```text
dagster._core.errors.DagsterInvalidDefinitionError: The following dbt resources
are configured with identical Dagster asset keys.

The following dbt resources have the asset key
`['kipptaf', 'amplify', 'stg_amplify__mclass__api__pm_student_summary']`:
  - `model.kipptaf.stg_amplify__mclass__api__pm_student_summary`
  - `source.kipptaf.amplify.stg_amplify__mclass__api__pm_student_summary`
```

It is raised inside `@dbt_assets` in `src/teamster/libraries/dbt/assets.py::build_dbt_assets`, so the location never finishes loading.

Why both sides produced one key. `CustomDagsterDbtTranslator.get_asset_key` returns `super().get_asset_key()` unchanged when the node pins `meta.dagster.asset_key`, and otherwise prefixes the code location. For a model, dagster-dbt's default is `[<config.schema>, <name>]`, and kipptaf's `dbt_project.yml` sets `amplify: +schema: amplify`, so the new models derive `[kipptaf, amplify, <model name>]`. The sources pinned that same 3-part list verbatim. Renaming the source to the real table name makes the default derivation, `[<source name>, <table name>]` plus the location prefix, produce `[kipptaf, amplify, stg_amplify__pm_student_summary]` — distinct, and identical to what a pin would say, which is why the pins are gone rather than rewritten.

The pins were not pointless. They came from the archive pattern in `src/dbt/kipptaf/CLAUDE.md`: disable the retired model, expose the table as a BQ-native source, and pin the key so the asset keeps its identity. That works only while no live model holds the name. #5322 reused both names, which is what broke it. The sibling case in the same tree, `source("amplify", "int_amplify__dds__data_farming_unpivot")`, is still safe because kipptaf disables the whole `amplify: dds:` tree.

Verification, all from the worktree:

- `dbt parse --no-partial-parse --target prod` — clean.
- `dagster definitions validate -m teamster.code_locations.kipptaf.definitions` against the shelved (unfixed) files — reproduces the error above verbatim.
- Same command with the fix — gets past the dbt assets and fails only on the known codespace Illuminate driver-name credential gap documented in `src/teamster/CLAUDE.md`.
- `pytest tests/test_dagster_definitions.py::test_definitions_kipptaf` — fails before, passes after (1 passed in 36.86s). That is the credentialed path, so it clears the Illuminate gap.
- `trunk check --force --no-fix` on the 3 changed files — no issues.

Not done, and deliberately: no GitHub issue (quick fix on a live outage, per `CLAUDE.local.md`), and no `stage_external_sources` run, since these are BQ-native sources and not external tables.

Checked boxes with nothing to do: this PR adds no models, no exposures, no contracted-column renames, and no external sources.

</details>


cbini:
## Review verdicts

**1. Asset-key pin dropped where repo precedent keeps it — disagree with the premise, no change.**

The finding rests on the other archive sources pinning a key that is redundant with the default. I checked all 4, and in every one the pin changes the key:

| Source entry | Default-derived key | Pinned key |
| --- | --- | --- |
| `google/sheets/sources-bigquery.yml:5` | `[kipptaf, google_sheets, stg_google_sheets__people__employee_numbers_archive]` | `[kipptaf, google, sheets, people, employee_numbers_archive]` |
| `google/sheets/sources-bigquery.yml:15` | `[kipptaf, google_sheets, stg_google_sheets__people__student_logins_archive]` | `[kipptaf, google, sheets, people, student_logins_archive]` |
| `google/sheets/sources-bigquery.yml:25` | `[kipptaf, google_sheets, stg_google_sheets__performance_management__leadership_development_metrics_archive]` | `[kipptaf, google, sheets, performance_management, leadership_development_metrics_archive]` |
| `performance_management/sources-bigquery.yml:7` and `:16` | `[kipptaf, performance_management, <table>]` | `[kipptaf, performance_management, archive, <table>]` |

The Google Sheets entries reshape a 3-segment default into the 5-segment ingestion key, and the performance management entries insert an `archive` segment. Each pin is load-bearing: drop it and the asset key moves.

So the precedent is not "archive sources always pin." It is "pin when the wanted key differs from the default." The mClass API entries are the only ones where the pin spelled out the default exactly — and only after this rename, which is what makes them removable. Re-adding a pin identical to the derivation would be config that cannot change behavior, and the next reader would have to diff it against `CustomDagsterDbtTranslator` to learn that.

On the translator-change risk: a change to the prefixing logic would move the key of every unpinned dbt node in all 5 code locations, so these 2 sources would not be a silent special case — they would move with everything else. That is the intended behavior for a source whose key is just its table.

**2–5. Informational, no action.** Agreed on all 4, including that the diff is already the lean form.

**On the verification gap.** Fair to flag. The before/after repro is `dagster definitions validate -m teamster.code_locations.kipptaf.definitions` in a worktree: with the change shelved it reproduces the prod `DagsterInvalidDefinitionError` verbatim, and with the change applied `pytest tests/test_dagster_definitions.py::test_definitions_kipptaf` passes in 36.86s. The prod code location coming back up after merge is the real confirmation.

Issue #4834, opened 2026-08-11
fix(kippmiami): cross-feed race leaves int_finalsite__enrollment_lifecycle stale, silently dropping new enrollments from the Focus delivery


cbini:
## What happened

On 2026-08-11, 9 newly-enrolled KIPP Miami students appeared in
`kippmiami_extracts.rpt_focus__demographics` but were absent from
`rpt_focus__student_enrollment` in both `kippmiami_extracts` and
`kipptaf_extracts`. The 12:45 ET Focus delivery shipped without them, and
enrollment ops hit it mid-import.

8 of the 9 were a real defect. The 9th (no `enrolled_date` in Finalsite yet) is
the documented deferral in `rpt_focus__student_enrollment` and is correct.

## Root cause

`int_finalsite__enrollment_lifecycle` joins two independently-ingested feeds:

- `stg_finalsite__contacts` — Finalsite **API** pull, scheduled 12:00 ET
- `stg_finalsite__status_report` — Finalsite **SFTP** drop via the couchdrop
  sensor, supplying `enrolled_date` and `assigned_school`

`kippmiami/CLAUDE.md` documents these as deliberately concurrent: they "feed
opposite sides of `int_finalsite__enrollment_lifecycle`". What the design did not
account for is that the dbt rebuild those inputs trigger is not one build. It is
several independent automation-condition runs, and they can interleave:

| time (ET)   | run        | asset                               |
| ----------- | ---------- | ----------------------------------- |
| 12:11:18    | `a4db12b6` | `int_finalsite__enrollment_lifecycle` |
| 12:11:26    | `8f74ab21` | `stg_finalsite__status_report`       |

The lifecycle run finished 8 seconds before the status report it depends on was
rebuilt. Its materialization metadata records
`input_event_pointer/stg_finalsite__status_report = 93677832558` — the 06:52 ET
version — proving it read the pre-SFTP-drop status report. So
`enrollment_start_date` came back null for all 8 students even though
`assigned_school` (from the same row of the same table) came through.

Then it did not self-heal. 75+ minutes later no kippmiami run had re-requested
the asset, despite `stg_finalsite__status_report` carrying a new data version.
The eager `any_deps_updated` trigger in `dbt_table_automation_condition` should
have caught this. Reading `core/automation_conditions.py`, the likely mechanism
is that `.since(newly_requested() | newly_updated())` reset the SINCE memory in
the same tick that set the trigger, swallowing it — that mechanism is inference;
[...]

PR #5332, opened 2026-09-15
fix(finalsite): gate the enrollment lifecycle dates to the current cycle

### LINKED ISSUE #5313: fix: stale Finalsite `enrolled_date` puts a withdrawn student on the Focus enrollment import extract
## What's happening

`kippmiami_extracts.rpt_focus__student_enrollment` is the import feed that creates new enrollments in Focus. It holds exactly 1 row today. Ops confirmed that the student in that row should not be enrolled at KIPP Miami at all. 

The student withdrew during the 2025-26 school year. Focus ended their `syear` 2025 enrollment on `2025-12-01` with `drop_code` `1520`, and every prior year on that student uses `1519`. Finalsite recorded a `mid_year_withdrawal_date` of `2026-01-21`. Neither signal reaches the extract.

Two separate gaps in `int_finalsite__enrollment_lifecycle` let the row through.

The first gap is `enrollment_start_date`. The model reads `enrolled_date` straight from `stg_finalsite__status_report` and never checks that the date belongs to the current enrollment cycle. Finalsite reuses one contact record across cycles, so this student still carries `enrolled_date` `2025-10-24` from their 2025-26 enrollment. Their Finalsite `status` is `assigned_school`, and Finalsite never set a 2026-27 `enrolled_date`. The model comment says pre-enrollment statuses wait until Finalsite sets `enrollment_start_date`. The stale date defeats that guard.

The second gap is `enrollment_end_date`. The model derives it only from the `withdrawal_last_attended_date` custom attribute, and ops left that attribute null on this record. The model never reads `mid_year_withdrawal_date` from `stg_finalsite__status_report`, so the withdrawal is invisible to the feed.

The dates prove the record is stale. `enrolled_date` is `2025-10-24`. `applicant_date`, `application_complete_date` and `assigned_school_date` are all `2026-08-16`. A student cannot enroll 10 months before they apply.

The scale is 1 row. Of the 1503 open desired-state rows for `syear` 2026, 1502 carry status `enrolled` with an `enrolled_date` later than `applicant_date`. The remaining 1 carries status `assigned_school` with an `enrolled_date` earlier than `applicant_date`. That row is the bug.

I verified every fact above against prod. I did not verify whether other regions hit the same pattern, because `int_finalsite__enrollment_lifecycle` currently unions kippmiami only.

## Steps to reproduce

1. Count the rows in `kippmiami_extracts.rpt_focus__student_enrollment`. It returns 1.
2. Run the pattern query in the "For Claude" fold-out. It returns the same student, with `status` `assigned_school`, `enrolled_date` `2025-10-24`, `applicant_date` `2026-08-16`, and `mid_year_withdrawal_date` `2026-01-21`.
3. Query `teamster-332318.kippmiami_focus.stg_focus__student_enrollment` for that student. The latest row is `syear` 2025, ended `2025-12-01`, and no `syear` 2026 row exists. The missing 2026 row is what clears the anti-join in the extract.

## Where

- **Code location / dbt project:** `kipptaf` and `kippmiami`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** none

<details>
<summary>For Claude</summary>

Root cause sits in `src/dbt/finalsite/models/api/intermediate/int_finalsite__enrollment_lifecycle.sql`. The `kipptaf` copy at `src/dbt/kipptaf/models/finalsite/intermediate/int_finalsite__enrollment_lifecycle.sql` is a `union_relations` wrapper and holds no logic.

`enrollment_end_date` already carries a contact-reuse guard (`withdrawal_last_attended_date >= sr.enrolled_date`). `enrollment_start_date` carries no equivalent guard. That asymmetry is the defect.

Two candidate fixes, not mutually exclusive:

1. Null out `enrollment_start_date` when it precedes the current cycle's `applicant_date`, mirroring the existing guard on `enrollment_end_date`. This alone suppresses the row, because the `where l.enrollment_start_date is not null` filter in `src/dbt/kipptaf/models/extracts/focus/rpt_focus__student_enrollment.sql` then drops it.
2. Fold `mid_year_withdrawal_date` and `summer_withdraw_date` from `stg_finalsite__status_report` into `enrollment_end_date`, gated the same way as `withdrawal_last_attended_date`. Treat this one carefully: 12 of the 1502 correct `enrolled` rows also carry a `mid_year_withdrawal_date` from a prior year, so an ungated read of that column suppresses 12 legitimate enrollments.

Reproduce query:

```sql
with sr as (
  select * except (rn)
  from (
    select
      *,
      row_number() over (
        partition by finalsite_enrollment_id order by _dagster_partition_key desc
      ) as rn
    from `teamster-332318.kippmiami_finalsite.stg_finalsite__status_report`
  )
  where rn = 1
)
select
  c.status,
  sr.applicant_date,
  sr.enrolled_date,
  sr.mid_year_withdrawal_date,
  e.syear,
  e.start_date
from `teamster-332318.kippmiami_extracts.rpt_focus__student_enrollment` as e
inner join `teamster-332318.kipptaf_finalsite.int_finalsite__contact_id_attributes` as ida
  on e.student_id = ida.focus_student_id_prefixed
inner join `teamster-332318.kippmiami_finalsite.stg_finalsite__contacts` as c
  on ida.finalsite_enrollment_id = c.finalsite_enrollment_id
inner join sr
  on c.finalsite_enrollment_id = sr.finalsite_enrollment_id
where sr.enrolled_date < sr.applicant_date
```

Fixed means `kippmiami_extracts.rpt_focus__student_enrollment` returns 0 rows, and a re-run of the cohort check still shows 1502 rows at status `enrolled` in the `syear` 2026 desired state. Add a unit test to `src/dbt/finalsite/models/api/intermediate/properties/` covering a reused contact whose `enrolled_date` precedes `applicant_date`.

</details>

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." keep a withdrawn student off the Focus enrollment import feed, and start reading two withdrawal dates the model never looked at.

Finalsite keeps one contact record per person and reuses it every year. That means `enrolled_date` can still hold last year's value when this year's enrollment has no date yet. `int_finalsite__enrollment_lifecycle` read that date straight through, so one student who withdrew during 2025-26 still looked enrolled for 2026-27 and landed on the import feed that creates new enrollments in Focus.

Two changes fix it. First, the model now ignores an enrolled date that falls before this cycle's applicant date, because a student cannot enroll before they apply. Second, the withdrawal date now comes from the earliest of three signals instead of one — the `withdrawal_last_attended_date` attribute the model already used, plus the status report's own `mid_year_withdrawal_date` and `summer_withdraw_date`. Ops left the first one blank on this student, so nothing marked the withdrawal.

Both dates are gated the same way. A date that belongs to a prior year on a reused contact is dropped instead of read as current.

Against production, the feed goes from 1 row to 0, and all 1,498 correctly enrolled 2026-27 students stay on it.

Closes #5313

## Reviewer Notes

- **104 contacts gain a withdrawal date.** Every one of them sits at status `mid_year_withdrawal` or `summer_withdraw` in Finalsite, so none is an active enrollment. They were invisible to the model before.
- **No enrolled student is suppressed.** 12 mid-year and 8 summer withdrawal dates sit on students who are currently enrolled, left over from an earlier year. The gate ignores all 20.
- **A second stale contact drops out of the model.** It is a prior-year withdrawal that the model currently labels a 2026-27 transfer-out. Nothing consumes it today.
- **The staleness test carries the same guard.** Without that, the test reports the date the model deliberately drops as a stale read, and the asset check fails on a row that is behaving exactly as designed.

### General

No Dagster changes.

- [x] Include a `[model_name].yml` properties file for all new models — no new models; the existing properties file is updated with the new column behavior and 4 new unit-test cases
- [x] Include (or update) an exposure for all models consumed by a dashboard, analysis, or application — no exposure change; the Focus feed exposure is unchanged
- [x] **Breaking change?** No columns renamed or removed. Values change only on rows carrying a prior cycle's date.
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — no external source change. This is a value-only edit to a package model, so no staging seed is needed.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The issue's diagnostic was re-run against prod before any code was written; 3 facts the issue body does not carry came out of that re-run and drove the design.

**1. The guard must be null-safe, and is.** 977 of the 1,499 correct `enrolled` 2026 rows carry a NULL `applicant_date`. `if(sr.enrolled_date < sr.applicant_date, cast(null as date), sr.enrolled_date)` leaves them untouched, because a NULL predicate falls to the else branch. Only 2 prod rows have `enrolled_date < applicant_date` across the whole model.

**2. The second flagged row.** The issue names 1 row; the guard fires on 2. The other is a `not_enrolling` contact with `enrolled_date` 2025-10-24, `applicant_date` 2026-08-30 and `enrollment_end_date` 2026-06-30 — both dates from the prior cycle, which the model currently presents as a 2026 `is_transfer_out`. With a null start date it fails the in-scope filter and leaves the model. Checked consumers: `rpt_google_directory__users_import` maps `is_transfer_out` to `suspended`, and neither flagged row reaches that extract today (`is_create` / `is_update` both null), so the suspension behavior does not move. `rpt_focus__addresses` / `_contacts` / `_demographics` use the model as an existence gate and are import-once anti-joins; both contacts already carry a Focus id.

**3. The end-date addition is bigger than the issue predicted, and safe.** The issue warned that 12 correct `enrolled` rows carry a prior-year `mid_year_withdrawal_date`. Measured: 104 rows gain an end date, not 12 — but the group-by shows all 104 at status `mid_year_withdrawal` (93) or `summer_withdraw` (11). They are new rows entering via the transfer-out branch, not suppressed enrollments. Among the active statuses (`accepted`, `enrollment_in_progress`, `assigned_school`, `enrolled`, `retained`), 0 rows gain an end date; the gate rejects 12 mid-year and 8 summer dates that predate their enrollment start. `enrolled` 2026 feed-eligible rows are 1,498 before and 1,498 after.

**Why `min()` over `unnest` with a `WHERE`, not `min()` then one gate.** Gate-then-min and min-then-gate differ when a contact carries both a stale and a current withdrawal date: min-then-gate takes the stale one, fails the gate, and loses the real withdrawal. The `where d >= enrollment_start_date` inside the array subquery filters first, then aggregates. It is also the null-start case for free — a null start means no row satisfies the predicate, so `min()` returns NULL and a stale enrolled date yields no enrollment to end. This is the blessed scalar-aggregate-over-`unnest` form in `.claude/rules/dbt-sql.md`.

**The test change is load-bearing, not cosmetic.** `int_finalsite__enrollment_lifecycle__status_report_not_stale` flags `enrollment_start_date is null and sr.enrolled_date is not null`, which is exactly the shape the fix produces. Simulated against prod: the test returns 1 row without the gate and 0 with it. The test declares `config: severity: warn` in `src/dbt/finalsite/tests/properties.yml`, deliberately — its own description says a delivery built from a stale read still beats no delivery — so the ungated version would not fail the build. What it would do is fail the asset check, which is the signal Ops acts on, for a row the model is handling exactly as designed. (An earlier revision of this section claimed the ungated test would resolve to `error` and block the delivery. That was wrong; the explicit `warn` is right there in the unmodified properties file.) The test's own header comment keeps its `#4834` reference; the new CTE carries `#5313`.

**Verification run.** Unit test passes (10 cases, 4 new: reused-contact stale date, mid-year withdrawal with no `withdrawal_last_attended_date`, prior-year mid-year date ignored, summer date winning as the earliest of two). `trunk check --force --no-fix` clean on all 3 files. The model and test were compiled with `--target prod` and their compiled SQL run against prod to produce the row counts above. No prod build was run.

**Not in this PR.** Nothing outstanding — both gaps the issue names are closed here.

</details>


cbini:
Verdicts on the review, one per finding.

**1. PR body's severity claim — valid, fixed.** Confirmed: `src/dbt/finalsite/tests/properties.yml:110-111` declares `config: severity: warn` on this test, and its description says why. The PR body now says the ungated test would not fail the build; what it would do is fail the asset check — the signal Ops acts on — for a row the model is handling exactly as designed. The correction is noted in the "For Claude" fold-out rather than silently overwritten.

**2. Stale `finalsite_status` bypassing the date gate — not reproduced, no change.** Three checks against prod:

- A null-date row cannot reach the feed. `rpt_focus__student_enrollment` filters `where l.enrollment_start_date is not null and l.assigned_school is not null`. Today 117 active-status rows carry a null `enrollment_start_date` under the new logic and none of them reaches the extract. The stated failure path — "re-enter the feed via this branch with null dates" — is closed by that filter, not by the new gate.
- The hypothesized class is empty. Exactly 1 row has an active status stamped with a prior `school_year_start` and is otherwise feed-shaped, and it is not a stale status: `enrollment_in_progress`, `school_year_start` 2025, `enrollment_start_date` 2026-02-03 — a real mid-year 2025-26 enrollment. The kippmiami wrapper's `where d.syear = {{ var("current_academic_year") }}` excludes it from the SFTP feed regardless.
- Stale statuses already have their own coverage. `stg_finalsite__contacts__no_stale_enrolled` tests exactly this condition — an `enrolled` contact stamped with a prior school year — and its description records it as a standing Ops worklist inside Finalsite. Gating `status` in this model would hide that worklist behind a silent filter instead of reporting it where someone can act on it.

**3. Guard duplicated between model and test — acknowledged, no change.** Deliberate, and the reason is the one the finding already identifies: the test recomputes `status_report_latest` independently for the same purpose. Its job is to detect drift between the model and the live status report, so it has to derive the comparison itself. A shared macro would couple the check to the thing it audits, and a bug in the macro would pass the test.

**4, 5, 6 — no action.** Confirmations, and the double `safe_cast` removal was indeed a side effect worth having.

On the tooling note: `trunk check --force --no-fix` ran clean on all 3 files locally, the unit test passes (10 cases), and the prod row counts in the fold-out come from running the `--target prod` compiled SQL of both the model and the test against the warehouse.

PR #5333, opened 2026-09-15
fix(powerschool): pin the student_course_grades_spine uniqueness test to warn

cbini:
## Summary & Motivation

When merged, this pull request will stop the PowerSchool build from failing in Newark, Camden, and Paterson.

A uniqueness test on `int_powerschool__student_course_grades_spine` is meant to warn, not fail. The note above it in the properties file says exactly that: it goes back to failing once Ops finishes the stored-grades cleanup tracked in #3915. But the test never declared `severity: warn`.

It warned anyway, by accident. The model used to live in `kipptaf`, where every test warns by default. #5321 moved the model into the `powerschool` package, which sets no default of its own, so dbt fell back to its own default of `error`. The test then failed the build in all three New Jersey districts.

This adds the one line the test was missing.

Refs #3915

## Reviewer Notes

The duplicate rows are the known upstream problem, not a new one. `stg_powerschool__storedgrades` really does hold duplicate rows per student, year, course, and store code, in every year the test flags. The term spine that feeds the model is clean, so the model is not creating the duplicates.

The matching test on the `kipptaf` wrapper model is missing the same line. It still resolves to warn through the kipptaf project default, so it is not failing, and I left it alone.

### General

No Dagster changes.

- [x] Include a `[model_name].yml` properties file for all new models — no new models; this edits an existing properties file
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — no model or column changes, so no exposure change
- [x] **Breaking change?** No. No columns renamed or removed, and no model SQL touched.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Diagnosed from Dagster run `3eb03abd-b648-4ef9-8723-1fb2cb7104c9` (kippcamden `__ASSET_JOB`, step `kippcamden_dbt_assets`), which failed with `Got 229 results, configured to fail if != 0` on `dbt_utils_unique_combination_of_columns_int_powerschool__student_course_grades_spine_studentid__yearid__quarter__course_number__category_name_code`.

Severity chain, the actual root cause. `src/dbt/powerschool/dbt_project.yml` declares no project-level `data_tests:` severity. `src/dbt/kippcamden/dbt_project.yml` scopes its `+severity: warn` under the `kippcamden:` key, so it does not reach package tests. The test declared no severity of its own, so it resolved to dbt's built-in default. Confirmed by reading the parsed manifest, not by inference: `dbt parse --no-partial-parse --project-dir src/dbt/kippcamden --target prod` on `main` gives `severity: ERROR` for that node. `src/dbt/kipptaf/dbt_project.yml` sets an unscoped `+severity: warn`, which is why the same test warned before #5321 moved the model.

Verification. Re-parsed `kippcamden` on this branch and diffed every test node's severity against the `main` manifest: the test node set is identical (184 tests both sides) and exactly one node changed, `error` to `warn`. Lowercase `warn` matches all 14 existing warn tests in that manifest, including `int_powerschool__ps_enrollment_all`, which is the same generic test with the same `config:`-sibling-of-`arguments:` shape.

Duplicate-row evidence, to check the TODO's claim rather than trust it. Duplicate groups on the tested key, per district: kippcamden 229, kippnewark 192, kipppaterson 98. All three assets read DEGRADED in Dagster with 1 failed check each. In kippcamden the groups fall in academic years 2015, 2017, 2018, 2019 (Q1 to Q4, at most 4 rows per group) and 2024, 2025 (Y1, at most 8 rows per group) — every one a prior year against `current_academic_year` 2026, which matches the TODO. `stg_powerschool__storedgrades` independently has duplicate groups on `(studentid, yearid, course_number, storecode)` in each of those years. The `term_spine` CTE has zero duplicates per `(schoolid, yearid, quarter)`, so the model is not the source of the fan-out.

One thing worth a follow-up, deliberately out of scope here. Prior-year rows read `stg_powerschool__storedgrades` twice — once through `y1_final_grades` and once through the prior-years branch of `quarter_grades` — and both join independently, so a 2x upstream duplicate becomes 4x in the spine, and 8x on Y1. That amplification is why 2x source duplicates show up as 4-row and 8-row groups. #3915 cleaning the source resolves it either way, so this PR does not touch the SQL.
</details>


cbini:
Verdict on the two items the review flagged as unverifiable from its sandbox. Both are now closed with evidence.

**1. CI status — confirmed green.** All 7 checks pass: `dbt Cloud`, `Trunk Check`, `Trunk Check Runner`, `CodeQL`, `Analyze (python)`, `Analyze (javascript)`, `Analyze (actions)`. `dbt Cloud` passing is the direct confirmation of the fix: the same test that failed the prod build now warns, and the CI build goes green. I have ticked those boxes in the self-review.

**2. Duplicate-row counts — confirmed, by warehouse query rather than a dbt build.** Duplicate groups on the tested key: kippcamden 229, kippnewark 192, kipppaterson 98. The 229 matches the `Got 229 results` in the failing Dagster run exactly. Upstream, `stg_powerschool__storedgrades` independently carries duplicate groups on `(studentid, yearid, course_number, storecode)` in every year the spine test flags, and the `term_spine` CTE has zero duplicates per `(schoolid, yearid, quarter)` — so the duplicates are upstream and pre-existing, not introduced by this model. The review is right that the fix does not depend on the exact counts, only on the duplicates being upstream.

One correction to the review's method, not its conclusion. It verified the severity-resolution chain by reading the three `dbt_project.yml` files statically and said so. That chain was also confirmed by an actual parse: `dbt parse --no-partial-parse --project-dir src/dbt/kippcamden --target prod` reports `severity: ERROR` for the test node on `main` and `warn` on this branch, with the test node set identical on both sides (184 nodes) and exactly one node changed. Same conclusion, stronger evidence.

The review's strongest claim also checks out verbatim, and I verified it rather than taking it on trust. `src/dbt/kipptaf/CLAUDE.md` lines 280-288 prescribe exactly this remediation for this exact upstream issue: "Downgrade the affected mart PK uniqueness test to `severity: warn` with a `TODO(#3915)` so it returns to error when source cleanup completes", and it forbids the alternative fix: "Do NOT add defensive dedupes (`qualify row_number() = 1` or `dbt_utils.deduplicate()`) for the residual fan-out."

No defects were raised, so there is nothing to implement.

One note for whoever merges, which the review did not cover. A properties-yml-only change does not fire `code_version_changed`, since `code_version` hashes the model's raw SQL. So the deploy alone will not rebuild the spine, and the three assets stay DEGRADED until their next data-change materialization. That happens routinely, because the model is a table and its upstreams change daily. To clear it immediately after merge, launch the three `int_powerschool__student_course_grades_spine` assets in Dagster.

PR #5334, opened 2026-09-15
fix(focus): warn instead of error on cross-table relationships tests

cbini:
## Summary & Motivation

When merged, this pull request will stop a Focus data load from failing the whole Miami dbt build when a teacher saves an assignment at the wrong moment.

The Focus loader copies each table separately. It does not read them all at one instant. Two related tables can be read half a minute apart, so a row a teacher writes in between arrives without its parent row. The parent arrives on the next load, about 15 minutes later.

That happened on 2026-09-15. Gradebook assignment 424551 was created at 21:13:36. The snapshot of the assignments table had already stopped at 21:13:05. The link row pointing at that assignment loaded with nothing to point at, and the foreign key test failed the run. The next load fixed it on its own, with no code change.

This change drops all 35 foreign key tests in the Focus package from error to warning. A real orphan still shows up, as a warning, and a real one stays across loads. Single-table tests that check for duplicates or missing values stay at error, because one table is read all at once.

## Reviewer Notes

The main call worth a second look is warning instead of error for every foreign key test in the package, not just the one that failed. All 35 read across two separately-copied tables, so all 35 can fail this same way. The one that failed first is simply the table teachers write to all day.

Retrying does not help here, which is why this needed a change at all. The failed run retried once and failed again in the same way, because a retry runs the same tests against the same data.

### General

### dbt

- [x] No new models, so no new properties files are needed.
- [x] No columns renamed or removed, so nothing downstream breaks.
- [x] No external sources touched, so no `stage_external_sources` run is needed.

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Diagnosed with `superpowers:systematic-debugging` from Dagster run `07a5e320-9a01-4322-8293-612d78579fd5` (retry 1 of root `4b72ebc4-60f7-4418-b658-7da3b087a4f1`), `kippmiami__automation_condition_sensor`, commit `41c0a7d0bd`.

**Failing test:** `relationships_stg_focus__gradebook_assignments_join_course_periods_assignment_id__assignment_id__ref_stg_focus__gradebook_assignments_`, `failed_row_count` 1.

**Evidence.** Orphan row `id` 453230, `assignment_id` 424551, `created_at` `2026-09-15T21:13:36.549Z`. At failure time `max(updated_at)` on `dagster_kippmiami_dlt_focus.gradebook_assignments` was `2026-09-15T21:13:05.410Z`, and both tables carried the same `_dlt_load_id` `1789506813.677`. The parent table's snapshot provably predates the child row's creation. Neither staging model filters rows — both are pure passthroughs of the dlt relation — so the orphan is in the raw load, not in dbt logic.

**Mechanism.** `src/teamster/libraries/dlt/focus/assets.py`: `_build_focus_resource` gives every table its own `@dlt.resource` with `write_disposition="replace"` and `parallelized=True`, and `_focus_table_items` opens a fresh engine per table. There is no shared transaction and no exported snapshot, so each table's read instant differs. `cursor_column` in the location's `focus.yaml` is only the sensor's probe signature (count plus max cursor) for change gating — it does not make the load incremental or consistent.

**Self-heal confirmed, not assumed.** Polled until the next load: `_dlt_load_id` `1789507786.189` brought assignment 424551 in and the orphan count went to 0 with no code change. That is what rules out a genuine Focus-side orphan.

**Not chronic.** Of the last 12 executions of this check, 10 passed; the 2 failures are the original run and its retry against identical data.

**Verification of this diff.** `dbt parse --no-partial-parse --target prod` on `kippmiami` for both `main` and this branch: 356 tests on each, none added, none removed, exactly 35 `config.severity` changes, all on `relationships` tests, all in `test.focus.*`, and no focus `relationships` test left above warn. Raw diff is 35 removed `severity: error` lines and 35 added `severity: warn` lines, nothing else. `trunk check --force --no-fix` clean on all 17 files.

**Alternative considered and not taken.** Making the extract snapshot-consistent (`pg_export_snapshot` plus `SET TRANSACTION SNAPSHOT` shared across resources) would remove the race at the source and let every foreign key test stay at error. That is a real change to the dlt source needing its own design pass and a live Focus connection to test, so it was deliberately deferred rather than bundled here.

**Convention recorded.** `src/dbt/focus/CLAUDE.md` now documents why these tests are warn, so a future reader does not "restore" them to error. It also states the signal that matters: an orphan still present under a later `_dlt_load_id` is a real defect.

</details>


cbini:
## Review verdicts

Verdict per finding. 7 confirmed, 1 partly incorrect with evidence below.

**1. Staging-severity rule doesn't apply here — CONFIRMED, and the new-precedent flag is fair.** Verified the claimed precedent on `origin/main`: `stg_focus__student_enrollment.yml` carries two explicit `severity: warn` tests and `stg_focus__students_join_people.yml` one, all `dbt_utils.unique_combination_of_columns`. Also verified the "first `relationships` test at warn repo-wide" claim: `grep -rn -A6 "relationships:" src/dbt/ --include='*.yml' | grep "severity: warn"` returns nothing outside `src/dbt/focus/`. So this PR does set a new precedent, which is the reason the rationale went into `src/dbt/focus/CLAUDE.md` rather than living only in this PR body.

**2. Pre-join FK rule doesn't apply — CONFIRMED.** `..._join_course_periods` is a Focus bridge table, not a dbt join. The staging model is `select <columns> from {{ source(...) }}` with no join and no filter.

**3. No altitude violation — CONFIRMED.** All 17 files are in the `focus` package.

**4. Scope and completeness — CONFIRMED.** Matches my `dbt parse` result exactly: 35 anchors, 35 edits, 35 severity changes.

**5. District override doesn't collide — CONFIRMED, and I checked it by parsing rather than reading.** `src/dbt/kippmiami/dbt_project.yml` `data_tests` keys are `['+store_failures', '+store_failures_as', 'kippmiami']`: no `focus` key and no unscoped `+severity`. The per-test `warn` resolves as declared.

**6. No persistent-vs-transient signal — PARTLY INCORRECT. The signal already exists; only the alerting is missing.**

The finding reasons from `store_failures_as: view` to "the failing rows are queryable," then concludes nothing distinguishes a repeat. The `view` part is what makes that conclusion wrong. Because the stored failure relation is a VIEW and not a table, it is not a frozen snapshot of the rows that failed — it re-evaluates against current staging every time it is read. A transient orphan drops out of the view by itself once the next dlt load lands the parent; a persistent one keeps returning rows indefinitely.

Verified on this exact incident. All 35 focus `relationships` failure relations exist in `kippmiami_dbt_test__audit` with `table_type = 'VIEW'`. The one for this test is `relationships_stg_focus__grade_bab1e7a5c476117beb8763e41cbfe778` (dbt truncates the 100-plus-character test name to 30 characters plus a hash, which is why a `like 'relationships_stg_focus__gradebook_assignments_join%'` lookup returns nothing). `select count(*)` against it right now returns **0** — it cleared itself when load `1789507786.189` brought assignment 424551 in, with no rebuild.

So the operational rule needs no new machinery: query that view: rows means a real defect, 0 means it already healed. What is genuinely absent is a scheduled check that reads these views and alerts on a non-zero count, which is a different and much smaller ask than the finding implies. Out of scope here; worth an issue if the team wants it.

**7. Ponytail pass — CONFIRMED.** No change.

**8. Could not run `trunk check` in CI — NOTED, and here is the evidence.** `/workspaces/teamster/.trunk/tools/trunk check --force --no-fix` over all 17 changed files reports `Checked 17 files / No issues`. One prettier reflow on `src/dbt/focus/CLAUDE.md` was caught and fixed before the commit; the pre-commit hook then re-checked all 17 and reported clean.

No code changes from this review.

Issue #3954, opened 2026-05-18
fix(powerschool): state_studentnumber collisions across distinct students


cbini:
## Summary

PowerSchool `state_studentnumber` values that are shared by multiple `student_number` records under suspicious circumstances. Two buckets:

- **Bucket A — Confirmed collisions (high confidence):** 2 `state_studentnumber` values each shared by records with different `dob` and different normalized name (one case also different `gender`). 4 records total, all KIPP Camden, all `enroll_status = 2`. State-issued IDs must be globally unique per student — these are PowerSchool data entry errors where a previously-assigned state ID was attached to a newly-enrolled distinct student.

- **Bucket B — Possible entry mistakes (lower confidence):** Additional `state_studentnumber` values shared by records with same `dob` and same `gender` but substantially different normalized name. Most such pairs are same-person name corrections (hyphenation, spelling, last-name change post-marriage/custody), but a subset shows name divergence large enough to warrant Ops verification — they may be either legitimate name changes OR distinct students mis-assigned to the same state ID. Two pairs in the current data are flagged for individual review.

## Background

Surfaced during investigation under [#3882](https://github.com/TEAMSchools/teamster/issues/3882) (Paterson Pearson state-ID resolution). Currently affects a small number of non-Paterson Pearson assessment rows that would mis-attribute under a state-ID-based identity-resolution fallback. The Pearson fix under design uses a date-window tiebreaker so these collisions don't silently corrupt the model output, but the underlying PowerSchool data is wrong and should be corrected at the source.

## Reproduce — Bucket A (confirmed collisions)

```sql
with ps as (
  select
    cast(state_studentnumber as string) as state_studentnumber,
    cast(student_number as string) as student_number,
    enroll_status,
    dob,
    gender,
    regexp_replace(
      regexp_replace(
        lower(coalesce(first_name, '') || coalesce(last_name, '')),
        r'_?dupe', ''
      ),
      r'[^a-z]', ''
    ) as name_normalized,
  from `teamster-332318.kipptaf_powerschool.stg_powerschool__students`
  where dcid >= 1
),
collisions as (
  select state_studentnumber
  from ps
  where state_studentnumber is not null
  group by state_studentnumber
  having count(distinct student_number) >= 2
     and count(distinct dob) > 1
     and count(distinct name_normalized) > 1
)
select
  ps.state_studentnumber,
  ps.student_number,
  ps.dob,
  ps.gender,
  ps.enroll_status,
from ps
[...]

PR #5343, opened 2026-09-16
docs(stat): document the STAT dashboard, fix the comps label mismatch, add a crosswalk matcher

GabyRangelB:
## Summary & Motivation

When merged, this pull request will fix a silent bug in the STAT comparison view, document the dashboard, and add a rules-based matcher for repairing student numbers on state assessment rows.

Separate commits, so they can be moved or cherry-picked one at a time.

**The comps bug.** The comparison sheet spells two demographic subgroups differently from the way our own results are labelled. The dashboard matches a comparison row to our own row on that exact text, so the match never happened. It did not show a gap — it showed a loss. No "Black Or African American" row and no "Non-Econ. Disadvantaged" row has ever shown us outperforming, in any year, for any region. A spelling that does match, like "Hispanic", shows us outperforming about half the time.

Newark, Camden and Paterson used the matching spelling through the 2023 year and switched at 2024, so their comparison worked and then quietly stopped. Miami never matched. The economic-disadvantage label has never matched anywhere.

Fixing it turns 239 rows from "we lost" to "we won". Nothing turns the other way, and no total, percentage or student count changes.

**The matcher.** A new `analyses/` query that proposes crosswalk rows from evidence rather than one at a time by hand. It requires an enrollment in the test's own year and district, checks grade where the test code determines it, and matches on state id, name and date of birth in four tiers. Validated against all 81 rows already in the sheet: it agrees with 66, declines on 15, and **disagrees with none**.

**The documentation.** A reference page and a matching skill, like the ones for FRESH, CARAT, gradebook audit and DIBELS.

**No change to the student crosswalk itself.** An earlier version of this PR renamed it and added a vendor column; that was dropped on request. The sheet is used exactly as it is, and Cambium corrections go into the existing tab — which already works with no code change, because `stg_cambium__njgpa` aliases `student_test_uuid` to `studenttestuuid` before the union. The rename and the rest are written up as deferred.

## Reviewer Notes

- **239 rows change on the Advanced Comps dashboard**, all in our favour. Worth a look before it surprises anyone in a meeting. No other STAT view moves. The exact breakdown by year and region is in the reference page, not in this thread, so it stays findable.
- **The matcher proposes; it does not decide.** Zero disagreements across 81 hand-entered rows is what makes it safe in that role. It declines rather than guesses when more than one student fits.
- **The docs describe failure modes, not vendor behaviours.** This is deliberate and is the main thing I would like checked. Today's outstanding rows split cleanly along vendor lines, but that is a coincidence of one Cambium administration, and writing it down as a vendor trait would mislead whoever triages the first Cambium wrong-value row.
- **One failure mode is invisible to every test.** A wrong identifier that is itself a valid `student_number` resolves to the wrong student and the join succeeds, so nothing fires. The Paterson rows in the ID-translation spec did exactly this — 432 of them FK'd into Newark students and passed.
- **The audit found 13 sheet rows that do not hold up** — 6 inert, 7 unexplained. Written up in the reference page rather than fixed here.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — no new models; the matcher is an `analyses/` query, which builds nothing
- [ ] Include (or update) an exposure — none needed; noted separately that this exposure is missing its `url`
- [x] **Breaking change?** No. No model renamed or removed, no contracted column changed. Two values within `comparison_demographic_subgroup` are rewritten; verified against the published workbook XML that no filter, group, set or calculated field pins either literal.
- [x] If adding or modifying an external source, run `stage_external_sources` — not applicable, no source changed

### Docs

- [x] New data model page under `models/` with an `mkdocs.yml` nav entry

## CI checks

- [x] **Trunk** — passes locally on every changed file, `--force --no-fix`, exit 0
- [ ] **dbt Cloud** — expected to pass; nothing depends on an unshipped sheet change
- [ ] **Dagster Cloud** — not expected to trigger

<details>
<summary>For Claude</summary>

### Comps label mismatch

`rpt_tableau__state_assessments_dashboard_comps` self-joins each row to its `comparison_entity = 'Region'` partner on ten columns including `comparison_demographic_subgroup`. Sheet-sourced entities (City, State, Neighborhood Schools) and derived entities (Region, KTAF NJ, KTAF FL) disagreed on two values: `Black Or African American` vs `African American`, and `Non-Econ. Disadvantaged` vs `Non Economically Disadvantaged`. The join returns null and `if(null, true, false)` collapses to `false`, so a missing comparison is indistinguishable from a lost one.

337 and 671 rows respectively, both at 0 outperformed; `Hispanic` as a spelling-matched control reads 145 of 296. Grain is preserved exactly — 13,897 rows, 13,897 distinct grouping keys before and after — so nothing merges and the measures are byte-identical. The `accepted_values` list was verified against production: 15 distinct values post-normalization, none missing, none unused.

Tableau impact was checked by downloading the `.twb`; the MCP cannot read calculated-field text and 500s on `get-datasource-metadata` for embedded extracts. **Zero occurrences of either literal** in 2.1 MB of XML. A `<manual-sort>` dictionary on this field lists the derived vocabulary, so normalizing also fixes an axis sort. Only `Advanced Comps - 3-Column` uses the boolean columns.

### The matcher

Three criteria. The enrollment gate is hard for every tier — the crosswalk overrides only the identifier, and the downstream join still needs year and district, so a candidate without an enrollment that year is unmatchable from the sheet and is reported `no_match`, never `resolved`. Grade is a hard gate only where deterministic: codes ending `03`-`08` encode it, HS codes do not (Algebra I is sat in grade 8 or 9, pathway tests in 11), so there it is informational. Identity is state id, first, last and DOB across four tiers; Tier D (nickname) never auto-resolves alone.

DOB availability is uneven and **not** a vendor property: `stg_pearson__njsla` / `_njsla_science` / `_parcc` carry `birthdate` only because they `select * except (...)`, so it rides through unnamed, while `stg_cambium__njgpa` and `stg_pearson__njgpa` use explicit lists that omit it. Cambium runs on Tier B. Adding `birth_date` to the cambium package model would lift it to Tier A and is worth doing; Pearson NJGPA is deprecated.

Validation replayed all 81 crosswalk rows through the tiers using each row's raw pre-repair identifier: 66 agree, 2 ambiguous, 7 `no_pick_identity`, 6 `no_pick_not_enrolled`, 0 disagreements. Of the 12 rows generated and entered during this session, all 12 reproduce as `agrees`.

### Cambium field analysis

`student_test_uuid` is an exact analogue of Pearson's `studenttestuuid`: 850 of 850 raw rows distinct, zero blank. `oppkey` and `vndrtstevent_id` are equally unique. `student_assessment_identifier` is 425 distinct — per student per administration, not per test. `student_uuid` exists in the schema but is 100% blank, so Cambium provides no per-student key at all.

`local_student_identifier`: 14 null of 850 raw (9 of 813 in scope), zero non-numeric, only 5- and 6-digit widths, 804 of 804 non-null values resolving 1:1. `state_student_identifier` is 0% blank. **One administration; not generalized.** An earlier draft described Cambium as structurally clean and used that to argue the sheet was unnecessary for it; that inference was not supported and has been removed.

### Why the crosswalk rename was dropped

The Paterson ID-translation spec already names the endpoint: *"move the per-localid corrections to kippnewark / kippcamden intermediates analogous to the Paterson pattern; then the crosswalk can be retired."* Hardening a sheet scheduled for retirement is work in the wrong direction, and it is not needed — Cambium corrections already flow through the existing sheet untouched.

Measurements taken while investigating, recorded so they need not be redone: 69 crosswalk rows at the time, 25 distinct students, all matching a real test row; 59 override a different raw value, 4 fill a null, 6 are no-ops. Raw values span 4 to 9 digits correcting to 5 or 6 — scattered data-entry error, not one translatable id space, which is why this is the one-off kind of problem rather than the Paterson kind. 72,021 assessment rows, 72,021 distinct `studenttestuuid`, zero UUIDs under both vendors.

### Deferred, with evidence in the reference page

1. Resolve the absent mode from `statestudentidentifier` — all 9 rows resolved 1:1, zero fan-out. Reduces volume; does not replace the sheet, since present-but-wrong has no derivable rule.
2. Rename the crosswalk off the Pearson name. Orphans two relations in `kipptaf_google_sheets` that dbt will not drop.
3. Move per-localid corrections into district intermediates and retire the sheet, per the Paterson spec. No follow-up issue found open as of 2026-09-17.
4. The crosswalk staging model has no uniqueness test and no column descriptions.
5. 195 comparison rows find no Region partner after normalization; 573 of 870 `Neighborhood Schools` rows have none. Likely `school_level` / `grade_range_band`, which reach sheet rows as typed values but derived rows through `any_value()`.
6. `test_incorrect_student_number_pearson` inherits `severity: warn`. Left alone deliberately.
7. The 13 audit rows that do not reproduce: 6 inert, 7 unexplained.

### Verification performed

- `dbt parse --no-partial-parse` and `dbt compile` clean; the matcher compiles and executes against production.
- `trunk check --force --no-fix` exit 0 on every changed file.
- The 239-row figure is a simulation against production data, not a build — the staging model reads a Drive-backed external the BigQuery service account cannot query and that dev builds cannot stage without authorization. The mapping and counts are verified; the compiled SQL executing against the real sheet is not.

### Note on branch history

This branch received a `Merge branch 'main'` commit from another actor mid-session while local commits were being rewritten. The local history was replayed on top of that merge rather than force-pushed over it, so the merge is preserved and the final push was a fast-forward.

---
  - https://app.asana.com/0/0/1218554489792126


GabyRangelB:
**Superseded — this detail now lives in the reference page.**

The expected-changes breakdown for the comps label fix (which 239 rows flip, by year and region, and why a zero-flip cell is a genuine loss rather than a failed match) has moved into `docs/models/stat-dashboard-data-model.md`, under *Resolved — the subgroup vocabulary was split*.

A pull request thread is the wrong home for evidence that should still be checkable in a year. The page also carries the query to re-run after a rebuild.

GabyRangelB:

Worth the most scrutiny, in order:

**1. The percentage fallback in `rpt_tableau__state_assessments_dashboard_comps`.** `grouped_comps` re-derives `percent_proficient` from summed counts, which destroys the value when a source publishes a percentage with no denominator. The fix falls back to the reported percentage, guarded on `source_rows = 1`. The guard matters more than the fallback: one-row groups are a property of today's grain (measured: 13,897 of 13,897), not a guarantee, and without the guard a future grain change would silently average percentages unweighted. Please check the guard is actually sufficient for that, and that the new `grouped_comps_raw` CTE split did not change anything else.

**2. The label normalization in `stg_google_sheets__state_test_comparison_demographics`.** Two subgroup spellings are rewritten so the Region self-join matches. Verified 239 rows flip `region_outperformed` false→true and zero flip the other way, no aggregate changes, and the published workbook XML contains neither literal. The `accepted_values` guard is the load-bearing part — please sanity-check the 15-value list is complete rather than just currently-passing.

**3. `analyses/state_assessment_tiered_crosswalk_match.sql`.** Identity matching, so a wrong match attaches a score to the wrong student. Specifically: is the enrollment gate genuinely inescapable across all four tiers, and is Tier B (state id + both names, no date of birth) too weak? Context that matters — state ids are known unreliable, see #3954 — so Tier B leaning on one is the part I am least sure of.

**4. Anything in the two docs that overstates what was measured.** Several claims in earlier drafts were wrong and got corrected in later commits; if any survived, they will read as confident and be wrong.

Not in scope, deliberately, and documented as such: the crosswalk rename, the state-id fallback, the null-versus-false booleans, and 13 pre-existing crosswalk rows that do not reproduce.

GabyRangelB:
Verdicts on the review, one per finding. Both defects confirmed by reproduction before fixing, in `a8a82dc6d`.

## 1. Percentage fallback guard — agreed, no change

Correct that `source_rows = 1` plus `any_value()` makes unweighted averaging structurally impossible, and correct on the all-NULL multi-row case resolving to NULL rather than picking arbitrarily. Nothing to change.

## 2. Label normalization — agreed, with one correction

The `'Black Or African American'` versus `'Black or African American'` casing inconsistency is real and worth knowing about. The reasoning about `accepted_values` as a loud backstop is right.

One correction: the completeness check **was** run against production, before the MCP dropped. 15 distinct post-normalization values, zero missing from the list, zero unused entries. So the list is exhaustive against the live sheet as of 2026-09-16, not merely currently-passing. Re-running near merge is still sensible.

## 3a. Enrollment gate — agreed, no change

Confirmed inescapable via the `inner join` in `candidates`. Matches the header.

## 3b. Tier B — agreed, and it was worse than described

The point that `gap_dob is null` fires on *any* row lacking a parseable DOB, not only structurally DOB-less feeds, led somewhere bigger. Checking it: **15,625 Pearson rows had an unparseable birth date, which is exactly the PARCC row count.** PARCC ships `M/D/YYYY`; NJSLA and NJSLA Science ship `YYYY-MM-DD`. Parsing only the ISO form nulled *every* PARCC birth date and silently demoted every PARCC gap to Tier B.

Fixed by parsing both formats — all 67,078 rows now resolve, zero unparsed. The Tier B header comment claiming "Cambium only" was never true and has been rewritten to say it is the weakest auto-resolving tier, leaning on an identifier #3954 says is unreliable.

## 3c. Ambiguous rows vanishing — confirmed, real bug, fixed

Reproduced by loosening a tier to force ambiguity: 8 gaps in, 4 out across `resolved` and `no_match`, **4 lost** — exactly the ambiguous count. The trace was right: the anti-join tests existence in `picked`, and an ambiguous gap does have a row there via `any_value()`.

Fixed as suggested, with one addition: the `ambiguous` bucket **withholds** `proposed_student_number` rather than emitting the arbitrary `any_value()` pick, since naming one of several candidates would present a coin flip as a proposal. Conservation now verified — 8 flagged rows, 8 output rows, every gap in exactly one bucket.

Worth noting why this survived: the live gap set contains no ambiguity, so the analysis returned a clean-looking `8 → 8` every time I ran it. The "2 ambiguous" figure came from a separate audit-replay script. Exactly as diagnosed.

## 4. Doc claims — agreed, fixed

The "2 ambiguous rows are working as designed" line described the audit-replay procedure, not the analysis. Both are now distinguished, and the skill's bucket list gains `ambiguous`.

## Note on the limitation

The BigQuery disconnection was real, but the production figures were verified earlier in the session while it was up — the 239-row flip, the 15-value completeness, the 81-row audit breakdown. The two new numbers here (15,625 PARCC rows, the 4-row reproduction) were run against production just now.

GabyRangelB:
@claude both defects are fixed in `a8a82dc6d` — please re-review that commit.

**Ambiguous rows.** Reproduced your trace exactly by loosening a tier to force ambiguity: 8 gaps in, 4 out, 4 lost. Added the `ambiguous` bucket and dropped the outer `where s.n_candidates = 1`. One deviation from your suggestion: the bucket **withholds** `proposed_student_number` rather than emitting the `any_value()` pick, since naming one of several candidates presents a coin flip as a proposal. Conservation verified against production — 8 flagged, 8 output, every gap in exactly one bucket.

**Tier B / DOB.** Your point about `gap_dob is null` firing on malformed dates led somewhere larger: 15,625 Pearson rows had an unparseable birth date, exactly the PARCC row count. PARCC ships `M/D/YYYY`, NJSLA and NJSLA Science ship `YYYY-MM-DD`, and parsing only the ISO form silently demoted every PARCC gap to Tier B. Now parses both — 67,078 rows resolve, zero unparsed.

Specifically worth a second look:

1. **Is conservation actually guaranteed, or only true for today's data?** The claim is that arm one now emits every row in `scored` (one per gap with ≥1 tier match) and arm two emits every gap absent from `picked`, so the union is total. Please check there is no gap shape that falls outside both — that is the same class of mistake as the original bug.
2. **The `ambiguous` bucket withholds the candidate.** Does that break the batching procedure in SKILL.md, which hands `resolved` rows over as paste blocks? A null `proposed_student_number` should never reach a paste block, but confirm the procedure says so clearly enough.
3. **The two-format DOB parse.** `coalesce(safe.parse_date('%Y-%m-%d', x), safe.parse_date('%m/%d/%Y', x))` — is there an ambiguous string that both patterns accept and parse to *different* dates? I checked coverage (zero unparsed) but not collision.
4. **Anything the PARCC finding invalidates upstream.** If PARCC birth dates never parsed, any earlier reasoning of mine that assumed DOB coverage across Pearson feeds may be wrong, including in the reference page.

GabyRangelB:
@claude all four points addressed in `dea689f9f`. Your open question is now answered with data.

**1. Conservation.** Agreed with the key-domain proof, and took the warning seriously — added a comment at the top of the analysis naming the exact way a future edit breaks it: filtering `gaps_typed` rather than `candidates` without matching the anti-join. That is the bug this structure replaced, so it belongs at the site rather than only in a commit message.

**2. `ambiguous` presentation step.** Agreed, and the gap was real. Added step 7: presented separately as a table, never a paste block, no `proposed_student_number` on purpose, a person picks among the candidates and never defaults to one. Renumbered the audit step.

**3. DOB collision.** Agreed — different literal separators mean at most one pattern can match, so a collision is impossible by construction rather than only unobserved. Better argument than my coverage check.

**4a. Stale Tier B paragraph.** Correct, that one was missed. Fixed, and it now also says a malformed or unexpectedly-formatted value lands in Tier B silently, which is how every PARCC row got there.

**4b. Did the audit-replay share the parse bug? Yes — and it changed nothing.**

I wrote that script, so I could check: it used the same single-format `safe.parse_date('%Y-%m-%d', ...)`. Re-ran the 81-row audit with both formats.

| | before | after |
| --- | ---: | ---: |
| agrees | 66 | 66 |
| ambiguous | 2 | 2 |
| `no_pick_identity` | 7 | 7 |
| `no_pick_not_enrolled` | 6 | 6 |

**Outcomes identical.** What moved is only DOB availability: rows with a usable date of birth in the `no_pick_not_enrolled` bucket went from 4 to 6. The two that gained one are exactly the PARCC rows, and both already fail the enrollment gate, so the date was never reachable for them.

So your hypothesis was sound — a PARCC row that should resolve on Tier C would have shown as `no_pick_identity` — it just has no instances here, because the only PARCC rows in the 81 are unenrolled. **The 7 unexplained rows are not explained by the parse bug**, and the one confirmed NJSLA DOB mismatch remains the only characterised case.

Worth noting the shape of that near-miss: the bug was invisible in the audit for the same reason it was invisible in the analysis — the affected population happened to be filtered out before the broken field mattered. Two different code paths, same accident.

src/dbt/kipptaf/models/exposures/tableau.yml
```diff
@@ -707,6 +707,7 @@ exposures:
     depends_on:
       - ref("rpt_tableau__state_assessments_dashboard")
       - ref("rpt_tableau__state_assessments_dashboard_comps")
+    url: http://SAC-RPT-01/#/site/KIPPNJ/workbooks/7102
```
cbini: FYI this doesn't require a URL, just the ID. It's also not real

GabyRangelB: wtf

cbini (approved):

PR #5345, opened 2026-09-16
feat(cambium): add NJSLA and NJSLA Science assets ahead of the first file

### LINKED ISSUE #5342: feat: add Cambium NJSLA and NJSLA Science assets, built ahead of the file
## What and why

New Jersey moved NJGPA score reporting to Cambium TIDE for the Spring 2026 administration, and NJSLA and NJSLA Science follow the same path. Nothing in this repository ingests a Cambium NJSLA file today. When the first file lands on Couchdrop, no sensor matches it and no asset picks it up.

The file does not exist yet. I listed both regions' Couchdrop folders on 2026-09-16. `/data-team/kippnewark/cambium/` and `/data-team/kippcamden/cambium/` each hold a single subfolder, `njgpa/`, with one file in it. The newest NJSLA file anywhere is still the Pearson one, `pcspr25_NJ-807325_District_Summative_Record_File_Spring.csv`, dropped 2025-07-24.

So this work is a best guess on purpose, built ahead of the file so the plumbing is in place when it arrives. Every guess is recorded below and in the code.

## Proposed approach

Build the Dagster half only. That means 2 assets per region, `njsla` and `njsla_science`, a stub Pydantic model, the sensor entries, the code location definitions, the deploy workflow path filters, and tests.

Stop before dbt. The dbt models read a BigQuery external table over a Cloud Storage prefix that holds no files, so `stg_cambium__njsla` cannot build, and a `not_empty` test like the NJGPA one fails continuous integration from day 1.

Use a stub Pydantic model rather than a full guessed one. The asset check `avro_schema_valid` reports every CSV column the schema lacks, so the first real file names its own missing fields and we replace the stub from evidence.

Follow the NJGPA build in `src/teamster/code_locations/kippnewark/cambium/` exactly. The 2 shared lists that feed both the filename regex and the partition values are the load-bearing part, and the reason is written in that file.

<details>
<summary>For Claude</summary>

> Context, not instructions.

**Reference implementation.** `src/teamster/code_locations/{kippnewark,kippcamden}/cambium/assets.py`, `src/teamster/libraries/cambium/schema.py`, and the design doc at `docs/superpowers/specs/2026-08-18-cambium-njgpa-ingestion-design.md` (Refs #4899). The design doc names NJSLA as an explicit non-goal, and says the project is structured so a later feed is additive: a new Pydantic model, a new asset, a new source entry, a new staging model. This issue does the first 2.

**Verified facts, checked 2026-09-16.**

| Fact | Value |
| --- | --- |
| Cambium files on Couchdrop | `njgpa/` only, both regions |
| Newark Cambium NJGPA file | `2026_Spring_7325_District_Summative_Record_File_GPA.csv` |
| Camden Cambium NJGPA file | `2026_Spring_1799_District_Summative_Record_File_GPA.csv` |
| Cambium `NJGPA` Pydantic fields | 226 |
| Pearson `NJSLA` Pydantic fields | 198 |
| Pearson `NJGPA` Pydantic fields | 176, a strict subset of Pearson `NJSLA` |
| Pearson NJSLA-only fields | 22 |
| Of those 22, already present in Cambium `NJGPA` | 5 |
| `njsla` and `njsla_science` Pearson assets | both use one `NJSLA_SCHEMA` |

**Guesses, in descending order of risk.**

1. The filename. Unknown. The Cambium NJGPA name is `<year>_<season>_<district>_District_Summative_Record_File_GPA.csv`, so the NJSLA suffix token is the only unknown part, and NJSLA Science needs its own. Keep both regexes strict. A loose regex that also matches the GPA file makes `build_sftp_file_asset` raise on multiple matches, and a token that matches but is not a declared partition fails the whole sensor tick, which stalls every Couchdrop asset in the region.
2. The column set. A stub covers identifiers, scores, subject, and grade. `check_avro_schema_valid` in `src/teamster/core/asset_checks.py` computes `extras = record_fields - schema_fields` at `severity=WARN`, so the first materialization succeeds, passes the extras into check metadata, and silently drops them from the Avro output. That is the intended discovery path, not a bug to fix.
3. 2 assets rather than 1. Mirrors the Pearson split. Cambium may ship 1 combined file instead.

**Acceptance criteria.**

- [ ] `njsla` and `njsla_science` assets exist in both `kippnewark` and `kippcamden`, keyed `[<code_location>, cambium, njsla]` and `[<code_location>, cambium, njsla_science]`.
- [ ] Both regions' `couchdrop/sensors.py` list the new assets in `asset_selection`.
- [ ] `uv run dagster definitions validate` passes for both code locations.
- [ ] A test proves every filename the regex matches captures a declared partition key, matching `tests/libraries/test_cambium_assets.py`.
- [ ] A test proves the NJSLA regexes do not match the NJGPA filename, and the reverse.
- [ ] `src/teamster/libraries/cambium/**` stays in the `pull_request` path filters of both deploy workflows, so a branch deployment builds when the schema changes.

**Follow-up, once a real file lands.** Replace the stub from the `avro_schema_valid` extras list, then build the dbt half: `src_cambium__njsla` and `src_cambium__njsla_science` external tables, staging models in `src/dbt/cambium/`, the kipptaf alignment model, the `int_pearson__all_assessments` union, and the `dim_assessment_administrations` and `dim_assessments` CTEs.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will ingest Cambium's NJSLA and NJSLA Science files as soon as they land on Couchdrop, in both Newark and Camden.

New Jersey moved NJGPA score reporting to Cambium TIDE for the Spring 2026 administration, and NJSLA follows the same path. Cambium has sent no NJSLA file yet. I listed Couchdrop on 2026-09-16: both regions' `cambium/` trees hold only `njgpa/`, and the newest NJSLA file anywhere is still the Pearson one from 2025-07-24. This pull request builds the plumbing ahead of the file on purpose, so nothing has to be written in a hurry when it arrives.

There are 2 new assets per region, `njsla` and `njsla_science`, matching the Pearson split. They share 1 Avro schema, the way the 2 Pearson assets already do.

The `NJSLA` Pydantic model is a deliberate stub of 31 fields. Every field is copied from `NJGPA`, which was generated from a real Cambium file, so the spelling is right even though the set is incomplete. Nothing NJSLA-specific is guessed.

This pull request adds no dbt models. They would read a BigQuery external table over a Cloud Storage prefix that holds no files, so `stg_cambium__njsla` cannot build, and a `not_empty` test like the NJGPA one would fail CI from day 1.

2 follow-on commits came out of review:

- `a2d32b8257` points the review workflows at the renamed `dagster` plugin. `dagster-io/skills` renamed it from `dagster-expert`, and the old one is now a stub with no Dagster guidance — which is why `claude-review` reported that invoking it produced nothing.
- `530f281da6` moves the filename regex composer into `src/teamster/libraries/cambium/assets.py` instead of duplicating it in each code location.

Refs #5342

## Reviewer Notes

**The filename tail is the one real guess.** Everything before it is copied from the NJGPA file that did arrive. Only the token after `Record_File` is unknown, so it is an optional `(_\w+)?` group and the asset matches whether Cambium appends a subject token or nothing at all.

**Permissive is safe only because the directory scopes each asset.** `test_one_feeds_asset_never_matches_another_feeds_file` proves it for every ordered pair of feeds in each region, including the case where `cambium/njsla` is a prefix of `cambium/njsla_science`.

**The stub self-corrects, and that is the point.** `check_avro_schema_valid` reports the columns a file has and the schema lacks, at warn severity. The first real file materializes, passes, and names every column the stub is missing.

**`build_remote_file_regex` reads its alternations off the partitions definition it is handed.** So a filename the regex matches always captures a declared partition key, by construction rather than by editing 2 parallel lists together in each code location. That invariant matters because an undeclared partition key raises inside `resolve_run_requests`, which processes every run request for a tick in one pass — the whole tick fails and every one of that region's Couchdrop assets stalls until a redeploy.

**The refactor is behavior-preserving, asserted not assumed.** All 6 generated regexes are byte-identical to the ones the first commit produced, checked against the literal strings rather than re-derived.

**The generator sorts each alternation longest-first.** `Fall|FallBlock` matches `Fall` and leaves `Block` to fail against the next literal. Moot while `ADMINISTRATIONS` holds 1 value, so `test_a_prefix_token_does_not_shadow_a_longer_one` is the only thing exercising it.

**The local Claude settings file still names the old plugin** and cannot be edited by Claude. That 1 line is handed over separately and is not in this branch.

**The automations catalog stays stale, and this pull request does not fix it.** `docs/reference/automations.md` lists no `cambium` entries at all, so it has been stale since #4899. Regenerating it needs a full local checkout, because `gen-automations-doc.py` skips code locations that fail to import and both `kipptaf` and `kippmiami` fail in a Codespace. Running it there would drop them from the catalog.

### General

### Dagster

      code location

### dbt

Skipped, no dbt changes.

### Docs

      catalog: `uv run scripts/gen-automations-doc.py` — **not done, needs a full
      local checkout.** See Reviewer Notes.
      `CLAUDE.md` — not needed, `cambium` is already listed in both. `src/teamster/CLAUDE.md`
      is updated, since `cambium` is no longer a schema-only library.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Charlie asked to build this before the file exists, after I reported that Couchdrop has no Cambium NJSLA file. The decisions below were his calls, made with the evidence attached.

**Decision 1 — a stub model, not a full guessed one.** I measured 3 candidates before asking.

| Candidate | Fields | Evidence |
| --- | --- | --- |
| Cambium `NJGPA` | 226 | generated from a real Cambium file |
| Pearson `NJSLA` | 198 | wrong header vocabulary, see below |
| Pearson `NJGPA` | 176 | a strict subset of Pearson `NJSLA` |
| This stub | 31 | every field copied from Cambium `NJGPA` |

Copying Pearson was rejected on measurement, not taste. Pearson ships camel-case headers and Cambium ships spaced headers, and `slugify(separator="_")` does not split camel case, so `LastOrSurname` becomes `lastorsurname` where `Last or Surname` becomes `last_or_surname`. The NJGPA design doc measured 11 of 225 names overlapping. Nearly every Pearson field would have landed in the `avro_schema_valid` extras list and been dropped from the Avro.

Subclassing Cambium `NJGPA` was the option I recommended and Charlie declined: 209 fields reused verbatim, 17 invented. Those 17 are the Pearson NJSLA-only fields with no Cambium counterpart: `claimcode`, `formid`, `sex`, `testformat`, `languagecode`, `percentofitemsattempted`, `first_high_school_math_assessment`, 3 ELAL variants, and 5 science performance levels. 5 further Pearson NJSLA-only fields already exist in Cambium `NJGPA` after de-underscoring — `answers_recorded_in_test_booklet`, `home_language`, `mathematics_science_accommodated_response`, and the 2 unit-4 timestamps — and 4 of those 5 are in this stub.

The stub's 31 fields are exactly what `stg_cambium__njgpa.sql` reads from its source: the `summative_flag` / `test_attemptedness_flag` filter pair, the identifiers, the score columns, the demographics, and the 5 timestamps that resolve `test_date`. That is the demonstrated minimum for an NJ-assessment staging model, and every name in it came out of a real Cambium file.

**Decision 2 — Dagster only, no dbt.** `src/dbt/cambium/models/sources-external.yml` would need `src_cambium__njsla` as an AVRO external over an empty prefix. `stage_external_sources` fails creating an AVRO external over an empty Cloud Storage prefix, because autodetect needs at least 1 file.

**Decision 3 — 2 assets, not 1.** Mirrors the Pearson `njsla` / `njsla_science` split. Cambium may ship 1 combined file instead, in which case 1 of the 2 assets never fires and gets deleted.

**Decision 4 — the shared composer derives the regex FROM the partitions definition.** The alternative Charlie asked for was a plain shared function taking the same year and season lists the partitions definition takes. Deriving instead means the 2 can no longer disagree, which is the whole reason the original code carried a 12-line comment about keeping them in step. `build_partitions_def` and `build_remote_file_regex` are the only 2 public names; the asset definitions stay in the code locations and still call `build_sftp_file_asset` directly, so they remain greppable alongside every other SFTP integration in the repo.

`FIRST_ADMINISTRATION_YEAR = 2026` and `ADMINISTRATIONS = ["Spring"]` now live once in the library instead of twice in the code locations, along with the reasoning comments.

**Decision 5 — the `dagster-expert` plugin rename.** Per `dagster-io/skills`, the plugin is now `dagster@dagster`; the marketplace (`dagster-io/skills`) and the skill name (`dagster-expert`) are both unchanged, so only the plugin spec moves and the "Apply dagster-expert" prompt line stays correct. Upstream warns that installing the new plugin without removing the old leaves 2 plugins providing a skill named `dagster-expert`, and the deprecated stub may win — so this replaces the spec rather than adding alongside it.

**Verification run.** `uv run pytest tests/libraries/test_cambium_assets.py` — 31 passed. `uv run pytest tests/test_dagster_definitions.py -k "kippnewark or kippcamden"` — 2 passed, after `dagster-dbt project prepare-and-package` for both projects. I also asserted the 2 real Couchdrop paths still match `njgpa`'s composed regex, that all 6 regexes are byte-identical across the refactor, and that the generated Avro carries 226 NJGPA fields and 32 NJSLA fields, the extra 1 being `source_file_name` from `SFTPFile`.

`trunk check --force --no-fix` on the 2 workflow files and `src/teamster/CLAUDE.md` reports 2 `pinact/config-error` findings (`unknown flag: -format`). Those are pre-existing and environmental, not from this change — an untouched `deploy-prod-kippnewark.yaml` reports the identical error, and the commit hook reports no issues on the same files.

**Follow-up when a real file lands.** Read the `avro_schema_valid` check's `extras` metadata on the first materialization, replace the stub from that list, then build the dbt half: the 2 external tables, staging models in `src/dbt/cambium/`, the kipptaf alignment model, the `int_pearson__all_assessments` union, and the `dim_assessment_administrations` and `dim_assessments` CTEs. `src/dbt/cambium/CLAUDE.md` says NJSLA is still on Pearson, which stays true until that work ships.

</details>


cbini:
Per-finding verdicts on the `claude-review` comment above. No code changes.

**1. Test execution not possible in the review job — accurate, no change.** Both suites ran locally on this branch:

- `uv run pytest tests/libraries/test_cambium_assets.py` — 30 passed
- `uv run pytest tests/test_dagster_definitions.py -k "kippnewark or kippcamden"` — 2 passed, after `dagster-dbt project prepare-and-package` for both projects

I also asserted separately that both real Couchdrop paths still match `njgpa`'s composed regex after the constant hoist, and that the generated Avro carries 226 NJGPA fields and 32 NJSLA fields.

**2. The directory-prefix invariant is not asserted structurally — accurate, and I am declining the fix.** Three reasons.

The failure mode is fail-closed, not fail-open. A trailing `/` on `remote_dir_regex` composes to `dir//file`, which matches no path at all, so `build_sftp_file_asset` raises `FileNotFoundError` and logs the pattern it tried. It cannot silently pick up another feed's file, which is the failure a guard would exist to prevent.

A guard inside `build_sftp_file_asset` reaches 50 `remote_dir_regex=` call sites across every SFTP integration in this repo. That is a wide change for a risk that cannot produce a wrong match.

The case where the invariant does carry weight — a permissive filename tail, where only the directory separates 2 feeds — is asserted for all 12 ordered pairs by `test_one_feeds_asset_never_matches_another_feeds_file`, which is in this PR.

I grepped the convention to confirm it holds. Of those 50 call sites, the only 2 whose value ends in `/` are `remote_dir_regex=r"/"` in the amplify mClass sensors — a different builder and a root path, not a Couchdrop asset.

**3. dbt and SQL conventions not applicable — agreed, no change.**

**4. `dagster-expert` is a deprecated stub — agreed, no change here.** Worth noting for whoever maintains the review job's skill list, since the job will keep invoking it for nothing.

**5. Automations catalog stale since #4899 — confirmed, out of scope.** It needs `uv run scripts/gen-automations-doc.py` from a full local checkout, not a Codespace: the script skips code locations that fail to import, and `kipptaf` and `kippmiami` both fail there, so running it in a Codespace would drop them from the catalog. Left as a follow-up.

PR #5350, opened 2026-09-16
fix(focus): raise dlt step-pod memory limit to 3.0Gi

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will give the Focus dlt ingest op its own 3.0Gi memory limit, instead of the 2.5Gi that every step pod shares.

A GCP alert fired today because `kippmiami__dlt__focus` used 92.6% of that 2.5Gi limit, or about 2.31 GiB. The same op set off the same alert yesterday. Both runs finished successfully and nothing ran out of memory, so nothing is broken right now. What is left is about 190 MiB of room on an op that can use more.

How much memory this op needs depends on how many Focus tables a sensor tick picks up. Both alerts came from ticks that loaded 8 tables. dlt reads 5 tables at a time by default and holds a 50,000-row chunk of each one in memory, so a tick that picks up more large tables at once will use more. 3.0Gi is about 20% above the highest number we have measured, which is the same rule the cpu limit for these pods already follows.

## Reviewer Notes

The override sets memory and nothing else, on purpose. Dagster merges these settings key by key, so the cpu limit and both requests that all step pods share stay exactly as they are. A new test pins that, because a future Dagster version that merged one level shallower instead would make these pods quietly lose their cpu limit.

2.31 GiB is the floor of what this op used, not its peak. GCP measures memory once a minute and these pods live under two minutes, so most runs of this op are never measured at all.

The tests read the Focus code location as text instead of importing it. That file asks for `FOCUS_DB` credentials the moment it loads, and those are not available to tests.

The alert staying open for its full 30 minutes is expected, not a second bug. A step pod's measurements stop when the pod does, so GCP never sees a reading back under the threshold and waits out its timer instead. We decided to leave that alert policy alone.

### General

### Dagster

- [ ] Run `uv run dagster definitions validate` for any modified code location — **not run, cannot run here.** `kippmiami.definitions` fails at module load in the codespace because `dlt/focus/assets.py` resolves `FOCUS_DB` credentials eagerly and they are unset. That is the file this pull request edits, so there is no import path around it. Verified instead with `py_compile` on all three files, plus the tests below.
- [ ] Run `uv run pytest tests/test_dagster_definitions.py` for any modified code location — **not run.** It needs dbt manifests for all 5 locations in a fresh worktree, and its `kippmiami` case fails on the same `FOCUS_DB` resolution regardless of this change.
- [x] New integrations follow the Library + Config pattern — not applicable, no new integration. `op_tags` is the factory's existing documented parameter.

What did run:

- `uv run pytest tests/libraries/test_dlt_focus_memory_limit.py` — 3 passed.
- `uv run pytest tests/libraries/ -k dlt` — 76 passed, covering the widened factory signature.
- `trunk check --force --no-fix` on all three files — no issues.

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Charlie handed over GCP monitoring incident `0.ocpb8eqhrs50` and chose the fix from a set of options; Claude did the investigation and the edit.

Exact evidence chain:

- Incident `0.ocpb8eqhrs50`, policy `17738234194304579603` "GKE Container - High Memory Limit Utilization (all containers)", threshold `> 0.9` on `kubernetes.io/container/memory/limit_utilization`, `duration: 0s`, `trigger.count: 1`, `autoClose: 1800s`.
- Pod `dagster-step-ef5c30db0aa99d5580e2cbbcbdc0bb5c-pw2bj`, resolved to run `951e7f87-6a27-4b62-8d5e-4be2c34c4a7e` / op `kippmiami__dlt__focus` via a `pods.create` audit-log lookup on the label `dagster/op`. That run SUCCEEDED, 17:48:30Z to 17:50:07Z, loading 8 tables.
- The time series holds exactly 2 points, 17:50Z and 17:51Z, both `0.9255935668945312` for `memory_type: non-evictable`. 0.9256 × 2.5Gi ≈ 2.31 GiB. Pod events are `FailedScheduling` → `TriggeredScaleUp` → `Scheduled` → `Pulling` → `Pulled` → `Created` → `Started`, and nothing after. No `OOMKilling`, `Evicted`, or `Preempted` on this pod, and no `OOMKilling` in `dagster-cloud` in the 14 days to 2026-09-16.
- Yesterday's incident `0.ocnugnt7lw90` (2026-09-15T15:47:38Z) resolved to the same op by the same lookup. It closed in 60 seconds because that pod lived one sample longer; today's hit `autoClose` because its last sample was above threshold.
- Baseline is `.k8s/dagster/values-override.yaml` step-pod `containerConfig`: requests `cpu: 500m` / `memory: 2.0Gi`, limits `cpu: 1750m` / `memory: 2.5Gi`. Note 2.31 GiB already exceeds the 2.0Gi request, so these pods are burstable and node-pressure eviction candidates at peak.

Merge semantics were verified at runtime, not just read: `UserDefinedDagsterK8sConfig.__new__` defaults `merge_behavior=K8sConfigMergeBehavior.DEEP` (`dagster_k8s/job.py:132`), and a probe through `K8sContainerContext._merge_k8s_config` returned `{'requests': {'cpu': '500m', 'memory': '2.0Gi'}, 'limits': {'cpu': '1750m', 'memory': '3.0Gi'}}`. Under SHALLOW the whole `resources` dict is replaced and the cpu limit vanishes silently, which is why `test_memory_only_override_preserves_the_helm_cpu_limit` asserts the default explicitly.

`op_tags` is stored on the op as a JSON **string**, not a dict — `assets.op.tags["dagster-k8s/config"]` needs `json.loads`. Confirmed by probe before writing the test.

**Corrected after review (a38b9a1ca9).** An earlier revision of this PR also widened `build_focus_dlt_assets`' `op_tags` from `dict[str, object]` to `Mapping[str, Any]`, and this section asserted that the nested `dagster-k8s/config` literal was a genuine Pyright error without it. That was wrong. Pyright's bidirectional inference types a dict literal passed directly as an argument from the parameter's declared type, so the inline literal this call site uses checks fine. The diagnostic behind the original claim came from a probe that passed the dict through an intermediate variable first — the one case that does error. Settled with the pyright CLI across four cases: inline literal clean, pre-typed variable errors, `dict[str, Any]` return (the test helper's type) clean, concretely-typed nested return errors. The widening was load-bearing for neither the call site nor the test, so it is reverted and that file is back to byte-identical with base. This also keeps it consistent with leaving `build_powerschool_dlt_assets`' identical annotation alone, which was the right call on "no nested caller today, don't widen speculatively" grounds that apply equally here.

Alternative fix considered and rejected for now: capping dlt extract concurrency via the existing `dlt_extract_workers` run tag, or lowering `FOCUS_CHUNK_SIZE`. Either reduces peak memory instead of raising the ceiling, and is arguably the truer root-cause fix, but both cost ingest throughput on every tick and the tag is documented as a diagnostic knob. That is the lever to reach for if 3.0Gi also proves short.

Not done, deliberately: the alert policy still has no retest duration and no exclusion for `dagster-step-` pods, so any ephemeral step pod whose last memory sample is hot will keep parking an incident for 30 minutes. Charlie chose to leave it.

</details>


cbini:
Verdicts on the three findings, all addressed in a38b9a1ca9.

**Finding 1 — `# noqa: SLF001` wrong suppression syntax. Accepted, fixed differently than suggested.** The rule is real and I broke it. But converting it to `trunk-ignore(ruff/SLF001)` would have been wrong too: this repo pins no ruff config (no `[tool.ruff]` in `pyproject.toml`, no `.ruff.toml`), so SLF001 is not in the enabled set and nothing was ever flagged. That is why `trunk check` passed with the suppression in place. A `trunk-ignore` for an unflagged rule trips `trunk/ignore-does-nothing`, per the `trunk-lint` skill. Deleted the suppression outright instead. `trunk check --force --no-fix` still clean on all three files.

**Finding 2 — narrative duplicated in three places. Accepted.** Trimmed the call-site comment from 19 lines to 13. Kept only the two things genuinely non-obvious at that spot: why the override names memory and nothing else (the DEEP merge), and why a sampled peak is a floor that must not be used to argue the number down. The incident detail now lives in the test docstring only, with a pointer to it from the comment, so the figures have one home.

**Finding 3 — the pyright claim. You were right, and I verified it properly.** The widening is reverted; `src/teamster/libraries/dlt/focus/assets.py` is now byte-identical to base and has dropped out of the diff.

Your reasoning was exactly correct. Bidirectional inference types a dict literal passed directly as an argument from the parameter's declared type, so the inline literal this call site uses is fine against `dict[str, object]`. My original probe passed the dict through an intermediate variable first — the one case that *does* error — so I measured the wrong thing and then asserted the result in the commit message.

Ran the pyright CLI (`uv run --with pyright`) on four isolated cases to settle it:

| Case | Result |
|---|---|
| inline nested literal → `dict[str, object]` (the production shape) | clean |
| pre-typed variable → `dict[str, object]` (what my probe did) | **error** |
| `dict[str, Any]` return → `dict[str, object]` (the new test's helper) | clean |
| concretely-typed nested return → `dict[str, object]` | **error** |

So the annotation was load-bearing for neither the call site nor the test. Reverting also removes an inconsistency you had already spotted from the other side: I declined to widen `build_powerschool_dlt_assets` on "no caller passes it nested, don't widen speculatively" grounds, which applies identically here.

The PR body has been corrected — it asserted the type error as fact.

On your point 6, the widened request/limit gap: noted and still intentional. Worth restating that peak usage already exceeded the 2.0Gi *request* before this change, so these pods were burstable and eviction-eligible at peak already; this widens that window rather than opening it. Raising the request too would reserve 3.0Gi on every Focus tick, including the frequent ones that load a table or two, so I left it.

Re-verified after all three changes: `pytest tests/libraries -k dlt` 76 passed, new file 3 passed, `trunk check` clean.

cbini:
**Encoding probe — result recorded for `.claude/context/github.md`.** Body replaced with the finding; the original test tokens are no longer needed. (Kept rather than deleted: `gh api -X DELETE` is outside this repo's `gh`-via-Bash allowlist and no `mcp__github__*` tool deletes a comment.)

**Finding: the MCP read tools do NOT mangle bodies. The "read side mangles" note in `.claude/context/github.md` is wrong and has been corrected.**

This comment originally carried `<role>`, `<col>`, `<notatag>`, `<zz>`, `<br>`, `<strong>`, `<=`, `>=`, `&`, `&&`, double and single quotes — each in prose, in inline code spans, and inside a fenced block. Ground truth came from `gh api .../issues/comments/<id> --jq .body`; the comparison read was `mcp__github__issue_read` with `method=get_comments`.

Result: every token survived at identical count, and the MCP-read body was byte-identical to storage after JSON decoding (the only delta was one trailing newline that `--jq` adds on shell redirect).

| Token | Stored | MCP read |
|---|---|---|
| `<role>` | 3 | 3 |
| `<col>` | 2 | 2 |
| `<notatag>` | 2 | 2 |
| `<zz>` | 1 | 1 |
| `<br>` / `<strong>` | 1 / 1 | 1 / 1 |
| `<=` / `>=` | 2 / 2 | 2 / 2 |
| `&` / `&&` | 3 / 1 | 3 / 1 |
| `"` / `'` | 6 / 4 | 6 / 4 |

HTML entities introduced by the read: `&amp;` 0, `&#34;` 0, `&#39;` 0, `&lt;` 0, `&gt;` 0, `&lt;=` 0, `&gt;=` 0.

What the old note mistook for corruption is **JSON transport escaping**: the read returns `<role>`, `&`, `\"` — Go's `encoding/json` with `SetEscapeHTML(true)`, which is the default, and the GitHub MCP server is Go. Those escapes decode losslessly to `<role>`, `&`, `"`. Nothing is stripped and nothing is entity-encoded.

Practical consequence: round-tripping a body through the MCP read tools into a write would NOT have stored entities, so that specific hazard does not exist. Using a raw GET plus `gh api -X PATCH` to edit bodies is still fine, just not for the reason the note gave.

Cross-check against the earlier `<details>` / `<summary>` observation on this PR's own body: those survived the same path too, which is what prompted this probe.

PR #5354, opened 2026-09-16
chore(dagster): authenticate the official MCP server with our own token

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make Dagster's official MCP server authenticate with the same Dagster Cloud API token the homebrew server already uses, instead of OAuth.

OAuth cost us two things. Every person had to run an interactive authenticate step in each fresh Codespace before the server would answer a single call. And each person got their own personal Dagster+ permissions, so what the server could do varied by who was asking.

`.mcp.json` stays the one place the server is configured. A new `headersHelper` field points at `scripts/dagster-plus-mcp-headers.sh`. Claude Code runs that script when it connects, reads a JSON object of headers from its output, and sends them with every request.

The script does not hold the token or the 1Password reference. It sources `scripts/dagster-mcp-launch.sh` with a new `--no-exec` flag, which fetches the credentials and then stops short of starting the homebrew server.

## Reviewer Notes

The `--no-exec` flag is the only change to a working launcher. It guards the final `exec` line and nothing else, so the normal spawn path is byte-for-byte the same as before.

Sourcing the launcher rather than copying its fetch is deliberate. The launcher stays the only file that names the 1Password reference, so a rotation touches one line.

The token never lands anywhere durable. It lives in the helper's short-lived subprocess and nowhere else — not in the environment, not in a config file. That is the same property the homebrew launcher was built for, and it is why a plain `headers` block will not work here: this container does no startup secret injection, so `${VAR}` in a header would resolve to nothing.

Claude Code kills the helper after 10 seconds. It takes about 0.7 seconds.

### General

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Charlie asked for token auth on the official server rather than OAuth; the design and verification are Claude's.

Verified against the live hosted server, not just the argument schemas:

- the helper exits 0, writes a JSON object of string headers to stdout, and writes nothing to stderr
- those headers complete a real MCP handshake against `https://mcp.agent.dagster.cloud/mcp` — `serverInfo` is `Dagster Plus` 3.4.2 — and `tools/list` returns all 36 tools
- the calling process's own variables still contain no token after the helper runs
- the homebrew launcher spawns unchanged: `initialize` plus `tools/list` returns its 35 tools, on 3 runs out of 3

One earlier probe run reported no `tools/list` reply from the homebrew launcher. That was a race in the probe, not a regression: it wrote all three messages and closed stdin at once, and the server can exit on end-of-file before answering the third. Reading each reply before sending the next message made it stable. Worth knowing if anyone writes a similar stdio probe.

`${CLAUDE_PROJECT_DIR:-.}` in the `headersHelper` path is safe whichever way Claude Code resolves it. If Claude Code expands the variable itself, the path is absolute. If it passes the string to a shell, the variable is unset there and the `:-.` default applies, which resolves against the project root — Claude Code's own working directory. The existing `dagster` entry already uses the same form for its `command`.

Documentation went to three places, each per this repo's placement rules: the official server's auth behavior and its 401 diagnostic to `.claude/context/dagster-plus.md`, the "source the launcher, do not add a second 1Password reference" rule to `.claude/context/dagster.md`, and the catalog row plus the hosted-server `headersHelper` pattern to `scripts/CLAUDE.md`.

`headersHelper` is present in the installed Claude Code build (2.1.273). A first `grep` for it appeared to find nothing; that was a misread of grep's binary-file exit code, and `grep -a` finds 55 matches.

</details>


cbini:
Verdicts on the 3 flagged findings. Fixes are in 6cfccc49df.

**1. Live auth path unverified — accepted, still open.**

Half of this is already done. The bearer token plus organization header authenticate against the hosted endpoint: a real MCP handshake returns `Dagster Plus` 3.4.2 and `tools/list` returns all 36 tools. That ran live from this container, which is why the 0.7s timing is in the doc.

What is genuinely unverified is the other half, and the review named it correctly: Claude Code's own `headersHelper` plumbing. This branch is unmerged, so the running session still has the OAuth-only `.mcp.json`. Nothing here proves Claude Code invokes the script, parses its output, and sends those headers. That check needs a merge and a restart, so it is Charlie's to run: after restarting, `/mcp` should show `dagster-plus` connected with no authenticate prompt, and a tool call should return data rather than a 401.

**2. Comment overstates isolation — accepted, fixed.**

Correct, and the citation is right. `scripts/dagster-mcp-launch.sh:20` exports the token, so it is in the helper process's own variables and is inherited by `jq`. "Never in the environment" was wrong; only the persistent-shell claim held. The comment now says the token is exported there, inherited by `jq`, and gone in under a second, never reaching a persistent shell.

**3. Debug-log capture of the bearer token — declined, measured.**

Checkable, so I checked rather than leaving it as a nice-to-check. Claude Code redacts the header before it writes the log. Across every MCP server log directory in this container, the `github` server's log — the other one using bearer auth — holds 515 header mentions and **0** matching a bearer followed by a token-shaped string. The logged form is `"Authorization":"[REDACTED]"`. The only two matches under `dagster-plus` are an OAuth authorization URL, not a credential.

So the token does not persist to the debug log. That is now recorded in `scripts/CLAUDE.md`, since the next person to add a hosted MCP server will ask the same question.

**4 through 7 — agreed, no action.** The shfmt tabs, the `dirname "$0"` sourcing precedent against `source-path=SCRIPTDIR`, the positional-parameter restore after `source file arg`, and the ponytail pass all match what I found.

One note for the record: `claude-review` does not re-run on `synchronize`, so it will not re-review 6cfccc49df. That commit changes 2 comments and a doc paragraph, no logic.

PR #5356, opened 2026-09-16
fix(dbt): read Titan eligibility for directly certified students

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make `lunch_status` show Reduced for students who are directly certified at the reduced level.

Ops reported that some students on the Direct Certification list show Reduced in LINQ but Free in PowerSchool. The cause is in `int_students__student_enrollments`. For the current year in the New Jersey regions, `lunch_status` forced `F` whenever the direct-certification flag was set, ignoring the eligibility level LINQ actually sent.

That rule held until this year. Every direct-certification benefit type before AY2026 was a free-level type, so forcing `F` always matched. AY2026 added "Medicaid Reduced", which certifies a student at the reduced level. Those students got `F` anyway.

`lunch_status` now reads the eligibility LINQ sends, for directly certified students too. This value feeds the PowerSchool student feed, so it changes what the feed writes for the affected students.

## Reviewer Notes

The fix removes a rule rather than adding an exception. I checked that the old rule never did anything: in every year of LINQ direct-certification records, the eligibility level is always present and always Free or Reduced, so forcing `F` only ever restated a value that was already `F`. Exact counts are in the fold-out below.

This changes an outbound feed, not just a report. 64 students get a new value in the PowerSchool feed on the next run.

`lunch_application_status` still uses the direct-certification flag and does not change. These students are genuinely direct certified, so that label stays correct.

### General

- [ ] Include (or update) an
      summary above.
      See

No exposure change: this is a value fix on an existing column, and no column was added, renamed or removed. No external source changed.

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Reported in Slack by Ops with four example Newark students; diagnosed with `superpowers:systematic-debugging`.

**The change.** One expression in `int_students__student_enrollments.sql`, in the `lunch_status` CASE:

`then coalesce(if(tpd.is_directly_certified, 'F', null), tpd.eligibility_name)` becomes `then tpd.eligibility_name`

**Why the override was vestigial.** `stg_titan__person_data.eligibility_name` maps LINQ `eligibility` to `F` / `R` / `P`. Over `is_directly_certified = true` rows in `kipptaf_titan.stg_titan__person_data`, grouped by academic year, `eligibility_name` is null in 0 rows and outside `('F','R')` in 0 rows, for all of AY2020 through AY2026 (31,970 rows). `dc_reduced` counts are 0 for every year before AY2026 and 65 in AY2026. So the `if(...)` arm of the coalesce only ever fired on rows whose `eligibility_name` was already `F`, until "Medicaid Reduced" arrived.

This is the `src/dbt/kipptaf/CLAUDE.md` "Verifying a coalesce/override layer is vestigial" check done the right way round: the comparison is against the raw upstream `stg_titan__person_data.eligibility_name`, not against the post-coalesce `int_students__student_enrollments.lunch_status`.

**Prod verification.** Compiled with `dbt compile --target prod` (confirmed the compiled SQL no longer contains the override, and that `current_academic_year` resolves to 2026). Then compared old against new over prod for `academic_year = 2026`, `region != 'Miami'`, `rn_year = 1`, `lunch_status is not null` — 7,556 rows:

| old | new | directly certified | rows |
| --- | --- | ------------------ | ---- |
| F | R | true | 64 |
| F | F | true | 5,535 |
| F | F | false | 909 |
| P | P | false | 656 |
| R | R | false | 456 |

Only the 64 change, all `F` to `R`, all directly certified. LINQ has 65 such students in AY2026; the 65th has no AY2026 enrollment row at all, so it never reaches this CASE arm (`no_enrollment_row = 1`, `not_primary_stint = 0`, `primary_but_null_status = 0`).

No row flips to null. That needs `is_directly_certified = true` with a null `eligibility_name`, which the year-by-year counts above rule out.

**No unit test added.** `int_students__student_enrollments` is a poor unit-test target: `dbt_utils.union_relations` resolves at compile time against persisted relations, and the model carries 13 joined refs. A data test asserting `lunch_status` equals `eligibility_name` would restate the SQL rather than test it. The prod old-against-new comparison above is the verification instead. A regression here would surface as a Reduced student reading `F`, which is the reported symptom.

**Not changed, deliberately.** `lunch_application_status` (same model, next CASE) branches on `is_directly_certified` to emit `'Direct Certification'`. That is correct for these students and stays. `dim_student_meal_eligibility_status` reads `eligibility_name` straight from Titan with no direct-certification override, so it already reported Reduced correctly — it is the working reference that pointed at this line.

**Downstream.** `lunch_status` flows to `int_extracts__student_enrollments`, then to `rpt_powerschool__autocomm_students` (as `eligibility_name`), which each NJ region pushes to its own PowerSchool instance. The corrected values land after the next prod rebuild and autocomm push.

</details>


cbini:
## Review verdicts

**Finding 1 — PII tagging gap on `lunch_status`: accepted, fixed in cc94e21ea3.**

Checked before applying. This file carries 20 `contains_pii` tags and every one is a tier-1 direct identifier (names, `dob`, address, phone, email, external ids). No tier-3 education-content column in the model carries one. That is the gap `.claude/rules/ferpa-pii.md` describes in its own words — "the existing staging tags are narrower than tier 3" — followed by "widen them when you touch a file for another reason; do not sweep." This PR changes `lunch_status`, so `lunch_status` gets the tag.

Scoped to that one column on purpose. The other tier-3 columns the review lists (`lunch_application_status`, `lunch_balance`, `spedlep`, `lep_status`, `special_education_code`, `homeless_code`) are untouched by this PR, and the rule forbids the sweep. They go to whoever edits them next.

**Finding 5 — downstream impact, flagged "needs local verification": verified, conclusion holds.**

The review could only grep. I re-ran it with repo and warehouse access and reached the same answer by a different route. Three consumers branch on the value, and all three treat `F` and `R` identically:

- `rpt_gsheets__csgf_enrollment.sql:40` — `in ('F', 'f', 'R', 'FDC')`
- `rpt_gsheets__csgf_hs_enrollment.sql:245` — `in ('F', 'R', 'FDC')`
- `int_tableau__state_assessments_demographic_comps.sql:105,195,265` — `in ('F', 'R')`

Every other consumer projects the column without branching on it. No consumer filters `lunch_status = 'F'` exclusively, which is the case that would have silently dropped these students from an FRL population. So no aggregate moves.

The prod old-against-new comparison is already done, pre-merge, and is in the PR body fold-out: 64 of 7,556 current-year NJ rows change, all `F` to `R`, all directly certified.

**Findings 2, 3, 4, 6, 7: no action.** These report no violations found, and I agree with each.

## Unrelated CI failure, now re-triggered

The first `dbt Cloud` run on ec7458ba01 failed, and not on anything in this PR. All 3 failures were `Not found: Table ... was not found in location US` against `zz_stg_kipptaf_powerschool.base_powerschool__final_grades` and `zz_stg_kipptaf_overgrad.int_overgrad__students`, in `rpt_tableau__gradebook_es_comments`, `rpt_tableau__gradebook_gpa` and `int_kippadb__roster` — models this branch never touches.

A `Clone - Staging` run (70403225044558, 19:44:00 to 19:46:15 UTC) was recreating those `zz_stg_*` tables while the CI run read them (19:43:01 to 19:45:59 UTC). That is the documented overlap failure, not a regression.

Both that clone and the `Build - Staging (Modified)` run behind it finished before the push above, so cc94e21ea3 gets a clean build.

PR #5360, opened 2026-09-16
feat(kfwd): spell the long sheet's headers, add three GPA columns

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will finish spelling out the headers on the College Assessments Long sheet, and add three GPA columns KIPP Forward asked for.

The sheet carried two kinds of header. Twelve already read like English, such as `SAT Composite Superscore`. The other 29 still read like database columns, such as `student_number` and `grad_iep_exempt_status_overall`. All 41 now read like English.

The three new columns are unweighted cumulative Y1 GPA, unweighted projected cumulative Y1 GPA, and the weighted Y1 GPA for the year in progress. The first two already sat on `int_extracts__student_enrollments`. The third is new to this model and comes from PowerSchool's current reporting term, so it moves as grades post, unlike the cumulative columns.

Rolling this back is a plain revert of the one commit. That restores the old header names, and the three new columns disappear. Nothing outside the sheet reads this model, so nothing else has to be coordinated.

## Reviewer Notes

- This renames 29 columns on a contracted model, which is normally a breaking change. Here it is not. No other dbt model selects from this one, so the sheet is the only consumer, and Anthony confirmed the sheet has no formulas pointing at the old headers.
- The running Y1 GPA is the only column needing new plumbing, through a left join to `int_powerschool__gpa_term`. Worth a look at whether that join can duplicate rows. I checked that it cannot.
- Miami students will read blank in `Current Y1 GPA Weighted`, because Miami runs on Focus rather than PowerSchool and so has no GPA row to join to. The sheet already leaves Miami out for unrelated reasons.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — the existing properties file is updated, with the renamed columns and the three new ones
- [x] Include (or update) an [exposure](https://teamschools.github.io/teamster/reference/dbt-conventions/#exposures) for all models consumed by a dashboard, analysis, or application — the `rpt_gsheets__college_assessments_long` exposure already exists and needs no change
- [x] **Breaking change?** Renaming or removing columns in a contracted model (`stg_`, `rpt_`, mart) will break downstream exposures. Coordinate with affected teams before merging and document your rollback plan in the summary above — done, see the last paragraph of Summary
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — not applicable, no external source changed

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

<details>
<summary>For Claude</summary>

**Header convention.** Title Case with spaces, acronyms left uppercase, chosen over a literal underscore-to-space substitution so the new headers match the twelve that already read like English. `schoolid` gained a space it never had, as `School ID`, and `salesforce_id` became `Salesforce ID`. The three `*_highest_score_by_test` flags were reshaped onto the suffix pattern the benchmark columns already use, so they read `Highest Score by Test - Composite` / `- EBRW` / `- Math`.

**The uniqueness test moved with the rename.** `dbt_utils.unique_combination_of_columns` names `student_number`, `test_type`, `administration_type` and `test_date`, and all four are renamed, so the test now names them backticked. Missing this would have failed CI.

**Column sources.**

| New column | Source |
| --- | --- |
| `Cumulative Y1 GPA Unweighted` | `e.cumulative_y1_gpa_unweighted` |
| `Projected Cumulative Y1 GPA Unweighted` | `e.cumulative_y1_gpa_projected_unweighted` |
| `Current Y1 GPA Weighted` | `gt.gpa_y1`, via the new join |

**The new join.** `int_powerschool__gpa_term` on `studentid`, `yearid`, `schoolid` and `_dbt_source_project`, with `gt.is_current`. Same shape `int_gpa__goal_student_metrics` uses. `is_current` stays in the `ON` clause on purpose — in `WHERE` it would collapse the left join to an inner and drop every student without a PowerSchool GPA row, Miami included. There is an inline comment saying so at the join.

**Fan-out check.** `int_powerschool__gpa_term` filtered to `is_current` is exactly 1:1 on the four join columns: 59,882 rows against 59,882 distinct keys. Against this model's own population (AY2026, HS, `rn_year = 1`, `enroll_status = 0`, graduation year 2027 or later) the join returns 1,812 rows for 1,812 distinct students, so it adds no rows. 1,800 of the 1,812 carry a non-null `gpa_y1`; the 12 blanks are students with no current-term GPA row.

**Contract check.** 44 backticked aliases in the final `select` against 44 backticked names in the properties file, with an exact set match on both sides. `dbt compile --target prod` renders clean. `trunk check --force --no-fix` reports no issues on both files. sqlfluff does not read `region as \`Region\`` as a self-alias, so AL09 needed no suppression.

**Not done, deliberately.** This model carries student names, emails, `student_number` and now five GPA columns, and has no `config.meta.contains_pii` tag at all. Adding one would be a reasonable widening under the FERPA rule's touch-the-file clause, but it was outside the approved scope here and changes metadata other tools may read. Worth a follow-up decision rather than a silent add.

</details>


anthonygwalters:
## Review verdicts

One of the three findings is fixed, two are declined with reasons below.

### 1. `int_powerschool__gpa_term` grain not enforced by a test — verified, no change

The measurement the finding asked me to re-run is fresh, and it holds. Filtered to `is_current`, that model is exactly 1:1 on the four join columns: 59,882 rows against 59,882 distinct keys. Against this model's own population it returns 1,812 rows for 1,812 distinct students, so the join adds no rows. 1,800 of the 1,812 carry a non-null `gpa_y1`.

One correction to the citation. The finding points at `src/dbt/powerschool/models/sis/intermediate/properties/int_powerschool__gpa_term.yml`, but the `ref()` in this model resolves to the kipptaf model of the same name, at `src/dbt/kipptaf/models/powerschool/intermediate/properties/int_powerschool__gpa_term.yml`. The gap claim survives the correction — neither file has a `data_tests:` block — so the substance is right even though the file named is not the one this join reads.

That missing test is a real standing gap against the intermediate-model rule in `.claude/rules/dbt-models.md`, and it predates this PR. I am not adding it here: `int_powerschool__gpa_term` is a heavily consumed intermediate, so a test on it pulls its whole descendant graph into `state:modified+` and turns a two-file change into a wide CI run. Happy to file it separately if that's wanted. The finding's own closing point is also correct — `gpa_y1` sits in both `GROUP BY` lists, so a real fan-out would show up as duplicate rows and fail this model's `dbt_utils.unique_combination_of_columns` test rather than pass quietly.

### 2. No `config.meta.contains_pii` tag — fixed in ebd6489fd

Six identity columns now carry the tag: `Student Number`, `Salesforce ID`, `Student Name`, `Student First Name`, `Student Last Name`, `Student Email`.

I stopped at direct identifiers rather than going as wide as the FERPA rule's tier 3, which would also cover the five GPA columns, the six score columns, the benchmark and highest-score flags, and the two IEP columns. Two reasons. The sibling extracts in this directory set the narrower convention — `rpt_gsheets__kippfwd_miami_roster` and `rpt_gsheets__gradebook_audit_student_flags` both tag names and ids and leave grades untagged — and `.claude/rules/ferpa-pii.md` says to widen an existing narrow tag when touching a file but not to sweep. Going tier-3-wide here would have been the sweep. Scope was confirmed with the author before tagging, per that rule's one-line-confirmation clause. The wider reading is a deliberate decline, not an oversight.

### 3. 29-column rename wants model versioning — declined

There is no versioning convention here to follow. `latest_version:` appears in no dbt properties file anywhere in this repo, so adding a `v2` to this model would invent a pattern rather than follow one.

Worth recording that dbt agrees the rename is breaking, and says so in CI. The passing run logs one warning: `Breaking change to contracted, unversioned model rpt_gsheets__college_assessments_long ... Columns were removed`. That is expected from a rename, it is a warning and not an error, and it will stay in the log for this PR. Nobody should read it later as a new problem.

The part the finding correctly says it cannot check statically — whether the sheet has formulas on the old headers — was confirmed by the sheet's owner before any of this was written. That is the whole basis for treating the rename as safe, and it is not something CI can ever verify.

### CI

`dbt Cloud` passed on 2af89ade3, with the one expected warning above and no test failures. The contract itself is only enforced at build time, so that green run is what proves the 44-column contract matches the SQL.

PR #5361, opened 2026-09-16
refactor(dbt): move the illuminate_subject mapping into a vendor subject crosswalk sheet

### LINKED ISSUE #5340: refactor(dbt): move the illuminate_subject mapping out of source models into the blended assessment layer
## What and why

`illuminate_subject` maps a vendor's or state's own subject name onto the Illuminate subject vocabulary. That mapping only matters where assessment sources are blended together, but today each source model defines its own copy. Six models carry one:

| Model | Project |
| --- | --- |
| `int_iready__diagnostic_results` | kipptaf |
| `int_amplify__all_assessments` | kipptaf |
| `stg_renlearn__star` | kipptaf |
| `int_pearson__all_assessments` | pearson, kipptaf |
| `int_fldoe__all_assessments` | kippmiami, kipptaf |
| `stg_cambium__njgpa` | cambium |

Two problems follow. A source model has to know about Illuminate, which it otherwise does not, and the vocabulary lives in six places, so a new subject means six edits and six chances to disagree.

Only two models consume the column: `int_assessments__score_anchors` reads it as `subject_area`, and `fct_assessment_scores_enrollment_scoped` reads it to join `int_assessments__resolved_section_enrollments`. Both sit in the blended layer. That is where the mapping belongs.

## Proposed approach

Define the mapping once in the blended assessment layer and drop the six per-source copies. Each source model keeps its own raw subject column; the blended layer translates it.

Check two things before starting. Some of the six are package models consumed by kipptaf through `source()`, so removing a column from one is a two-PR change per `.claude/rules/dbt-models.md`. And `stg_cambium__njgpa` is contract-enforced, so its properties file changes with it.

<details>
<summary>For Claude</summary>

> Context, not instructions.

Raised by @cbini as an inline review comment on PR #4710, on the new `int_iready__domain_unpivot` model: "pre-existing, but `int_iready__diagnostic_results` is not the right place to define this. it belongs downstream in blended models." Flagged there because that PR adds a seventh model that selects the column through, but the pattern predates it. Not fixed in #4710 — a six-model change across four dbt projects exceeds the additive-upstream-edits-only rule in `src/dbt/kipptaf/models/marts/CLAUDE.md`.

The i-Ready definition is `int_iready__diagnostic_results.sql:153-155`, a two-arm CASE: `Reading` → `Text Study`, `Math` → `Mathematics`.

Full reference list: `rg 'illuminate_subject' -g '*.sql' src/dbt`. Note the distinct `illuminate_subject_area` column on the course-enrollment side (`int_students__course_enrollments`, `int_assessments__course_enrollments`) — that is the join target, not a copy of this mapping, and it is out of scope.

Acceptance: one definition of the vendor-to-Illuminate subject mapping in the blended layer; no `illuminate_subject` column on any source-system staging or intermediate model; `fct_assessment_scores_enrollment_scoped` row count unchanged against the pre-change baseline.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will move the vendor-to-Illuminate subject mapping out of six source models and into one Google Sheet tab that analysts can edit.

`illuminate_subject` translates a vendor's or a state's own subject name into the Illuminate subject vocabulary. Six source models each defined their own copy, so a source model had to know about Illuminate, and a new subject meant six code edits. After this change, a new subject is one sheet row.

The new tab is `src_assessments__vendor_subject_crosswalk` on the existing assessments spreadsheet, and it holds 18 rows over 3 columns: `Source_System`, `Raw_Subject`, `Illuminate_Subject_Area`.

This is the first of 2 pull requests. A column drop reverses the usual order: the consumer stops reading before the producer removes. This one covers `kipptaf` only. A second pull request drops `illuminate_subject` from `pearson`, `kippmiami` and `cambium` once this merges and Dagster materializes it in production. The 3 models that still derive the column are correct at this point, not an oversight.

Row counts are unchanged. I verified each converted model against production, and the detail is in the "For Claude" fold-out below.

## Reviewer Notes

**The mapping resolves once per union, not once per source.** Every consumer already funnels its per-source branches into a single union CTE that carries a source discriminator, so the crosswalk joins at 5 sites, not 15.

**An unmapped subject falls back to its raw value.** Pearson, FLDOE and Cambium used to pass an unmapped subject through unchanged. A bare `LEFT JOIN` would yield NULL instead, and NULL never matches `illuminate_subject_area` at the resolver join, so those rows would drop silently. Every join site therefore reads `coalesce(x.illuminate_subject_area, <alias>.raw_subject)`. A warn-severity test lists any pair the crosswalk misses.

**The issue named 2 consumers. There are 4.** `int_extracts__student_enrollments_subjects` and `rpt_tableau__academic_goals_rollup` also read the column. Both are converted here.

**One deliberate behavior change.** Ren Learn's old `else 'Text Study'` meant a new Star product code was silently called Text Study. Under the crosswalk it passes through as its raw code and trips the warn test instead. `SEL` keeps mapping to Text Study, because it gets an explicit row.

**I already staged the new external source.** `src_google_sheets__assessments__vendor_subject_crosswalk` now exists in `zz_stg_kipptaf_google_sheets`, so dbt Cloud CI can read it.

### General

No Dagster changes.

- [x] Include (or update) an
      summary above.
      See

No schedule, sensor or integration changes. One published page is corrected: `docs/models/dibels-dashboard-data-model.md` counted `illuminate_subject` among the columns `int_amplify__all_assessments` adds.

## CI checks

## Troubleshooting

Refs #5340

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Design spec: `docs/superpowers/specs/2026-09-16-illuminate-subject-crosswalk-design.md`. Implementation plan: `docs/superpowers/plans/2026-09-16-illuminate-subject-crosswalk.md`.

### The 5 join sites

| Consumer | CTE | File |
| --- | --- | --- |
| `int_assessments__score_anchors` | `scores_resolved` | `models/assessments/intermediate/` |
| `fct_assessment_scores_enrollment_scoped` | `state_union` | `models/marts/facts/` |
| `fct_assessment_scores_enrollment_scoped` | `vendor_resolved` | `models/marts/facts/` |
| `int_extracts__student_enrollments_subjects` | `prev_yr_state_test_resolved` | `models/students/intermediate/` |
| `rpt_tableau__academic_goals_rollup` | `state_test_crosswalked` | `models/extracts/tableau/` |

All 5 use the same shape: `LEFT JOIN` the crosswalk as `x`, predicate `<alias>.source_system = x.source_system and <alias>.raw_subject = x.raw_subject`, then `coalesce(x.illuminate_subject_area, <alias>.raw_subject)`.

### The resolved column keeps its existing name per consumer

It stays `illuminate_subject` in the fact and `subject_area` in score anchors, because `fct_assessment_scores_enrollment_scoped` hashes `su.subject_area` into `assessment_score_key`. Renaming it would recompose that key.

### The crosswalk is keyed on the column each source's CASE actually read

`pearson` and `cambium` on `subject`, `fldoe` on `assessment_subject`, `iready` on `subject`, `renlearn` on `_dagster_partition_subject`. Amplify has no raw subject column at all, so its blended-layer CTE supplies the literal `'DIBELS'` and the crosswalk translates it.

### Verification against production

Row counts cannot detect a silently failed join, so I used sharper invariants.

`fct_assessment_scores_enrollment_scoped`, key-set diff in both directions: 0 keys in the branch that are not in production, which proves nothing was re-keyed or added. 725 keys in production not in the branch, all internal-branch rows the crosswalk never touches. That delta traces to `int_assessments__scaffold`, which rebuilds 5 times a day. `n_rows = n_keys = 14994389`, 0 null keys.

`int_assessments__score_anchors`, per-source counts identical to production: iready 264885, state_nj 71996, dibels 62491, state_fl 22181, star 7966. Zero null `subject_area`.

`int_extracts__student_enrollments_subjects`: rows where `state_test_proficiency != 'No Test'` come to 43810, identical to production. That figure would collapse toward zero if resolution broke while the row count held. `iready_proficiency_eoy` at 66091 is an untouched control, also identical.

`rpt_tableau__academic_goals_rollup`, per subject: Math 15852 rows and 3769 non-null `scale_score_state`, Reading 15394 and 3802. The asymmetry makes a Reading/Math flip detectable. Zero null `subject`.

`rg 'illuminate_subject\b' -g '*.sql' src/dbt/kipptaf` returns no hits.

`trunk check --force --no-fix` over all 23 changed files: no issues.

### The goals rollup has 4 union branches, not 3

The 4th is Star EOY off `stg_renlearn__star`. Its `star_discipline` is already Reading or Math, not an Illuminate value, so it stays outside the crosswalk. Routing it through the shared CASE would have remapped every Star `Reading` to `Math`. It opts out by setting `source_system` to NULL, which is the same convention `int_assessments__score_anchors` uses for its internal-scores branch.

### The `else 'Math'` arm

The NJSLA branch previously returned NULL for anything that was neither Text Study nor Mathematics, and now returns `'Math'`. Against production, `int_pearson__all_assessments` at `assessment_name = 'NJSLA'` with that branch's filters applied carries only Mathematics 19620, English Language Arts 18583, English Language Arts/Literacy 3708 and Geometry 299. Every one maps to Text Study or Mathematics through the crosswalk, so the change is behavior-identical on today's data, not merely count-identical. It becomes a divergence only if NJSLA-S ever lands under `assessment_name = 'NJSLA'`.

### Known and out of scope

`rpt_tableau__academic_goals_rollup.sql:94` labels its grade-3 `administration_window = 'PM1'` branch as `'FAST PM3'`. That mislabel predates this branch and is unchanged by it. It may be deliberate: the branch is grade-3 only and uses `academic_year` rather than `academic_year + 1`, so fall PM1 appears to stand in as the comparison point for students with no prior-year state test. It deserves its own issue.

The course side carries `Geometry`, `Algebra I` and `Algebra II` as distinct subject areas, but Pearson collapses all 3 to `Mathematics` before the resolver join, so only the 4 coarse values ever match. This refactor preserves that behavior either way. If it is unintended it deserves its own issue.

`stg_google_sheets__dibels__expected_assessments_by_levels` carries a sheet-supplied `illuminate_subject` column of its own. It is a 7th place the vocabulary appears and is not one of the 6 this issue names.

### Rollback

Revert the merge commit. The crosswalk sheet tab and its external table can stay: nothing else reads them.

</details>


cbini:
Verdict per finding. 1 fixed, 3 declined with evidence, 1 confirmed.

### 1. New external source needs `--target staging` — declined, already done

I ran `stage_external_sources --target staging` for `google_sheets.src_google_sheets__assessments__vendor_subject_crosswalk` before opening this pull request, because the table was absent from `zz_stg_kipptaf_google_sheets`. It now carries the 3 expected STRING columns.

CI settles it directly. `stg_google_sheets__assessments__vendor_subject_crosswalk` built as a `BASE TABLE` in `dbt_cloud_pr_70403104388001_5361_google_sheets` at 20:15:43. The run built 438 relations across 17 schemas, and `get_job_run_error(warning_only=true)` returns 0 warnings.

The finding was right to raise this as a rule, and right to mark its confidence Medium. The plan line it cites is the dev-target step; the staging step is real work I did outside the plan text.

### 2. Dict-format unit-test fixtures on a brand-new model — declined, CI disproves it

A fixture that fails introspection raises at compile time and aborts the whole `dbt build` step. This run succeeded, and `int_assessments__score_anchors` built in `dbt_cloud_pr_70403104388001_5361_assessments` at 20:16:37, 54 seconds after the crosswalk model it mocks.

The rule the finding cites is about a column ADDED to an existing upstream, where dbt reads the deferred old-schema relation. Here the whole model is new and `state:modified+` builds it first, so introspection resolves against the freshly built relation.

### 3. `else 'Math'` collapses an unmapped subject — declined, verified, and guarded

The behavior change is real and stated correctly: the old Pearson branch had no `else` and produced NULL. I verified the exposure the finding asked me to check.

Against production, `int_pearson__all_assessments` at `assessment_name = 'NJSLA'` with this branch's filters carries only 4 values: Mathematics 19620, English Language Arts 18583, English Language Arts/Literacy 3708, Geometry 299. Every one maps to Text Study or Mathematics through the crosswalk. The change is behavior-identical on today's data, not merely count-identical.

I am not restoring the raw-value fallback here, for a reason specific to this file. This column holds a Reading/Math goal label, not an Illuminate subject area, and the downstream join keys on it. A raw fallback would put a vendor string such as `Science` into that column, which matches nothing — the same outcome as the old NULL, with less signal. The `covers_all_sources` test is the actual guard: any unmapped source and subject pair trips it as a warning before it can reach this branch.

Verified: the 4 values above, and that the coverage test returns 0 rows today. Inferred: that a future NJSLA-S landing under `assessment_name = 'NJSLA'` is the only way this diverges.

### 4. Two CTEs where siblings use one — fixed in 25304fef

Correct, and the split was my own over-correction rather than a rule requirement. A `CASE` is not a function call, so `case when coalesce(a, b) = 'x'` nests nothing. Only the earlier outer `if(..., case when coalesce(...))` broke the one-level limit in `.claude/rules/dbt-sql.md`, and folding the Star EOY arm into the same `CASE` already fixed that. The extra CTE bought nothing.

`state_test_resolved` now joins the crosswalk and resolves the value in one CTE, matching every other call site in this change.

I verified the merge is behavior-neutral by querying the branch build and production in a single statement, so both views read the same live upstreams: Math 15852 rows / 3864 `scale_score_state` / 14712 `scale_score`, Reading 15394 / 3904 / 14475. Identical on all 6 figures, 0 null `subject`.

### 5. Old mapping left in 3 sibling files — confirmed, no action

Correct on every point, including that `Refs #5340` rather than `Closes` is deliberate. Pull request 2 drops the column from `pearson`, `kippmiami` and `cambium`, and it is blocked until this merges and Dagster materializes it in production.

Issue #5363, opened 2026-09-16
fix: the academic goals rollup labels grade-3 FAST PM1 rows as FAST PM3


cbini:
## What's happening

`rpt_tableau__academic_goals_rollup` labels a set of rows `FAST PM3` when those rows come from the PM1 administration window.

The grade-3 branch writes `'FAST PM3' as assessment_type` on line 94, and that same branch filters `administration_window = 'PM1'` on line 109. Every other grade is labelled by a different branch that really does read PM3.

The label is visible to whoever reads this model. `assessment_type` feeds `benchmark_assessment_type`, which is a contracted column. Anyone who groups or filters on `benchmark_assessment_type = 'FAST PM3'` gets grade-3 PM1 scores mixed in with every other grade's PM3 scores, and nothing says so.

The model is live and heavily read. Over the 30 days to 2026-09-16, BigQuery logged 1,748 jobs referencing it: 1,702 from a service account and 47 from named users, the most recent within the hour.

No computed value is wrong. The model never branches on this string — its own filters test `i-Ready BOY` and `Star BOY` only. Scale scores, proficiency flags and goal comparisons are all unaffected. The label is the whole defect.

This affects Miami. FAST is the Florida assessment, so no other region has these rows.

## A second problem found while checking this

The model has no dbt exposure. The only YAML that names it is its own properties file, and nothing under `src/dbt/kipptaf/models/exposures/` declares it.

That blocks the fix below. The right person to approve a category rename is whoever owns the dashboard, and no exposure records who that is. It also means a column rename or removal here would break that consumer with no warning, which is the exact failure an exposure exists to prevent.

Whether the service-account traffic is Tableau extract refreshes is an inference from the `rpt_tableau__` prefix and the `kipptaf_tableau` schema, not something I verified.

## Steps to reproduce

1. Open `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__academic_goals_rollup.sql`.
2. Read the branch at lines 87 to 111. Line 94 sets the label to `FAST PM3`. Line 109 filters `administration_window = 'PM1'`.
3. Compare it to the branch at lines 59 to 83, which carries the same label and filters `administration_window = 'PM3'`.
4. In prod, query `kipptaf_tableau.rpt_tableau__academic_goals_rollup` for `benchmark_assessment_type = 'FAST PM3'`. The grade-3 rows in that result came from PM1.

## Where

[...]

Issue #5362, opened 2026-09-16
refactor(dbt): remove the duplicated vendor score unions across 3 layers


cbini:
## What and why

`kipptaf` writes the same vendor score union many times. This issue is the umbrella. Each piece of work gets its own sub-issue, listed below.

The first version of this issue named 5 union sites and asked that no other `kipptaf` model read the 5 vendor score models. A map of the duplication, built on 2026-09-16 against `c14694871c`, disproves both numbers. There are 13 union sites, not 5. Of the 34 models that read the vendor sources, about 19 read exactly 1 source and union nothing, so they have no duplication to remove.

The duplication sits at 3 layers.

### Layer 1 normalizes each source family once, and is correct

| Model | How it unions | What it unions |
| --- | --- | --- |
| `int_pearson__all_assessments`, kipptaf copy | `dbt_utils.union_relations` over 5 explicit `source()` relations, with a 55-column `include` and `source_column_name="_dbt_source_relation_2"` | the Newark, Camden and Paterson district `int_pearson__all_assessments` models, plus `stg_cambium__njgpa` for Newark and Camden |
| `int_fldoe__all_assessments` | 4 inline `union all` | `fast`, `fsa`, `eoc`, `science` |
| `int_amplify__all_assessments` | 4 inline `union all` | benchmark summary, 2 progress-monitoring summaries, expected assessments |
| `int_iready__diagnostic_results` | no union | 1 select over `kippmiami_iready` and `kippnj_iready` |
| `stg_renlearn__star` | no union | 1 select over `kippmiami_renlearn` |

iReady and Star hold no layer 1 union, so "collapse the union" does not apply to them here.

`int_pearson__all_assessments` exists twice, once in the `pearson` package and once in `kipptaf`. A `ref()` from `kipptaf` resolves to the kipptaf copy. The package copy is district-scoped and driven by `var("pearson_state_assessment_relations")`, which differs per project: `kipppaterson` lists only 2 relations. Read the kipptaf copy when reasoning about what `kipptaf` consumers receive.

[...]

PR #5369, opened 2026-09-16
refactor(dbt): drop illuminate_subject from the source-model producers

### LINKED ISSUE #5340: refactor(dbt): move the illuminate_subject mapping out of source models into the blended assessment layer

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will delete the `illuminate_subject` column from the three source models that still build it.

Three models each held their own copy of the same vendor-to-Illuminate subject mapping. #5361 replaced all three with one Google Sheets crosswalk that the blended assessment layer joins to. Nothing reads the producer copies any more, so they come out.

The three:

- `int_pearson__all_assessments` (pearson package)
- `int_fldoe__all_assessments` (kippmiami)
- `stg_cambium__njgpa` (cambium package)

This is PR 2 of 2 for #5340. PR 1 (#5361) is merged, and prod already shows zero `illuminate_subject` columns on all six kipptaf source models, so the consumer side is clear.

**Breaking change.** `stg_cambium__njgpa` is contract-enforced, so this drops a contracted column. Its properties file changes in the same commit, and the declared list now matches prod's column set minus the dropped column exactly: 54 of 55, checked name by name. Nothing outside this repo reads the column — the Tableau and Illuminate consumers go through the blended layer, which #5361 already switched over.

**Rollback plan.** Revert this pull request. The column is derived in the model SQL from columns that stay, so a revert restores it on the next build with no backfill and no data loss.

## Reviewer Notes

- **Dagster must rematerialize the dependent views after merge.** Dropping a contracted column can leave a dependent prod view with its old stored SQL, which then fails at read time. I will check for stranded views after the merge and relaunch any that need it.
- **dbt Cloud CI cannot validate this pull request.** It builds only `kipptaf`, and every file here is in a district project or a source package, so the check goes green on a build that selects nothing. The real validation is local, and it ran — see the fold-out.

### General

### dbt

- [ ] Not applicable — no model here is newly consumed by a dashboard. Include (or update) an [exposure](https://teamschools.github.io/teamster/reference/dbt-conventions/#exposures) for all models consumed by a dashboard, analysis, or application
- [ ] Not applicable — no external source changed. If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** so the dbt Cloud CI job can find the table. See [Staging external sources](https://teamschools.github.io/teamster/guides/dbt-development/#staging-external-sources)

## CI checks

Closes #5340

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Ran through `superpowers:brainstorming` → `writing-plans` → `subagent-driven-development`; spec at `docs/superpowers/specs/2026-09-16-illuminate-subject-crosswalk-design.md`.

**What was removed, exactly.** A `case` over `` `subject` `` in `int_pearson__all_assessments`, a `case` over `t.assessment_subject` in `int_fldoe__all_assessments`, and an `if` over `` `subject` `` in `stg_cambium__njgpa`. All three mapped `English Language Arts%` to `Text Study` and the three math course names to `Mathematics`, passing everything else through. The crosswalk sheet (`stg_google_sheets__assessments__vendor_subject_crosswalk`, 18 rows across 5 source systems) now carries those pairs, and `stg_google_sheets__assessments__vendor_subject_crosswalk__covers_all_sources` tests that every source system's raw values are present.

**Left in place on purpose, all verified:**

- The DIBELS sheet has a column with the same name. It is a separate vocabulary with its own source entry, out of scope per the spec, and feeds `int_google_sheets__dibels__expected_assessments_by_levels` and `int_google_sheets__dibels_expected_assessments`.
- `fct_assessment_scores_enrollment_scoped` derives its own `illuminate_subject` at lines 178 and 540 as `coalesce(x.illuminate_subject_area, <raw_subject>)`. That is the crosswalk output, not a producer copy, and `assessment_score_key` hashes it — removing it would rehash every key.
- Two comments, at `fct_assessment_scores_enrollment_scoped.sql:718` and `rpt_tableau__academic_goals_rollup.sql:115`.

A repo sweep (`rg 'illuminate_subject\b'` over `src/**/*.{sql,yml}`) returns exactly those 14 hits and no producer reader.

**Local validation, since dbt Cloud CI builds only kipptaf.** `dbt deps` then `dbt compile --target prod` for `kippnewark` (selecting `int_pearson__all_assessments` and `stg_cambium__njgpa`) and for `kippmiami` (selecting `int_fldoe__all_assessments`). Both commands succeeded, and each compiled file carries 0 occurrences of the dropped column.

**Contract check without a shared write.** `assert_columns_equivalent` runs only inside `dbt build`, and a `--target staging` build of `stg_cambium__njgpa` would recreate a shared `zz_stg_*` table. Instead: the properties file's 54 declared names were diffed against `kippnewark_cambium.INFORMATION_SCHEMA.COLUMNS` for `stg_cambium__njgpa` (55 columns) minus `illuminate_subject`. Identical, name for name. The model ends in `select *, <case> as aligned_aggregate_ethnicity from aligned`, so its output column set is the `aligned` CTE's enumerated list plus one — nothing implicit can slip in, which is the failure mode a build would otherwise catch.

**Post-merge, still to do:** confirm every consuming district deployed (`kippnewark`, `kippcamden` for cambium and pearson; `kipppaterson` for pearson; `kippmiami` for fldoe), then check `INFORMATION_SCHEMA.VIEWS.view_definition` for dependent views still holding the old SQL and relaunch them through Dagster.

**Related issues filed this session:** #5362 (consolidate the 5 duplicated crosswalk join sites), #5363 (the `FAST PM3` label on the grade-3 PM1 branch of `rpt_tableau__academic_goals_rollup`, plus that model's missing exposure).

</details>


cbini:
Review reports no findings, so there is nothing to adjudicate. Closing out the two items it marked **needs local verification** — both ran before the PR opened, and here is the evidence.

**1. Contract enforcement on `stg_cambium__njgpa`.** The review took the second of the two options it named, the `INFORMATION_SCHEMA.COLUMNS` diff, rather than a `--target staging` build, because that build recreates a shared `zz_stg_*` table that CI and other developers read.

The 54 names declared in the properties file were diffed against `kippnewark_cambium.INFORMATION_SCHEMA.COLUMNS` for `stg_cambium__njgpa`, filtered to exclude the dropped column. Prod carries 55 columns; 55 minus 1 is 54. Sorted, the two lists are byte-identical — `diff` returns empty.

That agrees with the review's own independent count, reached a different way: 53 aliased columns in the `aligned` CTE plus `aligned_aggregate_ethnicity` in the outer select. Two methods, same 54.

The residual risk a real build would catch is a column arriving implicitly through a `select *`. It cannot happen here: the model's output set is the `aligned` CTE's enumerated list plus that one outer column, and the removed column was an enumerated member of that list.

**2. Local `dbt compile`.** Ran in the worktree, after `dbt deps` (fresh worktree, no `dbt_packages/`):

- `dbt compile --target prod --project-dir src/dbt/kippnewark --select int_pearson__all_assessments stg_cambium__njgpa`
- `dbt compile --target prod --project-dir src/dbt/kippmiami --select int_fldoe__all_assessments`

Both logged `Command dbt compile succeeded` (35.6s and 29.1s). Each of the 3 compiled files carries 0 occurrences of the dropped column, counted with `grep -c`.

`trunk check --force --no-fix` over all 6 changed files: `No issues`.

One correction to the summary above it: the review says it "couldn't run `dbt compile`/`dbt build` in this environment (no warehouse access)". That is right about its own sandbox and is the correct thing to flag, not a defect.

PR #5374, opened 2026-09-16
perf(powerschool): cron GPA snapshots at 23:00 and fix topline date cast

### LINKED ISSUE #5218: perf(powerschool): GPA snapshots merge 143 times a day each
## What and why

`snapshot_powerschool__gpa_term` and `snapshot_powerschool__gpa_cumulative` each run a `MERGE` about 143 times a day, roughly every 10 minutes. Together they cost 40 slot hours over 7 days. Each run is cheap. The cadence is the cost.

GPA changes when grades are stored, which is a few times a term, plus daily gradebook edits. A snapshot every 10 minutes records almost no new history.

## Proposed approach

Decide the cadence GPA history actually needs, likely once a day, and schedule the snapshot on that instead of on every upstream materialization.

<details>
<summary>For Claude</summary>

> Count from `JOBS_BY_PROJECT`, `statement_type = 'MERGE'`, 24 hours: 143 for `snapshot.kipptaf.snapshot_powerschool__gpa_term`. The snapshot is selected by whatever Dagster asset job includes the kipptaf powerschool dbt models. Find the automation condition or schedule that triggers it before changing anything.
>
> Verify: merge count per day drops to the agreed cadence and `dbt_valid_from` gaps in the snapshot table still cover every stored-grade change.
>
> Parent: #5212

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will cut the two PowerSchool GPA snapshots from about 143 BigQuery `MERGE` jobs a day each down to 1, and fix a date bug in the two models that read them.

`snapshot_powerschool__gpa_term` and `snapshot_powerschool__gpa_cumulative` have no automation condition, so each one inherits the eager table condition and re-runs on every upstream change. Measured over the 24 hours ending 2026-09-16: 138 and 147 merges, 6.13 slot hours between them. Each merge is cheap. The cadence is the cost.

Both snapshots now run on a `0 23 * * *` cron condition instead. GPA changes a few times a term plus daily gradebook edits, so a snapshot every 10 minutes records almost no new history.

This pull request also fixes a date bug the cadence work uncovered. `int_topline__gpa_term_weekly` and `int_topline__gpa_cumulative_weekly` derived their dates with `cast(dbt_valid_from as date)`. A bare cast truncates in UTC, so any evening capture in America/New_York landed on the next day. Both models now use `date(dbt_valid_from, '{{ var("local_timezone") }}')`, which matches `int_powerschool__gpa_term_lookback`.

## Reviewer Notes

The topline date fix changes values, not columns. It re-attributes stored history to the correct local day, so weekly GPA figures shift for the evening captures that were landing a day late. That is a correction, and 84% of row-versions were already correct before it.

`tests/test_automation_conditions.py` has 6 failures on `main` that this branch does not cause and does not fix. They are `test_view_requested_on_execution_failure`, `test_view_code_version_change_after_failed_materialization`, and the 4 `TestKipptafChainTopologies` tests. At least 2 of them assert an `execution_failed` trigger that no longer exists in `src/teamster/core/automation_conditions.py`.

Nothing rebuilds at deploy. A properties-only change does not alter a model's `code_version`, so both snapshots keep their current relation until the first cron tick at 23:00 America/New_York.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — no new models.
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — no exposure change; the `student_course_grades` Tableau exposure already covers the affected lineage.
- [x] **Breaking change?** No. No column is renamed or removed, and no contracted model changes shape.
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — no external source changes.

## CI checks

## Troubleshooting

Closes #5218

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Spec: `docs/superpowers/specs/2026-09-16-gpa-snapshot-cadence-design.md`. Plan: `docs/superpowers/plans/2026-09-16-gpa-snapshot-cadence.md`.

**Why the churn happened.** A dbt snapshot's `config.materialized` is `snapshot`, which is not `view` or `ephemeral`, so `CustomDagsterDbtTranslator.get_automation_condition` routes it to `dbt_table_automation_condition()`. The two snapshots reach different upstreams by different paths. `snapshot_powerschool__gpa_term` is declared on `int_powerschool__gpa_term_current`, which is `materialized: ephemeral` — but dagster-dbt creates no asset for an ephemeral model, verified against the built asset graph, so the snapshot's Dagster dep is `int_powerschool__gpa_term`, a view. `_VIEW_SELECTION` matches a view, so `_build_any_ancestor_updated` follows it down to the district source tables. `int_powerschool__gpa_cumulative` (the kipptaf copy, not the `src/dbt/powerschool` package copy — `ref()` resolves to the current project) is `materialized: table`, so its churn is plain `any_deps_updated()` on that table's own eager rebuilds.

**Why 23:00 and not the `0 0 * * *` used by the #4821 snapshots.** `int_powerschool__gpa_term_lookback` compares versions against `timestamp(date_add(date_sub(current_date(tz), interval N day), interval 1 day), tz)` for N of 7, 14, 28 — local-midnight boundaries. A 23:00 capture clears the boundary with an hour of margin and resolves to the end-of-day value, which is the contract the model's own inline comment states. A 00:00 capture lands just past the boundary and resolves one day stale, on sub-second run-start timing.

**The topline cast was a real latent bug, verified not assumed.** `cast(timestamp as date)` truncates in UTC. Checked in BigQuery: `23:00 ET 2026-09-14` casts to `2026-09-15`, and `00:00 ET 2026-09-15` also casts to `2026-09-15`. Over 14 days and 232,219 row-versions of `snapshot_powerschool__gpa_term`, every one of the 37,266 versions captured in local hours 20 through 23 stamped the wrong local date, and all 194,953 captured in hours 0 through 19 stamped the right one. Under a single daily 23:00 tick, 100% would have been wrong. The `date(..., local_timezone)` fix removes the cadence sensitivity entirely and re-attributes the whole stored history, because the date is derived at read time rather than stored.

**Accepted loss.** About 100 capture instants a day collapse to 1, so a grade stored and then revised the same day keeps only the end-of-day value. No consumer reads intra-day versions: all 3 resolve at day grain, and a repo-wide search found no Cube file and no other surface reading either snapshot. Separately, the lookback no longer sees a change stored between 23:00 and local midnight; that change lands on the next day's capture.

**Verification after the first tick.** Confirm the daily merge count for both tables in `JOBS_BY_PROJECT` drops to about 1 each from 147 and 138. Note the BigQuery MCP blocks the write-verb literal even inside a string, so filter `statement_type` with `like 'MER%'`. Confirm `dbt_valid_from` still carries a version on every day a grade store occurred. Confirm `int_powerschool__gpa_term_lookback` still returns non-null `gpa_y1_1_week_prior`, `gpa_y1_2_week_prior`, and `gpa_y1_4_week_prior`.

**Known gap, not this pull request's scope.** Neither topline weekly model carries a uniqueness test, which `.claude/rules/dbt-models.md` requires of every intermediate model. Measured against the full prod tables on 2026-09-16: of 442,459 student-week keys, `int_topline__gpa_term_weekly` duplicates 173,205 (39.1%) and `int_topline__gpa_cumulative_weekly` duplicates 299,399 (67.7%), up to 3 rows on one key. The duplicates disagree — 161,938 of the term model's duplicated keys carry a different `gpa_y1` per row, and on 23,063 the at-or-above-3.00 flag flips. `int_topline__student_metrics` scores each row per student, so those students are weighted up to 3 times. Root cause is the abutting-interval `BETWEEN` join that `.claude/rules/dbt-sql.md` warns about. Filed as #5375. Note that this PR shrinks the duplicate count without fixing it, because consecutive daily versions still straddle the boundary day.

**Stale by design.** The plan's Step 5 snippet still shows the pre-correction YAML comment. Plans under `docs/superpowers/` are working documents and are expected to go stale; the shipped YAML and the spec both carry the corrected rationale.

</details>


cbini:
Per-finding verdicts on the review above. Each claim was checked against the code before I acted on it. Fixes are in `9d0d150010`.

**1. `BETWEEN` fan-out on the topline weekly joins — agreed, and worse than either of us measured.**

The rule citation holds: `.claude/rules/dbt-sql.md:215-223` names this exact pattern, half-open intervals for abutting ranges.

The 24.5% figure in my PR body was scoped to one `yearid` and 8 Mondays, and it understated the problem. Measured against the full prod tables:

| Model | Student-week keys | Duplicated | Share | Max rows/key |
| `int_topline__gpa_term_weekly` | 442,459 | 173,205 | 39.1% | 3 |
| `int_topline__gpa_cumulative_weekly` | 442,459 | 299,399 | 67.7% | 3 |

The duplicates do not agree with each other. 161,938 of the term model's duplicated keys carry a different `gpa_y1` per row, and on 23,063 the at-or-above-3.00 flag flips between rows — the same student, same week, scored as both meeting and missing the bar. `int_topline__student_metrics` turns each row into a per-student score, so those students are weighted up to 3 times in every topline GPA figure.

Filed as #5375. I corrected the PR body to carry these numbers instead of the 24.5%.

One thing worth flagging for whoever picks up #5375: this PR shrinks the duplicate count without fixing the cause. Cutting roughly 100 snapshot versions a day to 1 reduces the overlap, but consecutive daily versions still straddle the boundary day. A lower count after this merges is not a fix.

**2. Missing uniqueness test and missing `description:` — descriptions added, test deferred with a reason.**

Both properties files now carry a model description and a description on all 5 columns. `.claude/rules/dbt-yaml.md` requires them on any modified model, and this PR modifies both, so the rule applies.

The uniqueness test stays with #5375 rather than landing here. Adding it now would fail immediately against the 173,205 and 299,399 duplicate keys already in prod. It belongs in the same change as the fan-out fix, where it is the regression guard. Both descriptions say so and link the issue.

**3. Stale churn figures in `int_students__gpa.yml` — declining.**

That comment is the rationale for `int_students__gpa`'s own cron cadence, and it reads as history: the model was one of several driving snapshot churn at the time #5082 was written. Its argument for that model's cron does not depend on the snapshots still churning today, so the comment is not made wrong by this PR.

Editing it also has a cost the finding does not price in. `int_students__gpa` is untouched by this PR, and `state:modified+` is branch-versus-deferred-environment, so a comment edit there pulls the model and its whole descendant graph into the CI selection for a doc change. Not worth it.

**4. The ephemeral-traversal doubt was well-founded, and my explanation was wrong.**

You flagged that `_build_any_ancestor_updated`'s `view_selection=_VIEW_SELECTION` matches assets tagged `dagster/materialized: view`, that `get_tags` would tag an ephemeral model `ephemeral`, and that you could not settle it from static reading. Correct on all 3 counts, and the conclusion is sharper than static reading could reach.

I probed the built asset graph. dagster-dbt creates **no asset at all** for `int_powerschool__gpa_term_current` — it does not appear in the specs under any key. So the tag question does not arise: `snapshot_powerschool__gpa_term`'s actual Dagster dep is `int_powerschool__gpa_term`, which is `materialized: view`. `_VIEW_SELECTION` does match that, so the view-chain recursion claim was right, but the model I named as driving it was not.

Corrected in 3 places: the `gpa_term` comment in `snapshots/powerschool.yml`, the `test_gpa_snapshots_get_cron_condition` docstring, and 2 passages in the design spec. The measured merge counts are unaffected, as you noted.

**On the commands you could not run:** `trunk check --force --no-fix` passes on all 5 files, and `test_gpa_snapshots_get_cron_condition` passes. `tests/test_automation_conditions.py` still has the 6 failures that exist on `main` and are called out in the PR body.

PR #5380, opened 2026-09-17
fix(powerschool): give HS sections a unique gradebook audit display label

### LINKED ISSUE #5379: fix: gradebook audit merges two high school sections that share a period
## What's happening

The gradebook audit dashboard lists one row per course and per section. At high school it labels that row with the class period. Two sections of the same course can meet in the same period with the same teacher, so both sections carry the same label. The dashboard then merges the two sections into a single row and adds their assignment counts together.

I expected each section to get its own row, counted against its own target. Instead one row shows the sum of both sections. A teacher who entered 3 of the 4 required Work Habits assignments in each of two sections sees "6/4" in red. The count looks met and the colour says it is not.

The warehouse data is correct. `rpt_tableau__gradebook_audit` already holds one row per section, each showing 3 of 4 and each flagged. Only the dashboard label merges them, so no teacher was ever miscounted. The display is wrong, not the audit.

This affects 14 teacher, course, and period groups across 28 sections. Every one of them is at KIPP Newark Collegiate Academy. No middle school is affected, and neither of the other 2 high schools is affected.

## Steps to reproduce

1. Open the "Gradebook Teacher View" sheet of the Academic & Gradebook Health Suite workbook.
2. Filter to KIPP Newark Collegiate Academy and quarter Q1.
3. Find a teacher who teaches 2 sections of one course in the same period. Physical Education and Driver's Education both qualify, because each rotates with another class.
4. Read the Work Habits count for that period. It shows the 2 sections added together, in red.

## Where

- **Environment:** prod
- **Run, PR, or dashboard link (if any):** Academic & Gradebook Health Suite, workbook `b3c14d67-3130-46ac-82a0-0637a5cc2da5`, sheet "Gradebook Teacher View". Reported in Zendesk ticket 482235.

## Also in scope

The `academic_gradebook_health_suite` exposure does not list 2 of the 5 models its workbook reads. It omits `rpt_tableau__gradebook_audit` and `rpt_tableau__gpa_goal_progress`. That exposure carries `cron_schedule: 0 4 * * *`, which Dagster runs as a real extract refresh. Because the audit model is missing from the dependency graph, nothing orders the 4am refresh after the audit model rebuilds. The dashboard can therefore refresh from a stale audit table. This is a separate defect from the section label, and I found it while tracing the label.

The `gradebook_audit` exposure points at workbook `f06a1938-9082-4c02-a0e4-f450b687dc43`. That workbook holds 1 sheet named "Sheet 1", has 0 views, and reads an extract taken on 2026-06-26 from a dbt Cloud CI schema that no longer exists in BigQuery. It is a dead artifact, and the exposure sends a reader to the wrong dashboard.

<details>
<summary>For Claude</summary>

### Root cause

`int_extracts__course_schedule_by_term.sql:138-140` derives the display label:

```sql
if(s.school_level_alt = 'HS', s.external_expression, s.section_number) as section_or_period,
```

At HS this resolves to `external_expression` (the period, e.g. `1(A)`), which is not unique across sections. At MS it resolves to `section_number` (the cohort name, e.g. `5Columbia`), which is. So the collision is structurally HS-only.

Four worksheets in the workbook put `section_or_period` on Rows as the finest grain with no `sectionid` in the LOD, which is what produces the merge: `Your sections grid`, `Your sections flags`, `Teacher sections panel`, `Sheet Card - shortfalls`.

Confirmed for `Phys Ed and Health II`, period `1(A)`, Q1: `sectionid` 37627 (`section_number` `1a`) and 37705 (`1b`), each `assignments_entered_count_no_flags = 3` against `expectation = 4`, each `not_enough_assignments = true`.

### Fix

HS section numbers across all 3 high schools already lead with the period and add a qualifier (`1a`, `1b`, `1ICR`, `1eleventh`, `2ACC`, `12DB`, `3anthro`), so the label needs no new source column:

```sql
if(
    s.school_level_alt = 'HS',
    array_to_string([s.external_expression, s.section_number], ' '),
    s.section_number
) as section_or_period,
```

`array_to_string` rather than `concat` because it skips nulls. 1,724 HS rows in that intermediate carry a null `external_expression` and so show a blank label today; `section_number` is never null, so this makes the label unconditionally populated. Those rows sit outside the audit's filters, so that part is latent robustness rather than a visible fix.

Changing the value in place rather than adding a column means no workbook edit. Blast radius is one report: this column feeds only the gradebook audit chain, and `int_extracts__gradebook_audit_student_flags` does not project it. The other dashboards showing a section or period each derive their own copy.

### Tableau side, verified from the `.twb` XML

Nothing is pinned to the current string values: no group, no bin, no filter, no manual sort, no member alias on the field, and no literal `1(A)`-shaped string anywhere in the file. Two narrative calculated fields reference the field and already use `COUNTD([sectionid])` for their counts, so they only get more precise. The workbook reads prod `kipptaf_tableau` via extracts, so the fix lands on the next 4am refresh.

### Files

1. `int_extracts__course_schedule_by_term.sql` — the label expression
2. its properties yml, plus `rpt_tableau__gradebook_audit.yml` — the `section_or_period` descriptions
3. `exposures/tableau.yml` — add the 2 missing `depends_on` refs to `academic_gradebook_health_suite`; disable the stale `gradebook_audit` exposure, following the `gradebook_audit_teacher_report` precedent
4. `docs/models/gradebook-audit-data-model.md` — record the label rule and correct the exposure pointer
5. `.claude/skills/gradebook-audit/SKILL.md` — its "List refs" procedure names the `gradebook_audit` exposure as authoritative, which sent this investigation to the dead workbook. It needs to name `academic_gradebook_health_suite`.

### Fixed means

- `uv run dbt build --select int_extracts__course_schedule_by_term+ --project-dir src/dbt/kipptaf --defer --state target/prod` passes.
- The collision query returns 0 groups where one teacher, course, quarter, and `section_or_period` covers more than 1 `sectionid`. It returns 14 today.
- Every section and quarter still has exactly 4 `category_summary` rows.
- MS `section_or_period` values are byte-identical before and after.

### Reply to the reporter

The reporter named 6 teachers. One more is affected and is not in the ticket: the AP Seminar sections in periods 7, 8, and 9, which split by grade rather than by rotation. Worth saying plainly that no gradebook was miscounted, because that is the part that matters to a teacher who saw a red 6/4.

</details>

anthonygwalters:
# Pull Request

## Summary & Motivation

When merged, this pull request will give every high school section its own row on the gradebook audit dashboard.

The dashboard groups its section rows on `section_or_period`. At high school that column held the class period alone, which is not unique — two sections of one course can meet in the same period with the same teacher. Both sections shared a label, so the workbook added their assignment counts together. A teacher who entered 3 of the 4 required Work Habits assignments in each of two sections saw "6/4" in red.

This change combines the period with the section number at high school. Section numbers at all 3 high schools already lead with the period and add a qualifier, so `1(A)` becomes `1(A) 1a` and `1(A) 1b`. Middle school keeps the section number alone, which names the cohort and is already unique.

The warehouse rows were always correct, one per section. Only the label merged them, so no gradebook was ever miscounted. The dashboard needs no edit — it groups on this column, so distinct values split the rows on their own.

This also fixes 2 exposure defects found while tracing the label. See "Reviewer Notes".

Refs #5379

## Reviewer Notes

**Every high school label changes, including the 290 that were never broken.** A plain section now reads `1(A) 1` instead of `1(A)`. That is the cost of a label that is correct by construction rather than only where a collision happens today. The alternative was a label that changes whenever a section is added or dropped, which breaks quietly.

**The `academic_gradebook_health_suite` exposure was missing 2 of the 5 models its workbook reads**, including `rpt_tableau__gradebook_audit`. That exposure carries a real Dagster extract refresh at `0 4 * * *`. With the audit model outside the dependency graph, nothing ordered the refresh after the audit rebuild, so the dashboard could refresh from a stale table. This is a live freshness bug, separate from the label.

**The `gradebook_audit` exposure is now disabled.** Its workbook holds 1 sheet, has 0 views, and reads an extract taken on 2026-06-26 from a dbt Cloud CI schema that no longer exists in BigQuery. The `gradebook-audit` skill named that exposure as the authoritative consumer, which sent this investigation to the dead workbook before I found the live one. The skill and the reference doc now name `academic_gradebook_health_suite`.

**No workbook edit, confirmed from the published XML rather than assumed.** Nothing is pinned to the old string values.

### General

      — not done, needs a human

### dbt

      models; updated the `section_or_period` description on both models that
      carry it
- [x] Include (or update) an
- [x] **Breaking change?** No column is renamed or removed. The change is
      value-only on an existing contracted string column, so no downstream
      exposure breaks. Rollback is a revert plus one extract refresh.

### Docs

- [x] No schedule, sensor, or integration change. Updated
      `docs/models/gradebook-audit-data-model.md` and the `gradebook-audit`
      skill.

## CI checks

- [ ] **Trunk** — passes
- [ ] **dbt Cloud** — passes

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Originated from Zendesk ticket 482235.

### The expression

`int_extracts__course_schedule_by_term.sql`, final `SELECT`:

```sql
if(
    s.school_level_alt = 'HS',
    array_to_string([s.external_expression, s.section_number], ' '),
    s.section_number
) as section_or_period,
```

`array_to_string` rather than `concat` because it skips nulls. 1,724 HS rows in this intermediate carry a null `external_expression` and so produce a null label today; `section_number` is never null, so the label becomes unconditionally populated. Those rows sit outside the audit's filters (`section_quarter_count >= 2`, `sections_no_of_students != 0`, `exclude_from_gpa = 0`), so that part is latent robustness, not a visible fix. One level of function nesting, within the repo limit, and `array_to_string` is already the idiom in this chain (`flag_reasons` in `rpt_tableau__gradebook_audit`).

Changing the value in place rather than adding a column is what avoids a workbook edit.

**Corrected after review — my original blast-radius claim here was wrong.** I wrote that this touched one report because `int_extracts__gradebook_audit_student_flags` does not project the schedule model's `section_or_period`. That is true and irrelevant: the flags model derives the column from `int_extracts__course_enrollments_by_term`, which held its own unfixed copy of the same expression and feeds the live `rpt_gsheets__gradebook_audit_student_flags` Google Sheet. `claude-review` caught it. Commit 3b26df585 fixes that second copy, adds a uniqueness test encoding the invariant, and documents the duplication; #5383 tracks deriving the label once. Full verdicts are in the review reply comment.

So the real reach is two consumers, both in the gradebook audit chain: `rpt_tableau__gradebook_audit` and `rpt_gsheets__gradebook_audit_student_flags`. `rpt_tableau__student_course_grades`, `rpt_tableau__gradebook_gpa`, `rpt_tableau__assignment_checks` and `rpt_tableau__gradebook_ms_hs_comments` each derive their own copy of a similar label and are untouched.

### Verification

Dev build of `int_extracts__course_schedule_by_term+`, `--target dev --defer --favor-state` against a freshly parsed prod manifest. PASS=8, WARN=0, ERROR=0, including 4 uniqueness tests.

| Check | Before | After |
| --- | --- | --- |
| Colliding teacher/course/quarter/label groups | 14 | 0 |
| Max sections sharing one label | 2 | 1 |
| Section-quarters off the 4-row `category_summary` floor | 0 | 0 of 1,574 |
| Null `section_or_period` | 0 | 0 |
| MS labels changed | — | 0 of 890 |
| HS labels changed | — | 684 of 684 |

No rows gained or lost on either side of a full join keyed on `sectionid`, `quarter`, `school_level`.

Worked case, `Phys Ed and Health II` Q1 period 1: `sectionid` 37627 (`1a`) and 37705 (`1b`), each 3 of 4 Work Habits, each `not_enough_assignments = true`. Before, one row reading 6 of 4. Period 4 of the same course is a better illustration — `4a` is at 2 of 4 and `4b` at 3 of 4, a real difference the merge was masking as 5 of 4.

### Tableau, read from the published `.twb`

Workbook `b3c14d67-3130-46ac-82a0-0637a5cc2da5`, reading prod `kipptaf_tableau` through extracts.

4 worksheets put `section_or_period` on Rows as the finest grain with no `sectionid` in the level of detail: `Your sections grid` (the sheet in the ticket screenshot), `Your sections flags`, `Teacher sections panel`, `Sheet Card - shortfalls`.

Nothing is pinned to the current string values: no Tableau group, no bin, no filter of any kind on the field, no manual sort, no member alias, and no literal `1(A)`-shaped string anywhere in the file. 8 calculated-field occurrences reference it; 4 are `[course_name] + " - " + [section_or_period]` display labels, and the other 4 are narrative strings that already use `COUNTD([sectionid])` for their counts, so they only become more precise.

Tooltips stop agglomerating by the same mechanism. `Tooltip - category reasons` has empty `<rows/>` and `<cols/>` — its whole slice comes from 4 action-filter tuples, and `section_or_period` is in all 4. Its calcs are all `SUM(...)` or `SUM(...) > 0` over the slice, so they currently aggregate both sections and afterwards do not. Across the audit datasource, all 20 action-filter tuples on 12 worksheets include `Section Or Period`, and `assignment_name` appears nowhere in the workbook, so no sheet renders assignment detail at a separate grain.

Side effect checked because it would change colours: `has_grade_above_100` and `has_grade_below_70_no_comment` are section-level booleans broadcast onto every row of a section, so they can bleed between two collided sections. Across all 68 colliding cells both sections agree on both booleans, so no cell changes colour — only the counts split. The bleed is structurally possible and this closes it.

Not verified here: a hover and a click need a human, and the read-only Tableau MCP returns no calculated-field text, so every claim above comes from the downloaded `.twb` XML. The fix reaches the dashboard on the next 4am extract refresh.

### For the ticket reply

The reporter named 6 teachers. One more is affected and is not in the ticket: `AP Seminar`, periods 7, 8 and 9, which splits by grade (`7eleventh` / `7twelfth`) rather than by rotation. Worth telling her plainly that no gradebook was miscounted, since that is what matters to a teacher who saw a red 6/4.

The field caption in the workbook is "Section Or Period", which gets a little inaccurate at HS now that it shows both. Renaming it is a workbook edit, not a dbt one, and is the owner's call.

</details>


anthonygwalters:
## Review verdicts

Addressed in 3b26df585. Verdict per finding.

### 1. Duplicate unfixed expression in `int_extracts__course_enrollments_by_term` — ACCEPTED, fixed

Confirmed exactly as described, and it invalidates a claim in my PR body. `int_extracts__gradebook_audit_student_flags.sql:57` projects `s.section_or_period` where `s` is `int_extracts__course_enrollments_by_term` (line 82), not the model this PR fixed. My "blast radius is one report" note was true of the schedule model's column and missed that the flags model sources this column from a different model entirely. The `rpt_gsheets__gradebook_audit_student_flags` exposure is live with no `enabled: false`. I have corrected the PR body.

Applied the same expression in the enrollments model.

**One correction to the finding.** The suggested verification query groups by `teacher_number, course_number, quarter, academic_year, section_or_period` with no `_dbt_source_project` and no `schoolid`. PowerSchool teacher numbers are instance-scoped, so that key merges different people across regions. Run as written it reports **9,465** colliding groups and a `max_sections` of 8. Correctly keyed it is **342** groups in AY2026 with a `max_sections` of 4 — about 28x smaller. I made the same mistake first and caught it on the `max_sections: 8` value, which is not reachable for one teacher, one course, one period.

Blast radius on the actual live consumer, which is what matters here, since the sheet applies the audit's own filters and shows flagged rows only:

| | Before | After |
| --- | --- | --- |
| Groups on the sheet | 756 | 762 |
| Colliding groups | 10 | 0 |
| Flagged rows affected | 90 of 3,332 | 0 |

Your read on severity was right — the sheet is row-level with `sectionid` present, so this never produced the arithmetic bug, only an ambiguous label. Worth fixing anyway as the same defect class in a live consumer.

### 2. `array_to_string` skips nulls but not empty strings — REJECTED, does not occur

Measured across all 72,890 rows of `int_extracts__course_schedule_by_term`: 0 rows where `external_expression = ''` and 0 where `section_number = ''`. So no leading space is reachable, and the uniqueness property would survive it regardless since the section number still disambiguates.

Not guarding it deliberately. `nullif(external_expression, '')` inside the `if` would push the expression to 2 levels of function nesting, past the limit in `.claude/rules/dbt-sql.md`, and would need a CTE — a real structural cost for a cosmetic case that does not occur. You flagged it as low confidence needing verification, which was the right call; this is the verification.

### 3. No test encodes the uniqueness invariant — ACCEPTED, fixed

Added a second `dbt_utils.unique_combination_of_columns` on `int_extracts__course_schedule_by_term`, `severity: error` to match the existing one. You were right that the existing test is keyed on `sectionid`, which was never the property that broke.

**One deviation.** I included `schoolid` in the key, which your suggestion omitted. Both forms hold on current data — 72,890 rows, 72,890 distinct keys either way — so I chose on future false-failure risk rather than on today's data. The dashboard's own grain includes school, so the stricter no-`schoolid` form would fail on a teacher assigned to 2 schools teaching the same course number in the same period, which is not a defect for any consumer. The key now matches the consumer's grain exactly.

### 4. Two independently-maintained copies will diverge again — ACCEPTED as follow-up

Opened #5383. Noting there what makes this non-trivial rather than a quick hoist: the 2 copies read from different upstreams (`int_students__course_sections` versus `base_powerschool__course_enrollments`), so there is no existing shared home, and the obvious shortcut of joining the schedule model for the label would null it for single-quarter-term sections, which the schedule model drops via `section_quarter_count >= 2` and the enrollments model keeps.

Also recorded the duplication as a warning admonition in `docs/models/gradebook-audit-data-model.md`, so the next person to edit either copy finds the other before shipping. Your finding is the proof that a doc note was needed.

### 5, 6, 7, 8, 9 — no action, and one of them earned its place

Agreed on all. #5 is the useful one: I asserted the exposure `depends_on` change fixes a real Dagster refresh-ordering bug, and you verified it against `build_tableau_workbook_refresh_asset` rather than taking the claim. That is the check I could not do from the PR body alone.

### 10. Stray local revert of `SKILL.md` in the review checkout — agreed, job artifact

Confirmed independently. `git show HEAD:.claude/skills/gradebook-audit/SKILL.md` matches the intended content on both commits.

### On your "not verified" list

Reasonable to hold those at arm's length. For the record, the warehouse figures came from a dev build with `--defer --favor-state`, and the Tableau claims came from the published `.twb` downloaded through `tableauserverclient` — not from the read-only Tableau MCP, which returns no calculated-field text. Two things still need a human regardless: a hover and a click cannot be rendered, and Desktop opening the workbook is the only real check that it is uncorrupted. Nothing in this PR edits the workbook, so neither is a blocker here.

Issue #5381, opened 2026-09-17
fix: seven more topline weekly models fan out, and star_assessment_weekly is empty


cbini:
## What's happening

`int_topline__student_metrics` unions 18 `int_topline__*_weekly` models and turns each row into one per-student score. A model that returns more than one row per student-week therefore weights that student more than once. #5375 fixed the 2 GPA models. Seven more have the same defect, and one of them is worse than either GPA model.

Measured against prod on 2026-09-17. These models carry no discriminator column — `int_topline__student_metrics` reads them at student-week grain, so every duplicate is a double-count:

| Model | Student-week keys | Duplicated | Most rows on one key |
| --- | --- | --- | --- |
| `int_topline__attendance_interventions_weekly` | 809,706 | 422,821 | 354 |
| `int_topline__dibels_pm_weekly` | 22,742 | 18,260 | 5 |
| `int_topline__college_matriculation_weekly` | 32,823 | 173 | 2 |
| `int_topline__dibels_benchmark_weekly` | 689,923 | 38 | 2 |
| `int_topline__truancy_weekly` | 861,248 | 23 | 2 |

`attendance_interventions_weekly` is the one to look at first. It feeds a numerator and denominator pair (`successful_call_count` / `total_anticipated_calls`), and 354 rows on a single student-week is not a boundary-overlap shape — it looks like a different, larger grain problem.

Four more models do carry a discriminator that `int_topline__student_metrics` uses, so their student-week repetition is expected. They still have a smaller residual at their real grain:

| Model | Real grain tested | Duplicated | Most rows on one key |
| --- | --- | --- | --- |
| `int_topline__iready_diagnostic_weekly` | plus `discipline` | 11,178 | 3 |
| `int_topline__iready_lessons_weekly` | plus `discipline` | 11,178 | 3 |
| `int_topline__state_assessments_weekly` | plus `discipline`, `test_round`, `region` | 684 | 2 |
| `int_topline__college_entrance_exams_weekly` | plus `test_type` | 183 | 2 |
| `int_topline__formative_assessment_weekly` | plus `discipline`, `formative_strategy` | 46 | 2 |

Separately, `int_topline__star_assessment_weekly` holds 0 rows in prod. `int_topline__student_metrics` reads it for the "Star Diagnostic" indicator, so that indicator is empty. Either the model is broken or it should be disabled.

[...]

PR #5382, opened 2026-09-17
fix(dbt): resolve topline GPA weekly models to one row per student-week

### LINKED ISSUE #5375: fix: topline GPA weekly models fan out, double-counting students in topline metrics
## What's happening

`int_topline__gpa_term_weekly` and `int_topline__gpa_cumulative_weekly` should return 1 row per student per week. Both return up to 3.

The extra rows are not harmless copies. They disagree with each other. In `int_topline__gpa_term_weekly`, 173,205 student-week keys are duplicated, and 161,938 of those carry a different `gpa_y1` on each row. For 23,063 of them the "at or above 3.00" test flips between the rows, so the same student in the same week counts as both meeting the bar and missing it.

`int_topline__student_metrics` reads both models and turns each row into a per-student score with `if(gpa_y1 >= 3.00, 1, 0)`. A student with 3 rows is weighted 3 times against a student with 1. Every topline GPA figure built on those scores is therefore wrong by an unknown amount.

Measured against prod on 2026-09-16:

| Model | Student-week keys | Duplicated | Share | Most rows on one key |
| `int_topline__gpa_term_weekly` | 442,459 | 173,205 | 39.1% | 3 |
| `int_topline__gpa_cumulative_weekly` | 442,459 | 299,399 | 67.7% | 3 |

Neither model has a uniqueness test, which is why this went unseen. `.claude/rules/dbt-models.md` requires one on every intermediate model.

## Steps to reproduce

1. Run this against prod:

```sql
with k as (
  select student_number, academic_year, week_start_monday, count(*) as n
  from `teamster-332318.kipptaf_topline.int_topline__gpa_term_weekly`
  group by 1, 2, 3
)
select count(*) as key_groups, countif(n > 1) as duplicated, max(n) as max_rows
from k
```

2. Read `duplicated`. It should be 0. It is 173,205.
3. Swap in `int_topline__gpa_cumulative_weekly` and run it again. It is 299,399.

## Where

- **Environment:** prod
- **Run, PR, or dashboard link (if any):** found while reviewing #5374

<details>
<summary>For Claude</summary>

> Context, not instructions.

**Suspected root cause.** Both models join the snapshot to the enrollment week spine with a date-range predicate and no tiebreaker:

```sql
and co.week_start_monday between gpa.dbt_valid_from_date and gpa.dbt_valid_to_date
```

The `deduplicate` CTE above it partitions by `dbt_valid_from_date`, so it keeps 1 row per snapshot day — but several days' versions can each span the same Monday. A version valid Saturday through Monday and a version valid Monday onward both satisfy `between` for that Monday. The dedupe is at the wrong grain: it dedupes the snapshot, not the join result.

**#5374 reduces this but does not fix it.** That PR puts both snapshots on a `0 23 * * *` cron, cutting roughly 100 versions a day to 1, so the fan-out shrinks. Consecutive daily versions still overlap on the boundary day, so a Monday can still match 2. Do not read a drop in the duplicate count after #5374 merges as this issue being resolved.

**What fixed looks like.** The reproduce query returns `duplicated = 0` for both models, and both properties files carry a `dbt_utils.unique_combination_of_columns` test on `student_number`, `academic_year`, `week_start_monday`. Add the test in the same change as the fix, not after — the repo rule wants it regardless, and it is the regression guard.

**Pick the row deliberately.** `.claude/rules/dbt-sql.md` governs row picking. The question this needs answered first is which version a week should resolve to when several span it: the one in effect at the start of the week, or at the end. `int_powerschool__gpa_term_lookback` resolves "value in effect at the end of the day" against a local-midnight boundary instant, and matching that convention here would be consistent. Confirm with the metric owner before choosing, since it changes published weekly figures either way.

**Related files.**

- `src/dbt/kipptaf/models/topline/intermediate/int_topline__gpa_term_weekly.sql`
- `src/dbt/kipptaf/models/topline/intermediate/int_topline__gpa_cumulative_weekly.sql`
- `src/dbt/kipptaf/models/topline/intermediate/int_topline__student_metrics.sql` (the consumer, lines 30-41 and 425-437)
- Both properties files under `.../topline/intermediate/properties/`, which currently hold only a Dagster config block, no `data_tests:` and no `description:`.

**Sibling models are worth a sweep.** `int_topline__student_metrics` unions many `*_weekly` models on the same student-week grain. Only these 2 were checked. The others may have the same gap.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make `int_topline__gpa_term_weekly` and `int_topline__gpa_cumulative_weekly` return exactly one row per student-week.

Both models returned up to 3 rows for the same student in the same week, and the extra rows disagreed on the GPA. `int_topline__student_metrics` scores each row against the 3.00 bar, so a student with 3 rows counted 3 times against a student with 1 row. Every topline GPA figure built on those scores was wrong.

Two separate things caused it, and this change fixes both.

First, the join to the snapshot matched on a date range with `between`. Consecutive daily snapshot versions both cover the boundary day, so one Monday matched 2 versions. The join now builds a single instant at the close of the week and compares it half-open against the snapshot's raw timestamps. That is the same convention `int_powerschool__gpa_term_lookback` already uses. Both snapshots turned out to be gap-free and overlap-free in timestamp space, so exactly one version matches each instant. The `dbt_utils.deduplicate` step is no longer needed and is gone.

Second, the prod snapshot held rows captured from developer dev schemas (#4318). The region code is pulled out with a regular expression, and `zz_cbini_kippnewark_powerschool` yields `kippnewark` just like the real relation does, so those rows matched the join too. Those rows have since been purged from both prod snapshots, so no filter is needed. This change also deletes the filter `int_powerschool__gpa_term_lookback` was carrying for the same reason, and its `TODO(#4318)` comment with it.

Both properties files gain a uniqueness test on `student_number`, `academic_year`, `week_start_monday`, set to error severity. That is the guard that stops this coming back.

Closes #5375

## Reviewer Notes

**This changes published numbers.** The share of student-weeks at or above 3.00 moves from 55.88% to 59.65% on the term model, and from 51.97% to 55.19% on the cumulative model. The old figures were wrong, so this is a correction, not a regression. Worth a heads-up to anyone reading the topline dashboard this week.

**The week now resolves to its close, not its start.** A week reports the GPA in effect at the instant the week ends, so a change mid-week shows up that same week rather than the next one. That choice was confirmed before implementing. It is also what makes each week match exactly one snapshot version.

**Seven more weekly models have the same defect.** Filed separately as #5381. One of them, `int_topline__attendance_interventions_weekly`, returns up to 354 rows on a single student-week — worse than either GPA model. This pull request deliberately does not touch them.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — both properties files already existed and are updated here
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — no exposure change; these are intermediate models behind `int_topline__student_metrics`
- [x] **Breaking change?** No columns renamed or removed. The column set is unchanged; only the row set and the values change.
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — no external source touched

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

> Context, not instructions.

Claude-assisted, human-directed. The end-of-week anchor and the decision to scope this to the 2 models #5375 names were both the user's calls.

**Verification, measured against prod on 2026-09-17.**

Re-ran #5375's diagnostic before changing anything. Numbers had drifted up from the issue body: term 175,513 duplicated keys (issue said 173,205), cumulative 301,532 (issue said 299,399). Max 3 rows on a key in both.

Attributed the term model's 173,200 duplicated keys to cause:

| Cause | Duplicated keys |
| --- | --- |
| A ghost dev-schema snapshot row joins alongside prod | 121,678 |
| Several prod versions span the same Monday | 93,997 |
| Prod overlap alone, no ghost row involved | 51,522 |

The two causes overlap, which is why the columns do not sum. #5375's "For Claude" note named only the second cause, so the date-join fix alone would have left roughly 121k keys duplicated. The ghost relations hold 54,560 rows across `zz_cbini_kippnewark_powerschool`, `zz_cbini_kippcamden_powerschool` and `zz_cbini_kippmiami_powerschool`.

Confirmed the spine is not a contributor: `int_extracts__student_enrollments_weeks` filtered the same way returns 442,474 keys with 0 duplicates and max 1 row. All fan-out came from the snapshot join.

Proved the contiguity assumption before deleting the `deduplicate` CTE. On prod relations only, with `lead(dbt_valid_from)` per snapshot key:

| Snapshot | Prod rows | Identical `dbt_valid_from` | Overlaps next version | Gap before next |
| `snapshot_powerschool__gpa_term` | 4,081,415 | 0 | 0 | 0 |
| `snapshot_powerschool__gpa_cumulative` | 2,467,978 | 0 | 0 | 0 |

Perfectly contiguous, so a boundary instant matches exactly one version by construction and no tiebreaker is needed. Both snapshots also set `dbt_valid_to_current` to `9999-12-31`, so the open version has no NULL `dbt_valid_to` and the current in-progress week still resolves.

Ran the compiled SQL of both models through BigQuery against prod. Both return 442,474 rows on 442,474 student-week keys, 0 duplicated, max 1 row.

Confirmed both new tests bind at the intended severity by reading `target/manifest.json` after `dbt parse --no-partial-parse`. Both report `severity: error`. The kipptaf project default is `warn`, which would not fail CI — that override is deliberate, and `.claude/rules/dbt-yaml.md` permits it where a warning is not acceptable.

`trunk check --force --no-fix` on all 4 files reported only sqlfmt formatting, which the commit hook fixed; the post-commit re-check reported no issues.

**The ghost rows are purged and all 3 filters are gone.** The purge ran 2026-09-17. Both prod snapshots now hold only the 4 production district relations, at 4,081,415 rows on `snapshot_powerschool__gpa_term` and 2,467,978 on `snapshot_powerschool__gpa_cumulative` — exactly the production-only counts measured before the purge, so nothing collateral was removed. Re-ran both topline join predicates against prod with no relation filter at all: 442,407 student-week keys, 442,407 rows, 0 duplicated, max 1 row per key. The uniqueness tests added here are what now catch a re-injection. Also dropped the lookback unit test's ghost-relation fixture row, since the filter it asserted is gone; `dbt test --select test_type:unit,int_powerschool__gpa_term_lookback` passes. `trunk check --force --no-fix` reported no issues on all 6 files.

**Sibling sweep detail, for #5381.** Tested all 18 `int_topline__*_weekly` models at the grain `int_topline__student_metrics` actually reads them at. Five models with no discriminator column fan out: `attendance_interventions` (422,821 duplicated, max 354), `dibels_pm` (18,260, max 5), `college_matriculation` (173), `dibels_benchmark` (38), `truancy` (23). Five more have smaller residuals at their extended grain: `iready_diagnostic` and `iready_lessons` 11,178 each with `discipline`, `state_assessments` 684 with `discipline` + `test_round` + `region`, `college_entrance_exams` 183 with `test_type`, `formative_assessment` 46 with `discipline` + `formative_strategy`. Also `int_topline__star_assessment_weekly` holds 0 rows in prod, so the "Star Diagnostic" indicator is empty.

</details>


cbini:
## Review verdicts

**Finding 1 — `regexp_contains` applied to a column inside `WHERE`. Valid rule tension, no change.**

The rule violation is real, and the suggested remedy costs more than it buys. BigQuery rejects a `SELECT`-list alias referenced by another item in the same list, so filtering on a precomputed named column needs a third CTE that exists only to hold one boolean. `int_powerschool__gpa_term_lookback.sql:44-46` carries the identical inline `regexp_contains` — that is the model this change is deliberately matching, so diverging here would make the two read differently for no behavioral gain.

The filter is also temporary. Both sites carry `TODO(#4318)`, and #4318 is being worked now; once the ghost rows are purged from the prod snapshots the filter is deleted outright rather than refactored. Building a CTE around something scheduled for removal is the wrong direction.

**Finding 2 — no `config.meta.contains_pii` tag. Valid, and fixed in 2119816.**

Not out of scope, and the review was too generous in calling it so. `.claude/rules/ferpa-pii.md` says to widen tags on a file touched for another reason and not to sweep the repo. This change already edits both properties files, so they are exactly the files the rule points at.

Tagged column-level on `student_number` (tier 1) and the GPA column in each model (tier 3), matching the format in `rpt_focus__contacts.yml` and the amplify mClass staging models. Nothing in `src/` or `core/` reads `config.meta.contains_pii`, so this is metadata only and no behavior changes.

The wider gap stands and is untouched here: 0 of 24 `topline/` properties files carry a tag, and `int_powerschool__gpa_term_lookback.yml` does not either. That is a sweep, and the rule says not to do one.

**On the two things the review could not verify.** Both are now confirmed by CI rather than by my report. dbt Cloud run `70403225163781` on commit f0508c91e3 built both models at 442.4k rows each and both new uniqueness tests returned `pass` with 0 failures, per `run_results.json`. The run had 0 warnings. Trunk passed on the same commit, and `trunk check --force --no-fix` was clean locally on all 4 files.

cbini:
## Post-review change in 9d0e9a4 — the ghost filters are deleted

Heads-up for anyone who already read this: the diff changed after the review above.

The ghost rows are purged from both prod snapshots, so all 3 relation filters are gone — the 2 this branch added, plus the one `int_powerschool__gpa_term_lookback` already carried. That also resolves review finding 1 by deletion rather than by argument: there is no longer an inline `regexp_contains` in a `WHERE` to justify.

**The purge, verified.** Both snapshots now hold only the 4 production district relations, at 4,081,415 rows on `snapshot_powerschool__gpa_term` and 2,467,978 on `snapshot_powerschool__gpa_cumulative`. Those are exactly the production-only counts measured before the purge, so nothing collateral went with the 54,560 + 27,967 ghost rows.

**Removal is safe, not just tidy.** Re-ran both topline join predicates against prod with no relation filter at all: 442,407 student-week keys, 442,407 rows, 0 duplicated, max 1 row per key. The half-open interval is doing all the work; the filter was only ever masking the ghosts. The uniqueness tests this branch adds are what now catch a re-injection, at error severity.

Also dropped the lookback unit test's ghost-relation fixture row, since the filter it asserted no longer exists. `dbt test --select test_type:unit,int_powerschool__gpa_term_lookback` passes, and `trunk check --force --no-fix` is clean on all 6 files.

Net effect on the diff: 6 insertions, 43 deletions.

#4318 keeps the remaining work — the root cause of the 2025-12-03 write is only partly established. The exact invocation is not recoverable, since BigQuery's job history retains 180 days and that date is 288 days back. What is established: no recurrence in 288 days, a single injection date, and every other snapshot clean.

PR #5386, opened 2026-09-17
fix(dbt): dedupe int_extracts__student_enrollments_subjects_weeks

### LINKED ISSUE #5385: fix: int_extracts__student_enrollments_subjects_weeks repeats a student-week
## What's happening

`int_extracts__student_enrollments_subjects_weeks` should hold one row per student, week, and discipline. It holds up to 3. Measured against prod on 2026-09-17: 11,178 of 1,704,894 keys repeat.

A student who enrolls more than once in a year gets one row per enrollment stint. The model joins every stint to every calendar week of that school, and nothing picks a single stint afterward. All 292 repeated `(student_number, academic_year, discipline)` keys in the upstream `int_extracts__student_enrollments_subjects` differ by `rn_year` and `entrydate`.

Its sibling `int_extracts__student_enrollments_weeks` is built the same way and is clean, because it ends with a step this model does not have:

```sql
{{ dbt_utils.deduplicate(
    relation="student_week",
    partition_by="_dbt_source_project, student_number, week_start_monday",
    order_by="is_enrolled_week desc",
) }}
```

Five models read the repeated rows. Four of them are topline weekly models that `int_topline__student_metrics` scores one row at a time, so a repeated student counts 2 or 3 times against a student who counts once:

| Model | Repeated keys | Share caused by this spine |
| --- | --- | --- |
| `int_topline__iready_diagnostic_weekly` | 11,178 | 100% |
| `int_topline__iready_lessons_weekly` | 11,178 | 100% |
| `int_topline__state_assessments_weekly` | 684 | 100% |
| `int_topline__formative_assessment_weekly` | 46 | 10 of 46 |

The fifth is `rpt_tableau__iready_apm`, a Tableau extract.

The repeated rows agree on every metric column — 0 of the 11,178 carry conflicting values. So this changes how much each student weighs, never what a student's answer is. The published figures move a little: iReady proficiency goes from 22.18% to 22.22%, state assessment proficiency from 29.57% to 29.62%, formative mastery from 47.77% to 47.78%.

Neither this model nor its upstream has a uniqueness test, which is why nobody saw it. `.claude/rules/dbt-models.md` requires one on every intermediate model.

## Steps to reproduce

1. Run this against prod:

```sql
select
  count(*) as keys,
  countif(n > 1) as repeated,
  max(n) as most_rows_on_one_key
from (
  select student_number, academic_year, week_start_monday, discipline, count(*) as n
  from `teamster-332318.kipptaf_extracts.int_extracts__student_enrollments_subjects_weeks`
  where academic_year >= 2025
  group by student_number, academic_year, week_start_monday, discipline
)
```

2. Read `repeated`. It should be 0. It returns 11,178.
3. Run the same query against `int_extracts__student_enrollments_weeks`, grouping without `discipline`, to see the sibling return 0.

## Where

- **Environment:** prod
- **Run, PR, or dashboard link (if any):** found while fixing #5381

<details>
<summary>For Claude</summary>

> Context, not instructions.

**The fix is the sibling's `deduplicate`, partitioned one column wider.** Add `dbt_utils.deduplicate` to the end of `int_extracts__student_enrollments_subjects_weeks` with `partition_by="_dbt_source_project, student_number, week_start_monday, discipline"` and `order_by="is_enrolled_week desc"`, matching `int_extracts__student_enrollments_weeks`.

**Do not filter on `is_enrolled_week` instead.** It looks equivalent and is not. It resolves 9,652 of the 11,178, leaves 6 still repeated, and drops 1,520 keys whose week falls between two stints — those keys leave the population entirely, which a `deduplicate` preserves.

**Stay on the macro, not the ranked-column form.** `.claude/rules/dbt-sql.md` gates the ranked rewrite on the input exceeding about 1M rows AND the model costing at least 1 slot hour over 7 days. The first bar is met at 1.7M keys; the second is not, because this model is a view and bills no slot time of its own. The sibling already uses the macro.

**Check `rpt_tableau__iready_apm` before merging.** It is the one consumer outside the topline cluster, so the row-count change reaches a dashboard.

**Add the uniqueness test in the same change**, at `severity: error`. The kipptaf project default is `warn`, which would not fail CI.

**Related files.**

- `src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments_subjects_weeks.sql`
- `src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments_weeks.sql` (the clean sibling)
- `src/dbt/kipptaf/models/students/intermediate/properties/int_extracts__student_enrollments_subjects_weeks.yml` (no `data_tests:` block today)
- `src/dbt/kipptaf/models/extracts/tableau/rpt_tableau__iready_apm.sql`

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will stop `int_extracts__student_enrollments_subjects_weeks` from repeating a student-week.

The model joined every calendar week to every enrollment stint at the same academic year and school. Nothing in the join tied the week to the stint's date range, so a student with 2 stints in a year got 2 rows for every week of that year, including weeks neither stint covered. The model already computed an `is_enrolled_week` flag, but only as an output column — it never filtered on it.

This change constrains the join to weeks a stint really covers. It compares the stint's dates to the week's **in-session** bounds, so a stint that touches a week only over a weekend or a school break stops producing a row for it.

Measured at one moment against prod, this cuts the model from 6,998,084 rows to 6,650,588. The 346 student-weeks that still repeat are not fan-out — they are students who really were enrolled twice in one week, at 2 schools or at the same school across an exit and a re-entry. The uniqueness test now covers that real grain, with `schoolid` and `entrydate` added.

An earlier commit on this branch solved the same symptom with `dbt_utils.deduplicate`. This replaces it. The dedupe kept 1 row per student-week by picking the "best" stint, which hid the double enrollments instead of representing them.

## Reviewer Notes

**The join uses in-session week bounds, not Monday to Sunday.** `int_students__calendar_week` carries both. Using the in-session pair drops 802 more rows: weeks where a stint overlapped only a non-school day. That looked like the honest reading of "enrolled during this week" for a model that feeds weekly academic metrics.

**The uniqueness test runs at `warn`, not `error`.** The true-grain key is clean except for 82 student-weeks, and all 82 trace to 3 duplicate rows in the upstream `int_extracts__student_enrollments_subjects`, whose own uniqueness test is disabled. #5388 tracks that. The `TODO` in the properties file says to raise this test back to `error` when #5388 lands.

**`exitdate` is inclusive here, and the second commit on this branch corrects a wrong reading of it.** The first attempt treated it as exclusive and used `exitdate > week_start`, citing the comment at `int_students__enrollment_daily.sql:15-16`. That comment describes that model's own raw PowerSchool input, not what reaches this model. The data disagrees: of the 80,130 abutting stint pairs, 73,600 have `exitdate` one day *before* the next stint's `entrydate` and only 144 equal it. The predicate is `>=`, and the 2 flags are unchanged from `main`.

**The 2 iReady weekly models lose 92,160 rows, and 17,819 of those carry real lesson activity.** Both drive straight from this model and filter to the last 2 academic years, so they lose exactly what this model loses there. The 17,819 are students doing iReady lesson work in weeks no enrollment stint covers — an upstream enrollment gap this change makes visible rather than causes. No consumer counted rows of this model as a denominator, so no metric changes definition.

**The sibling model has the same unconstrained join, and it is not fixed here.** `int_extracts__student_enrollments_weeks` pairs weeks to stints the same way, and it *is* the topline enrollment denominator. Left alone to keep this PR to one model.

### General

### dbt

- [x] Include (or update) an [exposure](https://teamschools.github.io/teamster/reference/dbt-conventions/#exposures) for all models consumed by a dashboard, analysis, or application
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** so the dbt Cloud CI job can find the table. See [Staging external sources](https://teamschools.github.io/teamster/guides/dbt-development/#staging-external-sources)

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Every number here is measured against prod. This model and its prod relation are views, so prod row counts drift between measurements — each before/after pair below was taken in one query at one moment.

**Evidence that `exitdate` is the last enrolled day, not exclusive.** Two independent sources agree. The abutting-pair distribution: 73,600 of 80,130 pairs put `exitdate` 1 day before the next `entrydate`, 144 equal it, 2 exceed it, and the other 92,926 stint pairs have a real gap. And `int_students__student_enrollments.sql:166-170` compares dates as `between enr.startdate and enr.exitdate` and tests `exitdate >= sye.last_day_school_year`. So a week overlaps a stint when `entrydate <= week_end and exitdate >= week_start`, which is what the join now says, and the 2 flags stay as `main` wrote them.

The first commit on this branch had this backwards and shipped `exitdate > week_start` plus rewritten half-open flags. That dropped 1,744 rows it should have kept — weeks whose first in-session day is a stint's last day. `74ec4774e3` reverts the flags and fixes the predicate.

**Why the issue's original approach was not taken.** #5385 prescribed `dbt_utils.deduplicate` on `_dbt_source_project, student_number, week_start_monday, discipline` and explicitly rejected filtering on `is_enrolled_week`. The dedupe shipped in `1c81a8a275`, `f22e632157` and `af40d37f63`, and `2e4aeae90a` reverts it.

Filtering on `is_enrolled_week` was the wrong tool for a separate reason: the flag tests only the Monday, so it drops student-weeks where the student entered or exited mid-week and was genuinely enrolled. 19,740 rows the join keeps have `is_enrolled_week` false for exactly that reason. The join predicate tests overlap rather than a single day, so it keeps them.

**Monday-to-Sunday vs in-session bounds, measured in one query.** Overlap on `week_start_monday` / `week_end_sunday` keeps 6,651,390 rows; overlap on `school_week_start_date` / `school_week_end_date` keeps 6,650,588. The 802-row difference is stints that overlapped a week only on a weekend or a break day. 0 rows have null in-session bounds, so the in-session comparison never silently drops a week.

**What the 346 residual groups are.**

| driver | groups |
| --- | --- |
| mid-week school transfer, 2 distinct stints at 2 schools | 208 |
| upstream duplicate rows (#5388) | 82 |
| same-school exit and re-entry inside 1 week | 56 |

70 of the 346 are AY2025 or later, and 42 of those are AY2026.

**Why the widened key is the right grain, not a workaround.** Adding `schoolid` and `entrydate` takes the duplicate-group count from 346 to 82, and those 82 are exactly the upstream-duplicate set: every group the widened key cannot separate is one where 2 source rows are identical, so no sort key and no join predicate could ever have separated them. There is no arbitrary pick hiding in the residual.

**Where the dropped rows fall, and what that costs each consumer.** 347,496 rows drop in total: 255,336 in AY2024 and earlier, 92,160 in AY2025 and AY2026. The old years carry most of it because 19 years of multi-stint students accumulate there.

| consumer | rows after | change |
| --- | --- | --- |
| `int_topline__iready_lessons_weekly` | 1,624,234 | −92,160 |
| `int_topline__iready_diagnostic_weekly` | 1,624,234 | −92,160 |
| `rpt_tableau__iready_apm` | 1,297,368 | smaller, prod-table drift confounds the exact figure |
| `int_topline__state_assessments_weekly` | 74,628 | smaller, same caveat |
| `int_topline__formative_assessment_weekly` | 74,258 | about 0; it already filters `where sw.is_enrolled_week` at line 59 |
| `int_topline__star_assessment_weekly` | 0 | 0; it returns 0 rows in prod in every academic year |

The 2 iReady figures are exact: both models are `select ... from this_model ... where academic_year >= current-1`, so their loss equals this model's loss over those years. The other 3 apply further filters, and their prod relations are tables materialized at an earlier moment, so a dev-versus-prod count mixes this change with upstream drift. Do not quote a delta for them without rebuilding both variants.

`int_topline__star_assessment_weekly` returning 0 rows is a separate problem, not filed. It feeds the "Star Diagnostic" topline indicator.

**No consumer needs a row for a week the student was not enrolled.** 3 independent checks: `int_topline__formative_assessment_weekly.sql:59` already filters `where sw.is_enrolled_week`; `int_topline__student_metrics` builds "Total Enrollment" from `if(is_enrolled_week, 1, 0)` off the *sibling* model, not from row presence here; and no model counts rows of this model. An earlier draft of this PR called this model the topline denominator. That was wrong — `int_topline__student_metrics.sql:585-590` drives from `int_extracts__student_enrollments_weeks` and left-joins the metrics onto it.

**Verification.** `uv run dbt build --select int_extracts__student_enrollments_subjects_weeks int_topline__iready_lessons_weekly int_topline__iready_diagnostic_weekly int_topline__formative_assessment_weekly int_topline__state_assessments_weekly int_topline__star_assessment_weekly rpt_tableau__iready_apm --defer --favor-state --state <prod manifest>` returns `PASS=7 WARN=1 ERROR=0 SKIP=0`. The 1 warning is the uniqueness test at 82 results, as expected.

A note for a future build: the changed model must be inside `--select`, not only its consumers. `--favor-state` resolves any unselected upstream to prod, so a consumer-only build reads the prod weeks model and reports no change at all.

The pre-existing `rpt_tableau__iready_apm` warning about numeric columns with unspecified precision is unrelated to this change and fires on the base commit too.

</details>


cbini:
## Review verdicts

**Finding 1 — `enroll_status asc` is a no-op tiebreak column: ACCEPTED, fixed in `f22e632157`.**

The lineage trace is correct and the data agrees. Measured on prod: of the 40,634 duplicate groups, 82 still tie after `is_enrolled_week desc, rn_year asc`, and `enroll_status` discriminates 0 of them.

```sql
countif(n > 1)                                as dup_groups                        -- 40,634
countif(n > 1 and n > n_top_key)              as tied_after_is_enrolled_and_rn_year -- 82
countif(n > 1 and n_plus_status > n_top_key)  as status_breaks_a_tie                -- 0
```

`enroll_status` does vary inside 390 duplicate groups, so it is not strictly constant per student in this model — Miami rows reach it through Focus, not `stg_powerschool__students`. But in every one of those 390, `is_enrolled_week` or `rn_year` already discriminates first, so the column is dead regardless. Dropped it, and corrected the PR body so no future reader credits it.

No rebuild was needed: `status_breaks_a_tie = 0` is itself the proof that output is unchanged.

**Finding 2 — cost gate applies to the table-materialized consumer, not this view: ACCEPTED as informational, no change.**

Fair point that a view's cost lands on whoever executes it. It does not move this PR, because `rpt_tableau__iready_apm` is also a view, so neither relation bills slot time of its own. The dedupe input clears the ~1M row bar but the slot-hour bar has nothing to measure against. Worth revisiting if any consumer in this chain flips to `table`.

**Findings 3 to 6 — no action needed.** Agreed, including the ST03 reasoning: the single-CTE shape where the macro call is the final top-level statement does not trip it, and Trunk Check passed.

## Root-cause question, asked separately and answered in the body

A reviewer may reasonably ask whether the dedupe is the right layer at all, so the "For Claude" fold-out now records the measurement. The join pairs every calendar week with every stint at the same `academic_year` + `schoolid`; constraining it to real stint-week overlap resolves 34,478 of the 40,634 duplicate groups, but deletes 305,298 keys (4.4% of the model) that no stint overlaps. This model is a spine, so those keys are intended. The dedupe picks the overlapping stint anyway.

Separately: `int_extracts__student_enrollments_subjects` has 3 rows duplicated at its own test key, and its uniqueness test carries `enabled: false`. That accounts for the 82 identical-stint duplicate groups here. It needs its own issue.

PR #5389, opened 2026-09-17
fix(dbt): collapse duplicate previous-year state test rows

### LINKED ISSUE #5388: fix: int_extracts__student_enrollments_subjects has duplicate rows and its uniqueness test is off
## What's happening

`int_extracts__student_enrollments_subjects` returns 3 more rows than it has distinct enrollment-subject keys. Every duplicate is in academic year 2024. The model's `dbt_utils.unique_combination_of_columns` test carries `enabled: false`, so nothing catches it.

The duplicates fan out downstream. In `int_extracts__student_enrollments_subjects_weeks` they become 82 duplicate student-weeks that no correct join or sort can resolve, because the duplicated rows are identical on every column the grain uses.

## Steps to reproduce

1. Run the count below against prod.

```sql
select
    count(*) as rows,
    count(distinct format("%T|%T|%T|%T", student_number, academic_year, entrydate, iready_subject)) as distinct_keys,
from `teamster-332318.kipptaf_extracts.int_extracts__student_enrollments_subjects`
```

2. `rows` returns 267,301 and `distinct_keys` returns 267,298.
3. Group by the same 4 columns and filter to `count(*) > 1` to see the 3 affected keys. All 3 are `academic_year = 2024`.

## Where

- **Code location / dbt project:** `src/dbt/kipptaf/models/students/intermediate/int_extracts__student_enrollments_subjects.sql`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** found while fixing #5385 (PR #5386)

<details>
<summary>For Claude</summary>

The uniqueness test was switched off in `b3ecd428c6` ("refactor: mv student enrollments our of extracts folder"). Re-enabling it is the acceptance check, but it will fail until the 3 duplicate rows are gone, so find where they enter first.

The model has 21 consumers, 15 of them Tableau `rpt_*` views, so a change to its grain is not cheap. Check whether the duplicates come from a source join that gained a second matching row, rather than from this model's own logic.

`int_extracts__student_enrollments_subjects_weeks` carries a uniqueness test at `severity: warn` with a `TODO` pointing here. When this is fixed, raise that test back to `severity: error` — it should then pass with 0 failures.

Fixed means: the test on `int_extracts__student_enrollments_subjects` is re-enabled and passes, and the weeks model's widened-key test passes at `severity: error`.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will remove 3 duplicate rows from `int_extracts__student_enrollments_subjects` and turn its uniqueness test back on.

The NJGPA test runs twice a year, once in the fall and once in the spring. A student who sat both windows has two rows for the same subject and year in the previous-year state test lookup. That lookup is joined to the enrollment spine, so those students came out of the model twice. All 3 duplicates are Camden students in academic year 2024. The duplicates spread downstream into `int_extracts__student_enrollments_subjects_weeks`, where they became 82 duplicate student-weeks.

The fix filters the previous-year state test lookup to NJSLA, the only Pearson assessment that carries a proficiency value. NJGPA, NJSLA Science and PARCC all carry none, so they could never supply a result and only ever added join weight. That leaves one row per student, year and subject by construction, with no tie to break. The uniqueness test on the subjects model is enabled again.

## Reviewer Notes

The change drops rows from the lookup, so the thing to check is that none of them could have supplied a value. `njsla_aggregated_proficiency` is non-null on exactly the 44,533 NJSLA rows and null on all 27,485 NJGPA, NJSLA Science and PARCC rows, so every dropped row would have coalesced to `No Test` anyway. Confirmed against prod: `state_test_proficiency` is identical on all 267,298 shared keys. Details in the fold-out.

A second, separate change rides along. The weeks model joined `int_students__calendar_week` on `academic_year` and `schoolid` only, with no `_dbt_source_project`. Calendar week is unique on `schoolid, yearid, week_start_date, _dbt_source_project`, so it may legitimately hold the same school number for two regions and its own test would not object; both rows would then match and fan out every student-week for that school. No region collides today (0 of 184 `schoolid` and `academic_year` keys span more than one project), and the join returns 6,650,462 rows with and without the predicate, so this changes no output. It closes the path rather than fixing a live duplicate.

A third change removes the uniqueness test on the weeks model. Issue #5388 asks for that test at `severity: error`, so this PR deliberately diverges from it. The reason is that the test cannot fail on its own: `discipline` is a 1-to-1 relabel of `iready_subject`, so the weeks key is the subjects key plus `schoolid` and `week_start_monday`, and both parents are uniqueness-tested at `severity: error`. Once the calendar join is keyed on `_dbt_source_project`, every remaining fan-out path fails a parent's test before it reaches this model. The reasoning moves into the model `description` so the absence does not read as an oversight. If you would rather keep the test as belt-and-braces, say so and I will put it back — it is a view, so the cost is a CI scan, not a nightly one.

### General

### dbt

- [x] Include (or update) an
      summary above.
      See

No columns change and no external source changes. The model keeps its full column list, so its 21 consumers and 15 Tableau views are unaffected.

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

**Root cause.** `prev_yr_state_test` unions Pearson and FLDOE. The Pearson branch reads `int_pearson__all_assessments`, which carries one row per administration window. NJGPA has a Fall and a Spring window, so a retaker has 2 rows sharing `(statestudentidentifier, academic_year, subject, _dbt_source_project)` — the exact key the `py` left join matches on. The model projects only `py.njsla_proficiency`, which is NULL on every NJGPA row, so the fan-out produced byte-identical duplicate rows rather than differing ones. That is why no downstream join or sort could resolve them.

**Blast radius, measured against prod.**

- Pearson branch: 3 keys with more than 1 row, network-wide, across all years. 0 of them have more than 1 distinct non-null `njsla_aggregated_proficiency`.
- FLDOE branch: 0 keys with more than 1 row. It is already narrowed by `assessment_name = 'FAST'` and `administration_window = 'PM3'`.
- Every other join in the model contributed 0 fan-out. The model had exactly 3 duplicate keys in total and all 3 trace to the Pearson branch, so nothing else needed touching.

**Why filter to NJSLA rather than pin the administration window.** Pinning to Spring also removes the fan-out today, but only incidentally: it works because NJGPA happens to be the sole assessment with a Fall window, so any future Fall administration re-breaks it. It would also keep 26,388 Spring rows that carry no proficiency and can never supply a value. Filtering to the assessment that supplies the value states the actual rule, mirrors the FLDOE branch in the same CTE, and removes the aggregate entirely — so the uniqueness test now fails loudly if the grain ever breaks, instead of a silent pick masking it.

| Pearson branch variant | Keys | Fanned-out keys |
| --- | ---: | ---: |
| No filter (pre-fix) | 72,018 | 3 |
| Spring only | 70,921 | 0 |
| `assessment_name = 'NJSLA'` | 44,533 | 0 |
| `njsla_aggregated_proficiency is not null` | 44,533 | 0 |

The last two are the same 44,533 rows, which is the evidence that NJSLA is exactly the proficiency-carrying set.

**Verification.** `dbt build --target dev --defer --favor-state` on both models: PASS=4, with both uniqueness tests still in place at `severity: error` at the time of that run. The dev table is 267,306 rows over 267,306 distinct keys — one row per key.

Output identity against prod, per key:

| Measure | Result |
| --- | ---: |
| Matched keys | 267,298 |
| `state_test_proficiency` differences | 0 |
| Keys only in dev | 8 |
| Keys only in prod | 0 |

The same comparison at CTE level, `max()` against the NJSLA filter across all 72,018 Pearson keys, also returns 0 differences. `max()` over an all-null group returns null and an absent row coalesces to `No Test`, so the two are provably equivalent.

Two numbers that look like regressions and are not. The 8 dev-only keys are spine drift — `int_extracts__student_enrollments` gained enrollments since prod last built. And `bucket_one` differs on 30 keys, all of them grade 3 and below, which is the `pr.iready_proficiency` branch of that `if()`; `iready_proficiency_eoy` itself differs on 101 keys from the same upstream drift. Zero `bucket_one` differences above grade 3, which is the branch `py` feeds.

Closes #5388

</details>


cbini:
## Review verdicts

One verdict per finding, in the order posted.

| # | Finding | Verdict |
| --- | --- | --- |
| 1 | Altitude check | Agreed, no change |
| 2 | Dedup pattern | Agreed, no change |
| 3 | Column order / structure | Agreed, no change |
| 4 | Narrowed CTE, no regression | Agreed, no change |
| 5 | Inline comment placement | Declined, keeping the comment |
| 6 | No dbt unit test for the collapse | Declined |
| 7 | Test-severity changes | Agreed, no change |
| 8 | Formatting not independently verified | Resolved, Trunk is green |

### 5. Inline comment placement — declined

The comment stays. `.claude/rules/dbt-sql.md` sends rationale to the properties `description:` but carves out "what a reader of that exact line cannot see — a non-obvious fallback". A reader of `max(p.njsla_proficiency)` cannot see that NJGPA runs two administration windows and carries no NJSLA proficiency, which is the whole reason the aggregate is safe. That is the carve-out, not the rule it excepts.

### 6. No dbt unit test — declined

The suggestion is reasonable in the abstract and expensive here. This model projects `co.*` from the enrollment spine, so its output is 115 columns. A dbt unit test builds `expect` rows as a `UNION ALL` and does not null-fill omitted keys, so every `expect` row must list all 115 columns. Controlling which rows reach the join also means mocking the spine itself, another wide fixture. Two fixtures of that size to pin one `max()` is a poor trade, and it is why the three existing unit tests in this directory sit on narrow models.

What covers the logic instead: the reinstated uniqueness tests at `severity: error` on both models fail the build the moment the collapse stops being one row per key. They do not pin `max` over `min`, and I am not claiming they do — what makes that choice safe is measured, not tested: across every year in the warehouse, no student, year and subject has two different non-null proficiency values, so `max` and `min` return the same answer for every key today. The ceiling if that ever stops holding is already written down in the "For Claude" fold-out.

### 8. Formatting — resolved

Trunk could not run in the review job. It ran here: `trunk check --force --no-fix` on all 3 changed files reported "No issues", and the PR's own Trunk Check and Trunk Check Runner checks are both green.

PR #5408, opened 2026-09-18
feat(amplify): add aimline_round_category for whole-round reporting

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will align the aimline progress-monitoring model to the internal one, so a single pair of Tableau selectors can drive every PM view instead of each method needing its own bespoke tab.

It started as one column, `aimline_round_category`, a round-level rollup of the aimline reporting category. It grew, because once the round-level question was answerable for aimline it became obvious that the two methods answered the same questions under different column names, different value vocabularies, and at different grains. A dashboard built on that needs a separate tab per method forever.

So the branch does three things:

**Aligns the two methods' published columns.** Same names, same grains, same shape. Where the values differ they differ on purpose: internal reads `Met` / `Not Met`, aimline reads `Meeting Aimline` / `Below Aimline` / `No Aimline Data`, because a reader of an aimline view should see what the verdict is measured against. The one exception is the benchmark standard, which reads identically on both methods, because the grade-level bar is the one thing here that does not depend on the method.

**Fills the granularity x comparison-item grid.** Five new columns so a Tableau parameter pair can pick measure standard / measure name code / round against own goal / benchmark / aimline+benchmark / trajectory.

**Gives the aimline method a visible target.** It had none on the dashboard — a reader could see a verdict but not the number behind it. It now has two, in separate columns, because the season endpoint and the moving per-probe target are different quantities.

## Breaking changes

**Yes. The Tableau workbook must be rebound before the extract refreshes**, or it will break silently while the extract itself builds clean.

Removed from the extract:

- `met_aimline_goal` — was already aliased to `met_measure_standard_goal` in the extract's own SELECT, so the two methods already shared one output name. The rename now happens upstream instead, so one Tableau field binds cleanly.
- `aimline_status` — still flows through the intermediates and is readable there; it is only dropped from the extract projection, because `measure_standard_goal_status` already carries the same verdict in academics' wording.

Renamed:

- `is_current_round` (boolean) -> `expected_round_selection` (string). Reads `Current` on each region's latest opened round and that round's label everywhere else, so one filter selection follows each region to wherever it actually is.

Changed values on live columns:

- `admin_benchmark_goal_status` now reads `Met Benchmark` / `Did Not Meet Benchmark`.
- The Aimline branch of `measure_standard_goal_status`, `measure_name_code_goal_status` and `pm_round_status` now reads `Meeting Aimline` / `Below Aimline` / `No Aimline Data`. Internal keeps `Met` / `Not Met`.

Added: roughly fifteen columns, listed in the extract's properties file.

## Reviewer Notes

- **`goal` now means the moving target on both methods.** `cumulative_growth_words` on Internal rows, `aimline_value_by_date` on Aimline rows, null on Benchmark. Both climb across a season and both are what their own method's verdict is computed against, so one column answers the same question either way. The season endpoint is a genuinely different quantity and stays separate in `aimline_season_student_goal`, with `aimline_season_student_goal_gap` beside it. This is the load-bearing judgment in the branch.
- **`aimline_value_by_date` was on hold and is not any more.** Academics were waiting on a definition; Amplify's own report documentation supplies one — the score that is on the aimline on the day the PM test is administered — and our data conforms to its stated ranges exactly. On AY2025, `measure_standard_score >= goal` reproduces `measure_standard_goal_status` on 32,170 of 32,170 scored Aimline rows carrying a target, zero disagreements either way.
- **`measure_standard_round_verdicts`** puts a student's whole season on every row as `B-B-A`: `A` at or above, `B` below, `?` no aimline data, `.` not tested. Scoped to the administration season and built over the expectation spine, so a skipped round shows as `.` rather than shortening the string.
- **Naming debt, deliberately incurred.** The five grid columns use `<grain>_<lens>_status`; four older ones do not. Aligning all nine was considered and deferred, because three of the older names are shared with the internal method and renaming them forces an unrelated workbook rebind mid-build. The reasoning is recorded in the reference doc.
- **One ordering choice still with academics.** Below Aimline ranks above No Aimline Data in the round rollup. Everything else in the precedence was already fixed by the per-measure column. It changes reported numbers if academics disagree.

### General

- [x] Review the **Claude Code Review** comments posted on this PR — three passes, all findings applied or answered with reasoning.

### dbt

- [x] Properties files updated for every new and changed column, in both the intermediates and the extract.
- [x] Exposure — the dashboard already has one; column changes do not alter it.
- [x] **Breaking change?** Yes — see the section above. Workbook rebinding required before the extract refreshes.

## CI checks

- [x] **Trunk** — passes.
- [x] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

**Verification method.** Every change was verified against prod before the next one started: union-branch balance, contract column count, a deferred dev build with the full upstream selection, and a dev-vs-prod value comparison per `model_type`. Where a change should have moved nothing, it moved nothing. Final state is 22 nodes building and testing clean at 125 contract columns per branch across all three UNION branches.

**Two verification failures worth recording, because both were mine and both were the same shape.**

The first: a check on the round-verdict string partitioned without `model_type`, merging Internal with Aimline and doubling every partition. It reported thousands of false mismatches. That is the exact double-count trap this model's own rules warn about.

The second, and worse: the verdict string shipped with its aimline half matching on `measure_standard_goal_status like 'Met%'`, which `Meeting Aimline` does not satisfy — `Mee`, not `Met`. All 13,886 met-aimline rows rendered as `?`. The check that passed it re-applied the same CASE to derive the expected value, so it compared the expression to itself and could not fail. Caught by review, not by me. The fix drives the token off `met_measure_standard_goal` directly, and `rpt_tableau__dibels_dashboard__round_verdict_token_reconciles` now derives the expected token from the flag instead; it was confirmed to reproduce the failure at exactly 13,886 rows before being accepted as green.

**Documented figures were re-measured against prod for this PR**, not against a dev build. Twelve stale numbers dating from 2026-09-15 are corrected here. The original error was believing a dev build authoritative: the whole aimline chain is views, so prod computes live and was never stale, while `--favor-state` does not defer a model that already exists in the dev schema, so a stale `zz_<user>_*` copy silently won.

**Known and not fixed here:** internal `pm_round_status` holds both `Met` and `Not Met` within 40 AY2025 round groups, unexplained by `pm_goal_criteria` or `completed_test_round`. Prod returns the same 40, so it is pre-existing rather than introduced. Recorded in the reference doc with a starting point, rather than left in a comment thread.

**Deferred on purpose:** the verdict-string CASE is duplicated across the two PM branches, and review argued for a macro. Not done — two call sites in one model, and the new reconciliation test makes drift loud rather than silent. Worth revisiting if it bites again.

</details>


GabyRangelB:
@claude Please review this branch end to end. It grew well past its original scope today, so a fresh pair of eyes on the whole diff would help more than a per-commit look.

**What it started as:** adding `aimline_round_category`, a round-level rollup of the aimline reporting category.

**What it became:** aligning the aimline PM model to the internal one so a single Tableau selector pair can drive every progress-monitoring view. Seven commits, and the later ones change published column names and values.

Areas where I would most value scrutiny:

1. **Column removals.** `met_aimline_goal` was a byte-identical duplicate of `met_measure_standard_goal` and is gone; `aimline_status` is no longer published to the extract (it stays upstream, where the verdict is translated from it). Both were verified as duplicates before removal, but these are breaking changes to a published extract and worth a second opinion.

2. **Value renames on live columns.** `admin_benchmark_goal_status` now reads `Met Benchmark` / `Did Not Meet Benchmark`; the Aimline branch of `measure_standard_goal_status`, `measure_name_code_goal_status` and `pm_round_status` now reads `Meeting Aimline` / `Below Aimline` / `No Aimline Data`. Internal keeps `Met` / `Not Met`. The divergence is deliberate — please check the reasoning holds rather than just that it is consistent.

3. **The rollup windows in the five new grid columns.** Partition keys are the thing most likely to be subtly wrong. I verified the coarser grains are strictly stricter (8,731 → 5,197 → 3,004 met, zero inversions), but that test would not catch a partition that is wrong in a way which happens to preserve monotonicity.

4. **The new warn test** `rpt_tableau__dibels_dashboard__measure_code_sat_all_or_none`. It guards an assumption the code-grain rollup depends on — that a student sits every measure standard under a code or none. Please sanity-check the assumption, not just the test.

5. **Naming debt, deliberately incurred.** The five new columns use `<grain>_<lens>_status`; four older ones do not. Aligning all nine was considered and deferred because three names are shared with the internal method and renaming forces a workbook rebind mid-build. Reasoning is in the reference doc — push back if you think the debt is not worth it.

Documentation lives in `docs/models/dibels-dashboard-data-model.md` and `.claude/skills/dibels-dashboard/SKILL.md`; both were updated alongside each change rather than at the end, and several sections record decisions that are deliberately NOT what T&L's guidance document says. Those are flagged as do-not-correct with the reasoning attached — if any of them read as drift rather than decisions, say so, because that means the note is not doing its job.

Every change was verified against prod before the next: union-branch balance, contract column count, deferred dev build, and a dev-vs-prod value comparison per `model_type`. Where a change should have moved nothing, it moved nothing.

Not yet done and known: the Tableau workbook still binds to the removed columns, so it needs rebinding before the extract refreshes.

GabyRangelB:
@claude Your end-to-end review is applied and the branch has grown again since. Please take another pass, weighted toward the new work rather than re-reading what you already cleared.

## Your findings, resolved

**1. PII tags — fixed.** All six columns now carry `contains_pii: true`: `aimline_round_category`, `measure_name_code_benchmark_status`, `round_benchmark_status`, `measure_name_code_aimline_benchmark_status`, `measure_name_code_trajectory_status`, `round_trajectory_status`. You were right that the closest analog being tagged is the test that matters, not the file's overall inconsistency.

**2. Seven undocumented intermediate columns — fixed.** Two on `int_amplify__pm_met_criteria.yml` and five on `int_amplify__pm_met_criteria_aimline.yml` now carry descriptions.

**3. ST06 column order — fixed, but your diagnosis was half right.** The interleaving is real (`if`, `if`, `case`, `if`, `if`, `case`) and I regrouped it. But it is **not** a lint failure: `trunk check --force` on that exact file passes clean both before and after. So the convention in `.claude/rules/dbt-sql.md` was the only thing holding here, not sqlfluff. Worth knowing for future reviews of this repo — ST06 does not appear to fire on this file, likely the templated-slice exemption.

**5. Broadcast-consistency check — ran it, and it found something.** You flagged this as optional and noted you had no warehouse access. I did, so I ran your query shape against the dev build. `aimline_round_category`, `round_benchmark_status` and `round_trajectory_status` are each single-valued across every `(academic_year, student_number, admin_season, round_number)` group — zero violations. But the internal `pm_round_status` returns **40 groups holding both `Met` and `Not Met` in the same round**, and neither `pm_goal_criteria` nor `completed_test_round` varies within any of them, so the obvious explanations do not cover it. Prod returns the same 40, so it is pre-existing and not introduced by this branch — I am flagging it rather than fixing it here. If you have a read on the mechanism from the SQL alone, I would take it.

Finding 4 (the "Below Aimline outranks No Aimline Data" precedence) is still with academics, unchanged.

## What is new since your review

Four commits. The first three are one arc: the aimline method had no visible target on the dashboard, and now it has two, in separate columns.

- **`aimline_season_student_goal`** — Amplify's per-student season endpoint, renamed from the vendor's bare `goal` at `int_amplify__mclass__pm_student_summary_aimline` and carried down. Renamed because `goal` says nothing about whose goal or over what horizon, and the extract's `goal` already meant the internal method's `cumulative_growth_words`. Published with `aimline_season_student_goal_gap` (score minus that goal) beside it.
- **`aimline_value_by_date` now fills the extract's `goal` column on Aimline rows**, opposite `cumulative_growth_words` on Internal rows. Both are moving targets that climb across a season and both are what their own method's verdict is computed against, so one column answers the same question for either method. This discharges the hold the docs recorded: academics were waiting on a definition and Amplify's report documentation supplies one.
- **`measure_standard_round_verdicts`** — the whole season on one row as a hyphen-separated string in round order, e.g. `B-B-A`. `A` at or above, `B` below, `?` no aimline data, `.` not tested.

## What I would most like you to check

1. **The shared `goal` column.** Internal gets `cumulative_growth_words`, Aimline gets `aimline_value_by_date`, Benchmark gets null. My argument is that both are the moving target their method's verdict is computed against, so merging them into one column is alignment rather than conflation — and that the season endpoint is a genuinely different quantity that therefore stays in its own column. That is the load-bearing judgment in this batch. If a single column whose meaning depends on `model_type` is a trap rather than a convenience, say so now, because the workbook will bind to it.

2. **`measure_standard_round_verdicts` is built with an analytic `string_agg` over a window**, in the two PM branches of the extract, partitioned by `(academic_year, region, student_number, expected_measure_standard, admin_season)` and ordered by round. Two things to scrutinize: whether that partition is right (your point that a partition can be wrong in a way that preserves monotonicity applies here too), and whether an analytic `string_agg` is acceptable in this repo at all. I confirmed BigQuery orders it by the window `ORDER BY` with an empirical test rather than trusting the docs, but the repo has no other instance of this construct that I found.

3. **Where the verdict string is built.** It is computed in the extract rather than upstream, which duplicates the status CASE in two branches. I chose that deliberately: the string must cover rounds the student did not sit, and those rows exist only on the extract's expectation spine, not in the criteria models. Building it upstream would have silently shortened the string and hidden the skip. Please check that reasoning rather than just the code — if there is a clean upstream home for it I missed, it would remove the duplication.

4. **A separator collision I caught late.** The not-tested token was originally `-`, the same character as the separator, so `B-B--` was unparseable. It is now `.`. Please check no other token can collide and that nothing downstream in the docs still shows the old alphabet.

5. **The `A`/`B` alphabet across two vocabularies.** Internal says Met/Not Met, Aimline says Meeting Aimline/Below Aimline, and both collapse to `A`/`B` in this one column because that is the pair Amplify's own reports use. This is the same deliberate-divergence question as the value renames you cleared earlier, one layer down — please check the reasoning holds rather than just that it is applied consistently.

## Verification already done

Every change built with `--defer --favor-state` against the prod manifest, 21 nodes including tests, clean. Contract column count matches the SQL at 125 per branch across all three branches. On AY2025 in dev: `measure_standard_score >= goal` reproduces `measure_standard_goal_status` on 32,170 of 32,170 scored Aimline rows carrying a target, zero disagreements; the character at each row's own round position in the verdict string equals that row's own status on all 89,730 PM rows, zero mismatches.

One caveat on my own verification worth recording: my first pass at that last check reported thousands of mismatches and the column looked broken. The query was wrong — I had left `model_type` out of the partition and merged Internal with Aimline, which is the double-count trap this model's own rules warn about. The column was correct. If you spot me making that mistake anywhere else in this branch, flag it.

Still outstanding and unchanged: the Tableau workbook binds to removed and renamed columns and needs rebinding before the extract refreshes.

GabyRangelB:
Fixed in `b3a911ad4`. You were right on every part of it, including the part about my verification.

## The bug

Confirmed before touching anything: `measure_standard_goal_status like 'Met%'` matches `Meeting Aimline` **0 times out of 13,886** on AY2025. Those rows were all rendering `?`, mixed in with the 3,312 rows where Amplify genuinely published nothing. On the aimline half, `A` could not occur.

Your read on the cause is right too — the `like 'Met%'` predates the `M` → `A` rename. `204a0ed37` changed the output character and left the condition that selects it, so the token's meaning and its selector drifted apart.

**Fix applied as you suggested**: both copies now branch on `met_measure_standard_goal` (`1` / `0` / `null`) instead of string-matching the status column. I checked first that the flag maps cleanly on both methods before relying on it — no row carries a null flag behind a real verdict on either method, so the "token equals this row's own status" property the column promises still holds by construction rather than by coincidence.

After the fix: aimline rows containing an `A` go 0 → 21,056, and rows containing a `?` fall to 3,612, which is the genuine no-value population.

## On your point about my verification

You called it exactly: my check re-applied the same `like 'Met%'` mapping to derive the expected token, so it compared the CASE to itself and passed unconditionally. It was not a weak check, it was a check of nothing. Second time this branch that I have verified something against itself or against the wrong partition, which is a pattern rather than an accident.

Two things came out of that rather than just a fix:

**A test that would have caught it.** `rpt_tableau__dibels_dashboard__round_verdict_token_reconciles` compares each row's character against a token derived from the **flag**, not from the string-building CASE. Before accepting it as green I ran it against the unfixed view and confirmed it reports exactly 13,886 failures, all Aimline, all `A` expected against `?` found. A test I had not seen fail would have been worth nothing here, given the circumstances.

**The trap is now in the skill**, named as a rule: never verify a derived column by re-applying its own derivation, and prefer flags over string prefixes anywhere the two methods' vocabularies diverge.

## Your other items

**`pm_round_status` anomaly — moved into the reference doc**, as you asked. You are right that a finding living only in a PR thread is lost. It now records the 40 groups, that `pm_goal_criteria` and `completed_test_round` are both ruled out, that prod returns the same 40 so it is pre-existing, that the aimline siblings are clean, and a starting point for whoever picks it up (the `met_pm_round_criteria` window partitions, not the CASE).

**Centralizing the token mapping** — taking the point, not acting on it yet. You are right that identical-and-identically-wrong pasted twice is what let this happen. I have kept the duplication for now because the reconciliation test makes the drift loud rather than silent, and a macro for two call sites in one model is a judgment call I would rather not make unilaterally on a branch this size. Flagging it for the author as a follow-up rather than deciding it here.

**Checks 1-4** — noted, no action. Useful that you independently confirmed the cohort gate pins the band before the window runs; that was the exact question the author raised separately, and two independent reads agreeing on it is worth more than one.

## One correction

Your environment note about `.claude/skills/dibels-dashboard/SKILL.md` being uncommitted-modified back to `origin/main`'s shorter version does not reproduce in the working checkout: `git diff HEAD -- .claude/skills/dibels-dashboard/SKILL.md` is empty, and the file on disk carries this PR's full set of updates. Whatever you observed looks local to the CI checkout rather than to the branch. Worth knowing in case it affects what you were reading elsewhere in that pass.

## State now

22 nodes build and test clean, including the new test. The alphabet is `A` at or above, `B` below, `?` no aimline data, `.` not tested, and it is now actually what the code produces on both halves.

GabyRangelB:
@claude You found a real one. Fixed in `b3a911ad4` — details in the comment directly above this, summarised here so you have it in context.

**The `A` token bug: confirmed and fixed.** `like 'Met%'` matched `Meeting Aimline` 0 times out of 13,886 on AY2025. Both copies of the CASE now branch on `met_measure_standard_goal` (`1`/`0`/`null`) as you suggested, rather than string-matching the status column. I verified first that no row on either method carries a null flag behind a real verdict, so the token-equals-this-row's-status property holds by construction. Aimline rows containing an `A` went 0 → 21,056; `?` fell to 3,612, the genuine no-value population.

**Your point about my verification was the more valuable half.** The check did re-apply the same broken mapping, so it compared the expression to itself and passed unconditionally. Added `rpt_tableau__dibels_dashboard__round_verdict_token_reconciles`, which derives the expected token from the flag instead, and I ran it against the unfixed view to confirm it reports exactly 13,886 failures before accepting it as green. The trap is now written into the skill as a rule.

**`pm_round_status` anomaly** moved from this thread into `docs/models/dibels-dashboard-data-model.md`, as you asked.

**One correction:** your note about `.claude/skills/dibels-dashboard/SKILL.md` being uncommitted-modified back to origin/main's shorter version does not reproduce here — `git diff HEAD` on that path is empty and the file carries this PR's full updates. Looks local to the CI checkout.

## What I'd like re-checked

1. **The flag-driven CASE itself**, in both PM branches of `rpt_tableau__dibels_dashboard.sql`. This is the code path you just caught, rewritten — please confirm the replacement is right rather than differently wrong. Specifically whether `pm.measure_standard_goal_status is null` is a sound test for "no criteria row at all, therefore not tested", given the extract coalesces that same column to `Not Tested` two lines away.

2. **The new test.** Does it actually test something independent, or have I fooled myself a second time? It derives the expected token from `met_measure_standard_goal` plus the coalesced status, while the column is built from the raw flag — I believe that is genuinely independent, but that is exactly the belief I held last time.

3. **Whether the duplication should become a macro.** You argued the pasted-twice CASE is what let this happen, and I agree. I have not done it — two call sites in one model felt like the author's call, not mine, and the new test makes the drift loud. If you think the test is insufficient mitigation and the macro should land before merge, say so plainly and I will do it.

4. **Anything else on this branch resting on a string prefix or a self-referential check.** Given I made the same class of mistake twice — the `model_type` partition miss and this one — a targeted sweep for that pattern would be worth more than another broad pass.

GabyRangelB:
@claude Your adjacent finding is real in the code path and empty in the data. Measured it, since you flagged the no-warehouse caveat — `d5b29d71f`.

**The conflation exists exactly as you describe.** The intermediate emits an explicit null on `measure_name_code_benchmark_status` / `round_benchmark_status` when a scored measure has no published benchmark goal, and the extract's `coalesce(..., 'Not Tested')` maps that to the same string a LEFT JOIN miss produces. Two real-world states, one label, on columns whose stated domain is three values.

**But it fires on nothing today.** AY2025: zero rows on either column in either method where a scored row reads `Not Tested`, and zero nulls on either column in either intermediate. Your estimate of ~15-30 rows does not hold — it is 0, because the population it depends on is the Miami `benchmark_goal` nulls and Miami produces no rows in this extract at all. So it is latent rather than live, which I think strengthens your "flag it, don't fix it here" call rather than weakening it.

**Documented, not fixed**, and the note says why: distinguishing the two states means a fourth value on a live column, which is a breaking change to a workbook that is about to be built, bought for a population of zero rows. Both column descriptions now name the two states, record the zero measurement, and say the honest fix when Miami appears is a fourth value rather than a wider `Not Tested`.

**On check 2 — you are right and I overstated it.** I wrote "independently is the point" without saying independent of what. Your distinction is the correct one: the test proves the window mechanics, that `string_agg` places each round's token at that round's position under the right partition and order. It does not independently prove the flag-to-token mapping, because it reads the same flag column the extract's CASE reads, so two copies changed identically and wrongly would still pass. I have rewritten the test's own description to say that, since a test whose reason for existing is that an earlier check overstated itself is the last place to overstate.

**Checks 1, 3 and 4 — closed, no action.** Your reasoning on `is null` being a sound not-tested probe matches what I found independently: neither source model can emit a null status on a row that exists, so a null can only be a join miss. The author has also decided against the macro, on your reasoning plus the test being adequate mitigation, so that is settled rather than deferred. And thank you for actually running the prefix sweep — `aimline_category like 'No Aimline Data%'` being safe because both real values are authored two lines above in the same CTE is the right kind of check, and not one I had done.

No open questions back to you on this pass. The remaining items on this PR are a product decision with academics (the Below-Aimline precedence, now documented in four places with its cost measured at 694 round groups) and a Tableau workbook rebind, neither of which is code.

PR #5407, opened 2026-09-18
fix(cambium): integrate the NJSLA district summative record file

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will put Cambium NJSLA and NJSLA Science scores
on the dashboard for Newark, Camden, and Paterson.

Cambium TIDE sent the first NJSLA District Summative Record File on 2026-09-18.
Two things stopped it. The Dagster ingestion captured 31 of the file's 225
columns, and no dbt model read the feed at all. So the scores reached nothing.

This pull request fixes both. On the ingestion side it gives the NJSLA feed the
full 225-field schema, narrows the Newark and Camden filename patterns so the
NJSLA file no longer competes with the Science file, and adds Cambium to
Paterson for the first time. On the dbt side it adds `stg_cambium__njsla`, which
splits the one file into the NJSLA and NJSLA Science assessments and maps both
into the shared New Jersey assessment shape. It then unions the 3 regions at
kipptaf and adds them to `int_pearson__all_assessments`, which is the model the
dashboard reads.

Paterson needs no student id translation. Cambium sends Paterson the KIPP
`student_number`, where Pearson sends the district SIS id. All 252 delivered ids
match `stg_powerschool__students.student_number` and none match `prevstudentid`,
so Paterson reads the package staging model directly.

## Reviewer Notes

3 calls worth a second look, with full reasoning in "For Claude" below.

- `assessment_version` repeats `assessment_name` instead of naming Cambium. This
  keeps the dashboard's existing values working.
- The `period` accepted values list is `[Spring]` only. That is tighter than the
  NJGPA model next to it, which also allows Fall.
- The model does not filter `test_status`. 1 invalidated row and 1 pending row
  survive, so they count toward proficiency. The NJGPA and Pearson models behave
  the same way, so this is not new.

2 more things to know.

- Paterson imports the cambium package with `stg_cambium__njgpa` disabled, and
  its 9 generic tests disabled one by one. A disabled model keeps its tests, and
  they would scan a relation Paterson never builds.
- I patched 3 rows of the automations catalog by hand instead of regenerating
  it. `gen-automations-doc.py` cannot load kippmiami in a codespace, so a full
  regeneration drops all 12 kippmiami automations.

### General

> If this is a same-day request, please flag that in the #data-team Slack

- [ ] Update **due date** and **assignee** on the
      [TEAMster Asana Project](https://app.asana.com/0/1205971774138578/1205971926225838)
[...]

PR #5412, opened 2026-09-18
refactor(cambium): remove the njsla_science placeholder assets

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will remove the two `cambium/njsla_science` assets that watch a folder Cambium never fills.

Cambium sends ELA, Mathematics and Science rows in the one NJSLA file. `stg_cambium__njsla` already splits the subjects itself. So the `njsla_science` assets in Newark and Camden match nothing and will keep matching nothing. This pull request drops both assets, their Couchdrop sensor wiring, and the two rows that named them in the automations catalog.

Pearson's `njsla_science` feed is untouched. It holds the 2022 through 2025 Science history that still flows through `int_pearson__all_assessments` to the state assessment dashboard.

## Reviewer Notes

2 things worth a second look.

- This fixes 2 tests that already failed on `main`. `NJSLA_TAILS` in `tests/libraries/test_cambium_assets.py` listed the permissive tails (empty, `_ELA`, `_MAT`, `_SCI`) that only `njsla_science`'s optional suffix group ever accepted, while the `njsla` asset matches a literal `_SLA`. With `njsla_science` gone, `njsla` takes a one-element tails list like `njgpa` does. The file goes from 29 passed and 2 failed to 17 passed.
- I patched the 2 automations catalog rows by hand instead of regenerating the file. `gen-automations-doc.py` cannot load kipptaf or kippmiami in a codespace, so a full regeneration deletes both sections. A regeneration also pulls in a lot of unrelated drift that has built up in the committed file, which belongs in its own change.

### General

      code location

      `CLAUDE.md`

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The user asked to remove `njsla_science` "assets and references" after the Pearson to Cambium transition, and chose the Cambium-only scope over a scope that would also delete the Pearson models.

Scope reasoning. `njsla_science` exists in 2 unrelated places. Under Cambium it is a placeholder asset added defensively in #5407, on the theory that a file dropped in an unwatched folder is ignored silently. Under Pearson it is a real historical feed: `src_pearson__njsla_science`, `stg_pearson__njsla_science` (kipptaf and the pearson package), `int_pearson__njsla_science` (kipppaterson), the SFTP assets in all 3 regions, and the `state_nj_njsla_science` branch of `int_pearson__all_assessments`. Only the Cambium placeholder is removed here. The `state_nj_njsla_science` assessment type stays, because `stg_cambium__njsla` still emits it for Science rows.

Docstring edit. `build_remote_file_regex`'s `filename_suffix_regex` docstring used `cambium/njsla` against `cambium/njsla_science` as its worked example of why a permissive suffix group is safe. That folder no longer exists, so the example is gone. The guidance itself stays, because the parameter still accepts a permissive group for a feed Cambium has not sent.

Test comment edits. `test_one_feeds_asset_never_matches_another_feeds_file` kept its 6 hardcoded tails and still guards the directory-scoping invariant between `njgpa` and `njsla`. Only its rationale comment changed, because both remaining feeds now carry a literal suffix and neither is permissive any more. `CROSS_FEED` drops from 6 pairs to 2.

Verification. `uv run pytest tests/libraries/test_cambium_assets.py` in the worktree: 17 passed, against 29 passed and 2 failed on `main`. `uv run pytest tests/test_dagster_definitions.py -k "kippnewark or kippcamden"`: 2 passed. Direct import confirms each region's cambium `assets` list is now exactly `njgpa` and `njsla`, and both Couchdrop sensors still construct. `trunk check --force --no-fix` on all 7 changed files: no issues.

</details>


cbini:
## Review verdicts

**1. Sensor cursor staleness after asset removal — accepted, not fixed here. Filed as #5414.**

The mechanism is right, and I confirmed it line by line in `src/teamster/libraries/couchdrop/sensors.py`: line 66 loads the saved cursor, line 69 takes `min` over all of its values, line 145 writes a value only for assets still in `asset_selection`, line 163 saves the whole dict back. Nothing drops a key, so removing the asset does not remove its entry.

Two things the finding gets wrong or leaves out.

It is pre-existing, not triggered by this PR. The `njsla_science` entry has held `min_cursor` at `0` since #5407 merged. The full re-list was already happening before this diff and would keep happening without it. This PR neither causes nor worsens it.

The suggested remedy has a cost the finding does not name. A cursor reset also drops each remaining asset's own `cursor_modified_timestamp` to `0` (line 92), so the next tick re-requests a run for every matching file still in the Drive folder. That re-materializes history, not just the recent partitions.

The durable fix is 2 lines in `build_couchdrop_sftp_sensor`: keep only the keys whose identifier is in the current `asset_selection` before taking the min. That is a shared-library change touching every code location's couchdrop sensor, so it does not belong in a removal PR. #5414 holds it.

I could not read the live cursor either, for the same reason the review job could not: the homebrew Dagster server has no cursor read tool, and `set_sensor_cursor` is denied in this repo's settings. What I did verify is that both remaining Cambium feeds have landed real files, so a reset would re-baseline them to a nonzero timestamp: `kippnewark_cambium.src_cambium__njgpa` holds 598 rows and `src_cambium__njsla` holds 8721.

**2. `docs/reference/automations.md` hand-edit — accurate, no action.**

Correct on both counts, and the PR body already discloses it. `gen-automations-doc.py` drops the kipptaf and kippmiami sections in a codespace because their definitions need 1Password credentials that a plain `uv run` does not get, and a full regeneration also pulls in unrelated drift that belongs in its own change.

**3. Unexercised permissive-suffix capability in `build_remote_file_regex` — accurate, declined.**

Every caller now passes a literal token, so the permissive branch has no live user. I am leaving it. `filename_suffix_regex` is a plain pass-through into the composed pattern with no branch or dead code behind it, so removing the capability would delete a docstring paragraph and nothing else, and it would force the next Cambium feed to re-derive it. This is not dead weight worth a diff.

**4. Scope check — matches my own.**

Same result: every surviving `njsla_science` reference is Pearson's, which is deliberately out of scope. Pearson holds the 2022 through 2025 Science history that still reaches the state assessment dashboard.

---

On the unverifiable test counts: `uv run pytest tests/libraries/test_cambium_assets.py` in the worktree reports 17 passed, against 29 passed and 2 failed on `main`. That matches the review's hand-traced arithmetic exactly.

PR #5420, opened 2026-09-18
fix(cambium): add the njsla administrations to the assessment dim

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make NJSLA and NJSLA Science scores from Cambium show up in the semantic layer.

New Jersey moved NJSLA score reporting from Pearson to Cambium TIDE with the Spring 2026 administration. `dim_assessment_administrations` still builds its NJSLA and NJSLA Science rows from the Pearson staging models only. Pearson has no rows for academic year 2025, so the dimension has no row for any Cambium NJSLA administration. Every Cambium NJSLA score then points at an `assessment_administration_key` that does not exist, and Cube drops those scores instead of showing them wrong: `count_scores`, `_sum_proficient` and `pct_proficient` all reach `student_assessments` through that join.

This adds one CTE, `state_nj_njsla_cambium_administrations`, that reads `stg_cambium__njsla`. It mirrors the NJGPA Cambium CTE that already sits next to it. `stg_cambium__njsla` sets `assessment_type` and `assessment_name` per subject, so the one CTE covers both NJSLA and NJSLA Science.

NJGPA was already wired and is unchanged. `dim_assessments` needs no change: its key leaves out academic year, and Pearson history already supplies every Cambium module code.

## Reviewer Notes

The new CTE passes `assessment_type` and `assessment_name` through from staging instead of hard-coding them, which is the one way it differs from the two Pearson CTEs above it. That is what lets one branch serve both subjects.

### General

### dbt

      models; the existing properties file is unchanged because the dimension's
      column list does not change
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — the
      Cube exposure already covers this mart
- [x] **Breaking change?** No. This adds rows to the dimension and renames or
      removes nothing.
      not applicable; no source changes

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The change started from the question "are all cambium scores wired into our semantic layer?"

Diagnosis, verified against prod BigQuery:

- Cambium NJGPA is wired end to end: 4 administration combos, 0 missing dimension rows, 425 scores in `fct_assessment_scores_enrollment_scoped` at test year 2026.
- Cambium NJSLA has 34 administration combos and NJSLA Science has 7. All 41 are missing from prod `dim_assessment_administrations`.
- `dim_assessments` has 0 missing rows for all three Cambium assessments. Its key hashes `[assessment_type, module_code, source_assessment_id, test_type]`, which leaves out academic year, and Pearson history already carries every Cambium module code including `SCI05`, `SCI08` and `SCI11`.
- Upstream volumes are present and clean: njgpa 813 rows, njsla 9,951, njsla_science 1,943, all scored, no null `localstudentidentifier` or `test_date`.

The fact does not join the dimension. Both hash the same inputs independently with `dbt_utils.generate_surrogate_key`, so a dimension branch that never reads a vendor's staging model produces no row and the fact's key dangles with no error anywhere.

Verification of this change, via a local dev build with `--defer --favor-state`:

- All 34 NJSLA and 7 NJSLA Science administration keys computed from `stg_cambium__njsla` resolve in the dev dimension. All 41 are missing in prod.
- `unique_dim_assessment_administrations_assessment_administration_key` passes, so the new rows do not break the grain. `module_code` encodes grade, so the extra `grade_level` and `subject_area` columns in the `select distinct` do not split a key.
- The `assessment_key` relationships warning reports 8 rows, spanning 3 distinct keys. Prod reports the same 3 distinct keys, so this change adds no orphans. All 16 Cambium NJSLA assessment keys resolve in `dim_assessments`.
- The two bridge relationships warnings (5 rows each) come from prod bridge tables against the new dimension. This change only adds dimension rows, so it cannot raise them.

Separately, at the time of the diagnosis `int_assessments__score_anchors` and `int_assessments__resolved_section_enrollments` had last built before the Cambium NJSLA staging tables landed, so no NJSLA row had reached the fact yet at all. That is staleness, not a code gap: the anchor path reads `int_pearson__all_assessments`, which already unions every Cambium relation. Expect the rows to reach the fact on the next build. Without this change they would then fail the error-severity `relationships` test on `assessment_administration_key`.

Deliberately not in this PR: a matching Cambium NJSLA CTE in `dim_assessments.sql`. The NJGPA Cambium CTE there carries the comment "not strictly required ... but kept so this dimension does not depend on Pearson history". The same reasoning would apply to NJSLA, but it is provably a no-op against today's data, so it is a separate topic.

</details>


cbini:
## Review verdicts

**Finding 1 — inline SQL comment carries rationale that belongs in `description:`. Accepted, fixed in 649c3f654f.**

The comment on `state_nj_njsla_cambium_administrations` is now 3 lines: the required `grain projection, not dup-masking` annotation plus a grain line naming the partition key, matching the two Pearson NJSLA CTEs above it. The vendor-cutover rationale moved to the model `description:` in `properties/dim_assessment_administrations.yml`, where it now covers both Cambium branches.

I did not touch the NJGPA sibling's pre-existing rationale block. Editing it widens the diff past the NJSLA fix, and the new description already explains why both Cambium branches exist, so the NJGPA comment is now redundant rather than wrong. That is a separate cleanup.

**Finding 2 — data-dependent claims unverifiable in the review environment. Answered; dbt Cloud CI now confirms them.**

CI run 70403225307546 passed on the prior commit with 4 warn tests, all pre-existing:

| Test | Result | Why it is not this PR |
| --- | --- | --- |
| `fct_assessment_scores_*` → dim | 2,414 | The dimension is the PARENT, so adding rows can only reduce orphans; the child is a stale `zz_stg` clone |
| bridge enrollment_scoped → dim | 484 | same |
| bridge student_scoped → dim | 5 | same |
| `dim_assessment_administrations.assessment_key` → `dim_assessments` | 8 | Matches prod exactly — 8 rows across 3 distinct keys, queried directly against prod |

The local dev build also confirmed the counts in the PR body: all 34 NJSLA and 7 NJSLA Science administration keys computed from `stg_cambium__njsla` resolve in the dev dim and are all missing from prod, the PK `unique` test passes (so `module_code` encoding grade does not split the grain), and all 16 Cambium NJSLA `assessment_key` values resolve in `dim_assessments` with zero new orphans.

Separately, and not part of this PR: `dim_assessments.sql` has a Cambium NJGPA CTE justified as keeping that dimension off Pearson history. The same argument applies to NJSLA, but mirroring it there is provably a no-op today, so I left it out.

PR #5431, opened 2026-09-21
fix: key the graduation cut score join to the soonest graduating class

### LINKED ISSUE #5429: fix: key the graduation cut score join to the soonest possible graduating class
## What's happening

Three high school students hold the wrong graduation pathway code. Two are seniors who passed NJGPA-A but read as "Not Grad Eligible". One of those two cleared the cut score in both English language arts and math.

`int_students__graduation_pathway_scores` joins the cut score sheet on `cohort`. `cohort` freezes at high school entry. That is correct for New Jersey's 4-year adjusted cohort graduation rate. NJDOE sets graduation requirements by graduating class instead. The two values disagree whenever a student skips or repeats a grade after entering high school.

A student whose `cohort` has no row in the cut score sheet matches no pathway at all. The model then returns `final_grad_path_code = 'R'`, which reads as "met no requirements". No test fails and no build breaks.

The wrong code reaches the state. `final_grad_path_code` flows through `rpt_powerschool__autocomm_students` into `s_nj_stu_x__graduation_pathway_ela` and `s_nj_stu_x__graduation_pathway_math`. A daily AutoComm file writes both fields to PowerSchool.

One of the three students is a junior who was retained. That student passes today and would lose the pathway under a naive fix. The section below explains why.

## Steps to reproduce

1. Query `int_students__graduation_path_codes` for students in grade 12 whose `cohort` is above 2027. Three rows come back.
2. Read `final_grad_path_code` and `grad_eligibility` on those rows. Both read `R` and "Not Grad Eligible".
3. Query `int_pearson__all_assessments` for the same students, filtered to `testcode` in `ELAGP` and `MATGP`. Two of them hold NJGPA-A scale scores at or above the 450 cut.

## Where

- **Code location / dbt project:** `src/dbt/kipptaf`, `int_students__graduation_pathway_scores`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** `rpt_tableau__graduation_requirements`. The model landed in #5162.

## What we propose

Join the cut score sheet on `least(cohort, cohort_primary)` in place of `cohort`.

`cohort_primary` is `(academic_year + 13) - grade_level`. It names the class the student's current grade puts them in. `least` picks the soonest class the student could graduate with.

Neither column alone works, because the two fail in opposite directions.

| Situation | `cohort` | `cohort_primary` |
| --- | --- | --- |
| A student skipped a grade, so the grade level already moved | wrong | right |
| A student repeated a grade and will recover credits over the summer | right | wrong |
| A student is on track | same | same |

The second row is the case that rules out `cohort_primary` on its own. Summer credit recovery produces no enrollment row, so the model cannot see the recovery until the grade level moves the next fall. `least` holds the student at the earlier class until then.

## What we rejected

**Use `cohort_primary` alone.** It fixes both accelerated seniors. It also drops the retained junior from "Math Only" to "Not Grad Eligible", because their new key of 2028 has no rows in the sheet.

**Add rows for the class of 2028 and change no code.** This also fixes both seniors. It requires writing cut scores NJDOE has not published, which the start-of-year procedure in `docs/models/hs-early-warning-data-model.md` already forbids.

**Fall back to the nearest cohort that has rows.** This keeps every student on the dashboard. It also applies another class's rules without saying so, which is the same silent wrong answer this issue exists to remove.

**Re-anchor `cohort` itself when a grade skip appears.** This breaks the 4-year adjusted cohort graduation rate for Newark, Camden, and Paterson. About 45 models read `cohort`.

## What it costs

**An 11th grader who graduates without ever being placed in grade 12 is never checked for FAFSA.** `fafsa_required` reads `grade_level = 12 and fafsa_season_12th`. Such a student is scored on pathways as a junior, then the graduation drops them from the model. `least` does not change this and does not fix it. Operations has to watch for those students. The warehouse cannot see them.

**A student retained more than once can key to a class that predates the current test version.** The sheet holds NJGPA-A rows for the class of 2027 only. `least` pins a twice-retained student to their older class, which will never gain NJGPA-A rows. Today 3 students sit in that position and all 3 scored below the 450 cut, so none of them lose anything. `cohort_primary` would age better for those students once the sheet advances.

**The fix does not populate the current junior class.** 411 of 418 juniors show a blank pathway because the class of 2028 has no cut scores yet. Only real cut scores from NJDOE fix those students.

## What we measured

We reproduced the whole chain against prod and ran the current join key as a control. The control returned 75,994 rows and 75,436 distinct keys, which matches prod exactly on both counts.

Under `least(cohort, cohort_primary)`, 3 students change and no other student in the network moves.

| Student | Region | Grade | `cohort` | `cohort_primary` | Today | After |
| A | Newark | 12 | 2028 | 2027 | Not Grad Eligible | Grad Eligible |
| B | Newark | 12 | 2028 | 2027 | Not Grad Eligible | Math Only |
| C | Camden | 11 | 2027 | 2028 | Math Only | Math Only |

No student joins or leaves any layer. `rpt_tableau__graduation_requirements` holds 2,789 students before and after. That view filters on `e.cohort`, which this change does not touch.

## Related work

This issue also covers two smaller changes that follow from the same finding.

The `int_students__student_enrollments__cohort_not_ahead_of_entry` warn test says a student ahead of their cohort means a skipped grade or a missing enrollment year, and that the fix is a records correction in PowerSchool. We found a third cause. A student can recover credits in summer school and advance a grade with no enrollment row to show for it, because `int_powerschool__student_enrollment_union` filters out `schoolid = 12345`. The records are right in that case. The test description needs to say so.

The cut score runbook in the `graduation-pathways` skill does not check its own work. After someone enters new cut scores, the runbook should report which students still match no pathway and why.

## This also affects CSGF, separately

The investigation behind this issue surfaced a second problem with the same root cause, in the Charter School Growth Fund graduation rate. `rpt_gsheets__csgf_hs_grad_data` counts a graduate only when `academic_year + 1 = cohort`, so a student who graduates ahead of their cohort year sits in the denominator and never reaches the numerator. In AY2025, 14 students graduated early.

That problem is tracked in #5432 and is fixed separately. This issue changes nothing in the CSGF model, and #5432 changes nothing here.

<details>
<summary>For Claude</summary>

The join is at `src/dbt/kipptaf/models/students/intermediate/int_students__graduation_pathway_scores.sql`, the `matched` CTE, currently `on s.cohort = c.cohort and s.discipline = c.discipline`. `cohort_primary` already reaches this model through `int_extracts__student_enrollments_subjects`, so no upstream column add is needed. Derive the key as a named column in the `students` CTE rather than calling `least()` in the join predicate: `.claude/rules/dbt-sql.md` bans one-sided calculations in join predicates.

Student identifiers stay out of this issue. Step 1 under "Steps to reproduce" regenerates the list, and the linked Asana task carries the same students.

Student A cleared the 450 cut in both subjects. Student B cleared math and fell 4 points short in English language arts. Student C cleared math and fell 15 points short in English language arts. A fourth accelerated senior holds no NJGPA record at all and stays blank under every variant, so the fix does not reach them.

Every cut score in `stg_google_sheets__student_graduation_path_cutoffs` is identical across all 5 cohorts: NJGPA 725, NJGPA-A 450, SAT 450 and 440, ACT 17, PSAT 420 and 21. The `cohort` column on that sheet only gates which rows exist and which assessment versions are available. That is why re-keying changes no pass or fail outcome for the 68 retained students whose key moves.

Variants simulated against prod, all six combinations of three keys (`cohort`, `cohort_primary`, `least`) against two sheets (current, and current plus class of 2028 rows cloned from 2027). Only the 3 students in the table above ever move. Stranded NJGPA-A holders by key: `cohort` 5, `cohort_primary` 5, `least` 3, out of 387 NJGPA-A holders. All 3 stranded under `least` scored below the 450 cut, so none of them lose a pathway. No null `cohort` or `cohort_primary` exists in the population, so `least` cannot produce a null key.

`rpt_tableau__graduation_requirements` filters `e.cohort between current_academic_year - 1 and current_academic_year + 5`. Leave that filter on `cohort`. It is a roster window, not a requirements key.

The Asana task "get someone to fix 6 students whose grade level is ahead of their cohort" in `Agenda: GR/AW O3` describes these students. Its premise was that the records were wrong. For at least one student the records are right and the acceleration was real. Close that task after the PR merges.

</details>

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will score a student's graduation pathways against the class they will actually graduate with, instead of the class they entered high school with.

`int_students__graduation_pathway_scores` joined the cut score sheet on `cohort`. `cohort` freezes at high school entry, which is correct for New Jersey's 4-year adjusted cohort graduation rate. NJDOE sets graduation requirements by graduating class instead. The two disagree whenever a student skips or repeats a grade after entering high school, and a student whose `cohort` has no row in the sheet matches no pathway at all. The model then returns `final_grad_path_code = 'R'`, which reads as "met no requirements", and that letter reaches the state through the daily AutoComm file.

Two seniors passed NJGPA-A and read as "Not Grad Eligible" today. One of them cleared the cut score in both subjects.

The join now keys on `least(cohort, cohort_primary)`, the soonest class the student could graduate with.

Refs #5429

## Reviewer Notes

- **`cohort_primary` alone does not work, which is the whole reason for `least`.** A student repeating a grade who recovers credits over the summer produces no enrollment row, because `int_powerschool__student_enrollment_union` filters out the summer school. The model cannot see that recovery until the grade level moves the next fall. Keying on `cohort_primary` alone drops one such junior from "Math Only" to "Not Grad Eligible" today. `least` holds them at the earlier class and keeps the pathway they earned.
- **Adding rows for the class of 2028 would also fix the two seniors, and was rejected.** NJDOE has not published those cut scores. The start-of-year procedure already forbids writing a threshold nobody has defined.
- **The key is derived as `cut_score_cohort` in the `students` CTE, not inlined in the join.** `.claude/rules/dbt-sql.md` bans one-sided calculations in join predicates. The column stays inside that CTE and is not projected, so no downstream column set changes.
- **Two costs ship with this, both documented rather than solved.** A student retained more than once keys to a class that can predate the assessment version they sat. And an 11th grader who graduates without ever being placed in grade 12 is never checked for FAFSA, because `fafsa_required` reads `grade_level = 12`. Operations owns watching for the second one.
- **The `cohort_not_ahead_of_entry` description was wrong, not the test.** It said a student ahead of their cohort means a records correction in PowerSchool. Summer credit recovery is a third cause where the records are already right. Severity stays `warn`.

### General

### dbt

- [x] Properties entry updated on the model and on the changed test.
- [x] Exposure — unchanged. No columns added or removed.
- [x] **Breaking change?** No. No column set changes anywhere in the chain.

### Docs

- [x] `docs/models/hs-early-warning-data-model.md` carries the rule, both costs, and the new coverage step.
- [x] The `graduation-pathways` skill carries the same rule and the runbook step.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Found while answering a question about students whose cohort sat above 2027.

**Verification.** The whole chain from `int_students__graduation_pathway_scores` through `int_students__graduation_path_codes` was reproduced against prod relations in one query, with the current join key run as a control. The control returned 75,994 rows and 75,436 distinct keys, matching prod exactly on both counts, which is what makes the diff trustworthy. Six variants were then simulated: three keys (`cohort`, `cohort_primary`, `least`) against two sheets (current, and current plus class-of-2028 rows cloned from 2027). Only 3 students move under any variant.

Results by key, all students in grades 8 through 12, Miami excluded. Under `cohort_primary`: 2 seniors gain, 1 junior loses. Under `least`: 2 seniors gain, nobody loses, and the same 3 students land identically whether or not the sheet has 2028 rows. That last property is why `least` needs no sheet change to ship safely.

Stranded NJGPA-A holders, meaning a student whose keyed cohort has no NJGPA-A row in the sheet: `cohort` 5, `cohort_primary` 5, `least` 3, out of 387 NJGPA-A holders. All 3 stranded under `least` scored below the 450 cut, so none of them lose a pathway today. The residual risk is a twice-retained student who passes NJGPA-A; `cohort_primary` would age better for exactly that shape once the sheet advances.

No null `cohort` or `cohort_primary` exists in the population, so `least` cannot produce a null key. Grades 10 and below are unaffected either way, since their eligibility is hardcoded and their code passes `ps_grad_path_code` through. `rpt_tableau__graduation_requirements` holds 2,789 students before and after; its own filter reads `e.cohort between current_academic_year - 1 and current_academic_year + 5`, which is a roster window and was deliberately left alone.

Every cut score in the sheet is identical across all 5 cohorts. The `cohort` column there gates only which rows exist and which assessment versions are available, which is why re-keying changes no pass or fail outcome for the 68 retained students whose key moves.

`dbt compile --target prod` resolves the model and the compiled SQL joins on `s.cut_score_cohort = c.cohort`. No warehouse write was run.

Student identifiers are deliberately absent from this PR and from #5429. Step 1 of the reproduce steps on the issue regenerates the list.

This does not populate the current junior class. 411 of 418 juniors show a blank pathway because NJDOE has not published the class of 2028 cut scores, and only real cut scores fix that.

</details>


GabyRangelB:
Per-finding verdicts on the review. All three accepted and fixed in e760cc771.

**Finding 1 — unit test fixture missing `cohort_primary`. Confirmed and fixed.**

Verified by running it rather than reading it: `dbt test --select test_njgpa_cut_score_boundaries_per_assessment_version` failed with `Database Error / Name cohort_primary not found inside e at [120:31]`. The diagnosis was exactly right, including the mechanism — a `format: sql` fixture substitutes a literal CTE, so a column the model reads but the fixture does not select cannot resolve. Every row of that fixture now carries `cohort_primary`, set equal to `cohort` so the existing expectations are unchanged. Both unit tests pass.

**Finding 2 — `least()` propagates NULL. Accepted and fixed.**

`.claude/rules/dbt-sql.md` prescribes `(select min(x) from unnest([...]) as x)` for exactly this shape, and I did not follow it. The consequence is the one described: a null on either side yields a null key, the cutoff join matches nothing, and the student lands on `final_grad_path_code = 'R'`, which is the failure this branch exists to remove. `cut_score_cohort` now uses the `min` over `unnest` form.

The calibration note is accurate — today's population has no null `cohort` or `cohort_primary`, so this closes a latent path rather than an active bug. Worth adding that `min` over `unnest` is also better behavior than a guard would be: if one column is ever null, the student keys off whichever value survives instead of matching nothing.

**Finding 3 — no test for the regression. Accepted.**

`test_cut_score_cohort_uses_soonest_graduating_class` covers both directions, which matters because they fail oppositely. Student 6 is a senior who skipped a grade, holding `cohort` 2028 against a sheet that stops at 2027. Student 7 is a junior who repeated one, holding `cohort_primary` 2028. The `expect` block also pins `cohort` itself to its unchanged value, so a future change that "fixes" the mismatch by moving `cohort` breaks this test — the 4-year adjusted cohort graduation rate depends on `cohort` staying frozen.

The test was verified in both directions. Against the old join key it fails, with student 6 matching no cut score row at all (null `cutoff`, null `scale_score`) instead of the 450 cut with a score of 500. Against the new key it passes. A test that only passes proves nothing about whether it guards anything.

**Finding 4 — convention checks.** No action; recorded.

One consequence worth naming: the expression is no longer literally `least()`, so the prose in the reference doc and in the `graduation-pathways` skill now describes it as the lesser of the two columns and says why it is not `least`. The issue body still says `least(cohort, cohort_primary)` as shorthand for the intent, which remains accurate.

GabyRangelB:
@claude all three findings accepted and fixed in e760cc771. Verdicts are in the comment above this one.

Finding 1 reproduced before fixing: `Name cohort_primary not found inside e at [120:31]`. The fixture now carries `cohort_primary` on every row.

Finding 2: `cut_score_cohort` moved from `least()` to `(select min(x), from unnest([e.cohort, e.cohort_primary]) as x)`, per the rule you cited.

Finding 3: added `test_cut_score_cohort_uses_soonest_graduating_class`, covering a senior who skipped a grade and a junior who repeated one. Verified failing on the old join key and passing on the new one.

Please re-review the delta. Two things specifically worth your attention:

1. Whether the `min` over `unnest` form still satisfies ST06 in that position. Trunk reports no rule violation, only sqlfmt reflow, but the expression changed category from a simple function to a scalar aggregate subquery and I want a second read on where it sorts in the select list.
2. Whether pinning `cohort` in the new test's `expect` block is the right guard. The intent is that a future change which "fixes" the mismatch by moving `cohort` itself breaks this test, because the 4-year adjusted cohort graduation rate depends on `cohort` staying frozen at high school entry.

PR #5435, opened 2026-09-21
fix: count early graduates in the CSGF 4-year graduation rate

### LINKED ISSUE #5432: fix: count early graduates in the CSGF 4-year graduation rate
## What's happening

`rpt_gsheets__csgf_hs_grad_data` does not count a student who graduates ahead of their cohort year. That student stays in the denominator and never reaches the numerator, so the graduation rate we report to the Charter School Growth Fund reads low.

The model decides who graduated with this expression.

```sql
case when co.academic_year + 1 = co.cohort and co.exitcode = 'G1' then 1 else 0 end as is_4yr_grad
```

The equality holds only for a student who graduates in the exact year their cohort implies. A student who graduates a year early fails it. The same student still counts toward `adjusted_cohort`, because `is_entry_cohort` only asks whether they started grade 9 at that school.

Early graduation is not rare and it is growing. Counting rows where `exitcode = 'G1'`, at a high school, outside Miami:

| Year | Graduated on time | Graduated early | Graduated late |
| --- | --- | --- | --- |
| AY2022 | 271 | 2 | 3 |
| AY2023 | 407 | 6 | 0 |
| AY2024 | 377 | 3 | 1 |
| AY2025 | 374 | 14 | 3 |

Late graduates are correctly left out of a 4-year rate. Early graduates are not, if CSGF follows New Jersey's adjusted cohort rule, which counts a student who graduates early in the cohort they entered with.

For the class of 2026, the class the model reports today, the fix moves the network total from 374 graduates to 377. Two of the three schools move; the third does not.

## Steps to reproduce

1. Query `int_extracts__student_enrollments` for rows where `exitcode = 'G1'`, `school_level = 'HS'` and `region != 'Miami'`.
2. Compare `academic_year + 1` against `cohort` on those rows. In AY2025, 14 rows have `academic_year + 1 < cohort`.
3. Rebuild the `grad_roster` CTE from `rpt_gsheets__csgf_hs_grad_data` for `cohort = 2026`, and sum `is_4yr_grad` against a version that uses `<=` instead of `=`. The totals differ by 3.

## Where

- **Code location / dbt project:** `src/dbt/kipptaf`, `rpt_gsheets__csgf_hs_grad_data`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** the `csgf_data` exposure

## How we found it

We found this while investigating #5429, which fixed a different problem with the same root cause. Students who skip or repeat a grade break assumptions that hold for everyone else. #5429 covers graduation pathway cut scores and changes nothing here. This issue covers the CSGF rate and is independent of it.

## What we propose

Change the equality to `academic_year + 1 <= co.cohort`.

An early graduate then counts in the cohort they entered with. A late graduate still does not, which is correct for a 4-year rate.

## What we rejected

**Key the graduate count on `cohort_graduated`.** `int_powerschool__student_enrollment_union` already computes it as `if(exitcode = 'G1', cohort_primary, null)`. That is the class the student's grade level implied at graduation, not the class they entered with, so it moves early graduates into a different class's numerator. The adjusted cohort rate is defined by entry, so this is the wrong anchor.

**Report early graduates as a separate figure and leave the rate alone.** This keeps the current number defensible under either reading. It also gives CSGF a number they did not ask for and still understates the rate they did ask for.

**Change nothing.** This is the right answer if CSGF's own definition excludes early graduates. We could not confirm their rule from the repo.

## What we are unsure about

We did not verify how CSGF defines a 4-year graduate. New Jersey's adjusted cohort rate counts early graduates in their entry cohort, and we assumed CSGF matches the state. The CSGF item-list documentation would settle it. Whoever owns the CSGF collection should confirm before the linked PR merges.

<details>
<summary>For Claude</summary>

The expression is in the `grad_roster` CTE of `src/dbt/kipptaf/models/extracts/google/sheets/rpt_gsheets__csgf_hs_grad_data.sql`. `is_cohort_grad_year` on the line above carries the same equality and feeds nothing downstream, so decide whether it needs the same treatment or whether it should go.

The model filters `co.cohort = {{ var("current_academic_year") }}`, so it reports one graduating class at a time and today that is the class of 2026. An early graduate's row survives the `rn_undergrad = 1` filter because that picks their most recent non-99 row, which is the grade 12 row carrying the `G1` exit.

Measured for `cohort = 2026`, rebuilding the model's own CTEs: adjusted cohort 96, 183 and 138 across the three high schools; graduates 374 now against 377 under `<=`. Every `G1` row inside the class-of-2026 roster is early or on time, so `<=` and a plain `exitcode = 'G1'` count agree exactly for this class. They will not agree for a class that has late graduates, which is why the predicate keeps the bound rather than dropping it.

Per-school deltas are small cells, so they are not written here. The reproduce query returns them.

</details>

GabyRangelB:
# Pull Request

## Summary & Motivation

When merged, this pull request will count a student who graduates ahead of their cohort year as a 4-year graduate in the Charter School Growth Fund submission.

`rpt_gsheets__csgf_hs_grad_data` decided who graduated with `academic_year + 1 = co.cohort and co.exitcode = 'G1'`. That equality holds only for a student who graduates in the exact year their cohort implies. An early graduate failed it, stayed in the `adjusted_cohort` denominator through `is_entry_cohort`, and never reached the numerator. The rate we send CSGF read low.

The predicate now reads `academic_year + 1 <= co.cohort`. A late graduate still does not count, because this is a 4-year rate.

Early graduation is not rare and it is growing: 2 in AY2022, 6 in AY2023, 3 in AY2024, and 14 in AY2025. For the class of 2026, the class the model reports today, the network total moves from 374 graduates to 377.

Refs #5432

## Reviewer Notes

- **This PR rests on an assumption somebody needs to confirm.** New Jersey's adjusted cohort rate counts an early graduate in the cohort they entered with, and I assumed CSGF matches the state. I could not find CSGF's own definition in the repo. If CSGF excludes early graduates, close this and leave the model alone. The assumption is written into the model description so the next reader sees it.
- **`is_cohort_grad_year` still carries the old equality and is deliberately untouched.** It is computed in `grad_roster` and never selected, so it changes nothing today. Dropping it is a reasonable follow-up, but it is not this fix, and whoever owns the model may want it for a denominator later.
- **`<=` rather than dropping the bound entirely.** Inside the class of 2026 every `G1` row is early or on time, so `<=` and a plain `exitcode = 'G1'` count agree exactly for this class. They diverge for a class with late graduates, which is the case the bound exists for.
- **The column descriptions are new, and the other 6 columns still have none.** That gap predates this change; I documented the 2 columns whose meaning this PR changes rather than sweeping the file.

### General

### dbt

- [x] Properties file updated — model description plus the 2 affected columns.
- [x] Exposure — `csgf_data` already exists and is unchanged. No columns added or removed.
- [x] **Breaking change?** No. Same columns, same types. Two of the three schools' values move for the class of 2026.

## CI checks

- [ ] **Trunk** — passes.
- [ ] **dbt Cloud** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Found while investigating #5429, which fixes a different consequence of the same underlying situation: students who skip or repeat a grade break assumptions that hold for everyone else. The two changes share no files and no models.

**Verification.** The model's own CTEs (`entry_cohort`, `transfer_in`, `transfer_out`, `grad_roster`) were rebuilt against prod `int_extracts__student_enrollments` for `cohort = 2026`, computing the current and proposed predicates side by side. Adjusted cohort is unchanged by construction, since the change touches only the numerator. Graduates move from 374 to 377 network-wide. Two of the three high schools move and one does not. Per-school deltas are cells of 1 and 2, so they are not written here or in #5432; the reproduce query on the issue returns them.

The early and late split across years came from counting `exitcode = 'G1'` rows at `school_level = 'HS'` outside Miami, comparing `academic_year + 1` against `cohort`. Late graduates: 3 in AY2022, 0 in AY2023, 1 in AY2024, 3 in AY2025. Those stay excluded under the new predicate, which is the intended behavior for a 4-year rate and the reason the bound was kept.

An early graduate's row survives `rn_undergrad = 1` because that picks the most recent non-99 row, which is the grade 12 row carrying the `G1` exit. This was checked rather than assumed.

`dbt compile --target prod` resolves the model. The compiled SQL shows `is_4yr_grad` on `<=` and `is_cohort_grad_year` still on `=`, which is the intended state.

</details>


GabyRangelB:
@claude per-finding verdicts. Four accepted and fixed in 09a1fdc15; one pushed back on.

**Finding 1 — inline comment duplicates the description. Accepted.** The two-line comment restated the rationale nearly verbatim. It now reads only that the inequality is deliberate and points at the description, which is the part a reader of that line cannot see.

**Finding 2 — not every column described. Accepted.** The rule is unconditional on a modified model, and my Reviewer Note argued from precedent rather than from the rule, which is the weaker position. All 6 remaining columns now have descriptions.

**Finding 3 — published doc not updated. Accepted.** `docs/models/csgf-data-model.md` gains a `csgf_hs_grad_data` did not count early graduates — resolved section, matching the shape of the existing cohort-scope section. It carries the early and late counts per year, the 374 to 377 movement, and the unconfirmed assumption about CSGF's own definition.

**Finding 4 — re-run the verification. Done.** Re-ran your query against prod just now: `old_grads` 374, `new_grads` 377 for cohort 2026. Unchanged from the original run.

**Finding 5 — unit test for the operator. Pushing back, for now.**

The reasoning is sound in general, and I added exactly this kind of test on #5431 when you asked for it there. The difference is that #5431's fix is settled and this one is not. This entire PR rests on an assumption nobody has confirmed: that CSGF counts an early graduate the way New Jersey does. You checked `docs/models/csgf-data-model.md` and the `csgf-data-collection` skill and found no answer, which matches what I found.

If that assumption turns out to be wrong, the predicate does not revert to `=` — the model needs a different shape, possibly reporting early graduates as their own figure. A unit test written now would encode the unconfirmed reading as the expected behavior and would have to be rewritten rather than kept.

The model also has no unit tests today and four CTEs with three self-joins, so the fixture is not free. I would rather add it once the CSGF owner confirms the definition, at which point it guards a decision instead of a guess. If you think the test earns its place before that confirmation, say so and I will add it.

PR #5438, opened 2026-09-21
feat: give Grow Regional Observers a whole-school observation group

### LINKED ISSUE #5437: feat: give Grow Regional Observers an observation group covering their whole school
## What and why

New Teacher Network Coordinators cannot enter an observation on anyone they do not personally coach. They need to observe any teacher at their school.

They already hold the `Regional Observer` role, and the extract already scopes them to their school. Neither does anything, because a seat to actually observe someone comes only from an observation group, and nothing puts a Regional Observer in one.

Today a school's observation groups are built two ways. There is one group per coach, where the coach is the sole observer and their own reports are the observees. There is also a fallback group named `Teachers`, whose observers are the school's School Admins and School Assistant Admins, and whose observees are the teachers who have no coach. That fallback is empty in practice, because every observee currently has a coach.

`Regional Observer` appears in neither. The role is never read in `grow_user_sync`.

28 active staff hold the role. Here is what each of them can observe right now:

| Seat today | Count |
| --- | --- |
| Nothing at all | 17 |
| Their own reports, because they are also a Coach | 7 |
| Their whole school, because they are also a School Admin or Assistant Admin | 4 |

Danielle Lisbon is in the middle row. She holds Teacher, Coach and Regional Observer, so she reaches her own coachees and nobody else.

## Proposed approach

Add a second fixed group per school, alongside `Teachers`, holding every observee at that school as an observee and that school's Regional Observers as observers.

Coach groups are left alone, so teachers still do not see peers through them. Each coordinator gains exactly one extra entry in their observation picker.

Pick the school's Regional Observers from `regional_admin_school_ids` rather than from their home school. That column is the scope the extract already writes, and reading it means the group follows a coordinator whose scope is widened beyond one campus.

<details>
<summary>For Claude</summary>

> Context, not instructions.

Files:

- `src/teamster/code_locations/kipptaf/level_data/grow/assets.py` — the school PUT builds `wanted` at roughly line 291. `ADMIN_ROLES`, `_can_anchor_group` and `_observes_fallback` are the existing role-to-seat plumbing.
- `src/dbt/kipptaf/models/extracts/schoolmint/rpt_schoolmint_grow__users.sql` — already emits `Regional Observer` in `role_names`, `observers` in `group_type`, and the school list in `regional_admin_school_ids`. No change expected here.

Design notes:

- The school PUT REPLACES `observationGroups`, so the new group needs the same match-and-reuse treatment the other groups get, or its `_id` churns daily and orphans history.
- Name it with a fixed per-school string and no `match_key`, exactly like `Teachers`, so no hand-made group can be claimed by accident.
- Do not exclude the group's own observers from its observee list. Most Regional Observers are also teachers, and excluding them would make a coordinator unobservable by their coordinator peers at the same school. A coordinator appearing in a list they can act on is cosmetic; the alternative is a real access gap.
- Reuse `_can_anchor_group` so a readonly, inactive, or non-observer account never becomes the group's observer. A Regional Observer who is also a Regional Admin is readonly and should be filtered out by it.
- Branch `cristinabaldor/feat/claude-grow-miami-slr-observer` widens `Regional Observer` scope to region-wide for Miami School Leaders in Residence. Keying the group off `regional_admin_school_ids` is what makes these two changes compose instead of conflict.

Acceptance: every active Regional Observer who can anchor a group is an observer on a group whose observees are all of their scoped schools' observees, and no coach group gains or loses a member.

Refs #5052, #5054.

</details>

cristinabaldor:
# Pull Request

## Summary & Motivation

When merged, this pull request will let New Teacher Network Coordinators enter an observation on any teacher at their school, instead of only on the people they personally coach.

Coordinators already hold the `Regional Observer` role in Grow, and the extract already records which schools they cover. Neither did anything, because the only thing that grants a seat to observe someone is an observation group, and nothing put a Regional Observer in one. `grow_user_sync` built a group for each coach, holding that coach's own reports, and a `Teachers` fallback group observed by school admins. The role was never read.

Of the 28 active Regional Observers, 17 could observe nobody at all. 7 more, Danielle Lisbon among them, reached only their own coachees because they happen to also be a Coach. The remaining 4 were covered only because they are also a School Admin.

This adds a second fixed group per school, named `Regional Observers`, alongside `Teachers`. It holds every observee at the school, and its observers are the Regional Observers whose scope names that school. Coach groups are not touched, so teachers still do not see peers through them, and each coordinator gains exactly one new entry in their picker.

Closes #5437.

## Reviewer Notes

**The group name.** `Teachers` is named for the people in it. `Regional Observers` is named for the people acting in it, which breaks that pattern. I chose it because a coordinator opening their picker can tell at a glance why they have the group. Say the word and I will rename it.

**Coordinators appear in their own group.** Most Regional Observers also teach, so they show up in the observee list of a group they observe. That is deliberate: excluding them would make a coordinator unobservable by their coordinator peers at the same school, which is a real access gap rather than a cosmetic oddity.

**Observers come from the scope column, not the home school.** That is what makes this compose with the Miami School Leaders in Residence change already on `main` rather than quietly ignore it. See the fold-out.

### General

### Dagster

- [x] Run `uv run dagster definitions validate` for any modified code location — `kipptaf` passes.
- [x] Run `uv run pytest tests/test_dagster_definitions.py` for any modified code location — `test_definitions_kipptaf` passes. The other four code locations fail in my worktree only because a fresh worktree has no dbt `target/` for their projects, which is a setup gap rather than a code defect. They pass in the main checkout. See the fold-out.
- [x] New integrations follow the Library + Config pattern — not applicable, no new integration.

### dbt

No dbt changes. The extract already emits everything this needs.

### Docs

## CI checks

- [ ] **Trunk** — passes.

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. The ask came in as "Danielle Lisbon and the other New Teacher Network Coordinators need observation groups in Grow so they can enter observations on anyone at their school."

**Why `regional_admin_school_ids` and not `school_id`.** Three Miami School Leaders in Residence — Angelica Ramirez, Mesra Pierre and Okevia Carter — hold `Regional Observer` with a home Grow school of `Room 11`, which has zero observees. Commit `03e13b1fc`, now on `main`, widened Regional Observer scope to region-wide for exactly this group, so their `regional_admin_school_ids` carries 8 real Miami schools. Keying the new group off `school_id` would have put all three in an empty `Room 11` group and left the Miami campuses uncovered. Keying off the scope column is what makes the two changes compose.

**Simulated payload against the live extract**, as of 2026-09-21. 22 schools gain a populated `Regional Observers` group. Counts are observees in the new group:

| School | Observers | Observees |
| --- | --- | --- |
| KIPP Newark Collegiate Academy | 2 | 66 |
| KIPP SPARK Academy | 2 | 47 |
| KIPP Upper Roseville Academy | 2 | 44 |
| KIPP Life Academy | 2 | 43 |
| KIPP THRIVE Academy | 2 | 43 |
| KIPP Newark Lab High School | 1 | 42 |
| KIPP Seek Academy | 3 | 41 |
| KIPP Cooper Norcross High | 1 | 39 |
| KIPP Royalty Academy | 4 | 36 |
| KIPP Lanning Square Primary | 1 | 35 |
| KIPP Purpose Academy | 1 | 33 |
| KIPP Justice Academy | 1 | 30 |
| KIPP Lanning Square Middle | 2 | 29 |
| KIPP Courage Academy | 4 | 28 |
| KIPP Rise Academy | 1 | 28 |
| KIPP TEAM Academy | 1 | 28 |
| KIPP Legacy Elementary | 4 | 13 |
| KIPP Legacy Middle | 3 | 3 |
| KIPP Miami Technical High | 3 | 3 |
| KIPP Miami - North Campus | 3 | 0 |
| KIPP Miami - Poinciana Campus | 3 | 0 |
| Room 11 | 3 | 0 |

The last three have no observees, so both lists empty together and the group stays out of every picker. That mirrors the existing `Teachers` handling: an empty group that kept its observers would show up holding nobody.

**Group identity.** `match_keys[REGIONAL_OBSERVER_GROUP]` is `None`, exactly like `Teachers`. A fixed name claims only an exact match and never falls back to a suffix, so a hand-made group can never be claimed by accident. Without the match-and-reuse treatment the school PUT would mint a fresh `_id` every morning and orphan the observation history recorded against it.

**`_observes_whole_school` reuses `_can_anchor_group`,** so an inactive, readonly, or non-observer account never anchors the group. The readonly branch matters in practice: a Regional Admin is readonly by design, and if one ever also held `Regional Observer` the role alone must not give them a writable seat. `tests/assets/test_grow_observation_groups.py` pins all four negative cases plus the two positives.

**Not covered by tests.** The group-building block is inline in the asset body, not a pure function, so only the helper is unit-tested. The payload shape was verified by simulating the same predicates in SQL against `kipptaf_extracts.rpt_schoolmint_grow__users`, which produced the table above.

**On the Dagster validation checkbox.** A fresh `git worktree add` has no `src/dbt/<project>/target/`, and every code location reads `manifest.json` at module import, so all five fail on `FileNotFoundError` before any of this code runs. I ran `dbt deps` and `dbt parse` for `kipptaf` in the worktree, after which `test_definitions_kipptaf` passes. The other four still fail on the same missing manifest for their own projects, and `test_definitions_kippnewark` passes unchanged in the main checkout, which is how I ruled this change out as the cause.

Refs #5052, #5054.

</details>


cristinabaldor:
Verdicts on the review findings.

**1. Efficiency — rescanning `users` per school. Declining.**

Measured the actual scale: the extract emits 1,128 rows and there are 28 active Grow schools, so the scan is about 31,600 predicate calls per run, once daily. The loop body around it makes one HTTP PUT to the Grow API per school, so the work this would save is nanoseconds inside a network-bound loop.

The shape also matches what is already there. `_at_school` runs per school over `school_users` and `users_by_grow_id` in the same loop, so precomputing a map for only the new lookup would make the two adjacent lookups inconsistent for no measurable gain. Revisit if the extract ever grows by an order of magnitude.

**2. Test coverage — the group-assembly block is untested. Acknowledged, no change.**

Correct, and called out in the PR body. The block is inline in the asset body rather than a pure function, so reaching it with a unit test means extracting it, and that refactor is wider than this change warrants. The four empty/non-empty combinations are covered by the hand-trace in the review plus the SQL simulation in the fold-out. Worth doing as its own change if this block grows a third group.

**3, 4, 5, 6, 7.** No findings or not applicable. Nothing to action.

**On the Trunk failure.** Not a lint finding. Trunk checked both modified files and reported `No issues, 1 failure` — the failure is `HTTP 504` fetching hermetic Python 3.14.4 from the `python-build-standalone` GitHub release. Transient CI network error. Re-triggering with an empty commit.

src/teamster/code_locations/kipptaf/level_data/grow/assets.py
```diff
@@ -112,6 +112,18 @@ def _observes_fallback(user: dict[str, Any]) -> bool:
     )
 
 
+# A Regional Observer observes every observee at each school their scope
+# names, not only their own reports, so they get a group of their own rather
+# than a seat in anyone else's.
+REGIONAL_OBSERVER_ROLE = "Regional Observer"
+REGIONAL_OBSERVER_GROUP = "Regional Observers"
```
cbini, line 118: this is redefined as a dict and as None below. what's up with that?

src/teamster/code_locations/kipptaf/level_data/grow/assets.py
```diff
@@ -112,6 +112,18 @@ def _observes_fallback(user: dict[str, Any]) -> bool:
     )
 
 
+# A Regional Observer observes every observee at each school their scope
+# names, not only their own reports, so they get a group of their own rather
+# than a seat in anyone else's.
+REGIONAL_OBSERVER_ROLE = "Regional Observer"
```
cbini: only used once. YAGNI?

cbini (changes requested):

src/teamster/code_locations/kipptaf/level_data/grow/assets.py
cristinabaldor: Agreed, inlined in 0d6dc91d7. It was named once and read once, so it bought nothing. `ADMIN_ROLES` earns its place because three call sites read it; this didn't.

src/teamster/code_locations/kipptaf/level_data/grow/assets.py
cristinabaldor: It isn't reassigned — it's a dict *key* in two different dicts, so what varies is the value, not the constant.

```python
wanted = {
    "Teachers": {"observees": ..., "observers": ...},
    REGIONAL_OBSERVER_GROUP: {"observees": ..., "observers": ...},   # membership
}

match_keys = {
    "Teachers": None,
    REGIONAL_OBSERVER_GROUP: None,                                    # match key
}
```

`"Teachers"` has the identical shape eleven lines up: a key in `wanted` holding a membership dict, and a key in `match_keys` holding `None`. The two dicts are kept separate on purpose — the comment above `match_keys` says why, so the match key never leaks into the payload sent to the Grow API.

That said, you misread it, which is the useful signal here. The constant exists precisely because the two dicts must agree on the string: if they drift, the group silently loses its `_id` match and the sync mints a fresh group every morning, orphaning the observations recorded against the old one. `"Teachers"` gets that wrong today and survives only because nobody has retyped it.

Happy to pull the `match_keys` value into `wanted` as a third field and drop the parallel dict entirely, which would remove the shape that confused you — but that rewrites the existing `Teachers` and per-coach paths too, so I'd rather do it as its own change than widen this one. Say which you'd prefer.

cbini (approved):

PR #5439, opened 2026-09-21
fix(kipptaf): key ops dashboard next-year columns to academic_year + 1

### LINKED ISSUE #5430: fix: is_enrolled_oct01_next reads the wrong academic year when a student skips a year
## What's happening

`rpt_tableau__ops_dashboard` should set `is_enrolled_oct01_next` to true only when a student was enrolled on 1 October of the next academic year. Instead the field copies `is_enrolled_oct01` from the student's next existing row, whatever year that row belongs to. The Operations "Attrition Analysis" dashboard reads the field as next-year enrollment, so it counts a student as retained for a year in which the student was not enrolled.

The model builds the field with one window function:

```sql
lead(se.is_enrolled_oct01, 1, false) over (
  partition by se.student_number order by se.academic_year
) as is_enrolled_oct01_next
```

There is no year arithmetic anywhere in it. `lead()` steps to a student's next existing row, not to `academic_year + 1`. When a student has no row for the following year, `lead()` silently reaches across the gap to whatever row comes next and copies that year's value. The `false` default only fires when there is no later row at all.

Student A shows the pattern. Student A has a row for every year from `academic_year` 2014 through 2024, no row for 2025, and a row for 2026. On the 2024 row `lead()` reaches the 2026 row, so `is_enrolled_oct01_next` is true. Student A was not enrolled in 2025-26.

I counted the rows that carry `is_enrolled_oct01_next = true` and have no row at `academic_year + 1`.

| Academic year | Newark | Camden |
| --- | --- | --- |
| 2023 | 55 | 18 |
| 2024 | 28 | 11 |
| 2025 | 0 | 0 |

The current year is clean because the defect needs a later row to reach. A student has to come back before the field can borrow a value from the wrong year. So the error in a closed year grows after that year closes, and the dashboard's history gets quieter over time.

Correcting the field raises reported attrition. For 2024-25 Newark rises 0.44 points and Camden rises 0.48 points. For 2023-24 Newark rises 0.87 points and Camden rises 0.81 points. The 2023-24 understatement is about twice the 2024-25 understatement, so correcting both shrinks the reported year-over-year rise in attrition by about 0.4 points.

Every other `lead()` column in the model shares the defect: `next_schoolid`, `next_exitdate`, `next_exitcode`, `next_exit_code_kf`, `next_exit_code_ts`, `next_exitcomment` and `is_enrolled_oct15_next`.

## Steps to reproduce

1. Run the first query in the "For Claude" fold-out. It lists the students that have a row at `academic_year = 2024`, no row at 2025, and a row at 2026.
2. Read one of those students' 2024 rows in `rpt_tableau__ops_dashboard`. `is_enrolled_oct01_next` is true and `next_exitdate` falls in 2027.
3. Read the same student's 2023 row. `next_exitdate` falls in 2025, which is 12 months after that row's own `exitdate`. The 24-month gap on the 2024 row is the visible tell.

## Where

- **Code location / dbt project:** `rpt_tableau__ops_dashboard` in `kipptaf`, the `lead()` expressions in the final `select`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** Operations "Attrition Analysis" dashboard, Exit Codes view

<details>
<summary>For Claude</summary>

Found while reconciling the student persistence report (`ldesimon_dev.student_retention_rollups`) against this dashboard for 2024-25. The two disagree by 39 students across Newark and Camden, and this defect accounts for all 39. `ldesimon_dev.student_retention_base` already computes the flag the year-aware way and its inline comment explains why, so that model is the reference implementation:

```sql
-- YEAR-AWARE, not lead(). A student with no AY+1 row at all scores false
-- here; lead() would hand them the AY+2 flag and credit them as retained.
coalesce(n1.is_enrolled_oct01, false) as is_enrolled_oct01_next,
```

Suspected fix: replace each `lead()` with a self-join on the next academic year. Do not add `_dbt_source_project` to the join key — a student who moves between source projects should still match, which is how `student_retention_base` does it.

```sql
left join
    `teamster-332318`.`kipptaf_extracts`.`int_extracts__student_enrollments` as n1
    on se.student_number = n1.student_number
    and n1.academic_year = se.academic_year + 1
    and n1.rn_year = 1
```

Query 1, the affected rows. Fixed means this returns 0 rows for every year, not just the current one:

```sql
with se as (
  select student_number, academic_year
  from `teamster-332318`.`kipptaf_extracts`.`int_extracts__student_enrollments`
  where rn_year = 1 and is_enrolled_oct01
)
select o.academic_year, o.region, count(*) as falsely_flagged_retained
from `teamster-332318.kipptaf_tableau.rpt_tableau__ops_dashboard` as o
where o.is_enrolled_oct01
  and o.is_enrolled_oct01_next
  and not exists (
    select 1 from se as n
    where n.student_number = o.student_number
      and n.academic_year = o.academic_year + 1
  )
group by o.academic_year, o.region
order by o.academic_year, o.region
```

Query 2, the cheap screen a dashboard user can run without changing any SQL: `date_diff(next_exitdate, exitdate, day) > 400`. Treat it as a screen, not a test. It returned 40 Newark rows for `academic_year` 2024 against 28 true positives, because a mid-year exit inflates the gap on its own.

The roster of affected students stays local. It holds student numbers and exit dates, so I am not attaching it here.

Two adjacent data-quality findings from the same reconciliation, both at KIPP Cooper Norcross High and both worth separate tickets if Camden confirms them:

- A grade 12 record carries `exitcode = 'G1'` and `enroll_status = 3` with an `exitdate` of 2024-12-05, about 6 months before the last day of school.
- A different grade 12 record has an `exitdate` of 2025-06-30, `enroll_status = 2`, and no `exitcode` at all.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make the `next_*` columns on `rpt_tableau__ops_dashboard` and the `next_year_school` / `next_year_schoolid` columns on `int_extracts__student_enrollments` read the student's row for the following academic year, instead of the student's next row whatever year it belongs to.

Both models built those columns with `lead()`. `lead()` steps to a student's next existing row. A student who left for a year and came back has no row for the missing year, so `lead()` reached past the gap and copied the values from the year the student returned in. The dashboard then counted that student as enrolled the following October when the student was not.

On the ops dashboard, 595 rows across academic years 2005 through 2024 carry `is_enrolled_oct01_next = true` for a year the student was not enrolled. That understates attrition on the Operations "Attrition Analysis" dashboard. Corrected attrition goes up. All 8 `next_*` columns had the same defect, so all 8 are fixed.

On `int_extracts__student_enrollments`, 867 of 132,847 rows carry a wrong `next_year_school` / `next_year_schoolid`. Every one of those 867 changes from a non-null value to null — no row swaps one wrong school for another. 29 of them lose a false `999999` graduated-next-year placeholder, which is what `int_students__retention_over_time` reads.

The fix in both models is a left join to a new `next_year` CTE keyed on `academic_year - 1`. A student with no following-year row now gets no match. On the dashboard, `is_enrolled_oct01_next` and `is_enrolled_oct15_next` coalesce to `false`, which keeps the default the old `lead()` calls had. Every other column goes null, which is what `lead()` gave at the end of a student's history anyway.

This also adds a scoped uniqueness test on `int_extracts__student_enrollments`, asserting that `(student_number, academic_year)` is unique where `rn_year = 1`. That is the invariant the new join relies on, and the model's declared key is wider, so nothing asserted it before.

Refs #5430

## Reviewer Notes

- The join key is `student_number` and `academic_year` only, with no `_dbt_source_project`, so a student who moves between regions still matches. Prod carries 0 duplicate groups on that key today, and the new test keeps it that way.
- The new test is `severity: warn`, not `error`. A genuine cross-region mid-year transfer would trip it, and the host table rebuilds hourly, so it should surface rather than halt a build.
- The second fix, on `int_extracts__student_enrollments`, was finding 2 of the `claude-review` pass. I first proposed deferring it to its own issue; Charlie asked for it in this PR instead. Its measured prod impact and the positional-binding check are in the fold-out.
- That model's `next_year` CTE reads the local `enrollments` CTE, not `base_powerschool__student_enrollments`. The base is a compatibility passthrough scheduled for removal under #3999, so a new consumer would move that migration backwards, and `enrollments` is a pure projection of it — same rows, four extra derived columns, no filter and no join. Reading it there makes the join's right side the same population as its left side by construction, and drops the compiled model from two references to that base down to one. `int_students__student_enrollments` is a view, so each reference re-expanded its two-branch union and twelve left joins; one reference plus a local CTE cannot expand more than two references did, whatever BigQuery decides about materializing the CTE.

### General

### dbt

- [x] Include a `[model_name].yml` properties file for all new models — no new models. Both existing properties files gain a description on each column whose meaning this change corrects.
- [x] Include (or update) an
      for all models consumed by a dashboard, analysis, or application — the exposure already exists and needs no change.
- [x] **Breaking change?** No. No column is renamed or removed, and the names and types are identical. In both models the touched columns move earlier in the select list. `rpt_tableau__ops_dashboard` is contract-enforced and a contract matches by name, not position; `int_extracts__student_enrollments` sets no contract at all. No consumer binds either model positionally — see the fold-out.
- [x] If adding or modifying an external source, run `stage_external_sources` with **`--target staging`** — no external source is touched.

## CI checks

- [x] **Trunk** — passes. CodeQL passes too.
- [x] **dbt Cloud** — passes on the head commit, run 70403225708770, which built 351 relations between 17:16 and 17:18 UTC. The earlier run 70403225705926 reported 9 test warnings, all pre-existing — see the fold-out.
- [x] **Dagster Cloud** — not triggered; no Dagster paths change.

## Troubleshooting

<details>
<summary>For Claude</summary>

Verified against prod, on the full population rather than a sample:

1. The issue's reproduce query still fires on prod, and wider than the issue reported: every academic year from 2005 through 2024, not only 2023 and 2024. Newark, Camden and Miami all appear.
2. Fan-out cannot happen on today's data. `(student_number, academic_year)` at `rn_year = 1` is unique over all 132,848 rows of `int_extracts__student_enrollments` — the grouped-having query returns 0 duplicate groups across all history. The same check against `base_powerschool__student_enrollments` — the relation the second fix's join reads, through the `enrollments` CTE that projects it unchanged — also returns 0 duplicate groups and 0 duplicate rows. That is why `_dbt_source_project` is correctly absent from both join keys: omitting it is what lets a genuine cross-region mover still count as enrolled next year. The new scoped test makes that a standing assertion rather than a point-in-time check.
3. Simulating the dashboard's new join against prod returns 132,848 rows for 132,848 rows in, equal to the current prod dashboard, and the reproduce query then returns 0. Rows with `is_enrolled_oct01 and is_enrolled_oct01_next` drop from 78,495 to 77,900, a delta of exactly 595, which reconciles to the sum of the per-year reproduce counts.
4. Simulating the second fix against prod, computing the old `lead()` value and the new join value side by side over all 132,847 rows: 867 rows change `next_year_schoolid`, 867 change `next_year_school`, and 867 go from non-null to null. Those three counts being equal is the proof — every change is a wrong value being withdrawn, never one wrong value replaced by another. The `999999` graduated placeholder falls from 21,563 rows to 21,534, so 29 rows lose a false graduated-next-year credit in `int_students__retention_over_time` line 160.
5. `dbt compile --target prod` resolves both models. The ops dashboard's compiled SQL passes a BigQuery dry run at 1,126,110,953 bytes; the `bq` CLI's own credentials expired before the second model could be dry run, so that one is validated by compile plus the dbt Cloud CI `--full-refresh` build. `dbt parse --no-partial-parse` is clean and confirms the new test node carries `where: rn_year = 1` and `severity: warn`.

The row counts 132,848 and 132,847 come from queries run about an hour apart against a table that rebuilds hourly. They are the same population, not a reconciliation error.

Positional-binding check for the `int_extracts__student_enrollments` reorder, since BigQuery binds `UNION ALL` branches by position: 112 models `ref()` this one, and only 4 use `select *` anywhere — `int_kippadb__roster` (1 star, no union), `rpt_tableau__crdc_roster` (6 stars, 15 unions), `rpt_gsheets__deanslist_mod_audit` (3 stars, 3 unions), `rpt_gsheets__njdoe_universal_screener_data` (2 stars, 1 union). Every one of those stars reads a local CTE, not this model. The only union over stars, in `rpt_gsheets__deanslist_mod_audit`, unions three CTEs that each enumerate their columns. The remaining 108 consumers project columns by name.

The 9 dbt Cloud test warnings are all pre-existing. The second commit widened `state:modified+` from a leaf report to a hub model with 112 consumers, so CI built models the earlier runs on this branch never touched, and their standing warn-tests surfaced. Seven are uniqueness warnings whose duplicate-group counts match prod exactly: `rpt_gsheets__deanslist_mod_audit` 798, `rpt_tableau__gradebook_gpa` 159, `rpt_tableau__kfwd_career_launch_survey` 128, `rpt_tableau__student_course_grades` 109, `rpt_gsheets__csgf_hs_postsec_pathways` 69, `rpt_appsheet__kfwd_career_launch_survey` 8, `rpt_tableau__ap_assessment_dashboard` 3. The `rpt_clever__enrollments` orphan test matches prod at 18. The ninth, `int_students__graduation_path_codes__scores_have_cutoffs`, reports 3 in CI against 7 in prod — CI is lower because this branch carries the merged grad cut-score fixes b0f415e307, e760cc771e and 927d6e8daa that prod has not rebuilt with yet. None of the 9 touches `next_year_school` or `next_year_schoolid`, and this change alters no row counts, only 867 values from non-null to null. The head run reports 0 warnings, but that is a reporting artifact rather than a real drop: re-querying the earlier run now also returns 0, where it returned 9 at the time. The head run is confirmed to have built by relation `creation_time` in `region-us.INFORMATION_SCHEMA.TABLES` filtered to `dbt_cloud_pr_%5439%` — 351 relations inside its 3-minute window.

No test was added on `rpt_tableau__ops_dashboard` itself, deliberately. After this change the invariant there is structural: `is_enrolled_oct01_next` can only be true when an `academic_year + 1` row matched the join. A test asserting it cannot fail unless someone rewrites the join, and `.claude/rules/dbt-yaml.md` bars a test that cannot fail by construction. The test that WAS added is a different claim: the upstream grain, which can fail and is not asserted anywhere else.

The reference implementation the issue names, `ldesimon_dev.student_retention_base`, is a BigQuery dataset — an analyst's dev copy, not a model in this repo. A grep of `src/dbt` for it correctly finds nothing.

Out of scope: `rpt_tableau__ops_dashboard.yml` carries no `contains_pii` tags despite `student_number`, name, ethnicity, gender and IEP/504/lunch status. Pre-existing. The columns this PR touches are consistent with their already-untagged siblings, so this change does not widen the gap.

The 2 adjacent Camden data-quality findings in the issue's fold-out are not addressed here. Both are single grade 12 records at KIPP Cooper Norcross High with inconsistent exit data, and the issue says they are worth separate tickets if Camden confirms them.

</details>


cbini:
Verdict per finding. Code changes are in `7180e20bf4`.

**1. Possible join fan-out on cross-region transfers — accepted, verified clean, guard added.**

Ran the exact query against prod. It returns 0 duplicate groups over all history:

```
dup_groups: 0   dup_rows: 0   worst_group: 0
```

So the fan-out has never happened. The structural point stands though: the model's declared key is `(_dbt_source_project, student_number, academic_year, entrydate)`, so nothing asserted the narrower invariant my join relies on. Added the standing test you suggested to `int_extracts__student_enrollments`, scoped `where: rn_year = 1`:

```yaml
- dbt_utils.unique_combination_of_columns:
    arguments:
      combination_of_columns:
        - student_number
        - academic_year
    config:
      where: rn_year = 1
      severity: warn
```

`warn`, not `error`, on purpose. A genuine cross-region mid-year transfer is a legitimate data event, and that table rebuilds hourly — it should surface, not halt the build. Confirmed the node parses with the right config via `dbt parse --no-partial-parse`.

**2. Same defect in `int_extracts__student_enrollments`, feeding a retention model — confirmed, out of scope here.**

Read the source. `next_year_school_lead` and `next_year_schoolid_lead` at lines 30-36 are `lead(...) over (partition by student_number, rn_year order by academic_year asc)` — identical gap-skipping shape. Grepped the consumers and they match your list: `int_students__retention_over_time` (lines 14, 104-106, 145, 160, 174, 202), `int_extracts__student_enrollments_subjects_weeks` (245-246), and `int_extracts__course_enrollments_by_term`.

Not fixing it in this PR. It is a different model with a wider blast radius, and `int_students__retention_over_time` needs its own before-and-after reconciliation. Proposing it as a follow-up issue rather than opening one unilaterally.

**3. Avoidable ST06 suppression — accepted, fixed.**

You are right that the contract matches by name, not position, so the reorder was available and my stated rationale did not hold. Moved the 6 plain `nxt.*` refs up to sit with the other plain refs, right after the `t.*` group. `trunk check --force --no-fix` now returns "No issues" with the `trunk-ignore` deleted. Re-verified after the change: `dbt compile --target prod` resolves, and the compiled SQL passes a BigQuery dry run at 1,126,110,953 bytes.

**4. `student_retention_base` citation — not a repo model, and the citation is accurate.**

`ldesimon_dev.student_retention_base` is a BigQuery dataset, the analyst's own dev copy named in #5430 as the reference implementation. A `grep` of `src/dbt` correctly finds nothing. Reworded the PR body to name it in full so it does not read as a repo model.

**5. Checked-clean list and the PII note — agreed on both.**

The PII gap on `rpt_tableau__ops_dashboard.yml` is real and pre-existing. The 2 columns this PR touches are booleans consistent with their already-untagged siblings, so this change does not widen it. Leaving it out of scope rather than starting a tagging sweep inside a bug fix.

cbini:
@claude the per-finding verdicts are in [the comment above](https://github.com/TEAMSchools/teamster/pull/5439#issuecomment-5764086256). Two findings changed the code, in `7180e20bf4`, which you have not seen — please re-review that commit.

What changed since your review:

- Finding 3 accepted. I deleted the `trunk-ignore(sqlfluff/ST06)` and moved the 6 plain `nxt.*` refs into the plain-ref group instead. You were right that a contract matches columns by name, not position, so the reorder was available and the suppression was not needed. `trunk check --force --no-fix` is clean on all 3 files, `dbt compile --target prod` resolves, and the BigQuery dry run is valid at 1,126,110,953 bytes.
- Finding 1 accepted and guarded. Prod carries 0 duplicate groups on `(student_number, academic_year)` at `rn_year = 1` over all 132,848 rows, so the fan-out you asked about has never happened — but nothing asserted it, so I added `dbt_utils.unique_combination_of_columns` scoped `where: rn_year = 1` on `int_extracts__student_enrollments`. It is `severity: warn`, not `error`, because a genuine cross-region mid-year transfer is a real data event and that model rebuilds hourly.
- Finding 2 confirmed, declined here. `int_extracts__student_enrollments.sql:30-36` carries the identical `lead()` defect in `next_year_school_lead` / `next_year_schoolid_lead`, feeding `int_students__retention_over_time`. Different model, wider blast radius, needs its own reconciliation. Proposed as a follow-up issue.
- Finding 4 rebutted. `ldesimon_dev.student_retention_base` is a BigQuery dev dataset named in issue 5430, not a model in this repo, so a grep of `src/dbt` correctly finds nothing.
- Finding 5 agreed, pre-existing and out of scope.

The PR body now matches the code.

cbini:
**Finding 2 — reversing my earlier verdict. Now fixed in this PR, commit 0617059f95.**

I previously said the `lead()` defect in `int_extracts__student_enrollments` would go to a separate issue because of its wider blast radius. Charlie asked for it here instead, so it is fixed here.

`next_year_school` and `next_year_schoolid` were built with `lead(...) over (partition by student_number, rn_year order by academic_year asc)` and then broadcast across the student's other rows for that year. Both now come from a left join to a `next_year` CTE keyed on `academic_year - 1`, the same shape as the ops dashboard fix.

Measured against prod over all 132,847 rows: 867 rows change `next_year_schoolid`, 867 change `next_year_school`, and 867 go from non-null to null. Those three counts being equal is the proof of correctness — every change withdraws a wrong value, and no row swaps one wrong school for another. The `999999` graduated placeholder falls from 21,563 to 21,534, so 29 rows lose a false graduated-next-year credit in `int_students__retention_over_time`.

`base_powerschool__student_enrollments` at `rn_year = 1` carries 0 duplicate groups on `(student_number, academic_year)` across all history, so the new join cannot fan out.

The 2 columns move earlier in the select list. This model sets no contract, and a contract matches by name rather than position regardless. I checked positional binding, since BigQuery binds `UNION ALL` branches by position: of the 118 models that `ref()` this one, only 4 use `select *` anywhere, and every one of those stars a local CTE rather than this model. The one union over stars unions three CTEs that each enumerate their columns.

The properties file gains a description on both columns. `trunk check --force --no-fix` is clean on both changed files, `dbt compile --target prod` resolves, and `dbt parse --no-partial-parse` is clean. The `bq` CLI's own credentials expired before I could dry-run the compiled SQL, so the real build validation is the dbt Cloud CI `--full-refresh` run on this commit.

Note for the reviewer run already in flight: it was triggered against `7180e20bf4` and will not see this commit.

cbini:
@claude re-review request for commit `0617059f95`, plus verdicts on your re-review of `7180e20bf4`.

**Your re-review: accepted, nothing to change.** Findings 1 and 3 confirmed fixed, and your non-blocking observation on `severity: warn` is a fair statement of the tradeoff — the test detects the condition rather than preventing the fan-out, and one hourly build could ship duplicated rows before anyone reads the warning. I am keeping `warn`. A genuine cross-region mid-year transfer is legitimate data, and halting an hourly build on it is worse than one build of duplicated dashboard rows.

Your addendum's last point is now moot in the best way. You noted that `int_extracts__student_enrollments.sql` lines 30-36 and 69-75 build `next_year_school_lead` / `next_year_schoolid_lead` with windows partitioned by `student_number` alone, and suggested folding that into the finding-2 follow-up issue. There is no follow-up issue: Charlie asked for finding 2 in this PR, so commit `0617059f95` deletes both `lead()` windows and both `max(if(rn_year = 1, ...))` broadcast windows outright and replaces them with a left join to a `next_year` CTE keyed on `academic_year - 1`, the same shape as the ops dashboard fix.

**What I would like you to check in `0617059f95`:**

1. The new join reads `base_powerschool__student_enrollments` at `rn_year = 1`, a different upstream than the dashboard's `next_year` CTE. I added no second uniqueness test there, on the reasoning that one is already covered: `enrollments` reads `base_powerschool__student_enrollments` with no joins, `windowed` reads `enrollments` through window functions only, and the final select is `from windowed as e`. The grain is 1:1 from base, so a duplicate at `(student_number, academic_year)` where `rn_year = 1` in base propagates into `int_extracts__student_enrollments` at `rn_year = 1` and trips the test I added in `7180e20bf4`. Please check that chain rather than take my word for it, and say if you think a second test is warranted anyway.
2. The 2 columns move earlier in the select list, from prod positions 125 and 126 to just after `potential_credits_cum`. This model sets no `contract:` key. I checked positional binding because BigQuery binds `UNION ALL` branches by position: of the 118 models that `ref()` this one, only 4 use `select *` anywhere, and every one of those stars a local CTE, not this model. Worth an independent trace, since the blast radius here is much larger than the dashboard's 2 consumers.
3. The properties file gains a description on both columns. The `next_year_schoolid` one mentions the `999999` graduate placeholder; check it reads correctly against how `int_students__retention_over_time` uses that value.

Measured against prod before the commit: 867 of 132,847 rows change, and `schoolid_changed`, `school_changed` and `schoolid_now_null` are all exactly 867 — every change withdraws a wrong value, and no row swaps one wrong school for another. The `999999` placeholder count falls from 21,563 to 21,534, so 29 rows lose a false graduated-next-year credit. `base_powerschool__student_enrollments` at `rn_year = 1` carries 0 duplicate groups and 0 duplicate rows on `(student_number, academic_year)` across all history.

Local checks: `trunk check --force --no-fix` clean on both changed files, `dbt compile --target prod` resolves, `dbt parse --no-partial-parse` clean. The `bq` CLI's own credentials expired mid-session so I ran no local dry run for this model; dbt Cloud CI run 70403225705926 built it for real and passed. That run carries 9 test warnings, all pre-existing — 8 match prod counts exactly, and the ninth is lower in CI than prod because this branch carries merged grad cut-score fixes that prod has not rebuilt with yet.

If the MCP servers are disconnected for your session again, say so and verify by reading source, the way you did last time. That was the right call and I would rather have it stated than guessed at.

cbini:
@claude Verdicts on your re-review of `0617059f95`, one correction of mine, and one new commit to look at.

**Check 1 (grain chain, no second test needed) — accepted, no action.** Your trace matches mine and reaches the conclusion independently, which is what I wanted from it rather than agreement with my reasoning.

**Check 2 (positional binding) — accepted, and you caught an error of mine.** You counted 112 consumer files; I wrote 118 in the PR body. You are right. `grep -rl` over `src/dbt` returns 112 files and `grep -ro` returns 152 ref occurrences, so 118 was neither figure. I corrected the body to 112, and the dependent "remaining 108 project by name". The conclusion does not move: we independently found the same 4 `select *` files and neither of us found a positional bind.

**Check 3 (`999999` description) — accepted, no action.** You traced the sentinel to `powerschool/CLAUDE.md` and to line 160, which is one hop further than I went.

**Bonus check (broadcast semantics) — accepted, and it closes a gap I had not written down.** The old code broadcast the `rn_year = 1` value across a student's other stint rows for that year, and the new join reproduces that only because the `e` side carries no `rn_year` filter. I verified the behavior but never stated why it holds. It is worth having on the record.

**New commit `4b8c8aa563`, pushed after your review.** Charlie asked whether `base_powerschool__student_enrollments` was the right source for the `next_year` CTE, or whether a purer one existed. It was not the right source. That model is a compatibility passthrough whose entire body is `select *, from {{ ref("int_students__student_enrollments") }}`, and its own description schedules it for removal once its consumers migrate, tracked in #3999 — so the fix added a new consumer to a model the repo is trying to delete.

The commit points the CTE at the local `enrollments` CTE instead, a one-line change. `enrollments` already reads that same base at line 8, adds four derived columns, and neither filters nor joins, so all five columns `next_year` needs are in it and the row set is identical. Two things follow. The join's right side is now the same population as its left side by construction, rather than the 1:1 chain you and I each had to trace by hand. And the compiled model references that base once instead of twice — verified with `grep -c` on the compiled SQL, 1 where it was 2 — which matters because `int_students__student_enrollments` is a view, its `materialized: table` commented out, so the second reference re-expanded a two-branch union and twelve left joins in a model that rebuilds hourly.

`dbt compile --target prod` resolves it and `trunk check` is clean. No column, row or value changes.

Three things worth your eyes on that commit, briefly, since it is one line. Whether reading `enrollments` rather than `windowed` is right — I chose `enrollments` because `windowed` adds four window functions the CTE does not need, but `windowed` is the literal left side of the join. Whether the plan-depth claim holds, given BigQuery does not guarantee CTE materialization and may re-evaluate `enrollments` at both reference sites anyway; I claimed only that it cannot be worse, and I did not measure it. And whether the two remaining `base_powerschool__student_enrollments` references in this model, at the `enrollments` CTE and in the ops dashboard's lineage, should migrate to `int_students__student_enrollments` in this PR or in a #3999 follow-up — my read is follow-up, because that is a migration rather than a bug fix.

cbini:
Verdicts on the re-review of `4b8c8aa563`. No findings to act on, so this records the three answers plus one thing I checked after you posted.

**Q1 (source CTE) — agreed, no change.** Same read as mine: `enrollments` is a pure projection, and `windowed`'s four window functions are overhead `next_year` does not need.

**Q2 (plan depth) — accepted, and I tightened the PR body.** Your argument settles the hedge I left. The old code made 2 references to a persisted view, which BigQuery inlines per reference with no sharing, so 1 view reference plus a local CTE is parity in the worst case and strictly better in the best case. The Reviewer Notes bullet now carries that reasoning instead of a bare "cannot be worse".

**Q3 (#3999 migration) — agreed, follow-up.** Your count of 14 other files is the stronger form of my own argument, so the remaining reference stays put.

**On your environment note.** CI is green on `4b8c8aa563`: all 7 checks pass — CodeQL, Trunk Check, Trunk Check Runner, the 3 Analyze jobs, and dbt Cloud.

I also chased something your note could not reach. The dbt Cloud head run 70403225708770 reported 0 test warnings where the previous run reported 9, which reads as a run that never built rather than as good news. It is a reporting artifact: re-querying the earlier run now also returns 0, where it returned 9 at the time. The head run did build — 351 relations created between 17:16 and 17:18 UTC, inside its 17:15:52 to 17:18:55 window, per `creation_time` in `region-us.INFORMATION_SCHEMA.TABLES` filtered to this PR's schemas. The PR body's CI line and the fold-out both say this now.

Issue #4774, opened 2026-08-07
perf(dbt): stale branches fan out state:modified+ and run the full test suite on PRs that change no dbt code


cbini:
## Symptom

A long-lived branch accumulates phantom `state:modified` selections proportional to how far `main` has advanced since its last merge. The PR's own diff is irrelevant. A docs-only PR can end up building several hundred models and running 600 plus tests per push.

## Evidence

PR #4673 (publish the Tableau permissions guide) changed 7 files, none under `src/dbt/`: 4 markdown docs, `mkdocs.yml`, root `CLAUDE.md`, and `.claude/context/tableau.md`.

Its per-invocation dbt selection, from `region-us.INFORMATION_SCHEMA.JOBS_BY_PROJECT`, against its commit history:

| When | Models selected | Branch event |
| --- | ---: | --- |
| Jul 31 20:24 | 7 | PR opened, fresh off `main` |
| Aug 3 18:56 to 19:26 | 284 (x3) | 3 docs commits, no `main` merge since Jul 31 |
| Aug 4 | 384, then 227 | 2 docs commits, still no merge |
| Aug 5 15:18 to 17:56 | 384 (x10) | 10 docs commits, still no merge |
| Aug 5 19:25 | -- | first `origin/main` merge (113c129ad) |
| Aug 6 15:18 | -- | second `origin/main` merge (a2ec3b969) |
| Aug 6 16:56 to 17:29 | 2 (x5) | back to normal |

The selection grew from 7 to 384 purely as `main` advanced under a branch that had not merged it, then collapsed to 2 immediately after two `main` merges.

Cost of that one PR over three days, on-demand at 6.25 USD per TiB billed:

- Aug 3: 3 runs at 0.78 USD
- Aug 4: 0.87 plus 0.60 USD
- Aug 5: 10 runs totalling 8.11 USD
- Total: about 11.90 USD for zero dbt changes

Model builds were nearly free (1.13 USD) because most marts are views and `create view` bills nothing. The spend is the tests attached to the fanned-out graph: 6.94 USD of the 8.07 USD on Aug 5, or 86 percent.

## Scope

[...]

PR #5465, opened 2026-09-21
feat(dbt): add a full-grain PowerSchool terms spine

### LINKED ISSUE #5397: fix: int_students__terms has no Miami quarters for 2 schools before AY2025
## What's happening

`int_students__terms` gets Miami's quarter terms from Focus alone. Focus carries quarters for 3 schoolids — 30200803, 30200804 and 30200805 — in academic years 2018 through 2024. It covers all 7 Miami schools only from AY2025.

Miami's frozen PowerSchool archive holds attendance days for 4 schoolids in those years. Two of them, 30200801 and 30200802, have no Focus quarter row at all before AY2025.

PR #5396 turned the term join in `int_students__attendance_daily` into a left join, so those days are no longer dropped. They now carry a null `term` and a null `semester` instead. 492,482 Miami membership days are affected, which is 26 percent of the 1,864,931 days that PR recovered network-wide.

The archive already holds the quarters that would fill almost all of them. Measured against prod on 2026-09-17:

| Coverage of the 492,482 days | Rows |
| --- | --- |
| An archive quarter covers the day | 491,971 |
| Nothing covers the day | 511 |

Two code comments and one source description on `main` say the opposite. Each states that Focus already supplies Miami terms across the whole archive range. PR #5398 corrects that text. It does not fix the gap.

## This needs a design change before any code

Do not fix this by adding `source("kippmiami_powerschool", "int_powerschool__terms")` to the kipptaf union wrapper. That fans out 793,259 days. Three decisions have to be made first, and they interact, so they belong in a spec rather than in a PR review.

**1. Where the precedence rule lives.** `int_students__terms` does not merge its two sources — it `UNION ALL`s them as independent arms. The PowerSchool arm full-joins `stg_powerschool__terms` to `int_powerschool__terms` and conforms the result; the Focus arm conforms `stg_focus__marking_periods` separately and pads every PowerSchool-only column with a typed null. The rule this fix needs — take an archive quarter only where Focus has none for that schoolid and academic year — is an anti-join *across* those two arms. The current shape has nowhere to put it. Either the arms get a shared pre-union spine, or Miami's precedence resolves upstream in the kipptaf wrapper before `int_students__terms` ever sees it. That choice drives everything else.

**2. The order of the wiring and the archive rebuild.** The frozen archive `int_powerschool__terms` predates #5396 and is missing 16 quarters worth 85,702 days. Wiring first leaves those days null and looks like a partial fix. Rebuilding first puts the overlapping rows in reach before the precedence rule exists. The rebuild also is not self-service: it needs the `powerschool` package re-included in `src/dbt/kippmiami/packages.yml` per `src/dbt/kippmiami/CLAUDE.md`, then removed again, and the prod build belongs in a human's terminal.

**3. The 511 residual days.** Decide whether they stay null with a comment at the derivation site, or get filed as a PowerSchool source defect.

Run `superpowers:brainstorming` and land a spec under `docs/superpowers/specs/` before touching a model.

## Steps to reproduce

1. Run the first query in the "For Claude" section. It lists which Miami schoolids have Focus quarters in each academic year. Schools 30200801 and 30200802 appear only in AY2025 and AY2026.
2. Run the second query. It joins the archive's attendance days to the archive's own quarters. The `n_days_no_term` column totals 511 across all school-years.
3. Compare that to the same join against Focus quarters, which leaves 492,482 days uncovered.

## Where

- **Code location / dbt project:** `kipptaf`, `src/dbt/kipptaf/models/students/intermediate/int_students__terms.sql` and `src/dbt/kipptaf/models/powerschool/sources-kippmiami.yml`
- **Environment:** prod
- **Run, PR, or dashboard link (if any):** found while reviewing #5396. Refs #4750 item 3 and #5390. Comment-only correction in #5398.

<details>
<summary>For Claude</summary>

### History

#4750 item 3 planned to wire 5 Miami archive relations into kipptaf: `int_powerschool__calendar_day`, `int_powerschool__calendar_week`, `int_powerschool__terms`, `int_powerschool__student_enrollment_union` and `stg_powerschool__terms`. The design doc is `docs/superpowers/specs/2026-09-17-miami-archive-and-warn-pins-design.md`.

#5395 shipped the 2 calendars. The other 3 got an exclusion comment instead. The stints exclusion is correct — Focus is Miami's sole enrollment source. The 2 terms exclusions rest on the false coverage claim.

#5390 said to re-run its reproduce query after item 3 merged and restate Miami's number. Item 3 merged without terms, both issues closed, and the re-run never happened.

### This is not a plain union

The 5395 design doc asserts "Date ranges are disjoint, so nothing double-counts." True for the calendars. False for terms.

Focus and the archive overlap on school-years: Focus covers 30200803, 30200804 and 30200805 from AY2018, while the archive covers 30200803 from AY2020 and 30200804 from AY2021. Their quarter dates disagree inside the overlap. For 30200803 in AY2022, Focus Q1 runs 2022-08-10 through 2022-10-15 and the archive Q1 runs 2022-08-01 through 2022-10-30.

Joined side by side against the archive's attendance days:

| Both sources match the day | Archive only | Focus only |
| --- | --- | --- |
| 793,259 | 492,482 | 0 |

So adding `source("kippmiami_powerschool", "int_powerschool__terms")` to the kipptaf wrapper would match 2 quarters on 793,259 days and fan out `int_students__attendance_daily`, which joins on a date range.

Focus matches nothing the archive misses. The archive is a strict superset across the archive range. The fix therefore needs a precedence rule — take an archive quarter only where Focus has none for that schoolid and year — which is the same anti-join shape #5396 just built for `termbins` against raw `terms`.

### Why the precedence rule has no home in the current model

`int_students__terms` is two `UNION ALL` arms over a shared column list, not a merge:

- The PowerSchool arm is `powerschool_canonical` (`stg_powerschool__terms` filtered to `rn = 1`) FULL JOINed to `powerschool_quarters` (`int_powerschool__terms`) on schoolid, yearid, abbreviation and `_dbt_source_project`. The full join is deliberate — a handful of pre-2018 non-instructional schoolids carry quarters in `int_powerschool__terms` with no raw `terms` row, and a left join would drop their dates.
- The Focus arm is `focus_conformed` over `focus_marking_periods`, floored at `mp.syear >= 2018` and filtered to `type in ('year', 'semester', 'quarter')`.

Miami is absent from BOTH kipptaf wrappers the PowerSchool arm reads, so no Miami row enters that arm at all today. Adding it there makes the precedence rule a cross-arm anti-join, which the union shape cannot express. The two candidate designs to weigh in the spec:

1. Resolve precedence in the kipptaf `int_powerschool__terms` wrapper — union Miami's archive in, then suppress archive rows for any schoolid and academic year Focus already covers. Needs the wrapper to read Focus, which crosses a layer boundary it does not cross today.
2. Give `int_students__terms` a shared pre-union quarter spine keyed on schoolid and academic year, resolve Focus-over-archive there, and keep both arms projecting from it.

Option 2 is the larger refactor and touches the NJ regions' behavior; option 1 is narrower but puts Focus knowledge in a powerschool-namespaced model. Neither is obviously right, which is the reason this is a spec and not a PR.

### The archive needs a rebuild, after the precedence rule is written

The archive's `int_powerschool__terms` relation is frozen at 72 rows, last written 2026-09-10, and predates #5396. Nothing reads it today: it is not among the 14 tables in `sources-kippmiami.yml`, and none of those 14 models refs it.

Simulating #5396's fallback against the archive's own staging tables adds 16 quarters — 30200801 in AY2018, AY2019 and AY2023, and 30200802 in AY2023. The AY2018 and AY2019 pair covers 85,702 of the 492,482 days. Without a rebuild the frozen relation covers 406,269 and those 85,702 stay null.

The rebuild recipe is the `dbt_project.yml` `powerschool:` block described in `src/dbt/kippmiami/CLAUDE.md`. The package is absent from `src/dbt/kippmiami/packages.yml` today, so no rebuild is in flight. The prod build is not Claude's to run.

### The 511 residual days

School 30200801 in AY2018, on 2019-06-12, 2019-06-13 and 2019-06-14. Q4's `lastday` is 2019-06-11 while the year record `18-19` runs to 2019-06-28, so PowerSchool's own quarter stops 3 instructional days short of its own year. No fallback that reads `terms.lastday` can close this. Leave the 511 null and say so in a comment, or file it as a source defect.

### Reproduce queries

Focus coverage by year:

```sql
select academic_year, string_agg(distinct cast(schoolid as string) order by cast(schoolid as string)) as schoolids
from `teamster-332318.kipptaf_students.int_students__terms`
where _dbt_source_project = 'kippmiami' and term is not null
group by academic_year
order by academic_year
```

Archive coverage of the uncovered days, with the #5396 fallback applied:

```sql
with termbins_quarters as (
  select t.schoolid, t.yearid, tb.storecode as term,
    tb.date1 as term_start_date, tb.date2 as term_end_date
  from `teamster-332318.kippmiami_powerschool.stg_powerschool__terms` as t
  inner join `teamster-332318.kippmiami_powerschool.stg_powerschool__termbins` as tb
    on t.id = tb.termid and t.schoolid = tb.schoolid
    and tb.storecode in ('Q1','Q2','Q3','Q4')
  where t.isyearrec = 1 and t.schoolid != 0
),
fallback_quarters as (
  select t.schoolid, t.yearid, t.abbreviation as term,
    t.firstday as term_start_date, t.lastday as term_end_date
  from `teamster-332318.kippmiami_powerschool.stg_powerschool__terms` as t
  left join termbins_quarters as tb
    on t.schoolid = tb.schoolid and t.yearid = tb.yearid and t.abbreviation = tb.term
  where t.abbreviation in ('Q1','Q2','Q3','Q4') and t.schoolid != 0 and tb.term is null
),
all_quarters as (
  select schoolid, yearid, term, term_start_date, term_end_date from termbins_quarters
  union all
  select schoolid, yearid, term, term_start_date, term_end_date from fallback_quarters
),
ctod as (
  select yearid, schoolid, calendardate
  from `teamster-332318.kippmiami_powerschool.int_powerschool__ps_adaadm_daily_ctod`
)
select
  c.yearid + 1990 as academic_year,
  c.schoolid,
  count(*) as n_days,
  countif(q.term is null) as n_days_no_term
from ctod as c
left join all_quarters as q
  on c.yearid = q.yearid and c.schoolid = q.schoolid
  and c.calendardate between q.term_start_date and q.term_end_date
group by academic_year, c.schoolid
order by academic_year, c.schoolid
```

Row counts before and after that join are identical in every school-year, so the archive's own quarters do not fan out.

### What fixed looks like

- A spec under `docs/superpowers/specs/` settles the 3 decisions above and is reviewed before any model changes.
- Miami membership days in AY2018 through AY2025 carry a non-null `term` and `semester`, except the 511 known residual days.
- `int_students__attendance_daily` row count does not rise. The recovered days already exist as rows after #5396; only their `term` and `semester` change from null to a value.
- The model's grain and uniqueness tests still pass, which is what proves the precedence rule did not fan out the 793,259 overlapping days.
- The NJ regions' term rows are byte-identical before and after, whichever design wins.
- The 2 code comments and 1 source description asserting full Focus coverage are gone or corrected (PR #5398).

</details>

cbini:
# Pull Request

> **Write this whole PR in plain language.** Read `.github/PLAIN_LANGUAGE.md`
> and follow it: every sentence does a job, one idea per sentence, active voice,
> common words. Only the "For Claude" fold-out at the bottom is exempt.

## Summary & Motivation

> What changed and why. A reviewer skimming just this section should understand
> the change. Save tradeoffs, edge cases, and verification detail for "Reviewer
> Notes" below.

When merged, this pull request will add one new model to the `powerschool` package: `int_powerschool__terms_spine`.

The new model holds every term row for one district at full grain. It stacks two things that today get joined together downstream. The first is the raw PowerSchool `terms` table, cut down to one row per school, year, and term abbreviation. The second is `int_powerschool__terms`, which is the quarter list with the dates PowerSchool stores grades against. The model stacks them with `union all` and joins nothing, so a quarter shows up as two rows: one raw row carrying the scheduling dates, and one derived row carrying the grade-storage dates. Those two date windows often disagree, which is why joining them loses information.

This is the first of two pull requests for issue #5397. Miami's frozen PowerSchool archive holds quarters that never reach `int_students__terms`, so 492,482 Miami membership days carry no term or semester. Getting those days back needs the term conform to live in the package, where the archive can reach it. That is what this model is. The second pull request wires it into `kipptaf` and is where the recovered days actually show up. Nothing downstream reads this model yet, so merging this changes no existing output.

`int_powerschool__terms` is deliberately untouched. That is the whole reason this is a new model rather than an edit — `int_powerschool__student_course_grades_spine` reads it and must not move.

## Reviewer Notes

> Optional. Name what's worth a second look and why, in a line or two each. Full
> technical detail, exact values, and edge cases belong in "For Claude" below.
> Delete if there's nothing here.

The model stacks its two inputs instead of joining them, and that is a real change in shape rather than a tidy-up. Today's downstream full join collapses a matched raw record and its quarter into one row. This model keeps them apart. I measured what that costs and checked every consumer; the detail is in the fold-out.

Both uniqueness tests are scoped with `where:`, because the two branches key on different columns. Quarter rows key on `term` and carry no `abbreviation`; raw rows key on `abbreviation` and carry no `term`. One test over both would have to key on a column that is null for half the model.

## Self-review

### General

No Dagster changes.

- [x] Include (or update) an
      summary above.
      See

Properties file included, with a description on the model and on all 33 columns. No exposure needed — nothing external reads this model, and nothing reads it at all yet. Not a breaking change: this adds a model and changes none, and the package intermediate layer is not contract-enforced. No external sources touched.

## CI checks

Expect the kipptaf dbt Cloud CI job to select no models. This pull request touches only the `powerschool` package and the planning documents, and CI is scoped to the kipptaf project. A green CI run here is not evidence the new model is correct — the local verification below is.

## Troubleshooting

<details>
<summary>For Claude</summary>

> Full technical detail behind the Reviewer Notes flags above — exact values,
> edge cases, full reasoning — plus anything else simplified or cut from Summary
> for plain-language readability, AI involvement, and anything a future
> `@claude` invocation needs. Delete if there's nothing to add.

Claude-assisted, human-directed. The [spec](https://github.com/TEAMSchools/teamster/blob/cbini/fix/claude-miami-archive-terms/docs/superpowers/specs/2026-09-21-miami-archive-terms-design.md) and the [plan](https://github.com/TEAMSchools/teamster/blob/cbini/fix/claude-miami-archive-terms/docs/superpowers/plans/2026-09-21-miami-archive-terms.md) are committed on this branch and are the first two commits here.

### The stack-versus-join measurement, and one correction

An earlier version of the spec claimed the downstream full join in `int_students__terms` produced exactly its two inputs stacked and merged nothing. That claim was wrong, and commit `2c320a69de` on this branch corrects it in the spec and the plan. The error came from comparing the raw-side CTE against this model's raw branch — both 1,925 rows, symmetric difference 0 — and reading that as model equivalence. The live model's output filtered `where term is null` is only 1,085 rows, because a merged row carries a non-null `term`.

What prod actually shows, measured 2026-09-21:

| Measurement | Value |
| --- | ---: |
| Quarter rows, both shapes | 926 = 926, symmetric difference 0 over 9 columns |
| Raw rows, both shapes | 1,925 = 1,925, symmetric difference 0 over 15 columns |
| Rows the full join MERGES | 840 — kippnewark 548, kippcamden 276, kipppaterson 16 |
| NJ `int_students__terms` rows, old versus new | 2,011 → 2,851 |

So the two shapes carry the same values but not the same rows. That is safe on two measurements.

First, the grain holds. No row carries both a null `dcid` and a null `term`, and both uniqueness keys hold over the FULL branch sets rather than only the subsets they are tested on today — 0 duplicate keys on `(schoolid, yearid, abbreviation)` across all 1,925 raw rows, and 0 on `(schoolid, yearid, term)` across all 926 quarter rows. Group per district when re-running this; omitting the district discriminator falsely reports 430 and 161 cross-district collisions.

Second, no consumer reads a merged row. All 9 consumers of `int_students__terms`: 5 read quarter-side columns only, 2 read raw-side only, and 2 (`rpt_tableau__gradebook_gpa`, `rpt_tableau__student_course_grades`) read both families but in separate single-sided `union all` branches. Nothing is at risk on columns or on cardinality. Two worth naming. `int_students__attendance_daily` guards its join with `and t.term is not null` and reads only `term_start_date` and `term_end_date`, so it sees the identical 926-row quarter set. `rpt_tableau__gradebook_gpa` branch 2 filters `isyearrec = 1`, and a year record's `abbreviation` is a year label, never `Q1` — so it was never a merged row, and the quarter branch pads `isyearrec` to null so that branch cannot pick one up.

The residual risk is a consumer the dbt graph cannot see: a hand-written Tableau extract or an ad-hoc query that reads `dcid` and `term` off one row would silently get nulls on one side. Rollback is to restore the full join in `int_students__terms` alone, in the second pull request, leaving this model in place.

### Why `semester` is a typed null on the raw branch

Deliberate. Today's `powerschool_canonical` CTE in `int_students__terms` does not project the raw `semester`, so projecting it here would change values rather than just shape. The quarter branch carries the real `semester`.

### Orphan quarters

86 quarters across the three NJ districts come from `termbins` with no matching raw `Q1`-`Q4` row in `terms`. They are the sole source of term and semester for 1,752,966 enrollment days, and a union preserves them without the orphan case needing any special handling, because a union never attempts a match. The counterfactual: schoolid 133570965 for AY2012 through AY2014 has no orphan quarter, and `term` is null on all 363,537 of its days.

### The row pick

`dbt_utils.deduplicate(relation=ref("stg_powerschool__terms"), partition_by="schoolid, yearid, abbreviation", order_by="dcid desc")`. `dcid` alone is a total order: it carries a unique test at error severity and a `primary_key` constraint in `stg_powerschool__terms`, so no tie-break is needed, and `id` is documented there as not guaranteed unique. Descending takes the most recently inserted record, and explicit `desc` also satisfies the macro's BigQuery constraint, which rejects nulls-last inside `array_agg`.

Every raw key is a singleton in prod today, so the reduction is defensive rather than corrective — and because the dedup guarantees the key, the raw-branch uniqueness test cannot fail while it is in place. The signal therefore moves upstream of the dedup: `stg_powerschool__terms` gains a model-level `dbt_utils.unique_combination_of_columns` on the same three columns at `severity: warn`, so a duplicate raw record surfaces to ops instead of being absorbed silently. `severity: warn` departs from the staging-severity rule in `.claude/rules/dbt-yaml.md`, at @cbini's direction: PowerSchool does not enforce this key, no district violates it today, and the spine still reduces to one row per key, so a duplicate is not a build emergency. Verified it does not fire today — 0 duplicate keys and 0 null `abbreviation` across all three NJ districts, and `dbt test --select stg_powerschool__terms` passes 2 of 2 in each.

### Verification run locally

Unit test and both data tests pass in kippnewark, kippcamden, and kipppaterson. The relation builds at 1,759 rows in kippnewark, which is 1,166 raw plus 593 quarter and matches the prod census exactly. `trunk check --force --no-fix` clean. `int_powerschool__terms` and its properties file are untouched across the whole branch range.

The unit test asserts three behaviors: a quarter present on both sides yields two rows rather than one merged row, the `rn = 1` guard drops a superseded raw record with a higher `id`, and an orphan quarter with no raw record survives. The third assertion does not discriminate this shape from the full join it replaces, since a full join also preserves an unmatched quarter; the first two carry the test.

### What comes next

The second pull request adds a kipptaf `union_relations` wrapper over this model for all four districts, rewrites `int_students__terms` to project it, and moves the Focus branch's floor from `syear >= 2018` to the literal `2026`. It cannot be pushed until this one merges and Dagster materializes the spine in all four district prod datasets, including the Miami archive rebuild. A separate issue will cover the 511 residual days, which need a PowerSchool server correction and a final sync through dlt.

Refs #5397

</details>


src/dbt/powerschool/models/sis/intermediate/int_powerschool__terms_spine.sql
```diff
@@ -0,0 +1,116 @@
...
+            suppresspublicview,
+            whomodifiedid,
+            fiscal_year,
+            schoolid,
+            yearid,
+            academic_year,
+
+            row_number() over (
+                partition by schoolid, yearid, abbreviation order by id, dcid
+            ) as rn,
```
cbini: - is this a real threat? if so, does upstream have a warn test so we can flag it for ops?
- if dcid is truly unique, use that alone. it needs an explicit order by and should be descending.

cbini: Both done in e95eac3973, and the answer to the first bullet is no.

**Is the duplicate a real threat?** Not today. Measured on prod: 0 duplicate `(schoolid, yearid, abbreviation)` keys across all 1,925 raw `stg_powerschool__terms` rows in the three NJ districts, and Miami's 193 archive rows are singletons too. Every key is already a singleton, so the guard corrects nothing that exists.

**Does upstream have a warn test?** No. `stg_powerschool__terms` tests only `unique` on `dcid` at severity error, plus the `primary_key` constraint. Nothing anywhere tests `(schoolid, yearid, abbreviation)`.

The part worth your call: with the dedup in place, this model's own `(schoolid, yearid, abbreviation) where term is null` test at severity error can never fail, because the dedup guarantees the key. So if PowerSchool ever did produce a duplicate, nothing would tell anyone — ops or us. Two ways to change that, and I did neither because both are your call:

1. Drop the dedup entirely. The spine's existing error-severity test then fails loudly the day a duplicate appears, and the model gets 12 lines shorter. Cost: it fails the build rather than warning, and it fails at the spine rather than at the source.
2. Add a `dbt_utils.unique_combination_of_columns` on `stg_powerschool__terms` at severity warn and keep the dedup. That flags ops without breaking a build. Cost: `.claude/rules/dbt-yaml.md` says staging-layer tests must set `severity: error`, so a deliberate warn there is against that rule as written — worth deciding rather than slipping in.

Say which and I will add it here.

**The row pick** now reads `order_by="dcid desc"` — `dcid` alone, explicit, descending, so it takes the most recently inserted record. `dcid` carries a unique test at error severity and a `primary_key` constraint upstream, so it is a total order on its own; `id` added nothing and PowerSchool does not guarantee `id` is unique anyway.

Once the pick was a single key, the manual `row_number()` CTE had no reason to exist, so it is now `dbt_utils.deduplicate()` — which `.claude/rules/dbt-sql.md` names the repo default below the 1M-row and 1-slot-hour gate. That also drops the 28-column passthrough CTE, the `rn` column, and the `rn = 1` filter: 53 lines out, 24 in.

Verified output-neutral: the relation still builds at 1,759 rows in kippnewark (1,166 raw plus 593 quarter), matching the previous build and the prod census exactly. Unit test and both data tests PASS. The unit-test fixture already carried a duplicate key where `id` ascending and `dcid` descending disagree, so it still discriminates the direction — the `expect` row moves from `dcid` 1001 to 1002.

cbini:
## Verdicts on the automated review

One finding accepted and fixed in e95eac3973. Two rejected, each against a command I ran rather than an argument. The three confirmed-clean items I agree with.

### 1. Unit test `expect` rows omit 19 of 33 columns — rejected, disproven

I ran the command the finding asked for, on the branch as reviewed:

```
uv run dbt build --select int_powerschool__terms_spine --target dev --project-dir src/dbt/kippnewark
...
1 of 4 PASS int_powerschool__terms_spine::unit_terms_spine_stacks_without_joining [PASS in 3.82s]
Done. PASS=4 WARN=0 ERROR=0 SKIP=0 NO-OP=0 REUSED=0 TOTAL=4
```

A dbt unit test compares only the columns `expect` lists — it selects that column set out of the actual result. A 14-column `expect` against a 33-column model asserts those 14 and ignores the rest, which is the point: this test exists to prove the two branches stack rather than merge, and the 19 padding columns carry no signal for that.

The rule the finding cites says something narrower than the finding reads it as. `.claude/rules/dbt-yaml.md`: "Every `expect` row must list the **same columns** — dbt builds them as `UNION ALL` and does NOT null-fill omitted keys." Same as each other, not same as the model. It is about the UNION ALL that assembles the `expect` block, and all four rows here list the identical 14 keys. The cited precedents enumerate every column because those models are contract-enforced marts whose tests assert the whole output; this is an uncontracted package intermediate.

### 2. ST06 column-order violation — rejected, the linter does not fire

```
trunk check --force --no-fix --filter=sqlfluff int_powerschool__terms_spine.sql
Checked 1 file
✔ No issues
```

sqlfluff does run on this file: `.trunk/trunk.yaml` enables `sqlfluff@4.3.0`, and its `src/dbt/**` ignore rule excludes every linter *except* sqlfluff, sqlfmt and prettier. So this is a real pass, not an absent linter.

The reason it passes is in the rule the finding quotes: "sqlfluff ST06 buckets `cast()` as a SIMPLE target, not a calculation." A `cast()` sits in the same bucket as a plain column ref, so plain refs adjacent to `cast(null as ...)` columns are not interleaving two buckets — which is what ST06 checks. The rule's following sentence ("put every `cast()` after the plain column refs") orders casts against *other function calls*; there are none in either branch.

Separately, the 33-column order is a fixed positional union alignment, locked as a Global Constraint because the second pull request's `int_students__terms` rewrite projects the same order. Reordering one branch to satisfy a lint that does not fire would mean reordering both branches and that constraint.

### 3. Manual `row_number()`/`rn = 1` reinvents `dbt_utils.deduplicate()` — accepted, fixed

Correct, and the gate makes it unambiguous. `.claude/rules/dbt-sql.md` switches to the ranked-column form only when the dedup input exceeds about 1M rows AND the model costs at least 1 slot hour in the 7-day prod ranking. This input is about 2,000 rows and the model is new, so it clears neither bar.

Fixed in e95eac3973, folded together with @cbini's row-pick feedback on the same lines. The pick is now `order_by="dcid desc"` — `dcid` alone, since it carries a unique test at error severity and a `primary_key` constraint upstream and is therefore a total order by itself. Net 53 lines out, 24 in: the 28-column passthrough CTE, the `rn` column and the `rn = 1` filter all go.

Verified output-neutral: the relation still builds at 1,759 rows in kippnewark (1,166 raw plus 593 quarter), matching both the previous build and the prod census. Unit test and both data tests PASS, trunk check clean on both files.

### 4, 5, 6 — agreed, no action

Altitude, PII and cross-branch `union all` type alignment: all three read the code correctly. Terms are reference data, tier 6 in `.claude/rules/ferpa-pii.md`, so no `contains_pii` tagging — consistent with the untagged `stg_powerschool__terms.yml` this model reads.

src/dbt/powerschool/models/sis/intermediate/int_powerschool__terms_spine.sql
cbini: Correcting my answer above: the dedup is pre-existing, and I should have said so when I laid out the two options.

It lives today in kipptaf's own `stg_powerschool__terms` wrapper, [lines 27-30](https://github.com/TEAMSchools/teamster/blob/main/src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__terms.sql#L27-L30), added in `293dd4c42c` and untouched by this branch:

```sql
row_number() over (
    partition by schoolid, yearid, abbreviation, {{ extract_source_project() }}
    order by id
) as rn,
```

So the spine is not adding a guard — it is relocating one from kipptaf to the package, which is why the original read `order by id` with no tie-break. The "all 1,925 keys are singletons, so this is defensive only" line I reported as a measurement is also already written in that wrapper's comment; I re-ran it against prod, but the claim predates me.

That makes option 1 the stronger of the two, for a reason I missed. The existing comment states the guard's purpose:

> Guards a quarter-grain join against a duplicate raw record for the same school/year/term, so consumers attaching termbins columns see one row per quarter.

That join is the full join in `int_students__terms` — the one this whole change replaces with a `union all`. The guard exists to keep a duplicate raw record from fanning out across the matched quarter. With no join, there is nothing to fan out: a duplicate would produce two raw rows, which is the honest representation of two raw records, and the spine's error-severity uniqueness test on `(schoolid, yearid, abbreviation)` would catch it.

So the guard's stated reason for existing does not survive the refactor. I would drop it, and let the test do the work — that is a smaller diff than carrying either the macro or a new upstream test, and it is the only option that would actually tell anyone a duplicate appeared.

The counter-argument for keeping it: `int_students__terms` today emits one row per raw key, and dropping the guard means a future duplicate changes that model's row count rather than failing a test in the package. If you would rather the network model stay stable and the package absorb the problem, keep the macro as it now stands.

Your call either way — say the word and I will strip it.

cbini: All three directions handled. Two shipped here, one recorded for PR B because it cannot land on this one.

**Warn test on the package staging model** — `2ccbbcb30e`. A model-level `dbt_utils.unique_combination_of_columns` on `(schoolid, yearid, abbreviation)` at `severity: warn`, in `src/dbt/powerschool/models/sis/staging/properties/stg_powerschool__terms.yml`. I checked prod first so it does not warn on day one: 0 duplicate keys and 0 null `abbreviation` in kippnewark (1,166 rows), kippcamden (702) and kipppaterson (57). `dbt test --select stg_powerschool__terms` passes 2 of 2 in each district. One flag: `severity: warn` on a staging test departs from the rule in `.claude/rules/dbt-yaml.md` that staging tests must set `severity: error`. That is the point of the ask, so I am not re-arguing it — just noting it once here and once in the commit message so a later reader does not read it as an oversight.

**Dedup stays in the spine** — no change. The `deduplicate` CTE is exactly as it stands in `e95eac3973`.

**Dropping the dedup from the kipptaf wrapper — PR B, not this one.** `int_students__terms.sql:111` is the only `ref("stg_powerschool__terms")` in kipptaf and the only reader of `rn`. This PR is package-only, so removing the window here fails dbt Cloud CI deterministically on `Name rn not found`. The wrapper edit and the `int_students__terms` rewrite are one atomic change, and that rewrite is Task 4.

That direction also simplifies the cleanup, which I recorded in the plan and the spec in `db8e6a1a13`. Task 6 disabled `stg_powerschool__terms` rather than deleting it *only* because the `rn` window made it more than a passthrough. With the window gone it reads `select *, extract_source_project() as _dbt_source_project` — structurally identical to `int_powerschool__terms`, which the plan already deletes under the #5162 exception. So Task 6 now deletes both wrappers, their two properties files and six source entries, instead of deleting one and disabling the other. Its Step 4 is no longer a disable but a note that the uniqueness coverage now lives upstream in two places: the new warn test, and the spine's two error-severity composite tests.

I corrected the spec too, not just the plan. Change 5 said in so many words to disable `stg_powerschool__terms`, and the spec is what the final whole-branch review compares the diff against — leaving it would have produced a finding on a decision you already made.

cbini:
## What the green dbt Cloud check does and does not prove

Worth one note before approval, because this PR's CI history reads worse than it is.

**The one red run was a race, not this branch.** Run `70403225723787` failed 5 nodes, all `Not found: Table teamster-332318:zz_stg_kipptaf_*` on `stg_google_sheets__reporting__terms`, `stg_google_sheets__topline_enrollment_targets` and `stg_people__student_logins` — none of which this branch touches. `JOBS_BY_PROJECT` shows dbt Cloud's own service account issuing `DROP_TABLE` then `CREATE_TABLE` on each at 20:00:14 through 20:00:28 UTC, inside that run's 19:59:05–20:02:19 window. #5466 merged to main at 19:57:30, 95 seconds before the run started, which fired `Parse - Staging` and the staging rebuild behind those drops. Re-triggered with a fresh build rather than a dbt Cloud retry, which would have replayed the same compiled SQL.

**The fan-out was branch staleness, and merging main fixed it.** Both pre-merge runs built roughly 376 to 391 relations into the PR schema, on a PR that touches zero kipptaf models. That is #4774: 29 commits had landed on main since the merge-base, including `int_extracts__student_enrollments`, `stg_powerschool__storedgrades` and `rpt_tableau__ops_dashboard` — the exact nodes the fan-out ran through. After merging `origin/main`, the run built **0** relations. #4774's first proposed fix, confirmed.

**So the green check is now trivially green.** kipptaf's `packages.yml` imports only `dbt_external_tables` and `dbt_utils` — it does not import the `powerschool` package, so nothing in this PR is reachable from kipptaf and `state:modified+` selects nothing. `int_powerschool__terms_spine` is validated only by the local district build: `dbt build --select int_powerschool__terms_spine --target dev --project-dir src/dbt/kippnewark` gives `PASS=4 WARN=0 ERROR=0`, the table at 1,759 rows (1,166 raw plus 593 quarter), matching the prod census, with both composite uniqueness tests and the unit test passing. Re-ran after the main merge; unchanged.

**The 9 warn-severity test failures in the last fanned-out run are pre-existing kipptaf noise**, not this branch: `rpt_clever__enrollments`, `rpt_gsheets__deanslist_mod_audit`, `rpt_gsheets__csgf_hs_postsec_pathways`, `rpt_tableau__ap_assessment_dashboard`, `rpt_tableau__student_course_grades`, `rpt_tableau__gradebook_gpa`, `rpt_tableau__kfwd_career_launch_survey`, `int_students__graduation_path_codes` and `rpt_appsheet__kfwd_career_launch_survey`. The fan-out surfaced them; the post-merge run selects none of them.

PR #5466, opened 2026-09-21
refactor(dbt): precompute Miami PowerSchool derivations in the powerschool package

### LINKED ISSUE #5413: perf: kipptaf recomputes 4 derivations on frozen Miami PowerSchool rows
## What and why

kipptaf reads Miami's frozen PowerSchool archive through 14 union wrappers. 11 of them are pure unions and add no work. The other 4 apply a join or a calculation to Miami rows that will never change again, and every rebuild redoes that work.

The archive is the `kippmiami_powerschool` source. Miami's SIS moved to Focus, so the archive stops at AY2025 and its rows are final.

| # | Where | What runs on Miami rows | Miami rows |
| --- | --- | --- | --- |
| 1 | `int_students__attendance_daily` | LEFT JOIN of the archive to Focus enrollment stints on 4 predicates, to re-key archive attendance onto the stint that contains each day | 1,285,741 |
| 2 | `stg_powerschool__storedgrades` | `agg_credittype`, a CASE over `credit_type` | 239,545 |
| 3 | `stg_powerschool__storedgrades` | `is_transfer_grade`, from a LEFT JOIN to `int_people__location_crosswalk` | 239,545 |
| 4 | `int_powerschool__gpa_cumulative` | 2 CASE expressions that band `cumulative_y1_gpa_unweighted` and `cumulative_y1_gpa_projected_unweighted` | 3,053 |

Item 1 is the expensive one by a wide margin. `int_students__attendance_daily` is a view, so BigQuery re-runs that 4-predicate join inside each of its 12 consumers on every build.

Items 2, 3 and 4 are cheap. Item 4 costs about 15 slot-minutes a month.

## Proposed approach

Each item needs its own decision. They do not share a fix.

**Item 1: materialize the view, do not touch the archive.** Flipping `int_students__attendance_daily` from a view to a table on a cron schedule stops the join re-running in 12 consumers. A table also stays correct when a Focus stint changes, which a value stored in the frozen archive would not.

**Items 2 and 3: decide whether the split is worth it.** Both derivations sit after the union, so there is no way to precompute only Miami's share without either pushing the CASE into all 4 districts' source models, or splitting the wrapper into a frozen branch and a live branch. Either shape costs more complexity than the compute it saves.

**Item 4: leave it.** The GPA band cut-offs are a network-wide rule documented in `src/dbt/kipptaf/models/students/CLAUDE.md`. Moving them into one district's archive splits a shared definition across 2 places for 15 slot-minutes a month.

<details>
<summary>For Claude</summary>

### Files

- Wrapper list: `rg -Ul 'source\(\s*\n?\s*"kippmiami_powerschool"' src/dbt/kipptaf/models` returns exactly 14, matching the 14 tables declared in `src/dbt/kipptaf/models/powerschool/sources-kippmiami.yml`. The non-`-U` form misses 4 multi-line `source()` calls.
- Item 1: `src/dbt/kipptaf/models/students/intermediate/int_students__attendance_daily.sql` lines 19-60. The `focus_stints` CTE reads `int_students__student_enrollment_union` filtered to `_dbt_source_project = 'kippmiami'`; `powerschool_conformed` LEFT JOINs `int_powerschool__ps_adaadm_daily_ctod` to it on `student_number`, `yearid`, and a half-open `calendardate` range. Refs #4803 and #5114.
- Items 2 and 3: `src/dbt/kipptaf/models/powerschool/staging/stg_powerschool__storedgrades.sql`. This is the only join at the wrapper layer.
- Item 4: `src/dbt/kipptaf/models/powerschool/intermediate/int_powerschool__gpa_cumulative.sql`.

### Measurements

- `int_powerschool__ps_adaadm_daily_ctod` holds 19,157,326 rows: kippnewark 13,915,605 (72.6%), kippcamden 3,612,017 (18.9%), kippmiami 1,285,741 (6.7%), kipppaterson 343,963 (1.8%). Miami's slice is 171 MB and is the largest single Miami contribution to any wrapper.
- `stg_powerschool__storedgrades` holds 1,882,154 rows; Miami is 239,545 (12.7%).
- `is_transfer_grade` is `false` for every Miami row. Counts of `true`: kippnewark 12,232, kippcamden 2,289, kippmiami 0, kipppaterson 0. The crosswalk always matches for Miami, so the value is a constant. The crosswalk is a live model, so it is not frozen by construction.
- `int_powerschool__gpa_cumulative` ran 2,985 builds and 155 slot-minutes over 30 days; Miami is 3,053 of 31,012 rows (9.8%).
- Miami row shares across the other wrappers run 5.9% to 11.6%.

### Limits of the measurement

Miami's share of kipptaf compute is inferred from row share. I did not isolate the Miami scan in a query plan. The shares are good enough to rank these 4 items and are not a savings estimate.

### Out of scope, recorded so nobody re-investigates

- `int_powerschool__calendar_week` derives `region` from `_dbt_source_relation` with `initcap(regexp_extract(...))`. That column only exists after the union, so the derivation cannot move upstream. The model costs 4.3 slot-minutes over 30 days.
- The Miami copies of `base_powerschool__final_grades` and `int_powerschool__final_grades_rollup` hold 0 rows. That is correct. `base_powerschool__final_grades` filters `cc_academic_year = {{ var("current_academic_year") }}`, which is 2026 for kippmiami, and the archive stops at AY2025. Miami's current-year grades come from Focus.

### Acceptance criteria

- `int_students__attendance_daily` no longer re-runs its Focus-stint join once per consumer, and its 12 consumers return the same rows as before the change.
- Items 2, 3 and 4 each carry a written decision: changed, or left alone with the reason.

### Constraints

- Before flipping item 1 to a table, map every consumer's refresh cadence. An intraday consumer vetoes a nightly cron table, and `rpt_tableau__ops_dashboard` is one of the 12. See the view-to-table rules in `.claude/rules/dbt-models.md`.
- The 12 consumers are `fct_student_attendance_enrollment_daily`, `int_powerschool__ada_term`, `int_reporting__promotional_status`, `int_students__fldoe_fte`, `int_topline__ada_running_weekly`, `int_topline__attendance_contacts`, `int_topline__truancy_weekly`, `rpt_tableau__attendance_chronic_absenteeism_log`, `rpt_tableau__attendance_dashboard`, `rpt_tableau__nj_school_register`, `rpt_tableau__okrts_referrals`, and `rpt_tableau__ops_dashboard`.
- `git log -S` the properties yml first. A reverted perf change reads as an obvious win.

</details>

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make the `powerschool` package compute 4 things that `kipptaf` recomputes today on every run.

Miami's PowerSchool data is frozen. The final pull ran on 2026-07-01, and the rows never change again. `kipptaf` still recalculates `academic_year`, 2 calendar flags, and a term-to-cumulative GPA join on top of those frozen rows. That is a design problem before it is a speed problem: data should arrive at `kipptaf` already shaped, and a calculation whose inputs are all frozen belongs in the package that owns them.

This pull request is the package half of a 2-part change. It only adds columns and 1 model, so nothing downstream breaks. The `kipptaf` half follows in a second pull request, once Dagster materializes these models in prod for all 4 districts. That wait is not optional: `dbt_utils.union_relations` reads its column list at compile time, so a new package column does not reach a `kipptaf` wrapper until the district projects rebuild.

The 4 changes:

- `int_powerschool__gpa_term` and `int_powerschool__attendance_streak` each gain `academic_year`, defined as `yearid + 1990`.
- `int_powerschool__calendar_day` gains `is_in_session` and `is_in_membership`.
- A new model, `int_powerschool__gpa`, joins `int_powerschool__gpa_term` to `int_powerschool__gpa_cumulative` so `kipptaf` stops doing that join itself.

1 `kipptaf` change rides along. `stg_powerschool__storedgrades` computed a column called `agg_credittype` that exactly 1 model reads, so the calculation moves into that model, `rpt_gsheets__award_ceremony_gpa`. It moves down to its consumer instead of up into the package, for 2 reasons given under "Reviewer Notes". It has no cross-project dependency, so it does not have to wait for the second pull request.

Refs #5413

## Reviewer Notes

`agg_credittype` moved down, not up. The original design pushed it into the package with the other 3. Two findings reversed that. First, the pushdown has no safe order to ship in: once the package emits the column, the `kipptaf` wrapper's own alias for it duplicates the column and BigQuery fails, but dropping the alias earlier leaves the 1 consumer without the column until the districts rebuild. Second, the calculation is a partial heuristic, not a canonical mapping. BigQuery's `like` is case-sensitive, so the calculation buckets `ENG-T1` and `MATH-T2` but misses `Eng`, `ELA`, `Math`, `MaTH`, `MA`, and `MAT`, which are all live values in `credit_type` today. A half-working heuristic does not belong in a source package that 4 districts inherit.

Do not compare a fresh build of `int_powerschool__gpa_term` against the copy in prod. They legitimately differ, and the difference is not from this change. `int_powerschool__gpa_term` branches on `y1.yearid < var("current_academic_year") - 1990`, and `current_academic_year` is now 2026. Prod's table was built before that rollover, so a fresh build routes `yearid` 35 through a different branch. On `kippnewark` that is a 13,028-row gap. This matters for the second pull request, which has to check its rewrites against a freshly built table, not against prod.

`int_powerschool__gpa` cannot fan out. `int_powerschool__gpa_cumulative` is unique on `(studentid, schoolid)`, which is the join key, so the new model inherits the grain of `int_powerschool__gpa_term` and carries a test for it.

### General

No Dagster changes.

- [ ] Include (or update) an
      summary above.
      See

`int_powerschool__gpa` is new and has no dashboard consumer, so it gets no exposure. The `agg_credittype` removal does drop a column from `stg_powerschool__storedgrades`, which is contracted. Exactly 1 model reads it, and that model changes in the same commit, so no exposure breaks. To roll back, revert commit `e3ab8a23d1`.

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Built with `superpowers:subagent-driven-development` from the spec at `docs/superpowers/specs/2026-09-21-miami-powerschool-pushdown-design.md` and the plan at `docs/superpowers/plans/2026-09-21-miami-powerschool-pushdown.md`, both on this branch. Tasks 1-6 of 12 ship here. Tasks 8-12 are the `kipptaf` half and wait on Dagster materializing these models in all 4 districts. Task 7, a re-bake of the frozen `kippmiami_powerschool` BigQuery archive to add the 15th table, is a prod write and belongs to the user.

The governing design test, from the spec: a derivation is pushable only when every one of its inputs is frozen. A join to live data cannot be frozen and stays in `kipptaf`. `is_transfer_grade` on `stg_powerschool__storedgrades` fails that test — it reads a LEFT JOIN to a live `kipptaf` view, `int_people__location_crosswalk` — so it stays where it is.

`academic_year = yearid + 1990` is verified, not assumed. The 2 expressions agree on all 126,319 non-null-`yearid` rows of `int_powerschool__calendar_day` across all 4 districts, with 0 disagreements.

Verification for each change, run against dev relations in `zz_cbini_kippnewark_powerschool` and re-run independently by a second agent: `int_powerschool__gpa_term` 159,776 rows, `int_powerschool__attendance_streak` 4,739,818 rows, `int_powerschool__calendar_day` 72,649 rows. 0 disagreements and 0 unexpected nulls on every new column. `int_powerschool__gpa` built to 159,776 rows, exactly matching the `int_powerschool__gpa_term` build it joins, and its `unique_combination_of_columns` test on `(studentid, schoolid, yearid, term_name)` passes. That grain is unique in all 4 districts: Newark 172,804, Camden 47,732, Miami 16,512, Paterson 2,548, each equal to its own row count.

Full per-`yearid` breakdown of the prod-versus-fresh gap flagged in Reviewer Notes, on `kippnewark`: `yearid` 12 through 33 identical; `yearid` 34 prod 14,528 against dev 14,484; `yearid` 35 prod 14,712 against dev 1,356; `yearid` 36 prod 15,420 against dev 15,792. 44 + 13,356 − 372 = 13,028, which is exactly 172,804 − 159,776. This is pre-existing date-relative behavior in `int_powerschool__gpa_term` and this pull request does not change it.

`int_powerschool__gpa` uses `unit_tests:` with `format: sql` inputs rather than dictionary fixtures. Dictionary fixtures fail here because dbt introspects the input relation, and on a same-pull-request column add it introspects the old deferred schema, which has no `academic_year`.

2 known cosmetic items, both judged not worth holding the pull request. `students_student_number` sorts to the top of the `columns:` list in `int_powerschool__gpa.yml` though it carries only a `contains_pii` tag and no per-column `data_tests:`, which is what the sorting rule actually keys on; neither source model does this. And `rpt_gsheets__award_ceremony_gpa` has a pre-existing bug this change neither causes nor fixes: its "Stored" branch buckets `credit_type` through the case expression while its "Live" branch reads `fg.credittype` raw, and the 2 branches union into 1 column that is then grouped. The 2 branches use largely disjoint vocabularies, and the raw side carries typos such as `CAREEER`, `CARER`, `PHSYED`, and the literal string `'NULL'`.

</details>


cbini:
@claude Verdict on each finding. Commit `66fd79bc65` carries the one code change.

**1. ST06 column-ordering violation — rejected. Making this change would break the build.**

`grade_source` feeds a positional `union all`, and `union all` binds by ordinal, not by name. Both branches currently produce 10 columns in the same order, with `credittype` at ordinal 7:

| # | Stored branch | Live branch |
| --- | --- | --- |
| 6 | `sg.course_number` | `fg.course_number` |
| 7 | `case ... end as credittype` | `fg.credittype` |
| 8 | `sg.potentialcrhrs` | `fg.courses_credit_hours as potentialcrhrs` |
| 9 | `sg.earnedcrhrs` | `if(...) as earnedcrhrs` |
| 10 | `sg.gpa_points` | `if(...) as gpa_points` |

The suggested fix moves `sg.potentialcrhrs`, `sg.earnedcrhrs`, and `sg.gpa_points` up and puts the `case` last. That shifts the Stored branch to `potentialcrhrs` at 7, `earnedcrhrs` at 8, `gpa_points` at 9, and `credittype` at 10, while the Live branch keeps its current order. `credittype` is `string` and the other 3 are `float64`, so BigQuery rejects the union on a type mismatch. If the types ever lined up it would bind the wrong values silently instead.

The premise is also wrong on the facts. The comment predicts the current code "will likely fail the lint/trunk check on push." It does not. I ran `trunk check --force --no-fix` on the file 3 times across 2 sessions and got `No issues` every time, and the **Trunk Check** job on this pull request passed. sqlfluff skips ST06 near the templated `{{ ref() }}` slice, so the rule never fires here.

This is the same change I caught and reverted while writing the plan. The plan now carries an explicit warning against it, and `66fd79bc65` corrects one detail in that warning: I had written that the misaligned union "compiles clean," and it does not — it fails on the type mismatch.

**2. YAML sort on `students_student_number` — agreed, no action.** Already disclosed in the pull request body. Leaving it.

**3. Missing uniqueness test on `int_powerschool__gpa_term` — accepted and fixed in `66fd79bc65`.** Correct catch. `.claude/rules/dbt-models.md` requires a uniqueness test on every intermediate model and this one had none. I added `dbt_utils.unique_combination_of_columns` on `(studentid, schoolid, yearid, term_name)`. That grain is unique in all 4 districts: Newark 172,804 of 172,804, Camden 47,732 of 47,732, Miami 16,512 of 16,512, Paterson 2,548 of 2,548. The test also passes against a fresh Newark dev build, which carries a different row count than prod for the reason the pull request body describes.

**4, 5, 6, 7 — agreed, no action.** Nothing to add.

One correction to the framing in finding 5. The comment reads the `agg_credittype` decision as "leave it at kipptaf." It is not a no-op: the calculation moved out of `stg_powerschool__storedgrades` and into `rpt_gsheets__award_ceremony_gpa`, its only consumer, and the column no longer exists on the staging model. The conclusion is right, the description of what shipped is not.

PR #3809, opened 2026-05-04
fix(dbt): batch L — join-path FK fixes

cbini:
## Summary & Motivation

When merged, this pull request will resolve the 8 originally-failing `relationships` tests in [PR batch L](https://github.com/orgs/TEAMSchools/projects/4) (join-path FK fixes).

- **`dim_courses`** rerouted from `base_powerschool__course_enrollments` to `stg_powerschool__courses` (canonical course catalog). Hash-compatible via `_dbt_source_relation` token replacement so `dim_course_sections.course_key` FK resolves.
- **`dim_students`** rerouted from `int_extracts__student_enrollments` to `stg_powerschool__students` (canonical, includes graduated/transferred). LEFT JOINs kippadb contact + PS extension tables to preserve `salesforce_contact_id`, `fleid`, `gifted_and_talented`, `lep_status`. Drops `meal_eligibility_status` and `has_iep` (enrollment attributes, not student-grain). `is_ell` simplified to current `studentcorefields.lep_status` flag (drops enrollment-date-window logic — relocate via follow-up).
- **PowerSchool phantom-row filter** — added `where dcid >= 1` at `stg_powerschool__students` to drop 4 PowerSchool placeholder rows (one per district, `dcid=-100`, `student_number=0`, `enroll_status=-100`) that broke `dim_students` PK uniqueness after the reroute.
- **New staging union model** `stg_powerschool__studentcorefields` (kipptaf-level union of the 4 region sources, mirrors `stg_powerschool__u_studentsuserfields` pattern). Required by `dim_students` for `lep_status`.
- **Pre-2000 date sanitization** at 5 staging/intermediate sites — wraps junk dates (year typos like `0001-01-01`, sentinels like `1900-01-01`) with `if(d < '2000-01-01', null, d)` so they don't propagate as orphan FKs against `dim_dates`. Adds 6 `dbt_utils.expression_is_true` regression-guard tests at `severity: warn`.

## AI Assistance

Claude Code authored this PR end-to-end (spec → plan → implementation → review). Human-directed at multiple inflection points:
- Direction to source dims from canonical staging rather than enrollment-derived intermediates.
- Decision to drop `meal_eligibility_status` / `has_iep` from `dim_students` (they're enrollment attributes).
- Choice of `dcid >= 1` as the phantom-row filter predicate.
- Override to use `severity: warn` for date-sanitization regression tests (per spec rationale: bad dates require source-system fixes, shouldn't block CI).

[...]

Issue #3719, opened 2026-04-22
fix(dbt): dim_dates range gaps on assessment/behavioral date FKs


cbini:
## Context

Low-severity gap surfaced by #3713 / PR #3714 (CI run 70403204236701). A handful
of `*_date_key` columns reference dates outside the `dim_dates` range.

## Failing relationships tests (5, 457 rows total)

| Child model | Column | Orphan rows |
|---|---|---:|
| `fct_assessment_scores_enrollment_scoped` | `test_date_key` | 347 |
| `fct_assessment_scores_student_scoped` | `test_date_key` | 100 |
| `dim_school_calendars` | `date_key` | 3 |
| `fct_behavioral_incidents` | `close_date_key` | 3 |
| `fct_behavioral_consequences` | `end_date_key` | 3 |
| `fct_behavioral_consequences` | `start_date_key` | 1 |

## Probable root cause

`dim_dates` range likely doesn't cover the full historical span of source data
(or future dates for open-ended intervals). Simple fix: extend `dim_dates`
generation bounds.

## Acceptance

- [ ] Identify min/max date across all fact `*_date_key` columns
- [ ] Extend `dim_dates` to cover that range
- [ ] All 6 relationships tests clean

PR #5468, opened 2026-09-21
refactor(kipptaf): read the precomputed PowerSchool derivations

### LINKED ISSUE #5413: perf: kipptaf recomputes 4 derivations on frozen Miami PowerSchool rows

cbini:
# Pull Request

## Summary & Motivation

When merged, this pull request will make `kipptaf` read the 4 derivations that the `powerschool` package already computes, instead of recomputing them on every run.

This is the `kipptaf` half of a 2-part change. [#5466](https://github.com/TEAMSchools/teamster/pull/5466) was the package half: it added the columns and 1 new model. Dagster has since materialized those models in prod for all 4 districts, which is what this half was waiting on. `dbt_utils.union_relations` reads its column list at compile time, so a new package column does not reach a `kipptaf` wrapper until the district projects rebuild.

Miami's PowerSchool data is frozen. The final pull ran on 2026-07-01 and the rows never change again. Recalculating a value whose inputs are all frozen is a design problem before it is a speed problem: the data should arrive at `kipptaf` already shaped.

The 4 changes:

- A new wrapper, `int_powerschool__gpa`, unions the 4 districts' copies of the package model.
- `int_students__gpa` reads that wrapper. It no longer joins `int_powerschool__gpa_term` to `int_powerschool__gpa_cumulative` itself, and no longer derives `academic_year` from `yearid`.
- `int_students__calendar_day` reads `is_in_session` and `is_in_membership` off the package model instead of recomputing them, and drops a filter that excluded dates before 2000.
- `int_students__attendance_streak` reads `academic_year` as a plain column instead of deriving it.

1 more change rides along. `int_students__gpa` used BigQuery's `full union all corresponding`. The repo SQL conventions reserve that form for cases the standard form cannot express, and this is not one of them, so the model now uses a positional `union all` with both branches listing the same 24 columns in the same order.

Refs #5413

## Reviewer Notes

There is 1 intentional behavior change, and it is the pre-2000 filter. `int_students__calendar_day` excluded calendar days dated before 2000 to hide 3 placeholder rows dated 1900-01-01 that PowerSchool carries. Hiding them in `kipptaf` was the wrong place for the fix. The plan expects the model to gain exactly 3 rows and `dim_school_calendars.date_key` to gain 3 new warns on its relationships test. I have not confirmed those 2 numbers yet — they need a built PR schema, so I will run the queries once dbt Cloud CI finishes and post the results as a comment. The 3 placeholder days are a school operations data-entry item and are tracked in Asana.

The union conversion is not cosmetic. sqlfluff cannot parse `full union all corresponding`, so it silently skipped its rules inside both branches of that model. Removing the clause exposed a real ST06 violation that main passes today: `g.syear - 1990 as yearid` sat ahead of the `cast(...)` targets in the Focus branch. That is fixed here. The same construct is still used by 8 other `int_students__*` models on main. Sweeping those is a separate change, not this one.

`int_students__gpa` reorders the columns inside its PowerSchool branch. That is safe because the final select lists every column explicitly in 1 order for both branches, so the union binds by position against a list this pull request controls.

### General

No Dagster changes.

- [ ] Include (or update) an
      summary above.
      See

`int_powerschool__gpa` is a new union wrapper with no dashboard consumer, so it gets no exposure. No contracted model loses a column here, so no exposure breaks. To roll back, revert the 5 commits on this branch.

## CI checks

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Built with `superpowers:subagent-driven-development` from the spec at `docs/superpowers/specs/2026-09-21-miami-powerschool-pushdown-design.md` and the plan at `docs/superpowers/plans/2026-09-21-miami-powerschool-pushdown.md`, both landed on the [#5466](https://github.com/TEAMSchools/teamster/pull/5466) branch. Tasks 8 through 12 of 12 ship here. Task 7, the re-bake of the frozen `kippmiami_powerschool` BigQuery archive, shipped as [#5467](https://github.com/TEAMSchools/teamster/pull/5467).

The governing design test, from the spec: a derivation is pushable only when every one of its inputs is frozen. A join to live data cannot be frozen and stays in `kipptaf`. That is why `int_students__attendance_daily` and `is_transfer_grade` are untouched here — both reach live `kipptaf` relations.

The shared `zz_stg_*` staging copies had to be refreshed before this pull request could pass CI, and were. Before the refresh, `INFORMATION_SCHEMA` showed `int_powerschool__gpa` absent from all 3 New Jersey staging datasets, `int_powerschool__calendar_day` missing `is_in_session` and `is_in_membership` in all 3, and `int_powerschool__attendance_streak` missing `academic_year` in all 3. A staging compile of the new wrapper emitted 63 `cast(null ...)` bindings, 21 for each of the 3 districts. After running `dbt clone --select int_powerschool__gpa int_powerschool__calendar_day int_powerschool__attendance_streak --target staging --full-refresh` against each district with a freshly parsed prod state manifest, all 9 relations carry their columns: `int_powerschool__gpa` at 21 columns, `int_powerschool__calendar_day` at 32 with all 3 target columns, `int_powerschool__attendance_streak` at 10 with `academic_year`. The wrapper's staging compile now emits 0 `cast(null ...)` bindings. The prod state manifests on disk were from Sep 10 and predated [#5466](https://github.com/TEAMSchools/teamster/pull/5466), so `dbt parse --target prod --target-path target/prod` had to regenerate them first, and `kipppaterson` needed `dbt deps` before it would parse.

What is verified locally and what is not. Verified: every touched model compiles clean under `--target staging`; `trunk check --force --no-fix` reports no issues on all touched files; the new wrapper's compiled SQL null-fills nothing. Not yet verified, because each needs a built PR schema: the row-count delta of 3 on `int_students__calendar_day`, 0 differing boolean values on `is_in_session` and `is_in_membership`, 0 differing rows on `int_students__gpa`, and the 3 new `dim_school_calendars.date_key` warns. Those queries run after dbt Cloud CI builds, and the results go in a comment on this pull request.

`int_powerschool__gpa_term` carries a trap for anyone re-checking this work. It branches on `y1.yearid < var("current_academic_year") - 1990`, and `current_academic_year` rolled to 2026 on 2026-07-17. Prod's table was built before that rollover, so a fresh build routes `yearid` 35 down a different branch. Do not compare a fresh build against prod for that model; compare against a freshly built table. The frozen Miami archive is unaffected — its `gpa_term` was last built on 2026-09-10, after the rollover, and the Task 7 re-bake returned the same 16,512 rows.

A stale cross-reference was cleared along the way. `dim_school_calendars.yml` carried a note on `date_key` naming PR #3809 and PR #3719 and a row count that no longer holds. Transient detail does not belong in a description, so the note is gone and nothing replaced it.

</details>


cbini:
@claude Review verdicts and the verification results the PR body promised. All 3 findings accepted and fixed in `67fe76df7b`; a live prod bug the verification uncovered is fixed in `ea161fdccc`.

## Review verdicts

**Finding 1 — `int_powerschool__gpa.yml` has no `columns:` block. ACCEPTED, fixed.** Correct, and the sibling comparison is the right test. All 23 columns now carry a `data_type` and a description. `students_student_number` also carries its own `contains_pii` tag alongside the model-level one.

**Finding 2 — stale `academic_year` description on `int_students__attendance_streak.yml`. ACCEPTED, fixed.** The description now says the value is read from `int_powerschool__attendance_streak`, where the powerschool package derives it from `yearid`.

**Finding 3 — `int_powerschool__calendar_day.yml` and `int_powerschool__attendance_streak.yml` omit the columns this PR reads. ACCEPTED, fixed, and it was more than documentation on one of the two.** `is_in_session` and `is_in_membership` are now listed on `calendar_day`, and `academic_year` on `attendance_streak`. The call that these files were already under-documenting before this PR is right, but the `attendance_streak` gap turned out to be load-bearing. See below.

Findings 4 through 10 were no-violation checks and need no response.

## The bug finding 3 was sitting next to

`kipptaf_powerschool.int_powerschool__attendance_streak` has been serving `academic_year` as NULL in prod for every New Jersey row: Newark 4,753,288 + Camden 1,397,035 + Paterson 68,833 = 6,219,156 rows. Miami's 469,176 rows are 0% null.

`dbt_utils.union_relations` resolves its column list at compile time. This view last compiled before the 3 NJ projects gained `academic_year`, so it baked in `cast(null as INT64)` for them. Miami escaped because its frozen archive is a BigQuery-native source that already carried the column when the view compiled. The same nulls were present in `zz_stg_kipptaf_powerschool`, so this was not a staging artifact.

Nothing downstream exposed it, because `int_students__attendance_streak` derived `yearid + 1990` itself. This branch is the first consumer of the wrapper's column, which is why it surfaced now.

`dbt_union_relations_automation_condition()` does not self-heal this. It keys on recursive ancestor `code_version_changed`, and the wrapper's upstream is a `source()` in a different Dagster code location, so a district-side column add is invisible to it.

The fix is a durable comment in the `.sql` — that marks the model `state:modified` for CI and changes its `code_version` for the Dagster prod deploy, which a properties-yml-only change does not do — plus the missing column in the properties file. After the CI rebuild the wrapper reads 0 nulls in all 4 districts.

`int_powerschool__calendar_day` was checked for the same defect and is healthy: 0 nulls on `is_in_session` and `is_in_membership` in all 4 districts.

## Verification results

**`int_students__calendar_day` — PASS.** PR 128,194 rows against prod 128,191. Delta 3, only-in-PR 3, only-in-prod 0. All 3 extra rows are `date_value = 1900-01-01`. Every one of the 128,191 prod rows survives byte-identical, which is the proof that the package's `is_in_session` and `is_in_membership` equal the old kipptaf recomputation.

**`dim_school_calendars.date_key` — PASS, exactly 3 new warns.** 3 orphan rows against `dim_dates`, all on `1900-01-01`, 1 distinct date. Matches the prediction.

**`int_students__attendance_streak` — PASS after the fix.** Before: identical row counts but 6,219,150 differing rows, isolated to `academic_year` alone (`diff_without_academic_year` was 0 across all 6.7M rows). After: only-in-PR 0, only-in-prod 10. Those 10 are AY2026 in-progress streaks; prod's table was last built 2026-09-09, 12 days before the PR build, so they are source churn in the open year rather than a logic gap.

**`int_students__gpa` — PASS, but prod is the wrong baseline and a naive diff misleads.** A direct PR-vs-prod comparison reports 30,364 and 30,348 differing rows spread across most GPA columns. That is not a refactor defect. `kipptaf_students.int_students__gpa` is a table last built at 2026-09-21T08:04:45Z, while the district `int_powerschool__gpa` rebuilt at 2026-09-21T21:46:24Z. About 13.7 hours of teacher grade entry sits between them, and the row deltas line up with it: Newark +12, Camden +4, Paterson 0, Miami 0.

So compare the two logics over the same inputs instead. Replicating the pre-refactor kipptaf join — `int_powerschool__gpa_term` left join `int_powerschool__gpa_cumulative` on `studentid` and `schoolid`, plus `yearid + 1990 as academic_year` — against the new district `int_powerschool__gpa`, both read from the same district prod build, over 21 columns with a symmetric `except distinct`:

| district | rows | only_in_new | only_in_old |
| --- | ---: | ---: | ---: |
| kippnewark | 172,804 | 0 | 0 |
| kippcamden | 47,732 | 0 | 0 |
| kipppaterson | 2,548 | 0 | 0 |
| kippmiami | 16,512 | 0 | 0 |

239,596 rows, byte-identical. The pushdown preserves every value.

One trap for anyone re-running that comparison: `_dbt_source_relation` differs by construction on both sides, because the union now names `int_powerschool__gpa` rather than `int_powerschool__gpa_term`, and staging names `zz_stg_*`. Exclude it and join on `_dbt_source_project`.

PR #5469, opened 2026-09-21
fix(dbt): serve Miami's pre-cutover terms from the PowerSchool archive

### LINKED ISSUE #5397: fix: int_students__terms has no Miami quarters for 2 schools before AY2025

cbini:
# Pull Request

## Summary & Motivation

"When merged, this pull request will..." give Miami a school quarter on every year before its 2026 switch to Focus.

Miami ran PowerSchool until July 2026, then moved to Focus. `int_students__terms` only ever read Focus for Miami, so Miami had no quarters in the older years. That left 492,482 Miami membership days in `int_students__attendance_daily` with no term and no semester. Anything that groups Miami attendance by quarter dropped those days.

Miami's PowerSchool data did not go away. We froze it as an archive when the SIS moved. This change reads that archive. A new model, `int_powerschool__terms_spine`, stacks the terms from all four regions, Miami included. `int_students__terms` reads it, and the Focus side now starts at 2026, the year of the switch. PowerSchool covers every year before the switch; Focus covers every year from the switch on. The two never overlap.

This also removes two older models that nothing reads any more. `int_powerschool__terms` is deleted. `stg_powerschool__terms` is turned off but its file stays, because it holds logic that moved elsewhere rather than nothing at all.

Nothing about the output changes shape. `int_students__terms` returns the same 35 columns in the same order, so its 8 consumers need no edit.

Refs #5397

## Reviewer Notes

The old model joined two sets of rows with a `full join`. This change stacks them with `union all` instead. Worth a look, because it reads like a behavior change and is not one. See "For Claude" for the numbers.

The Focus side now starts at the plain number 2026 rather than reading a model. That was deliberate, and "For Claude" says why.

The new spine model carries no uniqueness test. That matches the model it replaces, and it is a choice rather than an oversight. Push back if you disagree.

Miami's PowerSchool archive was rebuilt for this change, partially, outside a normal Dagster run. There is no Dagster asset for it and there cannot be one. "For Claude" has the detail.

### General

Skipped. No Dagster changes.

      code location

- [ ] Include (or update) an
      summary above.
      See

No contracted model loses a column here. `int_students__terms` keeps all 35 of its columns, in the same order. Two models are retired, and neither has a remaining reader.

Skipped. No schedule, sensor, or integration changes.

      `CLAUDE.md`

## CI checks

## Troubleshooting

<details>
<summary>For Claude</summary>

Claude-assisted, human-directed. Executed under `superpowers:subagent-driven-development` from the plan at `docs/superpowers/plans/2026-09-21-miami-archive-terms.md`, which argues from the spec at `docs/superpowers/specs/2026-09-21-miami-archive-terms-design.md`. This is PR B of two; PR A was #5465, which added the package-level `int_powerschool__terms_spine` and merged as `c9dffc4a0d`.

**Why the `full join` is gone rather than rewired.** The old `int_students__terms` joined `powerschool_quarters` to the raw terms rows with a `full join` and reconciled with `coalesce(p.x, q.x)`. Measured on prod against both shapes, that join produced exactly its two inputs stacked: 926 quarter rows and 1,925 raw rows, with a symmetric difference of 0 over every compared column. No consumer reads a merged row. So `union all` is not a simplification of the join's behavior, it IS the join's behavior, minus the reconciliation machinery.

**Why the Focus floor is the literal `2026`.** An earlier draft read it from an `int_students__sis_cutover` model. That was dropped on the user's call: the cutover already happened, so 2026 is a historical fact about this warehouse and not a moving parameter. The concrete hazard the literal guards against is a Focus backfill. Focus carries a full year/semester/quarter set for 2 Miami schools in every `syear` back to 1980, so a floor that tracks anything other than the cutover year would let Focus fabricate pre-cutover history on top of the archive. Both the `type in ('year', 'semester', 'quarter')` filter and the year floor stay in `int_students__terms` rather than moving to staging, because 321 report card grade rows point at pre-2018 marking periods and flooring the staging model orphans them.

**Why the kipptaf spine wrapper carries no uniqueness test and no column docs.** It mirrors the sibling `int_powerschool__terms` wrapper it replaces, and the repo treats a pure `union_relations` passthrough that way. The grain guarantee lives in two places that do run: two composite uniqueness tests on the package-level spine (registered as blocking asset checks per district, verified `severity: ERROR` with `failed_row_count` 0 in prod), and the two `where`-scoped uniqueness tests on `int_students__terms`. The `materialized: table` override in the wrapper's properties file IS load-bearing — kipptaf's `dbt_project.yml` `powerschool:` block sets only `+schema` for `intermediate/`, not `+materialized`.

**Miami's archive rebuild.** `kippmiami`'s `packages.yml` does not include the `powerschool` package; it is re-included temporarily for each archive rebuild and removed after. So `kippmiami` has no Dagster asset for any powerschool model and never will, and a local `--target prod` build is the only route to write into `kippmiami_powerschool`. The spine was built there that way, with user authorization, selecting `int_powerschool__terms_spine` alone so none of the 16 post-hooks on the staging models fired and the build issued no DELETE. Result: `PASS=4 WARN=0 ERROR=0`, 281 rows (193 raw + 88 quarter), `academic_year` 2017 through 2025. The temporary package re-include was fully reverted and `dbt parse --target prod` still succeeds; nothing from it is committed here.

**Prod spine row counts, all four regions, each decomposing exactly into raw + quarter:** kippnewark 1759 (1166 + 593), kippcamden 1019 (702 + 317), kipppaterson 73 (57 + 16), kippmiami 281 (193 + 88). Distinct `(schoolid, yearid, abbreviation)` keys are fewer than total rows by design — the raw and quarter halves can share an `abbreviation`, which is why the two tests are scoped `where term is not null` and `where term is null` rather than combined.

**CI seeding.** kipptaf `sources-kipp*` resolve to the `zz_stg_*` staging copies under `target=staging`, which is what dbt Cloud CI reads, and PR A's merge materialized the spine in each district's prod dataset without creating the staging copy. All three NJ copies were seeded with `dbt clone --select int_powerschool__terms_spine --target staging --full-refresh`, with user authorization, one district per call. Verified: `zz_stg_kippnewark_powerschool`, `zz_stg_kippcamden_powerschool` and `zz_stg_kipppaterson_powerschool` each hold the relation as a `CLONE`, and `kippmiami_powerschool` holds it as a `BASE TABLE`. Miami needed no clone — `sources-kippmiami.yml` is declared BQ-native with a plain schema and no target branch, so every target reads the prod archive directly.

**Reading the diff.** Git renders the wrapper swap as a rename (`R054`, `int_powerschool__terms.sql` to `int_powerschool__terms_spine.sql`) in the three-dot branch diff, because the new wrapper is the old one with Miami added. That is diff presentation only; the individual commits hold a real `git rm` and a real new file.

**Retirement treatment differs between the two old models, deliberately.** `int_powerschool__terms` is deleted outright, source entries included, under the #5162 exception in `.claude/rules/dbt-models.md` — it was a bare `select *` union passthrough (`union_relations` plus `extract_source_project`, nothing else) with no remaining `ref()`, and the district relations it read stay in place. `stg_powerschool__terms` is disabled instead, because it carries the `row_number()` dedup window that moved into the package spine, so it is not a bare passthrough and the default retire-is-a-disable rule applies. Its `.sql` stays on disk and the prod relation is untouched. Neither has tests to disable.

**Nine stale `config.meta.source_model` pointers** in `int_extracts__course_enrollments_by_term.yml` and `int_extracts__course_schedule_by_term.yml` named `int_powerschool__terms`, which this branch deletes. They were repointed at `int_powerschool__terms_spine`. dbt does not resolve `config.meta.source_model`, so this is documentation lineage only and nothing built or parsed differently before the fix.

**Verification run on this branch.** `trunk check --force --no-fix` over all 12 changed files: no new issues (the one existing issue is a pre-existing `trunk-ignore(sqlfluff/AM04)` no-op on `stg_powerschool__terms.sql`, present before this branch). `dbt parse --no-partial-parse --target staging --project-dir src/dbt/kipptaf`: clean, with `stg_powerschool__terms` under `disabled` and neither retired model in `nodes`. `dbt compile --select int_students__terms --target staging`: both union branches project 35 positionally-aligned columns, matching prod `INFORMATION_SCHEMA.COLUMNS` for `kipptaf_students.int_students__terms` name, type and position. `origin/main` was merged in before pushing, per the #4774 branch-staleness fan-out.

**Still open after this PR**, tracked separately: 511 residual Miami membership days that this change does not recover. Those need a correction on the PowerSchool server plus a final sync, which in turn needs Miami's decommissioned PowerSchool Dagster assets revived via dlt. That is a separate issue, not a gap in this one.

</details>


cbini:
## Pre-review pass: 10 findings, 8 fixed, 2 declined

I ran a whole-branch review before asking for one here. It found no correctness defect — every finding was a documentation or comment claim that was true of the old full-join shape and false under the union. Commit `baec9688` carries the fixes. All changes are prose: no SQL logic, tests, or model config changed.

### Fixed

| # | Where | What was wrong |
|---|---|---|
| 1 | `int_students__terms.yml` description | Claimed the two arms are "disjoint by year" without qualification. The 2026 floor is on the Focus arm alone, so all three NJ districts stay on the PowerSchool arm in every year. A reader trusting the old sentence could floor the PowerSchool arm too and delete every NJ AY2026 term. |
| 2 | same description | Said a quarter appears as two rows, unconditionally. A quarter that termbins carries but the raw terms table never had appears as one row. |
| 3 | same file, test comment | Said "quarter rows key on term". Spine rows key on term; everything with no term, including the raw Q1-Q4 records, keys on abbreviation. |
| 4 | same file, columns | `name`, `firstday` and `lastday` are null on spine rows and did not say so, while six sibling columns did. |
| 5 | `int_students__terms.sql` | The Focus-floor comment justified itself with "2 schools" and "321 report card rows … pre-2018" — both measured against the old pre-2018 floor. |
| 6 | `stg_powerschool__terms.sql` | The `row_number()` comment described guarding a quarter-grain join that no longer exists. |
| 7 | `rpt_illuminate__terms.sql` | The comment said the `name is not null` filter excludes synthetic rows for a handful of historical schoolids. Under the union it drops the entire spine arm. The filter is unchanged and the extract still returns 1,925 rows; only the comment was wrong. |
| 8 | `docs/models/dibels-dashboard-data-model.md` | Charted `stg_powerschool__terms` as a live upstream of `int_extracts__student_enrollments`. That model is now disabled, and `int_extracts__student_enrollments` refs no terms model at all (I checked all 16 of its refs). Removed the node, its two edges and classDefs, and corrected the layer counts. |

### Declined

**The `stg_powerschool__terms` source entries in `sources-kippnewark.yml`, `sources-kippcamden.yml` and `sources-kipppaterson.yml` stay.** The review asked to strip them since only the disabled model reads them. `.claude/rules/dbt-models.md` makes retiring a model a disable rather than a delete precisely so it stays reversible; removing its sources would break re-enabling it. Three unused source declarations have no build effect.

**The three `int_powerschool__terms` references in `docs/models/gradebook-audit-data-model.md` stay.** All three sit after line 838, `## Legacy: AY 2025-2026 and superseded`, whose next lines say the section is "kept collapsed for history; none of it describes the current pipeline." In that pipeline the model really was named `int_powerschool__terms`. Renaming it would falsify history the page deliberately preserves. A fix agent did make these edits before I caught the section boundary; they are reverted and the page is byte-identical to its state at `f86b1a4c`.

### Verification on `baec9688`

- `dbt parse --no-partial-parse --target staging` — exit 0, no warnings
- `trunk check --force --no-fix` on all 6 touched files — no new issues, 1 pre-existing
- The three NJ source files still carry both `stg_powerschool__terms` entries each

Earlier CI on `f86b1a4c` was fully green, including dbt Cloud run 70403225737963 with 9 pre-existing test warnings (two confirmed identical to prod).

cbini:

Context that will save you time:

The branch rewrites `int_students__terms` from a full join into a union of two arms — a PowerSchool arm (a spine row from `int_powerschool__terms_spine` carrying `term` and the term dates, plus a raw-terms row carrying `name`, `abbreviation`, `firstday`, `lastday`) and a Focus arm for Miami only, floored at `syear >= 2026`, one row per marking period. The goal is issue #5397: Miami had no quarters before AY2025, leaving 492,482 Miami membership days in `int_students__attendance_daily` with a null `term` and `semester`.

Two things are worth your attention more than the rest:

1. **The row split changes the grain.** Every spine row now carries a null `name`, so any consumer filtering on `name is not null` (`rpt_illuminate__terms` does) now drops the whole spine arm rather than the ~86 orphan quarters it used to drop. I verified that extract still returns 1,925 rows, and checked all 9 consumers, but a second read on the consumer analysis is the highest-value thing you can do here.

2. **The 2026 floor is on the Focus arm only.** The three NJ districts stay on the PowerSchool arm in every year including the cutover. 2026 is the literal SIS cutover year and a permanent feature of the warehouse, not a temporary guard.

This PR had a full pre-review pass already — see the comment directly above for its 10 findings, the 8 fixes, and the 2 declines with reasons. Please do not re-litigate the two declines unless you think the reasoning is actually wrong. Fresh eyes on the SQL and on the consumer impact are what I want.

cbini:
Verdicts on the review, one per item. Commit `1634882d`.

### `int_students__sis_cutover` duplication — partly accepted, code unchanged, both descriptions fixed

The literal stays, and the reason is a mechanism the review had backwards.

`int_students__sis_cutover` is `min(academic_year) from int_focus__attendance_daily where is_attendance_recorded`. The review argues it is "a `min()` floor over *recorded* attendance, not an unbounded backfill-prone value." A `min()` over recorded attendance is exactly backfill-prone in the direction that matters here: if Focus ever records real attendance for AY2025 or earlier, the value drops below 2026. That is not hypothetical — Miami's Focus instance gaining historical attendance is a live possibility, and it is the scenario the spec named.

For the eight siblings that is the correct behavior. If Focus gains earlier recorded attendance, the attendance and calendar conforms *should* move their boundary so Focus serves those years. For `int_students__terms` it is the wrong behavior, and it is wrong in the direction of the defect this whole branch exists to fix. The hazard here is not attendance coverage. It is that Focus's `marking_periods` carries a full year/semester/quarter set back to 1980 for schools with zero enrollment behind them — schoolid 30200805 has a 7-row Focus term set in every year from AY2018 with no attendance and no enrollment. Lowering the terms floor because attendance got backfilled would readmit exactly those fabricated rows.

So the two boundaries are the same year today and are not the same fact. Coupling them would make a Focus attendance backfill silently fabricate term history. The approved spec reached this conclusion already (§4, "The literal is deliberate"), and 2026 is a historical fact rather than a parameter — the cutover has happened.

What the review got right is the documentation gap, and it was worse than reported. `int_students__sis_cutover.yml` claimed *"Every `int_students__*` conform scopes both sides of its Miami union on it"*. This branch made that claim false, which puts it in the same class as the eight already fixed. Both descriptions now carry the truth: the cutover model names the one exception rather than claiming there are none, and `int_students__terms` explains why it takes no dependency.

For the record, I checked whether the suggested change was even legal before declining it: `int_focus__attendance_daily` reads only `source("kippmiami_focus", ...)`, so the cross join would introduce no cycle. The decline is on the semantics, not a constraint.

### `int_extracts__student_enrollments_subjects` mtss join — checked, resolves

The review flagged this as unverifiable from its session. Checked against prod:

| district | `schoolid = 0` rows | with `firstday` and `lastday` | distinct yearids |
|---|---|---|---|
| kippmiami | 32 | 32 | 8 |
| kippnewark | 64 | 64 | 22 |

Miami's archived raw terms carries district-level `schoolid = 0` rows with both date columns populated, the same shape Newark relies on. The `mtss` CTE resolves for Miami.

Also confirmed the arms meet cleanly: `kippmiami_powerschool.int_powerschool__terms_spine` holds 281 rows, 88 of them quarter rows, spanning academic years 2017 to 2025. The archive tops out at 2025 and the Focus floor is 2026 — no gap, no overlap. 88 matches what the plan predicted.

### The rest — no action needed

The confirmations on the `name is not null` filter, the Focus-only floor, the 35-column union alignment, the `stg_powerschool__terms` disable-not-delete call, ST06, and the ~10 consumer call sites all match my own analysis. The independent re-derivation of 1,925 as 1166 + 702 + 57 is a better check than the one I ran, which compared totals rather than per-district components.

### Pre-merge state

All 6 checks were green on `baec9688`. `1634882d` touches two `.yml` descriptions only — no SQL logic, tests, or model config — and re-runs `dbt parse` clean with `trunk check` reporting no issues on both files.

src/dbt/kipptaf/models/students/intermediate/properties/int_students__sis_cutover.yml
cbini, line 1: retire this model completely in this PR

src/dbt/kipptaf/models/powerschool/staging/properties/stg_powerschool__terms.yml
cbini, line 1: unused union models should be deleted, not only disabled

src/dbt/kipptaf/models/students/intermediate/properties/int_students__sis_cutover.yml
cbini: Done in 1e6267f180. All eight conforms that cross joined it now state `>= 2026` directly, along with the `int_students__calendar_day` singular test and six properties files. The derived value is 2026 in prod today, so every predicate is equivalent and no row set moves.

I disabled it rather than deleting it, per `.claude/rules/dbt-models.md` ("Retiring a model is always a disable, never a delete"), with its `not_null` test disabled alongside. Your other comment carved out deletion specifically for union wrappers and this isn't one, so I read "retire" as the rule's meaning — say the word and the files go too.

`dbt parse --no-partial-parse --target staging` is clean, which is also the proof that no `ref()` to it survives.

One consequence worth flagging: nine independent literals can drift apart where a shared model moved them together. The uniqueness tripwires on `int_students__category_grades` and `int_students__gradebook_assignments_scores` are what catch that, so I rewrote their comments to say so instead of describing a floor that drifts on its own.

src/dbt/kipptaf/models/powerschool/staging/properties/stg_powerschool__terms.yml
cbini: Deleted in 749c886dbb — model, properties, and all three NJ source entries.

I'd declined this in the pre-push review, reasoning that a disable should stay reversible and stripping the sources would block re-enabling. Your rule is the better one, and it matches the #5162 carve-out already written into `.claude/rules/dbt-models.md` for a kipptaf union passthrough with no remaining `ref()` — I read that carve-out too narrowly because this model carried a `row_number()` window.

Worth noting the powerschool package's own `stg_powerschool__terms` is a different model with fourteen live consumers. Untouched.

src/dbt/kipptaf/models/students/intermediate/properties/int_students__sis_cutover.yml
cbini: Deleted outright in dd8de546d3 — both files gone.

Recording the reasoning, since it is an exception to `.claude/rules/dbt-models.md`: that rule disables rather than deletes so an unknown consumer degrades to frozen data instead of vanishing. Your point is that the premise does not hold for this model — it is a recent addition, nothing is preserved by keeping it, and it will not be re-enabled. Noted on the commit.

For the record on the preceding commit: dbt Cloud CI passed green on 1e6267f180, so the eight rewritten conforms and their whole descendant graph build correctly with the literal.

cbini:
Verdicts on the second review pass, reviewing `dd8de546d3`. One finding actionable, declined; the rest confirmations with no action.

### Finding 4, no unit test on the union-vs-join grain — declined

Both factual claims verified before deciding. `int_students__terms.yml` carries no `unit_tests:` block, and the package-level `int_powerschool__terms_spine.yml` does ship `unit_terms_spine_stacks_without_joining` at line 174.

Declined because the regression it names is already covered, in three places that do run:

The union-not-join property itself is locked upstream. `unit_terms_spine_stacks_without_joining` exists on the package spine precisely to catch a union silently regressing to a join, and the PowerSchool arm of `int_students__terms` is that spine. A kipptaf-level test would re-assert at the consumer what the producer already guarantees.

What is genuinely unguarded at this layer is arm-stacking plus the 2026 floor, and a silent row drop there fails the two `where`-scoped `unique_combination_of_columns` tests on this model, or surfaces as one Miami academic year emitting from both arms — which is exactly what the uniqueness tripwires on `int_students__category_grades` and `int_students__gradebook_assignments_scores` are there to catch. Those tripwires got more load-bearing in this branch, not less, and their comments were rewritten to say so.

Cost against that: roughly 150 to 250 lines of fixture, given four refs and 35 output columns against the sibling's two inputs and 33. On a PR that is green, already four rounds deep, and waiting on review, that is a poor trade for a guard the reviewer itself rates nice-to-have. Recorded here rather than filed, so a future reader sees it was weighed and why.

### Finding 3, the cutover duplication — superseded, and resolved opposite to the earlier round

Worth flagging for anyone reading the two review rounds together. The first pass proposed centralizing `int_students__terms` onto `int_students__sis_cutover`. That was declined on the grounds that a `min()` over recorded attendance is backfill-prone in the one direction that would readmit Focus's fabricated pre-1980-forward term rows.

The resolution went the other way entirely: rather than one model reading the literal and eight reading the derived value, all nine now state `2026`, and `int_students__sis_cutover` is deleted. So there is one source of truth, which is what the first pass wanted, reached by removing the model rather than by adopting it. The tradeoff is that nine literals can drift where one model moved them together, which is what the tripwires above cover.

### Findings 1, 2, 5 through 11 — confirmed, no action

The `name is not null` ordering, the Focus-only floor, the `stg_powerschool__terms` deletion as an adjudicated call, the load-bearing `materialized: table` override, zero dangling references, the 35-column contract, ST06 on the eight rewritten conforms, altitude, and the ponytail pass all match my own analysis. Finding 7's independent grep for surviving references is the one I would have most wanted a second pair of eyes on, and it agrees.

### State

All checks green on `dd8de546d3`, including dbt Cloud. No code changes from this round.

